PR #44, opened 2021-02-12
Remove testing and storage dependencies

ljedrz:
This is the most straightforward way to remove the `storage` crate which is already a part of `snarkos`: remove the affected `snarkvm_dpc` tests and put them in `snarkos` instead.

The alternative solution would be to keep storage (or at least the main related traits) in `snarkvm` and depend on it in `snarkos`; some in-memory test storage solution could be used for test purposes.


ljedrz:
Closing the draft in order to concentrate on introducing the alternative solution.

PR #47, opened 2021-02-12
Deduplicate some common code under models

ljedrz:
This PR's content is the same as https://github.com/AleoHQ/snarkOS/pull/543; the split between the crates occurred before that one went in.


howardwu:
@collinc97 let's take a look at this change

collinc97 (approved):

howardwu (approved):

PR #48, opened 2021-02-12
ci(gha): add cargo test, simplify clippy

ljedrz:
It seems that this repo doesn't utilize CircleCI outside of merges (at least not yet), so a `cargo test` run is missing; perhaps GHA can handle it now that the repo is much smaller?


ljedrz:
It works, but is very slow, so probably not the best way to go forward, at least not until `storage` and `testing` are removed.

PR #53, opened 2021-02-18
Make the Ledger generic over its storage

ljedrz:
This PR accomplishes 2 main goals:
- allows the `Ledger` to use different storage options than `rocksdb`, especially in-memory ones
- removes the dependency on `rocksdb`, making compilation a lot faster

In addition, as long as the `dpc` tests dependent on storage can be moved to (or already have their counterparts in) snarkOS, both the `storage` and `testing` crates can be removed from snarkVM (as drafted in https://github.com/AleoHQ/snarkVM/pull/44). SnarkOS could then be the repo where `MemDb` and `RocksDb` storages are implemented, and snarkVM would only be the place where the `Storage` trait is implemented, plus some minor helper objects.

Some of the changes in this PR (the `storage` ones, with the exception of `catch_up_secondary`, which was removed here for simplicity) reflect the ones that would be necessary to implement on snarkOS side; I've already tested these changes with an adjusted snarkOS and they work fine, same as snarkVM tests with an in-memory storage, which are enabled in this draft.

The PR is a draft only due to the fact that the decision regarding `storage` and `testing` crates needs to be made (whether they should be fully removed from snarkVM, together with the associated `dpc` tests), and to demonstrate some of the changes that would follow on snarkOS side.


ljedrz:
I've pushed the removal of `storage` and `testing` crates as the last commit so that these changes can be merged sooner if desirable. The follow-up snarkOS-side changes can be seen at [my prototype branch](https://github.com/AleoHQ/snarkOS/compare/staging...ljedrz:decoupled_storage), where the storage used is dependent on the `snarkos-storage` crate feature, with `rocksdb` being the default backend. A simple in-memory storage is introduced there primarily for test purposes (though the tests are still run with `rocksdb` by default), but can later be extended to other use-cases as well.

PR #74, opened 2021-02-28
Make uint and int implementations more symmetric

Protryon:
## Motivation

Clean up uint implementation to be symmetric with int implementation in preparation for type casting


howardwu:
Will merge after I have a chance to add in some hardcoded transaction proofs from `testnet1` for sanity tests, to make sure these gadget changes do not change the `base_dpc`'s circuits.

On a separate note, this further pushes forward our need to begin designing a serialized program design, so we can have a compiled & versioned circuit available for use, instead of dynamically compiling the circuit every time we use it.

ljedrz:
I seem to have missed this PR when working on https://github.com/AleoHQ/snarkVM/pull/104 and https://github.com/AleoHQ/snarkVM/pull/108 (which builds on top of it) :sweat_smile:; however, I see that there's plenty of overlap in ideas (`Number` is very close to `Integral`) and there are some bits in https://github.com/AleoHQ/snarkVM/pull/108 that push code deduplication even further.

Feel free to merge/cherry-pick/rebase any bits of my implementation to further this goal :+1:.

collinc97:
Closing this because #108 was merged in.

PR #97, opened 2021-03-08
Isolate a Storage trait

ljedrz:
This is the least breaking way to introduce this change I could think of: isolate the trait in snarkVM so snarkOS can use it and adjust the necessary bits in the current `snarkvm-storage` crate to it. Once snarkOS is updated, snarkVM will be able to depend on `snarkos-testing` in order to run DPC tests, at which point `snarkvm-storage` and `snarkvm-testing` can be removed.

Filing this as a draft for now, so I can first make sure snarkOS can interact with it nicely once again.


objects/src/traits/storage.rs
```diff
@@ -0,0 +1,82 @@
...
+
+use crate::errors::StorageError;
+
+use std::{fmt::Debug, path::Path};
+
+pub trait Storage
+where
+    Self: Sized,
+{
+    /// Returns a `bool` indicating whether the storage is in-memory only
```
niklaslong, line 25:
```suggestion
    /// Returns a `bool` indicating whether the storage is in-memory only.
```

niklaslong (approved):

howardwu (approved):

raychu86 (approved):

raychu86:
I believe with this approach, we can have snarkVM can use an in-memory ledger for dpc testing rather than having to import `snarkos-testing`.

I'd imagine this would save us from having any cyclic dependencies between `snarkVM` and `snarkOS`.

ljedrz:
@raychu86 I think there were some `Ledger` functionalities defined only in `snarkos-storage` and I wasn't able to do it that way; in any case, this PR is a step forward.

ljedrz:
@raychu86 confirmed: it's primarily the `Ledger` struct from `storage` that 's needed for the DPC tests; it's also used in `setup_or_load_parameters` and `initialize_test_blockchain`, which are in the `testing` crate; it might be possible to leave a "trimmed" rendition of a `Ledger`, but it would still need to implement at least some of the functionalities already defined in `snarkos-storage` in order to be usable.

We don't need to have this as a circular dependency, actually; the new crate containing DPC tests could be excluded from the workspace members and tested in the CI completely separately, which I'd consider the cleanest solution - it doesn't even affect the `Cargo.lock` of snarkVM.

PR #99, opened 2021-03-09
Fix leo#602 & leo#735

ljedrz:
This seems like the most straightforward solution to the issue described in https://github.com/AleoHQ/leo/issues/602: call the closures passed to `KeypairAssembly::{alloc, alloc_input}`, so that all the `AllocatedBit.value`s can become `Some(x)`, so that the assumptions (that `value.unwrap()` is safe) in the `PartialEq` impl for `Boolean` always hold.

Some tests using `generate_random_parameters` needed an adjustment: the value passed to it needed to contain `Some(x)` values instead of `None`; an alternative would be to not require `ConstraintSystem::{alloc, alloc_input}` to return a concrete value, i.e. to return a `Result<Option<Self>, _>` instead (which would be a *much* larger change). I used random values there, which seems to pass the tests, but some specific values could be used as well.

Fixes https://github.com/AleoHQ/leo/issues/602, https://github.com/AleoHQ/leo/issues/735.

### DIFF
--- a/algorithms/tests/snark/mimc.rs
+++ b/algorithms/tests/snark/mimc.rs
@@ -172,8 +172,8 @@ fn test_mimc_groth_16() {
     // Create parameters for our circuit
     let params = {
         let c = MiMCDemo::<Fr> {
-            xl: None,
-            xr: None,
+            xl: Some(rng.gen()),
+            xr: Some(rng.gen()),
             constants: &constants,
         };
 
--- a/algorithms/tests/snark/mimc.rs
+++ b/algorithms/tests/snark/mimc.rs
@@ -172,8 +172,8 @@ fn test_mimc_groth_16() {
     // Create parameters for our circuit
     let params = {
         let c = MiMCDemo::<Fr> {
-            xl: Some(rng.gen()),
-            xr: Some(rng.gen()),
+            xl: None,
+            xr: None,
             constants: &constants,
         };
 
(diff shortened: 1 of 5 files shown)


algorithms/tests/snark/mimc.rs
```diff
@@ -172,8 +172,8 @@ fn test_mimc_groth_16() {
     // Create parameters for our circuit
     let params = {
         let c = MiMCDemo::<Fr> {
-            xl: None,
-            xr: None,
+            xl: Some(rng.gen()),
+            xr: Some(rng.gen()),
```
howardwu, line 176: These should not need to exist for the setup to run. For testing purposes, to ensure that is the case, it is best to leave it unassigned.

ljedrz: If either `xl` or `xr` is `None`, `generate_constraints` will fail at the first call to `alloc`, crashing the test with `AssignmentMissing`.

An alternative approach could be to only execute the closures in `alloc` and `alloc_input` outside of `#[cfg(test)]` conditions, if that's preferable.

ljedrz:
@howardwu as stated in the earlier reply, I adjusted the PR to only call the closures in non-test scenarios; this fixes the 2 Leo issues without breaking or changing any of the snarkVM tests.

collinc97:
Bugs  AleoHQ/leo#602, AleoHQ/leo#735 are related to the division gadget implementation. They should not be closed with this PR.

PR #103, opened 2021-03-12
Fix leo#738

ljedrz:
As discussed in https://github.com/AleoHQ/leo/pull/762 and outlined in https://github.com/AleoHQ/leo/issues/763, leo will be removing its `gadgets` and intstead use the ones in snarkVM; therefore, this PR replaces the one filed in leo, and adjusts the `pow` operation for signed integers to internally use a new `Mul::wrapping_mul` method in an intermediate operation that should be allowed to overflow.

In addition, the signed `pow` tests are adjusted and extended so that the bug can actually be triggered (which does happen without https://github.com/AleoHQ/snarkVM/pull/103/commits/556073c21da6afe7e97eb5a60e94227e5309dbf6). Some of them are also un-ignored.

Supersedes https://github.com/AleoHQ/leo/pull/762.
Fixes https://github.com/AleoHQ/leo/issues/738.

### DIFF
--- a/gadgets/src/traits/utilities/arithmetic/mul.rs
+++ b/gadgets/src/traits/utilities/arithmetic/mul.rs
@@ -25,4 +25,6 @@ where
     type ErrorType;
 
     fn mul<CS: ConstraintSystem<F>>(&self, cs: CS, other: &Self) -> Result<Self, Self::ErrorType>;
+
+    fn wrapping_mul<CS: ConstraintSystem<F>>(&self, cs: CS, other: &Self) -> Result<Self, Self::ErrorType>;
 }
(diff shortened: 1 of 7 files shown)


ljedrz:
@collinc97 as for your question in the original leo PR:
> Do you think we could make it more generic for both UInt and Int types?

I've gone over the current `pow` impl for unsigned integers, as well as its tests, and it seems that overflows are just perfectly allowed? Leo also seems to have no problem doing things like `let y = 32u8 ** 2u8;`. The same kind of solution might be applicable for them, but the `mul` implementation needs to be changed to one that actually involves the possibility of overflows in the first place, as the current one (`snarkvm_gadgets::traits::utilities::uint::mul`) is infallible (and not trait-based either).

howardwu (approved):

PR #119, opened 2021-03-19
Implements Fiat-Shamir RNG gadgets

raychu86:
## Motivation

This PR implements the `Poseidon Algebraic Sponge` and the  `Fiat Shamir Algebraic Sponge RNG Gadget`

## Test Plan

Tests have been added to both modules to ensure that the native and gadget implementation of each function is correct.


howardwu:
Updated to have the ChaCha case match the original approach.

PR #124, opened 2021-03-23
Fix signed integer subtraction involving iX::MIN

ljedrz:
The current state of affairs is that subtraction with `iX::MIN` as the second operand results in an overflow, as signed integer subtraction is implemented using the following algorithm: `a - b = a + (-b)` and if the value of `b` is the minimum boundary of that type, the negation results in an overflow. This is special-cased in tests, which skip operations where `b.checked_neg().is_none()`.

A possible solution is to adjust the subtraction algorithm for `b == iX::MIN` to the following: `a - b = -(-a + b)`, and to handle the one operation where it doesn't work, i.e. when both `a` and `b` are equal to `iX::MIN` by just returning `0`.

Fixes https://github.com/AleoHQ/leo/issues/800.

### DIFF
--- a/gadgets/src/traits/utilities/int/arithmetic/sub.rs
+++ b/gadgets/src/traits/utilities/int/arithmetic/sub.rs
@@ -17,8 +17,10 @@
 use crate::{
     errors::SignedIntegerError,
     utilities::{
+        alloc::AllocGadget,
         arithmetic::{Add, Neg, Sub},
         int::*,
+        Integer,
     },
 };
 use snarkvm_fields::PrimeField;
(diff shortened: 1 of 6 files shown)


ljedrz:
There is an argument for disallowing negation when the negated value would overflow when negated on its own, but since this is an intermediate operation in subtraction, it is not visible to the user and could thus be confusing, especially if it happens as part of some larger arithmetic operation where the affected operands are not immediately apparent.

ljedrz:
Closing due to inactivity and conflicts, happy to revisit if still applicable.

PR #132, opened 2021-03-27
Implement Marlin Polycommit gadgets

raychu86:
## Motivation

This PR implements the Verifier Key and Prepared Verifier Key gadgets for the Marlin polycommit module.


polycommit/src/marlin_pc/gadgets/verifier_key.rs
```diff
@@ -0,0 +1,453 @@
...
+    #[test]
+    fn test_to_bytes() {
+        unimplemented!()
+    }
+
+    #[test]
+    fn test_to_constraint_field() {
+        unimplemented!()
+    }
+}
```
howardwu, line 453: Just checking here, do we need to implement these before merging?

polycommit/src/marlin_pc/gadgets/mod.rs
```diff
@@ -0,0 +1,34 @@
...
+// pub use commitment::*;
+//
+// mod labeled_commitment;
+// pub use labeled_commitment::*;
+//
+// mod prepared_commitment;
+// pub use prepared_commitment::*;
+//
+// mod prepared_labeled_commitment;
+// pub use prepared_labeled_commitment::*;
```
howardwu, line 27: Do these files need to be implemented?

howardwu: Discussed offline, marking as resolved.

howardwu (approved):

PR #138, opened 2021-03-30
Speed up merkle tree growth

ljedrz:
This PR dramatically reduces the time needed to grow the `MerkleTree` object, allowing block syncing in snarkOS to happen in a reasonable time frame. Only when the new leaves share nearly no ancestors a large portion of the tree needs to be rebuilt, otherwise a lot of the pre-exsting hashes can be reused.

The results of this change were validated by using both `MerkleTree::{new, rebuild}` when synchronizing blocks and comparing the trees; no issues were found for the first few thousand blocks taken from the testnet. snarkOS tests also pass with `MerkleTree::rebuild` implicitly being used in tests (when blocks are committed).

In addition, memory use in `MerkleTree::new` is reduced by using an `ExactSizeIterator` instead of a slice for the `leaves` argument.

A snarkOS-side PR will follow once a new version of snarkVM is released.

Fixes https://github.com/AleoHQ/snarkOS/issues/459.

A comparison of the currently used `MerkleTree::new` with the new `MerkleTree::rebuild`:
merkle_tree_comparison

### DIFF
--- a/algorithms/src/merkle_tree/merkle_tree.rs
+++ b/algorithms/src/merkle_tree/merkle_tree.rs
@@ -30,7 +30,7 @@ pub struct MerkleTree<P: MerkleParameters> {
     tree: Vec<MerkleTreeDigest<P>>,
 
     /// The hash of each non-empty leaf in the Merkle tree.
-    hashed_leaves: Vec<MerkleTreeDigest<P>>,
+    hashed_leaves: Vec<MerkleTreeDigest<P>>, // TODO(ljedrz): this can be procured without allocations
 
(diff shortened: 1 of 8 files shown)


ljedrz:
Added 2 commits with extra optimizations that provide an additional ~19% average perf improvement; used a logarithmic scale this time so the difference is easier to see:

merkle_tree_comparison2

ljedrz:
I've run synchronization for a bit longer (53k blocks, 21 minutes) and `MerkleTree` growth complexity seems to be logarithmic (as opposed to linear) now:

merkle_tree_50k

howardwu (approved):

PR #140, opened 2021-04-01
Remove storage and testing crates

ljedrz:
This PR removes the `storage` and `testing` crates (they reside in snarkOS now), making snarkVM free of the dependency on `rocksdb` and greatly speeding up its compilation time. The tests that depend on `snarkos-testing` and `snarkos-storage` reside in a new `snarkos-integration` crate, which is _not_ a member of the snarkVM workspace.

In order to run the aforementioned tests, the CI needs to be adjusted so that `snarkos-integration` tests have their own dedicated run; the associated image should also contain the necessary parameters, otherwise they take a while to download.

note: since the snarkos crates need to be aligned with the current snarkVM APIs, the snarkOS repository linked in `Cargo.toml` within `snarkos-integration` currently points to a branch associated with https://github.com/AleoHQ/snarkOS/pull/677. This can be changed once that PR is merged (to a fresh `AleoHQ/snarkOS` commit) or to a released version once it's released.


ljedrz:
another note: `Cargo.lock` in `snarkos-integration` might be unnecessary - it can be removed if desired.

howardwu:
note to self: TODO - change the `snarkOS` dependencies to the next release in `snarkos-integration/Cargo.toml`

PR #155, opened 2021-04-29
Make rebuild return a new MerkleTree for better synchronization logic

Protryon:
This will allow us to go lock free for MerkleTree storage for what appears to be very low cost.

This PR also includes `Arc` wrapping for parameters to avoid cloning cost.

This PR includes a version bump for `v0.3.0`. Semver minor bump due to API change in rebuild. It's particularly necessary since previous callers will not get a compiler error for input types, but the return type has changed from nothing to a new struct.

No change in performance observed from this change. See graphs:
Before:
image
After:
image

Using 100 block test chain.

### DIFF
--- a/algorithms/src/merkle_tree/merkle_tree.rs
+++ b/algorithms/src/merkle_tree/merkle_tree.rs
@@ -20,6 +20,7 @@ use crate::{
     traits::{MerkleParameters, CRH},
 };
 use snarkvm_utilities::ToBytes;
+use std::sync::Arc;
 
 #[derive(Default)]
 pub struct MerkleTree<P: MerkleParameters> {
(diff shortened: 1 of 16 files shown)


algorithms/src/merkle_tree/merkle_tree.rs
```diff
@@ -118,10 +119,10 @@ impl<P: MerkleParameters> MerkleTree<P> {
     }
 
     pub fn rebuild<L: ToBytes, I: ExactSizeIterator<Item = L>, J: ExactSizeIterator<Item = L>>(
-        &mut self,
+        &self,
```
ljedrz, line 122: would it be possible to consume `self` here instead of just borrowing it?

ljedrz (approved):
LGTM with one question :+1:

algorithms/src/merkle_tree/merkle_tree.rs
Protryon: @ljedrz No. We have to borrow it since the caller will only have an `Arc<MerkleTree>`

ljedrz (approved):

Issue #158, opened 2021-05-07
[Feature] Introduce optimization goal to constraint systems


raychu86:
## 🚀 Feature

Every constraint system that implements the `ConstraintSystem` trait should be optimized by `constraints` or by `weight`. This optimization will be labeled with an enum in each constraint system.

## Motivation

The current constraint system implementations are currently all constraint optimized, but the incoming work in Marlin will require weight optimized operations. Adding this base infrastructure will be necessary for that logic.

## Implementation

Each constraint system will have a new `optimization_goal` and the `ConstraintSystem` trait will require two additional functions: `optimization_goal` and `set_optimization_goal`.

PR #159, opened 2021-05-07
Implement constraint system optimization goals

raychu86:
## Motivation

This PR introduces the optimization goals to constraint systems as outlined in #158

### DIFF
--- a/algorithms/src/snark/groth16/generator.rs
+++ b/algorithms/src/snark/groth16/generator.rs
@@ -19,7 +19,15 @@ use crate::{cfg_into_iter, cfg_iter, fft::EvaluationDomain, msm::FixedBaseMSM};
 use snarkvm_curves::traits::{Group, PairingEngine, ProjectiveCurve};
 use snarkvm_fields::{Field, One, PrimeField, Zero};
 use snarkvm_profiler::{end_timer, start_timer};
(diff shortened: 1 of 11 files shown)


algorithms/src/snark/groth16/generator.rs
```diff
@@ -130,6 +139,19 @@ impl<E: PairingEngine> ConstraintSystem<E::Fr> for KeypairAssembly<E> {
     fn num_private_variables(&self) -> usize {
         self.num_private_variables
     }
+
+    fn optimization_goal(&self) -> OptimizationGoal {
+        self.optimization_goal.clone()
```
ljedrz, line 144: since `OptimizationGoal` derives `Copy,` this clone (and all other clones of `OptimizationGoal` in this PR) shouldn't be necessary

ljedrz: I now noticed `clippy` already complained about this later on; weird that it didn't notice this here

howardwu (approved):

howardwu:
We need to update the `AHP` indexer and prover impls to account for this, otherwise, it isn't being utilized.

howardwu:
Closing this PR as the optimizations aren't fully integrated and somewhat deprecated now.

PR #174, opened 2021-05-24
Fix LinearCombination build-time nondeterminism

Protryon:
This PR fixes build-time nondeterminism in `UInt::addmany` through `LinearCombination::add_assign`. Issue was noticed during upgrade from rust 1.51.0 -> 1.52.0, where

[rust 18e44a1b](https://github.com/rust-lang/rust/commit/18e44a1be4d8aecf41d0b904d63f6c72f51d6b0d)
[rust 7d078cfb9](https://github.com/rust-lang/rust/commit/7d078cfb94fa75e5dee699535f3f9781d3a1d47d)

changed internals of `binary_search_by` leading to a change in emitted R1CS. `binary_search_by` doesn't guarantee stability if the items being sorted are not unique, and the excess appending of duplicates this PR fixes removes the issue encountered from changing internals of rust.

*warning* This PR will change emitted R1CS from that of rust `1.51.0, `1.52.0`, and all other versions of rust.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -1,6 +1,6 @@
 [package]
 name = "snarkvm"
-version = "0.3.2"
+version = "0.3.1"
 authors = [ "The Aleo Team <hello@aleo.org>" ]
 description = "A decentralized virtual machine"
 homepage = "https://aleo.org"
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -1,6 +1,6 @@
 [package]
 name = "snarkvm"
-version = "0.3.1"
+version = "0.3.2"
 authors = [ "The Aleo Team <hello@aleo.org>" ]
 description = "A decentralized virtual machine"
 homepage = "https://aleo.org"
--- a/r1cs/src/linear_combination.rs
+++ b/r1cs/src/linear_combination.rs
@@ -91,7 +91,11 @@ impl<F: Field> LinearCombination<F> {
                     found_index += 1;
                 }
             }
-            Err(found_index)
+            if self.0.get(found_index).map(|x| &x.0 == search_var).unwrap_or_default() {
+                Ok(found_index)
+            } else {
+                Err(found_index)
+            }
         } else {
             self.0.binary_search_by_key(search_var, |&(cur_var, _)| cur_var)
         }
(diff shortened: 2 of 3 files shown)


damirka:
@Protryon does it change existing outputs as well? Or only on older builds?

Protryon:
> @Protryon does it change existing outputs as well? Or only on older builds?

Outputs of leo programs? No, they are not effected. R1CS is changed for all builds depending on addmany.

r1cs/src/linear_combination.rs
```diff
@@ -91,7 +91,11 @@ impl<F: Field> LinearCombination<F> {
                     found_index += 1;
                 }
             }
-            Err(found_index)
+            if self.0.get(found_index).map(|x| &x.0 == search_var).unwrap_or_default() {
+                Ok(found_index)
+            } else {
+                Err(found_index)
+            }
```
Pratyush, line 98: Hmm I think this is better resolved by just removing the special case for linear combinations of size smaller than 6; that way the behaviour won't deviate again in the future. I'm sure the rust compiler has special cases for linear scans already.

Protryon: FWIW, it doesn't have any special cases there (I looked to debug this). The logic is tested -- we can keep it if we want it. I assume there was a performance advantage to be had, so I don't see a pressing need to remove it.

howardwu (approved):
LGTM

Cargo.toml
```diff
@@ -1,6 +1,6 @@
 [package]
 name = "snarkvm"
-version = "0.3.1"
+version = "0.3.2"
```
howardwu, line 3:
```suggestion
version = "0.3.1"
```

howardwu: Release script will handle bumping

r1cs/src/linear_combination.rs
Pratyush: In that case, maybe it's better to rewrite the logic along the lines of:
```rust
for (i, (var, _)) in self.0.iter().enumerate() {
	use core::ord::Ordering::*;
	match var.cmp(search_var) {
		Equal => return Ok(i),
		Greater => return Err(i),
		_ => (),
	}
}
return Err(self.0.len());
```

Pratyush: (It's also not clear to me whether this is actually faster in practice or not; maybe we can add a benchmark.)

Protryon: Lets not fix what isn't broken unless there is a specific reason to. Is there any point to this?

Pratyush: Sure, no problem. My rationale was just that it's not clear why the code was wrong earlier, whereas with the new for loop it's clearer what the behaviour is (at least IMO). But that's subjective.

Pratyush (approved):

howardwu (approved):

PR #177, opened 2021-05-26
Update the deps of the snarkos-integration crate

ljedrz:
This fixes the compliation of snarkOS integration tests after the snarkVM package versions in snarkOS were updated.

note: I'd have used `version = "1.3.6"` instead of its commit SHA, but it seems that `snarkos-testing` hasn't gotten a release since `1.1.4`.

Fixes https://github.com/AleoHQ/snarkVM/issues/176.


snarkos-integration/Cargo.toml
```diff
@@ -82,14 +82,14 @@ version = "0.10.0"
 version = "0.8"
 
 [dependencies.snarkos-storage]
-git = "https://github.com/ljedrz/snarkOS"
-rev = "2eca11a85989bab2f5ea00854f7efdfab7859d95"
+git = "https://github.com/AleoHQ/snarkOS"
+rev = "e72d3d9d03d1a053ae148608e3f5b3ae857a4edf" # version 1.3.6 of the package
```
howardwu, line 86: Can we switch these to `version` instead of `rev` going forward?

ljedrz: Only if `snarkos-testing` gets a new release on [crates.io](crates.io), otherwise it won't compile.

howardwu: Got it, I'll create a Github issue to flag this for now and make a note to include `snarkos-testing` in the release pipeline going forward, thanks!

howardwu (approved):

PR #178, opened 2021-05-27
Optimize Bowe Hopwood Pedersen Hash

Protryon:
## Motivation

Optimize Bowe Hopwood Pedersen hashing function

## Approaches taken
* Removed use of concurrency since it was hurting (bad use case)
* Removed 2 heap allocations (now zero)
* Enforced single copy of input only
* Created lookup table for bit lookups

## Approaches possible going forward
* Use a larger (I think most optimally 24-bit lookup table) so we can do less/no bit arithmetic
  * Note that since we usually do 8 windows, that would take 128 MB of static/on-load-init memory which is kinda big and probably costly to make at runtime. Definitely costly to store in memory.
* Create a specialized variant that requires input to be prepadded for zero-copy

Note that shrinking the stack buffer isn't listed since this aught not to be an expensive allocation (its not heap, just zeroing), and actual testing yielded nothing of significance.

## Effects
* Hash speed is about 70% faster in benchmarks matching merkle tree configuration
  * Should snarkOS improve load times by ~70%
  * Should improve block sync time by ~70%

### DIFF
--- a/algorithms/src/crh/bowe_hopwood_pedersen.rs
+++ b/algorithms/src/crh/bowe_hopwood_pedersen.rs
@@ -135,7 +135,12 @@ impl<G: Group, S: PedersenSize> CRH for BoweHopwoodPedersenCRH<G, S> {
             S::NUM_WINDOWS,
             BOWE_HOPWOOD_CHUNK_SIZE,
         );
+        assert_eq!(self.parameters.len(), S::NUM_WINDOWS);
         for bases in self.parameters.bases.iter() {
+            assert_eq!(bases.len(), S::WINDOW_SIZE);
+        }
+        assert_eq!(self.bowe_hopwood_parameters.base_lookup.len(), S::NUM_WINDOWS);
+        for bases in self.bowe_hopwood_parameters.base_lookup.iter() {
             assert_eq!(bases.len(), S::WINDOW_SIZE);
         }
         assert_eq!(BOWE_HOPWOOD_CHUNK_SIZE, 3);
--- a/algorithms/src/crh/bowe_hopwood_pedersen.rs
+++ b/algorithms/src/crh/bowe_hopwood_pedersen.rs
@@ -22,11 +22,11 @@ use crate::{
     errors::CRHError,
     traits::CRH,
 };
-use bitvec::{order::Lsb0, view::BitView};
 use snarkvm_curves::Group;
 use snarkvm_fields::{ConstraintFieldError, Field, PrimeField, ToConstraintField};
 use snarkvm_utilities::biginteger::biginteger::BigInteger;
 
+use bitvec::{order::Lsb0, view::BitView};
 use rand::Rng;
 
 #[derive(Debug, Clone, PartialEq, Eq, PartialOrd, Ord)]
--- a/algorithms/src/crh/bowe_hopwood_pedersen.rs
+++ b/algorithms/src/crh/bowe_hopwood_pedersen.rs
@@ -16,25 +16,23 @@
 
 pub use crate::crh::pedersen_parameters::PedersenSize;
 
+use super::bowe_hopwood_pedersen_parameters::*;
 use crate::{
     crh::{PedersenCRH, PedersenCRHParameters},
     errors::CRHError,
     traits::CRH,
 };
+use bitvec::{order::Lsb0, view::BitView};
 use snarkvm_curves::Group;
 use snarkvm_fields::{ConstraintFieldError, Field, PrimeField, ToConstraintField};
-use snarkvm_utilities::{biginteger::biginteger::BigInteger, bytes_to_bits};
+use snarkvm_utilities::biginteger::biginteger::BigInteger;
 
 use rand::Rng;
-
-#[cfg(feature = "parallel")]
-use rayon::prelude::*;
-
-pub const BOWE_HOPWOOD_CHUNK_SIZE: usize = 3;
 
 #[derive(Debug, Clone, PartialEq, Eq, PartialOrd, Ord)]
 pub struct BoweHopwoodPedersenCRH<G: Group, S: PedersenSize> {
     pub parameters: PedersenCRHParameters<G, S>,
+    pub bowe_hopwood_parameters: BoweHopwoodPedersenCRHParameters<G>,
 }
 
 impl<G: Group, S: PedersenSize> BoweHopwoodPedersenCRH<G, S> {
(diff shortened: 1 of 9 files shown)


algorithms/src/crh/bowe_hopwood_pedersen.rs
```diff
@@ -16,25 +16,23 @@
 
 pub use crate::crh::pedersen_parameters::PedersenSize;
 
+use super::bowe_hopwood_pedersen_parameters::*;
 use crate::{
     crh::{PedersenCRH, PedersenCRHParameters},
     errors::CRHError,
     traits::CRH,
 };
+use bitvec::{order::Lsb0, view::BitView};
```
howardwu, line 25: nit: could you move this import below with `rand`

algorithms/src/crh/bowe_hopwood_pedersen.rs
```diff
@@ -16,25 +16,23 @@
...
 use rand::Rng;
 
-#[cfg(feature = "parallel")]
-use rayon::prelude::*;
-
-pub const BOWE_HOPWOOD_CHUNK_SIZE: usize = 3;
-
 #[derive(Debug, Clone, PartialEq, Eq, PartialOrd, Ord)]
 pub struct BoweHopwoodPedersenCRH<G: Group, S: PedersenSize> {
     pub parameters: PedersenCRHParameters<G, S>,
```
howardwu, line 34: Do we still need these parameters? It appears `bowe_hopwood_parameters` are the only used params here. Perhaps we can remove the unused `parameters and rename `bowe_hopwood_parameters` to `parameters` .

algorithms/src/crh/bowe_hopwood_pedersen.rs
```diff
@@ -106,30 +110,21 @@ impl<G: Group, S: PedersenSize> CRH for BoweHopwoodPedersenCRH<G, S> {
...
-                current_length + BOWE_HOPWOOD_CHUNK_SIZE - current_length % BOWE_HOPWOOD_CHUNK_SIZE,
-                false,
-            );
-        }
-
-        assert_eq!(padded_input.len() % BOWE_HOPWOOD_CHUNK_SIZE, 0);
+        assert_eq!(bit_len % BOWE_HOPWOOD_CHUNK_SIZE, 0);
 
         assert_eq!(
             self.parameters.bases.len(),
```
howardwu, line 130: Should this be checking `bowe_hopwood_parameters`?

algorithms/src/crh/bowe_hopwood_pedersen.rs
Protryon: They are necessary for serialization as `bowe_hopwood_parameters` are just a view of `parameters`.

raychu86 (approved):

algorithms/src/crh/bowe_hopwood_pedersen.rs
```diff
@@ -106,30 +110,21 @@ impl<G: Group, S: PedersenSize> CRH for BoweHopwoodPedersenCRH<G, S> {
...
-        // Pad the input if it is not the current length.
-        let mut input_bytes = input;
-        let mut padded_input_bytes = vec![];
-        if (input.len() * 8) < S::WINDOW_SIZE * S::NUM_WINDOWS {
-            padded_input_bytes.extend_from_slice(input_bytes);
-            padded_input_bytes.resize((S::WINDOW_SIZE * S::NUM_WINDOWS) / 8, 0u8);
-            input_bytes = padded_input_bytes.as_slice();
+        // we cant use these in array sizes since they are from a trait (and cant be refered to at const time)
+        assert!(S::WINDOW_SIZE <= 256);
+        assert!(S::NUM_WINDOWS <= 296);
```
raychu86, line 115: Are these bounds from the declaration in `dpc/instantiated.rs`? These sizes are not entirely fixed and may be changed if we decide to update the proof systems or alter the circuits (which will lead to larger/smaller window requirements).

Protryon: Yeah, the intention is to change these when necessary. It's annoying, but this area is so hot it's worth it for the performance gains.

Protryon: FWIW, I tried using const generics on the struct instead, same obscure rust shenanigans about psuedo consts.

raychu86: In any case, this should be technically correct, however not sure if hard coding a bound here is the correct approach.

raychu86: For now, I think it's ok. But we may want to add more explicit documentation about why these bounds were selected and the cases for which they would need to be updated.

algorithms/src/crh/bowe_hopwood_pedersen.rs
```diff
@@ -106,30 +110,21 @@ impl<G: Group, S: PedersenSize> CRH for BoweHopwoodPedersenCRH<G, S> {
...
+
+        // overzealous but stack allocation
+        let mut buffer = [0u8; 296 * 256 / 8 + BOWE_HOPWOOD_CHUNK_SIZE + 1];
```
howardwu, line 118: @Protryon what do you think about setting the 256 and 296 as global consts in this file, so it's clearly denoted at the top that these are magical constants and aren't buried into the code itself?

algorithms/src/crh/bowe_hopwood_pedersen.rs
howardwu: See my other comment for a placeholder proposal for the time being

algorithms/src/crh/bowe_hopwood_pedersen.rs
Protryon: Sounds reasonable

algorithms/src/crh/bowe_hopwood_pedersen.rs
```diff
@@ -110,9 +114,8 @@ impl<G: Group, S: PedersenSize> CRH for BoweHopwoodPedersenCRH<G, S> {
...
             ));
         }
-        // we cant use these in array sizes since they are from a trait (and cant be refered to at const time)
-        assert!(S::WINDOW_SIZE <= 256);
-        assert!(S::NUM_WINDOWS <= 296);
+        assert!(S::WINDOW_SIZE <= MAX_WINDOW_SIZE);
+        assert!(S::NUM_WINDOWS <= MAX_NUM_WINDOWS);
 
         // overzealous but stack allocation
         let mut buffer = [0u8; 296 * 256 / 8 + BOWE_HOPWOOD_CHUNK_SIZE + 1];
```
howardwu, line 121: I believe you should incorporate usage of `MAX_WINDOW_SIZE` and `MAX_NUM_WINDOWS` here

Protryon: That would make sense lol

howardwu (approved):

PR #184, opened 2021-05-29
Parallelize Merkle Tree Build & Rebuild

Protryon:
This PR parallelizes merkle tree build and part of rebuild (only leaf hashing).

Saw a 4x improvement in tree build speed (800ms -> 200ms locally on benchmark). Subjectively, snarkOS loads instantly and syncs quite quickly, during CPU-bound bits.

Upper tree rebuild was not parallelized because A. more complex logic and B. far less hash iterations per rebuild than fresh building.

NOTE: This is a *breaking* change! This requires a minor version update to snarkVM, not a patch.

### DIFF
--- a/algorithms/src/merkle_tree/merkle_tree.rs
+++ b/algorithms/src/merkle_tree/merkle_tree.rs
@@ -41,10 +41,30 @@ pub struct MerkleTree<P: MerkleParameters> {
     parameters: Arc<P>,
 }
 
-impl<P: MerkleParameters> MerkleTree<P> {
+#[cfg(feature = "parallel")]
+use rayon::prelude::*;
+
+impl<P: MerkleParameters + Send + Sync> MerkleTree<P> {
     pub const DEPTH: u8 = P::DEPTH as u8;
 
-    pub fn new<L: ToBytes, I: ExactSizeIterator<Item = L>>(parameters: Arc<P>, leaves: I) -> Result<Self, MerkleError> {
+    fn hash_row<L: ToBytes + Send + Sync>(
+        parameters: &P,
+        leaves: &[L],
+    ) -> Result<Vec<Vec<<<P as MerkleParameters>::H as CRH>::Output>>, MerkleError> {
+        let hash_input_size_in_bytes = (P::H::INPUT_SIZE_BITS / 8) * 2;
+        cfg_chunks!(leaves, 500)
+            .map(|chunk| -> Result<Vec<_>, MerkleError> {
+                let mut buffer = vec![0u8; hash_input_size_in_bytes];
+                let mut out = Vec::with_capacity(chunk.len());
+                for leaf in chunk.into_iter() {
+                    out.push(parameters.hash_leaf(&leaf, &mut buffer)?);
+                }
+                Ok(out)
+            })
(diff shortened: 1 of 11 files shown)


algorithms/src/merkle_tree/merkle_tree.rs
```diff
@@ -41,10 +41,30 @@ pub struct MerkleTree<P: MerkleParameters> {
...
+impl<P: MerkleParameters + Send + Sync> MerkleTree<P> {
     pub const DEPTH: u8 = P::DEPTH as u8;
 
-    pub fn new<L: ToBytes, I: ExactSizeIterator<Item = L>>(parameters: Arc<P>, leaves: I) -> Result<Self, MerkleError> {
+    fn hash_row<L: ToBytes + Send + Sync>(
+        parameters: &P,
+        leaves: &[L],
+    ) -> Result<Vec<Vec<<<P as MerkleParameters>::H as CRH>::Output>>, MerkleError> {
+        let hash_input_size_in_bytes = (P::H::INPUT_SIZE_BITS / 8) * 2;
+        cfg_chunks!(leaves, 500)
```
Protryon, line 55: 500 was found experimentally

algorithms/src/merkle_tree/merkle_tree.rs
```diff
@@ -41,10 +41,30 @@ pub struct MerkleTree<P: MerkleParameters> {
...
 
-impl<P: MerkleParameters> MerkleTree<P> {
+#[cfg(feature = "parallel")]
+use rayon::prelude::*;
+
+impl<P: MerkleParameters + Send + Sync> MerkleTree<P> {
     pub const DEPTH: u8 = P::DEPTH as u8;
 
-    pub fn new<L: ToBytes, I: ExactSizeIterator<Item = L>>(parameters: Arc<P>, leaves: I) -> Result<Self, MerkleError> {
+    fn hash_row<L: ToBytes + Send + Sync>(
```
howardwu, line 50: nit: can you move private methods to the bottom of the impl scope?

algorithms/src/merkle_tree/merkle_tree.rs
howardwu, line 55: Can you clarify if `500` is a newly defined magical constant? We may want to parameterize this using a global const if so.

Protryon: Covered in above comment, it's experimentally found. I don't think this needs a global const.

algorithms/src/merkle_tree/merkle_tree.rs
Protryon: Is there a philosophy/meaning behind the nit here? I placed it here since it was nearest usage.

howardwu: not sure if its generally used, but in my experience private methods were moved further down as developers using the struct would never call it, so it doesn't serve a meaningful purpose to end-devs looking for methods to call

Protryon: I generally follow the approach of trying to cluster related stuff together, since then it's easier to build a mental model of whats going on with little/no context, without jumping around a lot.

howardwu (approved):

PR #201, opened 2021-06-11
MSM base changes for CUDA and ASM

Protryon:
This PR does some minor refactors for CUDA and ASM, which will be rebased on this PR.


algorithms/src/msm/variable_base/mod.rs
```diff
@@ -0,0 +1,73 @@
...
+
+use snarkvm_curves::traits::AffineCurve;
+use snarkvm_fields::{PrimeField, Zero};
+
+mod standard;
+
+pub struct VariableBaseMSM;
+
+impl VariableBaseMSM {
+    #[allow(unused)]
```
howardwu, line 25: Can you use a `deprecated` flag here so it is not used?

algorithms/src/msm/variable_base/mod.rs
```diff
@@ -0,0 +1,73 @@
...
+    fn msm_basic<G: AffineCurve>(bases: &[G], scalars: &[<G::ScalarField as PrimeField>::BigInteger]) -> G::Projective {
```
howardwu, line 26: Can you rename to `msm_naive`?

algorithms/src/msm/variable_base/mod.rs
Protryon: It's a private(unexposed) method, and serves as a reference implementation for tests that is more likely to be correct due to simplicity. It is used in tests. I think `deprecated` flag is not appropriate, if that's fine with you.

Protryon: Was also renamed to `naive` rather than `basic`, which I think is more appropriate.

howardwu: Just realized it's private. Can you move it below so it's clearly private?

howardwu (approved):

PR #207, opened 2021-06-16
Fix the Marlin verifier gadget with more than one input

weikengchen:
## Motivation

Per Slack channel communication with @raychu86, the Marlin verifier gadget fails when there are two inputs, as evidenced by the two failing tests in our [bug/marlin-verification-inputs](https://github.com/AleoHQ/snarkVM/commit/aa7acdd2ad6e1f9d6a734e3e291c2ebe1a8b2006) branch.

An investigation shows that the problem comes from here:

When input has 2 elements [a, b], Marlin pads it to 4 [1, a, b, 0].
- The prover/verifier in the non-constraint world: push [1, a, b, 0] to the sponge
- The verifier gadget in the constraint world: push [1, a, b] to the sponge (ignoring the zero-padding), so there is inconsistency

Therefore, since the sponge is different, all the challenges are different, and the verifier gadget rejects the proof.

## Test Plan

After the changes, the code passes the two tests in the `bug/marlin-verification-inputs` branch. In addition, it passes all the existing 38 tests in the `marlin` subdirectory.

## Related PRs

N/A

### DIFF
--- a/marlin/src/constraints/verifier.rs
+++ b/marlin/src/constraints/verifier.rs
@@ -45,6 +45,7 @@ use crate::{
     PoseidonSponge,
     PoseidonSpongeVar,
 };
+use snarkvm_algorithms::fft::EvaluationDomain;
 
 /// The Marlin verification gadget.
 pub struct MarlinVerificationGadget<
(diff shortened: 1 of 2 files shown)


weikengchen:
Reformatted and needs a reapproval for running workflow.

howardwu (approved):

PR #208, opened 2021-06-17
Fix native cpu flag on release

howardwu:
## Motivation

Fix native cpu flag on release


howardwu:
https://github.com/AleoHQ/snarkOS/pull/865/files

Issue #210, opened 2021-06-17
[Proposal] Use sort-by-point-name in poly-commit


weikengchen:
## 💥 Proposal

This is a follow-up of the PR. 

There is another issue causing the `marlin_test_nested_snark` test to fail. It requires big-picture changes to completely fix this one--both native and constraints must sort points by names. 

In snarkVM, it follows an older version's poly-commit implementation. It sorts the points by values so that it matches the native order. 
https://github.com/AleoHQ/snarkVM/blob/master/polycommit/src/marlin_pc/gadgets/marlin_kzg10.rs#L155

However, this is indeed problematic. In fact, the upstream poly-commit now intentionally sorts the points by names, both in the gadget and the native version. As you can see in the native batch_open, it is changed accordingly.
https://github.com/arkworks-rs/poly-commit/blob/constraints/src/marlin/marlin_pc/constraints.rs#L957
https://github.com/arkworks-rs/poly-commit/blob/master/src/marlin/marlin_pc/mod.rs#L490

Why sort by names? 

Because different proofs will have points being in different orders. Sometimes, alpha is smaller than beta. In other proofs, beta is smaller than alpha. Recalling that our circuit needs to be data-independent, and here, challenge-independent, to make sure that it can verify different proofs. 

Therefore, in arkworks-rs, we intentionally make the order by explicit names "alpha", "beta", so that it is not sensitive to specific values.

This also resolves Raymond's notes on the to-do about sorting the points by values but having trouble handling the setup mode. The sort-by-point-name approach is designed specifically to avoid leaving proof-specific things into the setup.

raychu86:
RIght. This also solves the potential `AssignmentMissing` issue during `point.value.value()?` https://github.com/AleoHQ/snarkVM/blob/bd37c5e0ed890ec2af13eefa9d9774f2e2e7a12c/polycommit/src/marlin_pc/gadgets/marlin_kzg10.rs#L159 will panic due do an unassigned variable in the setup phase

weikengchen:
Add a note to this: It is possible that we don't need the big-picture change, as just changing the code to be sort-by-point-name immediately fixes the test, though it has a side effect that all the other tests fail (I don't know why...). I will look into it in more detail.

Recently, arkworks-rs poly-commit may have some refactoring and the constraint code will be significantly cleaned up and improved. We may integrate those changes after it is done (likely in one month).

weikengchen:
This issue has been fixed.

PR #211, opened 2021-06-17
Fix Marlin gadget's risky verifying key preparation

weikengchen:
## Motivation

First thanks to @raychu86 for discovering this issue. This PR is made to the `bug/marlin-verification-inputs` branch, as there is more to do before merging to `master`.

When Marlin’s constraint gadget prepares the key, it uses a trick to compute a hash of `CommitmentVar<F>` in the native world; this trick breaks the wall between the constraint world and the native world.

This trick is used at that time because arkworks-rs did not have `ToConstraintField` widely implemented in the native world. So, that was the only way to do `to_constraint_elements`. Soon after that arkworks-rs has a wide implementation of `ToConstraintField` for all SW and TE curve points and snarkVM, but we haven't updated these implementations.

This trick would work fine unless the R1CS has another optimization… Recall that for efficiency, `Setup` does not store any values for efficiency; therefore, when key preparation breaks the wall, it sees all zeros in `Setup`, but meaningful real values in `Prove`.

Therefore, `Setup` believes that the `vk` has a very different hash, though `Prove` disagrees. 

The Marlin’s constraint gadget has a few other places also breaking this wall, which is left as a to-do for cleanup.

## Test is still failing

This is due to the same reason, but in the poly-commit, and requires a big-picture change. See https://github.com/AleoHQ/snarkVM/issues/210 for more discussion.

### DIFF
--- a/marlin/src/constraints/verifier.rs
+++ b/marlin/src/constraints/verifier.rs
@@ -117,6 +117,7 @@ where
     BaseField: PrimeField,
     PC: PolynomialCommitment<TargetField>,
     PCG: PCCheckVar<TargetField, PC, BaseField>,
+    PC::Commitment: ToConstraintField<BaseField>,
     PCG::VerifierKeyVar: ToConstraintFieldGadget<BaseField>,
     PCG::CommitmentVar: ToConstraintFieldGadget<BaseField>,
 {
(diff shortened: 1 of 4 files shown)


weikengchen:
@raychu86 I will be looking into #210 recently. There is a possibility that we don't need the big picture change. If you sort the points by name, the failing test immediately passes, though all the other tests fail. 

I will look into that in more detail, likely tomorrow night.

PR #215, opened 2021-06-24
Parallelize BH pedersen setup

Protryon:
For account creation speeds where we run setup but never hash. Also parallelizes setup.


collinc97 (approved):
Thanks @Protryon this is great. Speedup on sdk side is 5x

howardwu (approved):

PR #218, opened 2021-06-25
SnarkVM IR

Protryon:
General design notes:
* This PR creates an intermediate representation (IR) that abstracts over the proof system, constraint system, curve, and field.
* A follow-up PR is prepared for AleoHQ/leo repo, available on `ir` branch. Needs release of these changes.
* Tests are derived from Leo tests.
* In general, the use of `anyhow` crate for error management is significantly increased. Scope is used to determine instruction location, and the intention is to use source mapping in Leo for end user errors (not yet implemented). This is mostly done for convenience. Existing error types are used in some places.
* The `ir` subcrate this PR adds provides structure and serialization for the IR format.
* The `eval` subcrate provides tooling for creating an `Evaluator`, and a default `SetupEvaluator` implementation mirroring existing logic of Leo.
* IR generation, register allocation, type checking, etc, is all supplied by the IR generator -- i.e. Leo.
* Minimal runtime type checking is done, but proper implementation from Leo should avoid any compilation-level errors in generated IR.


Pratyush:
Does this actually touch 988 files? My github doesn't even load that large of a diff =/

Pratyush:
Great, thanks! BTW, is there a document that provides an overview for the IR design?

utilities/src/biginteger/macros.rs
```diff
@@ -213,6 +213,13 @@ macro_rules! bigint_impl {
 
                 res
             }
+
+            fn from_slice(input: &[u64]) -> Self {
```
howardwu, line 217: I would prefer if we didn't add this method, as it is prone to abuse by Field impls, which require normalization to Montgomery form for any BigInteger

howardwu (approved):
LGTM, pending a minor requested change.

The `eval` module mirrors the deprecated logic from Leo, as such any unintended side effects will need to be caught from existing bug testing pipelines.

The `ir` module matches the specifications. I still have questions about the "complex" instruction set, however we can get this in for now and begin refining the architecture.

utilities/src/biginteger/macros.rs
Protryon: Would it be OK to keep this, but only use it in field impls, enforcing a montgomery reduction? The problem this is trying to solve is instantiating a parameterized field from a slice.

Protryon: It also seems current `from_repr` implementation in fields forces a valid field.

PR #221, opened 2021-06-29
IR Tests

Protryon:


howardwu (approved):

howardwu:
@Protryon could you reverse the order of the test directory from `tests/ir` to `ir/tests`? If this is meant to be for more general purpose testing, then perhaps keeping it in `.resources` would work as well.

PR #223, opened 2021-06-29
Apply the changes that fix the Marlin verifier to `feat/testnet2-dpc`

### LINKED ISSUE #210: [Proposal] Use sort-by-point-name in poly-commit

weikengchen:
## Motivation

This PR mirrors #222, but it is pushed to `feat/testnet2-dpc`. 

GitHub and my local merge via `git apply -3` seem to have some conflicts, so I merge manually.

## Test Plan

Tests in `polycommit` and `marlin` all pass.

## Related PRs

This is essentially the same PR as #222, albeit solving the merge conflicts.

It closes #210.

### DIFF
--- a/marlin/src/ahp/verifier/verifier.rs
+++ b/marlin/src/ahp/verifier/verifier.rs
@@ -139,10 +139,10 @@ impl<TargetField: PrimeField> AHPForR1CS<TargetField> {
         //  LinearCombination::new("z_b", vec![(F::one(), z_b)])
         //  LinearCombination::new("g_1", vec![(F::one(), g_1)], rhs::new(g_1_at_beta))
         //  LinearCombination::new("t", vec![(F::one(), t)])
-        query_set.insert(("g_1".into(), beta));
-        query_set.insert(("z_b".into(), beta));
-        query_set.insert(("t".into(), beta));
-        query_set.insert(("outer_sumcheck".into(), beta));
+        query_set.insert(("g_1".into(), ("beta".into(), beta)));
+        query_set.insert(("z_b".into(), ("beta".into(), beta)));
+        query_set.insert(("t".into(), ("beta".into(), beta)));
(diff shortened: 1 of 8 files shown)


weikengchen:
A few checks are failing, but it seems to be formatting changes that are not part of this PR. So, the PR is ready for review & merge.

raychu86 (approved):
LGTM. Will also add the additional marlin verification tests defined in `bug/marlin-verification-inputs` in a separate PR.

PR #226, opened 2021-06-30
Implement Schnorr signature verification gadget

raychu86:
## Motivation

This PR implements Schnorr Signature verification gadget. Currently it lives in the `SignaturePublicKeyRandomizationGadget`, which should be renamed to something more generic like `SignatureGadget`.

## Test Plan

Tests have been added to test the Schnorr Signature verification gadget in valid and invalid input messages.


raychu86:
Need to verify that the bit comparisons is equivalent to the `from_random_bytes` call in the native impl of Schnorr signature verification.

Pratyush:
You might want to take a look at our impl in arkworks: https://github.com/arkworks-rs/r1cs-tutorial/blob/2fd35dd6a036e7a55cb034f68f94a34e6f8938e2/simple-payments/src/signature/schnorr/constraints.rs#L79

Re: the `from_random_bytes`, why is that required? I couldn't quite make out the logic?

Pratyush:
I see, it's because the signature scheme views the verifier challenge as a field element. I would get rid of that, and simply view it as a sequence of bytes (like we do in the arkworks link above). You can make sure the bytes fit within the field by just dropping bits from the hash output

raychu86:
Good point. I had used that trick in a previous attempt by dropping bits, but I believe the tests failed because I used the constraint field instead of the group scalar field. Will make the fix today.

howardwu (approved):

howardwu (approved):

PR #231, opened 2021-07-01
Supply standard Poseidon parameters for all the curves related to BLS12-377

weikengchen:
## Motivation

This is a follow-up to #224 and the Slack discussion today involving the need of having Poseidon for BLS12-377's Fr.

For the ease of future use, this PR additionally samples Poseidon for SW6's Fq, BW6-761's Fq, and the Fr of the two Edwards curves of BLS12-377 ~(which may have a use case for Schnorr signatures)~. 

All such Fr/Fq fields but SW6's Fq can do Poseidon with alpha 31 (as 31 is coprime with q-1 or r-1 depending on whether it is Fr or Fq). For SW6's Fq, we consider alpha 37 = 32 + 1 + 4 since q - 1 has a prime factor 31.

The parameters are sampled by running the reference implementation https://extgit.iaik.tugraz.at/krypto/hadeshash/-/blob/master/code/generate_parameters_grain.sage without the step that samples the round constants.

The intentional choice of larger alpha rather than smaller alpha like 3 and 5 is because Marlin, which we envision future proofs may be based on, cares about the R1CS matrix weights, so what matters more is indeed the number of partial rounds, not alpha.

31 and 37 are chosen with small NAF weights, though our current gadgets are unaware of it and will do it naively. 

## Test Plan

N/A. They will be the same as the BLS12-377's Fq.

## Related PRs

This may help #226 in the future as it may use Poseidon to replace Pedersen + Blake2.


howardwu:
> 31 and 37 are chosen with small NAF weights, though our current gadgets are unaware of it and will do it naively.

@weikengchen should we be filing a Github issue to update this in our gadget impl?

howardwu (approved):

weikengchen:
> > 31 and 37 are chosen with small NAF weights, though our current gadgets are unaware of it and will do it naively.
> 
> @weikengchen should we be filing a Github issue to update this in our gadget impl?

Yes. 

By the way, before this PR can be merged.
If we focus on Groth16, we would be using smaller alpha like 3, 5, 7. What would be the choice for now? 

In that case, an implementation of powering with low NAF cost would be not needed.

howardwu:
It depends on which part we're talking about. To summarize:

For Leo programs which operate in `edwards_bls12`, optimizing for Marlin on `bls12_377::Fr` is ideal.
For the inner circuit, optimizing for Groth16 on `bls12_377::Fr` is ideal.
For PoSW, optimizing for Marlin on `bls12_377::Fr` is ideal.
For the outer circuit, optimizing for Groth16 on `bw6_761::Fr` is ideal.

So it may make more sense to have contextually specific parameters/multiple parameter instantiations.

weikengchen:
Got it. Let me convert this back to a draft since some changes will be needed.

howardwu:
No worries, if it helps, you can find the instantiations here:

[testnet1 instantiations](https://github.com/AleoHQ/snarkVM/blob/master/dpc/src/testnet1/instantiated.rs#L142&L148)
[testnet2 instantiations](https://github.com/AleoHQ/snarkVM/blob/master/dpc/src/testnet2/instantiated.rs#L153&L174) (subject to change as we update and optimize)

weikengchen:
Remaining todos:

- Add the Poseidon constraint gadgets that are cross-checked with the reference implementation
- Replace the sponge used in Marlin with the new snarkvm-sponge repo
- Replace the PRF used in Schnorr signature

Pratyush:
Isn't it fine to increase the exponent, as that reduces the number of rounds? I.e. is the overall degree of the computed function the same?

weikengchen:
> Isn't it fine to increase the exponent, as that reduces the number of rounds? I.e. is the overall degree of the computed function the same?

Yes, the description in the PR is outdated (I will update it soon). Eventually, the alpha is found via optimization. An example of BLS12-377's Fq is as follows:

```
const PARAMS_OPT_FOR_CONSTRAINTS: [[usize; 5]; 7] = [
        [2, 17, 8, 31, 0],
        [3, 5, 8, 56, 0],
        [4, 5, 8, 56, 0],
        [5, 5, 8, 57, 0],
        [6, 5, 8, 57, 0],
        [7, 5, 8, 57, 0],
        [8, 5, 8, 57, 0],
    ];
    const PARAMS_OPT_FOR_WEIGHTS: [[usize; 5]; 7] = [
        [2, 257, 8, 13, 0],
        [3, 257, 8, 13, 0],
        [4, 257, 8, 13, 0],
        [5, 257, 8, 13, 0],
        [6, 257, 8, 13, 0],
        [7, 257, 8, 13, 0],
        [8, 257, 8, 13, 0],
    ];
```

It means that:
 - when optimized for constraints, given a rate of 2 (aka state size 3), use alpha = 17, 8 full rounds, 31 partial rounds.
 - however, when rate is higher than 2, switch to alpha = 5, 8 full rounds, 56 - 57 partial rounds.

- when optimized for weights, things are very different, as it prefers 257.

Pratyush:
Hmm how many constraints are required in the "weights-optimized" setting? I ask because IMO it seems that we'll run into subtle incompatibilities if there are different instantiations of Poseidon in the same system. For example, I might want to verify a Schnorr signature in both Marlin and in the inner SNARK (Groth16); the signature shouldn't change just because of the proof system.

weikengchen:
Good points. Let me take a look. I think we also need a "balanced" or at least "weighted" version, where Groth16 is okay to be slightly slower to give Marlin some space, but just do not go to the extremes.

weikengchen:
As this PR falls beyond the original scope, I will close this one and open a new one.

PR #241, opened 2021-07-06
Implements `hash_to_curve` and samples generators for BLS12-377 and BW6-761

### LINKED ISSUE #164: [Bug] Fix Bls12-377 scalar group generator
## 🐛 Bug Report

Cross referencing https://github.com/arkworks-rs/curves/pull/60

```
Previously we used 11 as a generator, which has order (p-1)/35.

Now we use 22, which has the right order.

Fixed the two-adic root of unity in accordance with the new generator.
```

howardwu:
## Motivation

This PR implements Blake2Xs and uses it to implement `hash_to_curve`. The generators for BLS12-377 and BW6-761 are sampled and updated using `try_hash_to_curve`.

Unit tests checking the correspondence between the input message and generated point, as well as the BigInteger representation are provided.

This PR fixes #164, closing the loop on the BLS12-377 generator bug.


howardwu:
There is a bug in the `fmt::Display` implementation of `BigInteger*`.

howardwu:
Closes #164

algorithms/src/hash_to_curve/hash_to_curve.rs
```diff
@@ -0,0 +1,83 @@
...
+/// Executes one round of hash-to-curve and returns a generator on success.
+#[inline]
+pub fn hash_to_curve<G: AffineCurve, const XOF_DIGEST_LENGTH: u16>(input: &str) -> Option<G> {
+    debug_assert!(G::SERIALIZED_SIZE > 0);
+    debug_assert!(G::SERIALIZED_SIZE <= XOF_DIGEST_LENGTH as usize);
+    debug_assert!(XOF_DIGEST_LENGTH % 32 == 0);
+    debug_assert!(XOF_DIGEST_LENGTH as usize - G::SERIALIZED_SIZE < 32);
+
+    // The number of Blake2Xs invocations needed.
+    let num_rounds: u16 = match G::SERIALIZED_SIZE % 32 > 0 {
```
kobigurk, line 50: Can simplify using '(x+y-1)/y'

curves/src/bls12_377/g1.rs
```diff
@@ -67,30 +67,38 @@ impl ShortWeierstrassParameters for Bls12_377G1Parameters {
     }
 }
 
+///
 /// G1_GENERATOR_X =
-/// 81937999373150964239938255573465948239988671502647976594219695644855304257327692006745978603320413799295628339695
+/// 180626689152567938232604559698399306213229526488826235533375072902470977089128486884883292831560455202598472059967
```
kobigurk, line 72: Why has this changed?

Or, if considering changing, how about deriving it in a nothing up your sleeve manner? Like taking the minimal x value that gives a valid point?

(I see that it was like that and was removed! Curious why?)

algorithms/src/prf/blake2xs.rs
```diff
@@ -0,0 +1,344 @@
...
+            // Offset 24 - Personalization
+            u32::from_le_bytes([PERSONALIZATION as u8, (PERSONALIZATION >> 8) as u8, (PERSONALIZATION >> 16) as u8, (PERSONALIZATION >> 24) as u8]),
+            // Offset 24 - Personalization (continued)
+            u32::from_le_bytes([(PERSONALIZATION >> 32) as u8, (PERSONALIZATION >> 40) as u8, (PERSONALIZATION >> 48) as u8, (PERSONALIZATION >> 56) as u8]),
+        ]
+    }
+}
+
+#[cfg(test)]
+mod tests {
```
kobigurk, line 152: Reference test vectors available here: https://raw.githubusercontent.com/BLAKE2/BLAKE2/master/testvectors/blake2-kat.json

algorithms/src/hash_to_curve/hash_to_curve.rs
```diff
@@ -0,0 +1,83 @@
...
+        true => ((G::SERIALIZED_SIZE / 32) + 1) as u16,
+        false => (G::SERIALIZED_SIZE / 32) as u16,
+    };
+    debug_assert!(num_rounds > 0);
+
+    // Compute the digest for sampling the generator.
+    let mut digest = Vec::with_capacity(XOF_DIGEST_LENGTH as usize);
+    for offset in 0..num_rounds {
```
kobigurk, line 58: This is not exactly blake2xs since it's missing the inner hash. If it's not too harsh on the cost (or if the input is larger than 512 bits anyway), might be worth just implementing it exactly as blake2xs. If not, it's worth noting that it's a bit different somewhere and inspired by blake2xs rather than exactly it.

kobigurk (changes requested):

curves/src/bls12_377/g1.rs
howardwu: This is derived from the impl of `hash_to_curve` in this PR, with tests in `snarkvm-algorithms/hash_to_curve` confirming the new generator matches the input message.

As to why we changed this one, it's mainly to be generated uniformly with  BLS12-377 G2, BW6-761 G1 and G2, which didn't have a corresponding "increment from 0" generator.

algorithms/src/hash_to_curve/hash_to_curve.rs
howardwu: Good catch! Thank you :)

algorithms/src/prf/blake2xs.rs
howardwu: Thanks! Added the test vectors to the unit tests

algorithms/src/hash_to_curve/hash_to_curve.rs
howardwu: Reimplemented with blake2xs proper!

emmorais (approved):

algorithms/src/hash_to_curve/hash_to_curve.rs
```diff
@@ -0,0 +1,59 @@
...
+    debug_assert!(G::SERIALIZED_SIZE > 0);
+
+    // Compute the digest for sampling the generator.
+    let digest = Blake2Xs::evaluate(input.as_bytes(), G::SERIALIZED_SIZE as u16, "AleoHtC0".as_bytes());
+    debug_assert!(digest.len() == G::SERIALIZED_SIZE);
+
+    // Attempt to use the digest to derive a generator.
+    G::from_random_bytes(&digest).and_then(|g| {
+        debug_assert!(g.is_on_curve());
+        debug_assert!(!g.is_in_correct_subgroup_assuming_on_curve());
```
howardwu, line 48:
```suggestion
```

algorithms/src/encoding/elligator2.rs
```diff
@@ -61,7 +61,7 @@ impl<P: MontgomeryParameters + TwistedEdwardsParameters, G: Group + ProjectiveCu
             let r = input;
 
             // Let u = D.
-            // TODO (howardwu): change to 11.
+            // TODO (howardwu): CRITICAL - change to 11, confirm in gadget impl as well.
```
Pratyush, line 64: Why 11?

algorithms/src/hash_to_curve/hash_to_curve.rs
```diff
@@ -0,0 +1,58 @@
...
+        // Construct a new message.
+        let message = format!("{} in {}", input, k);
+
+        // Output the generator if a valid generator was found.
+        if let Some(g) = hash_to_curve::<G>(&message) {
+            return Some((g, message, k));
+        }
+    }
+    None
+}
```
Pratyush, line 34: Do we expect in practice that we don't find a curve point within 8 * 32 iterations? Also, why is `8 * SERIALIZED_SIZE` a good number?

algorithms/src/hash_to_curve/hash_to_curve.rs
```diff
@@ -0,0 +1,58 @@
...
+
+/// Executes one round of hash-to-curve and returns a generator on success.
+#[inline]
+pub fn hash_to_curve<G: AffineCurve>(input: &str) -> Option<G> {
```
Pratyush, line 38: I would expect the names to be swapped; this method should be called `try_hash_to_curve`, and the other should be called  `hash_to_curve`.

algorithms/src/hash_to_curve/hash_to_curve.rs
howardwu: I believe the standard assumption is to find a solution within `k` bits iterations where `k := log2(modulus)`

howardwu: "To control bias, hash_to_field instead uses random integers whose
   length is at least ceil(log2(p)) + k bits, where k is the target
   security level for the suite in bits.  Reducing such integers mod p
   gives bias at most 2^-k for any p; this bias is appropriate when
   targeting k-bit security.
"
From Section 5.1 in https://datatracker.ietf.org/doc/draft-irtf-cfrg-hash-to-curve/

algorithms/src/hash_to_curve/hash_to_curve.rs
```diff
@@ -0,0 +1,58 @@
...
+
+    // Attempt to use the digest to derive a generator.
+    G::from_random_bytes(&digest).and_then(|g| {
+        debug_assert!(g.is_on_curve());
+
+        let g = g.mul_by_cofactor();
+        debug_assert!(g.is_on_curve());
+        debug_assert!(g.is_in_correct_subgroup_assuming_on_curve());
+
+        match !g.is_zero() {
```
Pratyush, line 53:
```suggestion
        (!g.is_zero()).then(|| Some(g))
```

curves/src/bls12_377/fr.rs
```diff
@@ -55,21 +55,12 @@ impl FftParameters for FrParameters {
...
     #[rustfmt::skip]
     const TWO_ADIC_ROOT_OF_UNITY: BigInteger = BigInteger([
-        0x3c3d3ca739381fb2,
-        0x9a14cda3ec99772b,
-        0xd7aacc7c59724826,
-        0xd1ba211c5cc349c,
+        12646347781564978760u64,
+        6783048705277173164u64,
+        268534165941069093u64,
+        1121515446318641358u64,
```
Pratyush, line 63: This is 22?

algorithms/src/hash_to_curve/hash_to_curve.rs
Pratyush: I would use `BaseField::MODULUS_BITS` in that case

algorithms/src/hash_to_curve/hash_to_curve.rs
howardwu: Switching them now!

algorithms/src/hash_to_curve/hash_to_curve.rs
Pratyush: From reading the PR, and based on the rejection probability in each iteration, I think we can just set the upper bound to some number like 128 or 256.

howardwu: Discussed offline, will use a for loop to `128` and `panic!` on failure

algorithms/src/hash_to_curve/hash_to_curve.rs
howardwu: Added with correct usage, thanks!

curves/src/bls12_377/fr.rs
howardwu: The `GENERATOR` is 22, however the `TWO_ADIC_ROOT_OF_UNITY` is 8065159656716812877374967518403273466521432693661810619979959746626482506078.

Here is the Sage output:
```
2-adic gen (g2 % q):  8065159656716812877374967518403273466521432693661810619979959746626482506078
2-adic gen (g2 * R % q):  7039866554349711480672062101017509031917008525101396696252683426045173093960
2-adic gen into_chunks(g2 * R % q):  [12646347781564978760, 6783048705277173164, 268534165941069093, 1121515446318641358]
```

howardwu: I've added comments to the const declaration to clarify this

weikengchen (approved):
The code looks good.

I personally prefer more constant-time hashing-to-curve, since random point searching has a probability of failure (though negligible) and can sometimes be slow. 

But implementing those maps, especially efficiently, is hard and burdensome. I support the Blake2xs approach.

PR #245, opened 2021-07-08
Fix display in BigInteger

howardwu:
## Motivation

This is a hotfix for the Display trait in BigInteger, which now outputs a radix 10 integer representation as a string.

## Test Plan

Adds sanity checks to confirm the first two limbs are converted correctly, and samples random integers, confirming their integer string representation is correct.


weikengchen:
Another PR #246 also makes this fix, which may be more general, as it simply invokes the BigUint to convert the number.

weikengchen:
Just found that our solutions are almost identical (but I didn't add those tests). Let me approve this one, but you might hit some PR conflicts with #246.

utilities/src/biginteger/biginteger.rs
```diff
@@ -57,6 +57,9 @@ pub trait BigInteger:
     + AsRef<[u64]>
     + From<u64>
 {
+    /// The number of limbs used in this BigInteger.
+    const NUM_LIMBS: usize;
```
weikengchen, line 61: I like this. This should be done!

utilities/src/biginteger/macros.rs
```diff
@@ -167,16 +169,15 @@ macro_rules! bigint_impl {
             #[inline]
             fn from_bits_be(mut bits: Vec<bool>) -> Self {
                 let mut res = Self::default();
-                let mut acc: u64 = 0;
```
weikengchen, line 170: My PR didn't fix this issue. This part should be added soon.

weikengchen (approved):

howardwu:
Thanks Weikeng! Glad we both caught it around the same time :) I can work with you to resolve any merge conflicts in PR #246, I'll keep an eye on it

weikengchen:
Sure, I would prefer you merge this one first since it is smaller, otherwise, it would be overwhelmed by a bigger PR.

PR #242, opened 2021-07-07
Signature scheme cleanup and add flexibility to signature verification gadget

raychu86:
## Motivation

This PR removes unnecessary trait requirements like `digest` from Schnorr parameters and Schnorr verification gadgets. 

Additionally, more support has been added to the Signature verification gadget in order to support the signature scheme used in the Account model (GroupEncryption), which is a wrapper around the Schnorr signature implementation.

PR #246, opened 2021-07-08
Add full-fledged Poseidon and use it everywhere (Marlin and Schnorr)

weikengchen:
## Motivation

Let me apologize to @raychu86 first, as this PR might still have a few conflicts with the Schnorr signature you are working on.

This PR adds the latest implementation from `ark-sponge` of Poseidon and matches the official Poseidon reference implementation (which implies that ours is also a reference implementation). This part is captured by `snarkvm-sponge`.

Then, it replaces the existing sponge in Marlin constraint gadgets, as well as replaces the Blake2s hash function used in the Schnorr signature (this is where I came across Raymond's PR #242 since we are both removing Blake2s).

## Test Plan

The code passes a number of existing tests in the repo.

## Related PRs

This is based on #231 and subsumes #242.

## Followup

As Pratyush pointed out, sometimes the sponge is used in more than one proof system, and one may want an optimized set of parameters that is not too bad for Groth16, not too bad for Marlin. This will be surely in a separate PR, and it also involves research on the detailed performance gap between Groth16 and Marlin, so that we can weigh the optimization goal properly.


weikengchen:
This PR also has overlapped with #245, but #245 fixes other issues and should also be merged into the master soon.

utilities/src/biginteger/macros.rs
```diff
@@ -190,6 +190,28 @@ macro_rules! bigint_impl {
...
+                    res.extend_from_slice(&limb);
+                }
+                res
+            }
+
+            #[inline]
+            fn to_biguint(&self) -> num_bigint::BigUint {
+                BigUint::from_bytes_le(&self.to_bytes_le())
+            }
+
```
howardwu, line 214: This is great! Let's keep these methods along with the impl in PR #245

weikengchen: (I copied them from arkworks-rs, which we added the `to_biguint` just recently because otherwise printing out field elements is hard.)

algorithms/src/signature/schnorr.rs
```diff
@@ -166,32 +173,53 @@ where
...
+            }
+        }
 
-            // Compute the supposed verifier response: e := H(salt || r || msg);
-            if let Some(verifier_challenge) = <G as Group>::ScalarField::from_random_bytes(&D::digest(&hash_input)) {
-                break (random_scalar, verifier_challenge);
-            };
+        // Hash everything to get verifier challenge.
+        let mut hash_input = Vec::<<G as AffineCurve>::BaseField>::new();
+        hash_input.extend_from_slice(&self.parameters.salt.to_field_elements().unwrap());
```
howardwu, line 192: @raychu86 heads up that the branch with replacement of `salt` to `public_key` will conflict with the changes here.

weikengchen: That should be doable, as the public key will be a point, which can be hashed into the sponge just like the "commitment" in the next line.

weikengchen: @raychu86 Do you want which branch it is? If the change is small, I can merge it into this PR.

weikengchen:
one moment fixing dependencies... it seems to be more complicated than I think so I will start my VM...

howardwu (approved):

PR #250, opened 2021-07-09
Updates trait `PrimeField` repr methods to correct convention

howardwu:
## Motivation

`PrimeField` traits currently provide `PrimeField::into_repr` and `PrimeField::into_repr_raw`, which do not consume `self`, but rather use `&self` to construct a new instance.

In addition the `PrimeField::into_repr_raw` and `PrimeField::from_repr_raw` methods are unsafe to use as each `PrimeField` is in Montgomery representation (not standard form). Attempting to use these methods and introduce them without Montgomery conversion results in elliptic curve points (one level above) that do not lie on the correct curve (as they do not satisfy the corresponding curve equation).

As such, this PR renames these 4 `PrimeField` methods as `PrimeField::to_repr`, `PrimeField::to_repr_unchecked`, `PrimeField::from_repr`, and `PrimeField::from_repr_unchecked`.

### DIFF
--- a/algorithms/benches/msm/variable_base.rs
+++ b/algorithms/benches/msm/variable_base.rs
@@ -33,7 +33,7 @@ fn variable_base(c: &mut Criterion) {
 
     let mut rng = XorShiftRng::seed_from_u64(234872845u64);
 
-    let v = (0..SAMPLES).map(|_| Fr::rand(&mut rng).into_repr()).collect::<Vec<_>>();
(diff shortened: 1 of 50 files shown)


Pratyush:
Instead of `unsafe`, I'd go for `unchecked`, as `unsafe` implies some violations of Rust memory safety laws, which is not the case here.

howardwu:
Thanks @Pratyush! I've made the switch to using `unchecked` for those methods. Will keep an eye out for similar updates in our code.

raychu86 (approved):

weikengchen (approved):

PR #252, opened 2021-07-10
Fix the setup failure involving the Div gadget

weikengchen:
## Motivation

As reported by rlifchitz in Discord, the setup fails when dividing a u32 by a constant u32 (here, 1u32).

A more careful look shows that the setup fails when computing "a / b" when either a or b is not a constant.

The reason is that the Div gadget aims to obtain a few values to aid the computation, but these values cannot be obtained in the setup mode. However, this would require large-scope changes, as the Div gadget seems to be heavily requiring this. 

## Test Plan

A fix is made to snarkVM 0.5.5 and run tests on Leo 1.5.2. The problem is solved.


algorithms/src/snark/gm17/generator.rs
```diff
@@ -155,6 +155,10 @@ impl<E: PairingEngine> ConstraintSystem<E::Fr> for KeypairAssembly<E> {
     fn num_private_variables(&self) -> usize {
         self.num_private_variables
     }
+
+    fn is_in_setup_mode(&self) -> bool {
+        true
+    }
```
weikengchen, line 161: Explain: In order to allocate Vec<Boolean> without opening the function, now all the constraint systems must declare whether or not it is in the setup mode.

gadgets/src/integers/alloc.rs
```diff
@@ -35,33 +35,44 @@ macro_rules! alloc_gadget_fn_impl {
             mut cs: CS,
             value_gen: Fn,
         ) -> Result<Self, SynthesisError> {
-            let value = value_gen().map(|val| *val.borrow());
-            let values = match value {
-                Ok(mut val) => {
-                    let mut v = Vec::with_capacity(<$gadget as Integer>::SIZE);
+            if cs.is_in_setup_mode() {
```
weikengchen, line 38: Now, when allocating Vec<Boolean>, it will not retrieve the value when it is in the setup mode. This is crucial for the Div gadget to be runnable in the setup mode.

gadgets/src/integers/int/arithmetic/div.rs
```diff
@@ -63,191 +61,80 @@ macro_rules! div_int_impl {
                 // else
                 //    !Q                      -- negative result
 
-                if other.eq(&Self::constant(0 as <$gadget as Integer>::IntegerType)) {
-                    return Err(SignedIntegerError::DivisionByZero);
+                // If `self` and `other` are both constants, return the constant results instead of generating constraints
```
weikengchen, line 64: This part of the gadget has been completely rewritten following the comments and information originally provided in the document. The new gadget works in the setup mode and avoids incorrect constant allocations.

gadgets/src/integers/int/int_impl.rs
```diff
@@ -29,6 +33,8 @@ macro_rules! int_impl {
 
         impl Integer for $name {
             type IntegerType = $type_;
+            type UnsignedGadgetType = $uname_;
+            type UnsignedIntegerType = $utype_;
```
weikengchen, line 37: Explain: When programming the Div gadget for signed integers, now it uses the unsigned integers for help. This reduces the code complexity and seems to be the correct way to implement.

gadgets/src/integers/int/arithmetic/div.rs
weikengchen: Explain: In addition, the new code fixes a previous error. Previously, dividing -128 by 2 will result in -63 instead of -64.

gadgets/src/integers/uint/arithmetic/div.rs
```diff
@@ -54,100 +56,89 @@ macro_rules! div_int_impl {
                 //   end
                 // end
 
-                if other.eq(&Self::constant(0 as <$gadget as Integer>::IntegerType)) {
-                    return Err(SynthesisError::DivisionByZero.into());
+                // Enforce D != 0
```
weikengchen, line 59: Explain: This gadget is also completely rewritten, which avoids the incorrect constant allocations.

gadgets/src/integers/uint/arithmetic/sub.rs
```diff
@@ -53,71 +46,6 @@ macro_rules! sub_int_impl {
 
                 Self::addmany(&mut cs.ns(|| "add_not"), &[self.clone(), other.negate()]).map_err(|e| e.into())
             }
-
```
weikengchen, line 56: Explain: Since the `sub_unsafe` function cannot work in the setup mode, the new Div gadget no longer uses it. Therefore, this function is dropped.

gadgets/src/integers/uint/tests/uint128.rs
```diff
@@ -225,8 +226,8 @@ fn test_uint128_sub_constants() {
     for _ in 0..1000 {
         let mut cs = TestConstraintSystem::<Fr>::new();
 
-        let a: u128 = rng.gen_range(u128::max_value() / 2u128..u128::max_value());
-        let b: u128 = rng.gen_range(0u128..u128::max_value() / 2u128);
+        let a: u128 = rng.gen_range(u128::MAX / 2u128..u128::MAX);
```
weikengchen, line 229: Explain: This change is due to a suggestion by Rust, as `max_value()` is no longer the recommended way.

.circleci/config.yml
```diff
@@ -42,7 +42,7 @@ commands:
             - .cargo
 jobs:
 
-  rust-stable:
+  rust-stable-integration:
```
weikengchen, line 45: Explain: Since the stable and nightly only covers the integration tests beyond other tests, so we cut the workload here.

.circleci/config.yml
```diff
@@ -286,12 +286,207 @@ jobs:
       - clear_environment:
           cache_key: snarkvm-utilities-cache
 
+  snarkvm-algorithms-nightly:
```
weikengchen, line 289: Add all the nightly tests.

weikengchen:
This PR is ready for review. In addition, it contains nice refactoring of the Circle CI scripts to remove repeated tests and reduce the total runtime.

gadgets/src/integers/uint/tests/uint128.rs
```diff
@@ -370,6 +370,7 @@ fn test_uint128_div_constants() {
 }
 
 #[test]
+#[ignore]
```
ljedrz, line 373: why ignore this test? is it too slow or tested elsewhere?

ljedrz:
I'd prefer the CI changes to be split out into a dedicated PR (especially since there are changes to `RUST_MIN_STACK`), and the commits to be more fine-grained (e.g. plenty of squashing seems to have gone into https://github.com/AleoHQ/snarkVM/pull/252/commits/5c1c600513ef1ec6af3465f357d31db73f5b389b, making it harder to review the individual changes).

> the setup fails when dividing a u32 by a constant u32 (here, 1u32)

Is the problem from the description a registered GitHub issue that this PR can close? If yes, then it should be linked.

gadgets/src/integers/uint/tests/uint128.rs
weikengchen: Yes. On Circle CI this appears to require a large stack size and randomly fails sometimes (fails without a Rust error/panic, but just a sigkill). 

Given that it passes the local tests, and we do not have a clever way to handle it, it is therefore marked as `ignore`.

weikengchen:
Let me split the changes on CI to another PR. 

Since the code is migrated from a fork on version snarkVM 0.5.5, there is no much log of the step-by-step changes. The squashing is partly an effort to debug the Circle CI on a weird failure, which later seems to be due to the stack size, but those are not very meaningful, so they are squashed.

gadgets/src/integers/int/arithmetic/div.rs
```diff
@@ -63,191 +61,80 @@ macro_rules! div_int_impl {
...
+
+                let is_a_negative = self.bits.last().unwrap();
+                let is_b_negative = other.bits.last().unwrap();
+                let positive_result = is_a_negative.evaluate_equal(cs.ns(|| "compare_msb"), &is_b_negative)?;
+
+                // Get the absolute value of each number.
+                let a_absolute : <$gadget as Integer>::UnsignedGadgetType = {
+                    let negated_bits = self.bits.neg(cs.ns(||"neg_self_bits"))?;
+
+                    let mut absolute_value_bits = Vec::new();
```
ljedrz, line 87:
```suggestion
                    let mut absolute_value_bits = Vec::with_capacity(self.bits.len());
```

gadgets/src/integers/int/arithmetic/div.rs
```diff
@@ -63,191 +61,80 @@ macro_rules! div_int_impl {
...
+                            &self.bits[i],
+                        )?);
+                    }
+
+                    <$gadget as Integer>::UnsignedGadgetType::from_bits_le(&absolute_value_bits)
+                };
+                let b_absolute : <$gadget as Integer>::UnsignedGadgetType = {
+                    let negated_bits = other.bits.neg(cs.ns(||"neg_other_bits"))?;
+
+                    let mut absolute_value_bits = Vec::new();
```
ljedrz, line 102:
```suggestion
                    let mut absolute_value_bits = Vec::with_capacity(self.bits.len());
```

gadgets/src/integers/int/arithmetic/div.rs
```diff
@@ -63,191 +61,80 @@ macro_rules! div_int_impl {
...
+                let negated_quotient_bits = quotient.bits.neg(cs.ns(||"neg_quotient_bits"))?;
 
-                // set to one if we know result is division of the minimum number by itself
-                q = Self::conditionally_select(
-                    &mut cs.ns(|| "one_result"),
-                    &both_min,
-                    &allocated_one,
-                    &q,
-                )?;
+                let mut result_bits = Vec::<Boolean>::new();
```
ljedrz, line 126:
```suggestion
                let mut result_bits = Vec::<Boolean>::with_capacity(quotient.bits.len());
```

gadgets/src/integers/uint/arithmetic/div.rs
```diff
@@ -54,100 +56,89 @@ macro_rules! div_int_impl {
...
+                if !other.is_constant() {
+                    other.evaluate_equal(cs.ns(||"is_divisor_zero"), &Self::constant(0 as <$gadget as Integer>::IntegerType))?.enforce_equal(cs.ns(||"divisor_is_not_zero"), &Boolean::constant(false))?;
                 }
+                if other.value.is_some() {
```
ljedrz, line 63:
```suggestion
                if let Some(value) = other.value {
```

gadgets/src/integers/uint/arithmetic/div.rs
```diff
@@ -54,100 +56,89 @@ macro_rules! div_int_impl {
...
+                    let value = other.value.unwrap();
```
ljedrz, line 64:
```suggestion
```

weikengchen:
Let me do a PR split, do a final pass of the code logic, and reach out to @ljedrz for a follow-up. 

Likely the commits cannot be made more fine-grained this time, but I will be careful and do it the next time.

weikengchen (commented):
I did a pass of the logic, and I think it is ready for a follow-up review @ljedrz.

The CI changes are moved to a separate PR. Some cleanup has been made.

ljedrz (approved):
Thanks for splitting out the CI changes, the PR is more self-contained now.

The new `is_in_setup_mode` method and its related alternative computation make sense, and to the extent I understand the gadget setup, making signed `div` depend on the unsigned implementation and the removal of `sub_unsafe` are welcome changes too.

Would it be possible to add some piece(s) of code triggered by the Leo issue you linked as new snarkVM test cases?

weikengchen:
Let me add a test back (some are deleted during the CI failure debug). Basically, we can run the circuit in the setup mode, and it should not fail. And also, -128 / 2 will now be -64.

I think it suffices to add for int8 only.

```
#[test]
fn test_int8_div_corner_case_1() {
    let mut cs = TestConstraintSystem::<Fr>::new();

    let a: i8 = i8::MIN;
    let b: i8 = 2;

    let expected = match a.checked_div(b) {
        Some(valid) => valid,
        None => return,
    };

    let a_bit = Int8::alloc(cs.ns(|| "a_bit"), || Ok(a)).unwrap();
    let b_bit = Int8::alloc(cs.ns(|| "b_bit"), || Ok(b)).unwrap();

    let r = a_bit.div(cs.ns(|| "division"), &b_bit).unwrap();

    assert!(cs.is_satisfied());
    assert_eq!(r.value.unwrap(), expected);
    check_all_allocated_bits(expected, r);
}

#[test]
#[should_panic]
fn test_int8_div_corner_case_2() {
    let mut cs = TestConstraintSystem::<Fr>::new();

    let a: i8 = i8::MIN;
    let b: i8 = -1;

    let a_bit = Int8::alloc(cs.ns(|| "a_bit"), || Ok(a)).unwrap();
    let b_bit = Int8::alloc(cs.ns(|| "b_bit"), || Ok(b)).unwrap();

    let _ = a_bit.div(cs.ns(|| "division"), &b_bit).unwrap();
}

#[test]
fn test_int8_div_setup_mode() {
    let mut cs = KeypairAssembly::<Bls12_377> {
        num_public_variables: 0,
        num_private_variables: 0,
        at: vec![],
        bt: vec![],
        ct: vec![],
    };

    let a: i8 = 1i8;
    let b: i8 = 2i8;

    let a_bit = Int8::alloc(cs.ns(|| "a_bit"), || Ok(a)).unwrap();
    let b_bit = Int8::alloc(cs.ns(|| "b_bit"), || Ok(b)).unwrap();

    let _ = a_bit.div(cs.ns(|| "division"), &b_bit).unwrap();
}
```

ljedrz:
>I think it suffices to add for int8 only.

Ideally we would be able to have more arithmetic tests happen for all sizes, but last time I was looking at it it was not feasible; with the simplification of `Div` in this PR we might soon be able to make it happen. If I recall correctly, some of the 128-bit calculations were especially different from the others.

weikengchen:
> > I think it suffices to add for int8 only.
> 
> Ideally we would be able to have more arithmetic tests happen for all sizes, but last time I was looking at it it was not feasible; with the simplification of `Div` in this PR we might soon be able to make it happen. If I recall correctly, some of the 128-bit calculations were especially different from the others.

It might be more challenging, as this `Div` gadget is much more expensive. At this moment, the SIGKILL issue of the Circle CI is still bothering me.

We may, in the future, replace our current implementations of integers, as the current implementation is too heavy.

PR #254, opened 2021-07-13
A candidate new CI procedure

weikengchen:
## Motivation

This PR replaces the CI procedure for development efficiency. 

## Summary of the changes:

- Move check-style to Circle CI, since GitHub CI sometimes has network failures. This change may be reverted in the future.
- Increase the Circle CI Rust minimal stack size to 64MB, which seems to be needed for certain tests.
- Split the tests as follows: 
   * Rust-nightly is broken down into individual crates' nightly and an integration-nightly. 
   * Rust-stable is replaced by an integration test on stable.
   * The integration test, which consists of the testnet1 part and the testnet2 part, is further split.
   * Names of the tests have been simplified.
- Remove `codecov` test temporarily since it is broken at this moment and seems to invariably demand a longer test time. It will be brought back later.

Expected that a full test procedure is now about 10m instead of 25m or longer.


weikengchen:
Adding @Protryon for a review since he has dealt with snarkOS min stack size and may provide some feedback on the choice of 64MB. The 64MB seems to be sufficient for this moment although some tests are still too heavy to run.

howardwu:
@weikengchen one of things we talked about was only supporting stable (and foregoing nightly) given the amount of breaking changes nightly tends to introduce, and frequently require two variants of syntax.

weikengchen:
Updated. The nightly tests are dropped.

.circleci/config.yml
```diff
@@ -282,29 +266,42 @@ jobs:
...
           name: Build and run tests
           no_output_timeout: 35m
-          command: cd utilities && RUST_MIN_STACK=8388608 cargo test
+          command: cd utilities && RUST_MIN_STACK=67108864 cargo test
       - clear_environment:
           cache_key: snarkvm-utilities-cache
 
+  fmt:
+    docker:
+      - image: howardwu/snarkos-ci:2021-03-25
```
howardwu, line 275:
```suggestion
      - image: cimg/rust:1.53.0
```

weikengchen: Oh nope! that image is for the `nightly` because our `fmt` is done over it.

howardwu (approved):

PR #255, opened 2021-07-14
Fix `GroupEncryptionGadget` Signature Verification

raychu86:
## Motivation

There was a minor issue with the signature verification check of the `GroupEncryptionGadget`, where the message length was not included in the Poseidon gadget hash input. 

## Test Plan

Tests have been added to check that the signature verification of the `GroupEncryptionGadget` is correct.

### DIFF
--- a/gadgets/src/algorithms/signature/group.rs
+++ b/gadgets/src/algorithms/signature/group.rs
@@ -398,6 +398,7 @@ where
             &claimed_prover_commitment
                 .to_constraint_field(cs.ns(|| "convert claimed_prover_commitment into field elements"))?,
         );
+        hash_input.push(FpGadget::<F>::Constant(F::from(message.len() as u128)));
         hash_input.extend_from_slice(&message.to_constraint_field(cs.ns(|| "convert message into field elements"))?);
 
         // Compute the hash on the base field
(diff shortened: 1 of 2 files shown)


raychu86:
@weikengchen Is there a reason the message length needs to be included in the hash input for the schnorr signatures?

weikengchen:
Thanks for catching this! 

The reason is that when we convert messages into field elements, different messages (those ending with 00....0) may result in the same vector of field elements. Depending on the application, this is sometimes okay, sometimes not.

In addition, when Poseidon absorbs two vectors of field elements (a, b, c) and (a, b, c, 0), the result may be the same, which is yet another concern that such input may not actually be equivalent.

Our current approach is to add the message length. An alternative is to pad the message with two bits "0" "1" at the end.

weikengchen (approved):

howardwu (approved):

howardwu:
@weikengchen I believe it would be more efficient to create a new Vec of `vec![ message.len() as u128, message].concat()` in order to possibly safe up to 1 field element for hash input.

howardwu:
There's also another thing we could possibly do, see #256

PR #259, opened 2021-07-14
A memory-efficient constraint system checker for CI

weikengchen:
## Motivation

This PR supplies a memory-efficient test constraint system, called `TestConstraintChecker`, that does not compute the full paths and does not store constraints, but still preserves the functionality of checking if a constraint system is satisfied.

## Test Plan

The code has been beta-tested on a few tests.

We will see if the CI time improves.

### DIFF
--- a/gadgets/src/integers/uint/tests/uint128.rs
+++ b/gadgets/src/integers/uint/tests/uint128.rs
@@ -373,7 +373,7 @@ fn test_uint128_div_constants() {
 fn test_uint128_div() {
     let mut rng = XorShiftRng::seed_from_u64(1231275789u64);
 
-    for _ in 0..10 {
+    for _ in 0..2 {
         let mut cs = TestConstraintChecker::<Fr>::new();
 
         let a: u128 = rng.gen();
@@ -393,6 +393,22 @@ fn test_uint128_div() {
         assert!(cs.is_satisfied());
         assert!(r.value == Some(expected));
         check_all_allocated_bits(expected, r);
+
+        // This test would not work in `TestConstraintChecker`, so we temporarily disable this
+        // part of the test.
+        //
+        // Flip a bit_gadget and see if the division constraint still works
+        /*
+        if cs
+            .get("division/r_sub_d_result_0/allocated bit_gadget 0/boolean")
+            .is_zero()
+        {
+            cs.set("division/r_sub_d_result_0/allocated bit_gadget 0/boolean", Fr::one());
+        } else {
+            cs.set("division/r_sub_d_result_0/allocated bit_gadget 0/boolean", Fr::zero());
+        }
+
+        assert!(!cs.is_satisfied());*/
     }
 }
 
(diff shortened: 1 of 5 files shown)


weikengchen:
It appears that the CI time does not improve. 

But the good thing is that memory usage is certainly reduced. Now it can even run `test_uint128_div` and `test_uint128_pow`, though both are time-consuming.

Let us have this PR sit for a while to collect some feedback.

r1cs/src/test_constraint_checker.rs
```diff
@@ -0,0 +1,173 @@
...
+                })
+                .sum::<F>()
+        };
+
+        let a = eval_lc(a(LinearCombination::zero()).0);
+        let b = eval_lc(b(LinearCombination::zero()).0);
+        let c = eval_lc(c(LinearCombination::zero()).0);
+
+        if a * b != c {
+            self.found_unsatisfactory_constraint = true;
```
ljedrz, line 125: nit: this can be moved inside the `if` block below

r1cs/src/test_constraint_checker.rs
```diff
@@ -0,0 +1,173 @@
...
+
+    fn push_namespace<NR: AsRef<str>, N: FnOnce() -> NR>(&mut self, name_fn: N) {
+        let new = name_fn().as_ref().to_string();
+        assert!(!new.contains('/'), "'/' is not allowed in names");
+
+        self.segments.push(new)
+    }
+
+    fn pop_namespace(&mut self) {
+        self.segments.pop();
```
ljedrz, line 146: doesn't the memory grow a lot if the variables associated with a namespace aren't removed from their respective vectors when a namespace is popped?

ljedrz: if each segment was associated with 2 values corresponding to the number of private and public variables, they could be used to truncate their respective vectors; this is under the assumption that once a namespace segment is popped, the variables that were declared within it can go out of scope.

weikengchen: This might be less an issue here, as (1) the segments are stored only in one place, and (2) the segment is not effectively associated with the counts of private and public variables, as there is only a global count of them.

weikengchen: (which also means that this checker now offers much less functionality for looking into the cost, so it may be just a checker for the `test`)

weikengchen:
Going to do some cleanup of the checker. I will update soon.

r1cs/src/test_constraint_checker.rs
weikengchen: Resolved.

weikengchen:
@ljedrz It is ready for a sequential review. 

I decided to use this sparsely, currently only for u128::div and u128::pow, since the memory issue only appears in those a few tests, and `TestConstraintChecker` does not enable one to modify the constraint matrix and check whether it is correct, so it cannot be used everywhere.

gadgets/src/integers/uint/tests/uint128.rs
```diff
@@ -394,18 +393,6 @@ fn test_uint128_div() {
...
-        if cs
-            .get("division/r_sub_d_result_0/allocated bit_gadget 0/boolean")
-            .is_zero()
-        {
-            cs.set("division/r_sub_d_result_0/allocated bit_gadget 0/boolean", Fr::one());
-        } else {
-            cs.set("division/r_sub_d_result_0/allocated bit_gadget 0/boolean", Fr::zero());
-        }
-
-        assert!(!cs.is_satisfied());
```
ljedrz, line 408: Perhaps we could comment this out (including the reasoning for doing it) instead of removing it, just in case the `TestConstraintSystem` eventually becomes fast enough to use on `u128`?

gadgets/src/integers/uint/tests/uint128.rs
```diff
@@ -454,22 +440,4 @@ fn test_uint128_pow() {
...
-            Fr::one(),
-        );
-    } else {
-        cs.set(
-            "exponentiation/multiply_by_self_0/partial_products/result bit_gadget 0/boolean",
-            Fr::zero(),
-        );
-    }
-
-    assert!(!cs.is_satisfied());
```
ljedrz, line 474: Ditto - if there is a chance we'll use `TestConstraintSystem` for `i128` in the future, I'd just comment this out instead. Otherwise, it can be removed.

weikengchen: Good points. Let me comment them out and leave a note on why they are temporarily hidden.

ljedrz (approved):
LGTM :ok_hand:; will require a bit of squashing to remove the CI-related commits, but other than that, it looks like a useful addition for "heavy" tests.

PR #262, opened 2021-07-15
Simplify rebuild

Protryon:
This PR removes the need to pass in all existing commitments to rebuild -- only the length is needed. Didn't alter the storage model in `.integration` since it is being changed in snarkOS.


weikengchen (approved):
Makes sense, as the rebuilding only uses `old_leaves.len()`.

ljedrz (approved):
Good catch!

PR #263, opened 2021-07-15
Implement CRH and CRHGadget for Poseidon

raychu86:
## Motivation

This PR implements the `CRH` trait for `PoseidonCryptoHash` and the `CRHGadget` trait for `PoseidonCryptoHashGadget`.  This is needed to enable Merkle trees to use the Poseidon hash function.

One thing that should be updated is the CRH attribute, `INPUT_SIZE_BITS`, which is currently hard-coded for the Poseidon impl.

## Test Plan

Tests have been added to check that Poseidon usage in Merkle Trees is valid.


howardwu:
`INPUT_SIZE_BITS` can be a const generic, but given our priorities I think this is fine for now

weikengchen (approved):
The code looks good. 

One thing to double-check: although we have implemented the FromBytes and ToBytes for PoseidonParameters, they are not going to be used anywhere right? Since that would increase unnecessary overhead.

weikengchen:
It would be preferred to make the FromBytes and ToBytes of PoseidonParameters as dummies, as we expect the parameters to come directly from the field.

weikengchen:
We may in the future extend the syntax of CRH. The current CRH syntax still forces the intermediate CRH values to be converted into bytes and then converted back.

It is slightly challenging, as Poseidon CRH is very different from existing CRHs, and our Merkle tree design is specialized for existing CRHs.

howardwu:
Agreed on all of the points you're making Weikeng. I'll open a Github issue to track this discussion.

In the meantime Ray, let's proceed and make sure to resolve this before `testnet2` goes to prod.

howardwu (approved):

howardwu:
#264

raychu86:
Following up on your question Weikeng, the ToBytes and FromBytes impls for the PoseidonParameters aren't actually being used anywhere.

It took me maybe a few minutes to write the serializations as it was quite trivial, but agreed that we may want to just make them dummies.

PR #268, opened 2021-07-16
[Ignore] `groth16` private branch

howardwu:
[Ignore] `groth16` private branch

## Deltas

`None`


algorithms/src/signature/schnorr.rs
```diff
@@ -89,8 +89,7 @@ impl<G: ProjectiveCurve + CanonicalSerialize + CanonicalDeserialize> ToBytes for
     /// Writes the x-coordinate of the encryption public key.
     #[inline]
     fn write_le<W: Write>(&self, mut writer: W) -> IoResult<()> {
-        let affine = self.0.into_affine();
-        let x_coordinate = affine.to_x_coordinate();
+        let x_coordinate = self.0.to_x_coordinate();
```
Pratyush, line 92: This is not always sound; it's sound only for the Edwards curves we're using. To make sure this works, I would change the trait bound to be `P: TEParameters`, and hard code the curve type to be Edwards

PR #276, opened 2021-07-18
Removes `program_commitment` and `local_data_root` from `Transaction`, moves transaction signatures to offline phase, increases `memo` size to 64 bytes

howardwu:
## Motivation

- Removes `program_commitment` and `local_data_root` from `Transaction`
- Removes `ledger_digest`, `program_commitment`, and `local_data_root` from transaction signatures
- Increases `memo` size to 64 bytes to support two record view keys or two account view keys

## Approach 

This PR preserves `program_commitment` and `local_data_root` in the `InnerCircuitVerifierInput`, however it removes it from `OuterCircuitVerifierInput`. Instead, these two items are allocated as witness in the outer circuit only.

This change then allows us to remove `program_commitment` and `local_data_root` from `Transaction`, and transaction signatures.

In addition, `ledger_digest` is removed from transaction signatures, to allow for offline signing of the transaction.

## Notes

This PR will require resampling parameters.


Pratyush:
Isn't `ledger_digest` required to create the transaction SNARKs in the first place? So it's required in the offline phase anyway, no?

howardwu:
That's right. So the offline phase (which requires a private key) is purely for signing. The online phase (the delegable DPC model) executes the inner and outer SNARKs with a ledger digest. This approach makes transaction generation a handoff procedure from caller to prover to chain.

Pratyush:
For delegable DPC don't we have to sign the transaction? Or am I missing something?

howardwu:
That was the original construction. I think it was around a year ago, we discussed making the outer proof generation a handoff procedure, by forgoing signing the final proof. Then we discussed removing the ledger digest from the signed message to allow for offline signing.

Pratyush:
I see, so now the signature is only over the commitments and serial numbers? (and maybe memo and some other non-proof stuff)

howardwu:
No worries :) Yes, here's the signature message now:

(Network ID, Serial Numbers, Commitments, Memorandum, Value Balance)

howardwu:
Resampling parameters now.

Pratyush (approved):

weikengchen (approved):
Did a pass of the files. I think it is good.

PR #279, opened 2021-07-20
Implements BHP commitments, switches all Pedersen commitments to BHP, updates testnet1 program SNARK to Groth16 and ID CRHs to Poseidon

howardwu:
## Motivation

- Implements BHP commitments
- Switches all Pedersen commitment usage in DPC `Parameters` to BHP commitments
- Updates `testnet1` program SNARK to Groth16, inner and program ID CRHs to Poseidon

## Test Plan

- Adds commitment parameter serialization tests for BHP native
- Adds native and gadget equivalence tests for BHP commitments, BHP compressed commitments, and Pedersen compressed commitments

## Notes

This PR resamples parameters for both `testnet1` and `testnet2`


howardwu:
Should we rename all Bowe Hopwood Pedersen primitives to BHP? e.g. BoweHopwoodPedersenCompressedCommitmentGadget is a *mouthful*

weikengchen:
Just to add that changing the DPC trait impl to using `PoseidonCryptoHash` would not give the full speed boost---since by default, it would call the bit interface of Poseidon and still incurs an overhead of bit conversions. Certain changes to make them into field elements first and then `hash_field_elements` are needed. I can take a look tomorrow and probably push to this branch (or after it is merged).

weikengchen:
> Should we rename all Bowe Hopwood Pedersen primitives to BHP? e.g. BoweHopwoodPedersenCompressedCommitmentGadget is a _mouthful_

I totally agree. Indeed, if its parameters are more easy to specify for end users, it can just be a drop-in replacement of Pedersen.

howardwu:
> Just to add that changing the DPC trait impl to using `PoseidonCryptoHash` would not give the full speed boost---since by default, it would call the bit interface of Poseidon and still incurs an overhead of bit conversions. Certain changes to make them into field elements first and then `hash_field_elements` are needed. I can take a look tomorrow and probably push to this branch (or after it is merged).

That sounds great, I will finish up my changes in here and after completing them hand it off to you.

(I will then resample parameters after the changes are ready)

weikengchen:
Yes, I suspect that the CI for integration testnet1 fails because the keys do not match.

howardwu:
I will resample parameters and update constraint sizes based on these changes just so we have a baseline here.

weikengchen (approved):
The changes look standard and reasonable.

Pratyush:
btw, separately, I would just drop the `Compressed` part from BHP; the specification of BHP in the Zcash spec includes the compression (i.e. projection to one coordinate) by default, because it's only defined over TE curves. I've made this change in arkworks already.

PR #284, opened 2021-07-21
Tracking issue for trimmable SRS

weikengchen:
## Motivation

This PR would consist of implementations toward making SRS trimmable. 

There are many things involved. We are currently at the first stage.
- make PolyCommit support SRS that is trimmable.
It is not finished yet.

This seems to already involve much changes, since SRS is not only about powers of `g`, but also:
- powers of `gamma_g` which in Marlin you only need 2 elements.
- negative powers of `g` in G2 which is not needed in Marlin but used in Sonic PC.
- powers of `g` but with degree `max_degree - i` which is used for degree bounds. In Marlin, all it needs has the shape of `2^t - 2`. 

Currently the setup is too long: it generates, up to the max_degree, of all of these.
And the SRS is too long: it contains all of these.

If we look at Marlin specifically, the actual effort of setup should be only 1/3, and the SRS would also be 1/3. 

Then, you can trim it further if you know that you only index and prove circuits of a specific level of complication.

## Test Plan

Will need several tests to ensure that the PC proof generated from the trimmed SRS matches those from the old, full SRS. Especially, pay attention to the hiding bounds.

### DIFF
--- a/marlin/src/parameters.rs
+++ b/marlin/src/parameters.rs
@@ -40,6 +40,16 @@ use std::{
     },
 };
 
+pub struct UniversalSetupConfig {
+    pub supported_degree : usize,
+    pub supported_hiding_bound : usize,
+}
+
+pub struct UniversalSetupParameters<E: PairingEngine> {
+    pub powers_of_g : Vec<E::G1Affine>,
+
+}
+
 #[derive(Derivative)]
 #[derivative(Clone(bound = ""))]
 #[derive(Debug, CanonicalSerialize, CanonicalDeserialize)]
--- a/algorithms/src/traits/snark.rs
+++ b/algorithms/src/traits/snark.rs
@@ -60,6 +60,13 @@ pub trait SNARK {
         unimplemented!()
     }
 
+    fn trim_universal_setup_parameters(
+        _long: &Self::UniversalSetupParameters,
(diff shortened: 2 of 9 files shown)


weikengchen:
Adding that we also need infrastructure for proving key.

Why? Since the majority of the proving key is the PC committer key, which is circuit-independent. So, indeed, `pk` would be also succinct.

marlin/src/parameters.rs
```diff
@@ -40,6 +40,16 @@ use std::{
...
 };
 
+pub struct UniversalSetupConfig {
+    pub supported_degree : usize,
+    pub supported_hiding_bound : usize,
+}
+
+pub struct UniversalSetupParameters<E: PairingEngine> {
+    pub powers_of_g : Vec<E::G1Affine>,
+
```
ljedrz, line 50:
```suggestion

```

ljedrz:
I started browsing the PR, but it seems like it's still a work in progress, right?

weikengchen:
Yes, it is halfway.

aphelionz:
@weikengchen Please mark it as a draft if you're still able to, that way we'll know not to jump and start reviewing

weikengchen:
This has been partially done in #317. So I closed this issue.

PR #285, opened 2021-07-21
Gracefully return an error message upon division by zero for signed integers

weikengchen:
## Motivation

This is to fix the bug caught by fuzzing in Leo: https://github.com/AleoHQ/leo/issues/1170

I personally consider it a minor bug, since when this happens, it should either be Rust panicking or Leo showing an error message. 

This PR does the latter instead of the former, as it may confuse the users and provide insufficient debug information.

## Test Plan

A similar test appears in the unsigned integer part. So this should work fine. I will wait for the CI.

## Related PRs

https://github.com/AleoHQ/leo/issues/1170

### DIFF
--- a/gadgets/src/integers/int/arithmetic/div.rs
+++ b/gadgets/src/integers/int/arithmetic/div.rs
@@ -61,6 +61,13 @@ macro_rules! div_int_impl {
                 // else
                 //    !Q                      -- negative result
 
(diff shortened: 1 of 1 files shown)


howardwu (approved):

ljedrz:
Whenever an edge case like this is detected, including via fuzzing, I would recommend adding it to the repo as a test case.

PR #286, opened 2021-07-21
Use Poseidon for Program/Inner Circuit ID in the field-element way

weikengchen:
## Motivation

This PR will be into #279. It replaces the testnet1's calls to Poseidon from the byte manner to the field-element manner.

Expect the integration tests would fail due to the need to resample parameters and changing constraint counts.

## Test Plan

Let us wait for the CI.

## Related PRs

#279

### DIFF
--- a/.integration/tests/dpc_testnet1.rs
+++ b/.integration/tests/dpc_testnet1.rs
@@ -36,6 +36,7 @@ use snarkvm_utilities::{to_bytes_le, FromBytes, ToBytes};
 use itertools::Itertools;
 use rand::SeedableRng;
 use rand_chacha::ChaChaRng;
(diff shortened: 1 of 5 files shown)


weikengchen:
Can confirm that after resampling outer_snark parameters for testnet1 locally, the integration test passes.

weikengchen:
actually let me resample the parameters again---I am expecting the constraint counts to go down further in #b1ef668.

howardwu (approved):

howardwu:
Sounds good. I can confirm it looks like all instances of the program executions and inner circuit VKs have been switched from `to_bytes` to field elements.

weikengchen:
Clear. I think it is good. The integration-testnet1 should disappear once parameters are resampled.

PR #287, opened 2021-07-21
Reduce the number of allocations used to deserialize snark proofs

ljedrz:
This PR makes the deserialization buffer handling in `Proof::read` for snarks faster (fewer allocations) and more idiomatic.

This was detected when profiling the snarkOS storage validator, as the `Proof::read` function was called a lot of times.


howardwu (approved):
LGTM

PR #288, opened 2021-07-21
Increase record payload size from 32 bytes to 128 bytes

raychu86:
## Motivation

This PR increases the record payload size from 32 bytes to 128 bytes. 

Two parameters need to be updated because of this change:

1. Record commitment CRH window is increased because the record size increases from `201` bytes to `297` bytes.
2. Encrypted record CRH window is increased because the number group elements in the encoded record increases from `7` to `10` elements.

  These two changes increase the inner circuit constraint size from `381519` to `436695` constraints.

## Test Plan

The record encoding and encryption tests pass. 

The integration tests are failing because the inner circuit parameters need to be regenerated.


dpc/src/record/payload.rs
```diff
@@ -18,19 +18,21 @@ use snarkvm_utilities::{FromBytes, ToBytes};
...
+pub struct Payload([u8; PAYLOAD_SIZE]);
 
 impl Payload {
     pub fn to_bytes(&self) -> &[u8] {
         &self.0[..]
     }
 
     pub fn from_bytes(bytes: &[u8]) -> Self {
-        assert_eq!(bytes.len(), 32);
+        assert_eq!(bytes.len(), PAYLOAD_SIZE);
```
ljedrz, line 32: this assertion is unnecessary, as `copy_from_slice` performs this check internally; the `payload` binding below could then be
```
let mut payload = [0u8; PAYLOAD_SIZE];
```

dpc/src/record/payload.rs
```diff
@@ -18,19 +18,21 @@ use snarkvm_utilities::{FromBytes, ToBytes};
...
 
-        let mut payload = [0u8; 32];
-        payload.copy_from_slice(&bytes[0..32]);
+        let mut payload = [0u8; 128];
+        payload.copy_from_slice(&bytes[0..PAYLOAD_SIZE]);
```
ljedrz, line 35:
```suggestion
        payload.copy_from_slice(bytes);
```
`bytes` is already a slice and it's guaranteed to be of `PAYLOAD_SIZE` due to the use of `copy_from_slice` and `payload` being of that size

dpc/src/record/payload.rs
```diff
@@ -18,19 +18,21 @@ use snarkvm_utilities::{FromBytes, ToBytes};
...
         &self.0[..]
     }
 
     pub fn from_bytes(bytes: &[u8]) -> Self {
-        assert_eq!(bytes.len(), 32);
+        assert_eq!(bytes.len(), PAYLOAD_SIZE);
 
-        let mut payload = [0u8; 32];
-        payload.copy_from_slice(&bytes[0..32]);
+        let mut payload = [0u8; 128];
```
ljedrz, line 34:
```suggestion
        let mut payload = [0u8; PAYLOAD_SIZE];
```

dpc/src/record/payload.rs
```diff
@@ -53,3 +55,9 @@ impl FromBytes for Payload {
         Ok(Self(FromBytes::read_le(&mut reader)?))
     }
 }
+
+impl Default for Payload {
+    fn default() -> Self {
+        Self([0u8; 128])
```
ljedrz, line 61:
```suggestion
        Self([0u8; PAYLOAD_SIZE])
```

dpc/src/record/tests.rs
```diff
@@ -46,7 +46,7 @@ fn test_record_serialization() {
 
             let sn_nonce_input: [u8; 32] = rng.gen();
             let value = rng.gen();
-            let payload: [u8; 32] = rng.gen();
```
ljedrz, line 49: nit: `PAYLOAD_SIZE` could be imported here

dpc/src/record/tests.rs
```diff
@@ -90,7 +90,7 @@ fn test_record_encryption() {
...
```
ljedrz, line 93: ditto

dpc/src/record/tests.rs
```diff
@@ -46,7 +46,7 @@ fn test_record_serialization() {
...
+            let payload = (0..128).map(|_| rng.gen()).collect::<Vec<u8>>();
```
ljedrz, line 49: nit: sampling of many values can be done in a more `clippy`-  and potentially performance-friendly way using
```
let mut payload = [0u8; PAYLOAD_SIZE]; // this also avoids the Vec allocation
rng.fill(&mut payload);
```
or
```
(&mut rng).sample_iter(Standard).take(PAYLOAD_SIZE).collect::<Vec<u8>>()
```
note: `Rng::gen` is using the standard distribution internally, while `Rng::fill` uses a uniform one

howardwu (approved):
LGTM, `PAYLOAD_SIZE` should be switched to a generic that takes `C: Parameters` with `C::RECORD_PAYLOAD_SIZE`, but this can wait for now

howardwu:
Resampled and tests are passing.

PR #289, opened 2021-07-22
Run testnet 2 integration tests on CI 

niklaslong:
The circle-ci "integration-testnet2" job was running testnet 1 integration tests.


ljedrz (approved):
Good catch :+1:

weikengchen:
Good catch! (I fixed this in the testnet2 branch but forgot that it is also in the master.)

PR #290, opened 2021-07-22
Use more macros in the bytes module

ljedrz:
This PR greatly shortens the `bytes` module in `snarkvm-utilities` by utilizing macros to implement `FromBytes` and `ToBytes`.


howardwu (approved):
LGTM, thank you for impl'ing this!

PR #293, opened 2021-07-24
Add ECIES Poseidon encryption

weikengchen:
## Motivation

This PR supples the ECIES Poseidon encryption, which is likely going to replace group encryption. 

Currently the group encryption has certain challenges: (1) elligator is not performing ideally for a few reasons and (2) it seems to complicate our serialization.

This one supplies an ECIES Poseidon encryption, which is likely to be faster than the group encryption, and can work securely when the same key is also used with Schnorr signatures. 

There is a possibility of using ElGamal hybrid encryption later, which will also be using Poseidon for the symmetric encryption part, and can reuse the code here.

## Test Plan

The ECIES Poseidon uses the same tests as used by group encryption.

### DIFF
--- a/algorithms/src/encryption/ecies_poseidon.rs
+++ b/algorithms/src/encryption/ecies_poseidon.rs
@@ -1,0 +1,191 @@
+// Copyright (C) 2019-2021 Aleo Systems Inc.
+// This file is part of the snarkVM library.
+
+// The snarkVM library is free software: you can redistribute it and/or modify
+// it under the terms of the GNU General Public License as published by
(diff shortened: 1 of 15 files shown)


weikengchen:
Putting this into DPC will require a massive change to the record part, since the record part has been specifically designed to embed data into group elements. It would be done after a few other PRs that unify the testnet1 and testnet2 DPC.

howardwu (approved):

howardwu:
LGTM, this PR doesn't plug into the existing record model yet, so there is no need to resample parameters for this PR.

PR #294, opened 2021-07-24
Adds `TryFrom` implementations for account `ViewKey` and `Address`

howardwu:
## Motivation

Adds `TryFrom` implementations for account `ViewKey` and `Address`

## Test Plan

- Adds tests for the new `TryFrom` methods
- Updates existing tests to include failure tests for derivation from strings


ljedrz (approved):
LGTM :+1:

PR #295, opened 2021-07-25
Initial attempts to merge the Testnet1 and Testnet2 traits

weikengchen:
## Motivation

Our DPC has two sets of parameter traits:
    Testnet1Components and Testnet2Components 

Ideally this should be reduced into one trait, which allows bringing the Testnet1 and Testnet2 together.

This PR makes the initial attempts, which removes several differences between `Testnet1Components` and `Testnet2Components`.

At this moment, they look like:

```
pub trait Testnet1Components: Parameters {
    type InnerSNARKGadget: SNARKVerifierGadget<Self::InnerSNARK, Input = Vec<Boolean>>;

    type ProgramSNARK: SNARK<
        ScalarField = Self::InnerScalarField,
        BaseField = Self::OuterScalarField,
        VerifierInput = ProgramLocalData<Self>,
    >;

    type ProgramSNARKGadget: SNARKVerifierGadget<Self::ProgramSNARK, Input = Vec<Boolean>>;
}
pub trait Testnet2Components: Parameters {
    type InnerSNARKGadget: SNARKVerifierGadget<Self::InnerSNARK, Input = Vec<Boolean>>;

    type ProgramSNARK: SNARK<
        ScalarField = Self::InnerScalarField,
        BaseField = Self::OuterScalarField,
        VerifierInput = ProgramLocalData<Self>,
    >;

    type ProgramSNARKGadget: SNARKVerifierGadget<
        Self::ProgramSNARK,
        Input = NonNativeFieldVar<Self::InnerScalarField, Self::OuterScalarField>,
    >;
}
```

As one can see, the only issue now is that the Input trait for ProgramSNARKGadget is different. This would be done in a followup PR, as it may need some changes to the Groth16 gadget and may break testnet1.

## Test Plan

Passes some local tests. Expect to have some changes needed for integration tests and the parameters setup.

### DIFF
--- a/algorithms/src/traits/snark.rs
+++ b/algorithms/src/traits/snark.rs
@@ -38,7 +38,7 @@ pub trait SNARK {
 
     // We can specify their defaults to `()` when `associated_type_defaults` feature becomes stable in Rust
     type UniversalSetupConfig: Clone;
-    type UniversalSetupParameters: Clone;
+    type UniversalSetupParameters: FromBytes + ToBytes + Clone;
 
     type VerifyingKey: Clone
         + ToBytes
--- a/dpc/src/testnet2/mod.rs
+++ b/dpc/src/testnet2/mod.rs
@@ -30,13 +30,7 @@ pub mod parameters;
 
 use crate::{Parameters, ProgramLocalData};
 use snarkvm_algorithms::prelude::*;
-use snarkvm_fields::ToConstraintField;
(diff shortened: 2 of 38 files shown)


weikengchen:
This one may face some blockers that require some other changes to DPC first, and is better to be merged at this shape now and let #297 merge through. Then after changes to DPC, another PR will follow up.

howardwu (approved):
LGTM

PR #297, opened 2021-07-25
Removes `serial_number_nonce_randomness` from `DPC` execution model

howardwu:
## Motivation

Removes serial number nonce randomness


howardwu:
This PR will await PR #295 given both PR's touch the `DPC` struct

weikengchen (approved):

PR #299, opened 2021-07-25
Use `alloc_constant` for parameters for crypto primitives

weikengchen:
## Motivation

This PR makes two changes:
- force all the crypto primitives parameters to be allocated only through `alloc_constant`, and `alloc`/`alloc_input` is now "unimplemented" and will cause a panic
- update the DPC to allocate the cryptographic primitive gadgets using `alloc_constant`
- remove the serialization of cryptographic primitive parameters (which at this moment is empty).

## Test Plan

Some tests have been executed in `gadgets` and related dirs.


howardwu (approved):

howardwu:
LGTM

PR #300, opened 2021-07-25
Unify DPC programs

howardwu:
## Motivation

- Unifies DPC `NoopProgram` and `NoopProgramCircuit` into one version
- Removes `ProgramSNARKUniversalSRS` struct
- Introduces `SRS` enum for `SNARK` setup
- Introduce `Parameters::program_srs()` in DPC
- Adds `CryptoRng` on all `SNARK` trait methods


howardwu:
This PR still needs clean up.

marlin/src/parameters.rs
```diff
@@ -14,7 +14,7 @@
 // You should have received a copy of the GNU General Public License
 // along with the snarkVM library. If not, see <https://www.gnu.org/licenses/>.
 
-use crate::{MarlinTestnet1, ProvingKey, VerifyingKey, SRS};
+use crate::{MarlinTestnet1, ProvingKey, VerifyingKey, URS};
```
weikengchen, line 17: Note: we use keep the name SRS, since URS often means uniform random string, which is a string that has no structured. Marlin still uses SRS, not an URS.

dpc/src/program/noop_program.rs
```diff
@@ -26,40 +26,40 @@ use crate::{
 };
 use snarkvm_algorithms::prelude::*;
 use snarkvm_parameters::{
-    testnet1::{NoopProgramSNARKPKParameters, NoopProgramSNARKVKParameters},
+    testnet2::{NoopProgramSNARKPKParameters, NoopProgramSNARKVKParameters},
```
weikengchen, line 29: Note: just to confirm that now DPC Noop Program for both testnet1 and testnet2 uses the testnet2's noop program parameters

weikengchen:
On a side note, I am working on a PR that resolves the other part of unifying---the outer circuit, now doing some tests. That one is likely to unify the OuterCircuit. These two PRs very likely would not conflict with each here for the large part.

marlin/src/parameters.rs
howardwu: Will clean up, had name collision with `SRS` struct

dpc/src/program/noop_program.rs
howardwu: Good catch, will add parameterization for this.

marlin/src/parameters.rs
weikengchen: `UnivesalSRS` can be a good name.

weikengchen:
I will point my PR to this branch and then merge the changes from it. 

On my side, OuterCircuit has just unified.

posw/src/lib.rs
```diff
@@ -117,14 +117,14 @@ mod tests {
 
     #[test]
     fn test_posw_marlin() {
-        let rng = &mut XorShiftRng::seed_from_u64(1234567);
+        let rng = &mut ChaChaRng::seed_from_u64(1234567);
```
weikengchen, line 120: Warning: This change causes the CI time for posw to go up a lot.

weikengchen: NOPE <- I guess ChaChaRng is not the reason. There is likely another reason why posw test becomes more expensive.

howardwu: It's likely due to the overpaying of to_bytes_le.

dpc/src/program/noop_program.rs
weikengchen: I will work on this. I think this can be solved by just including a name for the testnet (e.g., "testnet1") in the DPC Parametes.

marlin/src/parameters.rs
weikengchen: Solved in #301.

dpc/src/program/noop_program.rs
weikengchen: Solved in #301.

posw/src/lib.rs
weikengchen: I guess we can handle this in a separate PR. Let me think about a way to avoid `to_bytes_le`. Ideally we want some sort of zero-copying here. 

The easiest solution is this: We only have Marlin SRS so far, so we can just...use MarlinSRS. But I need to handle the trait parameters.

weikengchen: Solved in #301. 10m now down to 3m.

weikengchen:
It seems that all the remaining issues are solved in #301, though a resampling of the parameters is needed.

weikengchen:
I can help fix the clippy issues when you are working on parameter samplings.

howardwu:
Parameters have been resampled, currently uploading them.

weikengchen (approved):
Went through all the files for a pass.

PR #303, opened 2021-07-26
Randomized Merkle tree leaves in tests

ljedrz:
This PR was motivated by https://github.com/AleoHQ/snarkVM/issues/280, where a test was using technically invalid test values (as there were non-unique Merkle tree leaves), which initially led to some misleading results. While that specific test is going away when https://github.com/AleoHQ/snarkVM/pull/298 gets merged, there are still some tests that are using dummy values as inputs; with this PR, randomized values for Merkle tree leaves are used instead, improving the quality of the affected tests.

note: this PR was done against the upcoming `testnet2` branch, in order to avoid any conflicts with it.

### DIFF
--- a/algorithms/src/merkle_tree/tests.rs
+++ b/algorithms/src/merkle_tree/tests.rs
@@ -22,7 +22,24 @@ use crate::{
 };
 use snarkvm_utilities::{to_bytes_le, ToBytes};
 
+use rand::{thread_rng, Rng};
+
 use std::sync::Arc;
+
+/// Generates the specified number of random Merkle tree leaves.
+macro_rules! generate_random_leaves {
+    ($num_leaves:expr, $leaf_size:expr) => {{
+        let mut rng = thread_rng();
+
+        let mut vec = Vec::with_capacity($num_leaves);
+        for _ in 0..$num_leaves {
+            let mut id = [0u8; $leaf_size];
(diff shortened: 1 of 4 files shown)


ljedrz:
While working on this PR I noticed that there are many tests that are using a specific, seeded `Rng` (this PR changes one instance of `XorShiftRng::seed_from_u64(1234567)` to plain `thread_rng`); doing so results in the affected tests having the same input values with every run, which reduces the probability of detecting potential "problematic" values; I'll be happy to alter those if desirable, at least in tests which aren't checking for a very specific outcome.

howardwu:
I think this is a desirable change! For now, I would modestly switch from SeedableRng to a thread_rng. For things that are high-level abstractions (like PoSW), feel free to swap out for `thread_rng` everywhere.

ljedrz:
Rebased and added more `ThreadRng` in `posw` as suggested.

This PR can wait until https://github.com/AleoHQ/snarkVM/pull/298 goes in.

howardwu:
Once this PR is rebased, it looks ready to go

ljedrz:
Rebased.

howardwu (approved):

howardwu:
LGTM, thanks for adding the tests! They will improve our reliability :)

PR #311, opened 2021-07-27
Replace Group Encryption with ECIES Poseidon

### LINKED ISSUE #304: [Proposal] Plan to swap GroupEncryption with ECIES Poseidon in DPC inner circuit
## 💥 Proposal

This proposal replaces the group encryption in DPC with ECIES Poseidon. 

I did some analysis on pen and paper. Here is the execution plan. 

- Change the encryption to ECIES, so encryption results are now field elements.
- Drop the selectors.
- Since the ciphertext is now in field elements, it is probably best to replace the `EncryptedRecordCRH` from Bowe-Hopwood Pedersen to Poseidon as well. 

### Encryption side:

`candidate_encrypted_record_hash` ~<= `encrypted_record_hash_input` <= `candidate_encrypted_record_bytes`~ <= `candidate_encrypted_record_gadget` <= `encryption_plaintext_gadget` ~<= `record_group_encoding_gadgets`~ <= `record_field_elements_gadgets`

### Consistency check side: 

`record_field_elements_gadgets` 
- first 5 elements unchanged, 
- the remaining elements come from `payload_elements` ~<=  `payload_field_bits`~ <= `payload_bits`

The PR will be somewhat independent from #301 and #300 since they do not interfere the inner circuit, but better do it after since the DPC trait has changed.

weikengchen:
## Motivation

This PR replaces the group encryption in DPC with field encryption (ECIES Poseidon here).

It also fixes an issue that `ConditionalEqGadget` did not check the length and that nonnative did not truncate the bits properly.

## Test Plan

The code should pass the tests after a resampling of parameters.

## Related PRs

This closes #304


dpc/src/circuits/inner_circuit_gadget.rs
```diff
@@ -721,386 +690,134 @@ where
...
+                    record_field_elements_gadgets.len()
+                ];
+
+            let encryption_blinding_exponents_gadget = <C::AccountEncryptionGadget as EncryptionGadget<
+                C::AccountEncryptionScheme,
+                C::InnerScalarField,
+            >>::EncryptionWitnessGadget::alloc(
                 &mut encryption_cs.ns(|| format!("output record {} encryption_blinding_exponents", j)),
                 || Ok(encryption_blinding_exponents),
             )?;
```
howardwu, line 797: Can you clarify for me what is happening here with the blinding exponents?

I understand that this scheme doesn't require the blinding exponents (as before), and has an encryption witness that is added and subtracted to get the plaintext. Is this technically an allocation of those witnesses?

dpc/src/record/encoded.rs
```diff
@@ -16,45 +16,90 @@
...
-/// Decode a group into the byte representation of a base field element
-pub fn decode_from_group<P: MontgomeryParameters + TwistedEdwardsParameters, G: ProjectiveCurve>(
-    affine: <G as ProjectiveCurve>::Affine,
-    fq_high: bool,
-) -> Result<Vec<u8>, DPCError> {
-    let output = Elligator2::<P, G>::decode(&affine, fq_high)?;
-    Ok(to_bytes_le![output]?)
+/// Decode a group into the byte representation of a base field element.
+pub fn decode_from_field<F: PrimeField>(field_elements: &[F]) -> Result<Vec<u8>, DPCError> {
+    if field_elements.len() == 0 {
```
ljedrz, line 51:
```suggestion
    if field_elements.is_empty() {
```
otherwise `clippy` might complain about it

dpc/src/record/encoded.rs
```diff
@@ -16,45 +16,90 @@
...
+        )));
+    }
+    // There is at least one field element due to the additional true bit.
+
+    let capacity = <F::Parameters as FieldParameters>::CAPACITY as usize;
+
+    let mut bits = Vec::<bool>::with_capacity(field_elements.len() * capacity);
+    for elem in field_elements.iter() {
+        let elem_bits = elem.to_repr().to_bits_le();
+        bits.extend_from_slice(&elem_bits[0..capacity]); // only keep `capacity` bits, discarding the highest bit.
```
ljedrz, line 63:
```suggestion
        bits.extend_from_slice(&elem_bits[..capacity]); // only keep `capacity` bits, discarding the highest bit.
```

dpc/src/circuits/inner_circuit_gadget.rs
weikengchen: Yes, this is technically an allocation of dummy witnesses (for ECIES Poseidon, the encryption witness is empty).

The reason is somehow this: if one is given the randomness, the encryption key, and the plaintext, the person should be able to deduce the witnesses from it. (Therefore, technically the "witnesses" can be internalized by the gadgets and do not need a separate interface.)

dpc/src/record/encoded.rs
```diff
@@ -16,45 +16,90 @@
...
-    fq_high: bool,
-) -> Result<Vec<u8>, DPCError> {
-    let output = Elligator2::<P, G>::decode(&affine, fq_high)?;
-    Ok(to_bytes_le![output]?)
+/// Decode a group into the byte representation of a base field element.
+pub fn decode_from_field<F: PrimeField>(field_elements: &[F]) -> Result<Vec<u8>, DPCError> {
+    if field_elements.len() == 0 {
+        return Err(DPCError::EncodingError(EncodingError::Message(
+            "The encoded record must consist of at least one field element.".to_string(),
+        )));
```
howardwu, line 54:
```suggestion
        return Err(EncodingError::Message(
            "The encoded record must consist of at least one field element.".to_string(),
        ).into());
```

dpc/src/record/encoded.rs
```diff
@@ -16,45 +16,90 @@
...
+        if bits[i] == true {
+            last_bit_pos = Some(i);
+            break;
+        }
+    }
+
+    if last_bit_pos.is_none() {
+        return Err(DPCError::EncodingError(EncodingError::Message(
+            "The encoded record does not end with an expected termination bit.".to_string(),
+        )));
```
howardwu, line 79:
```suggestion
        return Err(EncodingError::Message(
            "The encoded record does not end with an expected termination bit.".to_string(),
        ).into());
```

dpc/src/record/encoded.rs
```diff
@@ -16,45 +16,90 @@
...
+    }
+
+    bits.truncate(last_bit_pos.unwrap());
+    // This also removes the additional termination bit.
+
+    if bits.len() % 8 != 0 {
+        return Err(DPCError::EncodingError(EncodingError::Message(
+            "The number of bits in the encoded record is not a multiply of 8.".to_string(),
+        )));
```
howardwu, line 88:
```suggestion
        return Err(EncodingError::Message(
            "The number of bits in the encoded record is not a multiply of 8.".to_string(),
        ).into());
```

dpc/src/record/encoded.rs
```diff
@@ -16,45 +16,90 @@
...
+
+    // Find the last bit of `true`, which should be within the last `CAPACITY` bits.
+    // If it does not exist, this is a `DPCError::EncodingError`.
+    let mut last_bit_pos = None;
+    for i in ((bits.len() - capacity)..(bits.len())).rev() {
+        if bits[i] == true {
+            last_bit_pos = Some(i);
+            break;
+        }
+    }
```
howardwu, line 74: To make this more readable, I believe you can do the following:
```rs
let mut found_last_bit = false;
while !found_last_bit {
    if let Some(true) = bits.pop() {
        found_last_bit = true;
    }
}
```

This forgoes needing to keep a track of the index, and allows you to remove the `bits.truncate()` call after.

Remember to account for the difference in `.rev()` in the final for loop.

dpc/src/record/encoded.rs
```diff
@@ -16,45 +16,90 @@
...
+    let mut bits = Vec::<bool>::with_capacity(field_elements.len() * capacity);
+    for elem in field_elements.iter() {
+        let elem_bits = elem.to_repr().to_bits_le();
+        bits.extend_from_slice(&elem_bits[0..capacity]); // only keep `capacity` bits, discarding the highest bit.
+    }
+
+    // Find the last bit of `true`, which should be within the last `CAPACITY` bits.
+    // If it does not exist, this is a `DPCError::EncodingError`.
+    let mut last_bit_pos = None;
+    for i in ((bits.len() - capacity)..(bits.len())).rev() {
```
ljedrz, line 69:
```suggestion
    for i in (0..bits.len()).rev().take(capacity) {
```

howardwu: @ljedrz looks like we came up with two approaches.

@weikengchen feel free to take your pick.

dpc/src/record/encoded.rs
ljedrz: This is better than my suggestion, considering `bits` is being truncated later on :+1:.

weikengchen: Lukasz just provided a way to rewrite it. Let me use that.

I limit the amount of bits to be considered to be <= `capacity` to ensure that an overpadding ciphertext is rejected.

weikengchen: (here by overpadding I mean a ciphertext whose last "1" bit is not in the last field element of the encoding, but in a prior one.)

howardwu: If the `capacity` is something you wish to enforce, I suggest a different approach.

`(0..bits.len()).rev().take(capacity)` will not enforce `capacity`. Notice it will take up to `bits.len()` and index directly into that. What I think you want is to `take` from `bits` itself then.

Otherwise @ljedrz approach and my approach functionally will fail under the same conditions.

howardwu: In fact, it may be better to explicitly check the length is `>= capacity` first, and then use an index-less while loop.

dpc/src/testnet1/dpc.rs
```diff
@@ -332,15 +330,6 @@ impl<C: Parameters> DPCScheme<C> for DPC<C> {
 
         // Prepare record encryption components used in the inner SNARK
```
howardwu, line 332:
```suggestion
```

howardwu: The code corresponding to this comment was removed.

dpc/src/testnet2/dpc.rs
```diff
@@ -333,15 +331,6 @@ impl<C: Parameters> DPCScheme<C> for DPC<C> {
 
         // Prepare record encryption components used in the inner SNARK
```
howardwu, line 333:
```suggestion
```
Ditto here, the comment can be removed now.

dpc/src/record/encoded.rs
ljedrz: There's also another longish option that avoids indexing:
```
let last_bit_pos = bits.iter().enumerate().rev().take(capacity).find(|(i_, &b)| b).map(|(i, _b)| i);
```

dpc/src/circuits/inner_circuit_gadget.rs
howardwu: Just to check then, do we actually need to use `record_field_elements_gadgets` in this implementation anymore? Given that we used this as a shortcut in the past, I want to see if this scheme still will benefit from it (since we do pay the constraint cost of checking `expected_plaintext_bits == record_field_elements_gadgets`.

weikengchen: It seems needed, as otherwise the ciphertext could contain a different plaintext.

dpc/src/record/encoded.rs
weikengchen: and yes, `explicitly check the length is >= capacity first` is done, as it requires the plaintext to have at least one element.

howardwu: ^That does work to get the index, however I believe it still does not enforce `bits.len >= capacity`. `take(capacity)` will still "fail gracefully" and `find` won't catch the issue.

ljedrz: It is at least equal, as `field_elements` is checked to be non-empty, and `bits` gets populated with `field_elements.len() * capacity`.

weikengchen: I think there is a guarantee: (let me correct my sentence).

the decode algorithm requires the encoded record to have at least one field element
```
    if field_elements.is_empty() {
        return Err(EncodingError::Message(
            "The encoded record must consist of at least one field element.".to_string(),
        )
        .into());
    }
```

and as a result, the minimal len of `bits` is capacity.

howardwu: Here is a simple example to demonstrate `capacity` isn't enforced.

```rs
fn main() {
    let bits = [true, false, true];
    let capacity = 5;
    
    let last_bit_pos = bits.iter().enumerate().rev().take(capacity).find(|(i_, &b)| b).map(|(i, _b)| i);
    assert_eq!(Some(2), last_bit_pos);
}
```

ljedrz: Indeed, it is not enforced by the iterator itself; that one only avoids indexing, which is a good thing in general, not just for security but often performance as well.

howardwu: Then I think this could work as well (just a revision on my original approach):
```rs
loop {
    if let Some(true) = bits.pop() {
        break;
    }
}
```

ljedrz: That one will go over `capacity`, though; `loop` would need to be `for _ in 0..capacity` instead to limit the maximum number of iterations, if the original behavior is to be preserved.

dpc/src/record/encoded.rs
```diff
@@ -16,45 +16,93 @@
...
+    }
+    // Here we do not use assertion since it can cause Rust panicking.
+
+    let mut bytes = Vec::with_capacity(bits.len() / 8);
+    for chunk in bits.chunks_exact(8) {
+        let mut byte = 0u8;
+        for bit in chunk.iter().rev() {
+            byte <<= 1;
+            byte += *bit as u8;
+        }
```
howardwu, line 101: I believe you can do something like this:
```rs
let bytes = u8::from_le_bytes(&chunk);
```

howardwu: That will also be a const function call then.

dpc/src/record/encoded.rs
howardwu: Spoke with @weikengchen offline, we now have a better picture of the issue.

TLDR - we need the last field element from the given array to be zero. A simple check of that will remove any of this capacity checks and simplify the code significantly.

weikengchen: The logic has been optimized in https://github.com/AleoHQ/snarkVM/pull/311/commits/dbc9dffe2c17e5544c093dec50cbfab780766cba

dpc/src/record/encoded.rs
weikengchen: nay, we need `from_bit_le` but it has not been implemented for `u8`.

dpc/src/record/encoded.rs
howardwu: LGTM

howardwu: Wait, actually, don't use the while loop. I take back my suggestion.

That while loop will infinite loop if given an array of all false's.

weikengchen: And it has been simplified again since we now enforce the last element to be nonzero, everything is simpler.

dpc/src/record/encoded.rs
howardwu: Ah shoot, good catch

dpc/src/record/encoded.rs
howardwu: Okay cool, you caught the issue I was gonna call out, which is the error case after is no longer needed.

howardwu (approved):

howardwu:
LGTM pending CI checks

PR #312, opened 2021-07-27
Update Schnorr signature public key to compressed representation

howardwu:
## Motivation

Updates Schnorr signature public key to compressed representation.

This PR unifies the parameter setup for the `AccountEncryptionScheme` and `AccountSignatureScheme` to use the same input message string, from `ACCOUNT_ENCRYPTION_AND_SIGNATURE_INPUT`.


weikengchen (approved):

howardwu:
The failing case in the integration tests is actually due to the change in key representation for deriving from `PrivateKey` to `ViewKey`. I believe I need to update the BHPCompressedCommitment preimage now to take the `pk_sig` in compressed form in the gadget world as well.

Converting back to draft to investigate.

weikengchen (approved):

howardwu:
I have updated the approach to minimally change how signature public keys work given we may change this application-level signature scheme very soon anyways.

As such, this does not change the inner circuit construction. However, as the encryption and signature setup message have been unified, the constants in the inner circuit have changed, and will require resampling.

howardwu:
Resampling parameters now.

PR #316, opened 2021-07-28
Remove Group Encryption and add a more compact encryption scheme

weikengchen:
## Motivation

This PR replaces the original ECIES Poseidon encryption ciphertext with a more compact representation, which converts `Vec<u8>` to `Vec<u8>`. by doing so, we remove the encoding, as the encryption can be done directly over bytes.

This simple interface comes with a cost of the number of constraints, but it is still in an acceptable range.

This PR simplifies various traits as well as DPC. 

## Test Plan

After a local resampling of the parameters, the test passes the integration tests.

## Related PRs

This is related to a discussion in the internal Slack channel #eng-snarkvm.


dpc/src/testnet1/parameters.rs
```diff
@@ -158,6 +164,7 @@ impl Parameters for Testnet1Parameters {
     dpc_setup!{local_data_crh, LOCAL_DATA_CRH, LocalDataCRH, "AleoLocalDataCRH0"}
     dpc_setup!{program_commitment_scheme, PROGRAM_COMMITMENT_SCHEME, ProgramCommitmentScheme, "AleoProgramIDCommitment0"} // TODO (howardwu): Rename to "AleoProgramCommitmentScheme0".
     dpc_setup!{program_id_crh, PROGRAM_ID_CRH, ProgramIDCRH, "AleoProgramIDCRH0"}
+    dpc_setup!{record_encoding_scheme, RECORD_ENCODING_SCHEME, RecordEncodingScheme, "AleoRecordEncoding0"} // TODO (howardwu): Rename to "AleoRecordEncodingScheme0".
```
weikengchen, line 167: As a side note, we have a few comments `// TODO (howardwu): Rename to "AleoProgramCommitmentScheme0".`

any plan on when such renaming would be done?

howardwu: Perhaps I can address all of the TODOs for input messages in this PR, given we will need to do a resampling of parameters.

weikengchen:
Updated the implementation and is ready for review. Resampling of the parameters as well as genesis would be needed. So better do it together with another PR or so.

Also updated the description in this PR.

howardwu (approved):

howardwu:
Resampling parameters now

PR #317, opened 2021-07-29
New-generation Marlin with SonicKZG10 as PC

weikengchen:
## Motivation

This is a tracking issue for the work-in-progress of a new generation of Marlin. It is expected to take about three days (or more).

The new generation of Marlin consists of two updates:
- a different AHP is introduced, which has fewer polynomials and opens fewer commitments, see Pratyush's work upstream (https://github.com/arkworks-rs/marlin/pull/80)
- a different PC is selected, which is almost better than MarlinKZG10 in any way

Progress as of 7/28:
- ~a new concept, "universal verification parameters", was introduced to SNARK and SNARKGadgets, which allows us to split the verifying key into a part that is universal, and another part that is circuit-specific.~ As of 8/1, this concept is withdrawn, as it is too complicated, and we may just use `alloc_constant` for PC VK and skip it during serialization.
- added all the data structures for SonicKZG10 gadgets.
- initial works in refactoring KZG10 to sample only the needed parameters.

Progress as of 7/29:
- KZG10 now only samples the needed parameters

Progress as of 7/30:
- debug

Progress as of 7/31:
- fixed a bug from Zexe about how G2 prepared gadget being incorrectly converted (https://github.com/arkworks-rs/r1cs-std/pull/77)

Progress as of 8/1:
- added initial works in SonicKZG10 gadget.
- initial tests show that SonicKZG10 works like a charm!
- end-to-end test with old Marlin + SonicKZG10
- replaced old Marlin native with new Marlin native, which passes the native tests

Progress as of 8/2:
- added the new Marlin gadgets
- PR review and cleanup

## Test Plan

Plans to use the same checks that we have today for this new generation.


weikengchen:
This commit is ready for review. 

The improvement on gadgets is not the major part, but still impressive. Outer circuit is now 878967 => 787899 constraints. 

The improvement on native is the major part. Proving will be much faster. Proofs are shorter.

weikengchen:
Parameter resampling is needed.

algorithms/src/fft/polynomial/dense.rs
```diff
@@ -90,17 +90,17 @@ impl<F: Field> DensePolynomial<F> {
     }
 
     /// Evaluates `self` at the given `point` in the field.
-    pub fn evaluate(&self, point: F) -> F {
+    pub fn evaluate(&self, point: &F) -> F {
```
ljedrz, line 93:
```suggestion
    pub fn evaluate(&self, point: F) -> F {
```
AFAIK `Field`s are `Copy`, which means it's preferable to pass them by value

gadgets/src/curves/templates/bls12/g2.rs
```diff
@@ -282,3 +323,74 @@ impl<P: Bls12Parameters> ConditionalEqGadget<<P as Bls12Parameters>::Fp> for G2P
...
+        }
+
+        Ok(Self { ell_coeffs })
+    }
+
+    fn sum<CS: ConstraintSystem<<P as Bls12Parameters>::Fp>>(
+        mut cs: CS,
+        elems: &[Self],
+    ) -> Result<Self, SynthesisError> {
+        assert!(!elems.is_empty());
```
ljedrz, line 348: is this assertion safe to make?

gadgets/src/curves/templates/bls12/g2.rs
```diff
@@ -282,3 +323,74 @@ impl<P: Bls12Parameters> ConditionalEqGadget<<P as Bls12Parameters>::Fp> for G2P
...
+
+        let mut res = elems.first().unwrap().ell_coeffs.clone();
+        for (i, elem) in elems.iter().skip(1).enumerate() {
+            for (j, (l_coeff, r_coeff)) in elem.ell_coeffs.iter().enumerate() {
+                res[j].0 = res[j].0.add(cs.ns(|| format!("sum_{}_{}_entry_0", i, j)), l_coeff)?;
```
ljedrz, line 353: is `res.len()` always greater than `j`?

gadgets/src/curves/templates/bls12/g2.rs
```diff
@@ -282,3 +323,74 @@ impl<P: Bls12Parameters> ConditionalEqGadget<<P as Bls12Parameters>::Fp> for G2P
...
+}
+
+impl<P: Bls12Parameters> CondSelectGadget<<P as Bls12Parameters>::Fp> for G2PreparedGadget<P> {
+    fn conditionally_select<CS: ConstraintSystem<<P as Bls12Parameters>::Fp>>(
+        mut cs: CS,
+        cond: &Boolean,
+        first: &Self,
+        second: &Self,
+    ) -> Result<Self, SynthesisError> {
+        let mut res = Vec::new();
```
ljedrz, line 369:
```suggestion
        let mut res = Vec::with_capacity(first.ell_coeffs.len());
```

.integration/tests/dpc_testnet1.rs
```diff
@@ -54,7 +54,7 @@ fn dpc_testnet1_integration_test() {
             previous_block_hash: BlockHeaderHash([0u8; 32]),
             merkle_root_hash: MerkleRootHash([0u8; 32]),
             pedersen_merkle_root_hash: PedersenMerkleRootHash([0u8; 32]),
-            proof: ProofOfSuccinctWork([0u8; 972]),
+            proof: ProofOfSuccinctWork([0u8; 771]),
```
ljedrz, line 57:
```suggestion
            proof: ProofOfSuccinctWork([0u8; 771]),
```
is this value `PROOF_SIZE` from `ledger`? it would be best if it was imported from there instead

.integration/tests/dpc_testnet2.rs
```diff
@@ -55,7 +55,7 @@ fn dpc_testnet2_integration_test() {
             previous_block_hash: BlockHeaderHash([0u8; 32]),
             merkle_root_hash: MerkleRootHash([0u8; 32]),
             pedersen_merkle_root_hash: PedersenMerkleRootHash([0u8; 32]),
-            proof: ProofOfSuccinctWork([0u8; 972]),
+            proof: ProofOfSuccinctWork([0u8; 771]),
```
ljedrz, line 58: ditto

ledger/src/posw/mod.rs
```diff
@@ -94,7 +103,7 @@ mod tests {
             .mine(&subroots, difficulty_target, &mut rand::thread_rng(), std::u32::MAX)
             .unwrap();
 
-        assert_eq!(proof.len(), 972); // NOTE: Marlin proofs use compressed serialization
+        assert_eq!(proof.len(), 771); // NOTE: Marlin proofs use compressed serialization
```
ljedrz, line 106: the `771` should ideally be replaced with an imported `const`; also, is this assertion safe to make?

marlin/src/ahp/indexer/mod.rs
```diff
@@ -26,3 +26,10 @@ mod constraint_system;
 pub(crate) use constraint_system::*;
 
 mod indexer;
+
+/// Represents a matrix.
+pub(crate) type Matrix<F> = Vec<Vec<(F, usize)>>;
+
+pub(crate) fn num_non_zero(joint_matrix: &Vec<Vec<usize>>) -> usize {
```
ljedrz, line 33:
```suggestion
pub(crate) fn num_non_zero(joint_matrix: &[Vec<usize>]) -> usize {
```

marlin/src/ahp/matrices.rs
```diff
@@ -131,41 +119,57 @@ pub struct MatrixArithmetization<F: PrimeField> {
...
-
-    /// Evaluation of `self.row_col` on an extended domain B (of size > `3K`).
-    pub row_col_evals_on_B: EvaluationsOnDomain<F>,
 }
 
 // TODO for debugging: add test that checks result of arithmetize_matrix(M).
 pub(crate) fn arithmetize_matrix<F: PrimeField>(
-    matrix_name: &str,
-    matrix: &mut Matrix<F>,
+    joint_matrix: &Vec<Vec<usize>>,
```
ljedrz, line 138:
```suggestion
    joint_matrix: &[Vec<usize>],
```

parameters/examples/genesis.rs
```diff
@@ -43,7 +43,7 @@ pub fn generate<C: Parameters>(recipient: &Address<C>, value: u64) -> Result<(Ve
             time: 0,
             difficulty_target: 0xFFFF_FFFF_FFFF_FFFF_u64,
             nonce: 0,
-            proof: ProofOfSuccinctWork([0u8; 972]),
+            proof: ProofOfSuccinctWork([0u8; 771]),
```
ljedrz, line 46: this value could be imported from the defining `const` too

polycommit/src/data_structures.rs
```diff
@@ -134,7 +137,7 @@ impl<F: Field> LabeledPolynomial<F> {
     }
 
     /// Evaluate the polynomial in `self`.
-    pub fn evaluate(&self, point: F) -> F {
+    pub fn evaluate(&self, point: &F) -> F {
```
ljedrz, line 140:
```suggestion
    pub fn evaluate(&self, point: F) -> F {
```
as long as `Field`s are still `Copy`

polycommit/src/kzg10/data_structures.rs
```diff
@@ -59,6 +65,10 @@ impl<E: PairingEngine> PCUniversalParams for UniversalParams<E> {
     fn max_degree(&self) -> usize {
         self.powers_of_g.len() - 1
     }
+
+    fn supported_degree_bounds(&self) -> Vec<usize> {
```
ljedrz, line 69: the call sites that I skimmed looked like iterations, so this should ideally return a `&[usize]` or - better - `impl Iterator<Item = usize>`

polycommit/src/lib.rs
```diff
@@ -672,11 +673,23 @@ pub mod tests {
...
+                    .copied()
+                    .filter(|x| *x >= degree && *x < supported_degree)
+                    .collect::<Vec<usize>>();
+
                 let degree_bound = if let Some(degree_bounds) = &mut degree_bounds {
-                    let range = rand::distributions::Uniform::from(degree..=supported_degree);
-                    let degree_bound = range.sample(rng);
-                    degree_bounds.push(degree_bound);
-                    Some(degree_bound)
+                    if supported_degree_bounds_after_trimmed.len() > 0 && rng.gen() {
```
ljedrz, line 683:
```suggestion
                    if !supported_degree_bounds_after_trimmed.is_empty() > 0 && rng.gen() {
```
otherwise `clippy` might complain

polycommit/src/lib.rs
```diff
@@ -711,7 +724,7 @@ pub mod tests {
                 let point = F::rand(rng);
                 for (i, label) in labels.iter().enumerate() {
                     query_set.insert((label.clone(), (format!("rand_{}", point_id), point)));
-                    let value = polynomials[i].evaluate(point);
+                    let value = polynomials[i].evaluate(&point);
```
ljedrz, line 727: in order to avoid indexing, `polynominals` could be `.zip()`ped with `labels` and iterated together

polycommit/src/lib.rs
```diff
@@ -794,10 +808,18 @@ pub mod tests {
...
+                    .iter()
+                    .copied()
+                    .filter(|x| *x >= degree && *x < supported_degree)
+                    .collect::<Vec<usize>>();
+
                 let degree_bound = if let Some(degree_bounds) = &mut degree_bounds {
-                    if rng.gen() {
-                        let range = rand::distributions::Uniform::from(degree..=supported_degree);
-                        let degree_bound = range.sample(rng);
+                    if supported_degree_bounds_after_trimmed.len() > 0 && rng.gen() {
```
ljedrz, line 818:
```suggestion
                    if !supported_degree_bounds_after_trimmed.is_empty() && rng.gen() {
```

polycommit/src/marlin_pc/gadgets/verifier_key/prepared_verifier_key.rs
```diff
@@ -64,50 +65,80 @@ where
...
+
+            for (i, (_, this_bound, _)) in prepared_degree_bounds_and_shift_powers.iter().enumerate() {
+                if this_bound.get_value().unwrap_or_default().eq(&desired_bound_value) {
+                    pir_vector[i] = true;
+                    break;
                 }
+            }
 
-                None
+            let mut pir_vector_gadgets = Vec::new();
```
ljedrz, line 90:
```suggestion
            let mut pir_vector_gadgets = Vec::with_capacity(pir_vector.len());
```

polycommit/src/sonic_pc/gadgets/sonic_kzg10.rs
```diff
@@ -0,0 +1,705 @@
...
+}
+
+impl<TargetCurve, BaseCurve, PG> SonicKZG10Gadget<TargetCurve, BaseCurve, PG>
+where
+    TargetCurve: PairingEngine<Fq = <BaseCurve as PairingEngine>::Fr>,
+    BaseCurve: PairingEngine,
+    PG: PairingGadget<TargetCurve, <BaseCurve as PairingEngine>::Fr>,
+{
+    #[allow(clippy::type_complexity, clippy::too_many_arguments)]
+    fn prepared_batch_check_evaluations<CS: ConstraintSystem<<BaseCurve as PairingEngine>::Fr>>(
```
ljedrz, line 128: this method is very long and complex - it would be great if it could be split into smaller chunks

polycommit/src/sonic_pc/gadgets/sonic_kzg10.rs
```diff
@@ -0,0 +1,705 @@
...
+        let mut combined_evals = Vec::new();
+
+        let mut combined_degree_bound_comms = Vec::new();
+        let mut combined_degree_bound_shift_powers = Vec::new();
+
+        // Now, for each point, go over all the linear combinations using this point
+        for (i, (_, (point, labels))) in query_to_labels_map.into_iter().enumerate() {
+            // Initialize the vectors to store the combination decisions.
+            let mut comms_to_combine = Vec::<Vec<LCInfoEntry<TargetCurve, BaseCurve, PG>>>::new();
+            let mut values_to_combine = Vec::new();
```
ljedrz, line 201:
```suggestion
            let mut values_to_combine = Vec::with_capacity(labels.len());
```

polycommit/src/sonic_pc/gadgets/sonic_kzg10.rs
```diff
@@ -0,0 +1,705 @@
...
+        let mut combined_queries = Vec::new();
+        let mut combined_evals = Vec::new();
+
+        let mut combined_degree_bound_comms = Vec::new();
+        let mut combined_degree_bound_shift_powers = Vec::new();
+
+        // Now, for each point, go over all the linear combinations using this point
+        for (i, (_, (point, labels))) in query_to_labels_map.into_iter().enumerate() {
+            // Initialize the vectors to store the combination decisions.
+            let mut comms_to_combine = Vec::<Vec<LCInfoEntry<TargetCurve, BaseCurve, PG>>>::new();
```
ljedrz, line 200:
```suggestion
            let mut comms_to_combine = Vec::<Vec<LCInfoEntry<TargetCurve, BaseCurve, PG>>>::with_capacity(labels.len());
```

polycommit/src/sonic_pc/gadgets/verifier_key/prepared_verifier_key.rs
```diff
@@ -0,0 +1,586 @@
...
+        Fn: FnOnce() -> Result<T, SynthesisError>,
+        T: Borrow<PreparedVerifierKey<TargetCurve>>,
+        CS: ConstraintSystem<<BaseCurve as PairingEngine>::Fr>,
+    >(
+        mut cs: CS,
+        value_gen: Fn,
+    ) -> Result<Self, SynthesisError> {
+        let obj = value_gen()?.borrow().clone();
+
+        let mut prepared_g = Vec::<PG::G1Gadget>::new();
```
ljedrz, line 197:
```suggestion
        let mut prepared_g = Vec::<PG::G1Gadget>::with_capacity(obj.prepared_vk.prepared_g.len());
```

polycommit/src/sonic_pc/gadgets/verifier_key/prepared_verifier_key.rs
```diff
@@ -0,0 +1,586 @@
...
+        for (i, g) in obj.prepared_vk.prepared_g.iter().enumerate() {
+            prepared_g.push(<PG::G1Gadget as AllocGadget<
+                <TargetCurve as PairingEngine>::G1Projective,
+                <BaseCurve as PairingEngine>::Fr,
+            >>::alloc_constant(cs.ns(|| format!("g_{}", i)), || {
+                Ok(g.into_projective())
+            })?);
+        }
+
+        let mut prepared_gamma_g = Vec::<PG::G1Gadget>::new();
```
ljedrz, line 207:
```suggestion
        let mut prepared_gamma_g = Vec::<PG::G1Gadget>::with_capacity(obj.prepared_vk.prepared_gamma_g.len());
```

polycommit/src/sonic_pc/gadgets/verifier_key/verifier_key.rs
```diff
@@ -0,0 +1,608 @@
...
+            let desired_bound_value = bound.get_value().unwrap_or_default();
+
+            for (i, (_, this_bound, _)) in degree_bounds_and_neg_powers_of_h.iter().enumerate() {
+                if this_bound.get_value().unwrap_or_default().eq(&desired_bound_value) {
+                    pir_vector[i] = true;
+                    break;
+                }
+            }
+
+            let mut pir_vector_gadgets = Vec::new();
```
ljedrz, line 91:
```suggestion
            let mut pir_vector_gadgets = Vec::with_capacity(pir_vector.len());
```

gadgets/src/curves/templates/bls12/g2.rs
weikengchen: Changed to another implementation that allows an empty `elems`:

```
        let mut res = Self::zero(cs.ns(|| "zero"))?;
        for (i, elem) in elems.iter().enumerate() {
            for (j, (l_coeff, r_coeff)) in elem.ell_coeffs.iter().enumerate() {
                res.ell_coeffs[j].0 = res[j].0.add(cs.ns(|| format!("sum_{}_{}_entry_0", i, j)), l_coeff)?;
                res.ell_coeffs[j].1 = res[j].1.add(cs.ns(|| format!("sum_{}_{}_entry_1", i, j)), r_coeff)?;
            }
        }

        Ok(res)
```

gadgets/src/curves/templates/bls12/g2.rs
weikengchen: Yes, should be safe to assume, as the length is a constant related to the curve parameters.

gadgets/src/traits/select.rs
```diff
@@ -33,6 +33,33 @@ where
...
 }
 
+impl<F: Field, T: CondSelectGadget<F>> CondSelectGadget<F> for Vec<T> {
+    fn conditionally_select<CS: ConstraintSystem<F>>(
+        mut cs: CS,
+        cond: &Boolean,
+        first: &Self,
+        second: &Self,
+    ) -> Result<Self, SynthesisError> {
+        assert_eq!(first.len(), second.len());
```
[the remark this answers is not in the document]
weikengchen: Should be, as all legit `G2PreparedGadget` for the same curve has the same length.

.integration/tests/dpc_testnet1.rs
weikengchen: Implemented the `Default` trait for `ProofOfSuccinctWork`. 
Replaced with `ProofOfSuccinctWork::default()`.

polycommit/src/sonic_pc/gadgets/sonic_kzg10.rs
weikengchen: I may do it in a future PR. The upstream arkworks-rs/poly-commit is doing some similar refactoring aiming to simplify the arguments passing to the function. And we may follow up with those changes later.

weikengchen:
@ljedrz comments have been mostly solved.

weikengchen:
And can confirm that the remaining tests should pass upon parameter resampling.

Pratyush:
The outer circuit is smaller by 100k constraints? That's great. Are there benchmarks on concrete circuits for the faster Marlin proving? I'm interested in how the new AHP affects prover time and memory consumption.

weikengchen:
I can try to get a number by running the inner circuit instead of our benchmark circuit (which is, how to say, too sparse), and see how much is the improvement.

howardwu (commented):
Leaving a quick note that I'm halfway through this PR, and reviewing it piece-wise.

This PR looks great, most notably, the nice things I see are:
- We have unified the various instantiations of Marlin, which was much needed tech debt reduction, many thanks for doing this.
- The changes to arithmetization, balancing matrices, and nonzero are incredibly nice, for developers these are abstract as lower-level details that are either now unnecessary or black boxed.
- A handful of gadget traits have been simplified or unified

weikengchen:
And in case necessary, I can add some explanations on certain parts. Just circle out those that confuses you.

Let me resolve the merge conflict.

weikengchen:
Below is a summary of a discussion with Pratyush. We analyze the improvement. The experiment is done on inner circuit for testnet2, although the inner circuit is indeed proved in Groth16.

**Shorter proof.** The proof for succinct work reduces to 771 bytes from 972 bytes.

**Fewer constraints.** The outer circuit, which verifies the program proof that is in Marlin, reduces to 787k constraints from 878k.

**Faster proving.** We expect the proving time to go down by 20%. This is due to the change of PC scheme to Sonic KZG10.

oldmarlin+marlinkzg10 takes 102s.
oldmarlin+sonickzg10 takes 85s.

newmarlin+marlinkzg10 takes 107s.
newmarlin+sonickzg10 takes 83s.

**Faster verification.** The verification time goes down slightly (so it is not that important actually) because the new Marlin AHP has fewer commitments to play with. The time for Sonic KZG10 is slightly higher, but we believe it can be optimized further (without changing the design), which we describe in https://github.com/arkworks-rs/poly-commit/issues/85.

oldmarlin+marlinkzg10 takes 45ms.
oldmarlin+sonickzg10 takes 49ms.

newmarlin+marlinkzg10 takes 42ms.
newmarlin+sonickzg10 takes 46ms.

**Ability to prove.** For an R1CS instance consisting of matrices A, B, and C, the old Marlin requires the SRS to have length at least `3 * max(|A|, |B|, |C|)` where |M| is the number of non-zero entries in that matrix. The new Marlin requires (|A| + |B| + |C|), which has a hope to be much smaller than the previous one.

For the inner circuit that we experiment on, the new Marlin will require a smaller SRS. To say it differently, if we have a setup of a fixed size, the new Marlin can prove more complicated statements.

oldmarlin is for a max-degree of 3145725
newmarlin is for a max-degree of 2097152


There are other changes.

**SRS is slightly longer.** Now the SRS also contains the negative powers of h in G2, but they are only a few.

**Verifier key is longer, or shorter.** Now the PC verifier key in all Marlin verifier keys are the same, regardless of the circuit, as long as they are in the same setup. If you transmit the PC verifier key all the time, then the key is longer. But you can choose not to transmit the PC verifier key, but only transfer the part that is circuit-dependent, then the verifier key is shorter. 

There is a possibility to reduce the verifier key size further, as three index comms (for vanishing polynomials) are indeed universal to the setup and is not circuit-dependent, and can be moved out of the verifier key. This is slightly messy and would be done in a future PR, with which the program may retrieve the PC verifier key from the `parameters` system and generate the commitments to vanishing polynomials deterministically.

howardwu (approved):

PR #321, opened 2021-07-29
Adds record serial number tree parameters, enforce new ledger block height of 1

howardwu:
## Motivation

This PR adds parameters for a record serial number tree, which after discussion is necessary for light client support. In addition, to enable consensus on the genesis block, we set the initialized ledger's block height to a convention of 1.

- Adds record serial number tree parameters
- Enforces a newly-initialized ledger has block height of 1
- Adds a basic convention for a genesis block header

### DIFF
--- a/.integration/tests/dpc_testnet1.rs
+++ b/.integration/tests/dpc_testnet1.rs
@@ -287,7 +287,7 @@ fn test_testnet1_dpc_execute_constraints() {
             previous_block_hash: BlockHeaderHash([0u8; 32]),
             merkle_root_hash: MerkleRootHash([0u8; 32]),
             time: 0,
-            difficulty_target: 0x07FF_FFFF_FFFF_FFFF_u64,
+            difficulty_target: 0xFFFF_FFFF_FFFF_FFFF_u64,
             nonce: 0,
             pedersen_merkle_root_hash: PedersenMerkleRootHash([0u8; 32]),
             proof: ProofOfSuccinctWork([0u8; 972]),
--- a/.integration/tests/dpc_testnet1.rs
+++ b/.integration/tests/dpc_testnet1.rs
@@ -71,10 +71,10 @@ fn dpc_testnet1_integration_test() {
             previous_block_hash: BlockHeaderHash([0u8; 32]),
             merkle_root_hash: MerkleRootHash([0u8; 32]),
             pedersen_merkle_root_hash: PedersenMerkleRootHash([0u8; 32]),
-            time: 0,
-            difficulty_target: 0x07FF_FFFF_FFFF_FFFF_u64,
-            nonce: 0,
             proof: ProofOfSuccinctWork([0u8; 972]),
+            time: 0,
+            difficulty_target: 0xFFFF_FFFF_FFFF_FFFF_u64,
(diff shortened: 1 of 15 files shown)


.integration/tests/dpc_testnet1.rs
```diff
@@ -287,7 +287,7 @@ fn test_testnet1_dpc_execute_constraints() {
             previous_block_hash: BlockHeaderHash([0u8; 32]),
             merkle_root_hash: MerkleRootHash([0u8; 32]),
             time: 0,
-            difficulty_target: 0x07FF_FFFF_FFFF_FFFF_u64,
+            difficulty_target: 0xFFFF_FFFF_FFFF_FFFF_u64,
```
ljedrz, line 290: is this value something that's tied to the testnet version? if it is, perhaps it should be some module-level `const`?

ledger/src/block/block_header.rs
```diff
@@ -59,6 +59,16 @@ impl BlockHeader {
...
     }
 
+    /// Returns `true` if the block header is uniquely a genesis block header.
+    pub fn is_genesis(&self) -> bool {
+        // Ensure the previous block hash in the genesis block is 0.
+        self.previous_block_hash == BlockHeaderHash([0u8; 32])
+            // Ensure the timestamp in the genesis block is 0.
+            && self.time == 0
+            // Ensure the difficulty target in the genesis block is 0xFFFF_FFFF_FFFF_FFFF_u64.
+            && self.difficulty_target == 0xFFFF_FFFF_FFFF_FFFF_u64
```
ljedrz, line 69: isn't any of these conditions sufficient to determine that something is a genesis block? if that's the case, this should be:
```
self.time == 0 || self.difficulty_target == 0xFFFF_FFFF_FFFF_FFFF_u64 || self.previous_block_hash == BlockHeaderHash([0u8; 32])
```
instead (with this altered order for faster short-circuiting)

ljedrz (approved):
Left some questions; other than this, LGTM :+1:.

.integration/tests/dpc_testnet1.rs
ljedrz: Having gone through the whole PR it seems that it's rather a genesis-specific quality; in that case it could also be some `const` (probably in `ledger`), e.g. `GENESIS_BLOCK_DIFFICULTY` so it's more clear.

ledger/src/block/block_header.rs
howardwu: Thats a great point, will definitely reorder for faster short circuiting.

The previous block hash and time options can definitely be an `||`. The odds of hitting an all 0's block hash would be astonishing.

Now the difficulty target, I'm not too sure about. If the network drops in mining power, technically the min difficulty would be hit again (0xFFFF...).

howardwu: For now, I will remove the difficulty requirement here since it isn't as relevant.

.integration/tests/dpc_testnet1.rs
howardwu: I've removed the convention requirement for now, and will revisit this before testnet2 lands

PR #324, opened 2021-07-29
Implement program selector tree

raychu86:
## Motivation

This PR implements the program selector merkle tree. 

Each record will now store the program selector root, which is the root of a depth 8 tree, where the leaves are the program id's of an authorized program. This will allow for records to be spent/created with different program executions without having to create a monolith circuit with all the permitted programs. 

## Test Plan

The dpc integration tests have been updated to reflect the new syntax and structure for generating a program execution.


weikengchen:
Just to learn a little bit: 
- in Zexe, a record has birth predicate and death predicate. In our case, the "predicate" is the "program id", aka the verifying key.
- now we change it into a tree of program ids, where each of the program id presents a candidate predicate. A record can die in, for example, now in 2^7 ways?

dpc/src/circuits/outer_circuit_gadget.rs
```diff
@@ -317,7 +342,7 @@ pub fn execute_outer_circuit<C: Parameters, CS: ConstraintSystem<C::OuterScalarF
         )?;
 
         let birth_program_vk_field_elements =
-            birth_program_vk.to_constraint_field(cs.ns(|| "birth_death_program_vk_field_elements"))?;
+            birth_program_vk.to_constraint_field(cs.ns(|| "birth_birth_program_vk_field_elements"))?;
```
weikengchen, line 345: Good catch. I feel that maybe our naming of variables in this loop is suboptimal. 

Is it true that in this part of the code, actually all `birth_program_vk_field_elements` is supposed to be something like `birth_or_death_program_vk_field_elements`?  Since the loop is over all the records:

```
for (j, input) in program_proofs
        .iter()
        .skip(C::NUM_INPUT_RECORDS)
        .enumerate()
        .take(C::NUM_OUTPUT_RECORDS)
```

should we change the names to make this clear?

dpc/src/program/execution.rs
```diff
@@ -15,15 +15,16 @@
...
 use crate::Parameters;
-use snarkvm_algorithms::SNARK;
+use snarkvm_algorithms::{merkle_tree::MerklePath, SNARK};
 use snarkvm_fields::{ConstraintFieldError, ToConstraintField};
 
 /// Program verifying key and proof.
 #[derive(Derivative)]
 #[derivative(Clone(bound = "S: SNARK"))]
-pub struct Execution<S: SNARK> {
+pub struct Execution<C: Parameters, S: SNARK> {
```
weikengchen, line 24: do you think it is possible AND reasonable to reduce the type parameters of `Execution` down to only `C: Parameters`? This means that the `Execution` struct will now only support one type of the proof. Currently in our system, it seems that `S: SNARK` will always be `C::ProgramSNARK`.

weikengchen (approved):
The changes look reasonable.

dpc/src/program/execution.rs
howardwu: I agree with this, we should switch `S: SNARK` for `C: Parameters` instead of incurring both. I am happy to make the change and then resample parameters.

howardwu:
> Just to learn a little bit:
> 
> * in Zexe, a record has birth predicate and death predicate. In our case, the "predicate" is the "program id", aka the verifying key.
> * now we change it into a tree of program ids, where each of the program id presents a candidate predicate. A record can die in, for example, now in 2^7 ways?

@weikengchen correct. This is a design choice we discussed at the offsite. Pratyush and Matt agree with this approach to allow us to select the output. This also greatly simplifies the Leo programming paradigm (at the cost of complicating the compiler's logic, though not significant in our opinion)

dpc/src/circuits/outer_circuit_gadget.rs
raychu86: The program proofs are ideally in the order or [death program 1, death program 2, birth program 1, birth program 2]. The for loop you pointed out should skip over the death programs, so the field elements should be for the birth programs only.

dpc/src/program/execution.rs
raychu86: Right. Should just be `C: Parameters`

dpc/src/circuits/outer_circuit_gadget.rs
raychu86: Oops, my fix had a typo in it as well. Should be fixed in my latest commit.

weikengchen: okay, I just realized it skipped all the input records. (and yes, then it is reasonable)

howardwu:
Wait Ray, let's change the branch of this PR. I fixed the additional Merkle tree issue in the `feat/initial-ledger` branch.

ljedrz (approved):
Saw no issues code-wise :+1:.

howardwu:
Closing this PR in favor of PR #326

PR #325, opened 2021-07-30
Revert the compact ECIES Poseidon back to the field-based version

### LINKED ISSUE #323: [Bug] Improve the security of the compact ECIES Poseidon encryption
## 🐛 Bug Report

Pratyush discovered that the ECIES Poseidon encryption based on byte (#316) has extracted too many bytes from a Poseidon field element than what is provably secure to do. 

Sorry that this was not discovered earlier. Therefore I have been working on a fix this morning. There are two solutions, both with a simple interface, but tradeoffs between num. of constraint and transaction size.

### Solution 1:  extract fewer bytes from Poseidon

Implemented in a branch, 
https://github.com/AleoHQ/snarkVM/commit/da7b277546522e879b3b71c4f8d0acb80eba42bb

If for each Poseidon output of a field element of 253 bits, we remove the highest 80 + 1 bits and use only the rest for the mask, it is secure. This increases the amount of Poseidon work.

The tradeoff:
 - constraints 293209 => 296773, 3744 constraints up
 - encrypted record size is unchanged, still 328 bytes

### Solution 2: revert to the field-based encryption (#311)

Implemented in a branch, 
https://github.com/AleoHQ/snarkVM/commit/f053b1e47ee6db53f94dcc7810017bf8a5bcd813

The encryption will still take input `Vec<u8>` and output `Vec<u8>`, so it does not affect the DPC.

The tradeoff:
 - constraints 293209 => 287289, 5920 constraints down
 - encrypted record size grows from 328 bytes to 352 bytes

I leave the decision to @pratyush and @howardwu to pick one. And then I will make a PR on one of them.

weikengchen:
## Motivation

Per discussion of the issue #323, we have reached a decision to use ECIES Poseidon based on fields.

This PR merges the changes from that branch that reflects the changes. The tests have passed. 

## Test Plan

It runs the same consistency checks as before. After resampling parameters locally, it passes the integration tests.

## Related PRs

This closes #323


weikengchen:
Base changed. Let me pull the code again and see if there is anything I need to update.

weikengchen:
I think it is ready for a resampling of parameters. 

No additional change is needed, since `feat/initial-ledger` seems to be very similar to `testnet2`.

howardwu:
Looks good at first pass. Will give more thorough review after resolving merge issues in PR #324.

howardwu (approved):

howardwu:
LGTM, resampling inner and genesis now

PR #328, opened 2021-07-31
Fix the issue that breaks the testnet1

weikengchen:
## Motivation

There were some commits that should not go to the `master` but should go to `testnet2` because they are breaking the testnet1.

This PR reverts those commits from the `master`. The Poseidon sponge will return to a less standard version. And the Marlin verifier will return to be sort-by-value version. Both are good for security. It is just not compatible with the Marlin gadgets and wide use of Poseidon in testnet2.

This PR will go to `master`. Then, a new version will be cut, which will be used for snarkOS. Before this works out, I will firstly check if this PR indeed solves the problem in snarkOS.

## Test Plan

I will wait for the CI and fix here and there. Then, I will compile a version of the snarkOS using this PR-ed branch as the patches, to see if it solves the problem.

## Related PRs

This is related to https://github.com/AleoHQ/snarkOS/issues/1004


weikengchen:
Note that this PR has the following side effects:

- as a result of this fix, snarkVM 0.7.6 will no longer have a functional Marlin gadget or testnet2. Therefore, in this version, all the integration tests for testnet2 has been disabled, as well as the Marlin gadget test.

- this is because testnet2 uses `MarlinVerificationGadget`, which will fail after the reversion of the fix. 

This also means that we cannot directly merge the `testnet2` to the `master` unless we are sure that testnet1 will not go up to the `master` again. Otherwise, we need to prepare for two copies of implementations.

weikengchen:
Running snarkOS with the patched version shows that the issue is fixed---it can successfully process the blocks with PoSW.

howardwu (approved):

weikengchen:
After the merge of this, we can cut a release for 0.7.6, and I will ping Max.

howardwu:
Thank you for making the backport!

PR #330, opened 2021-07-31
Improve `snarkVM` usability, expands on DPC and ledger usability

howardwu:
## Motivation

A modest set of changes to improve `snarkVM` ergonomics, expands on DPC and Ledger usability.

- Ledger with genesis block returns to block height of `0` (same as `master` branch)
- Adds `snarkVM`-level member access (e.g. `snarkvm::dpc::testnet1::*`)
- Unifies the parameter generation programs between networks
- Makes `Parameters::program_srs(rng)` a lazy static load
- Simplifies `snarkvm-integration` library


weikengchen (approved):
The changes look reasonable. I was also thinking about having a `snarkvm` crate that imports the rest. And I am glad that we have it now. It simplifies what `snarkos` needs to import. And if one wants to debug, just apply a patch over the `snarkvm` crate.

howardwu:
> The changes look reasonable. I was also thinking about having a `snarkvm` crate that imports the rest. And I am glad that we have it now. It simplifies what `snarkos` needs to import. And if one wants to debug, just apply a patch over the `snarkvm` crate.

@weikengchen you can see an example of snarkOS updated to use just `snarkvm` here: https://github.com/AleoHQ/snarkOS/pull/978/files

PR #331, opened 2021-08-01
Upgrades DPC, introduces dedicated structs for use in program executions

howardwu:
## Motivation

This PR simplifies much of the DPC execution logic to improve readability and prepare for a transaction builder.

As we use a tree to represent local data state, the record position is now an implicit part of the construction, and programs who need to know position will need to derive it by retracing the Merkle path siblings, or for input records only, recompute the serial number with a private key to verify an input record position. This PR updates the local data leaves to include the leaf index as part of the local data commitment. 

- Renames `DPC::execute_offline_phase()` to `DPC::authorize()`
    - `DPC::authorize()` now returns a `TransactionAuthorization` struct, which embeds the `TransactionKernel` 
- Renames `DPC::execute_online_phase()` to `DPC::execute()`
    - `DPC::execute()` now returns only the `Transaction` struct, as `new_records` are already in the kernel given by the caller
- Adds `LocalDataLeaf` and `leaf_index` to each leaf (requires resampling inner circuit parameters)
- Adds `TransactionAuthorization` struct


howardwu:
^The above commit is the only commit so far to change any circuit logic. Observe that it is a 2 line change the inner circuit successfully reaches the constraint size check error, meaning the inner circuit proof generated and verified successfully.

howardwu:
^The penultimate commit (2f2ae22) above passes the execute constraint test now that the inner circuit constraint count was updated.

weikengchen (approved):
I can see that the renaming makes it more clear.

PR #342, opened 2021-08-03
Expose position in testnet1 record

### LINKED ISSUE #341: [Feature] Expose position getter/constructable parameter in Record
We need to expose position in record to round trip records through snarkOS record serialization during transaction crafting, and it can neither be gotten or set at the moment.

Protryon:
Fixes #341

### DIFF
--- a/dpc/src/testnet1/record/encrypted.rs
+++ b/dpc/src/testnet1/record/encrypted.rs
@@ -233,6 +233,7 @@ impl<C: Testnet1Components> EncryptedRecord<C> {
             commitment,
             commitment_randomness,
             None,
+            None,
         ))
     }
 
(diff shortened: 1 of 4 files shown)


weikengchen:
So in summary, we add the getter for `position` and also add an argument in `from` about `position`.

weikengchen:
(does it mean that we need to cut another release for this to work?)

howardwu (approved):

howardwu:
Wanted to share my thoughts on this: I think this convention is fine for now. I do question (other than for consistency of state) why we need the record position value in snarkOS, given this is an internal helper that shouldn't be necessary.

In any case, I will make an effort to deprecate this and the `is_dummy` (by encoding rules about what a dummy is) so that these side effects are resolved altogether.

PR #344, opened 2021-08-05
Adds enhancements for the DPC program model

howardwu:
## Motivation

Adds enhancements for the DPC program model.

## Changes to Accounts

- Adds `Copy` trait to `Address` struct

## Changes to Records

- Removes `position` from `Record` struct (it is never used)
- Adds `Record::new_noop_input()` and `Record::new_noop_output()`
- Adds `Record::new_input()` and `Record::new_output()`

## Changes to Programs

- Includes #343 
- Reverses program public variables declaration to match allocation order
- Changes program circuits to be referenced from `circuit_id` instead of `circuit_index` (error-prone convention)
- Adds `Parameters::noop_circuit_id()` only for use in `load()` methods
- Adds helper method `fn program_circuit_id(verifying_key: &<Self::ProgramSNARK as SNARK>::VerifyingKey) -> Result<Self::ProgramCircuitID>` into `Parameters`
- Renames `Parameters::program_id_crh()` to `Parameters::program_circuit_id_crh()`
- Renames `Parameters::program_id_tree_parameters()` to `Parameters::program_circuit_tree_parameters()`

## Changes to DPC

- Introduces `Executable` to `DPC::execute()` which leverage trait objects to abstract the program executions
- Adds `DPCError::MissingCircuit`

### DIFF
--- a/dpc/src/program/public_variables.rs
+++ b/dpc/src/program/public_variables.rs
@@ -15,17 +15,7 @@
 // along with the snarkVM library. If not, see <https://www.gnu.org/licenses/>.
 
 use crate::Parameters;
-use snarkvm_algorithms::{merkle_tree::MerklePath, SNARK};
 use snarkvm_fields::{ConstraintFieldError, ToConstraintField};
-
-/// Program index, path, verifying key, and proof.
-#[derive(Derivative)]
-#[derivative(Clone(bound = "C: Parameters"))]
-pub struct Execution<C: Parameters> {
-    pub program_path: MerklePath<C::ProgramCircuitTreeParameters>,
-    pub verifying_key: <C::ProgramSNARK as SNARK>::VerifyingKey,
-    pub proof: <C::ProgramSNARK as SNARK>::Proof,
-}
 
 #[derive(Derivative)]
 #[derivative(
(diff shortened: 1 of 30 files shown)


dpc/src/program/public_variables.rs
```diff
@@ -15,41 +15,30 @@
...
 #[derive(Derivative)]
 #[derivative(
     Clone(bound = "C: Parameters"),
     Debug(bound = "C: Parameters"),
     Default(bound = "C: Parameters")
 )]
-pub struct ProgramPublicVariables<C: Parameters> {
-    pub local_data_root: C::LocalDataRoot,
+pub struct PublicVariables<C: Parameters> {
     pub record_position: u8,
```
weikengchen, line 27: Just to confirm: The PR said that we removed the record position from the `Record` struct, but `position` will still be used in other places to reference to this record right?

parameters/examples/genesis.rs
```diff
@@ -58,71 +58,51 @@ pub fn generate<C: Parameters>(recipient: &Address<C>, value: u64) -> Result<(Ve
...
-        recipient.clone(),
+        recipient,
         false,
         value,
         Payload::default(),
         C::NUM_INPUT_RECORDS as u8,
         joint_serial_numbers.clone(),
         rng,
     )?);
-    new_records.push(Record::new_full(
```
weikengchen, line 91: I agree the input/output records are better than old/new records, as the old/new records may give people a feeling that we are just updating existing records.

weikengchen (approved):
The changes look reasonable.

dpc/src/program/public_variables.rs
howardwu: Correct, there is a discrepancy that we need to resolve (via a standard convention) for the `leaf_index` and `record_position` as both are part of local data. They arose for different reasons, but are ordered in the same manner.

PR #346, opened 2021-08-07
Changes input record sn nonce allocation, adds noop empty payload check

howardwu:
## Motivation

This PR requires resampling parameters.

### Inner circuit changes

- [-2 constraints] Changes the input record's serial number nonce allocation:
```rs
// Before (pseudocode)

let serial_number_nonce = Circuit::alloc(record.serial_number_nonce()); // never used again.
let serial_number_nonce_bytes = Circuit::alloc(serial_number_nonce.to_bytes_le());

// After (pseudocode)

let serial_number_nonce_bytes = Circuit::alloc(record.serial_number_nonce().to_bytes_le());
```
- [+4096 constraints] Adds noop record empty payload checks (in addition to existing zero value checks):
```rs
given_payload.conditional_enforce_equal(cs, empty_payload, given_is_dummy);
```
- [-6 constraints] Changes the zero value alloc to constant.

### General changes

- Adds `Transaction::to_encrypted_record_hashes()`
- Adds `TransactionKernel::to_output_serial_number_nonces()`
- Adds `InnerPublicVariables::new()`
- Adds `Parameters::SerialNumberNonce` convenience type


dpc/src/circuits/inner_circuit.rs
```diff
@@ -148,7 +147,11 @@ pub fn execute_inner_circuit<C: Parameters, CS: ConstraintSystem<C::InnerScalarF
         )
     };
 
-    let zero_value = UInt8::alloc_vec(&mut cs.ns(|| "Declare record zero value"), &to_bytes_le![0u64]?)?;
+    let zero_value = UInt8::alloc_vec(&mut cs.ns(|| "Declare a record zero value"), &(0u64).to_bytes_le()?)?;
```
weikengchen, line 150: Shouldn't it be `constant_vec`?

dpc/src/circuits/inner_circuit.rs
```diff
@@ -148,7 +147,11 @@ pub fn execute_inner_circuit<C: Parameters, CS: ConstraintSystem<C::InnerScalarF
...
+    let empty_payload = UInt8::alloc_vec(
```
weikengchen, line 151: Shouldn't it be `constant_vec`?

weikengchen (approved):

dpc/src/circuits/inner_circuit.rs
howardwu: Good catch

dpc/src/circuits/inner_circuit.rs
howardwu: Good catch

dpc/src/circuits/inner_circuit.rs
howardwu: Fixed

dpc/src/circuits/inner_circuit.rs
howardwu: Fixed

howardwu:
Resampled inner circuit parameters

Pratyush:
Sorry I didn't comment on this earlier, but to reduce the cost of the payload = 0, check maybe we should pack the bytes into field elements, and check that those field elements are zero?

weikengchen:
Agree, it will be cheaper. 

It is unlikely that we change the underlying bit comparison implementation (there is a generic implementation for Vec<_>), but we can do it manually here, by packing them to field elements first.

PR #347, opened 2021-08-07
Updates accounts, records, programs, transactions, and DPC with new functionality

howardwu:
### Accounts
- Improves private key generation performance by 98%: removes unnecessary clones, removes unnecessary recomputation of `pk_sig` on every iteration of deriving a valid `r_pk`
```
account_private_key     time:   [7.7360 ms 8.4321 ms 9.1586 ms]                               
                        change: [-98.960% -98.518% -97.638%] (p = 0.00 < 0.05)
                        Performance has improved.
```
- Adds proper `Default` impl of `PrivateKey` as `Self::new(&mut thread_rng())`
- Adds a bench program for accounts

### Records
- Updates `Payload` struct to support constructing from byte arrays less than `PAYLOAD_SIZE`

### Programs
- Generalizes `Program` struct for common use across programs.

### Transactions
- Adds `network_id` and `value_balance` to preimage of transaction ID (prevents potential merge spending)
    - Transaction ID is now the hash of the transaction kernel
```
Transaction ID = Hash(network ID || serial numbers || commitments || value balance || memo)
```
- Adds `Transaction::from()`
- Removes `TransactionScheme::size()`

### DPC
- Adds `State` struct with builder
- Adds `Parameters::AccountSignature` associated type

## Test Plan

- Adds `test_private_key_to_decryption_key()` to sanity check the MSB bit of the scalar field element is 0
- Adds `test_payload_from()` which creates a random byte array, construct a payload from it, and check its byte array matches
- All existing tests pass

### DIFF
--- a/dpc/src/account/private_key.rs
+++ b/dpc/src/account/private_key.rs
@@ -39,9 +39,11 @@ pub struct PrivateKey<C: Parameters> {
     pub sk_prf: <C::PRF as PRF>::Seed,
     pub r_pk: <C::AccountCommitmentScheme as CommitmentScheme>::Randomness,
     pub r_pk_counter: u16,
-    // This dummy flag is set to true for use in the `inner_snark` setup.
+    // This dummy flag is set to true for use in the `inner_circuit` setup.
     #[derivative(Default(value = "true"))]
-    pub is_dummy: bool,
+    is_dummy: bool,
+    pk_sig: C::AccountSignaturePublicKey,
+    commitment_input: Vec<u8>,
 }
 
 impl<C: Parameters> PrivateKey<C> {
@@ -60,54 +62,53 @@ impl<C: Parameters> PrivateKey<C> {
             let seed: [u8; 32] = rng.gen();
 
             // Returns the private key if it is valid.
-            match Self::from_seed(&seed) {
-                Ok(private_key) => return Ok(private_key),
-                _ => continue,
-            };
+            if let Ok(private_key) = Self::from_seed(&seed) {
+                return Ok(private_key);
+            }
         }
     }
 
     /// Derives the account private key from a given seed and verifies it is well-formed.
     pub fn from_seed(seed: &[u8; 32]) -> Result<Self, AccountError> {
         // Generate the SIG key pair.
-        let sk_sig_bytes = Blake2s::evaluate(&seed, &Self::INPUT_SK_SIG)?;
-        let sk_sig = <C::AccountSignatureScheme as SignatureScheme>::PrivateKey::read_le(&sk_sig_bytes[..])?;
+        let sk_sig: <C::AccountSignatureScheme as SignatureScheme>::PrivateKey =
+            FromBytes::read_le(Blake2s::evaluate(&seed, &Self::INPUT_SK_SIG)?.as_ref())?;
+        let pk_sig = C::account_signature_scheme().generate_public_key(&sk_sig)?;
 
         // Generate the PRF secret key.
-        let sk_prf_bytes = Blake2s::evaluate(&seed, &Self::INPUT_SK_PRF)?;
-        let sk_prf = <C::PRF as PRF>::Seed::read_le(&sk_prf_bytes[..])?;
+        let sk_prf: <C::PRF as PRF>::Seed =
+            FromBytes::read_le(Blake2s::evaluate(&seed, &Self::INPUT_SK_PRF)?.as_ref())?;
 
-        let mut r_pk_counter = Self::INITIAL_R_PK_COUNTER;
-        loop {
-            if r_pk_counter == u16::MAX {
-                return Err(AccountError::InvalidPrivateKeySeed);
-            }
+        // Construct the commitment input for the account address.
+        let commitment_input = to_bytes_le![pk_sig, sk_prf]?;
 
-            // Derive the private key by iterating on the r_pk counter
-            let private_key = match Self::derive_r_pk(seed, r_pk_counter) {
-                Ok(r_pk) => Self {
-                    seed: *seed,
-                    sk_sig: sk_sig.clone(),
-                    sk_prf: sk_prf.clone(),
-                    r_pk,
-                    r_pk_counter,
-                    is_dummy: false,
-                },
-                _ => {
-                    r_pk_counter += 1;
-                    continue;
-                }
-            };
+        // Initialize a candidate private key.
+        let mut private_key = Self {
+            seed: *seed,
+            sk_sig,
+            sk_prf,
+            r_pk: Default::default(),
+            r_pk_counter: Self::INITIAL_R_PK_COUNTER,
+            is_dummy: false,
+            pk_sig,
+            commitment_input,
+        };
 
-            // Returns the private key if it is valid.
-            match private_key.is_valid() {
-                true => return Ok(private_key),
-                false => {
-                    r_pk_counter += 1;
-                    continue;
+        // Derive the private key by iterating on the r_pk counter, until a valid r_pk is found.
+        for r_pk_counter in Self::INITIAL_R_PK_COUNTER..u16::MAX {
+            if let Ok(r_pk) = Self::derive_r_pk(seed, r_pk_counter) {
+                // Update the candidate private key with the derived values.
+                private_key.r_pk = r_pk;
+                private_key.r_pk_counter = r_pk_counter;
+
+                // Returns the private key if it is valid.
+                if private_key.is_valid() {
+                    return Ok(private_key);
                 }
             }
         }
+
+        Err(AccountError::InvalidPrivateKeySeed)
     }
 
     /// Returns `true` if the private key is well-formed. Otherwise, returns `false`.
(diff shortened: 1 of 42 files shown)


dpc/src/state/mod.rs
```diff
@@ -0,0 +1,27 @@
...
+pub mod input;
+pub use input::*;
+
+pub mod state;
+pub use state::*;
+
+pub mod output;
+pub use output::*;
+
+// #[cfg(test)]
```
weikengchen, line 26: Just to confirm: so there is a plan to add some tests?

dpc/src/account/private_key.rs
```diff
@@ -39,9 +38,10 @@ pub struct PrivateKey<C: Parameters> {
     pub sk_prf: <C::PRF as PRF>::Seed,
     pub r_pk: <C::AccountCommitmentScheme as CommitmentScheme>::Randomness,
     pub r_pk_counter: u16,
-    // This dummy flag is set to true for use in the `inner_snark` setup.
-    #[derivative(Default(value = "true"))]
-    pub is_dummy: bool,
+    // TODO (howardwu): Move this into a `ProverKey` struct.
+    pk_sig: C::AccountSignaturePublicKey,
+    // TODO (howardwu): Move this into a `ProverKey` struct.
+    commitment_input: Vec<u8>,
```
weikengchen, line 44: Why is it related to `ProverKey`? It is more like signature key or?

dpc/src/state/mod.rs
howardwu: Yes, the plan in my next PR is to add these. The approach will be as follows:

Replace current approach to preparing inputs and outputs with `State` builder, and use the old approach as the sanity check in the unit tests against the new builder

dpc/src/account/private_key.rs
howardwu: This will be the key that an outsourced worker executing `DPC::execute()` can use (from authorization) to compute the transaction proof. (So no sk_sig, has pk_sig, as in the delegable DPC scheme)

dpc/src/account/private_key.rs
```diff
@@ -119,90 +117,92 @@ impl<C: Parameters> PrivateKey<C> {
...
+            if msb_bit {
+                return Err(AccountError::InvalidAccountCommitment);
             }
         }
 
+        // This operation implicitly enforces that the unused MSB bits
+        // for the scalar field representation are correctly set to 0.
+        let decryption_key: <C::AccountEncryptionScheme as EncryptionScheme>::PrivateKey =
+            FromBytes::read_le(&commitment_bytes[..])?;
+
```
weikengchen, line 152: We may want to qualify this "implicit check" more. It is weaker than the MSB check:
- In MSB check, we ensure that it is within the capacity of the TE scalar field.
- In this implicit check (due to ToBytes), it ensures that the number is smaller than the TE curve's scalar field prime number, which may still be larger than the capacity.

howardwu: Do you mean, add a comment that states this explicitly, or should there be something stricter being done?

I will add the comment right now.

dpc/src/account/tests.rs
```diff
@@ -17,16 +17,17 @@
...
 
-    const ALEO_TESTNET1_PRIVATE_KEY: &str = "APrivateKey1tn8cnHPNtcZ9pH89YBMmpPS3fP5kxooguzpbRz3pLWoSzhg";
-    const ALEO_TESTNET1_VIEW_KEY: &str = "AViewKey1m9cmnBtfWziAmT1SBC63a96fo18hLddrjweMxhcqhNo5";
-    const ALEO_TESTNET1_ADDRESS: &str = "aleo1h47qwdqqv25gwp0fkxgnqvm7ykrz0ud2vaw2cj4ac68w8wq5vqqqv58jvr";
+    const ALEO_TESTNET1_PRIVATE_KEY: &str = "APrivateKey1tyBgFoCXAq8RfZT3W2mzVV9XzJb2hVFL2LrHLToEC37tXrz";
+    const ALEO_TESTNET1_VIEW_KEY: &str = "AViewKey1gFRs4gG66wFW1FypYRfwy6pDqix6rtquQ8uJ15RgHCC8";
+    const ALEO_TESTNET1_ADDRESS: &str = "aleo1shhq355tyaptcej65tkrweej5wth7wqg5hcqg3hf8as5s9rtrypqs8vy3u";
 
-    const ITERATIONS: usize = 25;
+    const ITERATIONS: usize = 1000;
```
weikengchen, line 30: Note that several other tests in this file `tests.rs` also run the same test for `ITERATIONS` times. Would this number be too large? If it does not cause a performance issue (which seems to be the case), maybe we can keep 1000 here, for safety.

weikengchen: Since there is indeed a probability procedure in the code.

howardwu: I tried running it locally on 1000, and it appears to be "sufficiently performant", takes about 10-15 seconds

dpc/src/account/private_key.rs
howardwu: I went ahead and added this check into the function as well:
```rs
        // Enforce the MSB of the scalar field element is 0 by convention.
        debug_assert_eq!(Some(&false), decryption_key.to_bits_be().iter().next());
```

For now, I will leave it as a `debug_assert_eq` so it does not affect production performance. However, if we feel the need to elevate it to a release assertion, feel free to make the change.

dpc/src/account/tests.rs
```diff
@@ -55,6 +56,17 @@ mod testnet1 {
...
     }
 
+    #[test]
+    fn test_private_key_to_decryption_key() {
+        // Attempt to sample for a new account ITERATIONS times.
+        for _ in 0..ITERATIONS {
+            let private_key = PrivateKey::<Testnet1Parameters>::new(&mut thread_rng());
+            let decryption_key = private_key.to_decryption_key().unwrap();
+            // Enforce the MSB of the scalar field element is 0 by convention.
+            assert_eq!(Some(&false), decryption_key.to_bits_be().iter().next());
```
weikengchen, line 66: HOLD. 

I found a potential bug in my implementation of `to_bits_be`. The truncation is performed starting from the MSB direction, but it should be from the LSB direction.

https://github.com/AleoHQ/snarkVM/blob/testnet2/fields/src/fp_256.rs#L408

Will post a fix, likely to this branch.

howardwu: Great catch, will hold

weikengchen: And then, it is also curious why this test did not fail.

howardwu: Perhaps the implementation is correct because `BigInteger` does the following:
```rs
            /// Returns `self` as a boolean array in big-endian order, without leading zeros.
            fn to_bits_be(&self) -> Vec<bool> {
                BitIteratorBE::new(self).collect::<Vec<_>>()
            }
```

weikengchen: That makes sense. Although it is more concerning, now it is not fixed length, but we have used this in a few places where it is needed.

Let me use `resize` to make sure it is the right length.

howardwu: Oh wait, I think that implementation is wrong, it should be:
```rs
            /// Returns `self` as a boolean array in big-endian order, without leading zeros.
            fn to_bits_be(&self) -> Vec<bool> {
                BitIteratorBE::new_without_leading_zeros(self).collect::<Vec<_>>()
            }
```

weikengchen: ok, it seems resizing is not needed, and we may want to change that comment in `BigInteger`, as the leading zeros are still there.

weikengchen: There are two commits related to this that you will need to review:

https://github.com/AleoHQ/snarkVM/pull/347/commits/671982fb5d339cb18da0a1d43e88bed6f63a3770
https://github.com/AleoHQ/snarkVM/pull/347/commits/eb8f36f29be2882f29ee728afa23ced57867c132

howardwu:
Reviewer's review update: 671982f and eb8f36f look correct to me

dpc/src/state/state.rs
```diff
@@ -0,0 +1,122 @@
...
+            let position = (C::NUM_INPUT_RECORDS + outputs.len()) as u8;
+            // TODO (howardwu): Decide whether to "push" or "push_front" for program flow.
+            outputs.push(Output::new_noop(noop.clone(), position, rng)?);
+        }
+
+        Ok(State { inputs, outputs })
+    }
+
+    fn update_joint_serial_numbers(&mut self) -> Result<()> {
+        // Compute the joint serial numbers.
```
weikengchen, line 114: I feel maybe we can add an `assert_eq` that `self.inputs` is exactly `C::NUM_INPUT_RECORDS`. This is to ensure that the `joint_serial_numbers` always include the necessary padding of noop inputs.

weikengchen: <- Needs confirmation: is that the expected behavior?

weikengchen (approved):

dpc/src/state/state.rs
howardwu: Yes, that is the intended behavior.

I will add a check there.

howardwu: I'll also mention based on how the struct is designed, there shouldn't be a case whereby more than `C::NUM_INPUT_RECORDS` will be admitted into the struct

howardwu: As such, I will make the check a `debug_assert_eq!` if that is alright with you

howardwu: Done. Will add tests in the next PR to ensure behavior of `State` struct matches basic expectations

weikengchen:
Sounds good. `debug_assert` is sufficient.

Pratyush:
> Adds proper `Default` impl of `PrivateKey` as `Self::new(&mut thread_rng())`

I don't think it's safe to use `thread_rng` there; we should probably use a cryptographic rng

howardwu:
`thread_rng` is a `CryptoRng`, and I believe the `PrivateKey::new` method enforces the trait

PR #348, opened 2021-08-08
Replace the CRH used in testnet with Bowe-Hopwood Pedersen

weikengchen:
## Motivation

After a discussion with Pratyush, we feel it is better to keep Bowe-Hopwood Pedersen as CRH and leave Poseidon for those beyond CRH, such as ECIES encryption and Marlin sponge.

This PR makes the necessary changes for this, which basically changes the Program ID CRH back to BHP.

The original plan is to avoid introducing a new bit/byte serialization. It turns out that we cannot avoid---just because many of our implementations, such as BHP CRH, is defined over generic trait, and we cannot use the existing API to access the sufficient information. Therefore, we introduce a new trait `ToMinimalBitRepresentation` and its corresponding gadgets `ToMinimalBitRepresentationGadget`, which outputs a minimal bit/boolean array that uniquely describes a struct.

More details are described in the file changes.

## Overhead

Testnet1

InnerCircuit  287247 => 292173
OuterCircuit 422723 => 498968

Testnet2

InnerCircuit 287247 => 292173
OuterCircuit 787899 => 861124

## Test Plan

After a resampling of both the testnet1 and testnet2 parameters, it passes all the tests.


.integration/tests/dpc_testnet1.rs
```diff
@@ -419,7 +418,7 @@ fn test_testnet1_dpc_execute_constraints() {
 
     // NOTE: Do not change this to `Testnet1Parameters::inner_circuit_id()` as that will load the *saved* inner circuit VK.
     let inner_circuit_id = <Testnet1Parameters as Parameters>::inner_circuit_id_crh()
-        .hash_field_elements(&inner_snark_vk.to_field_elements().unwrap())
+        .hash_bits(&inner_snark_vk.to_minimal_bit_representation())
```
weikengchen, line 421: We now use `hash_bits` to hash a bit array, and use `to_minimal_bit_representation` to convert a struct to its minimal array representation.

algorithms/src/crh/bhp.rs
```diff
@@ -86,27 +86,25 @@ impl<G: ProjectiveCurve, const NUM_WINDOWS: usize, const WINDOW_SIZE: usize> CRH
         }
     }
 
-    fn hash(&self, input: &[u8]) -> Result<Self::Output, CRHError> {
+    fn hash_bits(&self, input: &[bool]) -> Result<Self::Output, CRHError> {
```
weikengchen, line 89: Now all the CRH implements the `hash_bits`, and the original `hash` will be mapped to `hash_bits`, by the trait definition.

algorithms/src/crh/mod.rs
```diff
@@ -26,9 +26,6 @@ pub use pedersen::*;
 pub mod pedersen_compressed;
 pub use pedersen_compressed::*;
 
-pub mod poseidon;
-pub use poseidon::*;
```
weikengchen, line 30: We drop the Poseidon CRH. It seems that the current implementation did not enforce `INPUT_SIZE_BITS` and simply truncates the part after it in the gadgets, so there is a security issue. Given that we are not using Poseidon CRH for now, I prefer to drop this implementation.

algorithms/src/traits/crh.rs
```diff
@@ -31,7 +31,22 @@ pub trait CRH: Clone + ToBytes + FromBytes + From<<Self as CRH>::Parameters> {
...
+        let mut bits = Vec::with_capacity(input.len() * 8);
+        for byte in input.iter() {
+            bits.push(byte & 1 != 0);
+            bits.push(byte & 2 != 0);
+            bits.push(byte & 4 != 0);
+            bits.push(byte & 8 != 0);
+            bits.push(byte & 16 != 0);
+            bits.push(byte & 32 != 0);
+            bits.push(byte & 64 != 0);
+            bits.push(byte & 128 != 0);
```
weikengchen, line 44: If anyone has a better way to do so, let me know.

curves/src/templates/short_weierstrass_jacobian/affine.rs
```diff
@@ -180,6 +188,16 @@ impl<P: Parameters> AffineCurve for Affine<P> {
     }
 }
 
+impl<P: Parameters> ToMinimalBitRepresentation for Affine<P> {
+    fn to_minimal_bit_representation(&self) -> Vec<bool> {
+        let mut res_bits = self.x.to_bits_le();
+        res_bits.push(self.y.to_bits_le().first().unwrap().clone());
+        res_bits.push(self.infinity);
```
weikengchen, line 195: For SW, we get the (1) full x, (2) the LSB of y (i.e., odd or even), and (3) the infinity bit.

We will see later that in TE, we only need the full x.

dpc/src/parameters/testnet1.rs
```diff
@@ -149,9 +148,13 @@ impl Parameters for Testnet1Parameters {
...
-    type ProgramCircuitIDCRH = PoseidonCryptoHash<Self::OuterScalarField, 4, false>;
-    type ProgramCircuitIDCRHGadget = PoseidonCryptoHashGadget<Self::OuterScalarField, 4, false>;
+    type ProgramCircuitIDCRH = BHPCompressedCRH<EdwardsBW6, 237, 16>;
+    type ProgramCircuitIDCRHGadget = BHPCompressedCRHGadget<EdwardsBW6, Self::OuterScalarField, EdwardsBW6Gadget, 237, 16>;
     type ProgramCircuitID = <Self::ProgramCircuitIDCRH as CRH>::Output;
+
+    type ProgramCircuitIDTreeCRH = BHPCompressedCRH<EdwardsBW6, 8, 48>;
+    type ProgramCircuitIDTreeCRHGadget = BHPCompressedCRHGadget<EdwardsBW6, Self::OuterScalarField, EdwardsBW6Gadget, 8, 48>;
+    type ProgramCircuitIDTreeDigest = <Self::ProgramCircuitIDTreeCRH as CRH>::Output;
     type ProgramCircuitTreeParameters = ProgramIDMerkleTreeParameters;
```
weikengchen, line 158: We separate the `ProgramCircuitIDCRH` and  `ProgramCircuitIDTreeCRH` because the former is too large, and due to a reason about the current BHP implementation, it has an additional cost. So, the tree would invoke BHP of a smaller window. This is the same as our treatment for record commitment trees.

dpc/src/parameters/testnet1.rs
```diff
@@ -127,12 +126,12 @@ impl Parameters for Testnet1Parameters {
...
-    type EncryptedRecordCRH = PoseidonCryptoHash<Self::InnerScalarField, 4, false>;
-    type EncryptedRecordCRHGadget = PoseidonCryptoHashGadget<Self::InnerScalarField, 4, false>;
+    type EncryptedRecordCRH = BHPCompressedCRH<EdwardsBls12, 72, 32>;
+    type EncryptedRecordCRHGadget = BHPCompressedCRHGadget<EdwardsBls12, Self::InnerScalarField, EdwardsBls12Gadget, 72, 32>;
     type EncryptedRecordDigest = <Self::EncryptedRecordCRH as CRH>::Output;
 
-    type InnerCircuitIDCRH = PoseidonCryptoHash<Self::OuterScalarField, 4, false>;
-    type InnerCircuitIDCRHGadget = PoseidonCryptoHashGadget<Self::OuterScalarField, 4, false>;
+    type InnerCircuitIDCRH = BHPCompressedCRH<EdwardsBW6, 296, 32>;
+    type InnerCircuitIDCRHGadget = BHPCompressedCRHGadget<EdwardsBW6, Self::OuterScalarField, EdwardsBW6Gadget, 296, 32>;
```
weikengchen, line 134: The InnerCircuit actually has a pretty long verifying key, because the input is long. We currently has no idea how to handle this.

gadgets/src/algorithms/crh/poseidon.rs
```diff
@@ -1,70 +0,0 @@
...
-
-    fn check_evaluation_gadget<CS: ConstraintSystem<F>>(
-        &self,
-        mut cs: CS,
-        input: Vec<UInt8>,
-    ) -> Result<Self::OutputGadget, SynthesisError> {
-        let params = F::get_default_poseidon_parameters(RATE, OPTIMIZED_FOR_WEIGHTS).unwrap();
-
-        let mut padded_input_bytes = input;
-        padded_input_bytes.resize(
```
weikengchen, line 46: This `resize` is my concern, as it may truncate the input.

gadgets/src/algorithms/merkle_tree/tests.rs
```diff
@@ -417,54 +417,3 @@ mod merkle_tree_bowe_hopwood_pedersen_compressed_crh_on_projective {
         update_merkle_tree::<EdwardsMerkleParameters, Fr, HG>(&leaves);
     }
 }
-
-mod merkle_tree_poseidon {
```
weikengchen, line 421: Drop the Poseidon CRH test since Poseidon CRH is dropped.

gadgets/src/fields/fp12.rs
```diff
@@ -672,15 +672,15 @@ where
     <P::Fp6Params as Fp6Parameters>::Fp2Params: Fp2Parameters<Fp = F>,
 {
     fn to_bits_be<CS: ConstraintSystem<F>>(&self, mut cs: CS) -> Result<Vec<Boolean>, SynthesisError> {
-        let mut c0 = self.c0.to_bits_be(&mut cs)?;
-        let mut c1 = self.c1.to_bits_be(cs)?;
+        let mut c0 = self.c0.to_bits_be(cs.ns(|| "c0"))?;
+        let mut c1 = self.c1.to_bits_be(cs.ns(|| "c1"))?;
```
weikengchen, line 676: This fixes an existing issue about `to_bits_le/be` for Fp2/Fp6/Fp12.

Before this change, it would cause the constraint system to complain that the path has been used.

parameters/src/testnet1/mod.rs
```diff
@@ -68,7 +68,7 @@ impl_params_remote!(
     "https://s3-us-west-1.amazonaws.com/aleo.parameters",
     "./",
     "outer_snark_pk",
-    447760217
+    486323933
```
weikengchen, line 71: InnerSNARK and OuterSnark PK are larger.

polycommit/src/marlin_pc/data_structures.rs
```diff
@@ -234,6 +241,19 @@ pub struct Commitment<E: PairingEngine> {
 }
 impl_bytes!(Commitment);
 
+impl<E: PairingEngine> ToMinimalBitRepresentation for Commitment<E> {
```
weikengchen, line 244: This is partly why this trait is needed, as we need to convert the commitment, but Marlin does not have access to the specific struct because Marlin is parameterized by a generic polynomial commitment.

utilities/src/bits.rs
```diff
@@ -52,3 +52,18 @@ pub trait FromBits: Sized {
     /// Reads `Self` from a boolean array in big-endian order.
     fn from_bits_be(bits: &[bool]) -> Self;
 }
+
+pub trait ToMinimalBitRepresentation: Sized {
```
weikengchen, line 56: This is the new trait that is being added.

dpc/src/parameters/testnet1.rs
Pratyush: If we compute (part of) the public input MSM outside, then we don't need to include those group elements in the hash

algorithms/src/crh/bhp.rs
```diff
@@ -86,27 +86,25 @@ impl<G: ProjectiveCurve, const NUM_WINDOWS: usize, const WINDOW_SIZE: usize> CRH
...
         }
         assert!(WINDOW_SIZE <= MAX_WINDOW_SIZE);
         assert!(NUM_WINDOWS <= MAX_NUM_WINDOWS);
 
         // overzealous but stack allocation
-        let mut buffer = [0u8; MAX_WINDOW_SIZE * MAX_NUM_WINDOWS / 8 + BOWE_HOPWOOD_CHUNK_SIZE + 1];
-        buffer[..input.len()].copy_from_slice(input);
-        let buf_slice = from_bytes_le_to_bits_le(&buffer[..]).collect::<Vec<_>>();
+        let mut buf_slice = input.to_vec();
+        buf_slice.resize(WINDOW_SIZE * NUM_WINDOWS, false);
```
Pratyush, line 100: Why don't we preserve the existing logic of avoiding allocations?

algorithms/src/traits/crh.rs
Pratyush: You can get the bit decomposition using `inputs.iter().flat_map(|byte| 0..8.map(|i| ((byte >> i) & 1) == 1)).collect::<Vec<_>>().`

algorithms/src/crh/bhp.rs
weikengchen: You mean allocate sufficient bits at the beginning (what the `buffer` does) and avoid possible reallocation?

curves/src/templates/short_weierstrass_jacobian/affine.rs
Pratyush: Does the bit decomposition still pad with extra bits?

marlin/src/marlin/circuit_verifying_key.rs
```diff
@@ -53,6 +60,26 @@ impl<F: PrimeField, CF: PrimeField, PC: PolynomialCommitment<F, CF>> ToBytes for
...
+    fn to_minimal_bit_representation(&self) -> Vec<bool> {
+        let domain_h = EvaluationDomain::<F>::new(self.circuit_info.num_constraints)
+            .ok_or(SynthesisError::PolynomialDegreeTooLarge)
+            .unwrap();
+        let domain_k = EvaluationDomain::<F>::new(self.circuit_info.num_non_zero)
+            .ok_or(SynthesisError::PolynomialDegreeTooLarge)
+            .unwrap();
+
+        let domain_h_size_bits = CF::from(domain_h.size() as u128).to_bits_le();
+        let domain_k_size_bits = CF::from(domain_k.size() as u128).to_bits_le();
```
Pratyush, line 75: We don't need to serialize things as u128; u64 should be more than enough.

algorithms/src/crh/bhp.rs
Pratyush: in the old code, the buffer allocated sufficient space as a *stack array* (i.e., no heap allocations), and used that everywhere. The new code instead reallocates twice.

algorithms/src/crh/mod.rs
howardwu: Actually we need this upstream, currently using it for a Merkle tree construction on the application level

howardwu (changes requested):
Overall looks good, two requests:

1. Can we bring back the poseidon CRH implementation? We need this upstream (and the associated MT safety tests)
2. What do you think about renaming `ToMinimalBitRepresentation` to `ToMinimalBits` for now? (Optional, was considering verbosity of `object.to_minimal_bit_representation()`)

howardwu:
Happy to resample parameters after open requests are addressed.

curves/src/templates/short_weierstrass_jacobian/affine.rs
Pratyush: bit decomposition of the field elements*

That, is is `bls12_377::Fq.to_bites` 377 bits or 384 bits?

gadgets/src/algorithms/crh/pedersen.rs
```diff
@@ -90,19 +90,25 @@ impl<F: PrimeField, G: ProjectiveCurve, GG: CurveGadget<G, F>, const NUM_WINDOWS
...
-
+        let input_in_bits = pad_input::<NUM_WINDOWS, WINDOW_SIZE>(input);
         GG::multi_scalar_multiplication(cs, &self.crh.bases, input_in_bits.chunks(WINDOW_SIZE))
     }
 }
 
+fn pad_input<const NUM_WINDOWS: usize, const WINDOW_SIZE: usize>(input: Vec<Boolean>) -> Vec<Boolean> {
+    let mut padded_input = input;
+    padded_input.resize(WINDOW_SIZE * NUM_WINDOWS, Boolean::Constant(false));
+    assert_eq!(padded_input.len(), WINDOW_SIZE * NUM_WINDOWS);
```
ljedrz, line 108: this assertion is not needed - `Vec::resize` makes sure this is the case

gadgets/src/traits/misc.rs
```diff
@@ -36,3 +37,21 @@ pub trait SumGadget<F: PrimeField>: Clone {
...
+}
+
+impl<F: PrimeField, T: ToMinimalBitRepresentationGadget<F>> ToMinimalBitRepresentationGadget<F> for Vec<T> {
+    fn to_minimal_bit_representation<CS: ConstraintSystem<F>>(
+        &self,
+        mut cs: CS,
+    ) -> Result<Vec<Boolean>, SynthesisError> {
+        let mut res_booleans = vec![];
+        for (i, elem) in self.iter().enumerate() {
+            res_booleans.extend(elem.to_minimal_bit_representation(cs.ns(|| format!("{}", i)))?);
```
ljedrz, line 52:
```suggestion
            res_booleans.extend(elem.to_minimal_bit_representation(cs.ns(|| i.to_string()))?);
```

polycommit/src/marlin_pc/data_structures.rs
```diff
@@ -234,6 +241,19 @@ pub struct Commitment<E: PairingEngine> {
...
+    fn to_minimal_bit_representation(&self) -> Vec<bool> {
+        let comm_bits = self.comm.to_minimal_bit_representation();
+        let shifted_comm_bits = if let Some(shifted_comm) = &self.shifted_comm {
+            shifted_comm.to_minimal_bit_representation()
+        } else {
+            vec![]
+        };
+
+        [comm_bits, shifted_comm_bits].concat()
```
ljedrz, line 253:
```suggestion
        if let Some(shifted_comm) = &self.shifted_comm {
            [comm_bits, shifted_comm.to_minimal_bit_representation()].concat()
        } else {
            vec![]
        }
```
this will avoid a reallocation in the `self.shifted_comm.is_none()` case

polycommit/src/marlin_pc/gadgets/commitment/commitment.rs
```diff
@@ -62,6 +64,28 @@ where
...
+        mut cs: CS,
+    ) -> Result<Vec<Boolean>, SynthesisError> {
+        let comm_booleans = self.comm.to_minimal_bit_representation(cs.ns(|| "comm"))?;
+        let shifted_comm_booleans = if let Some(shifted_comm) = &self.shifted_comm {
+            shifted_comm.to_minimal_bit_representation(cs.ns(|| "shifted_comm"))?
+        } else {
+            vec![]
+        };
+
+        Ok([comm_booleans, shifted_comm_booleans].concat())
```
ljedrz, line 85:
```suggestion
        let comm_booleans = self.comm.to_minimal_bit_representation(cs.ns(|| "comm"))?;
        if let Some(shifted_comm) = &self.shifted_comm {
            let shifted_comm_booleans = shifted_comm.to_minimal_bit_representation(cs.ns(|| "shifted_comm"))?;
            
            Ok([comm_booleans, shifted_comm_booleans].concat())
        } else {
            Ok(vec![])
        }
```

ljedrz (commented):
Left a few suggestions.

polycommit/src/marlin_pc/data_structures.rs
weikengchen: Edited. Note that the else branch would take `comm_bits` instead of `vec![]`.

algorithms/src/crh/bhp.rs
weikengchen: Revised the implementation back to the old one.

weikengchen: Decided to use the new implementation. The old one is somewhat complicated: it has to allocate `MAX_WINDOW_SIZE * MAX_NUM_WINDOWS` instead of `WINDOW_SIZE * NUM_WINDOWS`, so it is much larger. Given that this seems to be adding more complexity while the performance gain might be small, I will go ahead and simplify the logic instead.

weikengchen: Ah, change my mind, I will use the old implementation instead, as it does avoid additional allocation.

gadgets/src/algorithms/merkle_tree/tests.rs
weikengchen: Brought back.

curves/src/templates/short_weierstrass_jacobian/affine.rs
weikengchen: Now it is 377. There is an additional step in Fp256/384/... to truncate the last a few bits for `to_bits_le`. And for `to_bits_be`, it now simply uses the result of `to_bits_le` and reverse it.

marlin/src/marlin/circuit_verifying_key.rs
weikengchen: Updated. Now only the 64 bits are being included in the minimal bit representations, and the remaining bits are enforced to be smaller than u64::MAX, both in the outside and in the constraint world.

gadgets/src/algorithms/crh/poseidon.rs
```diff
@@ -35,20 +35,21 @@ impl<F: PrimeField + PoseidonDefaultParametersField, const RATE: usize, const OP
...
     type OutputGadget = FpGadget<F>;
 
-    fn check_evaluation_gadget<CS: ConstraintSystem<F>>(
+    fn check_evaluation_gadget_on_bits<CS: ConstraintSystem<F>>(
         &self,
         mut cs: CS,
-        input: Vec<UInt8>,
+        input: Vec<Boolean>,
     ) -> Result<Self::OutputGadget, SynthesisError> {
+        assert!(input.len() <= PoseidonCryptoHash::<F, RATE, OPTIMIZED_FOR_WEIGHTS>::INPUT_SIZE_BITS);
```
weikengchen, line 43: Poseidon is back, with additional safety checks.

weikengchen:
This PR is ready for a follow-up review. Almost all the concerns above should have been addressed.

Poseidon is back, with additional safety checks. The new trait is now `ToMinimalBis`.

.integration/tests/dpc_testnet1.rs
```diff
@@ -276,7 +275,7 @@ fn test_testnet1_dpc_execute_constraints() {
     println!("=========================================================");
     let num_constraints = inner_circuit_cs.num_constraints();
     println!("Inner circuit num constraints: {:?}", num_constraints);
-    assert_eq!(287247, num_constraints);
+    assert_eq!(321107, num_constraints);
```
weikengchen, line 278: This increase is due to the underpaying of record commitment and program ID commitment trees.

algorithms/src/merkle_tree/merkle_path.rs
```diff
@@ -37,7 +37,7 @@ impl<P: MerkleParameters> MerklePath<P> {
     pub fn verify<L: ToBytes>(&self, root_hash: &MerkleTreeDigest<P>, leaf: &L) -> Result<bool, MerkleError> {
         // Check that the given leaf matches the leaf in the membership proof.
         if !self.path.is_empty() {
-            let hash_input_size_in_bytes = (P::H::INPUT_SIZE_BITS / 8) * 2;
+            let hash_input_size_in_bytes = P::H::INPUT_SIZE_BITS / 8;
```
weikengchen, line 40: Previously, the Merkle tree never panics, because it always reserves twice the input size. 

So there is no panicking if the parameters is down by a half -> just somewhere in the code, the second half disappears.

algorithms/src/traits/merkle_tree.rs
```diff
@@ -32,6 +32,7 @@ pub trait MerkleParameters: Send + Sync + Clone {
 
     /// Returns the hash of a given leaf.
     fn hash_leaf<L: ToBytes>(&self, leaf: &L, buffer: &mut [u8]) -> Result<<Self::H as CRH>::Output, MerkleError> {
+        debug_assert_eq!(buffer.len(), Self::H::INPUT_SIZE_BITS / 8);
```
weikengchen, line 35: To prevent such overallocation of the Merkle tree buffer, a `debug_assert_eq` is left here.

dpc/src/parameters/testnet1.rs
```diff
@@ -150,21 +149,25 @@ impl Parameters for Testnet1Parameters {
...
     type ProgramCircuitTreeParameters = ProgramIDMerkleTreeParameters;
     
     type RecordCommitmentScheme = BHPCompressedCommitment<EdwardsBls12, 48, 50>;
     type RecordCommitmentGadget = BHPCompressedCommitmentGadget<EdwardsBls12, Self::InnerScalarField, EdwardsBls12Gadget, 48, 50>;
     type RecordCommitment = <Self::RecordCommitmentScheme as CommitmentScheme>::Output;
 
-    type RecordCommitmentTreeCRH = BHPCompressedCRH<EdwardsBls12, 8, 32>;
-    type RecordCommitmentTreeCRHGadget = BHPCompressedCRHGadget<EdwardsBls12, Self::InnerScalarField, EdwardsBls12Gadget, 8, 32>;
+    type RecordCommitmentTreeCRH = BHPCompressedCRH<EdwardsBls12, 16, 32>;
+    type RecordCommitmentTreeCRHGadget = BHPCompressedCRHGadget<EdwardsBls12, Self::InnerScalarField, EdwardsBls12Gadget, 16, 32>;
```
weikengchen, line 166: Increase the CRH size for the record commitment tree and the serial number tree.

dpc/src/parameters/testnet1.rs
```diff
@@ -181,15 +184,16 @@ impl Parameters for Testnet1Parameters {
...
     dpc_setup!{program_commitment_scheme, PROGRAM_COMMITMENT_SCHEME, ProgramCommitmentScheme, "AleoProgramIDCommitment0"} // TODO (howardwu): Rename to "AleoProgramCommitmentScheme0".
     dpc_setup!{program_circuit_id_crh, PROGRAM_CIRCUIT_ID_CRH, ProgramCircuitIDCRH, "AleoProgramIDCRH0"} // TODO (howardwu): Rename to "AleoProgramCircuitIDCRH0".
+    dpc_setup!{program_circuit_id_tree_crh, PROGRAM_CIRCUIT_ID_TREE_CRH, ProgramCircuitIDTreeCRH, "AleoProgramIDTreeCRH0"} // TODO (howardwu): Rename to "AleoProgramCircuitIDTreeCRH0".
     dpc_setup!{record_commitment_scheme, RECORD_COMMITMENT_SCHEME, RecordCommitmentScheme, "AleoRecordCommitment0"} // TODO (howardwu): Rename to "AleoRecordCommitmentScheme0".
     dpc_setup!{record_commitment_tree_crh, RECORD_COMMITMENT_TREE_CRH, RecordCommitmentTreeCRH, "AleoLedgerMerkleTreeCRH0"} // TODO (howardwu): Rename to "AleoRecordCommitmentTreeCRH0".
-    dpc_setup!{record_serial_number_tree_crh, RECORD_COMMITMENT_TREE_CRH, RecordCommitmentTreeCRH, "AleoRecordSerialNumberTreeCRH0"}
+    dpc_setup!{record_serial_number_tree_crh, RECORD_SERIAL_NUMBER_TREE_CRH, RecordSerialNumberTreeCRH, "AleoRecordSerialNumberTreeCRH0"}
```
weikengchen, line 190: Correct the parameters.

howardwu (approved):

howardwu:
LGTM, fantastic work, resampling parameters now, thanks @weikengchen

PR #350, opened 2021-08-09
Introduces `StateTransition` to `DPC` struct

howardwu:
## Motivation

Introduces `StateTransition` to `DPC` struct. This PR does not require resampling parameters.
- Updates `DPC::authorize` to reason over a state transition object
- Adds `ComputeKey` to the account model
- Updates `DPC::execute` to use compute keys for execution
- Updates `PrivateKey` to return `compute_key()` for execution
- Updates `genesis`, `testnet1_integration`, and `testnet2_integration` with state transition approach

## Test Plan

- Adds `snarkvm_dpc::state::Input` test for `new_noop`
- Adds `snarkvm_dpc::state::Output` test for `new_noop`
- Adds `snarkvm_dpc::state::StateTransition` test for `new_coinbase`
- Adds `snarkvm_dpc::state::StateBuilder` test for `add_noop_input` and `add_noop_output`

### DIFF
--- a/dpc/src/account/address.rs
+++ b/dpc/src/account/address.rs
@@ -14,7 +14,7 @@
 // You should have received a copy of the GNU General Public License
 // along with the snarkVM library. If not, see <https://www.gnu.org/licenses/>.
 
-use crate::{account_format, traits::Parameters, AccountError, PrivateKey, ViewKey};
+use crate::{account_format, AccountError, ComputeKey, Parameters, PrivateKey, ViewKey};
 use snarkvm_algorithms::{EncryptionScheme, SignatureScheme};
 use snarkvm_curves::AffineCurve;
 use snarkvm_utilities::{FromBytes, ToBytes};
@@ -42,7 +42,12 @@ pub struct Address<C: Parameters> {
 impl<C: Parameters> Address<C> {
     /// Derives the account address from an account private key.
     pub fn from_private_key(private_key: &PrivateKey<C>) -> Result<Self, AccountError> {
-        let decryption_key = private_key.to_decryption_key()?;
+        Self::from_compute_key(private_key.compute_key())
+    }
+
+    /// Derives the account address from an account compute key.
+    pub fn from_compute_key(compute_key: &ComputeKey<C>) -> Result<Self, AccountError> {
+        let decryption_key = compute_key.to_decryption_key()?;
         let encryption_key = C::account_encryption_scheme().generate_public_key(&decryption_key)?;
         Ok(Self { encryption_key })
     }
--- a/dpc/src/account/address.rs
+++ b/dpc/src/account/address.rs
@@ -82,8 +82,8 @@ impl<C: Parameters> Address<C> {
     }
 
     /// Returns the address as an encryption public key.
-    pub fn to_encryption_key(&self) -> &<C::AccountEncryptionScheme as EncryptionScheme>::PublicKey {
-        &self.encryption_key
+    pub fn to_encryption_key(self) -> <C::AccountEncryptionScheme as EncryptionScheme>::PublicKey {
+        self.encryption_key
     }
 }
 
--- a/dpc/src/account/address.rs
+++ b/dpc/src/account/address.rs
@@ -87,7 +87,7 @@ impl<C: Parameters> Address<C> {
     }
 
     /// Returns the address as an encryption public key.
-    pub fn to_encryption_key(self) -> <C::AccountEncryptionScheme as EncryptionScheme>::PublicKey {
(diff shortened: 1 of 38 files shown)


dpc/src/account/address.rs
```diff
@@ -82,8 +82,8 @@ impl<C: Parameters> Address<C> {
     }
 
     /// Returns the address as an encryption public key.
-    pub fn to_encryption_key(&self) -> &<C::AccountEncryptionScheme as EncryptionScheme>::PublicKey {
-        &self.encryption_key
+    pub fn to_encryption_key(self) -> <C::AccountEncryptionScheme as EncryptionScheme>::PublicKey {
```
Pratyush, line 85: Is the `PublicKey` `Copy`? If not, we should rename this to `into`, as per the API guidelines

dpc/src/dpc.rs
```diff
@@ -88,106 +88,67 @@ impl<C: Parameters> DPCScheme<C> for DPC<C> {
...
+            let private_key = match noop_private_key {
+                Some(noop_private_key) => noop_private_key,
+                None => {
+                    index += 1;
+                    &private_keys[index]
+                }
+            };
+
+            // Construct the prover key.
+            prover_keys.push((
```
Pratyush, line 121: This can clash with a SNARK prover key. Maybe `transaction_keys`?

howardwu: Will change to compute keys

dpc/src/account/address.rs
howardwu: Good catch

raychu86 (approved):

dpc/src/state/builder.rs
```diff
@@ -0,0 +1,371 @@
...
+    ///
+    /// Adds the given input into the builder.
+    ///
+    pub fn add_input(mut self, input: Input<C>) -> Self {
+        // Ensure there are no outputs assigned yet, as the builder computes joint serial numbers.
+        if !self.outputs.is_empty() {
+            self.errors.push("Builder cannot add new inputs after outputs".into());
+        }
+
+        match self.inputs.len() < C::NUM_INPUT_RECORDS {
```
weikengchen, line 58: Can the number of inputs "==" to `C::NUM_INPUT_RECORDS`?

dpc/src/state/builder.rs
```diff
@@ -0,0 +1,371 @@
...
+            self = self.add_input(input);
+        }
+        self
+    }
+
+    ///
+    /// Adds the given output into the builder.
+    ///
+    pub fn add_output(mut self, output: Output<C>) -> Self {
+        match self.outputs.len() < C::NUM_OUTPUT_RECORDS {
```
weikengchen, line 79: Same as above, can the number of outputs "==" `C::NUM_OUTPUT_RECORDS`?

dpc/src/state/builder.rs
```diff
@@ -0,0 +1,371 @@
...
+        }
+        self
+    }
+
+    ///
+    /// Appends the given data to the memo field in the builder.
+    ///
+    pub fn append_memo(mut self, data: &Vec<u8>) -> Self {
+        // TODO (howardwu): Change this to not be hardcoded to 64.
+        match self.memo.len() < 64 && (self.memo.len() + data.len()) < 64 {
```
weikengchen, line 101: Is is safe to have them "=="?

weikengchen (approved):
The code looks good. My only question is the one above: regarding the requirement that the number of inputs and outputs to be "<" than the limit, and why our tests do not hit this corner case.

dpc/src/state/builder.rs
howardwu: It cannot equal, that would exceed then

dpc/src/state/builder.rs
howardwu: Same, it would exceed

dpc/src/state/builder.rs
howardwu: Good catch the latter case can, the former cannot

howardwu: Fixed

dpc/src/state/builder.rs
weikengchen: Can you provide more info? Since intuitively it looks like a user can provide all the input records (i.e., no noop padding)?

weikengchen: Is there a compulsory padding or?

howardwu: This `<` just enforces there is at least 1 more input slot available

PR #352, opened 2021-08-10
Marlin proof early termination

### LINKED ISSUE #351: [Feature] Add early termination to POSW mine/prove
See [snarkOS#970](https://github.com/AleoHQ/snarkOS/issues/970).

Proposed implementation is to add a reference to an AtomicBool inside the mine & prove functions, and read to check abort status between major critical sections.

It's noticed one or few rounds are generally entered at lower difficulties, so we want sub-round aborting.

Protryon:
Fixes #351.

This enables snarkOS to terminate Marlin POSW proving early to help reduce orphan rate.

Benchmarking indicates no detectable performance decrease from added checks.

### DIFF
--- a/posw/src/posw.rs
+++ b/posw/src/posw.rs
@@ -50,7 +50,7 @@ use snarkvm_utilities::{to_bytes_le, FromBytes, ToBytes};
 
 use blake2::{digest::Digest, Blake2s};
 use rand::{rngs::OsRng, Rng};
-use std::marker::PhantomData;
+use std::{marker::PhantomData, sync::atomic::AtomicBool};
 
 /// Commits to the nonce and pedersen merkle root
 pub fn commit(nonce: u32, root: &PedersenMerkleRootHash) -> Vec<u8> {
@@ -217,6 +217,7 @@ where
         &self,
         subroots: &[[u8; 32]],
         difficulty_target: u64, // TODO: Change to Bignum?
+        terminator: &AtomicBool,
         rng: &mut R,
         max_nonce: u32,
     ) -> Result<(u32, Vec<u8>), PoswError> {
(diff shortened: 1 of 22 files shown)


weikengchen:
I can see this does require a large interface change. 

One question: Since each Marlin PoSW should only take a few seconds to prove, it is possible to not expose the level of SNARK verifier, but just check the termination bit every time one attempt to mine (i.e., generating a Marlin proof) has been made?

posw/src/posw.rs
```diff
@@ -225,15 +226,18 @@ where
...
-            proof = Self::prove(&pk, nonce, subroots, rng)?;
+            proof = Self::prove(&pk, nonce, subroots, terminator, rng)?;
+            i += 1;
 
             serialized_proof = to_bytes_le!(proof)?;
             if self.check_difficulty(&serialized_proof, difficulty_target) {
                 break;
             }
         }
+        println!("{} rounds", i);
```
weikengchen, line 240: Do we need to keep this debug info?

Protryon: removed

Protryon:
> I can see this does require a large interface change.
> 
> One question: Since each Marlin PoSW should only take a few seconds to prove, it is possible to not expose the level of SNARK verifier, but just check the termination bit every time one attempt to mine (i.e., generating a Marlin proof) has been made?

That was the question that debug code was written to answer. :P

No, it takes ~3 seconds on my laptop to do a single round, which is way too long to skip. We want a few milliseconds or less if possible.

howardwu:
@weikengchen I defer to you on design for this one. Looks okay to me, although you are better to evaluate the nuances and implications of early termination.

weikengchen (approved):
I think we can merge the code as it is. And thanks for simplifying the error logic of Marlin and PolyCommit.

I may take a deeper look in the future (I will add that to the optimization tracking list) to see if we can have more precise signal checks: FFT-ing takes time, MSM takes time. And in the end they are the major places.

Both of them go to other libraries. And as Max mentioned offline, if we use hardware optimization, it will be more complicated if we want to have early termination within MSM.

PR #358, opened 2021-08-12
Recognize mismatched keys in record decryption and add private key MSB check

weikengchen:
## Motivation

We first recall the DPC's record encryption framework. 

The blockchain contains encrypted records. A user, who has the viewing key, can decrypt records that belong to him/her. Note that the blockchain has everyone's records, so a user cannot decrypt most of the records on the blockchain. As a result, most of the times, the decryption result is wrong, because they are encrypted under another user's public key.

**Recognize mismatched keys in record decryption.** To enable this user to recognize ciphertexts that do not belong to him/her, so that the user can simply ignore the decryption result instead of trying to interpret the incorrect result, we commit to the public key in the ciphertext.

Now, each ciphertext additionally includes a commitment of the public key. The randomness of the commitment comes from the sponge (which absorbs the DH result). Now, a person who decrypts the ciphertext, performs the following steps:

1. given the encryption randomness rG and the private key, the person computes xrG (aka, the DH result).
2. the person creates a sponge and lets the sponge absorb xrG.
3. the person squeezes from the sponge and obtain the commitment randomness R.
4. given the public key commitment C and the public key xG, the person checks if C is a commitment of xG, under randomness R.
5. If so, the person decrypts the ciphertext, by squeezing more elements from the sponge and using them like in the counter mode. Otherwise, the person is sure that this ciphertext is not under the public key xG, and can skip it.

We use `commitment` rather than other cryptographic primitives for three reasons:
1. commitment provides hiding, so the transaction does not reveal who is the receiver (who can decrypt the record).
2. commitment provides binding, so even if the encryptor is malicious, it cannot commit to more than one users' public keys.
3. recall that the commitment is checked in the inner circuit, so therefore, we can be sure that the encryptor commits the right public key, exactly as the one used in the encryption.

This change increases the transaction sizes and number of constraints in the system, but not to a significant degree.

**Add the private key MSB check.** Now the ECIES Poseidon has an additional requirement about the private key: the private key must be, not just in the scalar field of the Edwards curve, but also within the capacity of the scalar field of Edwards curve. The same check is also enforced during allocation.

## Test Plan

The code passes all the existing tests (except the genesis and transaction size tests, which would be fixed in another PR) after parameter resampling.

## Related PRs

The fixing of the private key MSB check is a followup of #347.


algorithms/src/encryption/ecies_poseidon.rs
```diff
@@ -117,7 +117,26 @@ where
...
+            let mut flag = false;
+            for bit in bits
+                .iter()
+                .skip(<TE::ScalarField as PrimeField>::Parameters::CAPACITY as usize)
+            {
+                if *bit == true {
+                    flag = true;
+                    break;
+                }
+            }
```
ljedrz, line 134:
```suggestion
            let flag = bits
                .iter()
                .skip(<TE::ScalarField as PrimeField>::Parameters::CAPACITY as usize)
                .any(|bit| *bit);
```

algorithms/src/encryption/ecies_poseidon.rs
```diff
@@ -117,7 +117,26 @@ where
...
+
+            if flag == false {
```
ljedrz, line 136:
```suggestion
            if !flag {
```
(even if `clippy` doesn't complain about it yet, it will in the future, as it's a `bool` literal comparison)

algorithms/src/encryption/ecies_poseidon.rs
```diff
@@ -189,14 +218,32 @@ where
...
+        {
+            let bits = private_key.to_bits_le();
+            for bit in bits
+                .iter()
+                .skip(<TE::ScalarField as PrimeField>::Parameters::CAPACITY as usize)
+            {
+                if *bit == true {
+                    return Err(EncryptionError::InvalidPrivateKey);
+                }
+            }
```
ljedrz, line 244:
```suggestion
            if bits
                .iter()
                .skip(<TE::ScalarField as PrimeField>::Parameters::CAPACITY as usize)
                .any(|bit| *bit)
            {
                return Err(EncryptionError::InvalidPrivateKey);
            }
```

ljedrz (commented):
Left some minor suggestions.

howardwu:
It seems running the `testnet1/inner_snark.sh` script causes the following error now:
```
thread 'main' panicked at 'assertion failed: input.len() <= WINDOW_SIZE * NUM_WINDOWS', /snarkVM/gadgets/src/algorithms/crh/bhp.rs:92:9
stack backtrace:
   0: _rust_begin_unwind
   1: core::panicking::panic_fmt
   2: core::panicking::panic
   3: <snarkvm_gadgets::algorithms::crh::bhp::BHPCRHGadget<G,F,GG,_,_> as snarkvm_gadgets::traits::algorithms::crh::CRHGadget<snarkvm_algorithms::crh::bhp::BHPCRH<G,_,_>,F>>::check_evaluation_gadget_on_bits
   4: snarkvm_gadgets::traits::algorithms::crh::CRHGadget::check_evaluation_gadget
   5: snarkvm_dpc::circuits::inner_circuit::execute_inner_circuit
   6: snarkvm_algorithms::snark::groth16::generator::generate_random_parameters
   7: <snarkvm_algorithms::snark::groth16::snark::Groth16<E,V> as snarkvm_algorithms::traits::snark::SNARK>::setup
   8: inner_snark::main
note: Some details are omitted, run with `RUST_BACKTRACE=full` for a verbose backtrace.
```

weikengchen:
Let me handle it. Will get back soon.

howardwu:
@weikengchen it is very strange that the Transaction serialized size has regressed from 1278 to 1342 in the PR (despite the PR we just merged to master with the sizes working correctly)

gadgets/src/algorithms/prf/blake2s.rs
```diff
@@ -459,6 +461,16 @@ impl<F: PrimeField> AllocGadget<[u8; 32], F> for Blake2sOutputGadget {
     }
 }
 
+impl<F: PrimeField> ToBitsBEGadget<F> for Blake2sOutputGadget {
+    fn to_bits_be<CS: ConstraintSystem<F>>(&self, cs: CS) -> Result<Vec<Boolean>, SynthesisError> {
+        self.0.to_bits_le(cs)
```
howardwu, line 466: @weikengchen can you confirm this `to_bits_be` method should be returning `self.0.to_bits_le()`?

gadgets/src/algorithms/prf/blake2s.rs
```diff
@@ -459,6 +461,16 @@ impl<F: PrimeField> AllocGadget<[u8; 32], F> for Blake2sOutputGadget {
...
+    }
+
+    fn to_bits_be_strict<CS: ConstraintSystem<F>>(&self, cs: CS) -> Result<Vec<Boolean>, SynthesisError> {
+        self.0.to_bits_le_strict(cs)
```
howardwu, line 470: @weikengchen can you confirm this `to_bits_be_strict` method should be returning `self.0.to_bits_le_strict()`?

howardwu (approved):
LGTM once the above questions are resolved

weikengchen:
Pushed a fix. Follow up on this:

1. Yes, the tx size and block size increases are expected, because now the ciphertext includes an additional field element that commits the public key being used for encryption.

2. And yes, thanks for catching that. The `ToBitsBEGadget` is now correctly implemented for Blake2s. Indeed, the ToBitsBEGadget for UInt8 and [UInt8] are also missing. So, this PR adds them as well.

Ready for a follow-up review. It seems that we do not need to sample the parameters again.

howardwu (approved):

PR #360, opened 2021-08-13
Shorten the serial numbers

weikengchen:
## Motivation

A few PRs ago we discovered that the transaction sizes increase by 64. This happens because we use `to_bytes_le` to serialize the serial numbers, instead of `serialize` previously. These two APIs behave differently:
- `serialize`, from `CanonicalSerialize`, will keep minimal information, and therefore the final results is about 32 bytes, for the x coordinate of a point and a flag bit.
- `to_bytes_le`, from `ToBytes`, will keep both `x` and `y` of the point.

In this PR, we change how `ToBytes` works for the account signature public key, so that it also only keeps the x coordinate. Note that it is not always okay to strip away `y` and only keep `x`: commonly this is only okay in the twisted Edwards curve.

This PR, therefore, makes the following changes:
- propose a variant of Schnorr, called Schnorr compressed, that uses the observation above.
- keep the original variant as an option

The reason we keep the original variant is for compatibility, as the way how account signature public key is serialized, actually affects how viewing keys are generated (!), so changing the account signature public key serialization rule can cause a lot of keys to be invalid. It is a big breaking change that one should do with cautions.

So, basically, there are two versions:
- Schnorr, everything unchanged, with a small overhead in the transaction size.
- SchnorrCompressed, everything is changed, overhead is smaller.

## Test Plan

After resampling the parameterrrs, the tests should pass. 

## Related PRs

This is related to an old issue mentioned by Pratyush: https://github.com/AleoHQ/snarkVM/issues/277


algorithms/benches/signature/schnorr_compressed.rs
```diff
@@ -0,0 +1,113 @@
+// Copyright (C) 2019-2021 Aleo Systems Inc.
+// This file is part of the snarkVM library.
+
```
weikengchen, line 3: We originally have a bench program for Schnorr. This file is the bench program for Schnorr compressed.

algorithms/src/signature/schnorr_compressed.rs
```diff
@@ -0,0 +1,376 @@
...
+    Hash(bound = "TE: TwistedEdwardsParameters"),
+    Default(bound = "TE: TwistedEdwardsParameters")
+)]
+pub struct SchnorrCompressedPublicKey<TE: TwistedEdwardsParameters>(pub TEAffine<TE>);
+
+impl<TE: TwistedEdwardsParameters> ToBytes for SchnorrCompressedPublicKey<TE> {
+    #[inline]
+    fn write_le<W: Write>(&self, mut writer: W) -> IoResult<()> {
+        let x_coordinate = self.0.to_x_coordinate();
+        x_coordinate.write_le(&mut writer)
```
weikengchen, line 96: Note: SchnorrCompressed outputs only the x coordinate.

algorithms/src/signature/schnorr_compressed.rs
```diff
@@ -0,0 +1,376 @@
...
+            }
+        }
+
+        // Derive the public key.
+        let public_key = self.generate_public_key(private_key)?;
+
+        // Hash everything to get verifier challenge.
+        let mut hash_input = Vec::<TE::BaseField>::new();
+        hash_input.extend_from_slice(&prover_commitment.into_affine().x.to_field_elements().unwrap());
+        hash_input.extend_from_slice(&public_key.0.x.to_field_elements().unwrap());
```
weikengchen, line 251: Note: SchnorrCompressed's hash function only hashes the x coordinates.

algorithms/src/signature/tests.rs
```diff
@@ -14,12 +14,10 @@
...
 
-use crate::{signature::Schnorr, SignatureScheme};
-use snarkvm_curves::{edwards_bls12::EdwardsProjective as EdwardsBls12, edwards_bw6::EdwardsProjective as EdwardsBW6};
-use snarkvm_utilities::FromBytes;
-
-use rand::{Rng, SeedableRng};
+use crate::SignatureScheme;
+use rand::SeedableRng;
 use rand_chacha::ChaChaRng;
+use snarkvm_utilities::FromBytes;
```
weikengchen, line 21: Note: This file now tests both Schnorr and SchnorrCompressed.

algorithms/src/traits/signature.rs
```diff
@@ -28,17 +24,7 @@ pub trait SignatureScheme:
...
-        + Debug
-        + Default
-        + ToBytes
-        + FromBytes
-        + Hash
-        + Eq
-        + Send
-        + Sync
-        + CanonicalSerialize
-        + CanonicalDeserialize;
```
weikengchen, line 41: Note: Since DPC no longer uses `CanonicalSerialize` and `CanonicalDeserialize`, they are removed from the trait.

dpc/src/account/tests.rs
```diff
@@ -23,9 +23,9 @@ mod testnet1 {
     use rand_chacha::ChaChaRng;
     use std::{convert::TryInto, str::FromStr};
 
-    const ALEO_TESTNET1_PRIVATE_KEY: &str = "APrivateKey1tyBgFoCXAq8RfZT3W2mzVV9XzJb2hVFL2LrHLToEC37tXrz";
-    const ALEO_TESTNET1_VIEW_KEY: &str = "AViewKey1gFRs4gG66wFW1FypYRfwy6pDqix6rtquQ8uJ15RgHCC8";
-    const ALEO_TESTNET1_ADDRESS: &str = "aleo1shhq355tyaptcej65tkrweej5wth7wqg5hcqg3hf8as5s9rtrypqs8vy3u";
+    const ALEO_TESTNET1_PRIVATE_KEY: &str = "APrivateKey1y9jeNQybT9Mxk1AssbFmSXcFu9dG7sWkfYEsBUZrMin816z";
+    const ALEO_TESTNET1_VIEW_KEY: &str = "AViewKey1hNsfjkmrfiYWqMKtpKUW9LfGw93Pzz82UmmMn7pHHqZc";
+    const ALEO_TESTNET1_ADDRESS: &str = "aleo10kkut299n0fsmwu9z20hjmcdym4nv5s8adgtqaev3n3yqjlasqyqq6rk4d";
```
weikengchen, line 28: Note: The **first** valid private key under the new format is different, and therefore the first viewing key and the first address are different.

dpc/src/traits/parameters.rs
```diff
@@ -108,10 +106,7 @@ pub trait Parameters: 'static + Sized + Send + Sync {
         + PartialEq
         + Eq
         + Send
-        + Sync
-        + CanonicalSerialize
-        + CanonicalDeserialize
-        + PartialEq;
+        + Sync;
```
weikengchen, line 109: Note: DPC no longer uses CanonicalSerialize and CanonicalDeserialize for the account signing key.

gadgets/src/algorithms/signature/tests.rs
```diff
@@ -14,195 +14,398 @@
 // You should have received a copy of the GNU General Public License
 // along with the snarkVM library. If not, see <https://www.gnu.org/licenses/>.
 
-use crate::{
```
weikengchen, line 17: Note: Now we have gadget tests for both `Schnorr` and `SchnorrCompressed`.

parameters/src/testnet1/mod.rs
```diff
@@ -51,14 +51,14 @@ impl_params_remote!(
...
-    190452497
+    189389633
 );
 impl_params_local!(
     InnerSNARKVKParameters,
     inner_snark_vk_test,
     "./",
     "inner_snark_vk",
-    2426
+    2232
```
weikengchen, line 61: Note: Using `SchnorrCompressed` reduces the size of input to the inner circuit and outer circuit, so the size of the VK goes down.

dpc/src/circuits/inner_circuit.rs
```diff
@@ -412,17 +417,25 @@ pub fn execute_inner_circuit<C: Parameters, CS: ConstraintSystem<C::InnerScalarF
...
+                    &mut commitment_cs
+                        .ns(|| format!("If the input record {} is a noop, enforce it has a value of 0", i)),
+                    &zero_value_field_elements,
+                    &given_is_dummy,
+                )?;
+                given_payload_field_elements.conditional_enforce_equal(
+                    &mut commitment_cs
+                        .ns(|| format!("If the input record {} is a noop, enforce it has an empty payload", i)),
+                    &empty_payload_field_elements,
+                    &given_is_dummy,
```
weikengchen, line 436: Note: This part is a small change being piggybacked in this PR. @Pratyush mentioned that it is cheaper to compare the field elements that the bytes are packed into, rather than the bytes themselves.

howardwu (approved):

howardwu:
LGTM

parameters/src/testnet1/mod.rs
howardwu: @weikengchen How confident are you that these changes to VK size of inner and outer are correct?

Upon merging with `testnet2` (post #348), it seems the original sizes are correct, and I confirmed that SchnorrCompressed is being used in DPC circuits

weikengchen: I think I also need to double check---it is expected that the VK should be smaller because there are one fewer element in the input to the inner circuit.

weikengchen:
Follow up on the previous comment: 

- upon merging with the upstream, it seems that the inner SNARK VK is the short one (updated one). Can you take a look and check it is right?

- the Tx size is even smaller: now 1278 bytes, while the upstream testnet2 has 1342 bytes. So it is 64 bytes down (as expected, since we have two serial numbers, and each one is now one element shorter)

- the genesis block is smaller: 2230  -> 2166, which is also a gap of 64 bytes.

gadgets/src/algorithms/signature/schnorr_compressed.rs
```diff
@@ -0,0 +1,482 @@
...
+            <TEAffineGadget<TE, F> as GroupGadget<TEAffine<TE>, F>>::zero(cs.ns(|| "affine zero")).unwrap();
+
+        let prover_response_bytes = signature
+            .prover_response
+            .to_bytes(cs.ns(|| "prover_response_to_bytes"))?;
+
+        let prover_response_bits = prover_response_bytes
+            .iter()
+            .flat_map(|byte| byte.to_bits_le())
+            .collect::<Vec<_>>();
```
ljedrz, line 413:
```suggestion
```
it should be possible to keep this iterator un-collected to avoid an allocation

gadgets/src/algorithms/signature/schnorr_compressed.rs
```diff
@@ -0,0 +1,482 @@
...
+        let affine_zero: TEAffineGadget<TE, F> =
+            <TEAffineGadget<TE, F> as GroupGadget<TEAffine<TE>, F>>::zero(cs.ns(|| "affine zero")).unwrap();
+
+        let prover_response_bytes = signature
+            .prover_response
+            .to_bytes(cs.ns(|| "prover_response_to_bytes"))?;
+
+        let prover_response_bits = prover_response_bytes
+            .iter()
+            .flat_map(|byte| byte.to_bits_le())
```
ljedrz, line 412:
```suggestion
            .flat_map(|byte| byte.to_bits_le());
```

gadgets/src/algorithms/signature/schnorr_compressed.rs
```diff
@@ -0,0 +1,482 @@
...
+            <TE::ScalarField as PrimeField>::Parameters::CAPACITY as usize,
+            Boolean::Constant(false),
+        );
+
+        let hash = Boolean::le_bits_to_fp_var(cs.ns(|| "obtain the truncated hash"), &raw_hash_bits)?;
+        let result = signature
+            .verifier_challenge
+            .is_eq(cs.ns(|| "check verifier challenge"), &hash)?;
+
+        return Ok(result);
```
ljedrz, line 480:
```suggestion
        Ok(result)
```

ljedrz (commented):
Left a few suggestions.

gadgets/src/algorithms/signature/schnorr_compressed.rs
weikengchen: Yes, it works!

weikengchen:
Suggested changes from Lukasz have been applied.

weikengchen:
@howardwu This one is ready for parameter sampling.

This is better to put this in and then put the other one (#358), which I would clean it up after this one is merged.

howardwu:
Thanks! Got it, I will do this now.

PR #370, opened 2021-08-27
Arced system parameters

ljedrz:
This PR is a result of profiling done in snarkOS - the `SystemParameters` object is relatively expensive to clone, so wrapping it in an `Arc` can result in a noticeable performance improvement.


howardwu:
`SystemParameters` is removed in `testnet2` branch. Let me know if this is still useful for `testnet1` on the snarkOS side.

ljedrz:
If no further releases of snarkVM for `testnet1` are planned, this PR can be closed; the difference is noticeable in profiling, but it's not a blocker.

howardwu:
Since it's noticeable, let's merge it into master then, in case it becomes useful for posterity.

howardwu (approved):

PR #371, opened 2021-08-27
Parallel transaction verification

ljedrz:
While investigating block import time spikes in snarkOS, profiling showed that `DPCScheme::verify_transactions` accounted for a significant part of that process; I was wondering if it could be improved and noticed that it was run sequentially, while it could be parallelized - it just required some tweaks to traits. I was being conservative and only added `Sync` (and sometimes `Send`) where the compiler required it.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -1688,6 +1688,7 @@ dependencies = [
  "once_cell",
  "rand",
  "rand_chacha",
+ "rayon",
  "serde",
  "sha2",
  "snarkvm-algorithms",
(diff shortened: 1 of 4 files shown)


howardwu:
This looks to be desirable changes. Could you rebase this for `testnet2` branch?

ljedrz:
Done; it was much simpler with just `testnet2`, too :ok_hand:.

howardwu (approved):

howardwu:
Beautiful!

PR #372, opened 2021-08-27
Wrap pedersen parameters in Arcs

ljedrz:
Another finding from snarkOS profiling: cloning operations related to `BoweHopwoodPedersenCompressedCRH::hash` are costly. Wrapping them in `Arc`s improves hashing performance.

Ideally, the `Arc` would not even be needed here (it's only used so that the type can be changed from `BoweHopwoodPedersenCompressedCRH` to `BoweHopwoodPedersenCRH` for the purposes of hashing), but it's the simplest solution that doesn't affect the existing APIs.

### DIFF
--- a/algorithms/src/crh/bowe_hopwood_pedersen_parameters.rs
+++ b/algorithms/src/crh/bowe_hopwood_pedersen_parameters.rs
@@ -18,6 +18,8 @@ use snarkvm_curves::Group;
 
 use super::PedersenCRHParameters;
 
+use std::sync::Arc;
+
 pub const BOWE_HOPWOOD_CHUNK_SIZE: usize = 3;
 pub const BOWE_HOPWOOD_LOOKUP_SIZE: usize = 2usize.pow(BOWE_HOPWOOD_CHUNK_SIZE as u32);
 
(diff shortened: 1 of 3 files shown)


weikengchen:
One question: can the problem be solved if `BoweHopwoodPedersenCompressedCRH` has a different definition?

Currently, it is:
```rust
#[derive(Debug, Clone, PartialEq, Eq)]
pub struct BoweHopwoodPedersenCompressedCRH<
    G: Group + ProjectiveCurve,
    const NUM_WINDOWS: usize,
    const WINDOW_SIZE: usize,
> {
    pub parameters: PedersenCRHParameters<G, NUM_WINDOWS, WINDOW_SIZE>,
    pub bowe_hopwood_parameters: BoweHopwoodPedersenCRHParameters<G>,
}
```

but it can also just be simply a copy of `BoweHopwoodPedersenCRH`: (untested)
```rust
#[derive(Debug, Clone, PartialEq, Eq)]
pub struct BoweHopwoodPedersenCompressedCRH<
    G: Group + ProjectiveCurve,
    const NUM_WINDOWS: usize,
    const WINDOW_SIZE: usize,
> (
    pub BoweHopwoodPedersenCRH<G, NUM_WINDOWS, WINDOW_SIZE>
);
```

which seem to avoid the copying that you discovered.

ljedrz:
As long as as a result the `BoweHopwoodPedersenCompressedCRH::hash` method doesn't involve deep-cloning vectors - yes, this would solve the issue :+1:.

weikengchen:
(I actually don't know which solution would be better [and the second one may involve other complications], so it would leave it to your and others' discretion .)

howardwu:
Lets move this to the `testnet2` branch?

ljedrz:
Depending on how soon the `testnet2` is going to go live, it could still be practical in `testnet1`, but it's not a blocker - your call @howardwu :+1:.

PR #379, opened 2021-08-31
implement account from_private_key

collinc97:
## Motivation

I am currently refactoring the aleo sdk to be compatible with the testnet2 branch of snarkVM.  
One desired feature of the sdk is to import an account from a private key - example: `Account::from_private_key()`.  
The sdk uses the `Account<C: Parameters>` struct in snarkVM which does not provide this functionality.  
Rather than re-implement the `Account` struct in aleo sdk this PR adds the desired method in snarkVM.

## Changes
* add `Account::from_private_key()` method to `AccountScheme` trait.
* implement `Account::from_private_key()` method for generic `Account<C: Parameters>` struct.
* add test for `Account::from_private_key()` in account tests.

### DIFF
--- a/dpc/src/account/account.rs
+++ b/dpc/src/account/account.rs
@@ -36,6 +36,12 @@ impl<C: Parameters> AccountScheme for Account<C> {
     /// Creates a new account.
     fn new<R: Rng + CryptoRng>(rng: &mut R) -> Result<Self, AccountError> {
         let private_key = PrivateKey::new(rng);
+
+        Self::from_private_key(private_key)
+    }
+
+    /// Creates an account from a private key.
+    fn from_private_key(private_key: PrivateKey<C>) -> Result<Self, AccountError> {
         let view_key = ViewKey::try_from(&private_key)?;
         let address = Address::try_from(&view_key)?;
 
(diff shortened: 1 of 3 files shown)


weikengchen:
The PR looks good. (But I would leave it to Howard to take a look as well.)

The PR includes a test on testnet1. I am thinking, but unsure, whether we want the same test in testnet2, for the parameters in testnet2, as well. 

I believe that the account private key, at this moment, is the same in testnet1 and testnet2. But just would like to be more sure about it.

collinc97:
Added testnet2 `Account::from_private_key()` test.

Good catch @weikengchen thank you for reviewing.

weikengchen:
(Can you test it over the TESTNET2 constants? Actually that is the part that I am not sure about.)

```rust
    const ALEO_TESTNET2_PRIVATE_KEY: &str = "APrivateKey1y9jeNQybT9Mxk1AssbFmSXcFu9dG7sWkfYEsBUZrMin816z";
    const ALEO_TESTNET2_VIEW_KEY: &str = "AViewKey1hNsfjkmrfiYWqMKtpKUW9LfGw93Pzz82UmmMn7pHHqZc";
    const ALEO_TESTNET2_ADDRESS: &str = "aleo10kkut299n0fsmwu9z20hjmcdym4nv5s8adgtqaev3n3yqjlasqyqq6rk4d";
```

weikengchen (approved):
The PR looks good. And this is a simple PR that adds a method for SDK, so should be ready to merge.

raychu86 (approved):

howardwu (approved):

PR #389, opened 2021-09-09
Removes `signatures` from `Transaction` struct, now verified in `InnerCircuit`

howardwu:
## Motivation

This PR removes `signatures` from the `Transaction` struct, and moves it into the `InnerCircuit` for verification. As part of this, signature rerandomization is also removed, and the serial numbers are now updated to be a straight `PRF::Output`.

Transactions have been reduced from 1342 bytes to 1214 bytes.

### Transaction Signatures
- Removes randomization from transaction signatures
- Removes `signature_randomizers` from `StateTransition` and `Input`

### Inner Circuit
- Adds transaction signature verification without randomized public key

### Integer Gadgets
- Adds `ToBytesGadget` support for signed integers

## Test Plan

- Adds `test_int_to_bytes` for all signed integers
- Adds `test_uint_to_bytes` for all unsigned integers


howardwu:
As discussed offline, further improvement can be made using Poseidon. As this requires changing PRF input, seed, output types to a field, this requires updating how account private keys are represented as well.

Given the scope of change, I will checkpoint this PR here, and start a new PR for those changes as part of larger account upgrades.

raychu86 (approved):

PR #398, opened 2021-09-15
Adds account seed domain separators, adds input length encoding in PRF

howardwu:
## Motivation

- Adds account seed domain separators
- Adds input length encoding in `PoseidonPRF` and `PoseidonPRFGadget`

## Test Plan

- Adds `test_prf` for `PoseidonPRFGadget`
- Updates DPC account canaries in tests


howardwu:
Resampling parameters now.

raychu86 (approved):

.integration/tests/dpc_testnet2.rs
```diff
@@ -28,8 +28,8 @@ use std::time::{SystemTime, UNIX_EPOCH};
 #[test]
 fn test_testnet2_inner_circuit_id_sanity_check() {
     let expected_inner_circuit_id = vec![
```
Pratyush, line 30: btw maybe we should be encoding these as `b"blahblahblah"`?

dpc/src/account/account_format.rs
```diff
@@ -15,6 +15,8 @@
 // along with the snarkVM library. If not, see <https://www.gnu.org/licenses/>.
 
 pub static ACCOUNT_ENCRYPTION_AND_SIGNATURE_INPUT: &str = "AleoAccountEncryptionAndSignatureScheme0";
+pub static ACCOUNT_SEED_SK_SIG_DOMAIN: &str = "AleoAccountSeedSignatureSecretKey0";
+pub static ACCOUNT_SEED_R_SIG_DOMAIN: &str = "AleoAccountSeedSignatureRandomizer0";
```
Pratyush, line 19: Maybe this should be something like 
`pub const ACCOUNT_blah_blah: &'static [u8] = b"AleoAccountBlahBlahBlah"`?

dpc/src/account/account_format.rs
```diff
@@ -15,6 +15,8 @@
...
 
 pub static PRIVATE_KEY_PREFIX: [u8; 11] = [127, 134, 189, 116, 210, 221, 210, 137, 145, 18, 253]; // APrivateKey1
 pub static _COMPUTE_KEY_PREFIX: [u8; 10] = [109, 249, 98, 224, 36, 15, 213, 187, 79, 190]; // AComputeKey1
```
Pratyush, line 22: Same here, so that we don't have to encode it as numbers.

dpc/src/account/private_key.rs
```diff
@@ -72,12 +79,20 @@ impl<C: Parameters> PrivateKey<C> {
 impl<C: Parameters> From<&C::AccountSeed> for PrivateKey<C> {
     /// Returns the account private key from an account seed.
     fn from(seed: &C::AccountSeed) -> Self {
+        // Construct the sk_sig domain separator.
+        let sk_sig_input = ACCOUNT_SEED_SK_SIG_DOMAIN;
+        let sk_sig_domain = C::ProgramScalarField::from_bytes_le_mod_order(&sk_sig_input.as_bytes());
+
+        // Construct the r_sig domain separator.
+        let r_sig_input = format!("{}_{}", ACCOUNT_SEED_R_SIG_DOMAIN, 0);
+        let r_sig_domain = C::ProgramScalarField::from_bytes_le_mod_order(r_sig_input.as_bytes());
```
Pratyush, line 88: Maybe we should hash these using Blake2s first? If we're not checking this derivation in circuit (and we probably shouldn't be), then hashing makes sure it fits in 32 bytes.

PR #414, opened 2021-09-23
IR updates

0rphon:
## Motivation

This adds support for new leo features in the IR. Another branch currently exists on leo that implements IR using these changes.

Adds:
- support for reverse/inclusive ranges
- bug fixes for complex array indexing
- support for unsized arrays
- moves the `len` method from leo into the eval core

## Test Plan

all tests are passing. once this branch and the leo branch are merged for DP3 it will be tested against the fuzzer

## Related PRs

follow up to #218
developed alongside the [leo ir branch](https://github.com/AleoHQ/leo/tree/ir)


ir/src/values/mod.rs
```diff
@@ -41,6 +41,19 @@ pub enum Integer {
...
 }
 
+impl Integer {
+    pub fn get_unsigned(&self) -> u32 {
+        match self {
+            Self::U8(n) => *n as u32,
+            Self::U16(n) => *n as u32,
+            Self::U32(n) => *n,
+            Self::U64(n) => u32::try_from(*n).unwrap_or_else(|e| panic!("{}: {:?}", e, self)),
+            Self::U128(n) => u32::try_from(*n).unwrap_or_else(|e| panic!("{}: {:?}", e, self)),
```
howardwu, line 51: Just to confirm (without context), is downcasting the intended behavior here?

0rphon: to offer context: i wrote it so that array ranges could properly calculate intermediate slice length during complex array access (ie `x[0..6][0..2][1]`). it panics if the int is bigger than u32::MAX because array ranges use u32 for len.

that being said, yes it is essentially just downcasting to a u32. ill move it to a downcast impl now. thanks!

0rphon:
oh actually i almost forgot i still need to regenerate the IR tests. Ci just wasn't running them for some reason

Protryon (approved):

PR #416, opened 2021-09-28
Ir updates (fixed)

0rphon:
## Motivation

This is a second PR for the #414 PR that had to be reverted because i failed to update its cargo.toml versions before merging.

This adds support for new leo features in the IR. Another branch currently exists on leo that implements IR using these changes. 

Adds:
- support for reverse/inclusive ranges
- bug fixes for complex array indexing
- support for unsized arrays
- moves the len method from leo into the eval core

## Test Plan

all tests are passing. once this branch and the leo branch are merged for DP3 it will be tested against the fuzzer

## Related PRs

follow up to #218
developed alongside the [leo ir branch](https://github.com/AleoHQ/leo/tree/ir)

PR #418, opened 2021-09-29
[Feature] Bits Bytes IR core function

gluax:
## Motivation

To bits and to bytes core function calls in snarkvm for leo.

Right now on the master branch as far as I could tell only implements to_bits and from_bits for integer types. So the core calls exist on the other types to error out for now till those are in. 

## Test Plan

The IR is already thoroughly tested and all tests currently pass.

## Related PRs

Should be merged into #416, before 416 is merged into master, or merged into master after 416 is merged into master.


howardwu (dismissed):
Before proceeding further on bits and bytes, we need to change convention to `to_bits_le` and `to_bytes_le`, or `_be` if that's what's going on here

eval/src/value/value.rs
```diff
@@ -106,6 +109,36 @@ impl<F: PrimeField, G: GroupType<F>> ConstrainedValue<F, G> {
             (_, _) => false,
         }
     }
+
+    pub fn to_bits_le(&self) -> Result<Vec<Boolean>> {
+        use ConstrainedValue::*;
+
+        match self {
+            Address(_) => Err(anyhow!("the type `address` does not implement the to_bytes method")),
```
collinc97, line 117: should these error messages read "... to_bits_le method"

eval/src/value/value.rs
```diff
@@ -106,6 +109,36 @@ impl<F: PrimeField, G: GroupType<F>> ConstrainedValue<F, G> {
...
+            Array(_) => Err(anyhow!("the type `array` does not implement the to_bytes method")),
+            Tuple(_) => Err(anyhow!("the type `tupple` does not implement the to_bytes method")),
+        }
+    }
+
+    pub fn to_bytes_le(&self) -> Result<Vec<UInt8>> {
+        use ConstrainedValue::*;
+
+        match self {
+            Address(_) => Err(anyhow!("the type `address` does not implement the to_bytes method")),
```
collinc97, line 132: similarly "... to_bytes_le method"

collinc97 (changes requested):

eval/src/value/value.rs
```diff
@@ -106,6 +109,36 @@ impl<F: PrimeField, G: GroupType<F>> ConstrainedValue<F, G> {
...
+
+        match self {
+            Address(_) => Err(anyhow!("the type `address` does not implement the to_bits_le method")),
+            Boolean(_) => Err(anyhow!("the type `bool` does not implement the to_bits_le method")),
+            Char(_) => Err(anyhow!("the type `char` does not implement the to_bits_le method")),
+            Field(_) => Err(anyhow!("the type `field` does not implement the to_bits_le method")),
+            Group(_) => Err(anyhow!("the type `group` does not implement the to_bits_le method")),
+            Integer(integer) => Ok(integer.to_bits_le()),
+            Array(_) => Err(anyhow!("the type `array` does not implement the to_bits_le method")),
+            Tuple(_) => Err(anyhow!("the type `tupple` does not implement the to_bits_le method")),
```
collinc97, line 124:
```suggestion
            Tuple(_) => Err(anyhow!("the type `tuple` does not implement the to_bits_le method")),
```

eval/src/value/value.rs
```diff
@@ -106,6 +109,36 @@ impl<F: PrimeField, G: GroupType<F>> ConstrainedValue<F, G> {
...
+
+        match self {
+            Address(_) => Err(anyhow!("the type `address` does not implement the to_bytes_le method")),
+            Boolean(_) => Err(anyhow!("the type `bool` does not implement the to_bytes_le method")),
+            Char(_) => Err(anyhow!("the type `char` does not implement the to_bytes_le method")),
+            Field(_) => Err(anyhow!("the type `field` does not implement the to_bytes_le method")),
+            Group(_) => Err(anyhow!("the type `group` does not implement the to_bytes_le method")),
+            Integer(_) => Err(anyhow!("the type `int` does not implement the to_bytes_le method")),
+            Array(_) => Err(anyhow!("the type `array` does not implement the to_bytes_le method")),
+            Tuple(_) => Err(anyhow!("the type `tupple` does not implement the to_bytes_le method")),
```
collinc97, line 139:
```suggestion
            Tuple(_) => Err(anyhow!("the type `tuple` does not implement the to_bytes_le method")),
```

collinc97 (dismissed):
grammar changes

collinc97:
@gluax can we procedurally generate all these methods instead of hard coding them all?
We're expecting ourselves to go back and implement all of these methods at a later date under current assumptions.
But I fear those assumptions may change once we start designing the new programming model for testnet2.

gluax:
> 
> 
> @gluax can we procedurally generate all these methods instead of hard coding them all? We're expecting ourselves to go back and implement all of these methods at a later date under current assumptions. But I fear those assumptions may change once we start designing the new programming model for testnet2.

We cannot as the function does not have the context to what type it should be outputting at the IR for a generic solution. So we need a method per on the from ones.

0rphon (approved):

Protryon:
Can we use some rust macros to generate the boilerplate rather than making the IR dynamically typed? I.e. like we do for a `Integer` impls in `snarkvm_algorithms`.

PR #426, opened 2021-10-03
Enable WASM compilation for Marlin

julesdesmit:
## Motivation

Enable WASM compilation of Marlin, so that the DPC package can be used for the Inner setup.


howardwu (approved):

howardwu:
LGTM, thanks @julesdesmit !

PR #430, opened 2021-10-05
Update CommiterKey Serialization

raychu86:
## Motivation

This PR implements a custom serialization for the SonicKZG10 `CommiterKey` (to replace cannonical serialization/deserialization), which reduces the load time of Marlin Proving keys significantly. The most notable speedup reduces the load time of the POSW proving key by 98%.


howardwu:
Is it convenient to add a test to sanity check that the `ToBytes` and `FromBytes` impls work with each other?

howardwu (approved):

howardwu:
LGTM

PR #433, opened 2021-10-05
Reduce the number of allocations in some hot functions

ljedrz:
snarkOS profiling results indicate that allocations are non-negligible, especially while fast-forwarding blocks; this PR targets allocations in some hot functions.

### DIFF
--- a/algorithms/src/snark/groth16/mod.rs
+++ b/algorithms/src/snark/groth16/mod.rs
@@ -135,19 +135,19 @@ impl<E: PairingEngine> Proof<E> {
 
     /// Deserialize a proof from compressed or uncompressed bytes.
     pub fn read<R: Read>(mut reader: R) -> IoResult<Self> {
-        // Construct the compressed reader.
         let compressed_proof_size = Self::compressed_proof_size()?;
-        let mut proof_reader = vec![0u8; compressed_proof_size];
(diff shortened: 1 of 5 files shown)


algorithms/src/snark/groth16/mod.rs
```diff
@@ -135,19 +135,19 @@ impl<E: PairingEngine> Proof<E> {
...
     /// Deserialize a proof from compressed or uncompressed bytes.
     pub fn read<R: Read>(mut reader: R) -> IoResult<Self> {
-        // Construct the compressed reader.
         let compressed_proof_size = Self::compressed_proof_size()?;
-        let mut proof_reader = vec![0u8; compressed_proof_size];
-        reader.read_exact(&mut proof_reader)?;
+        let uncompressed_proof_size = Self::uncompressed_proof_size()?;
+
+        // Construct the compressed reader.
+        let mut proof_reader = vec![0u8; uncompressed_proof_size];
```
howardwu, line 142: Just to check, is it more efficient/cheaper to allocate double the expected size first, rather than resizing after the compressed case is checked?

ljedrz: Yes - it is faster, and for hot enough functions the difference in performance is detectable.

howardwu (approved):

PR #437, opened 2021-10-07
Remove log entry

damons:
## Motivation

Nodes print "Ledger already contains this transaction memo" all the time and clutter up the stdout for snarkOS.   I believe we should be using the standard log functions that we normally use at the proper logging level.  F


damons:
Nevermind on this PR.  Pulling out my insignificant changes.

PR #468, opened 2021-11-01
Update record commitments

howardwu:
## Motivation

- Removes `serial_number_nonce` from `Record` struct
- Formalizes the concept of a `record_view_key` for use in transactions


raychu86 (approved):

collinc97 (approved):

Pratyush:
Update so far: I believe I have things mechanically worked out in the circuit and in the native code, but I'm not sure of one thing: should `RecordCiphertext` include the `KeyCommitment` (which is `H(g^ar)`)? That's the main remaining inconsistency between the circuit and the main code.

Also, we should provide an abstraction for opening a record's commitment inside a circuit; I imagine this would be useful also for the program writers/Leo backend code.

dpc/src/virtual_machine/builder.rs
```diff
@@ -132,12 +132,12 @@ impl<N: Network> ResponseBuilder<N> {
...
+        let (output_records, encryption_randomness): (Vec<_>, Vec<_>) = outputs
             .iter()
             .enumerate()
             .take(N::NUM_OUTPUT_RECORDS)
             .map(|(i, output)| {
-                let record = output.to_record(serial_numbers[i], rng)?;
+                let record = output.to_record(rng)?;
 
                 // Add the record view key event if the output record is public.
                 if output.is_public() && events.len() < N::NUM_EVENTS as usize {
```
howardwu, line 143: @raychu86 we can add the RecordViewKey here now, after this PR lands.

dpc/src/network/testnet2.rs
```diff
@@ -192,9 +197,8 @@ impl Network for Testnet2 {
     type CiphertextIDCRHGadget = BHPCRHGadget<Self::ProgramProjectiveCurve, Self::InnerScalarField, Self::ProgramAffineCurveGadget, 41, 63>;
     type CiphertextID = AleoLocator<<Self::CiphertextIDCRH as CRH>::Output, { Self::RECORD_CIPHERTEXT_ID_PREFIX }>;
```
howardwu, line 198: Note to self: I believe I can delete this now, and its corresponding instantiation methods below.

dpc/src/network/testnet1.rs
```diff
@@ -120,6 +121,7 @@ impl Network for Testnet1 {
     const RECORD_SIZE_IN_BYTES: usize = 280;
     const RECORD_CIPHERTEXT_SIZE_IN_BYTES: usize = 320;
```
howardwu, line 122: Note to self: This might be wrong. Ensure there is a test that checks this value before merging the PR.

dpc/src/transition/transition.rs
```diff
@@ -231,12 +225,12 @@ impl<N: Network> Transition<N> {
     ///
     #[inline]
     pub(crate) fn compute_transition_id(
-        serial_numbers: &[N::SerialNumber],
-        commitments: &[N::Commitment],
-        ciphertexts: &[RecordCiphertext<N>],
+        serial_numbers: &Vec<N::SerialNumber>,
+        commitments: &Vec<N::Commitment>,
+        _ciphertexts: &Vec<N::RecordCiphertext>,
```
howardwu, line 230: Note to self: I can remove the unused `_ciphertexts` from computing the transition ID, as it is now a part of the commitment.

Separate note to self: Add a check in `Transition::verify` (if possible) to check `ciphertext -> commitment` relationship.

howardwu:
Resampling parameters now

PR #470, opened 2021-11-02
Update virtual machine initialization and expose transitions

raychu86:
## Motivation

The virtual machine only needs the ledger root, and not the entire ledger to operate. The `Transitions` struct is exposed to allow the nodes to construct local proofs.


howardwu (approved):
LGTM

PR #475, opened 2021-11-03
Switch to Blake2s-ChaCha-based Sponge for PoSW

Pratyush:
## Motivation

Greatly improves PoSW verification times by switching from a Poseidon-based sponge to a Blake2s+ChaCha-based sponge. This is fine as we don't recursively verify the PoSW proofs.

Before:
Screen Shot 2021-11-03 at 4 18 16 PM

After:
Screen Shot 2021-11-03 at 4 12 46 PM

## Test Plan

Existing PoSW tests should cover this.

## Related PRs

(Link your related PRs here)


Pratyush:
Hm the tests pass locally, so not sure why they're failing here. I updated the parameters as per the instructions in `snarkvm-parameters`.

Issue #496, opened 2021-11-19
[Bug] Non deterministic proofs


damirka:
## 🐛 Bug Report

We've encountered that issue while working on [Hangman game](https://github.com/AleoHQ/aleo-education/tree/main/hangman) implementation in Leo.

In short: proofs can be either valid or invalid and this behavior is unpredictable.

## Steps to Reproduce

Smallest Leo code that reproduces this behavior is this:
```ts
// [main]
// x: bool = false;
// y: bool = true;

function main(x: bool, y: bool) {
    if y {
        x = false;
        y = true;
    }
}
```

IR for it looks like this:
```asm
decl f0: <0>
  store &v0, ((), (), (), ())
  mask 2, v2
    store &v1, false
    store &v2, true
```

## Expected Behavior

Code should deterministically produce valid (or invalid) proof.

## Your Environment

- SnarkVM master 7ed56dd679931f14471b157857b701e6bb6bde41
- 1.56
- Big Sur

PR #497, opened 2021-11-19
Fix non-deterministic proofs (in IR)

damirka:
## Motivation

Described in the issue #496.

## Test Plan

Not quite sure how to test this behavior as it was hard to catch it in the first place. Ran multiple tests locally to verify that order is preserved.

> Fix looks super easy, but @d0cd and myself spent hours debugging and tracing it to this point.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -88,15 +88,6 @@ source = "registry+https://github.com/rust-lang/crates.io-index"
 checksum = "cf1de2fe8c75bc145a2f577add951f8134889b4795d47466a54a5c846d691693"
 
 [[package]]
-name = "bitmaps"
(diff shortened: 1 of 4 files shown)


gluax (approved):
LGTM. I would guess that `im::HashMap` is nondeterministic in order, and that's why this would fix the issue.

damirka:
@gluax yup!

d0cd (approved):
Looks good! 
Just curious, is there any significance of using the immutable HashMap? and do we have to worry about any fallout by using IndexMap?

acoglio (approved):

collinc97 (approved):

PR #518, opened 2021-12-06
Update MAXIMUM_FUTURE_BLOCK_TIME to 90 seconds

raychu86:
## Motivation

This PR updates the `MAXIMUM_FUTURE_BLOCK_TIME` from 2 hours to 90 seconds. Meaning a new block's timestamp can be at most 90 seconds ahead of the node's current time.

Additionally, the miner now has an additional consideration, where the new block timestamp must be greater than the previous block timestamp.

Additional Notes:
Bitcoin mining protocol allows for 2 hour future block timestamps, however that is not an issue due to block difficulty adjustments every 2016 blocks.

### DIFF
--- a/dpc/src/ledger/blocks.rs
+++ b/dpc/src/ledger/blocks.rs
@@ -22,7 +22,8 @@ use chrono::Utc;
 use itertools::Itertools;
 use std::collections::HashMap;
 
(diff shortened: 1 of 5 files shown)


howardwu (approved):

howardwu:
Note that this time is expected to come down further as additional testing demonstrates safety.

PR #529, opened 2021-12-16
Allow modification of the difficulty target

jules:
## Motivation

This method is necessary for pool operators in order
to restore proposed blocks to their original form, so they can be
properly checked and broadcast in case they are a valid solution.

## Related PRs

https://github.com/AleoHQ/snarkOS/pull/1427


ljedrz (approved):
Code-wise everything is fine :ok_hand:.

PR #543, opened 2021-12-21
Feature: Transactions to decrypted records

collinc97:
## Overview
Implements `to_decrypted_records` on the Transactions struct.

## Motivation
This will be used by the snarkOS cli to decrypt records belonging to the given account view key.

I know that `Transactions` implements `Deref` to convert a reference into a vector however, you cannot dereference a struct inside of a map. This prevents you from taking advantage of rayon's `into_par_iter` methods we want to use.


dpc/src/block/transactions.rs
```diff
@@ -176,6 +176,14 @@ impl<N: Network> Transactions<N> {
     ) -> Result<MerklePath<N::TransactionsRootParameters>> {
         Ok(self.tree.generate_proof(index, &leaf)?)
     }
+
+    /// Returns records from the transactions belonging to the given account view key.
+    pub fn to_decrypted_records(&self, account_view_key: &ViewKey<N>) -> Vec<Record<N>> {
+        self.transactions
+            .iter()
+            .flat_map(|transaction| transaction.to_decrypted_records(account_view_key))
+            .collect()
```
howardwu: Instead of collecting, can you return this as an `impl Iterator`?

(ditto for `Transaction::to_decrypted_records` if it is not an iterator already)

You will see significant performance gains in what you're doing.

collinc97: Yes I can remove the '.collect' calls.

Thoughts on calling '.into_par_iter'?

howardwu: If you can make it a conditional, maybe for now set it to around 500. So if its >500 use par_iter, if its below, use iter.

howardwu: @collinc97

howardwu (approved):

PR #565, opened 2022-01-05
Additional clippy overhaul

howardwu:
## Motivation

Additional clippy overhaul, this PR addresses benches, tests, examples, etc.

### DIFF
--- a/algorithms/src/msm/variable_base/cuda.rs
+++ b/algorithms/src/msm/variable_base/cuda.rs
@@ -16,7 +16,7 @@
 
 use snarkvm_curves::{
     bls12_377::{Fq, Fr, G1Affine, G1Projective},
-    traits::{AffineCurve, Group, ProjectiveCurve},
+    traits::{AffineCurve, Group},
 };
 use snarkvm_fields::{PrimeField, Zero};
 use snarkvm_utilities::BitIteratorBE;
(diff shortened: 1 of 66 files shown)


circuits/src/programs/mod.rs
```diff
@@ -101,7 +97,11 @@ impl<E: Environment> Memory<E> {
         match self.registers.get(locator.0 as usize) {
             Some(register) => match register.get().is_some() {
                 true => panic!("Register {} is already set", locator.0),
-                false => register.set(value.clone()),
+                false => {
```
Pratyush, line 100: The only non-trivial change seems to be this one; was this intended? If so, the rest LGTM

algorithms/src/msm/variable_base/cuda.rs
```diff
@@ -224,10 +224,9 @@ pub(super) fn msm_cuda<G: AffineCurve>(
         unimplemented!("trying to use cuda for unsupported curve");
     }
 
-    if bases.len() < scalars.len() {
-        scalars = &scalars[..bases.len()];
-    } else if bases.len() > scalars.len() {
-        bases = &bases[..scalars.len()];
```
howardwu, line 230: @raychu86 the equal case was not previously considered. Can you sanity check this?

raychu86: The equals case was not considered before because there was no need to truncate. 

The new changes will effectively have the same output. Looks fine to me.

raychu86 (approved):

PR #581, opened 2022-01-14
Update build.sh

unordered-set:
Add more supported archs, because benchmarks are failing on my A100.


raychu86:
Adding additional architectures could cause older machines to fail to compile. We'll likely look into a 1 time startup generation of the `fatbin` for each individual machine instead of having one provided in the codebase.

PR #588, opened 2022-01-15
[Ignore][Tracking] Experimental development

howardwu:
## Tracking

Experimental development

## TODOs

- [ ] Hardcode the noop program proof to improve performance
- [ ] Introduce a verifying key abstraction for transitions
- [ ] Abstract the `Transaction` struct to support deploy transactions

Need to check the following:
- [ ] Add test to check that a hiding PoSW proof fails DPC consensus


howardwu:
^This switches the `Transition` struct from an outer proof to a combination of an inner proof and program proof.

howardwu:
^This removes the outer circuit code and the outer circuit related types and helper methods from the `Network` trait.

howardwu:
^This merges in an FFT performance improvement

howardwu:
^These commits bring over desirable changes from the `testnet2` branch.

PR #598, opened 2022-01-17
Compute commitments with limited threads in high-thread scenarios

Pratyush:
## Motivation

When computing multiple MSMs, allocating an increasing number of threads to a single MSM results in diminishing returns. This PR allocates at most a fixed constant number of threads to a MSM, and instead computes multiple MSMs in parallel.

On a server machine with 24 cores, this results in a 25% speedup for PoSW proving.

## Test Plan

(Write your test plan here)

## Related PRs

(Link your related PRs here)


polycommit/src/sonic_pc/mod.rs
```diff
@@ -94,6 +94,80 @@ impl<E: PairingEngine> SonicKZG10<E> {
     }
 }
 
+macro_rules! execute_in_parallel {
```
Pratyush: This is the core macro that we rely on. Essentially, the idea is that, if we have, say, n MSMs, and m threads, then instead of devoting all m threads to a single MSM, and then computing the MSMs one-by-one, we instead devote a smaller number (say, 12) of threads to each MSM, and perform m/12 MSMs in parallel. 

This results in speed ups because increasing the number of threads per MSM provides diminishing returns, so it makes sense to cap the number of threads for any given MSM, and use the leftover threads for other MSMs. (Sounds like the Robin Hood story lol)

polycommit/src/sonic_pc/mod.rs
```diff
@@ -94,6 +94,80 @@ impl<E: PairingEngine> SonicKZG10<E> {
...
+    ($tasks:expr) => {{
+        #[cfg(feature = "parallel")]
+        {
+            use itertools::Itertools;
+            // At most 12 threads per task
+            const THREADS_PER_TASK: usize = 12;
```
Pratyush: This is a constant chosen empirically and without much theoretical backing; suggestions for a better division strategy are welcome.

Pratyush:
Oh and one other thing that I'm not sure about is how this interacts with the cuda feature; if it's enabled, then maybe we don't want this fancy parallelization after all.

Pratyush:
Commit [`6aab420` (#598)](https://github.com/AleoHQ/snarkVM/pull/598/commits/6aab420d1cab5f8169a3c9faa04c339058244b62) adds this speedup for Groth16 also.

polycommit/Cargo.toml
```diff
@@ -77,10 +77,19 @@ version = "0.9"
...
 [dependencies.rand_core]
 version = "0.6"
 default-features = false
 
+[dependencies.rand]
+version = "0.8"
+default-features = false
+features = [ "std_rng" ]
+
+
```
ljedrz:
```suggestion
```

utilities/src/parallel.rs
```diff
@@ -0,0 +1,209 @@
...
+    }
+
+    pub fn with_capacity(cap: usize) -> Self {
+        Self {
+            tasks: Vec::with_capacity(cap),
+        }
+    }
+
+    #[cfg(feature = "parallel")]
+    pub fn add_task<F: 'a + FnOnce() -> T + Send>(&mut self, f: F) {
```
ljedrz:
```suggestion
    pub fn add_job<F: 'a + FnOnce() -> T + Send>(&mut self, f: F) {
```
I'd change this name in order not to mix it up with asynchronous tasks.

utilities/src/parallel.rs
```diff
@@ -0,0 +1,209 @@
...
+    {
+        #[cfg(feature = "parallel")]
+        {
+            const THRESHOLD: usize = 12;
+            if max_available_threads() > THRESHOLD {
+                /// We have more threads than the threshold
+                use std::sync::{Arc, RwLock};
+                let mut task_results = Vec::with_capacity(self.tasks.len());
+                for _ in 0..self.tasks.len() {
+                    task_results.push(RwLock::new(None));
```
ljedrz: If we're not doing parallel reads, I'd change the `RwLock` to a `Mutex`

Pratyush: Done by moving to a `ParallelIterator`-based architecture (I was just poorly reimplementing that heh).

utilities/src/parallel.rs
Pratyush: Done, thanks!

utilities/src/parallel.rs
```diff
@@ -14,6 +14,87 @@
...
+        T: Send + Sync,
+    {
+        #[cfg(feature = "parallel")]
+        {
+            use rayon::prelude::*;
+            const THRESHOLD: usize = 12;
+            let num_threads = THRESHOLD.max(max_available_threads() / self.jobs.len());
+            self.jobs
+                .into_par_iter()
+                .map(|task| execute_with_threads(task, num_threads))
```
howardwu:
```suggestion
                .map(|job| execute_with_threads(job, num_threads))
```
minor nit

Pratyush: Fixed, thanks!

algorithms/src/fft/domain.rs
```diff
@@ -701,7 +683,7 @@ pub(crate) fn compute_powers<F: Field>(size: usize, g: F) -> Vec<F> {
         return compute_powers_serial(size, g);
     }
     // compute the number of threads we will be using.
-    let num_cpus_available = rayon::current_num_threads();
+    let num_cpus_available = max_available_threads();
```
Pratyush, line 686: This should improve performance when hyperthreading is enabled on Intel CPUs

utilities/src/parallel.rs
```diff
@@ -14,6 +14,86 @@
...
+    pub fn execute_all(self) -> Vec<T>
+    where
+        T: Send + Sync,
+    {
+        #[cfg(feature = "parallel")]
+        {
+            use rayon::prelude::*;
+            self.jobs
+                .into_par_iter()
+                .map(|job| execute_with_threads(job, max_available_threads()))
```
Pratyush, line 56: New logic for selecting the number of threads

howardwu (approved):

PR #600, opened 2022-01-18
Reduce memory overhead and FFT size in Marlin prover

Pratyush:
1. Sample the masking polynomial according to the distribution specified in [Lunar](https://eprint.iacr.org/2020/1069).
2. Reduce one FFT by splitting the computation of `z_a * z_b = (z_a' + r_a * v_H) * (z_b' * r_b * v_H)` as `z_a * z_b + r_a * r_b * v_H^2 + (r_a * z_b' + r_b * z_a') * v_H`. This is better as our FFT for the multiplication is smaller.
3. TODO: check if we can omit committing to the `t` polynomial
4. use execution pools for FFTs.

## Test Plan

(Write your test plan here)

## Related PRs

(Link your related PRs here)


algorithms/src/fft/polynomial/dense.rs
```diff
@@ -123,8 +123,7 @@ impl<F: Field> DensePolynomial<F> {
     }
 
     /// Perform a naive n^2 multiplication of `self` by `other`.
-    #[cfg(test)]
-    fn naive_mul(&self, other: &Self) -> Self {
+    pub fn naive_mul(&self, other: &Self) -> Self {
```
howardwu: Why was the `#[cfg(test)]` removed and the method made `pub`? Is there a case where we use it in prod now?

howardwu (approved):
LGTM

algorithms/src/fft/polynomial/dense.rs
Pratyush: Hm I used it at one point, but seem to have moved to a more efficient algorithm. I'll revert the change.

Pratyush:
```suggestion
	#[cfg(test)]
    fn naive_mul(&self, other: &Self) -> Self {
```

algorithms/src/fft/polynomial/dense.rs
```diff
@@ -123,7 +123,7 @@ impl<F: Field> DensePolynomial<F> {
     }
 
     /// Perform a naive n^2 multiplication of `self` by `other`.
-    #[cfg(test)]
+	#[cfg(test)]
```
Pratyush:
```suggestion
    #[cfg(test)]
```

PR #602, opened 2022-01-19
Feature/casting

gluax:
Adds casting to the IR, and a casting Gadget.

It's generic for future purposes in case we decide to add casting between more than just ints.

The number of commits was so that I could test with Leo. I can either trim them, or we can squash when merging to minimize them.


0rphon (approved):
IR/Eval code looks good

PR #603, opened 2022-01-19
Remove `t` polynomial from Marlin

Pratyush:
## Motivation

Improve Marlin proving speed and proof size by removing the commitment to the `t` polynomial. On my machine, this reduces PoSW time from 2.08s to 1.97s, so an improvement of roughly 5%. The proof size also reduces by 48 bytes (compressed). The security argument for this optimization is outlined in the Lunar paper.

Also introduces another change that switches our indexing from producing a single joint matrix to producing independent arithmetizations for each matrix. Furthermore, we reduce the size of our FFTs and witness polynomials by performing separate holographic linchecks for each matrix, and combining them via a new trick.

## Test Plan

(Write your test plan here)

## Related PRs

(Link your related PRs here)


marlin/src/ahp/prover/prover.rs
```diff
@@ -83,7 +81,7 @@ pub struct ProverSecondOracles<F: Field> {
 impl<F: Field> ProverSecondOracles<F> {
     /// Iterate over the polynomials output by the prover in the second round.
     pub fn iter(&self) -> impl Iterator<Item = &LabeledPolynomial<F>> {
-        vec![&self.t, &self.g_1, &self.h_1].into_iter()
+        vec![&self.g_1, &self.h_1].into_iter()
```
ljedrz: can we avoid the `vec!` here?

Pratyush: Yeah, now that we have `IntoIter` for arrays, we can avoid this. I'll push a fix.

howardwu (approved):

howardwu:
Note to self: resample parameters

Pratyush:
Note that test failures persist even after re-sampling parameters.

algorithms/src/fft/polynomial/dense.rs
```diff
@@ -256,6 +256,17 @@ impl<F: PrimeField> DensePolynomial<F> {
     }
 }
 
+impl<F: Field> From<super::SparsePolynomial<F>> for DensePolynomial<F> {
+    fn from(other: super::SparsePolynomial<F>) -> Self {
+        let mut result = vec![F::zero(); other.degree() + 1];
+        for (i, coeff) in other.coeffs {
+            result[i] = coeff;
+        }
```
ljedrz, line 264: if `other.coeffs` is a vector, this could be simplified as `result[..other.coeffs.len()].copy_from_slice(&other.coeffs)`

marlin/src/ahp/ahp.rs
```diff
@@ -111,39 +116,55 @@ impl<F: PrimeField, MM: MarlinMode> AHPForR1CS<F, MM> {
...
+        let non_zero_a_domain = EvaluationDomain::new(info.num_non_zero_a).unwrap();
+        let non_zero_b_domain = EvaluationDomain::new(info.num_non_zero_b).unwrap();
+        let non_zero_c_domain = EvaluationDomain::new(info.num_non_zero_c).unwrap();
+        Self::max_non_zero_domain_helper(non_zero_a_domain , non_zero_b_domain , non_zero_c_domain)
+    }
+    fn max_non_zero_domain_helper(domain_a: EvaluationDomain<F>, domain_b: EvaluationDomain<F>, domain_c: EvaluationDomain<F>) -> EvaluationDomain<F> {
+        let domains = [domain_a, domain_b, domain_c];
+        domains
+            .into_iter()
+            .max_by(|a, b| a.size().cmp(&b.size()))
```
ljedrz:
```suggestion
            .max_by_key(|domain| domain.size())
```

marlin/src/ahp/indexer/circuit.rs
```diff
@@ -59,12 +59,18 @@ impl<F: PrimeField, MM: MarlinMode> Circuit<F, MM> {
     /// Iterate over the indexed polynomials.
     pub fn iter(&self) -> impl Iterator<Item = &LabeledPolynomial<F>> {
         vec![
```
ljedrz: this `vec!` might be avoidable

marlin/src/ahp/matrices.rs
```diff
@@ -103,147 +99,99 @@ pub struct MatrixEvals<F: PrimeField> {
...
-    let c = c
-        .iter()
-        .enumerate()
-        .map(|(r, row)| row.iter().map(move |(f, i)| ((r, *i), *f)))
-        .flatten()
-        .collect::<BTreeMap<(usize, usize), F>>();
 
     let eq_poly_vals_time = start_timer!(|| "Precomputing eq_poly_vals");
-    let eq_poly_vals: BTreeMap<F, F> = output_domain
+    let eq_poly_vals: HashMap<F, F> = constraint_domain
```
ljedrz, line 146: just making sure that the order of elements in this collection can be random - `BTreeMap` enforces key ordering, while `HashMap` doesn't

ljedrz: this comment applies to all other places where a `BTreeMap` is changed to a `HashMap`

marlin/src/marlin/marlin.rs
```diff
@@ -489,11 +484,14 @@ impl<
...
+            &public_input,
+            &polynomials,
+            &prover_third_message,
+            &verifier_state,
+        )?;
 
-        if terminator.load(Ordering::Relaxed) {
-            return Err(MarlinError::Terminated);
-        }
+        Self::terminate(terminator)?;
```
ljedrz, line 407: optimization note: it seems to me that we create some objects well before they are used, by which point they might no longer be useful in case the `terminator` breaks the execution; an example is `commitment_randomnesses`, which is collected in line 478, and used only in line 533, with 2 calls to `Self::terminate(terminator)?;` in between.

marlin/src/marlin/marlin.rs
```diff
@@ -606,16 +628,19 @@ impl<
...
             }
 
             let mut new_input = vec![TargetField::one()];
             new_input.extend_from_slice(public_input);
-            new_input.resize(core::cmp::max(public_input.len(), domain_x.size()), TargetField::zero());
+            new_input.resize(
+                core::cmp::max(public_input.len(), input_domain.size()),
+                TargetField::zero(),
+            );
             assert!(new_input.first().unwrap().is_one());
```
ljedrz: this assertion doesn't seem useful - `new_input[0]` is set to `TargetField::one()` right above, and it doesn't get mutated in between

ljedrz (commented):
Left a few more comments and suggestions.

algorithms/src/fft/polynomial/dense.rs
Pratyush: Unfortunately `other.coeffs` is a `Vec<(usize, F)>`, but `result` needs to be `Vec<F>`, so we can't directly copy the contents. In practice, the loops should execute at most 10 times, so it's not a perf concern.

marlin/Cargo.toml
```diff
@@ -22,7 +22,7 @@ keywords = [
 categories = [ "cryptography::cryptocurrencies", "operating-systems" ]
 include = [ "Cargo.toml", "src", "README.md", "LICENSE" ]
 license = "GPL-3.0"
-edition = "2018"
+edition = "2021"
```
Pratyush, line 25: Upgrading to edition 2021 here.

marlin/src/ahp/indexer/circuit.rs
Pratyush:
```suggestion
        [
```

marlin/src/ahp/matrices.rs
Pratyush: Right, I don't think we're iterating over the contents of the map, only using it as a key-value store, so I think we can use the `HashMap` without issue.

marlin/src/ahp/verifier/first_message.rs
```diff
@@ -19,8 +19,6 @@
 pub struct VerifierFirstMessage<F> {
     /// Query for the random polynomial.
     pub alpha: F,
-    /// Randomizer for the lincheck for `A`.
-    pub eta_a: F,
```
Pratyush, line 23: Instead of computing `eta_a * A + eta_b * B + eta_c * C`, we can just do `A + eta_b * B + eta_c * C`; this is still linearly independent.

marlin/src/constraints/proof.rs
```diff
@@ -167,10 +167,9 @@ where
 
         let mut evaluation_gadgets = HashMap::<String, NonNativeFieldVar<TargetField, BaseField>>::new();
 
-        const ALL_POLYNOMIALS: [&str; 7] = [
+        const ALL_POLYNOMIALS: [&str; 6] = [
```
Pratyush: TODO: fix this, since we no longer have `g_2`

marlin/src/constraints/proof.rs
```diff
@@ -247,10 +246,9 @@ where
...
```
Pratyush: Ditto

marlin/src/marlin/marlin.rs
```diff
@@ -260,6 +260,14 @@ impl<
         Self::prove_with_terminator(circuit_proving_key, circuit, &AtomicBool::new(false), zk_rng)
     }
 
+    fn terminate(terminator: &AtomicBool) -> Result<(), MarlinError> {
```
Pratyush, line 213: Helper function to reduce early-termination boilerplate.

marlin/src/marlin/marlin.rs
Pratyush: Thanks, will keep in mind! I'm guessing the logic was to colocate similar pieces of code, however if it's expensive to construct these objects then I'll move the construction logic closer to the use site.

marlin/src/marlin/marlin.rs
Pratyush: The `resize` could truncate it to zero, right?

ljedrz: not here, because `new_input` gets resized to at least `public_input.len() + 1` (it starts with a single element and is extended with `public_input`)

marlin/src/ahp/prover/prover.rs
```diff
@@ -83,22 +82,38 @@ pub struct ProverSecondOracles<F: Field> {
...
 
 impl<F: Field> ProverThirdOracles<F> {
     /// Iterate over the polynomials output by the prover in the third round.
     pub fn iter(&self) -> impl Iterator<Item = &LabeledPolynomial<F>> {
-        vec![&self.g_2, &self.h_2].into_iter()
+        [&self.g_a, &self.g_b, &self.g_c].into_iter()
+    }
+}
+
+#[derive(Debug)]
```
Pratyush, line 107: Our protocol is now 4 rounds, instead of just 3.

marlin/src/marlin/marlin.rs
```diff
@@ -114,32 +110,9 @@ impl<
         let (committer_key, verifier_key) =
             PC::trim(&srs, circuit.max_degree(), supported_hiding_bound, Some(&coeff_support))?;
 
-        let mut vanishing_polys = vec![];
-        if MM::RECURSION {
```
Pratyush, line 118: We no longer support recursion in the default protocol. This means that the index does not include commitments to vanishing polynomials.

This change was made because recursive provers can always compute the appropriate vanishing polynomials.

marlin/src/marlin/marlin.rs
```diff
@@ -512,54 +423,36 @@ impl<
...
-            fs_rng.absorb_nonnative_field_elements(&evaluations, OptimizationType::Weight);
-        } else {
-            fs_rng.absorb_bytes(&to_bytes_le![&evaluations].unwrap());
-        }
+        fs_rng.absorb_nonnative_field_elements(&evaluations, OptimizationType::Weight);
 
-        let pc_proof = if MM::RECURSION {
-            let num_open_challenges: usize = 5;
+        let num_open_challenges: usize = 7;
```
Pratyush, line 430: We use individual opening challenges by default, since that's more efficient for our verifiers.

howardwu (approved):

marlin/src/constraints/ahp.rs
```diff
@@ -802,13 +783,13 @@ impl<
         }
 
         let h_minus_2 = index_pvk
-            .domain_h_size_gadget
+            .constraint_domain_size_gadget
             .clone()
-            .sub_constant(cs.ns(|| "domain_h_minus_2"), &BaseField::from(2u128))?;
+            .sub_constant(cs.ns(|| "constraint_domain_minus_2"), &BaseField::from(2u128))?;
 
         // 3 comms for beta from the round 2
```
howardwu:
```suggestion
        // 2 comms for beta from the round 2
```

howardwu:
Prior to merging #630, the benchmark of this branch is as follows:
```
Proof of Succinct Work: Marlin/mine                                                                          
                        time:   [1.9545 s 1.9892 s 2.0299 s]

Proof of Succinct Work: Marlin/verify                                                                           
                        time:   [50.427 ms 50.501 ms 50.624 ms]
```

howardwu:
After merging #630, the benchmark of this branch is as follows:
```
Proof of Succinct Work: Marlin/mine                                                                          
                        time:   [1.5326 s 1.5688 s 1.6051 s]

Proof of Succinct Work: Marlin/verify                                                                           
                        time:   [29.577 ms 29.661 ms 29.777 ms]
```

PR #609, opened 2022-01-21
A few deserialization improvements

ljedrz:
This PR does 3 things:
- removes some deserialization-related allocations (`Ciphertext`, `Payload`)
- returns an error when `CanonicalDeserialize::deserialize` tries to allocate more than the OS can provide
- bumps the CircleCI Rust images to 1.58 (the highest currently available stable version) in order to be able to use `Vec::try_reserve`

Supersedes https://github.com/AleoHQ/snarkVM/pull/592.

### DIFF
--- a/dpc/src/record/ciphertext.rs
+++ b/dpc/src/record/ciphertext.rs
@@ -16,14 +16,7 @@
 
 use crate::{Bech32Locator, DecryptionKey, Network, RecordError, ViewKey};
 use snarkvm_algorithms::traits::{EncryptionScheme, CRH};
-use snarkvm_utilities::{
-    io::{Cursor, Result as IoResult},
-    to_bytes_le,
-    FromBytes,
-    Read,
-    ToBytes,
-    Write,
-};
+use snarkvm_utilities::{io::Result as IoResult, to_bytes_le, FromBytes, Read, ToBytes, Write};
 
 use anyhow::{anyhow, Result};
 
@@ -133,13 +126,9 @@ impl<N: Network> FromBytes for Ciphertext<N> {
     /// Decode the ciphertext into the ciphertext randomizer, record view key commitment, and record ciphertext.
     #[inline]
     fn read_le<R: Read>(mut reader: R) -> IoResult<Self> {
-        let mut ciphertext = vec![0u8; N::RECORD_CIPHERTEXT_SIZE_IN_BYTES];
-        reader.read_exact(&mut ciphertext)?;
-
         // Decode the ciphertext bytes.
-        let mut cursor = Cursor::new(ciphertext);
-        let ciphertext_randomizer = N::RecordRandomizer::read_le(&mut cursor)?;
-        let record_view_key_commitment = N::RecordViewKeyCommitment::read_le(&mut cursor)?;
+        let ciphertext_randomizer = N::RecordRandomizer::read_le(&mut reader)?;
+        let record_view_key_commitment = N::RecordViewKeyCommitment::read_le(&mut reader)?;
 
         let mut record_bytes = vec![
             0u8;
@@ -147,7 +136,7 @@ impl<N: Network> FromBytes for Ciphertext<N> {
                 - N::RecordRandomizer::data_size_in_bytes()
                 - N::RecordViewKeyCommitment::data_size_in_bytes()
         ];
-        cursor.read_exact(&mut record_bytes)?;
+        reader.read_exact(&mut record_bytes)?;
 
         Ok(Self::from(
             ciphertext_randomizer,
--- a/dpc/src/record/payload.rs
+++ b/dpc/src/record/payload.rs
@@ -40,9 +40,7 @@ impl<N: Network> Payload<N> {
     }
 
     pub fn is_empty(&self) -> bool {
-        let mut payload = vec![0u8; N::RECORD_PAYLOAD_SIZE_IN_BYTES];
-        payload.copy_from_slice(&self.0);
-        payload == vec![0u8; N::RECORD_PAYLOAD_SIZE_IN_BYTES]
+        self.0.len() == N::RECORD_PAYLOAD_SIZE_IN_BYTES && self.0.iter().all(|byte| *byte == 0)
     }
 
     pub fn size() -> usize {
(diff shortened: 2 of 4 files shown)


dpc/src/record/payload.rs
```diff
@@ -107,7 +106,7 @@ impl<'de, N: Network> Deserialize<'de> for Payload<N> {
 
 impl<N: Network> Default for Payload<N> {
     fn default() -> Self {
-        Self::from(&[])
+        Self::from(vec![])
```
Pratyush: Hm should we just preallocate this with `N::RECORD_PAYLOAD_SIZE_IN_BYTES` empty bytes?

Pratyush:
Another strategy could be to add a size const generic parameter to `Payload` (since the size is statically known). This would completely avoid any allocations.

dpc/src/record/payload.rs
ljedrz: Depends on how the `Default` cases are handled - if they are populated with bytes "wholesale", it won't make a difference. I picked an empty `Vec`, as it doesn't allocate at all; in other words, while it's possible there's room for further improvement, this change doesn't have downsides compared to the current state.

ljedrz:
>Another strategy could be to add a size const generic parameter to Payload (since the size is statically known). This would completely avoid any allocations.

Depending on the number of `Payload`s and the size of `N::RECORD_PAYLOAD_SIZE_IN_BYTES`, this could be a strain on the stack (risking a stack overflow) - it's something to consider, but if there can be many of them in the memory at once, a vector is safer.

dpc/src/record/payload.rs
```diff
@@ -126,7 +125,9 @@ mod tests {
         // Create a random byte array, construct a payload from it, and check its byte array matches.
         for i in 0..Testnet2::RECORD_PAYLOAD_SIZE_IN_BYTES {
             let expected_payload = (0..i).map(|_| u8::rand(rng)).collect::<Vec<u8>>();
-            let candidate_payload = Payload::<Testnet2>::from(&expected_payload).to_bytes_le().unwrap();
+            let candidate_payload = Payload::<Testnet2>::from(expected_payload.clone())
+                .to_bytes_le()
+                .unwrap();
```
niklaslong, line 130: What are the guarantees around this call?

ljedrz: While I would usually say that I'm just preserving the existing logic, this time I'll note that this is in `mod tests`, so there don't need to be any :wink:.

niklaslong: Ah, missed the context here 👍

howardwu:
Could someone benchmark the impact of the CanonicalDeserialize change here? It looks good to me, just want to make sure it doesn't introduce new implications. That deserialization in general accumulates a significant runtime cost

ljedrz:
@howardwu there should be no change, but I can double-check tomorrow :+1:.

dpc/src/record/ciphertext.rs
```diff
@@ -133,21 +126,17 @@ impl<N: Network> FromBytes for Ciphertext<N> {
...
+        let record_view_key_commitment = N::RecordViewKeyCommitment::read_le(&mut reader)?;
 
         let mut record_bytes = vec![
             0u8;
             N::RECORD_CIPHERTEXT_SIZE_IN_BYTES
                 - N::RecordRandomizer::data_size_in_bytes()
                 - N::RecordViewKeyCommitment::data_size_in_bytes()
         ];
-        cursor.read_exact(&mut record_bytes)?;
+        reader.read_exact(&mut record_bytes)?;
```
howardwu, line 140: @ljedrz I like that the new logic is more concise. Can you help me build confidence that this ensures the maximal size read is no more than `N::RECORD_CIPHERTEXT_SIZE_IN_BYTES`?

howardwu: The concern here is whether we can accept malformed ciphertexts (by misreading some critical parameters that ultimately do introduce fund vulnerability)

dpc/src/record/payload.rs
howardwu: @Pratyush that's a great observation. It technically should be preallocated to adhere to the design assumptions, and I originally missed this.

howardwu:
```suggestion
        Self::from(vec![0u8; N::RECORD_CIPHERTEXT_SIZE_IN_BYTES])
```

ljedrz:
@howardwu as for the `CanonicalDeserialize` change, I checked the `dpc` benchmarks and saw no difference, as expected. The reason why that deserialization is taking a long time is likely just due to the crypto involved in the deserialization of the individual elements of the related vector.

dpc/src/record/payload.rs
ljedrz: this change appears to break some tests; I can either revert it to the current status quo, or adjust the tests - your call @howardwu.

dpc/src/record/ciphertext.rs
ljedrz: A limited size (`N::RECORD_CIPHERTEXT_SIZE_IN_BYTES - N::RecordRandomizer::data_size_in_bytes() - N::RecordViewKeyCommitment::data_size_in_bytes()`) is guaranteed for `record_bytes`, so I'm assuming you're referring to `ciphertext_randomizer` and `record_view_key_commitment` here; this should ideally be constrained in the `FromBytes` impl for `AleoLocator` (the concrete type for `N::RecordRandomizer` and `N::RecordViewKeyCommitment`), but if you want them to additionally be limited here, I can apply [Read::take](https://doc.rust-lang.org/std/io/trait.Read.html#method.take) limiting the reads for `ciphertext_randomizer` and `record_view_key_commitment` to `N::RecordRandomizer::data_size_in_bytes()` and `N::RecordViewKeyCommitment::data_size_in_bytes()` respectively.

Which do you prefer?

howardwu: The `Read::take` would be fine. If there is no performance difference, the old route is likely more clear (it requires less skill for the reader to deduce correctness of protocol)

ljedrz:
Superseded by https://github.com/AleoHQ/snarkVM/pull/735/files.

PR #612, opened 2022-01-21
Merkle tree perf tweaks

ljedrz:
I had a feeling that hashing done on each level of the Merkle tree could be done in parallel, and it seems that it does work and brings a nice performance boost - with this change, my snarkOS ledger loads in 33s instead of 208s, which is a **84%** decrease.

### DIFF
--- a/algorithms/src/merkle_tree/merkle_tree.rs
+++ b/algorithms/src/merkle_tree/merkle_tree.rs
@@ -168,8 +168,7 @@ impl<P: MerkleParameters + Send + Sync> MerkleTree<P> {
 
         // Compute the hash values for every node in the tree.
         let mut upper_bound = last_level_index;
-        level_indices.reverse();
-        for &start_index in &level_indices {
+        for start_index in level_indices.into_iter().rev() {
             // Iterate over the current level.
             for current_index in start_index..upper_bound {
                 let left_index = left_child(current_index);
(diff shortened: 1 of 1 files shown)


ljedrz:
Note: this will also improve the performance of any other operation which requires the Merkle tree to be rebuilt - the node launch is just the most visible improvement.

algorithms/src/merkle_tree/merkle_tree.rs
```diff
@@ -166,30 +166,44 @@ impl<P: MerkleParameters + Send + Sync> MerkleTree<P> {
                 .copy_from_slice(&subsection[..]);
         }
 
+        // Prepare a copy of the tree to allow each level to be hashed in parallel.
+        let mut tree_clone = tree.clone();
```
Pratyush: From what I understand, we need to do this `clone()` because we need immutable access to the child nodes, but at the same time we need mutable access to the parent node. Can we avoid the clone by doing a `split_mut` at `upper_bound`? This would give immutable access to the children, but mutable access to the parents.

Pratyush:
This is a fantastic improvement! It LGTM modulo the question about using `split_mut`.

algorithms/src/merkle_tree/merkle_tree.rs
ljedrz: I agree that this `clone` could possibly be improved upon; I'll try to figure something out :microscope:.

ljedrz:
Earlier on I felt a bit tired already (hence the simpler solution), but I felt evening inspiration and actually managed to pull it off :sunglasses:.

As expected, it's now even faster, and even with around 10% extra blocks (I needed to sync a bit), the ledger load time actually went down to **29s**.

Pratyush (approved):

Pratyush:
Thanks, this is awesome!

niklaslong (approved):

raychu86 (approved):

howardwu (approved):

PR #618, opened 2022-01-25
MSM tweaks

ljedrz:
I've been looking at the MSM impls, and have the following proposals for improvements:
- parallelize `FixedBaseMSM::windowed_mul`
- refactor `VariableBaseMSM::msm_standard` into 3 functions for greater readability; I initially did it hoping to be able to find performance improvements, but since there wasn't any room for them in the end, the last commit can be dropped if the original shape is more desirable

### DIFF
--- a/algorithms/src/msm/variable_base/standard.rs
+++ b/algorithms/src/msm/variable_base/standard.rs
@@ -21,6 +21,67 @@ use snarkvm_utilities::BigInteger;
 #[cfg(feature = "parallel")]
 use rayon::prelude::*;
 
+fn update_buckets<G: AffineCurve>(
+    base: &G,
+    mut scalar: <G::ScalarField as PrimeField>::BigInteger,
+    w_start: usize,
+    c: usize,
+    buckets: &mut [G::Projective],
+) {
+    // We right-shift by w_start, thus getting rid of the
+    // lower bits.
+    scalar.divn(w_start as u32);
+
+    // We mod the remaining bits by the window size.
+    let scalar = scalar.as_ref()[0] % (1 << c);
+
+    // If the scalar is non-zero, we update the corresponding
+    // bucket.
+    // (Recall that `buckets` doesn't have a zero bucket.)
+    if scalar != 0 {
+        buckets[(scalar - 1) as usize].add_assign_mixed(base);
+    }
+}
+
+fn process_window<G: AffineCurve>(
+    bases: &[G],
+    scalars: &[<G::ScalarField as PrimeField>::BigInteger],
+    w_start: usize,
+    c: usize,
+) -> G::Projective {
+    let mut res = G::Projective::zero();
+    let fr_one = G::ScalarField::one().to_repr();
+
+    // We only process unit scalars once in the first window.
+    if w_start == 0 {
+        scalars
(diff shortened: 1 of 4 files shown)


Pratyush:
This looks good!

For the fixed-base MSM, we can also parallelize window construction like here: https://github.com/arkworks-rs/algebra/blob/master/ec/src/msm/fixed_base.rs#L38

Regarding the perf improvement, maybe we should add an inline annotation to the helper methods in variable-base MSM? not sure if that would help or not.

ljedrz:
> maybe we should add an inline annotation to the helper methods in variable-base MSM? not sure if that would help or not

I tried it for `update_buckets`, but it didn't improve the associated benchmark; the other functions are rather large and force-inlining would probably not be very practical.

ljedrz:
> For the fixed-base MSM, we can also parallelize window construction like here: https://github.com/arkworks-rs/algebra/blob/master/ec/src/msm/fixed_base.rs#L38

I can check it out tomorrow :ok_hand:.

ljedrz:
@Pratyush done :+1:.

ljedrz:
This PR's extra parallelization in `FixedBaseMSM` has the following effect on the one related benchmark that I found:
```
snark_setup             time:   [12.182 ms 12.281 ms 12.412 ms]                       
                        change: [-40.292% -39.716% -38.969%] (p = 0.00 < 0.05)
                        Performance has improved.
Found 2 outliers among 50 measurements (4.00%)
  1 (2.00%) high mild
  1 (2.00%) high severe
```

curves/src/traits/group.rs
```diff
@@ -47,6 +48,7 @@ pub trait Group:
     + AddAssign<Self>
     + SubAssign<Self>
     + MulAssign<Self::ScalarField>
+    + Sum
```
Pratyush: Hm I think we should add this trait bound to the `ProjectiveCurve` trait, instead of here. The `Group` trait should probably be deprecated and removed.

ljedrz: You got it :ok_hand:.

Pratyush:
Thanks, LGTM!

Pratyush (approved):

algorithms/src/msm/variable_base/standard.rs
```diff
@@ -21,6 +21,67 @@ use snarkvm_utilities::BigInteger;
...
+    }
+
+    // We don't need the "zero" bucket, so we only have 2^c - 1 buckets
+    let mut buckets = vec![G::Projective::zero(); (1 << c) - 1];
+    scalars
+        .iter()
+        .zip(bases)
+        .filter(|(&s, _)| s > fr_one)
+        .for_each(|(&scalar, base)| update_buckets(base, scalar, w_start, c, &mut buckets));
+    // G::Projective::batch_normalization(&mut buckets);
```
niklaslong, line 73: Is this still needed?

ljedrz:
The CI failures were caused by a [CircleCI outage](https://status.circleci.com/incidents/287jc8v68v9v):

image

I'll re-run those jobs once it's resolved.

howardwu (approved):
LGTM

PR #620, opened 2022-01-26
Improve BitIterator

yujqiao:
## Motivation

`BitIterator` is used a lot in the calculation. However, each iteration will perform dividing and multiplying, which is an expensive operation.

## Test Plan

Tests already exist

## Related PRs
None


yujqiao:
Unfortunately, a primitive benchmark showed no measurable improvement.
And just realized that the divisor is 64 and can be efficiently calculated.
Close now.

PR #622, opened 2022-01-27
Improve BitIterator

yujqiao:
## Motivation

`BitIterator` is used a lot in the calculation. However, each iteration will perform multiplying and  subtraction to do modulo.

This PR eliminates the usage of multiplication via bit-wise AND operation. It's a micro optimization, though.

`(n % 64) == (n & 63)`

## Test Plan

Tests already exist.

## Related PRs

None

### DIFF
--- a/utilities/src/bititerator.rs
+++ b/utilities/src/bititerator.rs
@@ -48,7 +48,7 @@ impl<Slice: AsRef<[u64]>> Iterator for BitIteratorBE<Slice> {
         } else {
             self.n -= 1;
             let part = self.n / 64;
-            let bit = self.n - (64 * part);
+            let bit = self.n & 63;
 
             Some(self.s.as_ref()[part] & (1 << bit) > 0)
         }
@@ -94,7 +94,7 @@ impl<Slice: AsRef<[u64]>> Iterator for BitIteratorLE<Slice> {
             None
         } else {
             let part = self.n / 64;
-            let bit = self.n - (64 * part);
+            let bit = self.n & 63;
             self.n += 1;
 
             Some(self.s.as_ref()[part] & (1 << bit) > 0)


ljedrz:
Hi @rapiz1, and thanks for your contribution.

Is there a benchmark that confirms whether this results in a measurable improvement? I'm wondering if the compiler isn't already optimizing this operation this way automatically :thinking:.

yujqiao:
> Is there a benchmark that confirms whether this results in a measurable improvement?

I've made a little benchmark and run it on my machine. Unfortunately it's not a measurable on my machine. But I think it's indeed fast because I've checked out the assembly code.

> I'm wondering if the compiler isn't already optimizing this operation this way automatically thinking.

I'm pretty sure that the compiler doesn't optimize for this. I use godbolt to see what this snippet compiles to. [Here's the snippet with the disassembly result.](https://rust.godbolt.org/z/Tsd9d643q)

After the optimization:
```
        mov     rax, qword ptr [rdi + 8]
        and     rax, 64
        mov     qword ptr [rsp + 24], rax
```

Before the optimization, the corresponding assembly code is long and includes a multiplication and a subtraction. You can check out it using the link I provide above.

I think it's clear that the machine code after the optimization is more efficient, judging from the disassembly output and the fact that `and` is way more fast than `mul`.

ljedrz:
Note that in your godbolt snippet you didn't compile with optimizations, like you would via `cargo build --release`; I checked both variants with the appropriate flag and saw no difference in the resulting ASM: [godbolt](https://rust.godbolt.org/z/fj66qr4W7).

yujqiao:
@ljedrz You're right! I forgot to add the flag for godbolt 😱. Close now

PR #624, opened 2022-01-27
Algorithms: refactor the `blake2xs` module 

niklaslong:
This PR removes the dead code from the `blake2xs` module. This includes the now unnecessary parameter-block helpers and some obsolete tests. In addition, `blake2s_simd` was turned into an optional dependency and `blake2` was removed from the `algorithms` crate entirely. 

After the bump to `v0.10`, `blake2` remains a dependency (`Blake2s256`) on `FiatShamirChaChaRng` and is also bounded by `Digest` which the `simd` implementation is not.

Obviates #574 and #527.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -160,13 +160,11 @@ checksum = "bef38d45163c2f1dde094a7dfd33ccf595c92905c8f8f4fdc18d06fb1037718a"
 
 [[package]]
 name = "blake2"
-version = "0.9.2"
+version = "0.10.2"
 source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "0a4e37d16930f5459780f5621038b6382b9bb37c19016f39fb6b5808d831f174"
+checksum = "b94ba84325db59637ffc528bbe8c7f86c02c57cff5c0e2b9b00f9a851f42f309"
 dependencies = [
- "crypto-mac",
- "digest 0.9.0",
- "opaque-debug 0.3.0",
+ "digest 0.10.0",
 ]
 
 [[package]]
@@ -484,16 +482,6 @@ dependencies = [
 ]
 
 [[package]]
-name = "crypto-mac"
-version = "0.8.0"
-source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "b584a330336237c1eecd3e94266efb216c56ed91225d634cb2991c5f3fd1aeab"
-dependencies = [
- "generic-array 0.14.4",
- "subtle",
-]
-
-[[package]]
 name = "csv"
 version = "1.1.6"
 source = "registry+https://github.com/rust-lang/crates.io-index"
@@ -585,15 +573,6 @@ dependencies = [
 
 [[package]]
 name = "digest"
-version = "0.9.0"
-source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "d3dd60d1080a57a05ab032377049e0591415d2b31afd7028356dbf3cc6dcb066"
-dependencies = [
- "generic-array 0.14.4",
-]
-
-[[package]]
-name = "digest"
 version = "0.10.0"
 source = "registry+https://github.com/rust-lang/crates.io-index"
 checksum = "8549e6bfdecd113b7e221fe60b433087f6957387a20f8118ebca9b12af19143d"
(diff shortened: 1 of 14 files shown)


niklaslong:
Current failures are due to Circle CI's degraded service.

Cargo.lock
```diff
@@ -1929,10 +1937,10 @@ name = "snarkvm-circuits"
 version = "0.7.5"
 dependencies = [
  "anyhow",
- "blake2",
+ "blake2 0.9.2",
```
ljedrz: `snarkvm-circuits` and `snarkvm-polycommit` seem to still be using the older version of `blake2`

niklaslong: Good catch, it isn't being used in code so it can be removed from the deps 👍

ljedrz (approved):
LGTM with the aforementioned comment and a green CI.

edit: nice diff!

marlin/Cargo.toml
```diff
@@ -73,15 +73,15 @@ default-features = false
...
+version = "0.10"
 default-features = false
 
 [dependencies.derivative]
 version = "2"
 features = [ "use_core" ]
 
 [dependencies.digest]
-version = "0.9"
+version = "0.10"
```
howardwu: Can you move `blake2` to a `dev-dep` (and check if `digest` can be moved down as well)?

gadgets/Cargo.toml
```diff
@@ -54,7 +54,7 @@ version = "1.0.53"
 version = "2"
 
 [dependencies.digest]
-version = "0.9"
+version = "0.10"
```
howardwu: Can you check if `digest` can be a `dev-dep`?

algorithms/src/crypto_hash/blake2xs.rs
```diff
@@ -216,119 +129,4 @@ mod tests {
...
-            
-            // Sample a new input.
-            let input: [u8; 32] = rng.gen();
-            
-            // Compare the evaluation of the implementations.
-            assert_eq!(evaluate_blake2s(reference_a, input), evaluate_varblake2s(reference_b.clone(), input));
-            assert_eq!(evaluate_varblake2s(reference_b, input), evaluate_varblake2s(reference_c.clone(), input));
-            assert_eq!(evaluate_varblake2s(reference_c, input), evaluate_varblake2s(candidate, input));
-        }
-    }
```
howardwu: Keep the tests and implementation. They are used to check the test vectors match any updates to blake2s.

howardwu: If there is a mismatch, this allows us to debug what parameter changed

niklaslong: I'll reintroduce the `blake2s` impl and test just to make sure those vectors is still correct. Note that this isn't the implementation that's currently used elsewhere in the codebase, which is why I removed it. We use the default parameters (no `persona`) and it is a plain `blake2::Blake2s256::digest` (unless we want to unify that under our own wrapper implementation?). 

Regarding the parameter block test: the `blake2` crate no longer supports specifying the parameter block, this was a public but hidden part of the API in `v0.9` but has been removed in `v0.10`. The `blake2s_simd` implementation that we use for `Blake2Xs` also doesn't support a parameter block as input, the `Params` constructor must be used instead. I may have missed something but it seems that there's no parameter block left to test other than what is abstracted by the implementation itself and that's the dependency's concern?

marlin/Cargo.toml
niklaslong: `digest` still needs to be a `dep` as it is used in the `FiatShamirChaChaRng` impl.

gadgets/Cargo.toml
niklaslong: Done 👍

algorithms/src/crypto_hash/blake2xs.rs
howardwu: Thanks for investigating. That should be alright then. Thanks for adding the test vector sanity checks.

howardwu:
@niklaslong it looks like migrating this PR to `testnet3` is infeasible. Could you try applying a `git patch` on the `testnet3` branch with you changes here as is?

niklaslong:
Replaced by #634.

PR #636, opened 2022-01-31
Introduce a prototype program registry

raychu86:
## Motivation

This PR introduces a prototype program registry where programs must be deployed before use. In it's current form, the deployed program is stored in the `Execution` struct. 

How deployment transactions are constructed:
1. Create a program deployment `Request` that includes the program and it's corresponding function vks.
2. The Ledger checks that the program doesn't already exist in the registry.
3. Program is added to the registry when the block is mined. (Also included in the `Execution` struct of the deployment transaction).


Pratyush:
There is no check for whether the vk associated with a program is valid, right?

raychu86:
Not at the moment, I have a TODO for that check. Was thinking to hold off on it, since we'll be storing the IRs down the line anyways.

PR #640, opened 2022-02-02
Refactor marlin and polycommit into algorithms, adds back nonnative gadgets

howardwu:
## Motivation

Refactor marlin and polycommit into algorithms, adds back nonnative gadgets


algorithms/src/snark/marlin/mod.rs
```diff
@@ -26,14 +42,28 @@ pub use circuit_verifying_key::*;
...
 
+/// Parameters of non-native field gadget
+///
+/// Sample parameters for non-native field gadgets
+/// - `BaseField`:              the constraint field
+/// - `TargetField`:            the field being simulated
+/// - `num_limbs`:              how many limbs are used
+/// - `bits_per_limb`:          the size of the limbs
+///
+pub mod params;
```
Pratyush, line 65: I guess this can move back to the nonnative module?

howardwu: Open to suggestions how to handle. This was the tricky thing I mentioned. The Optimization type and params are used in Marlin it seems, so this would require snarkvm-algorithms to import snarkvm-gadgets which will slow down compile times

Pratyush: Maybe we move it to snarkvm_utilities or something?

howardwu: (Also feel free to push directly if you have an idea in mind)

howardwu: Maybe into snarkvm-r1cs?

Pratyush: that would also work. I guess also fine to leave it here; at some point soon we'll move to lookup-based non-native anyway =)

Pratyush (approved):

PR #645, opened 2022-02-07
Improve the quality of randomness in tests

ljedrz:
Instead of using hard-coded seeds, obtain them once using the `getrandom` feature and then seed the fast `XorShiftRandom` with them in order to have more random inputs in the related tests.
Note: this change is not possible everywhere, as some of the tests require a `CryptoRng`; in those cases, `StdRng` is used instead, and it's also seeded from the entropy provided by the OS.

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -159,7 +159,7 @@ jobs:
   gadgets:
     docker:
       - image: cimg/rust:1.58
-    resource_class: xlarge
+    resource_class: 2xlarge
     steps:
       - checkout
       - setup_environment:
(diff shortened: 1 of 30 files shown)


ljedrz:
This change already uncovered one issue:
```
failures:

---- integers::uint::tests::uint16::test_uint16_div_constants stdout ----
thread 'integers::uint::tests::uint16::test_uint16_div_constants' panicked at 'attempt to divide by zero', /rustc/db9d1b20bba1968c1ec1fc49616d4742c1725b4b/library/core/src/num/mod.rs:816:5


failures:
    integers::uint::tests::uint16::test_uint16_div_constants
```
I don't know if the related test should be adjusted (which would be trivial - `0` would just need to be excluded from the range for the divisor), or if we should merge this as-is, and file an issue for panicking division by zero.

niklaslong (approved):

howardwu (approved):

ljedrz:
Note: it seems that `snarkvm-gadgets` might need the 2xlarge resource class, otherwise it's flaky, especially with the higher-bit `pow` tests; I brought it back there in the second commit.

howardwu (approved):

PR #644, opened 2022-02-03
Remove Noop programs and record payload requirements

raychu86:
## Motivation

**This PR implements a few key features for testnet3:**
 - The ProgramID CRH curve updated from BW6 to EdwardsBls12
 - Record payloads are now optional
 - Program IDs are now optional in a transition.
     - Executions now do not need a program id, program path, program vk, and program proof for transfers/noop transitions.
     - This vastly reduces the size of transactions.
     - Actual applications outside of transfers will require real program executions will corresponding proofs.
 - Usage/reliance of the noop program has been removed.
    - The Noop program is no longer embedded into the Network trait
    - Noop parameters parameters are no longer regenerated and stored
 - Inner circuit updated to reflect the optional executions and optional record payloads.
    
**Additional updates:** 
 - Parameter have been regenerated
 - Updated constants (record size, record ciphertext size, program id size, etc)
 - Updated serialization size tests

PR #646, opened 2022-02-10
Record ciphertext attribute specific encryption

raychu86:
## Motivation

This PR updates the encryption scheme message type from [u8] to [[u8]] to to support attribute specific encryption in the record ciphertext rather than the original approach of bit packing. This allows us to easily add/remove the optional record payload when constructing records in transactions. 

The ciphertext commitment will be padding the "missing" bytes of the ciphertext if the payload is omitted. This is done so that the inner circuit can reason about the record ciphertexts properly.

This PR is built on #644


algorithms/src/encryption/ecies_poseidon.rs
```diff
@@ -226,121 +226,141 @@ where
     ///     ciphertext := to_bytes_le![C_1, ..., C_n], where C_i := R_i + M_i, and R_i := H_i(G^ar)
     /// ```
     ///
-    fn encrypt(&self, symmetric_key: &Self::SymmetricKey, message: &[u8]) -> Result<Vec<u8>, EncryptionError> {
+    fn encrypt(
+        &self,
+        symmetric_key: &Self::SymmetricKey,
+        message: &[Vec<u8>],
```
ljedrz: would it be possible to change `&[Vec<u8>]` to `&[&[u8]]`? we don't really need those elements to be owned vectors; otherwise, something like `&[T]` where `T: AsRef<[u8]>` would work too.

algorithms/src/encryption/ecies_poseidon.rs
```diff
@@ -226,121 +226,141 @@ where
...
     }
 
     ///
     /// Decrypts the given ciphertext with the given symmetric key.
     ///
-    fn decrypt(&self, symmetric_key: &Self::SymmetricKey, ciphertext: &[u8]) -> Result<Vec<u8>, EncryptionError> {
+    fn decrypt(
+        &self,
+        symmetric_key: &Self::SymmetricKey,
+        ciphertext: &[Vec<u8>],
```
ljedrz: ditto

dpc/src/record/ciphertext.rs
```diff
@@ -48,17 +41,20 @@ impl<N: Network> Ciphertext<N> {
...
-        record_bytes: Vec<u8>,
+        record_bytes: Vec<Vec<u8>>,
         program_id: Option<N::ProgramID>,
     ) -> Result<Self, RecordError> {
         let program_id_bytes = match program_id {
             Some(program_id) => program_id.to_bytes_le()?,
             None => vec![0u8; N::PROGRAM_ID_SIZE_IN_BYTES],
         };
 
+        let mut flattened_record_bytes = record_bytes.iter().flatten().collect::<Vec<&u8>>();
```
ljedrz:
```suggestion
        let mut flattened_record_bytes = record_bytes.iter().flatten().copied().collect::<Vec<u8>>();
```

dpc/src/record/tests.rs
```diff
@@ -26,9 +26,11 @@ pub(crate) const ITERATIONS: usize = 25;
 fn test_record_ciphertext() {
     let rng = &mut ChaChaRng::seed_from_u64(1231275789u64);
 
-    for _ in 0..ITERATIONS {
+    for i in 0..ITERATIONS {
         let account = Account::<Testnet2>::new(rng);
 
+        println!("i: {:?}", i);
```
ljedrz: should this remain here?

PR #648, opened 2022-02-10
Fix the failing clippy CI job, change the CircleCI resource class from 2xlarge to xlarge

ljedrz:
In addition, since there was the same breakage related to us using the `2xlarge` resource class in CircleCI, I'm adding a commit changing it to `xlarge` to see if it will suffice.


howardwu:
Ah sorry, I just noticed this! #651 had the same change, apologies!

PR #659, opened 2022-02-14
Extend BlockHeaderMetadata with LedgerRoot and TransactionsRoot

ljedrz:
This PR proposes that `LedgerRoot` and `TransactionsRoot` are inserted into `BlockHeaderMetadata` as opposed to being right before it in the `BlockHeader`. This will allow us to do 2 things with great snarkOS performance implications:
- use `BlockHeaderMetadata` (as opposed to the full `BlockHeader`) in `BlockLocators`, making their deserialization a lot faster, fixing issues like https://github.com/AleoHQ/snarkOS/issues/1632
- introduce a `DataMap<N::BlockHash, BlockHeaderMetadata<N>>` (in addition to the existing `DataMap<N::BlockHash, BlockHeader<N>>`) storage map that would allow us to perform all header-related storage operations that don't require the proof orders of magnitude faster, without any data duplication or low-level deserialization

This is a pure refactoring and shouldn't introduce any functional changes.

note: alternatives like splitting `BlockProof` (that would contain the nonce and the proof) out from the `BlockHeader` or any other variation where we can deserialize the current header partially into an object / objects without the proof would work; this PR just does that in the simplest way I could think of.

### DIFF
--- a/dpc/src/block/header.rs
+++ b/dpc/src/block/header.rs
@@ -34,7 +34,11 @@ use std::{mem::size_of, sync::atomic::AtomicBool};
 
 /// Block header metadata.
 #[derive(Clone, Debug, Eq, PartialEq, Serialize, Deserialize)]
-pub struct BlockHeaderMetadata {
+pub struct BlockHeaderMetadata<N: Network> {
+    /// The Merkle root representing the blocks in the ledger up to the previous block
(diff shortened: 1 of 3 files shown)


niklaslong (approved):
LGTM, certainly a desirable change 👍

Pratyush (approved):

PR #661, opened 2022-02-14
Initial improvements for `zip` safety

Pratyush:
## Motivation

Change many uses of `zip` to `zip_eq`, which enforces that the lengths of the zipped iterators are equal. In the remaining usages of `zip`, provide justification for why `zip` is sufficient. 

## Test Plan

(Write your test plan here)

## Related PRs

(Link your related PRs here)

### DIFF
--- a/algorithms/src/fft/domain.rs
+++ b/algorithms/src/fft/domain.rs
@@ -44,6 +44,9 @@ use std::fmt;
 #[cfg(feature = "parallel")]
 use rayon::prelude::*;
 
+#[cfg(not(feature = "parallel"))]
+use itertools::Itertools;
+
 /// Returns the ceiling of the base-2 logarithm of `x`.
 ///
 /// ```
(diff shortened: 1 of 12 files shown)


algorithms/src/fft/domain.rs
```diff
@@ -44,6 +44,9 @@ use std::fmt;
 #[cfg(feature = "parallel")]
 use rayon::prelude::*;
 
+#[cfg(not(feature = "parallel"))]
```
howardwu, line 47: Whats the purpose of the parallel cfg guard?

howardwu: (I presume it has something to do with Rayon compatibility)

Pratyush: `zip_eq` is available via both `Itertools` and `rayon::ParallelIterator`. In `parallel` mode, we use the latter impl, while in `not(parallel)` mode, the former. To avoid an unused import warning when `parallel` is set, I guard `Itertools` with `not(parallel)`.

howardwu (approved):

PR #664, opened 2022-02-16
Reduce the number of non-zero entries for the BHP circuit

Pratyush:
## Motivation

This reduces the maximum number of non-zero entries in the BHP gadget significantly.
* For the inner circuit, this results in a reduction from 538k to ~475k, which provides a big speed up when proving the inner circuit on Marlin (14s -> 11.6s on a server-class machine, which is a speedup of 17%). For context, proving the inner circuit with Groth16 takes ~3.8s on this machine, so with this PR, Marlin is only 3x slower than Groth16.

This change should also benefit any program SNARKs, as the transition hash will be cheaper to compute.

## Test Plan

(Write your test plan here)

## Related PRs

(Link your related PRs here)


howardwu:
LGTM. I witnessed 13.5s -> 11.1s

howardwu (approved):

PR #666, opened 2022-02-16
Adds batch addition for variable MSMs on BLS12-377

howardwu:
## Motivation

This PR adds batch addition for variable MSMs on BLS12-377. The technique used allows variable MSMs on short Weierstrass curves to exhibit a 20% performance improvements across all sizes.

Given two points `a := (x1, y1)` and `b := (x2, y2)`, this implementation computes `lambda := (y2 - y1) / (x2 - x1)` and subsequently computes the new point `c := (x3, y3)` where `x3 := lambda^2 - x1 - x2` and `y3 := lambda * (x1 - x3) - y1`. This approach is then generalized across a batch of points.

The size of the batch is determined empirically using performance benchmarks for BLS12-377 on Intel, AMD, and M1 machines. These values are determined by taking the L1 and L2 cache sizes and dividing them by the size of group elements (i.e. 96 bytes). As the algorithm itself requires caching additional values beyond the group elements, the ideal batch size is less than expectations, to accommodate those values. In general, it was found that undershooting is better than overshooting this heuristic.


algorithms/src/msm/variable_base/mod.rs
```diff
@@ -62,7 +80,36 @@ impl VariableBaseMSM {
...
+    ) -> G::Projective {
+        use rayon::prelude::*;
+        use snarkvm_fields::Zero;
+        use snarkvm_utilities::BitIteratorBE;
+
+        bases
+            .par_iter()
+            .zip_eq(scalars.par_iter())
+            .map(|(base, scalar)| base.mul_bits(BitIteratorBE::new(*scalar)))
+            .fold(|| G::Projective::zero(), |a, b| a + b)
```
Pratyush, line 111: This will still be slow, since I believe it's doing affine addition (which requires inverses)

algorithms/src/msm/variable_base/mod.rs
```diff
@@ -62,7 +80,36 @@ impl VariableBaseMSM {
...
+            .sum()
```
Pratyush: Can't we just directly call `sum()` here?

algorithms/src/msm/variable_base/standard.rs
```diff
@@ -75,39 +75,354 @@ fn process_window<G: AffineCurve>(
         res += running_sum;
     }
 
-    res
+    (res, window_size)
+}
+
+fn batched_window<G: AffineCurve>(
```
Pratyush: maybe make these as inherent methods of `MSMStrategy`? So you can call, e.g., `strategy.window(...)`. That would simplify the code below.

curves/src/traits/pairing_engine.rs
```diff
@@ -222,6 +222,20 @@ pub trait AffineCurve:
...
+    /// The element corresponding to the second operand becomes junk data.
+    fn batch_add_in_place_same_slice(bases: &mut [Self], index: &[(u32, u32)]);
+
+    /// Lookups up group elements according to index, and either adds and writes or simply
+    /// writes them to new_bases, using scratch space to store intermediate values. Scratch
+    /// space is always cleared after use.
+    fn batch_add_write(
+        bases: &[Self],
+        index: &[(u32, u32)],
+        new_bases: &mut Vec<Self>,
```
Pratyush:
```suggestion
        addition_result: &mut Vec<Self>,
```

curves/src/traits/pairing_engine.rs
```diff
@@ -222,6 +222,20 @@ pub trait AffineCurve:
...
     /// Checks that the current point is on the elliptic curve.
     fn is_on_curve(&self) -> bool;
+
+    /// Mutates bases in place and stores result in the first operand.
+    /// The element corresponding to the second operand becomes junk data.
+    fn batch_add_in_place_same_slice(bases: &mut [Self], index: &[(u32, u32)]);
+
+    /// Lookups up group elements according to index, and either adds and writes or simply
+    /// writes them to new_bases, using scratch space to store intermediate values. Scratch
+    /// space is always cleared after use.
```
Pratyush:
```suggestion
    /// If `(j, k)` is the `i`-th entry in `index`, then this method performs one of
    /// two actions:
    /// * `addition_result[i] = bases[j] + bases[k]`
    /// * `addition_result[i] = bases[j];
    ///
    /// It uses `scratch_space` to store intermediate values, and clears it after use.
```

curves/src/traits/pairing_engine.rs
```diff
@@ -222,6 +222,20 @@ pub trait AffineCurve:
 
     /// Checks that the current point is on the elliptic curve.
     fn is_on_curve(&self) -> bool;
+
+    /// Mutates bases in place and stores result in the first operand.
+    /// The element corresponding to the second operand becomes junk data.
```
Pratyush:
```suggestion
    /// If `(j, k)` is the `i`-th entry in `index`, then this method sets
    /// `bases[j] = bases[j] + bases[k]`. The state of `bases[k]` becomes unspecified.
```

curves/src/templates/twisted_edwards_extended/affine.rs
```diff
@@ -41,6 +41,50 @@ use std::{
     ops::{Add, AddAssign, Mul, MulAssign, Neg, Sub, SubAssign},
 };
 
+#[cfg(target_arch = "x86_64")]
+use crate::prefetch_slice;
+
+macro_rules! batch_add_loop_1 {
```
Pratyush: Let's make this into a function

curves/src/templates/twisted_edwards_extended/affine.rs
```diff
@@ -41,6 +41,50 @@ use std::{
...
+            $a.y *= &($inversion_tmp + &inversion_mul_d);
+
+            $b.x = P::BaseField::one() - &dx1x2y1y2.square();
+
+            $inversion_tmp *= &$b.x;
+        }
+    };
+}
+
+macro_rules! batch_add_loop_2 {
```
Pratyush: Same for this.

algorithms/Cargo.toml
```diff
@@ -127,6 +127,12 @@ features = [ "use_core" ]
 [dependencies.digest]
 version = "0.10"
 
+[dependencies.getrandom]
```
Pratyush, line 130: Should this be a dev-dependency?

howardwu: WASM requires this feature in order to instantiate an Rng

algorithms/benches/msm/variable_base.rs
```diff
@@ -14,39 +14,92 @@
...
+    let size = 1000000;
+    c.bench_function("Variable MSM on BLS12-377 (1,000,000)", |b| {
+        b.iter(|| VariableBase::msm(&bases[..size], &scalars[..size]))
+    });
+
+    let size = 2000000;
+    c.bench_function("Variable MSM on BLS12-377 (2,000,000)", |b| {
+        b.iter(|| VariableBase::msm(&bases[..size], &scalars[..size]))
+    });
+}
```
Pratyush:
```suggestion
    for size in [10_000, 100_000, 200_000, 300_000, 400_000, 500_000, 1_000_000, 2_000_000] {
    	c.bench_function(format!("VariableBase MSM on BLS12-377 ({})" size), |b| {
            b.iter(|| VariableBase::msm(&bases[..size], &scalars[..size]))
        });
    }
}
```

algorithms/src/msm/variable_base/mod.rs
```diff
@@ -14,95 +14,119 @@
...
     fn msm_naive<G: AffineCurve>(bases: &[G], scalars: &[<G::ScalarField as PrimeField>::BigInteger]) -> G::Projective {
-        let mut acc = G::Projective::zero();
+        use snarkvm_fields::Zero;
+        use snarkvm_utilities::BitIteratorBE;
 
+        let mut acc = G::Projective::zero();
         for (base, scalar) in bases.iter().zip(scalars.iter()) {
             acc += base.mul_bits(BitIteratorBE::new(*scalar));
         }
         acc
```
Pratyush:
```suggestion
        bases
            .iter()
            .zip_eq(scalars)
            .map(|(base, scalar)| base.mul_bits(BitIteratorBE::new(*scalar)))
            .sum()
```

algorithms/src/msm/variable_base/mod.rs
```diff
@@ -14,95 +14,119 @@
...
-                }
-            }
-        }
-        standard::msm_standard(bases, scalars)
+        use rayon::prelude::*;
+        use snarkvm_utilities::BitIteratorBE;
+
+        bases
+            .par_iter()
+            .zip_eq(scalars.par_iter())
```
Pratyush:
```suggestion
            .zip_eq(scalars)
```

Pratyush (approved):

PR #669, opened 2022-02-18
Create an indexable abstraction over powers of G on disk

### LINKED ISSUE #652: [Feature] Trimmable/expandable SRS
## 🚀 Feature

Implement logic which allows the SRS to be gradually downloaded as time goes on, rather than needing the whole thing from the get go.

## Motivation

This is mainly to avoid either bloating the binary with lots of preliminary data, or causing the node to get stuck for a long period of time as it downloads the transcript on startup.

## Implementation

We most likely will end up mapping the `powers_of_g` to disk, and creating a wrapper type that houses some read/write logic. This can then be used both for updating the disk storage of the SRS, as well as fetching what's needed from disk to memory in case we use it for proving/verifying later on.

The first version of this will see the SRS data being downloaded from chunked S3 buckets.

jules:
Fixes #652


algorithms/src/polycommit/data_structures/powers.rs
```diff
@@ -0,0 +1,375 @@
...
+impl<E: PairingEngine> CanonicalDeserialize for PowersOfG<E> {
+    fn deserialize<R: Read>(reader: &mut R) -> Result<Self, SerializationError> {
+        Ok(FromBytes::read_le(&mut *reader)?)
+    }
+}
+
+// NOTE: this drops the powers into a tmp file.
+// I assume this is only used for testing but this needs to be verified.
+impl<E: PairingEngine> From<Vec<E::G1Affine>> for PowersOfG<E> {
+    fn from(value: Vec<E::G1Affine>) -> Self {
```
howardwu: Could you clarify where this `from` method is used?

If we are confident it is exclusively for testing only, we should `cfg` guard it with `#[cfg(test)]` above the `impl` line (L153)

howardwu (approved):

algorithms/src/polycommit/data_structures/powers.rs
```diff
@@ -0,0 +1,375 @@
...
+    /// clones.
+    #[derivative(Debug = "ignore")]
+    file: Arc<RwLock<File>>,
+    /// Group elements of the form `{ \beta^i \gamma G }`, where `i` ranges from 0 to `degree`.
+    /// These are used for hiding.
+    pub powers_of_beta_times_gamma_g: Arc<RwLock<BTreeMap<usize, E::G1Affine>>>,
+    /// The degree up to which we have powers.
+    /// Mostly just used for ensuring we download in the proper order,
+    /// length checks are actually done by checking file metadata.
+    degree: Arc<RwLock<usize>>,
```
Pratyush: Should we make a new struct and put that entire struct under a `Arc<RwLock<...>>`?

algorithms/src/polycommit/data_structures/powers.rs
```diff
@@ -0,0 +1,375 @@
...
+}
+
+impl<E: PairingEngine> FromBytes for PowersOfG<E> {
+    fn read_le<R: Read>(mut reader: R) -> std::io::Result<Self> {
+        // Deserialize `powers_of_beta_times_gamma_g`.
+        let mut powers_of_beta_times_gamma_g = BTreeMap::new();
+        let powers_of_gamma_g_num_elements: u32 = FromBytes::read_le(&mut reader)?;
+        for _ in 0..powers_of_gamma_g_num_elements {
+            let key: u32 = FromBytes::read_le(&mut reader)?;
+            let power_of_gamma_g: E::G1Affine = FromBytes::read_le(&mut reader)?;
```
Pratyush, line 122: Doesn't this use the default impl which performs subgroup and on-curve checks?

algorithms/src/polycommit/data_structures/powers.rs
```diff
@@ -0,0 +1,375 @@
...
+    fn serialized_size(&self) -> usize {
+        todo!()
+    }
+}
+
+impl<E: PairingEngine> CanonicalDeserialize for PowersOfG<E> {
+    fn deserialize<R: Read>(reader: &mut R) -> Result<Self, SerializationError> {
+        Ok(FromBytes::read_le(&mut *reader)?)
+    }
+}
```
Pratyush, line 148: I would recommend we switch the impls: make `FromBytes` call `CanonicalDeserialize`, and ditto for `ToBytes`.

algorithms/src/polycommit/data_structures/powers.rs
Pratyush: Hm let me revive #459; it enables unchecked deserialization which enables omitting on-curve and in-subgroup checks when deserializing from trusted sources.

algorithms/src/polycommit/data_structures/powers.rs
```diff
@@ -0,0 +1,375 @@
...
+#[derive(Derivative)]
+#[derivative(Debug(bound = ""))]
+pub struct PowersOfG<E: PairingEngine> {
+    /// Filepath of the powers we're using.
+    file_path: String,
+    /// A handle to the file on disk containing the powers of G.
+    /// The handle is guarded to avoid read/write conflicts with potential
+    /// clones.
+    #[derivative(Debug = "ignore")]
+    file: Arc<RwLock<File>>,
```
Pratyush: Are these elements of the form `[G, \beta * G, \beta^2 * G, ..., \beta^{d} G]`? If so, we should specify that in the comment.

algorithms/src/polycommit/data_structures/powers.rs
```diff
@@ -0,0 +1,375 @@
...
+        }
+        alpha_powers_g1[3..].iter().chunks(3).into_iter().enumerate().for_each(|(i, c)| {
+            // Avoid underflows and just stop populating the map if we're going to.
+            if self.len() - 1 > (1 << i) {
+                let c = c.into_iter().collect::<Vec<_>>();
+                alpha_tau_powers_g1.insert(self.len() - 1 - (1 << i) + 2, *c[0]);
+                alpha_tau_powers_g1.insert(self.len() - 1 - (1 << i) + 3, *c[1]);
+                alpha_tau_powers_g1.insert(self.len() - 1 - (1 << i) + 4, *c[2]);
+            }
+        });
```
Pratyush, line 367: What is `alpha` and what is `tau` here?

algorithms/src/polycommit/sonic_pc/mod.rs
```diff
@@ -116,9 +116,10 @@ impl<E: PairingEngine> PolynomialCommitment<E::Fr, E::Fq> for SonicKZG10<E> {
         enforced_degree_bounds: Option<&[usize]>,
     ) -> Result<(Self::CommitterKey, Self::VerifierKey), PCError> {
         let trim_time = start_timer!(|| "Trimming public parameters");
-        let max_degree = pp.max_degree();
+        let mut max_degree = pp.max_degree();
         if supported_degree > max_degree {
-            return Err(PCError::TrimmingDegreeTooLarge);
+            pp.powers_of_beta_g.download_up_to(supported_degree).map_err(|_| PCError::TrimmingDegreeTooLarge)?;
+            max_degree = pp.max_degree();
```
Pratyush, line 122: We never hit this line because if we're in the error case then the `?` returns.

algorithms/src/polycommit/sonic_pc/mod.rs
```diff
@@ -146,17 +147,20 @@ impl<E: PairingEngine> PolynomialCommitment<E::Fr, E::Fq> for SonicKZG10<E> {
                     max_degree - lowest_shift_degree + 1
                 ));
 
-                let shifted_powers_of_beta_g = pp.powers_of_beta_g[lowest_shift_degree..].to_vec();
+                println!("lowest shift {}", lowest_shift_degree);
```
Pratyush:
```suggestion
```

algorithms/src/polycommit/data_structures/powers.rs
```diff
@@ -0,0 +1,375 @@
...
+    /// Filepath of the powers we're using.
+    file_path: String,
+    /// A handle to the file on disk containing the powers of G.
+    /// The handle is guarded to avoid read/write conflicts with potential
+    /// clones.
+    #[derivative(Debug = "ignore")]
+    file: Arc<RwLock<File>>,
+    /// Group elements of the form `{ \beta^i \gamma G }`, where `i` ranges from 0 to `degree`.
+    /// These are used for hiding.
+    pub powers_of_beta_times_gamma_g: Arc<RwLock<BTreeMap<usize, E::G1Affine>>>,
```
Pratyush: Let's make this field private, and offer only access via getter methods.

algorithms/src/polycommit/data_structures/powers.rs
```diff
@@ -0,0 +1,375 @@
...
+    /// up to the 15th degree, so this will always return false.
+    pub fn is_empty(&self) -> bool {
+        false
+    }
+
+    /// Returns an element at `index`.
+    /// NOTE: `std::ops::Index` was not used here as the trait requires
+    /// that we return a reference. We can not return a reference to
+    /// something that does not exist when this function is called.
+    pub fn index(&self, index: usize) -> E::G1Affine {
```
Pratyush:
```suggestion
    pub fn power_of_beta_g(&self, which_power: usize) -> E::G1Affine {
```

algorithms/src/polycommit/data_structures/powers.rs
```diff
@@ -0,0 +1,375 @@
...
+        let mut reader = BufReader::new(&*file);
+        reader.seek(SeekFrom::Start(index_start as u64)).expect("could not seek to element starting index");
+
+        // Now read it out, deserialize it, and return it.
+        E::G1Affine::deserialize(&mut reader).expect("powers of g corrupted")
+    }
+
+    /// Slices the underlying file to return a vector of affine elements
+    /// between `lower` and `upper`.
+    pub fn slice(&self, lower: usize, upper: usize) -> Vec<E::G1Affine> {
```
Pratyush:
```suggestion
    pub fn powers_of_beta_g(&self, lower: usize, upper: usize) -> Vec<E::G1Affine> {
```

algorithms/src/polycommit/kzg10/data_structures.rs
```diff
@@ -44,10 +45,7 @@ use std::{collections::BTreeMap, io};
 pub struct UniversalParams<E: PairingEngine> {
     /// Group elements of the form `{ \beta^i G }`, where `i` ranges from 0 to `degree`.
     /// These represent the monomial basis evaluated at `beta`.
-    pub powers_of_beta_g: Vec<E::G1Affine>,
-    /// Group elements of the form `{ \beta^i \gamma G }`, where `i` ranges from 0 to `degree`.
-    /// These are used for hiding.
-    pub powers_of_beta_times_gamma_g: BTreeMap<usize, E::G1Affine>,
+    pub powers_of_beta_g: PowersOfG<E>,
```
Pratyush:
```suggestion
    pub powers: PowersOfG<E>,
```

algorithms/src/polycommit/data_structures/powers.rs
jules: I'm pretty confident it is, that's a really good idea :)

algorithms/src/polycommit/data_structures/powers.rs
jules: Not entirely sure - I copied the code from some of the setup scripts. I can change the naming if needed, but I figured I'd stick with it for now!

algorithms/src/polycommit/data_structures/powers.rs
jules: I mostly picked this configuration since I stuck to the original code that uses `From/ToBytes` for everything. If we can get unchecked (de)serialization going through `CanonicalSerialize/Deserialize` though that would be great - the space efficiency won't be a massive issue but the deserialization speedup would be fantastic, and then I think swapping the functions would make a bit more sense.

algorithms/src/polycommit/sonic_pc/mod.rs
jules: I don't think we should hit that line in case there's an error, only if we succeed in downloading the needed powers.

algorithms/src/polycommit/data_structures/powers.rs
jules: Ah, it will actually prove quite tough, as the `setup` functions currently aren't `#[cfg(test)]` -- not sure if we want to guard those too?

algorithms/src/polycommit/data_structures/powers.rs
```diff
@@ -0,0 +1,377 @@
...
+    /// This file should never be empty since we always write at least
+    /// up to the 15th degree, so this will always return false.
+    pub fn is_empty(&self) -> bool {
+        false
+    }
+
+    /// Returns an element at `index`.
+    /// NOTE: `std::ops::Index` was not used here as the trait requires
+    /// that we return a reference. We can not return a reference to
+    /// something that does not exist when this function is called.
```
Pratyush:
```suggestion
    /// Returns the power of beta times G specified by `which_power`.
    // NOTE: `std::ops::Index` was not used here as the trait requires
    // that we return a reference. We can not return a reference to
    // something that does not exist when this function is called.
```

algorithms/src/polycommit/kzg10/data_structures.rs
```diff
@@ -65,33 +66,78 @@ pub struct UniversalParams<E: PairingEngine> {
     pub prepared_beta_h: <E::G2Affine as PairingCurve>::Prepared,
 }
 
+impl<E: PairingEngine> CanonicalSerialize for UniversalParams<E> {
+    fn serialize<W: Write>(&self, writer: &mut W) -> Result<(), SerializationError> {
+        self.powers.read().serialize(writer)?;
```
Pratyush, line 65: Can we not just copy from `self.powers` to `writer`? This would avoid a serialize <-> deserialize round trip

Pratyush:
This looks mostly good to me, modulo the couple of small comments!

algorithms/src/polycommit/kzg10/data_structures.rs
jules: What do you mean? In case you're talking about the `read` call, it's for holding a read lock

Pratyush: Ah I see. For some reason I thought the read call was reading from disk

Pratyush:
I think we can merge this! Maybe this will solve some of the parameter-related issues in the other PRs.

jules:
@Pratyush awesome! I guess let’s have @howardwu do a final sign off :)

howardwu (approved):

PR #675, opened 2022-02-22
Introduce value balance commitments

raychu86:
## Motivation

This PR introduces the concept of value balance commitments for value conservation. The scheme was derived from Zcash's redDSA binding signatures.


howardwu:
All of these `to_bytes_le()` calls are incredibly inefficient. The next step is to unify types to mitigate usage of `ToBytes`

howardwu:
@raychu86 update the circuit implementation to match the native version

howardwu:
@raychu86 please review my Network changes carefully. In particular, I still am not confident on the window sizes. Note that changing from (4, 350) -> (8, 125) doesn't seem to work despite the same size.

howardwu:
@raychu86 the Pedersen commitment gadget is missing its tests

howardwu:
@raychu86 is using `zero` as the randomness for the value commitment correct?

howardwu:
@raychu86 switch `i64` to `AleoAmount` (and rename `AleoAmount` to `Amount`)

howardwu:
> @raychu86 study the Network changes carefully

In particular, I still am not confident on the window sizes. Note that changing from (4, 350) -> (8, 125) doesn't seem to work despite the same size.

raychu86:
@howardwu 

> the Pedersen commitment gadget is missing its tests

Test readded.

> is using zero as the randomness for the value commitment correct?

Yes the scheme requires using a zero for the commitment of the value balance.

> switch i64 to AleoAmount (and rename AleoAmount to Amount)

This is change is formalized in #690

> please review my Network changes carefully. In particular, I still am not confident on the window sizes. Note that changing from (4, 350) -> (8, 125) doesn't seem to work despite the same size.

After some investigation, I found that the homomorphic properties of our Pedersen commitments don't seem to hold if the `WINDOW_SIZE` is less than `SCALAR_BITS` which in this case is 253. 

I believe it has to do with this operation: 
```
// Compute h^r.
let scalar_bits = BitIteratorLE::new(randomness.to_repr());
for (bit, power) in scalar_bits.into_iter().zip(&self.random_base) {
    if bit {
        output += power
    }
}
```
where the zip may terminate early because `self.random_base.len() = WINDOW_SIZE` and `scalar_bits.len() = 253`. This has me curious about if this operation is implemented correctly since those last bits are simply disregarded.

howardwu:
LGTM, I've moved the remaining issues to #695

howardwu (approved):

PR #678, opened 2022-02-22
Fix indexing error for `create_proof_no_zk`

raychu86:
## Motivation

Ensures that if a Groth16 proof is being constructed without zk or with an `r` value of 0, then the indexing of the execution pool results are correct.

## Test Plan

A test for `create_proof_no_zk` has been added.


raychu86:
Closed in favor of #680 for now

PR #685, opened 2022-02-25
[Testnet2] Enforce Merkle path length

raychu86:
## Motivation

Ensures that the Merkle path is only valid if the path length is equivalent to the depth of the tree. This enforces that Merkle paths are only checking the leaves and prevents verification of inner nodes.

Tracking PR: #679 

## Test Plan

A test has been added to check that the Merkle path length is valid.


Pratyush:
Q: should we also hash the node height at each level? i.e. for the i-th deep node, should we set the hash to be CRH(i || left_child || right_child)?

PR #690, opened 2022-03-03
Update `AleoAmount` to `Amount` and introduce standard operations

raychu86:
## Motivation

This PR renames`AleoAmount` to `Amount` and updates the `add` and `sub` functions. Additionally, the `std::ops::Add`, `std::ops::AddAssign`, `std::ops::Sub`, and `std::ops::SubAssign` traits were implemented.

## Test Plan

Tests have been added to check that overflows/underflows are handled and that the new syntax holds.


dpc/src/virtual_machine/amount.rs
```diff
@@ -53,34 +55,34 @@ impl fmt::Display for Denomination {
...
+        Ok(Self::from_gate(self.0.checked_add(b.0).ok_or(anyhow!("Amount addition overflow."))?))
     }
 
-    /// Subtract the value of two `AleoAmounts`
+    /// Subtract the value of two `Amount`
     #[allow(clippy::should_implement_trait)]
-    pub fn sub(self, b: AleoAmount) -> Self {
-        Self::from_gate(self.0 - b.0)
+    pub fn sub(self, b: Self) -> Result<Self> {
+        Ok(Self::from_gate(self.0.checked_sub(b.0).ok_or(anyhow!("Amount subtraction underflow."))?))
```
Pratyush:
```suggestion
        self.0.checked_sub(b.0).ok_or(anyhow!("Amount subtraction underflow.")).map(Self::from_gate)
```

dpc/src/virtual_machine/amount.rs
```diff
@@ -53,34 +55,34 @@ impl fmt::Display for Denomination {
...
         Self::from_gate(aleo_value * 10_i64.pow(Denomination::CREDIT.precision()))
     }
 
-    /// Add the values of two `AleoAmount`s
+    /// Add the values of two `Amount`s
     #[allow(clippy::should_implement_trait)]
-    pub fn add(self, b: Self) -> Self {
-        Self::from_gate(self.0 + b.0)
+    pub fn add(self, b: Self) -> Result<Self> {
+        Ok(Self::from_gate(self.0.checked_add(b.0).ok_or(anyhow!("Amount addition overflow."))?))
```
Pratyush:
```suggestion
        self.0.checked_add(b.0).ok_or(anyhow!("Amount addition overflow.")).map(Self::from_gate)
```

dpc/src/virtual_machine/amount.rs
```diff
@@ -106,19 +108,59 @@ impl AleoAmount {
     }
 }
 
-impl Sum for AleoAmount {
-    fn sum<I: Iterator<Item = AleoAmount>>(iter: I) -> AleoAmount {
-        iter.fold(AleoAmount::ZERO, |a, b| a.add(b))
+impl Add<Self> for Amount {
```
Pratyush: Should we also add an `impl Add<Result<Self>> for Amount` that simply propagates the error? Same for `Sub`.

raychu86: Do you mean something like this?
```
impl Add<Self> for Amount {
    type Output = Result<Amount>;

    fn add(self, other: Self) -> Self::Output {
        self.add(other)
    }
}
```

I believe `impl Add<Result<Self>> for Amount` means adding `Amount + Result<Amount>`

raychu86: I thought about setting, `type Output = Result<Amount>;` but ran into issues where we handle Amount arithmetic in iterators/maps/`sum()` and functions with no error handler. 

For simplicity, I went with the current `panic` approach. Open to other suggestions or potentially making all the functions handling Amount arithmetic require error handlers (if that's desired).

howardwu:
Whats the status on this?

PR #694, opened 2022-03-09
Integrate value balance commitments into the `InnerCircuit`

raychu86:
## Motivation

(Write your motivation here)

This PR integrates the value balance commitment into the inner circuit. The virtual machine builds each value commitment and the final value balance commitment in the `build` phase for each transition.


dpc/src/circuits/inner_circuit.rs
```diff
@@ -772,6 +825,43 @@ impl<N: Network> ConstraintSynthesizer<N::InnerScalarField> for InnerCircuit<N>
...
+                let given_value_commitment_gadget = <N::ValueCommitmentGadget as CommitmentGadget<
+                    N::ValueCommitmentScheme,
+                    N::InnerScalarField,
+                >>::OutputGadget::alloc_input(
+                    &mut vc_cs.ns(|| format!("Output value commitment {}", j)),
+                    || Ok(value_commitment),
+                )?;
+
+                let value_bytes =
+                    UInt8::alloc_vec(vc_cs.ns(|| format!("Output value bytes {}", j)), &record.value().to_bytes_le()?)?;
```
howardwu: This witness is unchecked and technically allows sidestepping the actual record value. (This applies for the input and output records)

howardwu: I've switched it to the derived bytes

PR #706, opened 2022-03-18
Refactor Marlin implementation

Pratyush:
## Motivation

This PR refactors the Marlin implementation to simplify the prover and verifier algorithms, with the aim of making future changes (batch proving + lookup tables) easier to implement. 

Builds upon #708; should be merged after that PR is merged.

## Test Plan

(Write your test plan here)

## Related PRs

(Link your related PRs here)


Pratyush:
This PR is ready for review. The proof size has gone down, which is good, and as a result the tests aren't passing any more (which makes sense). However I can't get tests to pass locally even after resampling parameters, so any insight would be valuable.

Pratyush:
Ah sorry, should have closed this; the PC refactor contains these commits. Can we revert this?

PR #707, opened 2022-03-19
Refactors CRH and Commitment, improves performance

howardwu:
## Motivation

This PR does not change the expected result of CRH or Commitment schemes. Rather, it removes inefficiencies that reduces the number of constraints.

- Improves Pedersen hashing by 4%, and BHP hashing by 41%
- Reduces the BHP CRH gadget constraint count by ~17%
- Fixes an incorrect enforcement of number of `bytes` as opposed to number of `bits`:
```
input.len() <= WINDOW_SIZE * NUM_WINDOWS // input is sometimes [u8], when we should be checking [bool]
```
- Fixes an incorrect enforcement of number of `bits`:
```
input.len() <= WINDOW_SIZE * NUM_WINDOWS // before
input.len() <= WINDOW_SIZE * NUM_WINDOWS * BHP_CHUNK_SIZE // after
```
- Fixes BHP input padding from a fixed constant to the actual input size.
    - Note: This did not change the native output, however it did reduce the number of constraints for the BHP hash circuit, thus requiring a resampling of parameters.
    
Remark: This PR improves transaction generation times by 4.7%.


algorithms/src/commitment/bhp.rs
```diff
@@ -56,77 +51,25 @@ impl<G: ProjectiveCurve, const NUM_WINDOWS: usize, const WINDOW_SIZE: usize> Com
...
-        for (bit, power) in scalar_bits.take(G::ScalarField::size_in_bits()).zip_eq(&self.random_base) {
+        let scalar_bits = BitIteratorLE::new(randomness.to_repr()).take(G::ScalarField::size_in_bits());
+        for (bit, power) in scalar_bits.zip_eq(&self.random_base) {
             if bit {
                 output += power
             }
         }
 
-        let affine = output.into_affine();
-        debug_assert!(affine.is_in_correct_subgroup_assuming_on_curve());
```
Pratyush: Why'd we remove this debug assertion?

algorithms/src/crh/bhp.rs
```diff
@@ -51,43 +40,69 @@ impl<G: ProjectiveCurve, const NUM_WINDOWS: usize, const WINDOW_SIZE: usize> CRH
...
-        }
-
-        let maximum_num_chunks_in_segment = calculate_num_chunks_in_segment::<G::ScalarField>();
-        if WINDOW_SIZE > maximum_num_chunks_in_segment {
-            panic!(
-                "BHP CRH must have a window size resulting in scalars < (p-1)/2, \
-                 maximum segment size is {}",
-                maximum_num_chunks_in_segment
-            );
+        debug_assert_eq!(BHP_CHUNK_SIZE, 3);
```
Pratyush: we can move this to a compile time assertion:

```rust
const ASSERT: () = assert!(BHP_CHUNK_SIZE == 3);
```

algorithms/src/crh/bhp.rs
```diff
@@ -96,164 +111,34 @@ impl<G: ProjectiveCurve, const NUM_WINDOWS: usize, const WINDOW_SIZE: usize> CRH
...
-
-    /// Precondition: number of elements in `input` == `num_bits`.
-    pub(crate) fn hash_bits_inner<S: Borrow<bool>>(
-        &self,
-        input: impl Iterator<Item = S>,
-        num_bits: usize,
-    ) -> Result<G, CRHError> {
-        if num_bits > WINDOW_SIZE * NUM_WINDOWS {
-            return Err(CRHError::IncorrectInputLength(num_bits, WINDOW_SIZE, NUM_WINDOWS));
+    pub(crate) fn hash_bits_inner(&self, input: &[bool]) -> Result<G, CRHError> {
```
Pratyush, line 112: This should just output `G::BaseField`.

algorithms/src/crh/bhp.rs
```diff
@@ -96,164 +111,34 @@ impl<G: ProjectiveCurve, const NUM_WINDOWS: usize, const WINDOW_SIZE: usize> CRH
...
+        // Ensure the input size is within the parameter size,
+        if input.len() > NUM_WINDOWS * WINDOW_SIZE * BHP_CHUNK_SIZE {
```
Pratyush, line 114: This is different from the earlier input.len() check

algorithms/src/commitment/bhp.rs
howardwu: I'll revert

howardwu: Done

algorithms/src/crh/bhp.rs
```diff
@@ -96,164 +111,34 @@ impl<G: ProjectiveCurve, const NUM_WINDOWS: usize, const WINDOW_SIZE: usize> CRH
...
-        }
-        let base_lookup = self.base_lookup(&self.bases);
-        debug_assert_eq!(base_lookup.len(), NUM_WINDOWS);
-        for bases in base_lookup.iter() {
-            debug_assert_eq!(bases.len(), WINDOW_SIZE);
+        // Pad the input to a multiple of `BHP_CHUNK_SIZE` for hashing.
+        let mut input = input.to_vec();
+        if input.len() % BHP_CHUNK_SIZE != 0 {
+            let padding = BHP_CHUNK_SIZE - (input.len() % BHP_CHUNK_SIZE);
+            input.extend_from_slice(&vec![false; padding]);
```
Pratyush:
```suggestion
            let padding = BHP_CHUNK_SIZE - (input.len() % BHP_CHUNK_SIZE);
            input.extend_from_slice(&[false; 3][..padding]);
```

Pratyush: Avoids an allocation

algorithms/src/crh/bhp.rs
```diff
@@ -96,164 +111,34 @@ impl<G: ProjectiveCurve, const NUM_WINDOWS: usize, const WINDOW_SIZE: usize> CRH
...
-        );
-        for bases in self.bases.iter() {
-            debug_assert_eq!(bases.len(), WINDOW_SIZE);
-        }
-        let base_lookup = self.base_lookup(&self.bases);
-        debug_assert_eq!(base_lookup.len(), NUM_WINDOWS);
-        for bases in base_lookup.iter() {
-            debug_assert_eq!(bases.len(), WINDOW_SIZE);
+        // Pad the input to a multiple of `BHP_CHUNK_SIZE` for hashing.
+        let mut input = input.to_vec();
```
Pratyush, line 119: We might want to keep the stack allocation logic

algorithms/src/crh/bhp.rs
howardwu: The BHP commitment needs the output to compute the randomness on top of it.

algorithms/src/crh/bhp.rs
howardwu: Wdym?

algorithms/src/crh/bhp.rs
howardwu: Done

algorithms/src/crh/bhp.rs
howardwu: The stack allocation logic requires a worst case allocation, which ends up being too expensive.

algorithms/src/crh/bhp.rs
Pratyush: ah good point.

howardwu:
I ended up removing `const ASSERT: () = assert!(BHP_CHUNK_SIZE == 3);` since it's now trivially true.

algorithms/src/crh/bhp.rs
Pratyush: Sorry, didn't realize you noted that already in the PR description. this aligns with the arkworks code.

Pratyush (approved):

PR #708, opened 2022-03-20
Removes `Group` trait, refactors Poseidon and related algorithms

howardwu:
## Motivation

This PR builds upon #707 .

This PR removes the `Group` trait, in favor of a projective and affine trait. In addition, Poseidon defaults are moved to `snarkvm-fields` in preparation for a `circuits` implementation.

There are more enhancements for the affine trait, however as this PR is already quite large, I am checkpointing here.

The parameters are resampled because the change to `GroupGadget` (to my knowledge) has slightly changed the layout of the circuit assignments. Note the # of constraints for all circuits remains unchanged, and the native-to-circuit tests show all results remain the same.

All tests continue to pass as expected.

#### Key Changes

- Removes the `Group` trait and enforces all gadgets use the affine form for group representation
- Refactors Poseidon defaults from `snarkvm_algorithms` to `snarkvm_fields` in preparation for a circuit implementation

#### Minor Changes

- Moves `sha256` from `crh/` to `crypto_hash/`
- Moves `hash_to_curve` from `hash_to_curve/` to `crypto_hash/`


howardwu:
^Need to resample parameters as byte allocations have improved in constraint count.

algorithms/src/merkle_tree/merkle_path.rs
```diff
@@ -31,10 +31,8 @@ pub type MerkleTreeDigest<P> = <<P as MerkleParameters>::H as CRH>::Output;
 
 /// Stores the hashes of a particular path (in order) from leaf to root.
 /// Our path `is_left_child()` if the boolean in `path` is true.
-#[derive(Derivative)]
```
Pratyush, line 34: Is there a reason to change this? In particular now we're comparing the `Parameters`, I think.

howardwu: The check on `Parameters` will check the instance of the CRH (which itself is empty) and the setup message matches.

howardwu: Ray is planning to tackle the Merkle tree with a 2-to-1 refactor this week

algorithms/src/snark/marlin/mode.rs
```diff
@@ -17,20 +17,20 @@
 use core::fmt::Debug;
 
 /// A trait to specify the Marlin mode.
-pub trait MarlinMode: Clone + Debug + 'static + Sync + Send {
+pub trait MarlinMode: 'static + Copy + Clone + Debug + PartialEq + Eq + Sync + Send {
```
Pratyush:
```suggestion
pub trait MarlinMode: 'static + Copy + Debug + Eq + Sync + Send {
```

fields/src/to_field_vec.rs
```diff
@@ -77,7 +83,7 @@ impl<F: PrimeField> ToConstraintField<F> for [u8] {
     #[inline]
     fn to_field_elements(&self) -> Result<Vec<F>, ConstraintFieldError> {
         // Derive the field size in bytes, floored to be conservative.
-        let floored_field_size_in_bytes = (<F as PrimeField>::Parameters::CAPACITY / 8) as usize;
+        let floored_field_size_in_bytes = (F::size_in_data_bits() / 8) as usize;
```
Pratyush, line 86: Where is this defined?

howardwu: `F::size_in_data_bits()` is a trait method that returns `<F as PrimeField>::Parameters::CAPACITY`, found in `snarkvm_fields` `PrimeField` trait

howardwu: The way we pack the `[u8]` isn't my favorite, however I tried modifying it, and couldn't get it to work in the brief attempt I made. The current changes here are merely for readability, the logic remains unchanged

Pratyush:
This mostly looks good to me, but the PR is quite large; is there a specific file/set of changes that I should look at?

howardwu:
The main change (which admittedly is a spider web) is the `twisted_edwards` native implementation changes.

I tried my best to make this PR as sane as possible, the challenge is that removing the `Group` trait is really an all-or-nothing type PR (couldn't split it as the curve impls, algo impls, GroupGadget, algo gadgets, all depend on it and wouldn't compile otherwise)

howardwu:
Regarding the need to resample parameters, the things that give me assurance of this PR's correctness are:

1. The native curve arithmetic and native hashes continue to compute the same results.
2. The gadget tests (which check against the native impls) continue to pass (meaning 1-1 correspondence).
3. The DPC tests (when it's not dependent on the serialized proving key file) all pass.
4. The inner and posw circuits continue to have the same constraint count and assignments.

The failing tests that motivated a resample were caused by the call to `N::InnerSNARK::verify(proving_key, input, proof)` failing. This implies (given the above) that there was likely a reordering of constraints or witness assignments because of the logic changes to `GroupGadget` impls, which was not accounted for in the `proving_key` file.

Pratyush (approved):

PR #711, opened 2022-03-21
Reduce the number of constraints for Field and Scalar comparison operations

d0cd:
- Uses parity check to implement a cheap compare for `Scalar`.
- Introduces a new method for bitwise `less_than`.


circuits/types/field/src/lsb.rs
```diff
@@ -0,0 +1,132 @@
...
+                Some(byte) => byte,
+                None => E::halt("Expected at least one byte."),
+            };
+            // Mask all but the lowest bit.
+            let mut lsb_bytes = vec![lsb_byte & 1u8];
+            // Zero-extend the field element.
+            lsb_bytes.extend(iter::repeat(0u8).take(bytes.len()));
+
+            // Mask lower bit.
+            bytes.push(lsb_byte & 254u8);
```
Pratyush:
```suggestion
            bytes.push(lsb_byte & 0b1111_1110u8);
```

circuits/types/field/src/lsb.rs
```diff
@@ -0,0 +1,132 @@
...
+use super::*;
+use snarkvm_utilities::{FromBytes, ToBytes};
+use std::iter;
+
+impl<E: Environment> LSB for Field<E> {
+    type Boolean = Boolean<E>;
+
+    /// Returns the LSB of the integer.
+    fn lsb(&self) -> Self::Boolean {
+        let mut bytes = match self.eject_value().to_bytes_le() {
```
Pratyush: if possible we should use `to_bits` here, to avoid dealing with weird byte padding issues.

circuits/types/field/src/lsb.rs
```diff
@@ -0,0 +1,132 @@
...
+            bytes.push(lsb_byte & 254u8);
+            bytes.reverse();
+
+            // Instantiate witnesses for the lsb and upper bits of self.
+            let lsb = &Self::new(self.eject_mode(), E::BaseField::from_bytes_le(&lsb_bytes).unwrap());
+            let upper_bits = Self::new(self.eject_mode(), E::BaseField::from_bytes_le(&bytes).unwrap());
+
+            // Ensure (1 - lsb) * lsb = 0.
+            // `lsb` must be either 0 or 1.
+            E::assert_eq((Self::one() - lsb) * lsb, Self::zero());
```
Pratyush: I can always make this pass by setting `upper_bits` to be `self - lsb`, for my choice of `lsb`. For example, for `self = 7`, I can set `lsb = 0` and `upper_bits = 7`. This would pass the existing constraints, but is clearly wrong.

EDIT: to clarify, the above means that the existing constraints are insufficient to prevent cheating.

circuits/types/field/src/lsb.rs
```diff
@@ -0,0 +1,132 @@
...
+
+            // Ensure that the witnesses for the lsb and upper bits sum up to the original value.
+            E::assert_eq(self, upper_bits + lsb);
+
+            lsb.is_equal(&Self::one())
```
Pratyush: This is unnecessary; since `lsb` is already boolean constrained, we should be able to convert the underlying low-level variable to the boolean type directly.

circuits/types/scalar/src/compare.rs
```diff
@@ -93,37 +83,46 @@ mod tests {
...
     fn test_private_is_less_than_public() {
-        check_is_less_than(Mode::Private, Mode::Public, 2, 0, 1250, 1250);
+        check_is_less_than(Mode::Private, Mode::Public, 0, 0, 5, 6);
     }
 
     #[test]
     fn test_private_is_less_than_private() {
-        check_is_less_than(Mode::Private, Mode::Private, 2, 0, 1250, 1250);
+        check_is_less_than(Mode::Private, Mode::Private, 0, 0, 5, 6);
     }
```
Pratyush: We can write these tests more concisely as `for (mode1, mode2) in [Mode::Private, Mode::Public, Mode::Constant].into_iter().product()`

circuits/types/field/src/lsb.rs
howardwu: Yes, this is castable for free

Pratyush:
As it stands I don't think the LSB constraints are sound, so we should think of alternative ways of obtaining the LSB.

d0cd:
Agreed, I'll mark this PR as a draft and resolve by EOW.

circuits/types/scalar/src/compare.rs
```diff
@@ -15,29 +15,51 @@
...
-        let bits_be = self.bits_le.iter().rev().zip_eq(other.bits_le.iter().rev());
-
-        for (index, (self_bit, other_bit)) in bits_be.enumerate() {
-            // Determine if `self` is less than `other` up to the `index`-th bit.
-            is_less_than |= &are_previous_bits_equal & (!self_bit & other_bit);
-
-            // Skip the update to the LSB, as this boolean is subsequently discarded.
-            if index != self.bits_le.len() - 1 {
-                are_previous_bits_equal &= self_bit.is_equal(other_bit);
+        // TODO: Will ScalarField ever outsize BaseField?
```
Pratyush: No, at least not for our curves.

circuits/types/scalar/src/compare.rs
```diff
@@ -15,29 +15,51 @@
...
+        let scalar_modulus = match E::ScalarField::modulus().to_bytes_le() {
+            Ok(modulus_bytes) => match E::BaseField::from_bytes_le(&modulus_bytes) {
+                Ok(modulus) => modulus,
+                Err(error) => E::halt(format!("Failed to load the scalar modulus as a constant: {error}")),
+            },
+            Err(error) => E::halt(format!("Failed to retrieve the scalar modulus as bytes: {error}")),
+        };
+        if scalar_modulus.to_repr() <= E::BaseField::modulus_minus_one_div_two() {
```
Pratyush: For us, this condition always holds. (We can add a debug_assert to ensure that.)

circuits/types/scalar/src/compare.rs
```diff
@@ -15,29 +15,51 @@
...
+            // If all elements of the scalar field are less than (p - 1)/2, where p is the modulus of
+            // the base field, then we can perform an optimized check for `less_than`.
+            // We compute the less than operation by checking the parity of 2 * (self - other) mod p.
+            // If a < b, then 2 * (self - other) mod p is odd.
+            // If a >= b, then 2 * (self - oher) mod p is even.
+            if self.is_constant() && other.is_constant() {
```
Pratyush: This uses the MSB <-> LSB relation? It would be good to document it here.

circuits/environment/src/lib.rs
```diff
@@ -79,3 +80,152 @@ pub mod prelude {
     pub use num_traits::{Inv, One as NumOne, Pow, Unsigned};
     pub use once_cell::unsync::OnceCell;
 }
+
+#[cfg(test)]
+pub use test_utilities::*;
```
howardwu: Can you move this back to integers? Would prefer to remove this altogether

circuits/types/field/src/compare.rs
```diff
@@ -19,27 +19,45 @@ use super::*;
...
+        } else {
+            // Initialize an iterator over `self` and `other` from MSB to LSB.
+            let self_bits_le = self.to_bits_le();
+            let other_bits_le = other.to_bits_le();
+            let bits_le = self_bits_le.iter().zip_eq(&other_bits_le);
+
+            bits_le.fold(Boolean::constant(false), |rest_is_less, (self_bit, other_bit)| {
+                Boolean::ternary(&self_bit.bitxor(other_bit), other_bit, &rest_is_less)
+            })
         }
```
howardwu: These 3 cases can just be folded into the last case.

circuits/environment/src/lib.rs
```diff
@@ -15,6 +15,7 @@
 // along with the snarkVM library. If not, see <https://www.gnu.org/licenses/>.
 
 #![forbid(unsafe_code)]
+#![allow(clippy::too_many_arguments)]
```
howardwu: Was there a specific case that required this clippy lint?

circuits/types/integers/src/compare.rs
```diff
@@ -88,6 +87,9 @@ mod tests {
         num_private: usize,
         num_constraints: usize,
     ) {
+        // Note that we need to use "fresh" circuits for each of the test cases, otherwise number of
+        // variables and constraints are incorrectly counted.
+
```
howardwu: What is this comment addressing specifically?

Circuit::scope creates a new environment each time

circuits/types/field/src/lib.rs
```diff
@@ -37,13 +37,13 @@ pub mod to_lower_bits;
...
-#[cfg(test)]
-use snarkvm_circuits_environment::assert_scope;
-
 use snarkvm_circuits_environment::prelude::*;
 use snarkvm_circuits_types_boolean::Boolean;
 use snarkvm_utilities::ToBits as TBits;
 
+#[cfg(test)]
+use snarkvm_circuits_environment::assert_scope;
+
```
howardwu: nit: move back please

circuits/types/integers/src/lib.rs
```diff
@@ -61,15 +61,15 @@ pub type U32<E> = Integer<E, u32>;
...
-#[cfg(test)]
-use snarkvm_circuits_environment::assert_scope;
-
 use snarkvm_circuits_environment::prelude::*;
 use snarkvm_circuits_types_boolean::Boolean;
 use snarkvm_circuits_types_field::Field;
 
 use core::marker::PhantomData;
 
+#[cfg(test)]
```
howardwu: nit: move this back

circuits/types/integers/src/lib.rs
```diff
@@ -95,9 +95,13 @@ impl<E: Environment, I: IntegerType> Inject for Integer<E, I> {
...
 
-// TODO (@pranav) Document
+///
+/// Casts an integer as its `dual` type.
+/// The `dual` type is the type that is the same size as the original type, with the opposite signedness.
+/// For example, the `dual` type of `I8` is `U8` and the `dual` type of `U8` is `I8`.
+///
 impl<E: Environment, I: IntegerType> Integer<E, I> {
-    fn cast_as_dual(self) -> Integer<E, I::Dual> {
+    pub(crate) fn cast_as_dual(self) -> Integer<E, I::Dual> {
```
howardwu: Why is this `pub(crate)` now?

circuits/types/field/src/from_bits.rs
```diff
@@ -15,11 +15,12 @@
 // along with the snarkVM library. If not, see <https://www.gnu.org/licenses/>.
 
 use super::*;
+use std::iter;
 
 impl<E: Environment> FromBits for Field<E> {
     type Boolean = Boolean<E>;
 
-    /// Initializes a new base field element from a list of little-endian bits *without* trailing zeros.
+    /// Initializes a new base field element from a list of little-endian bits.
```
howardwu, line 22: Please revert this comment

circuits/types/field/src/from_bits.rs
```diff
@@ -95,7 +94,7 @@ impl<E: Environment> FromBits for Field<E> {
         output
     }
 
-    /// Initializes a new base field element from a list of big-endian bits *without* leading zeros.
+    /// Initializes a new base field element from a list of big-endian bits.
```
howardwu, line 102: Please revert this comment

circuits/types/field/src/from_bits.rs
```diff
@@ -15,11 +15,12 @@
 // along with the snarkVM library. If not, see <https://www.gnu.org/licenses/>.
 
 use super::*;
+use std::iter;
```
howardwu: I believe you can use `core::iter::{your_module}` directly in the code

circuits/types/field/src/from_bits.rs
```diff
@@ -138,11 +138,12 @@ mod tests {
                 assert_eq!(expected_size_in_bits, candidate_bits.len());
                 assert_scope!(num_constants, num_public, num_private, num_constraints);
             });
+            Circuit::reset();
```
howardwu: nit: this reset is not necessary

circuits/types/field/src/from_bits.rs
```diff
@@ -155,6 +156,46 @@ mod tests {
                     }
                 };
             });
+            Circuit::reset();
```
howardwu: nit: this reset is not necessary

circuits/types/field/src/from_bits.rs
```diff
@@ -155,6 +156,46 @@ mod tests {
...
+                true => {
+                    let result = std::panic::catch_unwind(|| Field::<Circuit>::from_bits_le(&candidate));
+                    assert!(result.is_err());
+                }
+                _ => {
+                    Circuit::scope(format!("Excess One: {mode}"), || {
+                        let _ = Field::<Circuit>::from_bits_le(&candidate);
+                        assert_scope_fails!(num_constants, num_public, num_private + i, num_constraints + i + 1);
+                    });
+                    Circuit::reset();
```
howardwu, line 174: nit: this reset is not necessary

circuits/types/field/src/from_bits.rs
```diff
@@ -155,6 +156,46 @@ mod tests {
...
+            Circuit::reset();
+
+            // Add excess one bits.
+            let candidate = vec![given_bits, vec![Boolean::new(mode, true); i + 1]].concat();
+
+            match mode.is_constant() {
+                true => {
+                    let result = std::panic::catch_unwind(|| Field::<Circuit>::from_bits_le(&candidate));
+                    assert!(result.is_err());
+                }
```
howardwu: nit: you can one-line this

circuits/types/field/src/from_bits.rs
```diff
@@ -31,7 +32,45 @@ impl<E: Environment> FromBits for Field<E> {
             // Check if all excess bits are zero.
             let should_be_zero = bits_le[size_in_bits..].iter().fold(Boolean::constant(false), |acc, bit| acc | bit);
             // Ensure `should_be_zero` is zero.
-            E::assert_eq(E::zero(), should_be_zero);
+            // TODO: `E::assert_eq` does not enforce constant constraints.
```
howardwu: This is intentional behavior, remove the comment

circuits/types/field/src/from_bits.rs
```diff
@@ -31,7 +32,45 @@ impl<E: Environment> FromBits for Field<E> {
...
+            if should_be_zero.is_constant() && should_be_zero.eject_value() {
```
howardwu: nit: switch this to a `match` statement

circuits/types/field/src/from_bits.rs
```diff
@@ -31,7 +32,45 @@ impl<E: Environment> FromBits for Field<E> {
...
+                E::halt("The excess bits are not zero.")
```
howardwu:
```suggestion
                E::halt("Detected nonzero excess bits while initializing an integer from bits")
```

circuits/types/field/src/from_bits.rs
```diff
@@ -31,7 +32,45 @@ impl<E: Environment> FromBits for Field<E> {
...
+            // (For advanced users) BaseField::MODULUS - 1 is equivalent to -1 in the field.
+            let modulus_minus_one = -E::BaseField::one();
+            let modulus_minus_one_bits_le = modulus_minus_one.to_bits_le();
+
+            // Pad `bits_le` with zeros to the size of the base field modulus, if necessary.
+            let boolean_false = Boolean::constant(false);
+            let padded_bits_le = bits_le.iter().chain(iter::repeat(&boolean_false)).take(size_in_bits);
+
+            // Initialize an iterator over `self` and `other` from MSB to LSB.
+            let bit_pairs_le = modulus_minus_one_bits_le.iter().zip_eq(padded_bits_le);
```
howardwu:
```suggestion
            let bit_pairs_le = modulus_minus_one.to_bits_le().iter().zip_eq(padded_bits_le);
```

circuits/types/field/src/from_bits.rs
```diff
@@ -31,7 +32,45 @@ impl<E: Environment> FromBits for Field<E> {
...
+            // TODO: `E::assert_eq` does not enforce constant constraints.
+            if should_be_zero.is_constant() && should_be_zero.eject_value() {
+                E::halt("The excess bits are not zero.")
+            } else {
+                E::assert_eq(E::zero(), should_be_zero)
+            }
+        }
+
+        // If the number of bits is greater than `size_in_data_bits`, then check that it is a valid field element.
+        if num_bits > size_in_data_bits {
```
howardwu, line 48: If we have an `is_less_than` and `is_less_than_equal_to`, how come we're rewriting the logic here?

circuits/types/field/src/from_bits.rs
```diff
@@ -155,6 +156,46 @@ mod tests {
...
+            true => {
+                let result = std::panic::catch_unwind(|| Field::<Circuit>::from_bits_le(&modulus));
+                assert!(result.is_err());
+            }
+            _ => {
+                Circuit::scope(format!("Modulus: {mode}"), || {
+                    let _ = Field::<Circuit>::from_bits_le(&modulus);
+                    assert_scope_fails!(num_constants, num_public, num_private, num_constraints);
+                });
+                Circuit::reset();
```
howardwu, line 197: nit: this reset is not necessary

circuits/types/field/src/from_bits.rs
```diff
@@ -155,6 +156,46 @@ mod tests {
...
+            .iter()
+            .map(|b| Boolean::new(mode, *b))
+            .take(<Circuit as Environment>::BaseField::size_in_bits())
+            .collect::<Vec<_>>();
+
+        match mode.is_constant() {
+            true => {
+                let result = std::panic::catch_unwind(|| Field::<Circuit>::from_bits_le(&modulus));
+                assert!(result.is_err());
+            }
```
howardwu: nit: you can one-line this

circuits/types/integers/src/lib.rs
```diff
@@ -611,8 +620,24 @@ mod test_utilities {
         Circuit::reset();
     }
 
+    pub fn check_unary_operation_fails_without_counts<IN, OUT>(
```
howardwu: Remove this method and use the logic directly where it is needed.

circuits/types/scalar/src/compare.rs
```diff
@@ -15,29 +15,31 @@
...
-            if index != self.bits_le.len() - 1 {
-                are_previous_bits_equal &= self_bit.is_equal(other_bit);
-            }
+        debug_assert!(match E::ScalarField::modulus().to_bytes_le() {
+            Ok(modulus_bytes) => match E::BaseField::from_bytes_le(&modulus_bytes) {
+                Ok(modulus) => modulus.to_repr() <= E::BaseField::modulus(),
+                Err(error) => E::halt(format!("Failed to load the scalar modulus as a constant: {error}")),
+            },
+            Err(error) => E::halt(format!("Failed to retrieve the scalar modulus as bytes: {error}")),
+        });
```
howardwu: Remove this debug

howardwu: If you need to, you can look at `to_bits` for a 1 line sanity check

circuits/types/scalar/src/compare.rs
```diff
@@ -15,29 +15,31 @@
...
+
+        // If all elements of the scalar field are less than (p - 1)/2, where p is the modulus of
+        // the base field, then we can perform an optimized check for `less_than`.
+        // We compute the less than operation by checking the parity of 2 * (self - other) mod p.
+        // If a < b, then 2 * (self - other) mod p is odd.
+        // If a >= b, then 2 * (self - other) mod p is even.
+        if self.is_constant() && other.is_constant() {
+            Boolean::new(Mode::Constant, self.eject_value() < other.eject_value())
+        } else {
+            (self.to_field() - other.to_field()).double().to_bits_be().pop().unwrap()
```
howardwu: nit: usage of `.unwrap()` here should be handled

circuits/types/scalar/src/compare.rs
```diff
@@ -64,66 +66,103 @@ mod tests {
...
+
+            // Check `is_less_than`.
+            let a = Scalar::<Circuit>::new(mode_a, first);
+            let b = Scalar::<Circuit>::new(mode_b, second);
+            Circuit::scope(&format!("Less Than: {} {}", mode_a, mode_b), || {
+                let candidate = a.is_less_than(&b);
+                assert_eq!(first < second, candidate.eject_value());
+                assert_scope!(num_constants, num_public, num_private, num_constraints);
+            });
+            Circuit::reset();
```
howardwu: nit: reset is not necessary here

circuits/types/scalar/src/compare.rs
```diff
@@ -64,66 +66,103 @@ mod tests {
...
+
+            // Check `is_less_than_or_equal`
+            let a = Scalar::<Circuit>::new(mode_a, first);
+            let b = Scalar::<Circuit>::new(mode_b, second);
+            Circuit::scope(&format!("Less Than Or Equal: {} {}", mode_a, mode_b), || {
+                let candidate = a.is_less_than_or_equal(&b);
+                assert_eq!(first <= second, candidate.eject_value());
+                assert_scope!(num_constants, num_public, num_private, num_constraints);
+            });
+            Circuit::reset();
```
howardwu: nit: reset is not necessary here

circuits/types/scalar/src/compare.rs
```diff
@@ -64,66 +66,103 @@ mod tests {
...
+
+            // Check `is_greater_than`
+            let a = Scalar::<Circuit>::new(mode_a, first);
+            let b = Scalar::<Circuit>::new(mode_b, second);
+            Circuit::scope(&format!("Greater Than: {} {}", mode_a, mode_b), || {
+                let candidate = a.is_greater_than(&b);
+                assert_eq!(first > second, candidate.eject_value());
+                assert_scope!(num_constants, num_public, num_private, num_constraints);
+            });
+            Circuit::reset();
```
howardwu: nit: reset is not necessary here

circuits/types/scalar/src/compare.rs
```diff
@@ -64,66 +66,103 @@ mod tests {
...
+
+            // Check `is_greater_than_or_equal`
+            let a = Scalar::<Circuit>::new(mode_a, first);
+            let b = Scalar::<Circuit>::new(mode_b, second);
+            Circuit::scope(&format!("Greater Than Or Equal: {} {}", mode_a, mode_b), || {
+                let candidate = a.is_greater_than_or_equal(&b);
+                assert_eq!(first >= second, candidate.eject_value());
                 assert_scope!(num_constants, num_public, num_private, num_constraints);
             });
             Circuit::reset();
```
howardwu: nit: reset is not necessary here

circuits/types/scalar/src/from_bits.rs
```diff
@@ -15,14 +15,15 @@
...
 use super::*;
-use snarkvm_utilities::{FromBytes, ToBytes};
+
+use std::iter;
 
 impl<E: Environment> FromBits for Scalar<E> {
     type Boolean = Boolean<E>;
 
-    /// Initializes a new scalar field element from a list of little-endian bits *without* trailing zeros.
+    /// Initializes a new scalar field element from a list of little-endian bits.
```
howardwu, line 24: Look at the field comments to mirror please

circuits/types/field/src/compare.rs
```diff
@@ -21,25 +21,46 @@ impl<E: Environment> Compare<Field<E>> for Field<E> {
...
+        if self.is_constant() && other.is_constant() {
+            Boolean::constant(self.eject_value() < other.eject_value())
+        } else if self.is_constant() {
+            // For advanced users: this implementation saves us from instantiating 253 constants for
+            // the bits of `self`.
+            let self_bits_le = self.eject_value().to_bits_le();
+
+            let bit_pairs_le = self_bits_le.into_iter().zip_eq(other.to_bits_le());
+
+            bit_pairs_le.fold(Boolean::constant(false), |rest_is_less, (self_bit, other_bit)| {
```
howardwu: Unify this into 1 line

circuits/types/field/src/compare.rs
```diff
@@ -21,25 +21,46 @@ impl<E: Environment> Compare<Field<E>> for Field<E> {
...
+                }
+            })
+        } else if other.is_constant() {
+            // For advanced users: this implementation saves us from instantiating 253 constants for
+            // the bits of `self`.
+            let self_bits_le = self.to_bits_le();
+
+            let bit_pairs_le = self_bits_le.iter().zip_eq(other.eject_value().to_bits_le());
+
+            bit_pairs_le.fold(Boolean::constant(false), |rest_is_less, (self_bit, other_bit)| {
```
howardwu: Unify this into 1 line

circuits/types/field/src/compare.rs
```diff
@@ -21,25 +21,46 @@ impl<E: Environment> Compare<Field<E>> for Field<E> {
...
+                    Boolean::ternary(self_bit, &!self_bit, &rest_is_less)
+                }
+            })
+        } else {
+            // Initialize an iterator over `self` and `other` from MSB to LSB.
+            let self_bits_le = self.to_bits_le();
+            let other_bits_le = other.to_bits_le();
+            let bits_le = self_bits_le.iter().zip_eq(&other_bits_le);
+
+            bits_le.fold(Boolean::constant(false), |rest_is_less, (self_bit, other_bit)| {
```
howardwu: Unify this into 1 line

circuits/types/field/src/compare.rs
```diff
@@ -21,25 +21,46 @@ impl<E: Environment> Compare<Field<E>> for Field<E> {
...
-
-            // Skip the update to the LSB, as this boolean is subsequently discarded.
-            if index != self_bits_be.len() - 1 {
-                are_previous_bits_equal &= self_bit.is_equal(other_bit);
-            }
+        if self.is_constant() && other.is_constant() {
+            Boolean::constant(self.eject_value() < other.eject_value())
+        } else if self.is_constant() {
+            // For advanced users: this implementation saves us from instantiating 253 constants for
+            // the bits of `self`.
```
howardwu: Why does this save you 253 constants? Documentation isn't clear

circuits/types/field/src/compare.rs
```diff
@@ -64,68 +85,105 @@ mod tests {
...
+
+            // Check `is_less_than`.
+            let a = Field::<Circuit>::new(mode_a, first);
+            let b = Field::<Circuit>::new(mode_b, second);
+            Circuit::scope(&format!("Less Than: {} {}", mode_a, mode_b), || {
+                let candidate = a.is_less_than(&b);
+                assert_eq!(first < second, candidate.eject_value());
+                assert_scope!(num_constants, num_public, num_private, num_constraints);
+            });
+            Circuit::reset();
```
howardwu: nit: the reset can be removed

circuits/types/field/src/compare.rs
```diff
@@ -64,68 +85,105 @@ mod tests {
...
+
+            // Check `is_less_than_or_equal`
+            let a = Field::<Circuit>::new(mode_a, first);
+            let b = Field::<Circuit>::new(mode_b, second);
+            Circuit::scope(&format!("Less Than Or Equal: {} {}", mode_a, mode_b), || {
+                let candidate = a.is_less_than_or_equal(&b);
+                assert_eq!(first <= second, candidate.eject_value());
+                assert_scope!(num_constants, num_public, num_private, num_constraints);
+            });
+            Circuit::reset();
```
howardwu: nit: the reset can be removed

circuits/types/field/src/compare.rs
```diff
@@ -64,68 +85,105 @@ mod tests {
...
+
+            // Check `is_greater_than`
+            let a = Field::<Circuit>::new(mode_a, first);
+            let b = Field::<Circuit>::new(mode_b, second);
+            Circuit::scope(&format!("Greater Than: {} {}", mode_a, mode_b), || {
+                let candidate = a.is_greater_than(&b);
+                assert_eq!(first > second, candidate.eject_value());
+                assert_scope!(num_constants, num_public, num_private, num_constraints);
+            });
+            Circuit::reset();
```
howardwu: nit: the reset can be removed

circuits/types/field/src/compare.rs
```diff
@@ -64,68 +85,105 @@ mod tests {
...
+
+            // Check `is_greater_than_or_equal`
+            let a = Field::<Circuit>::new(mode_a, first);
+            let b = Field::<Circuit>::new(mode_b, second);
+            Circuit::scope(&format!("Greater Than Or Equal: {} {}", mode_a, mode_b), || {
+                let candidate = a.is_greater_than_or_equal(&b);
+                assert_eq!(first >= second, candidate.eject_value());
                 assert_scope!(num_constants, num_public, num_private, num_constraints);
             });
             Circuit::reset();
```
howardwu: nit: the reset can be removed

circuits/types/field/src/from_bits.rs
d0cd: I believe you added support for trailing zeros in this commit: https://github.com/AleoHQ/snarkVM/commit/ecce1aa57b50d48395d6db49a260e81b7dc681b0

circuits/types/field/src/from_bits.rs
d0cd: I believe you added support for leading zeros in this commit: https://github.com/AleoHQ/snarkVM/commit/ecce1aa57b50d48395d6db49a260e81b7dc681b0

circuits/types/field/src/from_bits.rs
howardwu: Sure, can you update it to enumerate the cases that this method supports then?

circuits/types/field/src/from_bits.rs
howardwu: Sure, can you update it to enumerate the cases that this method supports then?

circuits/types/field/src/from_bits.rs
d0cd: What was the rationale for making it intentional?
Personally, I've run into this problem a few times while writing circuits.
I figure some sort of warning would help prevent errors.

Pratyush:
We could make assert_eq call an actual assert in the constant case. That would catch bugs, and wouldn’t harm correct execution by adding constraints.

howardwu:
> We could make assert_eq call an actual assert in the constant case. That would catch bugs, and wouldn’t harm correct execution by adding constraints.

I have a `TODO` in the `Circuit::assert_eq` method to contemplate the cases here.

The one I'm unsure about is for `SNARK::setup` usage. Are there any cases we can think of where dummy cases would cause it to false-positive error?

circuits/types/field/src/from_bits.rs
d0cd: `Field::is_less_than` is called on `Field` circuits. 
`Field::is_less_than` invokes `to_bits` which produces 254 constraints.
Since we already have `bits_le`, we shouldn't pay for redundant computation.

circuits/types/scalar/src/compare.rs
d0cd: There was a bug in this debug assert.
The line `Ok(modulus) => modulus.to_repr() <= E::BaseField::modulus()`, should actually read 
`Ok(modulus) => modulus.to_repr() <= E::BaseField::modulus_minus_one_div_two()`.

I believe the 1 line sanity check you are referring to is in `to_field`:
`debug_assert!(E::ScalarField::size_in_bits() < E::BaseField::size_in_bits());`

It is worth noting that `modulus.to_repr() <= E::BaseField::modulus_minus_one_div_two()` and `E::ScalarField::size_in_bits() < E::BaseField::size_in_bits()` are not equivalent.

That being said, would you still like the `debug_assert` to be removed?

circuits/environment/src/lib.rs
d0cd: Used for `test_utilities`. Removed.

circuits/types/field/src/lib.rs
d0cd: This is used for tests.
Introduced in https://github.com/AleoHQ/snarkVM/commit/ee2a3f98ff422cab5a8ff8e113dd30667356792d

circuits/types/integers/src/compare.rs
d0cd: Removing comment.

circuits/types/integers/src/lib.rs
d0cd: Mix-up. Removing.

circuits/types/field/src/compare.rs
```diff
@@ -21,25 +21,52 @@ impl<E: Environment> Compare<Field<E>> for Field<E> {
...
+        } else if self.is_constant() {
+            // (For advanced users) this implementation saves us from instantiating 253 constants for
+            // the bits of `self`. The implementation in the `else` case invokes `to_bits_le` on
+            // `self` which would allocate 253 constants. Since `self` is constant, we can directly
+            // inspect its bits and construct an equivalent but reduced ternary expression.
+            self.eject_value().to_bits_le().into_iter().zip_eq(other.to_bits_le()).fold(
+                Boolean::constant(false),
+                |rest_is_less, (self_bit, other_bit)| {
+                    if self_bit {
+                        Boolean::ternary(&!&other_bit, &other_bit, &rest_is_less)
```
Pratyush: This expression is a bit difficult to read; might be worth explaining what is going on here.

circuits/types/field/src/compare.rs
```diff
@@ -21,25 +21,52 @@ impl<E: Environment> Compare<Field<E>> for Field<E> {
...
+                    } else {
+                        Boolean::ternary(&other_bit, &other_bit, &rest_is_less)
+                    }
+                },
+            )
+        } else if other.is_constant() {
+            // (For advanced users) this implementation saves us from instantiating 253 constants for
```
Pratyush: We can just call `other.is_less_than(&self)` and avoid duplicating the logic.

circuits/types/field/src/compare.rs
```diff
@@ -21,25 +21,52 @@ impl<E: Environment> Compare<Field<E>> for Field<E> {
...
+            // Zip `self.to_bits_le()` and `other.to_bits_le()` together and construct a check on the sequence of bit pairs.
+            // The bitwise check begins with the most-significant bit pair and ends with the least-significant bit pair.
+            // For each bit pair,
+            // - If `self_bit` and `other_bit` are different signs, then if `self_bit` is `true`, return false.
+            // - If `self_bit` and `other_bit` are different signs, then if `self_bit` is `false`, return true.
+            // - If `self_bit` and `other_bit` are the same sign, then check the following bits.
+            self.to_bits_le().iter().zip_eq(other.to_bits_le()).fold(
+                Boolean::constant(false),
+                |rest_is_less, (self_bit, other_bit)| {
+                    Boolean::ternary(&self_bit.bitxor(&other_bit), &other_bit, &rest_is_less)
```
Pratyush, line 65: Can we combine the ternary with the bitxor here somehow? By, say, writing out the truth table. That would save some constraints.

circuits/types/field/src/from_bits.rs
Pratyush: Might be worth extracting out a common helper function to avoid duplication.

circuits/types/field/src/from_bits.rs
```diff
@@ -31,7 +35,47 @@ impl<E: Environment> FromBits for Field<E> {
             // Check if all excess bits are zero.
             let should_be_zero = bits_le[size_in_bits..].iter().fold(Boolean::constant(false), |acc, bit| acc | bit);
             // Ensure `should_be_zero` is zero.
-            E::assert_eq(E::zero(), should_be_zero);
+            match should_be_zero.is_constant() {
```
Pratyush: Maybe make this is a `match (should_be_zero().is_constant(), should_be_zero.eject_value()) { ... `?

Not sure if that would help with readability

circuits/types/field/src/from_bits.rs
```diff
@@ -31,7 +35,47 @@ impl<E: Environment> FromBits for Field<E> {
...
+        }
+
+        // If the number of bits is greater than `size_in_data_bits`, then check that it is a valid field element.
+        if num_bits > size_in_data_bits {
+            // Retrieve the modulus & subtract by 1 as we'll check `bits_le` is less than or *equal* to this value.
+            // (For advanced users) BaseField::MODULUS - 1 is equivalent to -1 in the field.
+            let modulus_minus_one = -E::BaseField::one();
+
+            // Pad `bits_le` with zeros to the size of the base field modulus, if necessary.
+            let boolean_false = Boolean::constant(false);
```
Pratyush, line 54: It might be worth it to add a constant function `Boolean::false()` that returns `Boolean::constant(false)`

circuits/types/scalar/src/compare.rs
```diff
@@ -15,29 +15,34 @@
...
+
+        // If all elements of the scalar field are less than (p - 1)/2, where p is the modulus of
+        // the base field, then we can perform an optimized check for `less_than`.
+        // We compute the less than operation by checking the parity of 2 * (self - other) mod p.
+        // If a < b, then 2 * (self - other) mod p is odd.
+        // If a >= b, then 2 * (self - other) mod p is even.
+        if self.is_constant() && other.is_constant() {
+            Boolean::new(Mode::Constant, self.eject_value() < other.eject_value())
+        } else {
+            match (self.to_field() - other.to_field()).double().to_bits_be().pop() {
```
Pratyush: We can write this as `(self.to_field() ...).unwrap_or_else(|| E::halt("..."))`

circuits/types/scalar/src/compare.rs
```diff
@@ -64,66 +69,94 @@ mod tests {
 
     const ITERATIONS: usize = 100;
 
-    fn check_is_less_than(
+    #[rustfmt::skip]
```
Pratyush: Do we intend to skip this?

circuits/types/scalar/src/from_bits.rs
```diff
@@ -32,40 +37,65 @@ impl<E: Environment> FromBits for Scalar<E> {
...
             // Ensure `should_be_zero` is zero.
-            E::assert_eq(E::zero(), should_be_zero);
+            match should_be_zero.is_constant() {
+                true => match should_be_zero.eject_value() {
+                    false => (), // Constraint is satisfied.
+                    true => {
+                        E::halt("Detected nonzero excess bits while initializing a scalar field element from bits.")
+                    }
+                },
+                false => E::assert_eq(E::zero(), should_be_zero),
```
Pratyush:
```suggestion
                false => E::assert(!should_be_zero),
```

circuits/types/scalar/src/from_bits.rs
```diff
@@ -97,17 +127,18 @@ mod tests {
...
+            Circuit::scope(format!("{mode} {i}"), || {
                 let candidate = Scalar::<Circuit>::from_bits_le(&given_bits);
                 assert_eq!(expected, candidate.eject_value());
                 assert_eq!(expected_size_in_bits, candidate.bits_le.len());
                 assert_scope!(num_constants, num_public, num_private, num_constraints);
             });
 
             // Add excess zero bits.
-            let candidate = vec![given_bits, vec![Boolean::new(mode, false); i]].concat();
+            let candidate = vec![given_bits.clone(), vec![Boolean::new(mode, false); i]].concat();
```
Pratyush:
```suggestion
            let candidate = [given_bits.clone(), vec![Boolean::new(mode, false); i]].concat();
```

circuits/types/scalar/src/from_bits.rs
```diff
@@ -159,36 +222,68 @@ mod tests {
                     }
                 };
             });
+
+            // Add excess one bits.
+            let candidate = vec![vec![Boolean::new(mode, true); i + 1], given_bits].concat();
```
Pratyush:
```suggestion
            let candidate = [vec![Boolean::new(mode, true); i + 1], given_bits].concat();
```

Pratyush:
> I have a `TODO` in the `Circuit::assert_eq` method to contemplate the cases here.
> 
> The one I'm unsure about is for `SNARK::setup` usage. Are there any cases we can think of where dummy cases would cause it to false-positive error?

I think we want to catch those edge cases anyway, and decide how to handle them.

circuits/types/field/src/from_bits.rs
howardwu: Unfortunately, `false` is a reserved keyword in Rust

circuits/types/field/src/from_bits.rs
howardwu: I would avoid the helper functions altogether. It was a big factor in contributing to `gadgets/`' readability issues

circuits/types/field/src/from_bits.rs
Pratyush: we could try `fn True` and `fn False` (slightly unorthodox naming, but IMO clearer than `Boolean::constant(true)` and `Boolean::constant(false)`)

howardwu:
> I think we want to catch those edge cases anyway, and decide how to handle them.

I'll prepare a PR this week to introduce this check.

circuits/types/scalar/src/compare.rs
howardwu: You can do the following:
```rs
debug_assert!(E::ScalarField::modulus() < E::BaseField::modulus_minus_one_div_two());
```
and in `environment.rs`, change line 27 to read:
```rs
type ScalarField: PrimeField<BigInteger = <Self::BaseField as PrimeField>::BigInteger> + Copy;
```

howardwu: This is technically safer anyways, since your version decomposes into bytes and recomposes without assumptions. This version will catch size mismatches should one change curves in the future.

circuits/types/scalar/src/compare.rs
d0cd: Removed.

circuits/types/field/src/from_bits.rs
d0cd: Could we push this discussion to a new issue?
Edit: ^Noted to consider in the future.

circuits/types/field/src/compare.rs
d0cd: I'm not sure if there is a reduction for the ternary and bitxor.
I checked with a boolean function minimizer and the minimal form uses 3 AND gates and 2 OR gates.
I did minimize the logic for the cases where exactly one of the operands is constant.

d0cd:
Turns out `Circuit::reset` can be useful while testing.
It initializes a new instance of `R1CS`, effectively clearing scopes created by previous iterations of tests.
This is useful during CI since the machines are capped in size.
If you do not use `Circuit::reset`, you may have to reduce the number of iterations run by tests so as to avoid an OOM issue.

howardwu:
> Turns out `Circuit::reset` can be useful while testing. It initializes a new instance of `R1CS`, effectively clearing scopes created by previous iterations of tests. This is useful during CI since the machines are capped in size. If you do not use `Circuit::reset`, you may have to reduce the number of iterations run by tests so as to avoid an OOM issue.

For the cases where reset is necessary, make sure to do a double pass on whether there is expensive logic that can be sped up. Often there are ways to improve the performance so that it is no longer necessary, a benefit that also translates to real runtime environments for programs. Often, the `eject` may be the expensive operation, which can either 1) be reduced in usage, and 2) cached to improve performance.

Lastly, try to minimize usage of reset as much as possible. We don't want to bury the symptoms of not scaling only to face it in consensus.

circuits/types/field/src/lib.rs
howardwu: nit: move back please

circuits/types/integers/src/lib.rs
howardwu: nit: move back please

circuits/types/integers/src/lib.rs
howardwu: Why is this pub(crate) now?

d0cd: It shouldn't be.
Will remove.

d0cd:
> > Turns out `Circuit::reset` can be useful while testing. It initializes a new instance of `R1CS`, effectively clearing scopes created by previous iterations of tests. This is useful during CI since the machines are capped in size. If you do not use `Circuit::reset`, you may have to reduce the number of iterations run by tests so as to avoid an OOM issue.
> 
> For the cases where reset is necessary, make sure to do a double pass on whether there is expensive logic that can be sped up. Often there are ways to improve the performance so that it is no longer necessary, a benefit that also translates to real runtime environments for programs. Often, the `eject` may be the expensive operation, which can either 1) be reduced in usage, and 2) cached to improve performance.
> 
> Lastly, try to minimize usage of reset as much as possible. We don't want to bury the symptoms of not scaling only to face it in consensus.

To clarify, is `eject` an expensive operation in runtime or memory?

PR #712, opened 2022-03-22
Introduces a `witness!` macro, updates circuit traits and linear combination

howardwu:
## Motivation

This PR updates the circuits environment design to introduce better support for witnesses.
- Introduces a `witness!` macro that takes as input any circuit that implements `Eject` returning circuit primitive(s) that support `Inject` to automatically convert to and from the circuit environment.
- Adds support for `Vec` and `Tuples` with the `Inject` trait to auto inject multiple primitives.
- Updates the linear combination design to use an `IndexMap` for better performance characteristics
- Introduces a `TypeName` trait to move `fn type_name()` out of `Inject`

This PR also adds additional benchmarks for linear combination and adds 1 benchmark for Boolean.


circuits/environment/src/traits/inject.rs
```diff
@@ -0,0 +1,107 @@
...
+    #[inline]
+    fn new(mode: Mode, value: Self::Primitive) -> Self {
+        value.into_iter().map(|v| C::new(mode, v)).collect()
+    }
+}
+
+/********************/
+/****** Tuples ******/
+/********************/
+
```
d0cd, line 66: This could be implemented using macros for larger sequences of tuples.

d0cd (approved):
LGTM overall.
Consider using a macro for implementing inject for tuples.

PR #715, opened 2022-03-23
Refactor PC Scheme infrastructure to enable future optimizations

Pratyush:
## Motivation

This PR builds upon #706 to simplify the PC scheme implementation. In the process of doing so, it also speeds up things like signature verification as well as the value commitment handling.

Speed ups for signature operations

## Test Plan

(Write your test plan here)

## Related PRs

(Link your related PRs here)


algorithms/src/polycommit/sonic_pc/data_structures.rs
```diff
@@ -477,5 +488,279 @@ impl<E: PairingEngine> Prepare<PreparedVerifierKey<E>> for VerifierKey<E> {
...
+/// That is, if `(label, query)` is an element of `Q`, then `evaluation.get((label, query))`
+/// should equal `p[label].evaluate(query)`.
+pub type Evaluations<'a, F> = BTreeMap<(String, F), F>;
+
+/// Evaluate the given polynomials at `query_set`.
+pub fn evaluate_query_set<'a, F: PrimeField>(
+    polys: impl IntoIterator<Item = &'a LabeledPolynomial<F>>,
+    query_set: &QuerySet<'a, F>,
+) -> Evaluations<'a, F> {
+    let polys: BTreeMap<_, _> = polys.into_iter().map(|p| (p.label(), p)).collect();
```
ljedrz: does this have to be a `BTreeMap`? lookups are much faster in a `HashMap` and it doesn't seem like ordering is of importance here

algorithms/src/polycommit/sonic_pc/mod.rs
```diff
@@ -338,91 +306,102 @@ impl<E: PairingEngine> PolynomialCommitment<E::Fr, E::Fq> for SonicKZG10<E> {
...
+        commitments: impl IntoIterator<Item = &'a LabeledCommitment<Commitment<E>>>,
+        query_set: &QuerySet<E::Fr>,
+        rands: impl IntoIterator<Item = &'a Randomness<E>>,
+        fs_rng: &mut S,
+    ) -> Result<BatchProof<E>, PCError>
+    where
+        Randomness<E>: 'a,
+        Commitment<E>: 'a,
+    {
+        let poly_rand_comm: BTreeMap<_, _> = labeled_polynomials
```
ljedrz: same as before: if the collection is not iterated over (or a specific order is expected for any other reason), a `HashMap` will be faster (~`O(1)` as opposed to `O(log(n))`) for lookup purposes

algorithms/src/polycommit/sonic_pc/mod.rs
```diff
@@ -432,15 +411,24 @@ impl<E: PairingEngine> PolynomialCommitment<E::Fr, E::Fq> for SonicKZG10<E> {
...
 
-        let mut combined_comms: BTreeMap<Option<usize>, E::G1Projective> = BTreeMap::new();
-        let mut combined_witness: E::G1Projective = E::G1Projective::zero();
-        let mut combined_adjusted_witness: E::G1Projective = E::G1Projective::zero();
+        let mut combined_comms = BTreeMap::new();
+        let mut combined_witness = E::G1Projective::zero();
+        let mut combined_adjusted_witness = E::G1Projective::zero();
+
+        let mut batch_kzg_check_fs_rng = S::new();
+        batch_kzg_check_fs_rng.absorb_nonnative_field_elements(
```
ljedrz: It seems like `absorb_nonnative_field_elements` only needs to iterate over its first argument (I checked the definition); it would avoid the vector allocations at the callsites

algorithms/src/polycommit/sonic_pc/mod.rs
```diff
@@ -767,221 +605,54 @@ impl<E: PairingEngine> PolynomialCommitment<E::Fr, E::Fq> for SonicKZG10<E> {
...
-
-                comms.push(commitment);
-                values.push(*v_i);
-            }
-
-            let proof_time = start_timer!(|| "Checking per-query proof");
-            result &= Self::check_individual_opening_challenges(vk, comms, point, values, &proof, opening_challenges)?;
-            end_timer!(proof_time);
+    fn normalize_commitments(commitments: Vec<E::G1Projective>) -> impl Iterator<Item = Commitment<E>> {
+        let mut comms = Vec::with_capacity(commitments.len());
```
ljedrz: doesn't this vector contain the same contents as `commitments`? if I'm not missing anything, `comms` can be removed and `commitments` used directly

algorithms/src/snark/marlin/marlin.rs
```diff
@@ -353,87 +349,89 @@ impl<
...
-                lc_s.iter().find(|lc| &lc.label == label).ok_or_else(|| AHPError::MissingEval(label.to_string()))?;
-            let evaluation = polynomials.get_lc_eval(lc, *point)?;
-            if !AHPForR1CS::<TargetField, MM>::LC_WITH_ZERO_EVAL.contains(&lc.label.as_ref()) {
-                evaluations_unsorted.push((label.to_string(), evaluation));
+        let mut evaluations = std::collections::BTreeMap::new();
+        for (label, (_, point)) in query_set.to_set() {
+            if !AHPForR1CS::<E::Fr, MM>::LC_WITH_ZERO_EVAL.contains(&label.as_str()) {
+                let lc = lc_s
+                    .iter()
+                    .find(|lc| *lc.label == label)
```
ljedrz: this looks like linear search; perhaps `lc_s` can be made into a `HashMap` for faster lookups, especially since these happen in a loop?

algorithms/src/snark/marlin/ahp/prover/round_functions/first.rs
```diff
@@ -0,0 +1,214 @@
...
+    ) -> Result<(prover::FirstOracles<'a, F>, prover::State<'a, F, MM>), AHPError> {
+        let round_time = start_timer!(|| "AHP::Prover::FirstRound");
+        let constraint_domain = state.constraint_domain;
+        let zk_bound = state.zk_bound;
+
+        let mut job_pool = snarkvm_utilities::ExecutionPool::with_capacity(3);
+        let z_a = state.z_a.take().unwrap();
+        let z_b = state.z_b.take().unwrap();
+        let mut private_variables = vec![];
+        core::mem::swap(&mut state.private_variables, &mut private_variables);
```
ljedrz: nit: `mem::take` does the same thing (as long as there are no `Default` impl issues)

algorithms/src/polycommit/sonic_pc/data_structures.rs
Pratyush: Hm ordering is indeed not important here. Happy to change it to a HashMap, but generally the size of this Map is going to be quite small (20-30 items *at most*), so I don't think it will have a big perf impact.

algorithms/src/polycommit/sonic_pc/mod.rs
Pratyush: (Same comment applies here too)

algorithms/src/polycommit/sonic_pc/mod.rs
Pratyush: The suggestion is to change the argument to `impl IntoIterator<Item = F>`? I think that makes sense.

algorithms/src/polycommit/sonic_pc/mod.rs
Pratyush: Oh good point.

algorithms/src/snark/marlin/ahp/prover/round_functions/first.rs
Pratyush:
```suggestion
        let private_variables = core::mem::take(&mut state.private_variables);
```

algorithms/src/snark/marlin/marlin.rs
Pratyush: Will do!

algorithms/src/polycommit/sonic_pc/mod.rs
ljedrz: yes, that or `impl Iterator<Item = F>` or `I: Iterator<Item = F>`

algorithms/src/polycommit/sonic_pc/data_structures.rs
ljedrz: if it's that small, then a vector would probably be fine too

algorithms/src/polycommit/sonic_pc/mod.rs
ljedrz: what about the number of iterations in the enclosing loop? is it also expected to be small?

Pratyush:
This PR is ready for review!

howardwu (approved):

PR #720, opened 2022-03-26
Implements `Signature` circuit, `Poseidon::hash_to_scalar`, and variety of traits for `Address`, `Field`, and `Literal`

howardwu:
## Motivation

- Implements a `Signature` circuit with a `Signature::verify()` method
- [Adds `Poseidon::hash_to_scalar` matching the impl for native signature operations](https://github.com/AleoHQ/snarkVM/pull/720/commits/b93cd2e35a5c6817a93d7cbe047c7d2993933005)
- [Adds `to_x_coordinate` and `to_y_coordinate` for `Group`](https://github.com/AleoHQ/snarkVM/pull/720/commits/ceb71e20cf257c956bb5a8818d1c8d82ae91dafd)
- Adds support for excess bits in `FromBits` impls on `Field`, `Integer`, and `Scalar` (with tests)
- Implements `ToBits` for `Literal`
- Implements `FromBoolean` for `Field`
- Implements `Equal`, `ToBits`, `ToGroup` for `Address`
- Removes the `DataType` trait

### Resampling Parameters

Commit [4d51d6b](https://github.com/AleoHQ/snarkVM/pull/720/commits/4d51d6bb9d56e253cb7a46c3ae3189a4d79a668e) requires resampling parameters. A synopsis of the changes are:
1. Changes signature messages from `[u8]` to `[bool]`
2. Implements the `Signature::verify()` circuit and confirms it matches native
3. Fixes a Poseidon PRF message length bug

Note that all `algorithms/` and `gadgets/` tests pass in that commit.

PR #721, opened 2022-03-27
Implements `Scalar::add()` circuit

howardwu:
## Motivation

#721 builds on #720

- Implements `Scalar::add()` circuit, with tests


circuits/types/scalar/src/add.rs
```diff
@@ -0,0 +1,200 @@
...
+    fn add_assign(&mut self, other: &Scalar<E>) {
+        // Determine the variable mode.
+        if self.is_constant() && other.is_constant() {
+            // Compute the sum and set the new constant in `self`.
+            *self = witness!(|self, other| self + other);
+        } else {
+            // Instead of adding the bits of `self` and `other` directly, the scalars are
+            // converted into a field elements, and summed, before converting back to scalars.
+            // Note: This is safe as the base field is larger than the scalar field.
+            let sum = self.to_field() + other.to_field();
```
d0cd: Is it always case that the addition of two `Scalar`s in the base field will never wrap around the base field modulus?

d0cd (approved):
LGTM.
PR #711 should also reduce the number of constraints produced for this operation.

circuits/types/scalar/src/add.rs
howardwu: For our case, yes

PR #722, opened 2022-03-27
Refactors `Literal` and `core/`

howardwu:
## Motivation

Refactors `Literal` and `core/`.
- Adds a `Primitive` enum for `Literal` to become circuit-compliant (implementing `Inject` and `Eject`)
- Moves `Literal` from `types/` into `core`
- Creates a `account/` and `algorithms/` folder in `core/`


circuits/environment/src/traits/to_bits.rs
```diff
@@ -69,12 +73,36 @@ impl<C: ToBits<Boolean = B>, B: BooleanTrait> ToBits for &[C] {
...
+        // The slice is order-preserving, meaning the first circuit in is the first circuit bits out.
         self.iter().flat_map(|c| c.to_bits_be()).collect()
     }
 }
+
+/********************/
+/****** Tuples ******/
+/********************/
+
+impl<'a, C0: ToBits<Boolean = B>, C1: ToBits<Boolean = B>, B: BooleanTrait> ToBits for (&'a C0, &'a C1) {
```
d0cd, line 92: Consider macros.

d0cd (approved):
LGTM.
One comment I have is that the move of `Literal` to `core` might motivate a clearer name.
The definition of literal (from wiki) is: "a notation for representing a fixed in source code", but we seem to be using it for more than that.
One suggestion is that we could rename `Literal` and `Primitive` to `CircuitPrimitive` and `NativePrimitive` respectively.

PR #727, opened 2022-03-28
Add Pedersen gadget

jules:


circuits/core/src/pedersen/hash.rs
```diff
@@ -0,0 +1,140 @@
...
+            // Ensure the input size is within the parameter size,
+            false => E::halt("incorrect input length for pedersen hash"),
+        }
+
+        // Compute sum of h_i^{m_i} for all i.
+        input
+            .chunks(WINDOW_SIZE)
+            .zip_eq(&self.bases)
+            .flat_map(|(bits, powers)| {
+                bits.iter().zip_eq(powers).flat_map(|(bit, base)| match bit.eject_value() {
```
howardwu: This should be a ternary, since the hash is hardcoded in the circuit otherwise.

howardwu (changes requested):

circuits/core/src/pedersen/hash.rs
```diff
@@ -0,0 +1,140 @@
...
+
+    const ITERATIONS: usize = 10;
+    const WINDOW_SIZE_MULTIPLIER: usize = 8;
+
+    fn check_hash<const NUM_WINDOWS: usize, const WINDOW_SIZE: usize>(mode: Mode, message: &str, input_value: &[bool]) {
+        let native_hasher = PedersenCRH::<EdwardsProjective, { NUM_WINDOWS }, { WINDOW_SIZE }>::setup(message);
+        let circuit_hasher = Pedersen::<_, { NUM_WINDOWS }, { WINDOW_SIZE }>::setup(message);
+
+        // Check for equivalency of bases.
+        native_hasher.parameters().iter().zip(circuit_hasher.parameters().iter()).for_each(
```
howardwu: Let's move this to a `check_setup` and `test_setup_constant` function in `mod.rs`

circuits/core/src/pedersen/hash.rs
```diff
@@ -0,0 +1,140 @@
...
+                native_bases.iter().zip(circuit_bases.iter()).for_each(|(native_base, circuit_base)| {
+                    assert_eq!(native_base.into_affine(), circuit_base.eject_value())
+                })
+            },
+        );
+
+        for i in 0..ITERATIONS {
+            let native_hash: EdwardsAffine = native_hasher.hash(input_value).expect("should be able to hash input");
+            Circuit::scope(format!("Pedersen {mode} {i}"), || {
+                let circuit_input =
```
howardwu: This should be moved out of `scope` otherwise the constraint counts are incorrect (as they will factor these into the count)

circuits/core/src/pedersen/hash.rs
```diff
@@ -0,0 +1,140 @@
...
+        input
+            .chunks(WINDOW_SIZE)
+            .zip_eq(&self.bases)
+            .flat_map(|(bits, powers)| {
+                bits.iter().zip_eq(powers).flat_map(|(bit, base)| match bit.eject_value() {
+                    true => Some(base.clone()),
+                    false => None,
+                })
+            })
+            .fold(Group::<E>::zero(), |acc, x| acc + x)
```
howardwu: I think `.sum()` might work here

jules: It wouldn't let me, apparently `Iterator::Sum` wasn't implemented for `Group`

howardwu: That's helpful, I will add that in a future PR.

circuits/core/src/pedersen/hash.rs
```diff
@@ -0,0 +1,131 @@
...
+        }
+
+        // Compute sum of h_i^{m_i} for all i.
+        input
+            .chunks(WINDOW_SIZE)
+            .zip_eq(&self.bases)
+            .flat_map(|(bits, powers)| {
+                bits.iter()
+                    .zip_eq(powers)
+                    .map(|(bit, base)| Group::<E>::ternary(bit, &base.clone(), &Group::<E>::zero()))
```
howardwu: Out of curiosity, does `&base.clone()` here need a `.clone()`? I thought pass by reference would be sufficient. (I assume its needed since Clippy CI didn't seem to complain about it)

howardwu (approved):
LGTM, let's wait to merge this after the base branch is merged into `testnet3`

circuits/core/src/pedersen/hash.rs
jules: You're absolutely right, I don't need it, will remove it

circuits/core/src/pedersen/hash.rs
```diff
@@ -0,0 +1,130 @@
...
+
+        // Compute sum of h_i^{m_i} for all i.
+        input
+            .chunks(WINDOW_SIZE)
+            .zip_eq(&self.bases)
+            .flat_map(|(bits, powers)| {
+                bits.iter()
+                    .zip_eq(powers)
+                    .map(|(bit, base)| Group::<E>::ternary(bit, base, &Group::<E>::zero()))
+                    .collect::<Vec<Group<E>>>()
```
ljedrz: it the `collect` necessary here? `flat_map` might work without it

jules: I couldn't get it to work with `flat_map` as it was complaining `Group` wasn't an iterator

PR #728, opened 2022-03-29
Implements a macro to automatically derive ToBytes and FromBytes for enums

d0cd:
## Motivation

Implementing `ToBytes` and `FromBytes` for enums requires hard-coded constants for enum tags.
This is hard to maintain as enum variants are introduced or removed.
This PR introduces two proc macros that derive the implementations for enums.

## Test Plan

Tests in `bytecode` verifying that instructions are given the correct tags.

### DIFF
--- a/utilities/derives/src/test_with_metrics.rs
+++ b/utilities/derives/src/test_with_metrics.rs
@@ -1,0 +1,44 @@
+// Copyright (C) 2019-2022 Aleo Systems Inc.
+// This file is part of the snarkVM library.
+
+// The snarkVM library is free software: you can redistribute it and/or modify
+// it under the terms of the GNU General Public License as published by
+// the Free Software Foundation, either version 3 of the License, or
+// (at your option) any later version.
+
+// The snarkVM library is distributed in the hope that it will be useful,
+// but WITHOUT ANY WARRANTY; without even the implied warranty of
+// MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE. See the
+// GNU General Public License for more details.
+
+// You should have received a copy of the GNU General Public License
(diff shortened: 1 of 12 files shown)


utilities/derives/src/test_with_metrics.rs
```diff
@@ -0,0 +1,44 @@
...
+
+// You should have received a copy of the GNU General Public License
+// along with the snarkVM library. If not, see <https://www.gnu.org/licenses/>.
+
+use proc_macro2::{Ident, Span};
+use proc_macro_crate::crate_name;
+use quote::quote;
+use syn::ItemFn;
+
+pub(crate) fn generate_test_function(function: ItemFn) -> proc_macro::TokenStream {
```
howardwu, line 22: @ljedrz @niklaslong Do we know why this is here? I think we should remove it. If `snarkOS` requires it, seems like it should be moved there.

ljedrz: I don't recall this, and agree that any snarkos-specific functionality should be implemented there.

niklaslong: I don't recall either 🤔

niklaslong: This is unused in `snarkVM` and `snarkOS`, it's only ever called in `test_with_metrics` which is also never called.

howardwu:
Keep the `derive` refactor as this is valuable.

Let's remove the `bytecode` example of the `enum_bytes` stuff, and we can get this PR merged in first, since the proc_macro clean up is useful.

I'm hesitant to introduce more proc_macros of this complexity at this stage, and we should consider alternative options or off the shelf solutions instead.

d0cd:
Sounds good.
Will remove this PR and introduce a new one with just the `derive` refactor.

PR #732, opened 2022-04-02
Implement Merkle path gadget

raychu86:
## Motivation

This PR implements an initial version of a Merkle path gadget with the new circuits library. 

One note is that this current implementation is hard-coded to the only existing CRH gadget in the library - Poseidon. This will need to be generalized to support other CRH implementations like Pedersen and BHP.

Additionally, because the Merkle path gadget must match the native Merkle implantation that uses the native `PoseidonCRH`, there is a lot of unnecessary padding being done. This is due to `PoseidonCRH` requiring a padded bit implementation (using 
a predetermined `INPUT_SIZE_FE`). The struggle here is that the native Merkle tree implementation deals with bytes for the hash inputs, while Poseidon prefers Field elements.


circuits/core/src/merkle_path/mod.rs
```diff
@@ -0,0 +1,248 @@
...
+    /// Ejects the account private key as `(sk_sig, r_sig)`.
+    ///
+    fn eject_value(&self) -> Self::Primitive {
+        (&self.traversal, &self.path).eject_value()
+    }
+}
+
+impl<E: Environment, const INPUT_SIZE_FE: usize> MerklePath<E, INPUT_SIZE_FE> {
+    /// Calculate the root of the merkle tree given the leaf node.
+    pub fn calculate_root(&self, leaf: Field<E>) -> Field<E> {
```
howardwu: This should be moved into a new file called `to_root.rs` with tests

circuits/core/src/merkle_path/mod.rs
```diff
@@ -0,0 +1,248 @@
...
+    //         let left_hash = Field::ternary(bit, sibling, &curr_hash);
+    //         let right_hash = Field::ternary(bit, &curr_hash, sibling);
+    //
+    //         curr_hash = self.crh.hash(&[left_hash, right_hash]);
+    //     }
+    //
+    //     curr_hash
+    // }
+
+    pub fn verify(&self, leaf: Field<E>, root: Field<E>) {
```
howardwu: This should be moved to a new file called `verify.rs` with tests

circuits/core/src/merkle_path/mod.rs
```diff
@@ -0,0 +1,248 @@
...
+        let expected_root = self.calculate_root(leaf);
+        E::assert_eq(&expected_root, &root);
```
howardwu: `verify()` should return a `Boolean<E>`

circuits/core/src/merkle_path/mod.rs
```diff
@@ -0,0 +1,248 @@
...
+    }
+
+    pub fn conditionally_verify(&self, leaf: Field<E>, root: Field<E>, condition: Boolean<E>) {
```
howardwu: This should be removed.

circuits/core/src/merkle_path/mod.rs
howardwu: Alternatively, we can remove this method. The caller can easily check that `path.to_root() == root` themselves

circuits/types/field/src/lib.rs
```diff
@@ -156,6 +156,18 @@ impl<E: Environment> From<&Field<E>> for LinearCombination<E::BaseField> {
...
+        field.to_bits_le()
+    }
+}
+
+impl<E: Environment> From<&Field<E>> for Vec<Boolean<E>> {
+    fn from(field: &Field<E>) -> Self {
+        field.to_bits_le()
+    }
+}
+
```
howardwu: Remove this and use `.to_bits_le()`. This call has no notion of endianness and should be avoided at all costs.

circuits/core/src/traits.rs
```diff
@@ -14,10 +14,14 @@
...
+use snarkvm_circuits_environment::{Eject, FromBits, Inject, Ternary, ToBits};
+
+use core::fmt::Debug;
+
 /// A trait for a hash function.
 pub trait Hash {
-    type Input;
-    type Output;
+    type Input: Inject + Eject + Ternary + ToBits + FromBits + Clone + Debug;
+    type Output: Inject + Eject + Ternary + ToBits + FromBits + Clone + Debug;
```
howardwu: Can you confirm every trait declared here is necessary?

circuits/core/src/algorithms/merkle_path/to_root.rs
```diff
@@ -0,0 +1,158 @@
...
+                assert_eq!(root, candidate_root.eject_value());
+
+                let case = format!("mode = {mode}");
+                assert_scope!(case, num_constants, num_public, num_private, num_constraints);
+            });
+        }
+    }
+
+    #[test]
+    fn test_good_root_constant() {
```
howardwu: nit: rename these method to `test_to_root_{mode}` and `test_to_root_{mode}_fails`

circuits/core/src/algorithms/merkle_path/to_root.rs
```diff
@@ -0,0 +1,158 @@
...
+
+use super::*;
+
+use crate::traits::Hash;
+
+impl<E: Environment, H: Hash> MerklePath<E, H>
+where
+    <<H as Hash>::Output as Ternary>::Boolean: From<Boolean<E>>,
+    <H as Hash>::Output: From<<<H as Hash>::Output as Ternary>::Output>,
+    Vec<<H as Hash>::Input>: From<<H as Hash>::Output>,
```
howardwu: Is there a way to use implied trait bounds?

circuits/core/src/traits.rs
raychu86: The minimum required seems to be:
```
type Input: Inject + Clone;
type Output: Inject + Eject + Ternary<Output = Self::Output> + ToBits + Clone;
```

circuits/core/src/algorithms/merkle_path/to_root.rs
raychu86: Currently reduced to:
```
<<H as Hash>::Output as Ternary>::Boolean: From<Boolean<E>>,
Vec<<H as Hash>::Input>: From<Vec<<<H as Hash>::Output as ToBits>::Boolean>>,
```

Boolean conversion is a bit trickier because each object has it's own `Boolean: BooleanTrait` type.

howardwu (approved):

PR #733, opened 2022-04-02
Implement BHP gadget

jules:
This PR adds the following changes:
- Implements a BHP `hash` and `hash_uncompressed` method
- Reduces the constraint cost of `Field::div` from `2` to `1` constraint
- Improves the performance of the native `Fp::half()` method by removing a call to `Fp::inverse()`


jules:
@howardwu It seems the hashing tests fail randomly, I'm still investigating as to why, but I may need some help - this seems quite esoteric.

howardwu:
^ The above commit contains the hardcoded circuit logic for BHP to compare against for debugging (tests pass on hardcoded).

Here is a copy of the core logic below for easier reference:
```rs
input
    .chunks(WINDOW_SIZE * BHP_CHUNK_SIZE)
    .zip(&self.bases)
    .flat_map(|(bits, bases)| {
        bits.chunks(BHP_CHUNK_SIZE).zip(bases).map(|(chunk_bits, base)| {
            // base[(chunk_bits[0] as usize) | (chunk_bits[1] as usize) << 1 | (chunk_bits[2] as usize) << 2]
            let mut encoded = base.clone();
            if chunk_bits[0].eject_value() {
                encoded += base;
            }
            if chunk_bits[1].eject_value() {
                encoded += &base.double();
            }
            if chunk_bits[2].eject_value() {
                encoded = -encoded;
            }
            encoded
        })
    })
    .fold(Group::zero(), |acc, x| acc + x)
```
_(For Future Readers: Do **NOT** use the above logic in your circuit. It is hardcoded to sanity check a proper implementation)_

howardwu:
Here is a naive (but correct) implementation of the circuit (test passes):
```rs
input
    .chunks(WINDOW_SIZE * BHP_CHUNK_SIZE)
    .zip(&self.bases)
    .flat_map(|(bits, bases)| {
        bits.chunks(BHP_CHUNK_SIZE).zip(bases).map(|(chunk_bits, base)| {
            let mut encoded = base.clone();
            encoded = Group::ternary(&chunk_bits[0], &(encoded + base), base);
            encoded = Group::ternary(&chunk_bits[1], &(&encoded + &base.double()), &encoded);
            encoded = Group::ternary(&chunk_bits[2], &(-&encoded), &encoded);
            encoded
        })
    })
    .fold(Group::zero(), |acc, x| acc + x)
```

howardwu:
Verified that your implementation of the three-bit lookup is correct.

The issue was that the arithmetic was in Edwards form, when the logic was for Montgomery form.

Here is an un-optimized version of the correct logic (test passes):
```rs
let zero = Field::<E>::zero();
input
    .chunks(WINDOW_SIZE * BHP_CHUNK_SIZE)
    .zip(self.bases.iter())
    .flat_map(|(bits, bases)| {
        bits.chunks(BHP_CHUNK_SIZE).zip(bases).map(|(chunk_bits, base)| {
            let mut x_coeffs = Vec::with_capacity(4);
            let mut y_coeffs = Vec::with_capacity(4);
            let mut acc_power = base.clone();
            for _ in 0..4 {
                let x = (Field::one() + acc_power.to_y_coordinate()) / (Field::one() - acc_power.to_y_coordinate());
                let y = &x / acc_power.to_x_coordinate();

                x_coeffs.push(x);
                y_coeffs.push(y);
                acc_power += base;
            }

            let precomp = &chunk_bits[0] & &chunk_bits[1];
            let mut x = Field::<E>::zero();
            x += &x_coeffs[0];
            x += Field::ternary(&chunk_bits[0], &(&x_coeffs[1] - &x_coeffs[0]), &zero);
            x += Field::ternary(&chunk_bits[1], &(&x_coeffs[2] - &x_coeffs[0]), &zero);
            x += Field::ternary(&precomp, &(&x_coeffs[3] - &x_coeffs[2] - &x_coeffs[1] + &x_coeffs[0]), &zero);

            // Three bit conditional neg lookup for Y
            let y = Field::ternary(
                &chunk_bits[0],
                &Field::ternary(&chunk_bits[1], &y_coeffs[3], &y_coeffs[1]),
                &Field::ternary(&chunk_bits[1], &y_coeffs[2], &y_coeffs[0]),
            );

            let y = Field::ternary(&chunk_bits[2], &-&y, &y);

            let edwards_x = &x / y;
            let edwards_y = (&x - &Field::one()) / (&x + Field::one());

            Group::from_xy_coordinates(edwards_x, edwards_y)
        })
    })
    .fold(Group::zero(), |acc, x| acc + x)
```

circuits/core/src/bhp/hash.rs
```diff
@@ -0,0 +1,214 @@
...
+        // Ensure the input size is within the parameter size,
+        if input.len() > NUM_WINDOWS * WINDOW_SIZE * BHP_CHUNK_SIZE {
+            E::halt("Incorrect input length")
+        }
+
+        // Pad the input to a multiple of `BHP_CHUNK_SIZE` for hashing.
+        let mut input = input.to_vec();
+        if input.len() % BHP_CHUNK_SIZE != 0 {
+            let padding = BHP_CHUNK_SIZE - (input.len() % BHP_CHUNK_SIZE);
+            input.extend_from_slice(&vec![Boolean::constant(false); BHP_CHUNK_SIZE][..padding]);
```
ljedrz:
```suggestion
            input.resize(input.len() + padding, Boolean::constant(false));
```

circuits/core/src/bhp/hash.rs
```diff
@@ -0,0 +1,214 @@
...
+
+    // #[test]
+    // fn test_hash_constant() {
+    //     check_hash::<8, 32>(Mode::Constant);
+    // }
+
+    // #[test]
+    // fn test_hash_public() {
+    //     check_hash::<8, 32>(Mode::Public);
+    // }
```
ljedrz: these should probably be uncommented or removed?

howardwu: Yeah, this PR is still a WIP

circuits/core/src/bhp/hash.rs
howardwu: @ljedrz the compiler doesn't seem to like the suggestion. Any alternative options?

ljedrz: What does the compiler complain about? I just double-checked it locally and it works.

howardwu: It said it doesn't implement `Copy`

howardwu: Okay strange, I confirm it works locally on my machine too. It does not work on CI with the `Copy` error.

howardwu: I was able work around it, I'm not sure what went wrong, but cleared cache and started working again. Perhaps CircleCI had a hiccup

howardwu (approved):

PR #735, opened 2022-04-03
Deserialization fixes

ljedrz:
While running fuzz tests for the snarkOS network layer, I detected a few deserialization-related vulnerabilities that needed to be patched; I planned to finalize them and release the results on Monday, but seeing the extra weekend changes, I decided to be proactive about it and provide the results I have already.

Most of these are either overallocations or unwraps, which will cause a crash when deserializing broken/malicious network messages.

Feel free to adjust the numbers I used, depending on the expected maximum values. I initially thought using `Vec::try_reserve` or tweaking the allocator would be preferable, but in the end I believe we should instead be imposing limits on a per-case basis.

Supersedes https://github.com/AleoHQ/snarkVM/pull/592 and https://github.com/AleoHQ/snarkVM/pull/609.


algorithms/src/merkle_tree/merkle_path.rs
```diff
@@ -125,16 +125,22 @@ impl<P: MerkleParameters> FromBytes for MerklePath<P> {
         //  If you are seeing this message, please be proactive in bringing it up :)
         let parameters = {
             let setup_message_length: u64 = FromBytes::read_le(&mut reader)?;
+            if setup_message_length > 1024 * 1024 * 10 {
```
howardwu: To sanity check, if I were to reduce this from a `u64` to a `u16`, would this mitigate the need for checking here?

(We can safely assume this will not exceed `u16`; it won't exceed `u8` within reason, though we don't know what the future holds for us)

howardwu: I think this question actually applies to most (if not all) of these cases

howardwu:
Thanks for catching these Lukasz! Looks like we should redesign our serialization infrastructure to account for this.

algorithms/src/merkle_tree/merkle_path.rs
ljedrz: Yes, reducing the size of the length prefix from `u64` to `u16` will completely fix the issue for vectors of bytes (and other small objects), and reduce the scale of the issue in the rest of cases, but note that a `u16::MAX` (65535) number of large objects in vectors might also be an issue.

howardwu:
Alright, this should handle the serialization cases that were caught.

howardwu (approved):

ljedrz:
I can confirm that the fuzzing artifacts still pass with the additional changes :+1:; I'll run the tests a bit more tomorrow just to be safe.

PR #739, opened 2022-04-04
Metrics for circuits

d0cd:
## Motivation

- Introduces two traits `Metrics` and `OutputMode` which are used to define metadata on circuit operations.
- Introduces `Count`, which is used to define the number of constants, number of variables, and number of constraints for circuit operations.
- Sample implementations are available in `circuits/types` and `circuits/core/pedersen`
- Uses the implementations of `Metrics` and `OutputMode` to statically determine the number of constants, variables, and constraints for bytecode programs. See `bytecode/src/main.rs` for an example of how counts can be determined. 

## Test Plan

- Uses implementations of `Metrics` and `OutputMode` in tests for `circuits`.
- Unit tests for `Measurement`s.
- Tests of bytecode for nontrivial programs.


circuits/types/field/src/neg.rs
```diff
@@ -34,11 +34,32 @@ impl<E: Environment> Neg for &Field<E> {
...
+        }
+    }
+}
+
 #[cfg(test)]
 mod tests {
     use super::*;
     use snarkvm_circuits_environment::Circuit;
 
+    // TODO: Add tests checking the count and output mode.
```
howardwu: Note: This needs to be addressed

bytecode/src/main.rs
```diff
@@ -56,15 +93,44 @@ mod tests {
...
+        // Get estimated circuit count for the `HelloWorld` program.
+        let count = HelloWorld::count::<Process>();
+
+        // Run the `HelloWorld` program with the given inputs.
         let first = Value::from_str("1field.public");
         let second = Value::from_str("1field.private");
 
-        let expected = Value::<Process>::from_str("2field.private");
+        // Store the circuit counts before running the program.
+        let old_num_constants = <Process as Program>::Aleo::num_constants();
```
howardwu: Note: Introduce a `count()` method in `Environment` trait, and use that instead so we don't end up with this

howardwu:
TODO: @d0cd Remove`OutputMode`, it's not a productive check here. If you'd like to hardcode the check in tests (inside `fn check_*` that is fine)

howardwu (approved):

PR #740, opened 2022-04-04
Add Pedersen commitment gadget

jules:


howardwu (commented):
This looks great. One outstanding request:

Can you merge `PedersenCommitment` into `Pedersen` (along with their typedef variants)?

I understand the random base will also need to added, that is okay since this will only be loaded once at initialization for our runtime.

howardwu:
Actually, one more item:

If you take a look at the `Pedersen::hash_uncompressed` tests, you'll see there are tests to check for homomorphic addition.

This property also holds for Pedersen commitments. Can you add a test to implement this check for the `Pedersen*` variants?

The property you want to enforce is:
```
Commit(i0 + i1, r0 + r1) == Commit(i0, r0) + Commit(i1, r1)
```

howardwu (approved):
LGTM

PR #750, opened 2022-04-12
Add ECIES Poseidon Encryption/Decryption gadget

jules:


circuits/core/src/algorithms/poseidon/encryption.rs
```diff
@@ -0,0 +1,77 @@
...
+        let mut bits = Vec::<Boolean<E>>::with_capacity(encoded_message.len() * capacity);
+        for element in encoded_message.iter() {
+            bits.extend_from_slice(&element.to_bits_le()[..capacity]);
+        }
+
+        // Drop all the ending zeros and the last "1" bit.
+        // Note that there must be at least one "1" bit because the last element is not zero.
+        // TODO: how do i do this without ejecting
+        loop {
+            if bits.pop().unwrap().eject_value() == true {
```
raychu86: I think an `enforce` call would work here. 

For example:
`E::assert_eq(bit, Boolean::constant(false));`

jules: Hmm, would this not then create an invalid constraint when the first `1` bit is hit and therefore produce an invalid circuit? Not really sure how to use it as a conditional..

jules:
Not entirely sure about the comments, but I think the logic should all be good.

howardwu (approved):

PR #753, opened 2022-04-15
Introduce `TwoToOneCRH` and `LeafCRH` for Merkle trees

raychu86:
## Motivation

This PR introduces an additional CRH to Merkle trees to optimize the input sizes for each required hash. Each MerkleParameter will have a `LeafCRH` and `TwoToOneCRH` to be used for leaves and parent nodes, respectively. 

The inner (two-to-one) hashes require double the input space of the leaf hashes, so the old model would add unnecessary padding for each leaf hash.


algorithms/src/merkle_tree/masked_merkle_tree_parameters.rs
```diff
@@ -37,16 +50,22 @@ impl<H: CRH, const DEPTH: usize> MerkleParameters for MaskedMerkleTreeParameters
     //  If you are seeing this message, please be proactive in bringing it up :)
     /// Returns the saved `message` from calling the `MerkleParameters::setup()` function.
     fn setup_message(&self) -> &str {
-        &self.2
+        &self.3
     }
 
-    fn crh(&self) -> &Self::H {
+    fn crh(&self) -> &Self::LeafCRH {
```
howardwu: nit: Rename to `leaf_crh` so it is clearly distinct from the `two_to_one_crh` (you should be able to do an `fn + F6`)

algorithms/src/merkle_tree/merkle_path.rs
```diff
@@ -82,9 +82,9 @@ impl<P: MerkleParameters> MerklePath<P> {
     /// Returns: (left, right)
     fn select_left_right_bytes(
         index: u64,
-        computed_hash: &<P::H as CRH>::Output,
-        sibling_hash: &<P::H as CRH>::Output,
-    ) -> Result<(<P::H as CRH>::Output, <P::H as CRH>::Output), MerkleError> {
+        computed_hash: &<P::LeafCRH as CRH>::Output,
+        sibling_hash: &<P::LeafCRH as CRH>::Output,
+    ) -> Result<(<P::LeafCRH as CRH>::Output, <P::LeafCRH as CRH>::Output), MerkleError> {
```
howardwu: Just to sanity check, shouldn't this be using a `TwoToOneCRH`? (I think only the first instance should be using the `LeafCRH`)

algorithms/src/merkle_tree/merkle_tree.rs
```diff
@@ -296,14 +296,28 @@ impl<P: MerkleParameters + Send + Sync> MerkleTree<P> {
     fn hash_row<L: ToBytes + Send + Sync>(
```
howardwu: nit: Rename to `hash_leaf_row` to delineate

algorithms/src/merkle_tree/merkle_tree.rs
```diff
@@ -296,14 +296,28 @@ impl<P: MerkleParameters + Send + Sync> MerkleTree<P> {
...
+    ) -> Result<Vec<Vec<<<P as MerkleParameters>::LeafCRH as CRH>::Output>>, MerkleError> {
         match leaves.len() {
             0 => Ok(vec![]),
             _ => {
                 Ok(vec![crate::cfg_iter!(leaves).map(|leaf| parameters.hash_leaf(&leaf).unwrap()).collect::<Vec<_>>()])
             }
         }
     }
+
+    fn hash_row_inner<L: ToBytes + Send + Sync>(
```
howardwu: nit: Rename to `hash_two_to_one_row`

algorithms/src/traits/merkle_tree.rs
```diff
@@ -37,34 +38,42 @@ pub trait MerkleParameters: Clone + Debug + PartialEq + Eq + Send + Sync {
     fn setup_message(&self) -> &str;
 
     /// Returns the collision-resistant hash function used by the Merkle tree.
-    fn crh(&self) -> &Self::H;
+    fn crh(&self) -> &Self::LeafCRH;
```
howardwu: nit: Rename to `leaf_crh`

algorithms/src/merkle_tree/merkle_tree.rs
```diff
@@ -296,14 +296,28 @@ impl<P: MerkleParameters + Send + Sync> MerkleTree<P> {
...
+        parameters: &P,
+        inner: &[L],
+    ) -> Result<Vec<Vec<<<P as MerkleParameters>::TwoToOneCRH as CRH>::Output>>, MerkleError> {
+        match inner.len() {
+            0 => Ok(vec![]),
+            _ => Ok(vec![
+                crate::cfg_iter!(inner)
+                    .map(|inner_nodes| parameters.hash_two_to_one(&inner_nodes).unwrap())
```
ljedrz: since the result is a `Result<Vec<T, MerkleError>>`, perhaps we can avoid the calls to `unwrap` here (and remove the enclosing `Ok()`) to improve this method's safety?

ljedrz (commented):
Left a suggestion :+1:

algorithms/src/merkle_tree/merkle_tree.rs
raychu86: Agreed, I've removed the `unwraps` in [cab5f14](https://github.com/AleoHQ/snarkVM/pull/753/commits/cab5f149842eadd899ee2285e4c48980b74f2f75).

I also noticed that we unnecessarily had a double Vec that was seemingly unused in our `hash_leaf_row` and `hash_two_to_one_row` impls. I've updated the output to `Result<Vec<T, MerkleError>>` from `Result<Vec<Vec<T, MerkleError>>>` and made the corresponding changes above.

howardwu (approved):

PR #760, opened 2022-04-25
Marlin batch-proving via parallel composition

Pratyush:
## Motivation

Adds batch proving to Marlin, so that we can produce a single proof for multiple instances of the same circuit.

## Test Plan

(Write your test plan here)

## Related PRs

(Link your related PRs here)


algorithms/src/snark/marlin/ahp/ahp.rs
```diff
@@ -185,17 +149,22 @@ impl<F: PrimeField, MM: MarlinMode> AHPForR1CS<F, MM> {
...
+        let public_input =
+            public_input.iter().map(|p| prover::ConstraintSystem::format_public_input(p)).collect::<Vec<_>>();
+
+        for public_input in &public_input {
+            if !Self::formatted_public_input_is_admissible(public_input) {
+                return Err(AHPError::InvalidPublicInputLength);
+            }
         }
-        let input_domain = EvaluationDomain::new(public_input.len()).ok_or(AHPError::PolynomialDegreeTooLarge)?;
+        let input_domain = EvaluationDomain::new(public_input[0].len()).ok_or(AHPError::PolynomialDegreeTooLarge)?;
```
ljedrz: we might want to guard against empty `public_input` before this call, in order to not potentially panic on indexing

algorithms/src/snark/marlin/ahp/prover/oracles.rs
```diff
@@ -14,45 +14,72 @@
...
+
+    /// Iterate over the polynomials output by the prover in the first round.
+    /// Intended for use when opening.
+    pub fn iter_for_open(&'a self) -> impl Iterator<Item = &'a LabeledPolynomial<F>> {
+        self.batches.iter().flat_map(|b| b.iter_for_open()).chain(self.mask_poly.as_ref())
+    }
+
+    pub fn matches_info(&self, info: &BTreeMap<PolynomialLabel, PolynomialInfo>) -> bool {
+        self.batches.iter().all(|b| b.matches_info(info))
+            && self.mask_poly.as_ref().map_or(true, |p| p.info() == &info[p.label()])
```
ljedrz:
```suggestion
            && self.mask_poly.as_ref().map_or(true, |p| Some(p.info()) == info.get(p.label()))
```
haven't tested it (might require a call to `.as_ref()` or an extra reference), but the suggestion won't panic if there is no `info[p.label()]`

algorithms/src/snark/marlin/ahp/prover/oracles.rs
```diff
@@ -14,45 +14,72 @@
...
-        [Some(&self.w_poly), Some(&self.z_a_poly), Some(&self.z_b_poly), self.mask_poly.as_ref()].into_iter().flatten()
+        [(&self.w_poly), &self.z_a_poly, &self.z_b_poly].into_iter()
+    }
+
+    pub fn matches_info(&self, info: &BTreeMap<PolynomialLabel, PolynomialInfo>) -> bool {
+        self.w_poly.info() == &info[self.w_poly.label()]
+            && self.z_a.info() == &info[self.z_a.label()]
+            && self.z_b.info() == &info[self.z_b.label()]
+            && self.z_a_poly.info() == &info[self.z_a_poly.label()]
+            && self.z_b_poly.info() == &info[self.z_b_poly.label()]
```
ljedrz: the same suggestion as https://github.com/AleoHQ/snarkVM/pull/760/files#r857827628 applies here and in the other cases of indexing in this file, if there is any chance that it might fail

algorithms/src/polycommit/sonic_pc/polynomial.rs
```diff
@@ -33,18 +33,50 @@ use itertools::Itertools;
...
+}
+
+impl PolynomialInfo {
+    /// Construct a new labeled polynomial by consuming `polynomial`.
+    pub fn new(label: PolynomialLabel, degree_bound: Option<usize>, hiding_bound: Option<usize>) -> Self {
+        Self { label, degree_bound, hiding_bound }
+    }
+
+    /// Return the label for `self`.
+    pub fn label(&self) -> &String {
```
ljedrz:
```suggestion
    pub fn label(&self) -> &str {
```
nit: it's better to return `&str` instead of `&String` whenever possible

howardwu (approved):

howardwu (approved):

PR #763, opened 2022-04-26
feat: impl bytecode instructions And, Equal, Not, Or, XOR, NAND, NEQ, NOR 

amousa11:
## Motivation

Implement bytecode instructions for And,Equal, Not, Or, XOR, NAND, NEQ, NOR.

Add opcodes to reserved keywords list in `identifiers.rs`.

Refactored testing macro implementations to cut down on duplication and increase clarity. Generated tests also have clearer names. For example,`test_modes!(boolean, or, "true", "false", "false")` would produce a test named `test_or_boolean_modes`.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -2033,6 +2033,7 @@ dependencies = [
  "indexmap",
  "nom",
  "once_cell",
+ "paste",
  "snarkvm-algorithms",
  "snarkvm-circuits",
  "snarkvm-curves",
--- a/bytecode/Cargo.toml
+++ b/bytecode/Cargo.toml
@@ -30,6 +30,9 @@ version = "7.1"
 [dependencies.once_cell]
 version = "1.8.0"
 
+[dependencies.paste]
+version = "1.0.7"
+
 [dev-dependencies.snarkvm-algorithms]
 path = "../algorithms"
 version = "0.7.5"
--- a/bytecode/src/function/instructions/and.rs
+++ b/bytecode/src/function/instructions/and.rs
@@ -135,7 +135,6 @@ mod tests {
         ["public", "constant", "constant"],
         ["public", "private", "private"],
         ["private", "public", "private"],
-        // Why does this pass?? Feels like private & constant should yield a private
         ["private", "constant", "constant"],
         ["private", "private", "private"],
         ["constant", "public", "public"],
--- a/bytecode/src/function/instructions/and.rs
+++ b/bytecode/src/function/instructions/and.rs
@@ -128,7 +128,7 @@ impl<P: Program> Into<Instruction<P>> for And<P> {
 #[cfg(test)]
 mod tests {
     use super::*;
-    use crate::{test_instruction_halts, test_modes};
+    use crate::{test_instruction_halts, test_modes, Identifier, Process};
 
     const SIGNED_INTEGER_MODE_TESTS: [[&str; 3]; 9] = [
         ["public", "public", "private"],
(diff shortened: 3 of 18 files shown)


bytecode/src/function/instructions/add_wrapped.rs
```diff
@@ -0,0 +1,178 @@
...
+impl<P: Program> AddWrapped<P> {
+    /// Returns the operands of the instruction.
+    pub fn operands(&self) -> Vec<Operand<P>> {
+        self.operation.operands()
+    }
+
+    /// Returns the destination register of the instruction.
+    pub fn destination(&self) -> &Register<P> {
+        self.operation.destination()
+    }
```
ljedrz: these 2 methods look like good candidates for `Operation` membership or a new dedicated trait that would apply to all objects containing a `BinaryOperation`

bytecode/src/function/instructions/add_wrapped.rs
```diff
@@ -0,0 +1,178 @@
...
+            (Literal::I32(a), Literal::I32(b)) => Literal::I32(a.add_wrapped(&b)),
+            (Literal::I64(a), Literal::I64(b)) => Literal::I64(a.add_wrapped(&b)),
+            (Literal::I128(a), Literal::I128(b)) => Literal::I128(a.add_wrapped(&b)),
+            (Literal::U8(a), Literal::U8(b)) => Literal::U8(a.add_wrapped(&b)),
+            (Literal::U16(a), Literal::U16(b)) => Literal::U16(a.add_wrapped(&b)),
+            (Literal::U32(a), Literal::U32(b)) => Literal::U32(a.add_wrapped(&b)),
+            (Literal::U64(a), Literal::U64(b)) => Literal::U64(a.add_wrapped(&b)),
+            (Literal::U128(a), Literal::U128(b)) => Literal::U128(a.add_wrapped(&b)),
+            _ => P::halt(format!("Invalid '{}' instruction", Self::opcode())),
+        };
```
ljedrz: a macro could shorten `match` expressions like this; I haven't tested it, but it would look something like
```
macro_rules! impl_evaluate {
    ($($ty:ident),*) => {
        #[inline]
        fn evaluate(&self, registers: &Registers<P>) {
            (...)

            let result = match (first, second) {
                $( (Literal::$ty(a), Literal::$ty(b)) => Literal::$ty(a.add_wrapped(&b)), )*
                _ => P::halt(format!("Invalid '{}' instruction", Self::opcode())),
            };

            registers.assign(self.operation.destination(), result);
        }
    }
}
```
and called like
```
impl<P: Program> Operation<P> for AddWrapped<P> {
    impl_evaluate!(I8, I16, I32, I64, I128, U8, U16, U32, U64, U128);
}
```
it might not apply to all implementations of `Operation`, but it's there's possibly some room for code deduplication

bytecode/src/function/instructions/add_wrapped.rs
```diff
@@ -0,0 +1,178 @@
...
+}
+
+impl<P: Program> ToBytes for AddWrapped<P> {
+    fn write_le<W: Write>(&self, mut writer: W) -> IoResult<()> {
+        self.operation.write_le(&mut writer)
+    }
+}
+
+#[allow(clippy::from_over_into)]
+impl<P: Program> Into<Instruction<P>> for AddWrapped<P> {
```
ljedrz: (applies to all `Into` implementations) the docs for `std` always recommend implementing `From` instead of `Into`: https://doc.rust-lang.org/std/convert/trait.Into.html (or https://doc.rust-lang.org/std/convert/trait.From.html)

ljedrz: i just noticed `#[allow(clippy::from_over_into)]`, meaning this had been reported by `clippy`; what's the reason behind overriding it?

amousa11: The first few instructions were implemented this way so I followed the same pattern for the rest. @howardwu might have more insight as to why. Happy to update this if we decide to use From over Into.

bytecode/src/function/instructions/add_wrapped.rs
amousa11: I think in this case I'd prefer to keep this simple and readable, even if there is code duplication.

bytecode/src/function/instructions/add_wrapped.rs
amousa11: Yes, perhaps we could add these 2 methods to Operation. Though the way the Instruction enum is currently implemented, it does implicitly require all variants to implement these methods anyways. 

@howardwu Do you think this behavior should be captured under Operation or maybe under a separate trait?

bytecode/src/function/instructions/add_wrapped.rs
howardwu: AddWrapped should not be invoked outside of this module

howardwu (approved):

PR #776, opened 2022-05-02
Decompress powers of G

### LINKED ISSUE #769: [Sanity Check] Universal SRS serialization form
## 🐛 Bug Report

A number of unit tests in `snarkvm-algorithms` and `snarkvm-dpc` slowed down following the integration of the new universal SRS infrastructure.

Can we confirm that the files are stored in **uncompressed** form? (@julesdesmit) If they're in compressed form, we should update them to uncompressed form. 

If the serialization format isn't the cause of the performance regression, we should look into the cause. There might be something happening that we will want to address before shipping this version of snarkVM.

jules:
Fixes #769


jules:
The current CI errors should be resolved once the parameters are uploaded on S3 and the proper naming format can be used.

howardwu (approved):

PR #800, opened 2022-05-11
Fix bug in GPU blst_377_cuda

arsadmin:
## Motivation
bug in blst_p1_add_projective_to_projective function, The two points are equal, so is double,is not zero.

## Test Plan
image
The array has two items and their values are the same. The test result is wrong. The PR result is correct

## Related PRs
(Link your related PRs here)

### DIFF
--- a/algorithms/src/msm/variable_base/blst_377_cuda/blst_377_ops.cu
+++ b/algorithms/src/msm/variable_base/blst_377_cuda/blst_377_ops.cu
@@ -417,6 +417,12 @@ __device__ void blst_p1_add_projective_to_projective(blst_p1 *out, const blst_p1
     blst_fp_mul(s2, p2->Y, p1->Z);
     blst_fp_mul(s2, s2, z1z1);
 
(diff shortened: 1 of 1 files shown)


282548573 (approved):
Yes, the bug exists. PR is OK

howardwu:
Is this PR still relevant?

algorithms/src/msm/variable_base/blst_377_cuda/blst_377_ops.cu
```diff
@@ -417,6 +417,12 @@ __device__ void blst_p1_add_projective_to_projective(blst_p1 *out, const blst_p1
     blst_fp_mul(s2, p2->Y, p1->Z);
     blst_fp_mul(s2, s2, z1z1);
 
+    if (is_blst_fp_eq(u1, u2) && is_blst_fp_eq(s1, s2)){
+        // The two points are equal, so we double
+        blst_p1_double(out, p1);
+        return;
+    }
+
```
howardwu, line 425: You can frankly check this before converting to affine coordinates.

howardwu:
Closing as we no longer use this code for CUDA.

PR #807, opened 2022-05-15
Modify Pedersen to use a 4-bit window

jules:


jules:
Closing, as there's no efficiency gain to be had (lookup is no longer free in a 4-bit window)

PR #813, opened 2022-05-20
Implement Sinsemilla hash

jules:


jules:
To note:

Sinsemilla with the default parameters `k := 10` yields `1537` constraints for a `1030` bit input.
Sinsemilla with `k := 14` yields `1102` constraints for a `1036` bit input.

The math needs to be checked on whether we can modify the window size, but we can really drop a lot of constraints if it turns out to be the case, so an investigation is worth it.

jules:
Testing constraint counts for both bit windows:
```
sinsemilla 14 bit window:
1024 (1036) yields 1102
768 (770) yields 817
512 (518) yields 547
256 (266) yields 277

sinsemilla 10 bit window:
1024 (1030) yields 1537
768 (770) yields 1147
512 (520) yields 772
256 (260) yields 382
```

I'll do some more research to figure out if this is safe to do or not.

PR #815, opened 2022-05-20
Add support for batch proving in transaction generation

Pratyush:
## Motivation

Adds batch proving to `dpc`.

Benchmark numbers:

<table>
<tr>
	<td> Arity
	<td> Old Input
	<td> New Input
	<td> Old Output
	<td> New Output
<tr>
	<td> 1x1
	<td> 3.23
	<td> 3.3
	<td> 0.953
	<td> 0.976
<tr>
	<td> 2x2
	<td> 6.40
	<td> 3.53
	<td> 1.79
	<td> 1.04
<tr>
	<td> 4x4
	<td> 12.68
	<td> 4.18
	<td> 3.58
	<td> 1.15
<tr>
	<td> 1x8
	<td> 3.34
	<td> 3.15
	<td> 7.43
	<td> 1.55
<tr>
       <td> 8x1
	<td> 25.2
	<td> 4.84
	<td> 0.84
	<td> 0.88
</table>

Overall, no case is slower, and we see at least 2x improvements across the board. In some cases we see 5-6x improvements!

## Test Plan

I have updated some tests, but more eyes as well as new test suggestions would be great.

## Related PRs

(Link your related PRs here)


dpc/src/virtual_machine/virtual_machine.rs
```diff
@@ -22,20 +22,33 @@ use itertools::Itertools;
 use rand::{CryptoRng, Rng};
 use std::sync::Arc;
 
-#[derive(Clone, Debug)]
+#[derive(Clone, Derivative)]
+#[derivative(Debug(bound = "N: Network"))]
```
ljedrz, line 26: this probably applies to a lot more objects, but do we even need `Debug` (at least the default impl, which might already be unreadable even without the circuits) here?

Pratyush: I added it mostly to satisfy existing trait bounds. Probably a thorough evaluation of this belongs in another PR.

Pratyush:
Q: now that the sizes of the input and output proofs depend on the number of events in a transaction, these constants are no longer constants: 

https://github.com/AleoHQ/snarkVM/blob/8ceff2a07605a3daa9c073650e2ef19fe154c81a/dpc/src/traits/network.rs#L152

https://github.com/AleoHQ/snarkVM/blob/8ceff2a07605a3daa9c073650e2ef19fe154c81a/dpc/src/traits/network.rs#L153

Should we replace these with methods of the form `fn {in,out}put_proof_size(num_events: usize) -> usize`?

dpc/src/circuits/tests.rs
```diff
@@ -150,27 +150,32 @@ fn dpc_execute_circuits_test<N: Network>(
 
         //////////////////////////////////////////////////////////////////////////
 
-        let input_proof = <N as Network>::InputSNARK::prove(&input_proving_key, &input_circuit, rng).unwrap();
-        assert_eq!(N::INPUT_PROOF_SIZE_IN_BYTES, input_proof.to_bytes_le().unwrap().len());
-
-        // Verify that the inner circuit proof passes.
-        assert!(<N as Network>::InputSNARK::verify(&input_verifying_key, &input_public, &input_proof).unwrap());
+        input_circuits.push(input_circuit);
```
Pratyush, line 153: We now just store the inputs and circuits, instead of proving them immediately.

dpc/src/transaction/transaction.rs
```diff
@@ -199,6 +204,27 @@ impl<N: Network> Transaction<N> {
...
+        let input_public_variables: Vec<_> = self
+            .transitions
+            .iter()
+            .flat_map(|t| t.input_public_variables(self.ledger_root, transitions.root()))
+            .collect();
+
+        let output_public_variables: Vec<_> =
+            self.transitions.iter().flat_map(|t| t.output_public_variables()).collect();
+
+        match self.kernel_proof.verify(&input_public_variables, &output_public_variables) {
```
Pratyush, line 216: This is the only place that verifies the new batch proofs. Is there any other method that needs to be updated?

dpc/src/transaction/transition.rs
```diff
@@ -554,13 +574,13 @@ mod tests {
             let transaction = Testnet1::genesis_block().to_coinbase_transaction().unwrap();
             let transition = transaction.transitions().first().unwrap().clone();
             let transition_bytes = transition.to_bytes_le().unwrap();
-            assert_eq!(1389, transition_bytes.len());
+            assert_eq!(501, transition_bytes.len());
```
Pratyush, line 577: Transitions no longer contain input/output proofs, and so the transition is smaller.

dpc/src/virtual_machine/execution.rs
```diff
@@ -53,64 +52,16 @@ impl<N: Network> ProgramExecution<N> {
...
-                    }
-                }
-                Err(error) => {
-                    eprintln!("Failed to validate output proof {}: {:?}", i, error);
-                    return false;
-                }
-            };
-        }
-
+    pub fn verify(&self, transition_id: N::TransitionID) -> bool {
```
Pratyush, line 64: We no longer have any proof verification logic for transitions

dpc/src/virtual_machine/virtual_machine.rs
```diff
@@ -154,8 +157,24 @@ impl<N: Network> VirtualMachine<N> {
     }
 
     /// Finalizes the virtual machine state and returns a transaction.
-    pub fn finalize(&self) -> Result<Transaction<N>> {
-        Transaction::from(*N::input_circuit_id(), *N::output_circuit_id(), self.ledger_root, self.transitions.clone())
+    pub fn finalize<R: Rng + CryptoRng>(&self, rng: &mut R) -> Result<Transaction<N>> {
```
Pratyush, line 160: This requires an `rng` now to produce zero-knowledge batch proofs

dpc/src/transaction/transaction.rs
howardwu: This is the only place! I've tried to keep the surfaces as deduped as possible

howardwu (approved):

howardwu:
LGTM, resampling parameters

howardwu:
> Q: now that the sizes of the input and output proofs depend on the number of events in a transaction, these constants are no longer constants:
> 
> https://github.com/AleoHQ/snarkVM/blob/8ceff2a07605a3daa9c073650e2ef19fe154c81a/dpc/src/traits/network.rs#L152
> 
> https://github.com/AleoHQ/snarkVM/blob/8ceff2a07605a3daa9c073650e2ef19fe154c81a/dpc/src/traits/network.rs#L153
> 
> Should we replace these with methods of the form `fn {in,out}put_proof_size(num_events: usize) -> usize`?

These can be removed now. I saw you're also using the serde size encoder now too, so these truly don't need to exist anymore (which is great!)

PR #819, opened 2022-05-21
Reduce the number of constraints for Field and Scalar comparison operations

d0cd:
- Uses parity check to implement a cheap compare for `Scalar`.
- Introduces a new method for bitwise `less_than`.

### DIFF
--- a/circuits/types/field/src/compare.rs
+++ b/circuits/types/field/src/compare.rs
@@ -23,20 +23,32 @@ impl<E: Environment> Compare<Field<E>> for Field<E> {
     fn is_less_than(&self, other: &Self) -> Self::Output {
         if self.is_constant() && other.is_constant() {
             Boolean::constant(self.eject_value() < other.eject_value())
-        } else if other.is_constant() {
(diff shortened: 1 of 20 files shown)


circuits/types/field/src/compare.rs
```diff
@@ -21,25 +21,63 @@ impl<E: Environment> Compare<Field<E>> for Field<E> {
...
-
-        for (index, (self_bit, other_bit)) in bits_be.enumerate() {
-            // Determine if `self` is less than `other` up to the `index`-th bit.
-            is_less_than |= &are_previous_bits_equal & (!self_bit & other_bit);
-
-            // Skip the update to the LSB, as this boolean is subsequently discarded.
-            if index != self_bits_be.len() - 1 {
-                are_previous_bits_equal &= self_bit.is_equal(other_bit);
-            }
+        if self.is_constant() && other.is_constant() {
```
ljedrz, line 25: a `match` expression against `(self.is_constant(), other.is_constant())` might be a bit more readable here

howardwu (approved):

PR #829, opened 2022-05-28
Updates Pedersen and BHP

howardwu:
# Motivation

- Updates BHP to support variable-length inputs (beyond the original total window size)
- Adds `Inject` on `Pedersen`, `BHP`, and `Poseidon`

## BHP

BHP is a collision-resistant hash function that takes a variable-length input.

### Design
The BHP hash function splits the given input into blocks, and processes them iteratively.

The first iteration is initialized as follows:
```text
DIGEST_0 = BHP([ 0...0 || DOMAIN || LENGTH(INPUT) || INPUT[0..BLOCK_SIZE] ]);
```
Each subsequent iteration is initialized as follows:
```text
DIGEST_N+1 = BHP([ DIGEST_N[0..DATA_BITS] || INPUT[(N+1)*BLOCK_SIZE..(N+2)*BLOCK_SIZE] ]);
```

### DIFF
--- a/circuit/algorithms/build.rs
+++ b/circuit/algorithms/build.rs
@@ -1,0 +1,21 @@
+// Copyright (C) 2019-2022 Aleo Systems Inc.
+// This file is part of the snarkVM library.
+
+// The snarkVM library is free software: you can redistribute it and/or modify
+// it under the terms of the GNU General Public License as published by
+// the Free Software Foundation, either version 3 of the License, or
+// (at your option) any later version.
+
+// The snarkVM library is distributed in the hope that it will be useful,
+// but WITHOUT ANY WARRANTY; without even the implied warranty of
+// MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE. See the
+// GNU General Public License for more details.
+
+// You should have received a copy of the GNU General Public License
+// along with the snarkVM library. If not, see <https://www.gnu.org/licenses/>.
+
+fn main() {
+    if cfg!(feature = "enable_console") {
+        println!("cargo:rustc-cfg=console");
+    }
+}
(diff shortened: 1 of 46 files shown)


circuit/algorithms/build.rs
```diff
@@ -0,0 +1,21 @@
...
+// The snarkVM library is distributed in the hope that it will be useful,
+// but WITHOUT ANY WARRANTY; without even the implied warranty of
+// MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE. See the
+// GNU General Public License for more details.
+
+// You should have received a copy of the GNU General Public License
+// along with the snarkVM library. If not, see <https://www.gnu.org/licenses/>.
+
+fn main() {
+    if cfg!(feature = "enable_console") {
```
Pratyush, line 18: Why not just use the `enable_console` feature, instead of cfg(console)? It’s generally better to avoid build.rs if we can

howardwu: Good to know. What's the reason it's better?

circuit/algorithms/src/bhp/hasher/hash_uncompressed.rs
```diff
@@ -0,0 +1,232 @@
...
+            // `(lambda * (this_x - sum_x)) == (this_y + sum_y)`
+            E::enforce(|| (&lambda, this_x - &sum_x, this_y + &sum_y));
+
+            (sum_x, sum_y)
+        };
+
+        // Compute sum of h_i^{sum of (1-2*c_{i,j,2})*(1+c_{i,j,0}+2*c_{i,j,1})*2^{4*(j-1)} for all j in segment}
+        // for all i. Described in section 5.4.1.7 in the Zcash protocol specification.
+        //
+        // Note: `.zip()` is used here (as opposed to `.zip_eq()`) as the input can be less than
```
Pratyush, line 90: Is this comment still true? Now that we’re padding everywhere, we might be able to use `zip_eq`

howardwu: Yes, for the final round of BHP, it is still the case.

circuit/algorithms/build.rs
Pratyush: Sometimes build.rs can interfere with incremental compilation, which slows down iteration times

howardwu: I'll address in a future PR

PR #845, opened 2022-06-09
PoC FROST implementation

raychu86:
## Motivation

This PR introduces a PoC implementation of FROST threshold signatures - https://eprint.iacr.org/2020/852.pdf

The current keygen scheme depends on a trusted dealer, and integration with the current Aleo address scheme is a WIP.


frost/Cargo.toml
```diff
@@ -0,0 +1,115 @@
...
+[dependencies.digest]
+version = "0.10"
+
+[dependencies.getrandom]
+version = "0.2"
+optional = true
+features = [ "js" ]
+default-features = false
+
+[dependencies.hashbrown]
```
ljedrz, line 71: is this dependency needed? I don't see it used, and the standard library already uses it internally for `HashMap` and `HashSet`

frost/src/utils.rs
```diff
@@ -0,0 +1,145 @@
...
+use anyhow::{anyhow, Result};
+use std::{
+    collections::HashMap,
+    ops::{Mul, Sub},
+};
+
+/// Calculate the Lagrange coefficient for a given participant index.
+pub fn calculate_lagrange_coefficients<G: AffineCurve>(
+    participant_index: u64,
+    all_participant_indices: &[u64],
```
ljedrz, line 32: looking at the callsites, this could be an `impl Iterator<Item = u64>` to avoid `Vec` allocations (or an `I: Iterator<Item = u64>`)

frost/src/utils.rs
```diff
@@ -0,0 +1,145 @@
...
+) -> Result<G::ScalarField> {
+    let mut numerator = G::ScalarField::one();
+    let mut denominator = G::ScalarField::one();
+
+    let participant_index_scalar = match G::ScalarField::from_repr(participant_index.into()) {
+        Some(scalar) => scalar,
+        None => Err(anyhow!("Failed to convert participant index to scalar"))?,
```
ljedrz, line 39:
```suggestion
        None => bail!("Failed to convert participant index to scalar"),
```
this [bail](https://docs.rs/anyhow/latest/anyhow/macro.bail.html) suggestion applies to all other places like this

frost/src/frost.rs
```diff
@@ -0,0 +1,299 @@
...
+
+            if !partial_signature.is_valid(signer_public_key, challenge, lambda_i, commitment_i.to_affine()) {
+                return Err(anyhow!("Invalid partial signature"));
+            }
+        }
+
+        // Aggregate the partial signatures.
+        let mut signature = G::ScalarField::zero();
+        for partial_signature in &partial_signatures {
+            signature += partial_signature.partial_signature;
```
ljedrz, line 181: a good candidate for `Sum`, if it applies to `G::ScalarField`

frost/src/keys.rs
```diff
@@ -0,0 +1,276 @@
...
+    num_participants: u8,
+    threshold: u8,
+    secret: &G::ScalarField,
+    rng: &mut R,
+) -> (Vec<SignerShare<G>>, PublicKeys<G>) {
+    if num_participants < 1 {
+        panic!("The number of participants must be greater than 0.");
+    }
+
+    if threshold < 1 || threshold > num_participants {
```
ljedrz, line 99:
```suggestion
    if threshold == 0 || threshold > num_participants {
```
`clippy` is likely to suggest this

ljedrz: the same applies to line 90

PR #846, opened 2022-06-10
Add lookups to Marlin

jules:


algorithms/src/snark/marlin/ahp/prover/round_functions/second.rs
```diff
@@ -98,7 +98,7 @@ impl<F: PrimeField, MM: MarlinMode> AHPForR1CS<F, MM> {
 
         let sumcheck_lhs = Self::calculate_lhs(&state, t, summed_z_m, z, *alpha);
 
-        debug_assert!(
+        assert!(
```
ljedrz: note: this was a `debug_assert` presumably for performance reasons; might make sense to revert back to it, unless `evaluate_over_domain_by_ref` became fast enough in the meantime

algorithms/src/snark/plookup/table.rs
```diff
@@ -0,0 +1,40 @@
...
+use snarkvm_fields::Field;
+
+pub struct LookupTable<F: Field>(Vec<Vec<F>>);
+
+impl<F: Field> LookupTable<F> {
+    pub fn new(rows: usize, cols: usize) -> Self {
+        Self(vec![vec![F::zero(); cols]; rows])
+    }
+
+    pub fn push(&mut self, row: usize, col: usize, element: F) {
```
ljedrz: this name might suggest that the table grows in the process; I'd suggest something like `fill` instead

algorithms/src/snark/marlin/ahp/ahp.rs
```diff
@@ -189,26 +189,73 @@ impl<F: PrimeField, MM: MarlinMode> AHPForR1CS<F, MM> {
...
 
         let mut linear_combinations = BTreeMap::new();
 
         // Lincheck sumcheck:
+        let z_a_s = (0..state.batch_size)
+            .map(|i| {
+                let z_a_i = witness_label("z_a", i);
+                LinearCombination::new(z_a_i.clone(), [(F::one(), z_a_i)])
+            })
+            .collect::<Vec<_>>();
```
ljedrz: do these need to be collected? seems like they get iterated over immediately afterwards

ljedrz: ah, I see that they are later reused too; there's still a possibility the allocations can be avoided, but it's not as trivial, so might be best to leave it for later :+1:

algorithms/src/snark/marlin/ahp/indexer/constraint_system.rs
```diff
@@ -74,11 +87,24 @@ impl<F: Field> ConstraintSystem<F> {
...
+        } else {
+            Err(SynthesisError::LookupTableMissing)
+        }
+    }
 }
 
 impl<F: Field> CS<F> for ConstraintSystem<F> {
     type Root = Self;
 
+    fn add_lookup_table(&mut self, lookup_table: LookupTable<F>) -> Result<(), SynthesisError> {
```
Pratyush: Right now this just replaces the existing lookup table; I presume the reason it's named this way is because in the future we want to support multiple lookup tables? If so we should add a TODO comment or something.

jules: yeah thats right! just left it like this for testing mostly - will annotate it

algorithms/src/snark/marlin/ahp/indexer/constraint_system.rs
```diff
@@ -27,6 +34,9 @@ pub(crate) struct ConstraintSystem<F: Field> {
     pub(crate) num_public_variables: usize,
     pub(crate) num_private_variables: usize,
     pub(crate) num_constraints: usize,
+    pub(crate) lookup_table: Option<LookupTable<F>>,
+    pub(crate) s_mul: Vec<F>,
+    pub(crate) s_lookup: Vec<F>,
```
Pratyush: Instead of `s_mul` and `s_lookup` which are `num_constraints`-sized vectors containing `F`-sized elements, we can save memory as well as generalize to a future with multiple lookup tables as follows:
```rust
...
    pub(crate) mul_constraints: HashSet<ConstraintIndex>,
    pub(crate) lookup_constraints: Vec<LookupConstraints<F>>,
}

pub type ConstraintIndex = usize;
pub struct LookupConstraints<F> {
    table: LookupTable<F>,
    indices: HashSet<ConstraintIndex>
}
```

If we know we're going to have at most `u32::MAX` constraints, we can even set `ConstraintIndex = u32` to save space further.

Pratyush: We can construct the selector polynomials using this compactified information later =)

algorithms/src/snark/marlin/ahp/indexer/constraint_system.rs
```diff
@@ -124,6 +150,21 @@ impl<F: Field> CS<F> for ConstraintSystem<F> {
         self.c.push(Self::make_row(&c(LinearCombination::zero())));
 
         self.num_constraints += 1;
+        self.s_mul.push(F::one());
+        self.s_lookup.push(F::zero());
+    }
+
+    fn lookup(&mut self, val: LinearCombination<F>) -> Result<Variable, SynthesisError> {
```
Pratyush: This is a unary lookup, right? I think this suffices for Sinsemillia, but we should also support binary lookups `lookup(a, b) = c`, since those are useful for things like XOR, ANDs, etc.

algorithms/src/snark/marlin/ahp/prover/constraint_system.rs
```diff
@@ -16,14 +16,24 @@
...
 
 pub(crate) struct ConstraintSystem<F: Field> {
     pub(crate) public_variables: Vec<F>,
     pub(crate) private_variables: Vec<F>,
     pub(crate) num_public_variables: usize,
     pub(crate) num_private_variables: usize,
     pub(crate) num_constraints: usize,
+    pub(crate) lookup_table: Option<LookupTable<F>>,
+    pub(crate) s_mul: Vec<F>,
+    pub(crate) s_lookup: Vec<F>,
```
Pratyush: Same suggestion from the `indexer`

algorithms/src/snark/marlin/ahp/prover/round_functions/first.rs
```diff
@@ -74,36 +74,53 @@ impl<F: PrimeField, MM: MarlinMode> AHPForR1CS<F, MM> {
...
         let z_a = state.z_a.take().unwrap();
         let z_b = state.z_b.take().unwrap();
+        let z_c = state.z_c.take().unwrap();
         let private_variables = core::mem::take(&mut state.private_variables);
         assert_eq!(z_a.len(), batch_size);
         assert_eq!(z_b.len(), batch_size);
         assert_eq!(private_variables.len(), batch_size);
         let mut r_b_s = Vec::with_capacity(batch_size);
+        let s_m = state.s_m.clone();
+        let s_l = state.s_l.clone();
```
Pratyush: We should take these by reference.

algorithms/src/snark/marlin/ahp/prover/oracles.rs
```diff
@@ -55,12 +55,22 @@ pub(in crate::snark::marlin) struct SingleEntry<'a, F: PrimeField> {
...
     /// The LDE of `Bz`.
     pub(super) z_b_poly: LabeledPolynomial<F>,
+    /// The LDE of `Cz`.
+    pub(super) z_c_poly: LabeledPolynomial<F>,
+    /// The multiplication constraint selector polynomial.
+    pub(super) s_m_poly: LabeledPolynomial<F>,
+    /// The lookup constraint selector polynomial.
+    pub(super) s_l_poly: LabeledPolynomial<F>,
+    /// The query vector polynomial.
+    pub(super) f_poly: LabeledPolynomial<F>,
```
Pratyush: `s_m_poly` and `s_l_poly` are preprocessed polynomials and should be part of the `Index`, not part of the prover polynomial oracles.

algorithms/src/snark/marlin/ahp/prover/round_functions/first.rs
```diff
@@ -74,36 +74,53 @@ impl<F: PrimeField, MM: MarlinMode> AHPForR1CS<F, MM> {
...
         let mut job_pool = snarkvm_utilities::ExecutionPool::with_capacity(3 * batch_size);
         let state_ref = &state;
-        for (i, (z_a, z_b, private_variables, x_poly)) in
-            itertools::izip!(z_a, z_b, private_variables, &state.x_poly).enumerate()
+        for (i, (z_a, z_b, z_c, private_variables, x_poly, s_m, s_l)) in
+            itertools::izip!(z_a, z_b, z_c, private_variables, &state.x_poly, s_m, s_l).enumerate()
         {
             job_pool.add_job(move || Self::calculate_w(witness_label("w", i), private_variables, x_poly, state_ref));
-            job_pool.add_job(move || Self::calculate_z_m(witness_label("z_a", i), z_a, false, state_ref, None));
+            let z_a_clone = z_a.clone();
```
Pratyush: Why was this clone necessary?

algorithms/src/snark/marlin/ahp/prover/round_functions/first.rs
```diff
@@ -74,36 +74,53 @@ impl<F: PrimeField, MM: MarlinMode> AHPForR1CS<F, MM> {
...
+            job_pool.add_job(move || Self::calculate_z_m(witness_label("z_a", i), z_a_clone, false, state_ref, None));
             let r_b = F::rand(rng);
-            job_pool.add_job(move || Self::calculate_z_m(witness_label("z_b", i), z_b, true, state_ref, Some(r_b)));
+            let z_b_clone = z_b.clone();
```
Pratyush: Ditto

algorithms/src/snark/marlin/ahp/prover/round_functions/first.rs
```diff
@@ -74,36 +74,53 @@ impl<F: PrimeField, MM: MarlinMode> AHPForR1CS<F, MM> {
...
+            job_pool
+                .add_job(move || Self::calculate_z_m(witness_label("z_b", i), z_b_clone, true, state_ref, Some(r_b)));
             if MM::ZK {
                 r_b_s.push(r_b);
             }
+            let z_c_clone = z_c.clone();
```
Pratyush: Ditto

algorithms/src/snark/marlin/ahp/prover/round_functions/first.rs
```diff
@@ -74,36 +74,53 @@ impl<F: PrimeField, MM: MarlinMode> AHPForR1CS<F, MM> {
...
+            job_pool
+                .add_job(move || Self::calculate_z_m(witness_label("z_c", i), z_c_clone, true, state_ref, Some(r_b)));
+            let s_l_clone = s_l.clone();
```
Pratyush: Same

algorithms/src/snark/marlin/ahp/prover/round_functions/first.rs
```diff
@@ -238,12 +255,61 @@ impl<F: PrimeField, MM: MarlinMode> AHPForR1CS<F, MM> {
...
+        state: &prover::State<'a, F, MM>,
+    ) -> PoolResult<'a, F> {
+        let constraint_domain = state.constraint_domain;
+        // let v_H = constraint_domain.vanishing_polynomial();
+        // let should_randomize = MM::ZK && will_be_evaluated;
+        let label = label.to_string();
+        // let poly_time = start_timer!(|| format!("Computing {label}"));
+
+        let evaluations = z_a
+            .iter()
```
Pratyush: This can be made into a `cfg_iter`

algorithms/src/snark/marlin/ahp/indexer/constraint_system.rs
Pratyush: Edited the original comment to change the `Vec<ConstraintIndex>` to a `HashSet<ConstraintIndex>`.

algorithms/src/snark/marlin/ahp/prover/round_functions/first.rs
```diff
@@ -238,12 +255,61 @@ impl<F: PrimeField, MM: MarlinMode> AHPForR1CS<F, MM> {
...
+            )
+            .collect::<Vec<F>>();
+
+        let evals = EvaluationsOnDomain::from_vec_and_domain(evaluations, constraint_domain);
+
+        let poly = evals.interpolate_with_pc_by_ref(state.ifft_precomputation());
+        PoolResult::Query(LabeledPolynomial::new(label, poly, None, None))
+    }
+
+    fn calculate_selector<'a>(
```
Pratyush: This should be done during indexing

algorithms/src/snark/marlin/ahp/prover/round_functions/first.rs
```diff
@@ -238,12 +255,61 @@ impl<F: PrimeField, MM: MarlinMode> AHPForR1CS<F, MM> {
...
+        PoolResult::Selector(LabeledPolynomial::new(label, poly, None, None))
+    }
 }
 
 #[derive(Debug)]
 pub enum PoolResult<'a, F: PrimeField> {
     Witness(LabeledPolynomial<F>),
     MatrixPoly(LabeledPolynomial<F>, LabeledPolynomialWithBasis<'a, F>),
+    Query(LabeledPolynomial<F>),
+    Selector(LabeledPolynomial<F>),
```
Pratyush: Once we move the selector polynomial computation to indexing, we can get rid of this variant.

algorithms/src/snark/marlin/ahp/prover/round_functions/first.rs
```diff
@@ -260,4 +326,18 @@ impl<'a, F: PrimeField> PoolResult<'a, F> {
...
+            _ => None,
+        }
+    }
+
+    fn selector(self) -> Option<LabeledPolynomial<F>> {
+        match self {
+            Self::Selector(poly) => Some(poly),
+            _ => None,
+        }
+    }
```
Pratyush: Ditto

algorithms/src/snark/marlin/ahp/prover/round_functions/second.rs
```diff
@@ -193,7 +194,16 @@ impl<F: PrimeField, MM: MarlinMode> AHPForR1CS<F, MM> {
                     // we rewrite this as  r_i * (z_a * (eta_c * z_b + 1) + eta_b * z_b);
                     // This is better since it reduces the number of required
                     // multiplications by `constraint_domain.size()`.
+                    //
+                    // LOOKUPS EDIT
+                    // turning this back into a simpler lincheck sumcheck to allow easy
+                    // modification of the rowcheck.
                     let mut summed_z_m = {
+                        cfg_iter_mut!(z_b.coeffs).for_each(|b| *b *= eta_b);
```
Pratyush: I suspect we can change this into a single expression like

`let summed_z_m = cfg_iter!(z_a.coeffs).zip(z_b.coeffs).zip(z_c.coeffs).map(|((a, b), c)| a + eta_b * b + eta_c * c).collect();`

algorithms/src/snark/marlin/ahp/prover/state.rs
```diff
@@ -68,6 +68,22 @@ pub struct State<'a, F: PrimeField, MM: MarlinMode> {
...
+    /// The length of this list must be equal to the batch size.
+    pub(super) z_c: Option<Vec<Vec<F>>>,
+
+    /// The list of multiplication selector vectors for each instance in the batch.
+    /// The length of this list must be equal to the batch size.
+    pub(super) s_m: Vec<Vec<F>>,
+
+    /// The list of lookup selector vectors for each instance in the batch.
+    /// The length of this list must be equal to the batch size.
+    pub(super) s_l: Vec<Vec<F>>,
```
Pratyush: Since we support only the same circuit for all entries in a batch, this is shared across the entire batch, and should live only in the index.

algorithms/src/snark/plookup/table.rs
```diff
@@ -0,0 +1,40 @@
...
+// but WITHOUT ANY WARRANTY; without even the implied warranty of
+// MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE. See the
+// GNU General Public License for more details.
+
+// You should have received a copy of the GNU General Public License
+// along with the snarkVM library. If not, see <https://www.gnu.org/licenses/>.
+
+use snarkvm_fields::Field;
+
+pub struct LookupTable<F: Field>(Vec<Vec<F>>);
```
Pratyush: For any circuit of interest, I believe that (at least) the number of columns is statically known. So I believe we can change this to `LookupTable<F: Field, const C: usize>(Vec<[F; C]>);`, which should reduce the number of redirections when accessing this vector.

algorithms/src/snark/plookup/table.rs
```diff
@@ -0,0 +1,40 @@
...
+
+    pub fn push(&mut self, row: usize, col: usize, element: F) {
+        self[row][col] = element;
+    }
+
+    pub fn lookup(&self, row: F) -> F {
+
+
+
+    pub fn compress(&self, zeta: F) -> Vec<F> {
```
Pratyush: Where is this method called? It needs to be called in the prover algorithm, right? Also, for efficiency, it's probably better to take a vector of length `num_columns - 1`.

algorithms/src/snark/marlin/ahp/prover/round_functions/second.rs
jules: So it turns out that `z_a` is actually 1 below the degree of `z_b` and`z_c`. Would simply pushing a `F::zero()` to the end of `z_a` work out in this case or could the degree mismatch fluctuate?

algorithms/src/snark/plookup/table.rs
jules: Yeah, it should be - this was left over from my first approach and I mostly neglected it as I started delving into the PIOP stuff. I'll use it and update the API so the caller provides a `Vec::with_capacity(num_columns - 1)`

algorithms/src/snark/marlin/ahp/prover/round_functions/second.rs
jules: I've added a bit of code that will just fill up `z_a` with zero coefficients until it matches `z_b`, though maybe this might not work in the non-ZK case (since the degree of `z_b` is otherwise just the size of the constraint domain, but i'm assuming this could vary?)

algorithms/src/snark/plookup/table.rs
jules: On second thought - this method is actually quite useless. It serves a good purpose in PLONK-style proof systems since it generates the query vector but our query vector is actually contingent on the generation of the matrix-vector products when initializing the prover. It may make more sense to fully contain the lookup table within the R1CS crate and remove the plookup mod altogether.

algorithms/src/snark/plookup/table.rs
jules: It makes sense - however, this might kind of get in the way when we want to have multiple lookup tables with different column amounts. If that's no issue, then I'll opt for this but otherwise it may be better to stay unconstrained in this regard.

jules: Although from an API point of view it is a bit nicer to constrain here since we won't have to deal with edge cases that come with needing to take in a dynamic-sized vector of lookup values - I'm also assuming we may get some performance gain out of being able to leverage arrays.

In order to save on space, I assume I can just manually implement the serialization-deserialization process to ignore zero elements in tables where the arity is smaller than the constrained size. It will sadly still be inefficient in-memory, but maybe we can find another way around this later.

jules: After quite a bit more experimentation, it seems that adding this const generic has an enormous cascading effect on the required generics for every single type related to calculating SNARKs, including the `SNARK` trait itself (as it uses `ConstraintSynthesizer`s which directly extend `ConstraintSystem`, which contains the `LookupTable<F, C>`).

If this is really desired, I can proceed on finalizing this work, but it has a bad design smell IMO. It may be more convenient to instead opt for an `Indexmap` which will just allow map-like lookups, and constrain key lengths internally.

algorithms/src/snark/marlin/ahp/ahp.rs
```diff
@@ -176,38 +176,66 @@ impl<F: PrimeField, MM: MarlinMode> AHPForR1CS<F, MM> {
...
         let g_1 = LinearCombination::new("g_1", [(F::one(), "g_1")]);
 
         let r_alpha_at_beta = constraint_domain.eval_unnormalized_bivariate_lagrange_poly(alpha, beta);
         let v_H_at_alpha = constraint_domain.evaluate_vanishing_polynomial(alpha);
         let v_H_at_beta = constraint_domain.evaluate_vanishing_polynomial(beta);
         let v_X_at_beta = input_domain.evaluate_vanishing_polynomial(beta);
 
+        let z_a_s_at_beta = z_a_s.iter().map(|z_a| evals.get_lc_eval(z_a, beta)).collect::<Result<Vec<_>, _>>()?;
         let z_b_s_at_beta = z_b_s.iter().map(|z_b| evals.get_lc_eval(z_b, beta)).collect::<Result<Vec<_>, _>>()?;
+        let z_c_s_at_beta = z_c_s.iter().map(|z_c| evals.get_lc_eval(z_c, beta)).collect::<Result<Vec<_>, _>>()?;
```
Pratyush: I think we do not need to have `z_c_s_at_beta`?

algorithms/src/snark/marlin/ahp/ahp.rs
```diff
@@ -176,38 +176,66 @@ impl<F: PrimeField, MM: MarlinMode> AHPForR1CS<F, MM> {
...
+            })
+            .collect::<Vec<_>>();
         let g_1 = LinearCombination::new("g_1", [(F::one(), "g_1")]);
 
         let r_alpha_at_beta = constraint_domain.eval_unnormalized_bivariate_lagrange_poly(alpha, beta);
         let v_H_at_alpha = constraint_domain.evaluate_vanishing_polynomial(alpha);
         let v_H_at_beta = constraint_domain.evaluate_vanishing_polynomial(beta);
         let v_X_at_beta = input_domain.evaluate_vanishing_polynomial(beta);
 
+        let z_a_s_at_beta = z_a_s.iter().map(|z_a| evals.get_lc_eval(z_a, beta)).collect::<Result<Vec<_>, _>>()?;
```
Pratyush: We don't need this either.

algorithms/src/snark/marlin/ahp/ahp.rs
```diff
@@ -176,38 +176,66 @@ impl<F: PrimeField, MM: MarlinMode> AHPForR1CS<F, MM> {
...
         let z_b_s_at_beta = z_b_s.iter().map(|z_b| evals.get_lc_eval(z_b, beta)).collect::<Result<Vec<_>, _>>()?;
+        let z_c_s_at_beta = z_c_s.iter().map(|z_c| evals.get_lc_eval(z_c, beta)).collect::<Result<Vec<_>, _>>()?;
+        let f_s_at_beta = f_s.iter().map(|f| evals.get_lc_eval(f, beta)).collect::<Result<Vec<_>, _>>()?;
+
+        let batch_z_a_at_beta: F =
+            z_a_s_at_beta.iter().zip_eq(batch_combiners).map(|(z_a_at_beta, combiner)| *z_a_at_beta * combiner).sum();
```
Pratyush:
```suggestion
```

algorithms/src/snark/marlin/ahp/ahp.rs
```diff
@@ -176,38 +176,66 @@ impl<F: PrimeField, MM: MarlinMode> AHPForR1CS<F, MM> {
...
         let batch_z_b_at_beta: F =
             z_b_s_at_beta.iter().zip_eq(batch_combiners).map(|(z_b_at_beta, combiner)| *z_b_at_beta * combiner).sum();
+        let batch_z_c_at_beta: F =
+            z_c_s_at_beta.iter().zip_eq(batch_combiners).map(|(z_c_at_beta, combiner)| *z_c_at_beta * combiner).sum();
```
Pratyush:
```suggestion
```

algorithms/src/snark/marlin/ahp/ahp.rs
```diff
@@ -176,38 +176,66 @@ impl<F: PrimeField, MM: MarlinMode> AHPForR1CS<F, MM> {
...
+        let beta = state.third_round_message.unwrap().beta;
         let gamma = state.gamma.unwrap();
+        let zeta = state.zeta.unwrap();
 
         let mut linear_combinations = BTreeMap::new();
 
         // Lincheck sumcheck:
+        let z_a_s = (0..state.batch_size)
+            .map(|i| {
+                let z_a_i = witness_label("z_a", i);
```
Pratyush, line 199: I suggest instead of constructing the linear combination for the rowcheck as `(z_a_beta * z_b_beta - z_c_beta) * s_m`, we should construct it as `(z_a * z_b_beta - z_c) * s_m_beta`. This saves  proof size: for a batch size of `n`,we only need to provide 1 evaluation of `s_m_beta` and `n` evaluations of `z_b_beta`, as opposed to `n` evaluations each of `z_a_beta`, `z_b_beta`, and `z_c_beta`.

algorithms/src/snark/marlin/ahp/ahp.rs
```diff
@@ -217,29 +245,51 @@ impl<F: PrimeField, MM: MarlinMode> AHPForR1CS<F, MM> {
...
-            for (i, (z_b_i_at_beta, combiner)) in z_b_s_at_beta.iter().zip_eq(batch_combiners).enumerate() {
+            for (i, combiner) in batch_combiners.iter().enumerate() {
                 lincheck_sumcheck
-                    .add(r_alpha_at_beta * combiner * (eta_a + eta_c * z_b_i_at_beta), witness_label("z_a", i))
                     .add(-t_at_beta * v_X_at_beta * combiner, witness_label("w", i));
             }
             lincheck_sumcheck
+                .add(r_alpha_at_beta * batch_z_a_at_beta, LCTerm::One)
                 .add(r_alpha_at_beta * eta_b * batch_z_b_at_beta, LCTerm::One)
+                .add(r_alpha_at_beta * eta_c * batch_z_c_at_beta, LCTerm::One)
```
Pratyush: These should be replaced with `sum(r_alpha_at_beta * eta_m * z_m)`

algorithms/src/snark/marlin/ahp/ahp.rs
```diff
@@ -217,29 +245,51 @@ impl<F: PrimeField, MM: MarlinMode> AHPForR1CS<F, MM> {
...
         let lincheck_sumcheck = {
-            let mut lincheck_sumcheck = LinearCombination::empty("lincheck_sumcheck");
+            let mut lincheck_sumcheck = LinearCombination::empty("placeholder");
+            // Add rowcheck
+            let mut rowcheck = LinearCombination::empty("lincheck_sumcheck");
+            for (i, combiner) in batch_combiners.iter().enumerate() {
+            rowcheck
+                .add((z_a_s_at_beta[i] * z_b_s_at_beta[i] - z_c_s_at_beta[i]) * combiner, "s_m")
+                .add((z_a_s_at_beta[i] + zeta * z_b_s_at_beta[i] + zeta_squared
+                    * z_c_s_at_beta[i] - f_s_at_beta[i]) * combiner, "s_l");
```
Pratyush: I suggest instead of constructing the linear combination for the rowcheck as `(z_a_beta * z_b_beta - z_c_beta) * s_m`, we should construct it as `(z_a * z_b_beta - z_c) * s_m_beta`. This saves  proof size: for a batch size of `n`,we only need to provide 1 evaluation of `s_m_beta` and `n` evaluations of `z_b_beta`, as opposed to `n` evaluations each of `z_a_beta`, `z_b_beta`, and `z_c_beta`. This results in a savings of ~2n field elements.

Similarly for the lookup part, we should do it as `(z_a + zeta * z_b + zeta * z_c) * s_l_beta`.

algorithms/src/snark/marlin/ahp/indexer/constraint_system.rs
```diff
@@ -123,7 +141,18 @@ impl<F: Field> CS<F> for ConstraintSystem<F> {
         self.b.push(Self::make_row(&b(LinearCombination::zero())));
         self.c.push(Self::make_row(&c(LinearCombination::zero())));
 
+        self.mul_constraints.insert(self.num_constraints);
+        self.num_constraints += 1;
+    }
+
+    fn lookup(&mut self, _: &[LinearCombination<F>], table_index: usize) -> Result<Variable, SynthesisError> {
+        self.a.push(Self::make_row(&LinearCombination::zero()));
```
Pratyush: Shouldn't the various `self.m.push(...)` be identical to `enforce` above?

algorithms/src/snark/marlin/ahp/indexer/constraint_system.rs
```diff
@@ -123,7 +141,18 @@ impl<F: Field> CS<F> for ConstraintSystem<F> {
         self.b.push(Self::make_row(&b(LinearCombination::zero())));
         self.c.push(Self::make_row(&c(LinearCombination::zero())));
 
+        self.mul_constraints.insert(self.num_constraints);
+        self.num_constraints += 1;
+    }
+
+    fn lookup(&mut self, _: &[LinearCombination<F>], table_index: usize) -> Result<Variable, SynthesisError> {
```
Pratyush: I think the API for this should be similar to the one for `enforce`: we should take in three linear combinations `a, b, c`, and check that the table contains the triple `(az, bz, cz)`. So the API would look like `enforce_lookup(a, b, c)`.

We could additionally keep the existing API of `c = lookup(a, b)` as sugar atop the core API.

algorithms/src/snark/marlin/ahp/prover/constraint_system.rs
```diff
@@ -56,11 +70,34 @@ impl<F: Field> ConstraintSystem<F> {
         make_matrices_square(self, num_variables);
         assert_eq!(self.num_public_variables + self.num_private_variables, self.num_constraints, "padding failed!");
     }
+
+    fn lookup(&mut self, key: &[LinearCombination<F>], table_index: usize) -> Result<F, SynthesisError> {
```
Pratyush: To confirm, this is a "native" lookup, right?

algorithms/src/snark/marlin/ahp/prover/round_functions/mod.rs
```diff
@@ -30,21 +30,22 @@ use snarkvm_utilities::println;
 #[cfg(feature = "parallel")]
 use rayon::prelude::*;
 
+mod fifth;
 mod first;
 mod fourth;
 mod second;
 mod third;
```
Pratyush: Nit: let's sort this by round (if rustfmt re-orders things, let's use a `#[rustfmt::skip]`)

algorithms/src/snark/marlin/ahp/prover/round_functions/third.rs
```diff
@@ -14,222 +14,89 @@
...
-        });
-
-        let [(sum_a, lhs_a, g_a), (sum_b, lhs_b, g_b), (sum_c, lhs_c, g_c)]: [_; 3] =
-            pool.execute_all().try_into().unwrap();
-
-        let msg = prover::ThirdMessage { sum_a, sum_b, sum_c };
-        let oracles = prover::ThirdOracles { g_a, g_b, g_c };
-        state.lhs_polynomials = Some([lhs_a, lhs_b, lhs_c]);
-        state.sums = Some([sum_a, sum_b, sum_c]);
+            .expect("prover::State should include verifier_first_msg when prover_fourth_round is called");
```
Pratyush:
```suggestion
            .expect("prover::State should include verifier_first_msg when prover_third_round is called");
```

algorithms/src/snark/marlin/ahp/prover/round_functions/third.rs
```diff
@@ -14,222 +14,89 @@
...
+            })
+            .sum::<DensePolynomial<F>>();
+
+        let mut lin = state.lin.as_ref().unwrap().clone();
+        cfg_iter_mut!(lin.coeffs).for_each(|c| *c *= delta);
+        let mut x_g_1 = state.x_g_1.as_ref().unwrap().clone();
+        cfg_iter_mut!(x_g_1.coeffs).for_each(|c| *c *= delta);
+
+        let q_1 = &(&row + &lin) - &x_g_1;
+        let (h_1, rem) = q_1.divide_by_vanishing_poly(constraint_domain).unwrap();
```
Pratyush: We can just save `_h_1` from the previous round and use it here.

algorithms/src/snark/marlin/ahp/prover/state.rs
```diff
@@ -83,6 +91,12 @@ pub struct State<'a, F: PrimeField, MM: MarlinMode> {
     /// The challenges sent by the verifier in the first round
     pub(super) verifier_first_message: Option<verifier::FirstMessage<F>>,
 
+    /// The lincheck sumcheck polynomial produced in the second round.
+    pub(super) lin: Option<DensePolynomial<F>>,
+
+    /// The Xg_1(X) polynomial produced in the second round.
+    pub(super) x_g_1: Option<DensePolynomial<F>>,
```
Pratyush: We *should* be able to replace this with `_h_1`

algorithms/src/snark/marlin/ahp/prover/round_functions/mod.rs
jules: to do this we'd either have to bury them in a separate mod (so that we can scope the directive), or tag each line with `#[rustfmt::skip]` - if there is no problem adding an extra mod layer then i'll go for that!

algorithms/src/snark/marlin/ahp/prover/constraint_system.rs
jules: yeah!

julesdesmit:
NOTE: a lot of the code currently being pushed is heavily unoptimized - this will change in the near future once the protocol itself is functioning and I can redirect my focus towards cutting corners. Leaving it in this state currently makes it a little easier to reason with since it's more verbose.

algorithms/src/snark/marlin/ahp/indexer/constraint_system.rs
jules: yep, good catch! seems i was just pushing zeros this whole time

jules:
This work should now be mostly wrapped up, cleaned and optimized for as far as I can manage naively. Next up to finish this will be to make it fit back into the CI properly (some integration stuff fails because proving keys have changed and they will no longer deserialize properly), but since there is ongoing work on the parameter infrastructure, I'll leave this PR to be revisited when appropriate.

PR #876, opened 2022-07-07
Fixes bug in div_wrapped

d0cd:
This PR
- Fixes a bug in div_wrapped.
- Redefines shl_checked as an arithmetic shift left.
- Annotates necessary uses of `E::halt`
- Uses witness macro in place of eject/inject 
for some integer circuits.


circuit/types/integers/src/shl_checked.rs
```diff
@@ -68,27 +68,16 @@ impl<E: Environment, I: IntegerType, M: Magnitude> ShlChecked<Integer<E, M>> for
...
-
-            // Determine if any of the upper bits of the RHS are nonzero.
-            let is_nonzero = rhs.bits_le[trailing_zeros_index..].iter().fold(Boolean::constant(false), |a, b| a | b);
-
-            // Ensure the upper bits of the RHS are zero.
-            E::assert(!is_nonzero);
+        let two = Self::one() + Self::one();
+        let base = match I::is_signed() {
+            // If `self` is a signed integer and is negative, then negate the base.
+            true => Self::ternary(&self.is_less_than(&Self::zero()), &(Self::zero() - &two), &two),
```
howardwu: If we introduce `is_negative` and `is_positive` trait methods, we can further reduce the cost of these `self.is_less_than(&Self::zero())` checks.

For signed cases, we can inspect that `self.msb().is_one()` when `I::is_signed()` for `is_negative`, and noop for `!I::is_signed()`. A similar case can be done for `is_positive`.

In Rust (and in math), 0 is neither positive nor negative. So we do need to perform a `.fold()` to ensure the case that all bits are 0 is `false` for `is_positive`.

d0cd: Agreed and thanks for the catch! Optimizing this implementation directly.

howardwu (approved):

PR #894, opened 2022-07-20
Unifies `is_reserved_opcode`

howardwu:
## Motivation

Unifies `is_reserved_opcode`.

This PR builds upon the prior PR https://github.com/AleoHQ/snarkVM/pull/893


howardwu:
There remains some more Clippy lints, which should be addressed

PR #899, opened 2022-07-23
Faster `sqrt` impl

jules:


howardwu:
Screen Shot 2022-07-28 at 4 38 33 PM
^This is the pseudocode algorithm for reference.

howardwu:
@ljedrz can you check the correctness of the Rust syntax and API usage (i.e. we use `.floor()` and `.sqrt()` among many operators) for edge cases that we have not have contemplated?

fields/src/macros.rs
```diff
@@ -86,55 +86,81 @@ macro_rules! impl_primefield_from_int {
...
+                let k_2_field = $Self::from(k_2 as u64);
+                let k_1 = k - k_2;
+                let l_minus_one_times_k = n_field - $Self::one() - k_2_field;
+                let l_minus_one = l_minus_one_times_k / k_field;
+                let l = l_minus_one + $Self::one();
+                let mut l_s: Vec<$Self> = Vec::with_capacity(k);
+
+                for _ in 0..k_1 {
+                    l_s.push(l_minus_one);
+                }
```
ljedrz:
```suggestion
                l_s.resize(l_s.len() + k_1, l_minus_one);
```

fields/src/macros.rs
```diff
@@ -86,55 +86,81 @@ macro_rules! impl_primefield_from_int {
...
+                for _ in 0..k_2 {
+                    l_s.push(l);
```
ljedrz: ditto

fields/src/macros.rs
```diff
@@ -86,55 +86,81 @@ macro_rules! impl_primefield_from_int {
...
+                let n = $P::TWO_ADICITY;
+                // `T` is equivalent to `m` in the paper.
+                let v = $self.pow($P::T_MINUS_ONE_DIV_TWO);
+                let x = $self.pow($P::T);
+                let g = $Self::two_adic_root_of_unity().pow($P::T);
+                let two = $Self::from(2u8);
+                let n_field = $Self::from(n);
+
+                let k = ((n - 1) as f64).sqrt().floor() as usize;
+                let k_field = $Self::from(k as u64);
```
ljedrz, line 103:
```suggestion
                let k = ((n - 1) as f64).sqrt().floor() as u64;
                let k_field = $Self::from(k);
```

fields/src/macros.rs
```diff
@@ -86,55 +86,81 @@ macro_rules! impl_primefield_from_int {
...
+                let k_2 = ((k as f64) / 2.0).floor() as usize;
+                let k_2_field = $Self::from(k_2 as u64);
```
ljedrz, line 105:
```suggestion
                let k_2 = ((k as f64) / 2.0).floor() as u64;
                let k_2_field = $Self::from(k_2);
```

fields/src/macros.rs
```diff
@@ -86,55 +86,81 @@ macro_rules! impl_primefield_from_int {
...
+                    l_s.push(l);
                 }
 
-                while !b.is_one() {
-                    let mut k = 0usize;
+                let mut x_s: Vec<$Self> = Vec::with_capacity(k);
+                for i in 0..k {
+                    x_s.push(x.pow(
+                        two.pow((n_field - $Self::one() - (l_s.iter().take(i + 1).sum::<$Self>())).to_repr()).to_repr(),
+                    ));
```
ljedrz: some extra bindings would help readability here

fields/src/macros.rs
```diff
@@ -86,55 +86,81 @@ macro_rules! impl_primefield_from_int {
...
-                    z = w.square();
-                    b *= &z;
-                    x *= &w;
-                    v = k;
+                    s
+                };
+
+                let mut s = $Self::zero();
+                let mut t = $Self::zero();
+                for i in 0..k {
```
ljedrz: it would be best to work on a `l_s.iter().zip(x_s.iter())` here to remove index checking

ljedrz:
Left some suggestions; as long as subtractions and `pow`s don't overflow, I don't see any obvious footguns here. As with any arithmetic implementations, I'd recommend using something like [quickcheck](https://github.com/BurntSushi/quickcheck) to verify that the results match the expectations.

fields/src/macros.rs
ljedrz: also, the subsequent sums could be memoized, so that only a single addition is performed with each iteration

jules:
@ljedrz I've managed to squeeze out a lot of optimization, and managed around a 4-10% decrease in time taken to compute square roots (depending on field size, with `Fr` having the best decrease). Maybe you might find some more optimizations I can apply to shave it down even further.

More generally, the algorithm will become even faster if lookups are added. Since the theoretical improvement applies to both TS and Sarkar, we will see a more significant improvement by applying it here, so I think it may be worthwhile to continue working on this with the intention of merging it in.

fields/src/macros.rs
```diff
@@ -86,55 +86,106 @@ macro_rules! impl_primefield_from_int {
...
-
-                    let j = v - k - 1;
-                    w = z;
-                    for _ in 0..j {
-                        w.square_in_place();
+                    s
+                };
+
+                let calc_kappa = |i: usize, j: usize, l_s: &Vec<u64>| -> u64 {
+                    l_s.iter().take(j).fold(1, |acc, x| acc + x) + l_s.iter().skip(i + 1).fold(0, |acc, x| acc + x)
```
ljedrz:
```suggestion
                let calc_kappa = |i: usize, j: usize, l_s: &[u64]| -> u64 {
                    l_s.iter().take(j).sum::<u64>() + 1 + l_s.iter().skip(i + 1).sum::<u64>()
```

fields/src/macros.rs
```diff
@@ -86,55 +86,106 @@ macro_rules! impl_primefield_from_int {
...
+                };
+
+                let calc_gamma = |i: usize, q_s: &Vec<Vec<bool>>, last: bool| -> $Self {
```
ljedrz:
```suggestion
                let calc_gamma = |i: usize, q_s: &[Vec<bool>], last: bool| -> $Self {
```

jules:
For now we've decided to forgo the lookup variant of the Sarkar algorithm as the speedup may not be worth it.

howardwu (approved):

PR #908, opened 2022-07-28
Fixes bug in `shl_checked` circuit

d0cd:
This PR
- Fixes a bug in the `shl_checked` circuit.

For reference, the specification for `shl_checked` is given below.
Suppose we have `a.shl_checked(b)`, where a is an integer of size `N`. 
The operation must enforce that:
- `b` < `N`.
- the **result** of `a * 2 ^ b` is a valid integer.

One noteworthy case is `-1i8 << 7u8`. 
In this case we have that `2i8 ^ 7u8 = 128i8`, which is not a valid i8.
However the result of the operation is `-128i8`, which is a valid i8.
The operation must not error on this case.

### DIFF
--- a/circuit/types/integers/src/shl_checked.rs
+++ b/circuit/types/integers/src/shl_checked.rs
@@ -165,14 +165,7 @@ mod tests {
         let a = Integer::<Circuit, I>::new(mode_a, first);
         let b = Integer::<Circuit, M>::new(mode_b, second);
 
-        let two = I::one().wrapping_add(&I::one());
-        let (first, base) = if I::is_signed() {
-            if (*first).lt(&I::zero()) { (first.wrapping_abs(), I::zero().wrapping_sub(&two)) } else { (*first, two) }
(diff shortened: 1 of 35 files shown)


circuit/types/integers/src/shl_checked.rs
```diff
@@ -194,11 +207,17 @@ mod tests {
...
+        mode_a: Mode,
+        mode_b: Mode,
+    ) {
         for i in 0..ITERATIONS {
             let first = Uniform::rand(&mut test_rng());
             let second = Uniform::rand(&mut test_rng());
 
+            let name = format!("Shl Zero: {} << {} {}", mode_a, mode_b, i);
+            check_shl::<I, M>(&name, first, second, mode_a, mode_b);
+
```
howardwu: What's going on here? It's a duplicate

d0cd: Resolved.

PR #925, opened 2022-08-02
Remove allocations from the Map iterators

ljedrz:
This PR removes all the allocations related to iteration temporarily introduced alongside `Cow` in the `Map`, using a few wrapper objects and extra helper methods.

### DIFF
--- a/vm/compiler/src/ledger/block/transactions/mod.rs
+++ b/vm/compiler/src/ledger/block/transactions/mod.rs
@@ -118,7 +118,7 @@ impl<N: Network> Transactions<N> {
 
     /// Returns an iterator over all transitions.
     pub fn transitions(&self) -> impl '_ + Iterator<Item = &Transition<N>> {
-        self.transactions.values().flat_map(Transaction::transitions)
+        self.transactions().flat_map(Transaction::transitions)
     }
 
     /// Returns a consuming iterator over all transitions.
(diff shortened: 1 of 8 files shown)


howardwu:
Why is this approach necessary? Is there no other way to do it upstream?

howardwu:
What does this approach remove in terms of allocations?

ljedrz:
> Why is this approach necessary?

The first commit is not necessary, but it should be helpful with the persistent storage impl.

> Is there no other way to do it upstream?

If you're worried about the number of extra methods, they can be hidden from documentation etc. They would also not be needed if the related object members were not private.

> What does this approach remove in terms of allocations?

In the diff you'll notice the removal of some calls to `to_owned` and `collect`, which would allocate in the persistent impl; this PR removes them all, making iterations almost as fast as in the in-memory impl (with key deserialization being the only perf penalty).

ljedrz:
An additional bit of rationale:
- with the storage being either in-memory or persistent _and_ working on end types `K` and `V` (as opposed to some `enum` composed of an object or raw bytes), the former doesn't require deserialization during iteration, and the latter does
- while in-memory objects can be iterated over by reference, the deserialized ones can't, as they are deserialized on demand during iteration and don't persist as `K` and `V` objects unless passed by value
- `Cow` allows borrowed and owned objects to be handled together (including being iterated over), which is the natural solution here
- the initial `Cow` implementation used some calls to `collect` and `to_owned` as a temporary workaround, but this meant that all the affected iterations incurred needless _double_ (cloning + collecting) allocations, which slow them down
- this PR removes them both: adding the `into_` methods removes the needless clones (from references to owned values), while adding the wrapper `enum`s for iteration purposes removes the need for collecting (due to different iterators being returned from the 2 branches)
- other than the extra required code there are no downsides to this solution

ljedrz:
Rebased, and added one commit that followed the convention from the recent changes to `Transactions`.

ljedrz:
Rebased.

ljedrz:
Closing, as https://github.com/AleoHQ/snarkVM/pull/944 integrated it.

PR #934, opened 2022-08-03
[CU-383p3pb] Fix build skipping when program code has changed

facundoolano:
## Motivation

I noticed when trying the aleo cli that after updating my instructions code, I needed to run `aleo clean` to wipe the build dir for changes to pick up. Otherwise the build would be skipped but I would get an error e.g. when running a new function, because the verifier was missing:

```
facundo@entropy ~/d/e/p/aleo_instructions aleo run hello 1u32

🚀 Executing 'foo.aleo/hello'...

 • Calling 'foo.aleo/hello'...
 • Executed 'hello' (in 4656 ms)

➡️  Output

 • 120u32

✅ Executed 'foo.aleo/hello' (in "/Users/facundo/dev/entropy1729/platform-tools/aleo_instructions")

# (updated my main.aleo with a new function here)

facundo@entropy ~/d/e/p/aleo_instructions $ aleo run another 1u32

🚀 Executing 'foo.aleo/another'...

⚠️  The prover file is missing: '/Users/facundo/dev/entropy1729/platform-tools/aleo_instructions/build/another.prover'

aleo build

✅ Built 'foo.aleo' (in "/Users/facundo/dev/entropy1729/platform-tools/aleo_instructions")

facundo@entropy ~/d/e/p/aleo_instructions $ aleo clean

✅ Cleaned the build directory (in "/Users/facundo/dev/entropy1729/platform-tools/aleo_instructions/build")

facundo@entropy ~/d/e/p/aleo_instructions $ aleo build

⏳ Compiling 'foo.aleo'...

 • Loaded universal setup (in 2611 ms)
 • Built 'hello' (in 15228 ms)
 • Built 'another' (in 15899 ms)
```

## Test Plan

I tested this manually by repeating the steps that were producing the error before, but we need to add unit tests to the is_build_required module before merging this.

### DIFF
--- a/vm/package/is_build_required.rs
+++ b/vm/package/is_build_required.rs
@@ -39,34 +39,32 @@ impl<N: Network> Package<N> {
             Err(_) => return true,
         };
 
-        // Initialize a boolean indicator if we need to build the circuit.
-        let mut is_complete = true;
-
         // Check if the program ID in the manifest matches the program ID in the AVM file.
-        if avm_file.program().id() == &self.program_id {
-            // Retrieve the main program.
-            let program = self.program();
+        if avm_file.program().id() != &self.program_id {
+            return true;
+        }
 
-            // Check if the program matches.
-            if avm_file.program() == program {
-                // Next, check if the prover and verifier exist for each function.
-                for function_name in program.functions().keys() {
-                    // Check if the prover file exists.
-                    if !ProverFile::exists_at(&build_directory, function_name) {
-                        // If not, we need to build the circuit.
-                        is_complete = false;
-                        break;
-                    }
-                    // Check if the verifier file exists.
-                    if !VerifierFile::exists_at(&build_directory, function_name) {
-                        // If not, we need to build the circuit.
-                        is_complete = false;
-                        break;
-                    }
-                }
+        // Check if the main program matches.
+        let program = self.program();
+        if avm_file.program() != program {
+            return true;
+        }
+
+        // Next, check if the prover and verifier exist for each function.
+        for function_name in program.functions().keys() {
+            // Check if the prover file exists.
+            if !ProverFile::exists_at(&build_directory, function_name) {
+                // If not, we need to build the circuit.
+                return true;
+            }
+            // Check if the verifier file exists.
+            if !VerifierFile::exists_at(&build_directory, function_name) {
+                // If not, we need to build the circuit.
+                return true;
             }
         }
 
-        !is_complete
+        // Package hasn't changed, no need to build
+        return false;
     }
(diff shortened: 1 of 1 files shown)


vm/package/is_build_required.rs
```diff
@@ -39,34 +39,32 @@ impl<N: Network> Package<N> {
...
         };
 
-        // Initialize a boolean indicator if we need to build the circuit.
-        let mut is_complete = true;
-
         // Check if the program ID in the manifest matches the program ID in the AVM file.
-        if avm_file.program().id() == &self.program_id {
-            // Retrieve the main program.
-            let program = self.program();
+        if avm_file.program().id() != &self.program_id {
```
howardwu, line 43: Are we sure this is correct? If the program ID doesn't match, I believe we shouldn't be building because it is the wrong program.

facundoolano: I see, I may be lacking some context here, I thought this was something like a lighter version of the "has this program changed" check. Is there a legitimate use case for skip building on different program id, or is it more like an invalid condition where the cli should throw an error?

howardwu: The latter. Your intuition is correct here. This is like another sanity checking spot, which should never technically be hit . I suppose we could just remove the check?

facundoolano: Yeah I think so. I'll take a look at the sorrounding code and run some checks to be sure, but if it looks good I'll just remove it.

ncontinanza: I tried @facundoolano's changes in the roulette project removing this check.
When the program IDs match, the result is the expected:
```$ aleo build

⏳ Compiling 'bets.aleo'...

 • Loaded universal setup (in 1610 ms)
 • Built 'psd_hash' (in 6252 ms)
 • Built 'mint_casino_token_record' (in 11657 ms)
 • Built 'make_bet' (in 26166 ms)
 • Built 'psd_bits_mod' (in 5980 ms)

✅ Built 'bets.aleo' (in "/Users/ncontinanza/Projects/Entropy1729/aleo_roulette/circuits/bets")
```

When the IDs do not match (I edited the ID in the program.json file), we get a descriptive error:
```$ aleo build

⚠️  The program ID from `program.json` does not match in '/Users/ncontinanza/Projects/Entropy1729/aleo_roulette/circuits/bets/main.aleo'
```

facundoolano: Nice we can remove the check then

facundoolano:
Closing this one in favor of https://github.com/Entropy1729/snarkVM/pull/3

PR #946, opened 2022-08-05
Use an IndexMap inside the MemoryMap

ljedrz:
This change has the benefit of allowing to index the `MemoryMap` by keys like the block height, (making lookups faster if we tweak the API), and making the iteration order deterministic.


howardwu (approved):

howardwu:
LGTM

PR #957, opened 2022-08-12
[Tracking] Testnet3 security updates 

raychu86:
## Motivation

This PR will track all the in progress and completed security related PRs.

The initial plan is to merge all the security fixes into testnet3-audit before migrating them into the testnet3 branch of snarkVM.

Tracker:

- [ ] #958
- [ ] #959
- [ ] #960 
- [ ] #974
- [ ] #975 
- [ ] #976
- [ ] #977

PR #959, opened 2022-08-12
[Testnet3] Add `is_on_curve` check for `from_coordinates` call

raychu86:
## Motivation

This PR checks that the short Weierstrass Jacobian affine element is on the curve when constructed via `from_coordinates`. This check is already done in the Twisted Edwards variant of `from_coordinates`.

Tracking PR: #957


Pratyush:
I think we should also check subgroup membership here (and also in the TE case). If we know for certain that something is in the subgroup, then we should add a `from_coordinates_unchecked` method that omits the subgroup check.

howardwu (approved):

PR #967, opened 2022-08-15
feat: introduce atomic batch writes

howardwu:
## Motivation

feat: introduce atomic batch writes

### DIFF
--- a/vm/compiler/src/ledger/map/memory_map.rs
+++ b/vm/compiler/src/ledger/map/memory_map.rs
@@ -107,6 +107,14 @@ impl<
     }
 
     ///
+    /// Aborts the current atomic operation.
+    ///
+    fn abort_atomic(&self) {
+        self.atomic_batch.lock().clear();
+        self.batch_in_progress.store(false, Ordering::SeqCst);
+    }
+
+    ///
     /// Finishes an atomic operation, performing all the queued writes.
     ///
     fn finish_atomic(&self) {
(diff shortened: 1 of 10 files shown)


howardwu:
@ljedrz Can we add a unit test for `MemoryMap` with the atomic op?

Also clippy may have some lint requests.

howardwu:
I have one request and two questions.

Request: can we add an `abort_atomic` method as well?

Question 1: what happens if a method `return Err(..)` in between `start_atomic` and `finish_atomic`? Is it possible to rollback/abandon the unsaved changes when leaving the scope? I presume we may need an object with some drop logic to perform this.

Question 2: is it possible to continue 1 atomic scope from one method into another? For example:
```rs
fn foo() {
    start_atomic();
    bar();
    end_atomic();
}
fn bar() {
    // do stuff atomically still
}
```
There is a legitimate use case for this if it is possible.

ljedrz:
Rebased, fixed the related `clippy` lint, added `Map::abort_atomic` and 2 unit tests.

> Question 1: what happens if a method return Err(..) in between start_atomic and finish_atomic? Is it possible to rollback/abandon the unsaved changes when leaving the scope? I presume we may need an object with some drop logic to perform this.

It could be done via `Drop`, but a new object would be required, and it could accidentally be misused (e.g. when bound to `_` or when done in some extra helper scope). The best way to handle this would IMO be to have a wrapper function/method, so that the following could be performed:
```
impl Ledger {
    (...)

    fn some_method(...) {
        self.start_atomic();

        if let Err(e) = self.some_method_apply_storage_changes(...) {
            self.abort_atomic();
            return Err(e);
        }

        self.finish_atomic();
    }
}
```

> Question 2: is it possible to continue 1 atomic scope from one method into another? For example:

Yes; right now `Map::start_atomic` persists until `Map::{abort_atomic, finish_atomic}` is called, regardless of what else happens in the meantime.

PR #971, opened 2022-08-17
Adds subgroup check at initialization

### LINKED ISSUE #805: [Feature] Add cofactor check to `Group::new`
## 🚀 Feature

```rs
struct Group {
    fn new(point: affine) -> Self {
        let output: Group = CS::alloc(point * COFACTOR_INV);
        Self::on_curve_check(output);

        // let output = output * Scalar::constant(COFACTOR);
        let output = output.double().double();

        output
    }
}
```

howardwu:
## Motivation

Closes #805.

Adds a subgroup check at initialization for `Group` circuits.


console/types/group/src/lib.rs
```diff
@@ -90,10 +90,6 @@ impl<E: Environment> Group<E> {
 
     /// Returns `self / COFACTOR`.
     pub fn div_by_cofactor(&self) -> Self {
-        // (For advanced users) The cofactor for this curve is `4`. Thus doubling is used to be performant.
```
howardwu, line 94: This was deleted because it was a copypasta from `mul_by_cofactor` (the method above).

The `mul_by_cofactor_inv` does not use this trick/assumption (it does the full `mul_bits`).

circuit/types/group/src/helpers/from_xy_coordinates.rs
```diff
@@ -20,28 +20,8 @@ impl<E: Environment> Group<E> {
...
-            let second = (d * &x2) - &Field::one();
-            let third = (a * x2) - Field::one();
-
-            // Ensure y^2 * (dx^2 - 1) = (ax^2 - 1).
-            E::enforce(|| (first, second, third));
-        }
-
-        Self { x, y }
+        // Recover point from the `(x, y)` coordinates as a witness.
+        witness!(|x, y| console::Group::from_xy_coordinates(x, y))
```
howardwu, line 24: `witness!` does the following:
1. It ejects the `mode` of `x` and `y`, and determines whether to allocate the recovered `point` as `Mode::Constant` or `Mode::Private`
2. It then calls `circuit::Group::new(mode, point)`

circuit/types/group/src/lib.rs
```diff
@@ -57,12 +57,43 @@ impl<E: Environment> Inject for Group<E> {
...
+        // Inject the point.
+        let point = {
+            // Initialize the (x, y) coordinates of the point as field elements.
+            let (x, y) = group.to_xy_coordinate();
+            // Inject the `(x, y)` coordinates as field elements.
+            Self { x: Field::new(mode, x), y: Field::new(mode, y) }
+        };
+
+        // Inject `(point / COFACTOR)` as a witness.
+        let point_inv: Self = witness!(|point| point.div_by_cofactor());
```
howardwu: `witness!` does the following:
1. It ejects the `mode` of `point`, to determine whether to allocate `point_inv` as `Mode::Constant` or `Mode::Private`
2. It computes `point_inv` as `point.div_by_cofactor()` natively (via `console`)
3. It allocates `point_inv` as `circuit::Group::new(mode, point_inv)`

circuit/types/group/src/lib.rs
```diff
@@ -57,12 +57,43 @@ impl<E: Environment> Inject for Group<E> {
...
+        //
+        // Ensure ax^2 + y^2 = 1 + dx^2y^2
+        // by checking that y^2 * (dx^2 - 1) = (ax^2 - 1)
+        //
+        {
+            let a = Field::constant(console::Field::new(E::AffineParameters::COEFF_A));
+            let d = Field::constant(console::Field::new(E::AffineParameters::COEFF_D));
+
+            let x2 = point_inv.to_x_coordinate().square();
+            let y2 = point_inv.to_y_coordinate().square();
```
howardwu: Observe we check `point_inv` (not `point`) is on the curve here

circuit/types/group/src/lib.rs
Pratyush: We don't need to separately allocate both `point_inv` and `point`; we just allocate `point_inv`, check that the result is on the curve, and then output `point_inv.mul_by_cofactor()`. This avoids extra allocations and also the equality check.

Pratyush: Doesn't the call to `circuit::Group::new(mode, point_inv)` lead to an infinite loop?

Pratyush: nvm, I see that you replaced it with individually allocations of the coordinates.

Pratyush (approved):

circuit/types/group/src/lib.rs
howardwu: Yeah, it was my bad there :)

Pratyush:
Should we merge this?

PR #973, opened 2022-08-17
Adds atomic on the map stores

howardwu:
## Motivation

Adds atomic on the map stores

### DIFF
--- a/vm/compiler/src/ledger/store/block/mod.rs
+++ b/vm/compiler/src/ledger/store/block/mod.rs
@@ -163,26 +163,32 @@ pub trait BlockStorage<N: Network>: Clone + Sync {
             None => bail!("Failed to remove block: missing transactions for block '{height}' ('{block_hash}')"),
         };
 
+        // Start an atomic batch write operation.
+        self.start_atomic();
+
         // Remove the block hash.
-        self.id_map().remove(&height)?;
+        self.id_map().remove(&height).or_abort(|| self.abort_atomic())?;
         // Remove the block height.
-        self.reverse_id_map().remove(block_hash)?;
+        self.reverse_id_map().remove(block_hash).or_abort(|| self.abort_atomic())?;
         // Remove the block header.
-        self.header_map().remove(block_hash)?;
+        self.header_map().remove(block_hash).or_abort(|| self.abort_atomic())?;
 
         // Remove the transaction IDs.
-        self.transactions_map().remove(block_hash)?;
+        self.transactions_map().remove(block_hash).or_abort(|| self.abort_atomic())?;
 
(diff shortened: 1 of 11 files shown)


vm/compiler/src/ledger/store/block/mod.rs
```diff
@@ -83,6 +83,116 @@ pub trait BlockStorage<N: Network>: Clone + Sync {
...
+        }
+
+        // Store the block signature.
+        self.signature_map().insert(block.hash(), *block.signature()).or_abort(|| self.abort_atomic())?;
+
+        // Finish the atomic batch write operation.
+        self.finish_atomic();
+
+        Ok(())
+    }
```
ljedrz, line 151: alternatively, these functions could look as follows (example for `BlockStorage::insert`):
```
/// Stores the given `block` into storage.
fn insert(&self, block: &Block<N>) -> Result<()> {
    // Start an atomic batch write operation.
    self.start_atomic();

    let atomic_ops = || -> Result<()> {
        // Store the block hash.
        self.id_map().insert(block.height(), block.hash())?;
        // Store the block height.
        self.reverse_id_map().insert(block.hash(), block.height())?;
        // Store the block header.
        self.header_map().insert(block.hash(), *block.header())?;

        // Store the transaction IDs.
        self.transactions_map().insert(block.hash(), block.transaction_ids().copied().collect())?;

        // Store the block transactions.
        for transaction in block.transactions().values() {
            // Store the reverse transaction ID.
            self.reverse_transactions_map().insert(transaction.id(), block.hash())?;
            // Store the transaction.
            self.transaction_store().insert(transaction)?;
        }

        // Store the block signature.
        self.signature_map().insert(block.hash(), *block.signature())?;

        Ok(())
    };

    if atomic_ops().is_err() {
        self.abort_atomic();
    }

    // Finish the atomic batch write operation.
    self.finish_atomic();

    Ok(())
}
```
This would obviate `OrAbort`.

ljedrz:
Since the atomic operations execute on all the underlying objects, it would suffice that only the `Ledger`-level operations use them (as long as the `Ledger` is the only point where the storage operations are performed), but other than the extra code, I don't see an issue with them being implemented underneath too.

vm/compiler/src/ledger/store/block/mod.rs
howardwu: Agreed! Would you mind updating each of these cases to this approach? I think it may be easier to checkout an earlier commit and hash and add the wrapping closure.

howardwu: Lets tackle this as a follow up

PR #974, opened 2022-08-17
[Testnet3] Remove redundant rerun condition

raychu86:
## Motivation

This PR removes a rerun condition in `build.rs` which will cause unnecessary recompilations for any change in the `target` directory. The default `Cargo` behavior will already rerun the `build.rs` script if any Rust file in the source tree has been changed. 

This particular condition is more meant for use when there are other language files to track that `Cargo` can not, for example `C++` files used in FFI bindings.

### DIFF
--- a/build.rs
+++ b/build.rs
@@ -52,9 +52,6 @@ fn check_file_licenses<P: AsRef<Path>>(path: P) {
             );
         }
     }
-
-    // Re-run upon any changes to the workspace.
-    println!("cargo:rerun-if-changed=.");
 }
 
 // The build script; it currently only checks the licenses.


howardwu:
@ljedrz Can you review if this is correct with your original implementation?

howardwu:
@ljedrz friendly reminder to check if this is correct (and to reflect this in snarkOS/Leo/Aleo repos)

ljedrz:
Apologies, must have missed the notification the first time around.

I'm aware of the default behavior of the build script, but IIRC this extended check is actually not redundant - during development I did some manual tests with the snarkVM workspace (changing the license text in some Rust source files at random), and without this line, the script failed to detect some of the changes.

ljedrz:
I'm pretty sure the root cause is that the license texts are technically only **code comments**, which means that incremental recompilation will not trigger if the only change is to the license text, causing the build script to not be rerun.

While a change to the license text only is probably a rare occurrence, the script is just being thorough, and it hasn't caused any slowdowns in my tests.

Also, this doesn't have any real security implications one way or the other (and this matter is listed among security updates).

PR #975, opened 2022-08-17
[Testnet3] Remove bidirectional code points support from `StringType`s

raychu86:
## Motivation

This PR removes support for "invisible" bidirectional code points in `StringType`s. This is important because these characters will effectively be hidden when displaying Aleo programs on Github (some IDEs will be able to display these characters). 

Currently the invalidates support for the following bidirectional override characters - https://unicode.org/reports/tr9/: 
```
U+202A, U+202B, U+202C, U+202D, U+202E, U+2066,
U+2067, U+2068, U+2069
```

### Example
Here is an example to explain the significance of this issue:
Program displayed on Github: 
```
function compare:
    hash.psd2 "aleo" into r0;
    hash.psd2 "aleo" into r1; // Looks like the same string
    is.eq r0 r1 into r2;
    output r3 as boolean;
```

Same program on an IDE that shows the code points:
```
function compare:
    hash.psd2 "aleo" into r0;
    hash.psd2 "aleo[U+202E]" into r1; // Is definitely NOT the same string
    is.eq r0 r1 into r2;
    output r3 as boolean;
```

### DIFF
--- a/console/network/environment/src/traits/string.rs
+++ b/console/network/environment/src/traits/string.rs
@@ -32,11 +32,16 @@ pub mod string_parser {
         bytes::streaming::{is_not, take_while_m_n},
         character::streaming::{char, multispace1},
         combinator::{map, map_opt, map_res, value, verify},
-        error::{FromExternalError, ParseError},
+        error::{ErrorKind, FromExternalError, ParseError},
         multi::fold_many0,
         sequence::{delimited, preceded},
+        Err::Error,
         IResult,
     };
+
+    /// The list of unsupported "invisible" code points.
+    const UNSUPPORTED_CODE_POINTS: [&str; 9] =
+        ["\u{202a}", "\u{202b}", "\u{202c}", "\u{202d}", "\u{202e}", "\u{2066}", "\u{2067}", "\u{2068}", "\u{2069}"];
 
     /// Parse a unicode sequence, of the form u{XXXX}, where XXXX is 1 to 6
     /// hexadecimal numerals. We will combine this later with parse_escaped_char
@@ -106,6 +111,13 @@ pub mod string_parser {
 
     /// Parse a non-empty block of text that doesn't include \ or "
     fn parse_literal<'a, E: ParseError<&'a str>>(input: &'a str) -> IResult<&'a str, &'a str, E> {
+        // Return an error if the literal contains an unsupported code point.
+        for code_point in UNSUPPORTED_CODE_POINTS {
+            if input.contains(code_point) {
+                return Err(Error(E::from_error_kind("String literal contains invalid codepoint", ErrorKind::Char)));
+            }
+        }
+
         // `is_not` parses a string of 0 or more characters that aren't one of the
         // given characters.
         let not_quote_slash = is_not("\"\\");
--- a/console/types/string/src/parse.rs
+++ b/console/types/string/src/parse.rs
@@ -92,4 +92,18 @@ mod tests {
         }
         Ok(())
     }
+
+    #[test]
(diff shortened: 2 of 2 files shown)


acoglio:
This is good. We also did it in Leo (also for comments, not only strings):
https://github.com/AleoHQ/leo/blob/1370c4a91c0517ff8dc1a0933045db4fded0eb3c/docs/grammar/abnf-grammar.txt#L24

Related to this, you may consider also restricting allowed ASCII control characters. This shows a similar issue with ASCII backspace: https://github.com/AleoHQ/leo/issues/2003

howardwu:
@acoglio would you happen to know if there are other code points or string cases where this type of ambiguity can exist? We want to make sure we're hitting every case that we know of :)

acoglio:
@howardwu Also tagging @bendyarm, who first brought up the Unicode bidi override issue for Leo a while ago.

This is my assessment for ASCII:
- 32 (space) is fine.
- 33-126 (visible chars) are fine.
- 10 (line feed) and 13 (carriage return) are fine.
- 9 (horizontal tab), 11 (vertical tab), 12 (form feed) should be fine, but I see no good reason to allow them.
- The others should be disallowed, particularly 8 (backspace —  see above) and 127 (delete).

For Unicode, the bidi overrides that this PR excludes are well-known dangers, but I don't know offhand if there may be possible issues with others. I believe we should make an assessment for all non-ASCII Unicode characters (well, by ranges, not one by one 😄) similarly to the above list for ASCII, and only allow what we have good reasons to allow.

This is an instance of the general security principle to prefer whitelists to blacklists, i.e. thinking explicitly of what we want to allow and why, rather than saying "anything but...", since the latter may lead to unexpected vulnerabilities, cf. CWEs such as [this general one](https://cwe.mitre.org/data/definitions/20.html).

While we are on this topic, I'd like to advocate a [LangSec (= language-theoretic security) approach](http://langsec.org) (also mentioned in the above CWE) for our handling of all inputs to our tools, not just for "proper" languages like Leo and Aleo instructions, but also for other input data. We use 3rd-party crates for parsing things like TOML and YAML, which I think is fine for expediency, but I'm not sure how well those crates follow LangSec principles (maybe they do, but it's something we should look into).

bendyarm:
@howardwu  Yes, there are definitely many other code points where this ambiguity can exist.  
First, I think there are two concepts conflated by the Motivation section of this PR.  
One concept is the bidi overrides and the other is characters that cannot be seen.

In general, you cannot look at text rendered from Unicode and know what codepoints there are.  There are numerous invisible codepoints.  There are different codepoints that are usually rendered the same or almost the same.  There are single codepoints that look like multiple joined codepoints.  A popular example is the homoglyph attacks like `ebау.com` which is not the same as `ebay.com` (paste it into a web browser), and client-side Punycode mitigates those attacks.   If we want the Aleo Instructions strings to be unambiguous when read by a human, then we must limit the allowable contents.

Note that if we wish, we can allow general Unicode codepoints in a string at runtime while disallowing them at parse time (and at print time, when displaying in IDEs), by requiring escapes for all except standard printable ASCII.  That would make strings more burdensome for people who want messages in other languages than English.

The bidi override issue is actually much worse than invisible characters, it is about making code appear different from what the code actually is.  There are links to the original paper and a shorter writeup [from this Leo issue. ](https://github.com/AleoHQ/leo/issues/1726).

@acoglio mentioned that the bidi override issue is also relevant to comments.

One other note: it is also important to make sure any user-facing tool understands these concepts so we have a consistent view.

howardwu (approved):

PR #977, opened 2022-08-17
[Testnet3] Downcast truncation protection

raychu86:
## Motivation

This PR prevents 3 cases of possible truncation when downcasting integer types. Two fixes use `try_from` and one uses a manual `if` check. 

There is many more potential downcasting issues in the rest of the repo, and should be addressed with https://github.com/AleoHQ/snarkVM/issues/970.

### DIFF
--- a/console/program/src/data/ciphertext/bytes.rs
+++ b/console/program/src/data/ciphertext/bytes.rs
@@ -38,7 +38,7 @@ impl<N: Network> ToBytes for Ciphertext<N> {
     #[inline]
     fn write_le<W: Write>(&self, mut writer: W) -> IoResult<()> {
         // Ensure the number of field elements does not exceed the maximum allowed size.
-        if self.0.len() as u32 > N::MAX_DATA_SIZE_IN_FIELDS {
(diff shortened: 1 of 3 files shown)


console/program/src/data/ciphertext/bytes.rs
```diff
@@ -38,7 +38,7 @@ impl<N: Network> ToBytes for Ciphertext<N> {
     #[inline]
     fn write_le<W: Write>(&self, mut writer: W) -> IoResult<()> {
         // Ensure the number of field elements does not exceed the maximum allowed size.
-        if self.0.len() as u32 > N::MAX_DATA_SIZE_IN_FIELDS {
+        if self.0.len() as u32 > N::MAX_DATA_SIZE_IN_FIELDS || self.0.len() > u16::MAX as usize {
```
pablodeymo, line 41: I think that instead of `16::MAX as usize` this form could be better to avoid the `as` casting: 
`usize::from(u16::MAX)`

howardwu (approved):

PR #982, opened 2022-08-17
Tweak the atomic write batch setup

### LINKED ISSUE #979: Update atomics to closures
alternatively, these functions could look as follows (example for `BlockStorage::insert`):
```
/// Stores the given `block` into storage.
fn insert(&self, block: &Block<N>) -> Result<()> {
    // Start an atomic batch write operation.
    self.start_atomic();

    let atomic_ops = || -> Result<()> {
        // Store the block hash.
        self.id_map().insert(block.height(), block.hash())?;
        // Store the block height.
        self.reverse_id_map().insert(block.hash(), block.height())?;
        // Store the block header.
        self.header_map().insert(block.hash(), *block.header())?;

        // Store the transaction IDs.
        self.transactions_map().insert(block.hash(), block.transaction_ids().copied().collect())?;

        // Store the block transactions.
        for transaction in block.transactions().values() {
            // Store the reverse transaction ID.
            self.reverse_transactions_map().insert(transaction.id(), block.hash())?;
            // Store the transaction.
            self.transaction_store().insert(transaction)?;
        }

        // Store the block signature.
        self.signature_map().insert(block.hash(), *block.signature())?;

        Ok(())
    };

    if atomic_ops().is_err() {
        self.abort_atomic();
    }

    // Finish the atomic batch write operation.
    self.finish_atomic();

    Ok(())
}
```
This would obviate `OrAbort`.

_Originally posted by @ljedrz in https://github.com/AleoHQ/snarkVM/pull/973#discussion_r947567328_

ljedrz:
A follow-up to https://github.com/AleoHQ/snarkVM/pull/973.

In the previous atomic write batch PR, I considered the utility of being able to run lower-level operations individually; however, this requires a bit of extra logic in order for the "smaller" atomic operations to not cause the ones that run them underneath to perform (partial) writes prematurely. That's why I added `Map::is_atomic_in_progress` that can allow us to determine if there is already a batch write in progress, which is then used in a new `atomic_write_batch` macro that's used in all the atomic operations.

This means that any atomic batch write operation that executes several other ones underneath starts a large common batch that is "reused" by all the operations involved, so e.g. when `BlockStorage::insert` is called (via `Ledger::add_next_block`), it starts a write batch that's later used by the `IDMap`, `TransactionsMap`, `TransactionStore` etc., but if any of those were to have `insert` called on them in isolation (i.e. not as part of `BlockStorage::insert`), those operations would be atomic as well, creating their own batch in the process (with the previous rule applying to any possible `Map` operations underneath).

In addition, `Map::finish_atomic` now returns a `Result<()>`, in order to accommodate potential related persistent storage errors.

Closes https://github.com/AleoHQ/snarkVM/issues/979.

### DIFF
--- a/vm/compiler/src/ledger/map/memory_map.rs
+++ b/vm/compiler/src/ledger/map/memory_map.rs
@@ -118,6 +118,15 @@ impl<
     }
 
     ///
+    /// Checks whether an atomic operation is currently in progress. This can be done to ensure
+    /// that lower-level operations don't start and finish their individual atomic write batch
+    /// if they are already part of a larger one.
+    ///
+    fn is_atomic_in_progress(&self) -> bool {
+        self.batch_in_progress.load(Ordering::SeqCst)
+    }
+
+    ///
     /// Aborts the current atomic operation.
     ///
     fn abort_atomic(&self) {
--- a/vm/compiler/src/ledger/map/memory_map.rs
+++ b/vm/compiler/src/ledger/map/memory_map.rs
@@ -130,7 +130,7 @@ impl<
     ///
     /// Finishes an atomic operation, performing all the queued writes.
     ///
-    fn finish_atomic(&self) {
+    fn finish_atomic(&self) -> Result<()> {
         // Retrieve the atomic batch.
         let operations = core::mem::take(&mut *self.atomic_batch.lock());
 
@@ -145,8 +145,11 @@ impl<
                 };
             }
         }
+
         // Set the atomic batch flag to `false`.
         self.batch_in_progress.store(false, Ordering::SeqCst);
+
+        Ok(())
     }
 }
 
@@ -264,7 +267,7 @@ mod tests {
         assert!(map.iter().next().is_none());
 
         // Finish the current atomic write batch.
-        map.finish_atomic();
+        map.finish_atomic().unwrap();
 
         // Check that the items are present in the map now.
         for i in 0..NUM_ITEMS {
@@ -285,7 +288,7 @@ mod tests {
         assert_eq!(map.iter().count(), NUM_ITEMS);
 
         // Finish the current atomic write batch.
-        map.finish_atomic();
+        map.finish_atomic().unwrap();
 
         // Check that the map is empty now.
         assert!(map.iter().next().is_none());
@@ -328,7 +331,7 @@ mod tests {
         }
 
         // Finish the current atomic write batch.
-        map.finish_atomic();
+        map.finish_atomic().unwrap();
 
         // The map should contain NUM_ITEMS items now.
         assert_eq!(map.iter().count(), NUM_ITEMS);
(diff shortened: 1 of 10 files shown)


vm/compiler/src/ledger/map/memory_map.rs
```diff
@@ -264,7 +276,7 @@ mod tests {
         assert!(map.iter().next().is_none());
 
         // Finish the current atomic write batch.
-        map.finish_atomic();
+        map.finish_atomic().unwrap();
```
niklaslong, line 279: Maybe worth commenting on the safety of the `unwrap`s here?

ljedrz: it's in a `test` :wink:

vm/compiler/src/ledger/map/mod.rs
```diff
@@ -119,3 +110,34 @@ pub trait MapRead<
...
+    ($self:expr, $ops:block) => {
+        // Check if an atomic batch write is already in progress.
+        let is_part_of_atomic_batch = $self.is_atomic_in_progress();
+
+        // Start an atomic batch write operation IFF it's not already part of one.
+        if !is_part_of_atomic_batch {
+            $self.start_atomic();
+        }
+
+        // Wrap the operations in a closure to know if any of them has failed.
```
niklaslong:
```suggestion
        // Wrap the operations in a closure to detect if any of them have failed.
```

ljedrz: technically only one can fail at a time, but I'm ok with the wording adjustment :ok_hand:

ljedrz:
The CI error seems to be unrelated, I'll retry:
```
running 1 test
test program::instruction::operation::is::tests::test_is_eq_fails has been running for over 60 seconds
error: test failed, to rerun pass '--lib'

Caused by:
  process didn't exit successfully: `/home/circleci/project/target/debug/deps/snarkvm_compiler-e1b225389cbff33f 'program::instruction::operation::is::tests::test_is_eq_fails'` (signal: 9, SIGKILL: kill)
```

vm/compiler/src/ledger/map/memory_map.rs
niklaslong: oh, so it is 😄

ljedrz:
A different CI error, also seemingly unrelated:
```
thread 'program::instruction::operation::assert::tests::test_assert_eq_succeeds' panicked at 'assertion failed: `(left == right)`
  left: `0`,
 right: `1`: Constant constraint failed: (0 * 1) =?= 1', /home/circleci/project/circuit/environment/src/circuit.rs:156:29


failures:
    program::instruction::operation::assert::tests::test_assert_eq_succeeds
```

howardwu (approved):

PR #983, opened 2022-08-17
Faster `console::algorithms::Poseidon`

Pratyush:
## Motivation

Improves the speed of `snarkvm_console_algorithms::Poseidon` by anywhere between 10-25% by using `F::sum_of_products`.

## Test Plan

Added tests for `{Fp256, Fp384}::sum_of_products`

### DIFF
--- a/console/algorithms/src/poseidon/helpers/state.rs
+++ b/console/algorithms/src/poseidon/helpers/state.rs
@@ -44,7 +44,7 @@ impl<E: Environment, const RATE: usize, const CAPACITY: usize> State<E, RATE, CA
 
 impl<E: Environment, const RATE: usize, const CAPACITY: usize> State<E, RATE, CAPACITY> {
     /// Returns an immutable iterator over the state.
-    pub fn iter(&self) -> impl Iterator<Item = &Field<E>> {
+    pub fn iter(&self) -> impl Iterator<Item = &Field<E>> + Clone {
         self.capacity_state.iter().chain(self.rate_state.iter())
     }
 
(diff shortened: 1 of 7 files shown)


Pratyush:
I don't think the test failure is related to my PR, but I could be wrong on that.

console/algorithms/src/poseidon/helpers/state.rs
```diff
@@ -44,7 +44,7 @@ impl<E: Environment, const RATE: usize, const CAPACITY: usize> State<E, RATE, CA
 
 impl<E: Environment, const RATE: usize, const CAPACITY: usize> State<E, RATE, CAPACITY> {
     /// Returns an immutable iterator over the state.
-    pub fn iter(&self) -> impl Iterator<Item = &Field<E>> {
+    pub fn iter(&self) -> impl Iterator<Item = &Field<E>> + Clone {
```
howardwu, line 47: It could be possible to implement `Copy` here, instead of `Clone`, and use copy in the sum of products (I could be wrong but worth checking)

howardwu:
> I don't think the test failure is related to my PR, but I could be wrong on that.

Yes, this one is a flaky test it seems

console/algorithms/src/poseidon/helpers/state.rs
Pratyush: Unfortunately the output of `chain` is not `Copy`, so we have to use `Clone` instead.

raychu86 (approved):

PR #989, opened 2022-08-19
A minor logic tweak to Ledger::check_next_block

ljedrz:
There's no need to check the block's height for every transition if it's what determines whether to do anything with them; it makes more sense to only iterate over them if the height is right in the first place.


howardwu:
Closing as the intention is to add logic for when the case is 0.

ljedrz:
As long as it happens before going over transactions :+1:.

PR #990, opened 2022-08-19
Don't verify transactions twice

ljedrz:
When running `Ledger::add_next_block`, the transactions are currently verified twice:
- in `Ledger::check_next_block` (called at `vm/compiler/src/ledger/mod.rs:508`)
- in `VM::finalize` (called at `vm/compiler/src/ledger/mod.rs:556`)

Since this isn't a fast operation, we should only be doing it once, likely just in `Ledger::check_next_block`.

### DIFF
--- a/vm/compiler/src/ledger/vm/finalize.rs
+++ b/vm/compiler/src/ledger/vm/finalize.rs
@@ -21,8 +21,6 @@ impl<N: Network, P: ProgramStorage<N>> VM<N, P> {
     /// This method assumes the given transaction **is valid**.
     #[inline]
     pub fn finalize(&mut self, transaction: &Transaction<N>) -> Result<()> {
-        // Ensure the transaction is valid.
-        ensure!(self.verify(transaction), "Invalid transaction: failed to verify");
         // Finalize the transaction.
         match transaction {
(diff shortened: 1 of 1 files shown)


ljedrz:
If extra safety is desired, we should instead use types, e.g. introduce a `VerifiedBlock` and/or a `VerifiedTransaction` that have the same contents as their unverified counterparts and are returned by e.g. `Ledger::check_next_block` (or, alternatively introduce some `Verified<T>` wrapper struct). They would later used by methods that require verified input, and thus don't need to verify it on their own anymore.

ljedrz:
Cc https://github.com/AleoHQ/snarkVM/issues/986

vm/compiler/src/ledger/vm/finalize.rs
```diff
@@ -21,8 +21,6 @@ impl<N: Network, P: ProgramStorage<N>> VM<N, P> {
     /// This method assumes the given transaction **is valid**.
```
ljedrz, line 21: BTW, I just noticed this comment; this change adheres to the doc :smile:

howardwu:
This has been fixed in `testnet3.2`.

PR #991, opened 2022-08-19
Simplify Ledger::add_next_block

ljedrz:
Don't know how I missed it earlier, but here it is - a pretty straightforward simplification to the `Ledger` replacement operation.


howardwu:
Closing as explicit convention ensures we update every struct member.

PR #996, opened 2022-08-24
Extend ledger API

jrchatruc:
## Motivation

We need a few more functions from the ledger so nodes can implement their corresponding http endpoints

### DIFF
--- a/vm/compiler/src/ledger/get.rs
+++ b/vm/compiler/src/ledger/get.rs
@@ -76,16 +76,16 @@ impl<N: Network, B: BlockStorage<N>, P: ProgramStorage<N>> Ledger<N, B, P> {
     }
 
     /// Returns the transaction for the given transaction id.
-    pub fn get_transaction_by_hash(&self, hash: &str) -> Result<Transaction<N>> {
-        let transaction_id = match N::TransactionID::from_str(hash) {
+    pub fn get_transaction(&self, transaction_id_string: &str) -> Result<Transaction<N>> {
+        let transaction_id = match N::TransactionID::from_str(transaction_id_string) {
             Ok(transaction_id) => transaction_id,
-            Err(_error) => bail!("Invalid transaction id {hash}"),
+            Err(_error) => bail!("Invalid transaction id {transaction_id_string}"),
         };
 
         // Retrieve the transaction.
         match self.transactions.get_transaction(&transaction_id)? {
             Some(transaction) => Ok(transaction),
-            None => bail!("Missing transaction for id {hash}"),
+            None => bail!("Missing transaction for id {transaction_id_string}"),
         }
     }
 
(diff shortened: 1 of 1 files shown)


facundoolano (approved):

vm/compiler/src/ledger/get.rs
```diff
@@ -75,6 +75,20 @@ impl<N: Network, B: BlockStorage<N>, P: ProgramStorage<N>> Ledger<N, B, P> {
         }
     }
 
+    /// Returns the transaction for the given transaction id.
+    pub fn get_transaction_by_hash(&self, hash: &str) -> Result<Transaction<N>> {
```
raychu86: We should have the input as a transaction id. 

For example, something like this:
`pub fn get_transaction(&self, transaction_id: N::TransactionID)`

vm/compiler/src/ledger/get.rs
```diff
@@ -75,6 +75,20 @@ impl<N: Network, B: BlockStorage<N>, P: ProgramStorage<N>> Ledger<N, B, P> {
         }
     }
 
+    /// Returns the transaction for the given transaction id.
+    pub fn get_transaction(&self, transaction_id_string: &str) -> Result<Transaction<N>> {
```
raychu86:
```suggestion
    pub fn get_transaction(&self, transaction_id: N::TransactionID) -> Result<Transaction<N>> {
```

This should be ` transaction_id: N::TransactionID` instead of a `&str`. The subsequent functions that call this can parse a string beforehand.

vm/compiler/src/ledger/get.rs
jrchatruc: Renamed the function here [68c8f71](https://github.com/AleoHQ/snarkVM/pull/996/commits/68c8f71f29765f5952535ad4a0fe9342cfb923eb)

vm/compiler/src/ledger/get.rs
jrchatruc: Changed here [c6e67b7](https://github.com/AleoHQ/snarkVM/pull/996/commits/c6e67b7b363c5517adabff993563d293058ea3c0)

vm/compiler/src/ledger/get.rs
```diff
@@ -75,6 +75,15 @@ impl<N: Network, B: BlockStorage<N>, P: ProgramStorage<N>> Ledger<N, B, P> {
         }
     }
 
+    /// Returns the transaction for the given transaction id.
+    pub fn get_transaction(&self, transaction_id: &N::TransactionID) -> Result<Transaction<N>> {
```
raychu86:
```suggestion
    pub fn get_transaction(&self, transaction_id: N::TransactionID) -> Result<Transaction<N>> {
```

Do not pass by reference. Since N::TransactionID implements Copy

jrchatruc: Whoops, you're right; done

raychu86 (approved):

vm/compiler/src/ledger/get.rs
raychu86: Thanks!

PR #999, opened 2022-08-25
Fix the use of test_rng in tests

ljedrz:
I've noticed this while working on https://github.com/AleoHQ/snarkVM/issues/986; with the way `test_rng` is currently used, we're neither benefiting from its speed nor the determinism it provides in case a test fails.

While I continue to work on the aforementioned issue, this adjustment is useful on its own.

### DIFF
--- a/utilities/src/rand.rs
+++ b/utilities/src/rand.rs
@@ -42,6 +42,16 @@ where
 pub fn test_rng() -> XorShiftRng {
     // Obtain the initial seed using entropy provided by the OS.
     let seed = StdRng::from_entropy().gen();
+
+    // Print the seed, so it's displayed if any of the tests using `test_rng` fails.
+    println!("rng seed: {}", seed);
+
+    // Use the seed to initialize a fast, non-cryptographic Rng.
+    XorShiftRng::seed_from_u64(seed)
+}
+
+/// A function that can be used in order to reproduce any failing test cases that use `test_rng`.
+pub fn test_rng_seeded(seed: u64) -> XorShiftRng {
     // Use the seed to initialize a fast, non-cryptographic Rng.
     XorShiftRng::seed_from_u64(seed)
 }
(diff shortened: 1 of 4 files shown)


ljedrz:
Cc https://github.com/AleoHQ/snarkVM/issues/986 for the CI failure :slightly_smiling_face: 

rng seed: `2013169090568970952`

utilities/src/rand.rs
```diff
@@ -42,6 +42,16 @@ where
...
+
+    // Print the seed, so it's displayed if any of the tests using `test_rng` fails.
+    println!("rng seed: {}", seed);
+
+    // Use the seed to initialize a fast, non-cryptographic Rng.
+    XorShiftRng::seed_from_u64(seed)
+}
+
+/// A function that can be used in order to reproduce any failing test cases that use `test_rng`.
+pub fn test_rng_seeded(seed: u64) -> XorShiftRng {
```
howardwu: nit: let's call this test_rng_fixed to adhere to existing convention

utilities/src/rand.rs
```diff
@@ -42,6 +42,16 @@ where
 pub fn test_rng() -> XorShiftRng {
     // Obtain the initial seed using entropy provided by the OS.
     let seed = StdRng::from_entropy().gen();
+
+    // Print the seed, so it's displayed if any of the tests using `test_rng` fails.
+    println!("rng seed: {}", seed);
```
howardwu:
```suggestion
    println!("\nInitializing 'test_rng' with seed '{seed}'\n");
```

Pratyush:
This is a fantastic and simple solution to this problem :). Maybe we can add a `test_rng_from_seed` method to make it easy to use the printed seed?

ljedrz:
@Pratyush it's already there: https://github.com/AleoHQ/snarkVM/pull/999/files#diff-4e91c8c3a15ee7a262cfe30f8d4b8e486f1c13d5ea7e08ff92a549bbbd5b46c7R54 :slightly_smiling_face:

Pratyush:
Oops, I should read more carefully before commenting. This looks fantastic then :)

howardwu (changes requested):

ljedrz:
@howardwu applied the requested changes :+1:

raychu86 (approved):

howardwu (approved):

PR #1004, opened 2022-08-26
Fix imposibility to deploy Aleo programs that have imports

ilitteri:
## Motivation

There was not possible to deploy Aleo programs with imports from the Aleo CLI.

## The problem

Only the main program was in the process when deploying.

## Solution

Added all the main program imports to the process before deploying.

## Related PRs

This issue was mentioned in [this comment](https://github.com/AleoHQ/aleo/pull/326#issuecomment-1227898092) in the Aleo repo.

### DIFF
--- a/vm/package/deploy.rs
+++ b/vm/package/deploy.rs
@@ -14,7 +14,8 @@
 // You should have received a copy of the GNU General Public License
 // along with the snarkVM library. If not, see <https://www.gnu.org/licenses/>.
 
-use snarkvm_compiler::Deployment;
+use indexmap::map::Keys;
+use snarkvm_compiler::{Deployment, Import};
 use snarkvm_console::types::Address;
 use snarkvm_utilities::test_crypto_rng;
 
@@ -138,24 +139,12 @@ impl<N: Network> Package<N> {
         &self,
         endpoint: Option<String>,
     ) -> Result<()> {
-        // Retrieve the main program.
         let program = self.program();
-
-        // Construct the process.
         let mut process = Process::<N>::load()?;
-
-        let imports_directory = self.imports_directory();
-
-        program.imports().keys().try_for_each(|program_id| {
-            let import_program_file = AleoFile::open(&imports_directory, program_id, false)?;
-            process.add_program(import_program_file.program())?;
-            Ok::<_, Error>(())
-        })?;
-
+        self.add_imported_programs_to_process(&mut process, program.imports().keys())?;
         if let Some(ref endpoint) = endpoint {
             Self::deploy_program::<A>(&process, program, endpoint)?;
         }
-
         Ok(())
     }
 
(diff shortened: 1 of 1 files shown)


vm/package/deploy.rs
```diff
@@ -137,35 +138,46 @@ impl<N: Network> Package<N> {
     pub fn deploy<A: crate::circuit::Aleo<Network = N, BaseField = N::Field>>(
```
raychu86, line 137: The logic for the subsequent functions are quite straightforward, try to keep them all contained the the `deploy` function.

vm/package/deploy.rs
```diff
@@ -137,35 +138,46 @@ impl<N: Network> Package<N> {
...
         let program = self.program();
+        let mut process = Process::<N>::load()?;
+        self.add_imported_programs_to_process(&mut process, program.imports().keys())?;
+        if let Some(ref endpoint) = endpoint {
+            Self::deploy_program::<A>(&process, program, endpoint)?;
+        }
+        Ok(())
+    }
 
-        // Retrieve the program ID.
```
raychu86, line 144: Try to document each step like it's done previously. For example:

```
pub fn deploy() {
// Retrieve the main program.
...

// Retrieve the program ID.
...

// Add the imported programs.

...

```

facundoolano: I actually suggested him not to do that 😅 I think it makes sense when the comment helps clarifying the code below, but they are distracting when they just repeat what the code already states, e.g.

``` 
// Retrieve the program ID.
let program_id = program.id();
```

raychu86: In general, we like to err on the side of over documenting, so when api's change, we still know the exact intentions of the logic.

vm/package/deploy.rs
```diff
@@ -137,35 +137,40 @@ impl<N: Network> Package<N> {
     pub fn deploy<A: crate::circuit::Aleo<Network = N, BaseField = N::Field>>(
         &self,
         endpoint: Option<String>,
-    ) -> Result<Deployment<N>> {
```
raychu86, line 140: If we remove this `Deployment` output, what is the purpose of the case where `endpoint` is None.

When would the user tries to call `deploy` without an endpoint? The deployment is never generated, but the function runs to completion without issue.

vm/package/deploy.rs
facundoolano: Yeah I prefer to err on the side of over-documenting as well, but comments should complement or augment the information provided by the code, not reiterate it. E.g. establish intent, context and assumptions, especially at the interfaces (module, struct and function-level), and explaining/clarifying the complex parts of the implementation. From that perspective I'd say this module is under-documented even though it has many comments. Comments that just repeat the code are distracting to the reader and add maintenance burden without providing value.

PR #1005, opened 2022-08-26
Fix address for deploy request

IAvecilla:
## Motivation

The address inside the body of the request sent to the explorer to deploy a program, wasn't the expected one. 
The used address was the one associated with the program and not the one of the developer/owner of it.
Now the deploy command get the address from the manifest file and use that to send it on the request.

### DIFF
--- a/vm/package/deploy.rs
+++ b/vm/package/deploy.rs
@@ -26,17 +26,6 @@ pub struct DeployRequest<N: Network> {
     program_id: ProgramID<N>,
 }
 
-impl<N: Network> TryFrom<Deployment<N>> for DeployRequest<N> {
-    type Error = anyhow::Error;
-
-    fn try_from(deployment: Deployment<N>) -> Result<Self, Self::Error> {
-        let program_id = *deployment.program_id();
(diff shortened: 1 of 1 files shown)


ilitteri (approved):
LGTM

vm/package/deploy.rs
```diff
@@ -67,7 +56,7 @@ impl<N: Network> DeployRequest<N> {
 impl<N: Network> Serialize for DeployRequest<N> {
     /// Serializes the deploy request into string or bytes.
     fn serialize<S: Serializer>(&self, serializer: S) -> Result<S::Ok, S::Error> {
-        let mut request = serializer.serialize_struct("DeployRequest", 1)?;
```
facundoolano, line 70: this looks suspicious why do we need to manually keep the amount of fields in sync?

howardwu: https://serde.rs/impl-serialize.html#serializing-a-struct

It's a size hint that some libraries implementing `Serialize` will use to determine the map allocation

howardwu (approved):

PR #1011, opened 2022-08-29
Remove redundant parser call.

acoglio:
Since each of `Mapping::parse`, `Interface::parse`, etc. call `Sanitizer::parse`
as their first action, there is no need to call `Sanitizer::parse` before
calling the `alt` of those parsers.


acoglio:
Not sure why the `parameters` CI failed. Previously a different one failed. But `cargo run` works on my machine. Perhaps they are false alarms?

acoglio:
The only reason to keep the redundant call could be a very slight performance advantage, but it seems better to me to keep the code simpler; furthermore, it doesn't look like the parser is currently a bottleneck.

howardwu (approved):

PR #1015, opened 2022-08-30
[WIP] Add execute feature

jrchatruc:
## Motivation

This PR contains the functionality to execute program functions, needed by the CLI for the `aleo execute` feature.

Still WIP, for now it's mostly a port of the `run` code; a lot of the code can probably be abstracted as it will be shared between `run` and `execute`.


vm/package/execute.rs
```diff
@@ -0,0 +1,219 @@
...
+        Self { execution }
+    }
+
+    /// Returns the execution.
+    pub const fn execution(&self) -> &Execution<N> {
+        &self.execution
+    }
+}
+
+impl<N: Network> Serialize for ExecutionResponse<N> {
```
facundoolano, line 102: not necessary to solve this in this PR, but I check if we can make this work with a derive or something like that, I'd expect not to have to write custom serializers for simple structs like this one

PR #1016, opened 2022-08-30
Support for deploying programs with imports

IAvecilla:
## Warning!

This PR doesn't work in its actual state. It prints the following error:

```
ERROR unhandled custom rejection, returning 500 response: Request("channel closed")
Error: Invalid transaction found in the transactions list
```

## Motivation

The imported programs are not being deployed before the main program and that causes an error when deploying. 
The error is the following one:

```
WARN Deployment verification failed: Cannot add program 'foo.aleo' because its import 'bar.aleo' must be added first
Error: Invalid transaction found in the transactions list
```

This PR includes the logic to add all the imported programs to the process and make the deployment of them before the main program.

### DIFF
--- a/vm/compiler/src/process/deploy.rs
+++ b/vm/compiler/src/process/deploy.rs
@@ -30,6 +30,21 @@ impl<N: Network> Process<N> {
         stack.deploy::<A, R>(rng)
     }
 
+    /// Deploys all the programs that are in the process's stacks.
+    #[inline]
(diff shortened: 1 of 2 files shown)


IAvecilla:
This PR contains logic that is not accurate to achieve the correct deployment of imported programs. The idea is to not deploy every imported program on a `deploy` run instead deploy a program and then import them in another one checking if the imported program was previously uploaded.

PR #1019, opened 2022-08-31
Pad the underlying fields when comparing StringTypes

### LINKED ISSUE #986: [Bug] CI job failure (test_assert_eq_succeeds)
Detected in https://github.com/AleoHQ/snarkVM/pull/982:
```
thread 'program::instruction::operation::assert::tests::test_assert_eq_succeeds' panicked at 'assertion failed: `(left == right)`
  left: `0`,
 right: `1`: Constant constraint failed: (0 * 1) =?= 1', /home/circleci/project/circuit/environment/src/circuit.rs:156:29


failures:
    program::instruction::operation::assert::tests::test_assert_eq_succeeds
```

ljedrz:
While I'm not super savvy with the related crypto, I tracked the issue behind #986 down to the difference in the number of underlying fields when comparing strings of different sizes; I tried right-padding the shorter list of fields with the zero field for comparison purposes, and it seems to fix the issue.

Fixes #986.

### DIFF
--- a/circuit/types/string/src/equal.rs
+++ b/circuit/types/string/src/equal.rs
@@ -15,6 +15,9 @@
 // along with the snarkVM library. If not, see <https://www.gnu.org/licenses/>.
 
 use super::*;
+use snarkvm_circuit_types_field::Field;
+
+use core::cmp::Ordering;
 
 impl<E: Environment> Equal<Self> for StringType<E> {
     type Output = Boolean<E>;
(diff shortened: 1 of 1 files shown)


ljedrz:
As for picking `Field::zero` as the padding value (i.e. if it could inadvertently cause the comparison of strings of different lengths to result in `true`), I haven't been able to hit that value when testing different string values, and the `Field::from_bits_le` impl (specifically the mixing with the coefficient) suggests that it's impossible that it would ever be one of the underlying fields.

howardwu:
@Pratyush we should discuss the approach to fix the `StringType`, so we can resolve the ongoing CI failure.

The issue with this PR's approach is that padding leads to ambiguity in the length, causing false-positives (string match collisions). I believe we should employ the `terminus bit` approach by appending a `1` at the end of the bitstring.

Wdyt?

ljedrz:
Can this PR now be closed since https://github.com/AleoHQ/snarkVM/pull/1136 (which seems like an alternative) was merged?

howardwu:
Yes, thanks for flagging. Closing PR now.

PR #1025, opened 2022-08-31
Minor updates to validator selection

raychu86:
## Motivation

This PR adds basic functionality for getting, adding, and removing validators. Additionally, initializing a fresh ledger or loading a ledger from storage will now select the address from the genesis block signature as the initial validator, instead of the hard-coded address.

NOTE: The current validator model is just a placeholder and will be updated for Phase 3 of Testnet3.

### DIFF
--- a/vm/compiler/src/ledger/mod.rs
+++ b/vm/compiler/src/ledger/mod.rs
@@ -594,9 +594,36 @@ impl<N: Network, B: BlockStorage<N>, P: ProgramStorage<N>> Ledger<N, B, P> {
         Ok(())
     }
 
+    /// Adds a given address to the validator set.
+    pub fn add_validator(&mut self, address: Address<N>) -> Result<()> {
+        if self.validators().contains_key(&address) {
+            bail!("'{address}' is already in the validator set.")
+        }
+
+        // Insert the address into the validator set.
+        self.validators.insert(address, ());
+        Ok(())
+    }
+
+    /// Removes a given address from the validator set.
+    pub fn remove_validator(&mut self, address: Address<N>) -> Result<()> {
+        if !self.validators().contains_key(&address) {
+            bail!("'{address}' is not in the validator set.")
+        }
+
+        // Remove the address from the validator set.
+        self.validators.remove(&address);
+        Ok(())
+    }
+
     /// Returns the block tree.
     pub const fn block_tree(&self) -> &BlockTree<N> {
         &self.block_tree
+    }
+
+    /// Returns the validator set.
+    pub const fn validators(&self) -> &IndexMap<Address<N>, ()> {
+        &self.validators
     }
 
     /// Returns the memory pool.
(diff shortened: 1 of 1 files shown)


facundoolano (approved):

vm/compiler/src/ledger/mod.rs
```diff
@@ -594,11 +591,38 @@ impl<N: Network, B: BlockStorage<N>, P: ProgramStorage<N>> Ledger<N, B, P> {
         Ok(())
     }
 
+    /// Adds a given address to the validator set.
+    pub fn add_validator(&mut self, address: Address<N>) -> Result<()> {
+        if self.validators().contains_key(&address) {
```
ljedrz: since `insert` indicates whether the key had already existed and the value doesn't change, there's no need to use `contains_key`

vm/compiler/src/ledger/mod.rs
```diff
@@ -594,11 +591,38 @@ impl<N: Network, B: BlockStorage<N>, P: ProgramStorage<N>> Ledger<N, B, P> {
...
+        }
+
+        // Insert the address into the validator set.
+        self.validators.insert(address, ());
+        Ok(())
+    }
+
+    /// Removes a given address from the validator set.
+    pub fn remove_validator(&mut self, address: Address<N>) -> Result<()> {
+        if !self.validators().contains_key(&address) {
```
ljedrz: ditto: `remove` indicates whether the key had existed, so a double lookup is not necessary

vm/compiler/src/ledger/mod.rs
raychu86: Logically, we wouldn't even need the `Result`s  here, but since the validator logic will be quite sensitive in subsequent testnet3 phases, I wanted to have dedicated error handling here.

vm/compiler/src/ledger/mod.rs
raychu86: Same comment as above, however I guess using the `Option<V>` output of the `remove` call could be better than the double lookup.

vm/compiler/src/ledger/mod.rs
ljedrz: the `Result` is fine; what I'm suggesting is:
```
pub fn add_validator(&mut self, address: Address<N>) -> Result<()> {
    if self.validators.insert(address, ()).is_some() {
        bail!("'{address}' is already in the validator set.")
    } else {
        Ok(())
    }
}
```

vm/compiler/src/ledger/mod.rs
ljedrz: it would look like this:
```
pub fn remove_validator(&mut self, address: Address<N>) -> Result<()> {
    if self.validators.remove(&address).is_none() {
        bail!("'{address}' is not in the validator set.")
    } else {
        Ok(())
    }
}
```

vm/compiler/src/ledger/mod.rs
```diff
@@ -146,12 +144,7 @@ impl<N: Network, B: BlockStorage<N>, P: ProgramStorage<N>> Ledger<N, B, P> {
             transitions: blocks.transition_store().clone(),
             blocks,
             // TODO (howardwu): Update this to retrieve from a validators store.
-            validators: [(
-                Address::<N>::from_str("aleo1q6qstg8q8shwqf5m6q5fcenuwsdqsvp4hhsgfnx5chzjm3secyzqt9mxm8")?,
-                (),
-            )]
-            .into_iter()
-            .collect(),
+            validators: [].into_iter().collect(),
```
ljedrz:
```suggestion
            validators: Default::default(),
```

PR #1029, opened 2022-09-02
Add transaction ID for the deployment process

IAvecilla:
## Motivation

The user of the Aleo CLI tool needs to know the transaction ID associated with the deployment of their programs. This PR includes the `transaction_id` field on the deployment response to send it to the Aleo CLI tool and show it to the user.

### DIFF
--- a/vm/package/deploy.rs
+++ b/vm/package/deploy.rs
@@ -85,25 +85,32 @@ impl<'de, N: Network> Deserialize<'de> for DeployRequest<N> {
 
 pub struct DeployResponse<N: Network> {
     deployment: Deployment<N>,
+    transaction_id: String,
 }
 
 impl<N: Network> DeployResponse<N> {
     /// Initializes a new deploy response.
-    pub const fn new(deployment: Deployment<N>) -> Self {
-        Self { deployment }
+    pub const fn new(deployment: Deployment<N>, transaction_id: String) -> Self {
+        Self { deployment, transaction_id }
     }
 
     /// Returns the program ID.
     pub const fn deployment(&self) -> &Deployment<N> {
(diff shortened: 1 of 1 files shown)


vm/package/deploy.rs
```diff
@@ -192,14 +200,14 @@ mod tests {
         let (directory, package) = crate::package::test_helpers::sample_package();
 
         // Deploy the package.
-        let deployment = package.deploy::<CurrentAleo>(None).unwrap();
```
facundoolano, line 195: shouldn't these tests be updated to include the transaction id field you added?

IAvecilla: Since the test cases use `None` for the endpoint (because there isn't an actual deployment) the `transaction_id` will always be an empty string because there is not an actual transaction. Do you think is worth adding it to the test anyway?

facundoolano: Can the test be rewritten to send a Some endpoint and yield a non empty transaction? Otherwise it seems we're only testing the getters and setters

IAvecilla: You're right that is not that useful to have those tests here. Maybe we have to consider doing them in another repo where we can test a real transaction with a real endpoint.

howardwu:
Instead of returning `{ deployment, transaction_id }`, can we return `{ transaction }` since it will already contain the `deployment` and the `ID` (in addition to all other details)?

(Related comment here: https://github.com/AleoHQ/snarkOS/pull/1910)

IAvecilla:
These changes are deprecated with the new versions of the Aleo CLI, snarkOS, and snarkVM. If necessary, these changes will be introduced again in the future.

PR #1030, opened 2022-09-02
Speed up add_mixed_assign in TE Projective

jules:
This PR adds an optimization which should speed up the calls to `add_assign_mixed` - it adds a `t` coordinate on `Affine` points, which saves a multiplication.


howardwu:
@julesdesmit do we have any benchmarks to confirm the speedup and %?

curves/src/templates/twisted_edwards_extended/affine.rs
```diff
@@ -250,7 +260,7 @@ impl<P: Parameters> Neg for Affine<P> {
     type Output = Self;
 
     fn neg(self) -> Self {
-        Self::new(-self.x, self.y)
+        Self::new(-self.x, self.y, -self.x * self.y)
```
Pratyush:
```suggestion
        Self::new(-self.x, self.y, -self.t)
```

curves/src/traits/group.rs
```diff
@@ -224,6 +224,9 @@ pub trait AffineCurve:
 
     /// Performs the second half of batch addition in-place.
     fn batch_add_loop_2(a: &mut Self, b: Self, inversion_tmp: &mut Self::BaseField);
+
+    /// Sets the `t` coordinate to the correct value.
+    fn correct_t_coordinate(&mut self);
```
Pratyush: why can't we do the correction directly in `batch_add_loop_2`? That way we don't have to expose this method for short weierstrass curves.

jules: I assumed we should do it as little as possible since it would incur a multiplication, so I exposed it to be able to call it after running the add loops

Pratyush: Hm for TE curves I believe we use the standard MSM, not the one with the batch-affine, so we don't call these methods anyway

jules: ah sure, in that case I will just do the correction in `batch_add_loop_2` then :)

Pratyush (approved):

howardwu (approved):
LGTM

Issue #1039, opened 2022-09-03
LangSec compliance


howardwu:
@howardwu Also tagging @bendyarm, who first brought up the Unicode bidi override issue for Leo a while ago.

This is my assessment for ASCII:
- 32 (space) is fine.
- 33-126 (visible chars) are fine.
- 10 (line feed) and 13 (carriage return) are fine.
- 9 (horizontal tab), 11 (vertical tab), 12 (form feed) should be fine, but I see no good reason to allow them.
- The others should be disallowed, particularly 8 (backspace —  see above) and 127 (delete).

For Unicode, the bidi overrides that this PR excludes are well-known dangers, but I don't know offhand if there may be possible issues with others. I believe we should make an assessment for all non-ASCII Unicode characters (well, by ranges, not one by one 😄) similarly to the above list for ASCII, and only allow what we have good reasons to allow.

This is an instance of the general security principle to prefer whitelists to blacklists, i.e. thinking explicitly of what we want to allow and why, rather than saying "anything but...", since the latter may lead to unexpected vulnerabilities, cf. CWEs such as [this general one](https://cwe.mitre.org/data/definitions/20.html).

While we are on this topic, I'd like to advocate a [LangSec (= language-theoretic security) approach](http://langsec.org) (also mentioned in the above CWE) for our handling of all inputs to our tools, not just for "proper" languages like Leo and Aleo instructions, but also for other input data. We use 3rd-party crates for parsing things like TOML and YAML, which I think is fine for expediency, but I'm not sure how well those crates follow LangSec principles (maybe they do, but it's something we should look into).

_Originally posted by @acoglio in https://github.com/AleoHQ/snarkVM/issues/975#issuecomment-1218260751_

acoglio:
As a follow-up, the above should apply to all our parsers of text files: code files, input files, configuration files, etc. Every input to the system should be thoroughly checked.

PR #1044, opened 2022-09-04
Reject unsafe characters in `.aleo` file parser.

acoglio:
Rejects (1) all control ASCII characters except HT and LF and CR, and (2)
Unicode bidi characters, in both comments and string literals. These syntactic
entities (comments and string literals) should be the only part of Aleo
instructions that could otherwise allow these characters.

Prior to this commit, bidi characters were rejected from string literals, but
not from comments. And all ASCII characters were accepted.

Added tests to demonstrate rejection.

Also added some positive tests for string literals, which were not present in
the file for parsing string literals.

This addresses #1039, for `.aleo` files.

### DIFF
--- a/console/network/environment/src/helpers/sanitizer.rs
+++ b/console/network/environment/src/helpers/sanitizer.rs
@@ -18,9 +18,9 @@ use crate::ParserResult;
 
 use nom::{
     branch::alt,
-    bytes::complete::{tag, take_until},
+    bytes::complete::tag,
     character::complete::{anychar, char, line_ending, multispace1},
-    combinator::{cut, map, recognize, value},
+    combinator::{cut, map, recognize, value, verify},
     error::{ErrorKind, VerboseError, VerboseErrorKind},
     multi::fold_many0,
     sequence::{preceded, terminated},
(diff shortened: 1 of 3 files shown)


howardwu:
This looks great @acoglio. Where are you planning to plug this sanitize checker in?

My guess is we could benefit from it in the parser of `Identifier` and `String`, which will then propagate to everything upstream (such as `ProgramID`, `Locator`, etc.)

acoglio:
@howardwu I think that it should be already integrated into the parsing of `.aleo` files, but I may be wrong because I'm not familiar with the rest of the snarkVM code yet. The new `Sanitize::parse_safe_char` function is called (indirectly) by `Sanitizer::parse`, i.e. any time we parse comments and whitespace, so it should apply to all comments. For string literals, given that there was already an `UNSUPPORTED_CODE_POINTS` list, I added the forbidden ASCII codes to it as an initial step, but I think we should rework that code to call `Sanitizer::parse_safe_char` too (so it's a single point of validation), as noted in the comment there. identifiers should not need this because we are already using `nom`'s parsers for ASCII letters and digits and underscore, so anything outside that narrow range would just not parse as part of an identifier.

howardwu:
Rebased here

howardwu:
Rebased again for clippy.

howardwu (approved):
LGTM

PR #1049, opened 2022-09-05
[Feature] get_record method for the Ledger

ilitteri:
## Motivation

We need to be able to retrieve a record ciphertext from the ledger for a given commitment.

## Related PRs

This feature will be used in https://github.com/AleoHQ/snarkOS/pull/1907

### DIFF
--- a/vm/compiler/src/ledger/get.rs
+++ b/vm/compiler/src/ledger/get.rs
@@ -105,6 +105,15 @@ impl<N: Network, B: BlockStorage<N>, P: ProgramStorage<N>> Ledger<N, B, P> {
             None => bail!("Missing signature for block {height}"),
         }
     }
+
+    /// Returns the record for the given `commitment`.
+    ///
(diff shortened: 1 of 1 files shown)


vm/compiler/src/ledger/get.rs
```diff
@@ -105,6 +105,11 @@ impl<N: Network, B: BlockStorage<N>, P: ProgramStorage<N>> Ledger<N, B, P> {
             None => bail!("Missing signature for block {height}"),
         }
     }
+
+    /// Returns the record for the given `commitment`.
```
facundoolano, line 109: I will just comment for future consideration that these pass-through methods are generally a bad sign, suggesting that there may be room for refactoring how the structs are nested and accessed by the client code (not part of this PR or functionality, but something that we need to consider for improvement in the future).

facundoolano (approved):

jrchatruc (approved):

PR #1084, opened 2022-09-19
Implement GLV scalar multiplication algorithm

jules:


curves/src/templates/short_weierstrass_jacobian/projective.rs
```diff
@@ -506,15 +506,116 @@ impl<P: Parameters> Mul<P::ScalarField> for Projective<P> {
...
+
+        let glv_endomorphism = |t: Affine<P>| -> Affine<P> {
+            if t.is_zero() {
+                return t;
+            }
+
+            let mut r = Affine::zero();
+            r.y = t.y;
+            r.x = phi(t.x);
+            r
```
Pratyush:
```suggestion
        let glv_endomorphism = |mut t: Affine<P>| -> Affine<P> {
            t.x = phi(t.x);
            t
```

jules:
Some preliminary bench results:

```
bls12_377: g1_mul_assign
                        time:   [87.016 µs 87.050 µs 87.091 µs]
                        change: [-47.347% -47.283% -47.229%] (p = 0.00 < 0.05)
                        Performance has improved.
bls12_377: g1_is_in_correct_subgroup
                        time:   [53.454 µs 53.485 µs 53.516 µs]
                        change: [-11.952% -11.651% -11.381%] (p = 0.00 < 0.05)
                        Performance has improved.
bls12_377: g2_mul_assign
                        time:   [268.61 µs 268.80 µs 269.00 µs]
                        change: [-56.031% -55.960% -55.888%] (p = 0.00 < 0.05)
                        Performance has improved.
```

curves/src/templates/short_weierstrass_jacobian/projective.rs
```diff
@@ -499,22 +500,107 @@ impl<'a, P: Parameters> SubAssign<&'a Self> for Projective<P> {
...
+        const L: usize = 1 << (GLV_WINDOW_SIZE - 1);
+
+        let decomposition = other.decompose();
+
+        // Prepare tables.
+        let mut t_1 = Vec::with_capacity(L);
+        let double = Affine::from(self.double());
+        t_1.push(self);
+        for i in 1..L {
+            t_1.push(t_1[i - 1].add_mixed(&double));
```
ljedrz: perhaps [iter::scan](https://doc.rust-lang.org/std/iter/trait.Iterator.html#method.scan) could be used here?

curves/src/templates/short_weierstrass_jacobian/projective.rs
```diff
@@ -499,22 +500,107 @@ impl<'a, P: Parameters> SubAssign<&'a Self> for Projective<P> {
...
+                wnaf_1.iter_mut().for_each(|e| *e = -*e);
+            }
+            if !s2 {
+                wnaf_2.iter_mut().for_each(|e| *e = -*e);
+            }
+
+            (wnaf_1, wnaf_2)
+        };
+
+        let naf_add = |table: &Vec<Affine<P>>, naf: i32, acc: &mut Self| {
```
ljedrz:
```suggestion
        let naf_add = |table: &[Affine<P>], naf: i32, acc: &mut Self| {
```

curves/src/templates/short_weierstrass_jacobian/projective.rs
jules: Possibly, though it might be a bit annoying to wrap all of the elements in `Option`s

ljedrz: it's possible that the removal of bounds checks would make it faster, but only if `L` can be large

jules: I see, I think L is fixed at 1 << 3 = 8. Is that large enough for it to have an impact?

ljedrz: nope, most likely not :+1:

fields/src/fp_256.rs
```diff
@@ -376,6 +376,86 @@ impl<P: Fp256Parameters> PrimeField for Fp256<P> {
...
     type Parameters = P;
 
+    #[inline]
+    fn decompose(&self) -> (Self, Self, bool, bool) {
+        const Q1: [u64; 4] = [9183663392111466540, 12968021215939883360, 3, 0];
+        const Q2: [u64; 4] = [13, 0, 0, 0];
+        let b1 = Self::from_repr(BigInteger([725501752471715840, 4981570305181876225, 0, 0])).unwrap();
+        let b2 = Self::from_repr(BigInteger([725501752471715841, 4981570305181876225, 0, 0])).unwrap();
+        let r128 = Self::from_repr(BigInteger([0xffffffffffffffff, 0xffffffffffffffff, 0, 0])).unwrap();
+        const HALF_R: [u64; 8] = [0, 0, 0, 0x8000000000000000, 0, 0, 0, 0];
```
Pratyush: These constants are specific to BLS12-377, right? We should move these to `BLS12Parameters`, and pass these in to `decompose`

fields/src/fp_384.rs
```diff
@@ -434,6 +434,11 @@ impl<P: Fp384Parameters> PrimeField for Fp384<P> {
     type BigInteger = BigInteger;
     type Parameters = P;
 
+    #[inline]
+    fn decompose(&self) -> (Self, Self, bool, bool) {
+        (Self::zero(), Self::zero(), false, false)
```
Pratyush:
```suggestion
        unimplemented!()
```

curves/src/bls12_377/tests.rs
```diff
@@ -616,6 +617,31 @@ fn test_fq12_mul_by_034() {
...
+    //     7829691328573998628,
+    //     14277141908032039045,
+    //     12498936762639071483,
+    //     4759044196164073388,
+    //     3898951160843530891,
+    //     117220051666006481,
+    // ]))
+    // .unwrap();
+    // println!("{:?}", acc.x);
+    // println!("{:?}", x);
```
Pratyush: Is the commented out part testing on a special edge-case?

curves/src/bls12_377/g2.rs
```diff
@@ -89,6 +104,13 @@ impl ShortWeierstrassParameters for Bls12_377G2Parameters {
     ) -> bool {
         p.mul_bits(BitIteratorBE::new(Self::ScalarField::characteristic())).is_zero()
     }
+
+    fn glv_endomorphism(
```
Pratyush, line 127: Something that we might want to follow up on is that for G2 there exists an even more efficient decomposition relying on the GLS endomorphism (https://eprint.iacr.org/2008/194)

curves/src/bls12_377/tests.rs
jules: Oh yeah, this was part of some debugging efforts. I'll take it out :)

fields/src/fp_256.rs
jules: I'll put them on the `ShortWeierstrassParameters` as it's not possible to access `BLS12Parameters` from `Fp256` or `Projective` as far as I can tell.

curves/src/bls12_377/g2.rs
jules: Would you know the speedup this would give? If it's big I would consider doing it for this PR, otherwise might just want to create an issue.

Pratyush: Let's create an issue, since it's not really on the critical path for anything (there's only one G2 mul in Marlin verification)

jules: Done!

fields/src/fp_256.rs
Pratyush: Hm I would rather put a `mul` method in `ShortWeierstrassParameters`, so that we can specify it when instantiating `Bls12377G1Parameters`. See for example https://github.com/arkworks-rs/algebra/blob/402e7f9603fca7a68b86baf296b6feaf904939f5/ec/src/models/short_weierstrass/mod.rs#L82

jules: Hmm, so as in a `mul_by_b1`, `mul_by_q1` etc. sort of thing?

Pratyush: I mean by a `mul_projective(a: Projective<Self>, b: Self::ScalarField) → Projective<Self>`. This way you can encapsulate the GLV portion in the curve-specific parameters

howardwu:
^Rebased here.

Pratyush (approved):
This LGTM! The last (optional) change that might be good would be to use a helper method to share code between `G1Parameters::mul_projective` and `G2Parameters::mul_projective`. However this can be done in a second PR.

PR #1095, opened 2022-09-22
Deduplicate macro code in parameters

ljedrz:
While looking at the issue of parameters not downloading, I noticed that plenty of code is duplicated in the related macros, making it harder to analyze; it will also cause trouble if in the future something needs to be adjusted there and the dev doesn't know or forgets that there are two spots where the change needs to be made.

### DIFF
--- a/parameters/src/macros.rs
+++ b/parameters/src/macros.rs
@@ -105,6 +105,23 @@ macro_rules! impl_store_and_remote_fetch {
     }
 }
 
+macro_rules! impl_load_bytes_logic_local {
+    ($buffer: expr, $expected_size: expr, $expected_checksum: expr) => {
+        // Ensure the size matches.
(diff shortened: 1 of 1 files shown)


ljedrz:
Cc @julesdesmit, as I see your comments in this code.

jules (approved):
Good catch - LGTM, fine to merge IMO if CI is all green :)

ljedrz:
Cc https://github.com/AleoHQ/snarkVM/pull/1092 for the failed `clippy` job

ljedrz:
The parameters-related CI job failures seem to be unrelated, but it would be best to merge this change only once they can be downloaded again.

ljedrz:
Rebased.

howardwu (approved):
LGTM

PR #1100, opened 2022-09-23
Integrate new coinbase puzzle into blocks

raychu86:
## Motivation

Based on the new coinbase puzzles implemented here: https://github.com/AleoHQ/snarkVM/pull/994

This PR cleans up the new Coinbase Puzzle implementations and begins to integrate its use into the ledger and blocks.


howardwu:
^The above commit now includes the `degree` as a preimage for the **prover polynomial**.

howardwu:
^This commit makes `CoinbaseSolution` --> `Option<CoinbaseSolution>` in `Block`, so that we can support coinbase-less blocks

howardwu:
^Rebased here.

algorithms/src/fft/domain.rs
```diff
@@ -375,6 +375,15 @@ impl<F: FftField> EvaluationDomain<F> {
         self.fft_helper_in_place_with_pc(x_s, FFTOrder::II, &pc)
     }
 
+    pub fn in_order_fft_with_pc<T: DomainCoeff<F>>(&self, x_s: &[T], pc: &FFTPrecomputation<F>) -> Vec<T> {
+        let mut x_s = x_s.to_vec();
+        if self.size() != x_s.len() {
+            x_s.extend(core::iter::repeat(T::zero()).take(self.size() - x_s.len()));
```
ljedrz, line 381: is `self.size()` guaranteed to be larger than `x_s.len()`?

ljedrz: also, this can be done via `x_s.resize(self.size(), T::zero())`

algorithms/src/polycommit/kzg10/mod.rs
```diff
@@ -401,16 +398,38 @@ impl<E: PairingEngine> KZG10<E> {
...
+    /// outputs an evaluation proof for the same.
+    pub fn open_lagrange(
+        lagrange_basis: &LagrangeBasis<E>,
+        domain_elements: &[E::Fr],
+        evaluations: &[E::Fr],
+        point: E::Fr,
+        evaluation_at_point: E::Fr,
+    ) -> Result<KZGProof<E>, PCError> {
+        Self::check_degree_is_too_large(evaluations.len() - 1, lagrange_basis.size())?;
+        assert_eq!(evaluations.len().next_power_of_two(), lagrange_basis.size());
```
ljedrz: is `evaluations.len().next_power_of_two()` guaranteed to not overflow? otherwise `checked_next_power_of_two` would be a better pick

ljedrz: though I don't suppose it would be equal to `lagrange_basis.size()` under overflow conditions, which is grounds for a panic due to `assert_eq`; in any case, it would be a more meaningful error

algorithms/src/polycommit/kzg10/mod.rs
```diff
@@ -401,16 +398,38 @@ impl<E: PairingEngine> KZG10<E> {
...
+        let mut divisor_evals = cfg_iter!(domain_elements).map(|&e| e - point).collect::<Vec<_>>();
```
ljedrz, line 425: is every `e` guaranteed to be greater or equal to `point`? (not sure if `E::Fr` can overflow, just double-checking; there are other operations like this further down)

algorithms/src/polycommit/kzg10/mod.rs
```diff
@@ -401,16 +398,38 @@ impl<E: PairingEngine> KZG10<E> {
...
+        snarkvm_fields::batch_inversion(&mut divisor_evals);
+        cfg_iter_mut!(divisor_evals).zip(evaluations).for_each(|(divisor_eval, &eval)| {
```
ljedrz: should the length of `divisor_evals` be checked for equality with the length of `evaluations`?

vm/compiler/benches/coinbase_puzzle.rs
```diff
@@ -0,0 +1,151 @@
...
+use snarkvm_compiler::{CoinbasePuzzle, EpochChallenge, PuzzleConfig};
+
+use criterion::Criterion;
+use rand::{self, thread_rng, CryptoRng, RngCore};
+
+type CoinbasePuzzleInst = CoinbasePuzzle<Testnet3>;
+
+fn sample_inputs(
+    degree: u32,
+    rng: &mut (impl CryptoRng + RngCore),
```
ljedrz, line 32: nit: unless it's required for compatibility with some API, there is no need to require `CryptoRng` in tests or benchmarks, as it's only a marker trait

vm/compiler/benches/coinbase_puzzle.rs
```diff
@@ -0,0 +1,151 @@
...
+#![allow(clippy::single_element_loop)]
+
+#[macro_use]
+extern crate criterion;
+
+use console::{account::*, network::Testnet3};
+use snarkvm_compiler::{CoinbasePuzzle, EpochChallenge, PuzzleConfig};
+
+use criterion::Criterion;
+use rand::{self, thread_rng, CryptoRng, RngCore};
```
ljedrz, line 26:
```suggestion
use rand::{self, CryptoRng, RngCore};
```
we should instead be using `snarkvm_utilities::TestRng` for greater test/benchmark performance and easier reproducibility; this applies to all other tests and benchmarks as well

vm/compiler/src/coinbase_puzzle/hash.rs
```diff
@@ -0,0 +1,81 @@
...
+use snarkvm_curves::PairingEngine;
+use snarkvm_fields::PrimeField;
+use snarkvm_utilities::CanonicalSerialize;
+
+use blake2::Digest;
+
+#[cfg(feature = "parallel")]
+use rayon::prelude::*;
+
+pub fn hash_to_coefficients<F: PrimeField>(input: &[u8], num_coefficients: u32) -> Result<Vec<F>> {
```
ljedrz:
```suggestion
pub fn hash_to_coefficients<F: PrimeField>(input: &[u8], num_coefficients: u32) -> Vec<F> {
```
it doesn't seem like the `Result` is needed here - all the operations are infallible

vm/compiler/src/coinbase_puzzle/hash.rs
```diff
@@ -0,0 +1,81 @@
...
+        .map(|counter| {
+            let mut input_with_counter = [0u8; 36];
+            input_with_counter[..32].copy_from_slice(&hash);
+            input_with_counter[32..].copy_from_slice(&counter.to_le_bytes());
+            F::from_bytes_le_mod_order(&blake2::Blake2b512::digest(input_with_counter))
+        })
+        .collect())
+}
+
+pub fn hash_to_polynomial<F: PrimeField>(input: &[u8], degree: u32) -> Result<DensePolynomial<F>> {
```
ljedrz:
```suggestion
pub fn hash_to_polynomial<F: PrimeField>(input: &[u8], degree: u32) -> DensePolynomial<F> {
```
if `hash_to_coefficients` doesn't return a `Result`, this function won't need to either

vm/compiler/src/coinbase_puzzle/hash.rs
```diff
@@ -0,0 +1,81 @@
...
+    // Hash the input into coefficients.
+    let coefficients = hash_to_coefficients(input, degree + 1)?;
+    // Construct the polynomial from the coefficients.
+    Ok(DensePolynomial::from_coefficients_vec(coefficients))
+}
+
+pub fn hash_commitment<E: PairingEngine>(commitment: &KZGCommitment<E>) -> Result<E::Fr> {
+    // Convert the commitment into bytes.
+    let mut bytes = Vec::with_capacity(96);
```
ljedrz, line 51: perhaps the `96` could be some `const` (associated or not)?

vm/compiler/src/coinbase_puzzle/hash.rs
```diff
@@ -0,0 +1,81 @@
...
+    let num_commitments = match u32::try_from(commitments.len()) {
+        Ok(num_commitments) => num_commitments,
+        Err(_) => bail!("Cannot hash more than 2^32 commitments: found {}", commitments.len()),
+    };
+    ensure!(!num_commitments.is_zero(), "No commitments provided for hashing");
+
+    // Convert the commitments into bytes.
+    let bytes = commitments
+        .flat_map(|commitment| {
+            let mut bytes = Vec::with_capacity(96);
```
ljedrz, line 72: ditto for the `96`

vm/compiler/src/coinbase_puzzle/helpers/coinbase_solution/bytes.rs
```diff
@@ -0,0 +1,77 @@
...
+            let address = Address::try_from(private_key)?;
+
+            partial_solutions.push(PartialSolution::new(address, u64::rand(&mut rng), KZGCommitment(rng.gen())));
+        }
+        let expected = CoinbaseSolution::new(partial_solutions, KZGProof { w: rng.gen(), random_v: None });
+
+        // Check the byte representation.
+        let expected_bytes = expected.to_bytes_le()?;
+        assert_eq!(expected, CoinbaseSolution::read_le(&expected_bytes[..])?);
+        // assert!(CoinbaseSolution::<CurrentNetwork>::read_le(&expected_bytes[1..]).is_err());
```
ljedrz, line 73: do we want to preserve this additional check? otherwise it can be removed

vm/compiler/src/coinbase_puzzle/helpers/coinbase_solution/serialize.rs
```diff
@@ -0,0 +1,121 @@
...
+}
+
+impl<'de, N: Network> Deserialize<'de> for CoinbaseSolution<N> {
+    /// Deserializes the coinbase solution from a JSON-string or buffer.
+    fn deserialize<D: Deserializer<'de>>(deserializer: D) -> Result<Self, D::Error> {
+        match deserializer.is_human_readable() {
+            true => {
+                let combined_puzzle_solution = serde_json::Value::deserialize(deserializer)?;
+                Ok(Self::new(
+                    serde_json::from_value(combined_puzzle_solution["partial_solutions"].clone())
```
ljedrz:
```suggestion
                    serde_json::from_value(combined_puzzle_solution["partial_solutions"].take())
```
([Value::take](https://docs.rs/serde_json/latest/serde_json/enum.Value.html#method.take)) can be used to avoid the clone here and down below

vm/compiler/src/coinbase_puzzle/helpers/epoch_challenge/mod.rs
```diff
@@ -0,0 +1,84 @@
...
+    /// The evaluations of the epoch polynomial over the product domain.
+    epoch_polynomial_evaluations: EvaluationsOnDomain<<N::PairingCurve as PairingEngine>::Fr>,
+}
+
+impl<N: Network> EpochChallenge<N> {
+    /// Initializes a new epoch challenge.
+    pub fn new(epoch_number: u32, epoch_block_hash: N::BlockHash, degree: u32) -> Result<Self> {
+        // Construct the 'input' as '( epoch_number || epoch_block_hash )'
+        let input: Vec<u8> =
+            epoch_number.to_le_bytes().into_iter().chain(epoch_block_hash.to_bytes_le()?.into_iter()).collect();
```
ljedrz:
```suggestion
            epoch_number.to_le_bytes().into_iter().chain(epoch_block_hash.to_bytes_le()?).collect();
```
that call shouldn't be necessary, since [`Iterator::chain`](https://doc.rust-lang.org/std/iter/trait.Iterator.html#method.chain) already expects an `IntoIterator`

vm/compiler/src/coinbase_puzzle/helpers/epoch_challenge/mod.rs
```diff
@@ -0,0 +1,84 @@
...
+        let epoch_polynomial = hash_to_polynomial::<<N::PairingCurve as PairingEngine>::Fr>(&input, degree)?;
+        let domain = EvaluationDomain::new((2 * degree - 1) as usize).unwrap();
```
ljedrz: note: `degree` can't be equal to `0`; if this is guaranteed, feel free to mark this comment as resolved

vm/compiler/src/coinbase_puzzle/helpers/epoch_challenge/mod.rs
```diff
@@ -0,0 +1,84 @@
...
+
+    /// Returns the evaluations of the epoch polynomial over the product domain.
+    pub const fn epoch_polynomial_evaluations(&self) -> &EvaluationsOnDomain<<N::PairingCurve as PairingEngine>::Fr> {
+        &self.epoch_polynomial_evaluations
+    }
+
+    /// Returns the number of coefficients of the epoch polynomial.
+    pub fn degree(&self) -> Result<u32> {
+        // Cast the degree into a u32.
+        match u32::try_from(self.epoch_polynomial.degree()) {
```
ljedrz: `self.epoch_polynomial.degree()` returns an `usize`; it would probably be a good idea to explicitly make it `u64`, especially if it's the only valid size

vm/compiler/src/coinbase_puzzle/helpers/epoch_challenge/mod.rs
```diff
@@ -0,0 +1,84 @@
...
+    }
+
+    /// Returns the evaluations of the epoch polynomial over the product domain.
+    pub const fn epoch_polynomial_evaluations(&self) -> &EvaluationsOnDomain<<N::PairingCurve as PairingEngine>::Fr> {
+        &self.epoch_polynomial_evaluations
+    }
+
+    /// Returns the number of coefficients of the epoch polynomial.
+    pub fn degree(&self) -> Result<u32> {
+        // Cast the degree into a u32.
```
ljedrz:
```suggestion
        // Convert the degree into a u32.
```
(technically not a cast)

vm/compiler/src/coinbase_puzzle/helpers/mod.rs
```diff
@@ -0,0 +1,90 @@
...
+    /// The verifying key of the coinbase puzzle.
+    pub verifying_key: CoinbaseVerifyingKey<N>,
+}
+
+impl<N: Network> CoinbaseProvingKey<N> {
+    /// Obtain powers for the underlying KZG10 construction
+    pub fn powers(&self) -> Powers<N::PairingCurve> {
+        Powers {
+            powers_of_beta_g: self.powers_of_beta_g.as_slice().into(),
+            powers_of_beta_times_gamma_g: Cow::Owned(vec![]),
```
ljedrz:
```suggestion
            powers_of_beta_times_gamma_g: Default::default(),
```
also applicable to other such cases

vm/compiler/src/coinbase_puzzle/helpers/mod.rs
```diff
@@ -0,0 +1,90 @@
...
+        }
+    }
+
+    /// Obtain elements of the SRS in the lagrange basis powers.
+    pub fn lagrange_basis(&self) -> LagrangeBasis<N::PairingCurve> {
+        LagrangeBasis {
+            lagrange_basis_at_beta_g: Cow::Borrowed(self.lagrange_basis_at_beta_g.as_slice()),
+            powers_of_beta_times_gamma_g: Cow::Owned(vec![]),
```
ljedrz, line 76:
```suggestion
            powers_of_beta_times_gamma_g: Default::default(),
```
(ditto)

vm/compiler/src/coinbase_puzzle/helpers/partial_solution/mod.rs
```diff
@@ -0,0 +1,68 @@
...
+
+impl<N: Network> PartialSolution<N> {
+    /// Initializes a new instance of the partial solution.
+    pub const fn new(address: Address<N>, nonce: u64, commitment: KZGCommitment<N::PairingCurve>) -> Self {
+        Self { address, nonce, commitment }
+    }
+
+    /// Returns the address of the prover.
+    pub const fn address(&self) -> &Address<N> {
+        &self.address
```
ljedrz:
```suggestion
    pub const fn address(&self) -> Address<N> {
        self.address
```
`Address` is `Copy`

vm/compiler/src/coinbase_puzzle/helpers/partial_solution/mod.rs
```diff
@@ -0,0 +1,68 @@
...
+    }
+
+    /// Returns the nonce for the solution.
+    pub const fn nonce(&self) -> u64 {
+        self.nonce
+    }
+
+    /// Returns the commitment for the solution.
+    pub const fn commitment(&self) -> &KZGCommitment<N::PairingCurve> {
+        &self.commitment
```
ljedrz:
```suggestion
    pub const fn commitment(&self) -> KZGCommitment<N::PairingCurve> {
        self.commitment
```
`KZGCommitment` is `Copy`

vm/compiler/src/coinbase_puzzle/helpers/partial_solution/serialize.rs
```diff
@@ -0,0 +1,103 @@
...
+}
+
+impl<'de, N: Network> Deserialize<'de> for PartialSolution<N> {
+    /// Deserializes the partial solution from a JSON-string or buffer.
+    fn deserialize<D: Deserializer<'de>>(deserializer: D) -> Result<Self, D::Error> {
+        match deserializer.is_human_readable() {
+            true => {
+                let partial_prover_solution = serde_json::Value::deserialize(deserializer)?;
+                Ok(Self::new(
+                    serde_json::from_value(partial_prover_solution["address"].clone()).map_err(de::Error::custom)?,
```
ljedrz, line 42: same comment as https://github.com/AleoHQ/snarkVM/pull/1100/files#r996780312

vm/compiler/src/coinbase_puzzle/helpers/prover_solution/mod.rs
```diff
@@ -0,0 +1,104 @@
...
+        let claimed_value = epoch_evaluation * prover_evaluation;
+
+        // Check the KZG proof.
+        Ok(KZG10::check(verifying_key, self.commitment(), challenge_point, claimed_value, self.proof())?)
+    }
+
+    /// Returns the address of the prover.
+    pub const fn address(&self) -> &Address<N> {
+        self.partial_solution.address()
+    }
```
ljedrz:
```suggestion
    pub const fn address(&self) -> Address<N> {
        self.partial_solution.address()
    }
```
same as https://github.com/AleoHQ/snarkVM/pull/1100/files#r996810536

vm/compiler/src/coinbase_puzzle/helpers/prover_solution/mod.rs
```diff
@@ -0,0 +1,104 @@
...
+
+    /// Returns the nonce for the solution.
+    pub const fn nonce(&self) -> u64 {
+        self.partial_solution.nonce()
+    }
+
+    /// Returns the commitment for the solution.
+    pub const fn commitment(&self) -> &KZGCommitment<N::PairingCurve> {
+        self.partial_solution.commitment()
+    }
```
ljedrz:
```suggestion
    pub const fn commitment(&self) -> KZGCommitment<N::PairingCurve> {
        self.partial_solution.commitment()
    }
```

vm/compiler/src/coinbase_puzzle/helpers/prover_solution/mod.rs
```diff
@@ -0,0 +1,104 @@
...
+
+    /// Returns the proof for the solution.
+    pub const fn proof(&self) -> &KZGProof<N::PairingCurve> {
+        &self.proof
+    }
```
ljedrz:
```suggestion
    pub const fn proof(&self) -> KZGProof<N::PairingCurve> {
        self.proof
    }
```

vm/compiler/src/coinbase_puzzle/helpers/prover_solution/serialize.rs
```diff
@@ -0,0 +1,113 @@
...
+}
+
+impl<'de, N: Network> Deserialize<'de> for ProverSolution<N> {
+    /// Deserializes the prover solution from a JSON-string or buffer.
+    fn deserialize<D: Deserializer<'de>>(deserializer: D) -> Result<Self, D::Error> {
+        match deserializer.is_human_readable() {
+            true => {
+                let prover_puzzle_solution = serde_json::Value::deserialize(deserializer)?;
+                Ok(Self::new(
+                    serde_json::from_value(prover_puzzle_solution["partial_solution"].clone())
```
ljedrz, line 44: same as https://github.com/AleoHQ/snarkVM/pull/1100/files#r996780312

vm/compiler/src/coinbase_puzzle/helpers/prover_solution/string.rs
```diff
@@ -0,0 +1,65 @@
...
+        let mut rng = TestRng::default();
+        let private_key = PrivateKey::<CurrentNetwork>::new(&mut rng)?;
+        let address = Address::try_from(private_key)?;
+
+        // Sample a new prover puzzle solution.
+        let partial_solution = PartialSolution::new(address, u64::rand(&mut rng), KZGCommitment(rng.gen()));
+        let expected = ProverSolution::new(partial_solution, KZGProof { w: rng.gen(), random_v: None });
+
+        // Check the string representation.
+        let candidate = format!("{expected}");
```
ljedrz:
```suggestion
        let candidate = expected.to_string();
```

vm/compiler/src/coinbase_puzzle/mod.rs
```diff
@@ -0,0 +1,363 @@
...
+
+#[cfg(feature = "parallel")]
+use rayon::prelude::*;
+
+#[derive(Clone)]
+pub enum CoinbasePuzzle<N: Network> {
+    /// The prover contains the coinbase puzzle proving key.
+    Prover(Arc<CoinbaseProvingKey<N>>),
+    /// The verifier contains the coinbase puzzle verifying key.
+    Verifier(Arc<CoinbaseVerifyingKey<N>>),
```
ljedrz, line 55: it might be a good idea to comment on why these are `Arc`ed

vm/compiler/src/coinbase_puzzle/mod.rs
```diff
@@ -0,0 +1,363 @@
...
+
+        Ok(Self::Prover(Arc::new(pk)))
+    }
+
+    // TODO (raychu86): Create a "candidate_prove", just output the commitment -> then finalize the prove.
+    /// Returns a prover solution to the coinbase puzzle.
+    pub fn prove(
+        &self,
+        epoch_challenge: &EpochChallenge<N>,
+        address: &Address<N>,
```
ljedrz:
```suggestion
        address: Address<N>,
```

vm/compiler/src/coinbase_puzzle/mod.rs
```diff
@@ -0,0 +1,363 @@
...
+        let product_evaluations = {
+            let polynomial_evaluations = pk.product_domain.in_order_fft_with_pc(&polynomial, &pk.fft_precomputation);
+            let product_evaluations = pk.product_domain.mul_polynomials_in_evaluation_domain(
+                &polynomial_evaluations,
+                &epoch_challenge.epoch_polynomial_evaluations().evaluations,
+            );
+            product_evaluations
+        };
+        let (commitment, _rand) =
+            KZG10::commit_lagrange(&pk.lagrange_basis(), &product_evaluations, None, &AtomicBool::default(), None)?;
```
ljedrz:
```suggestion
            KZG10::commit_lagrange(&pk.lagrange_basis(), &product_evaluations, None, &Default::default(), None)?;
```

vm/compiler/src/coinbase_puzzle/mod.rs
```diff
@@ -0,0 +1,363 @@
...
+            pk.product_domain_elements(),
+            &product_evaluations,
+            point,
+            product_eval_at_point,
+        )?;
+        ensure!(!proof.is_hiding(), "The prover solution must contain a non-hiding proof");
+
+        debug_assert!(KZG10::check(&pk.verifying_key, &commitment, point, product_eval_at_point, &proof)?);
+
+        Ok(ProverSolution::new(PartialSolution::new(*address, nonce, commitment), proof))
```
ljedrz:
```suggestion
        Ok(ProverSolution::new(PartialSolution::new(address, nonce, commitment), proof))
```

vm/compiler/src/coinbase_puzzle/mod.rs
```diff
@@ -0,0 +1,363 @@
...
+            Self::Prover(coinbase_proving_key) => coinbase_proving_key,
+            Self::Verifier(_) => bail!("Cannot accumulate the coinbase puzzle with a verifier"),
+        };
+
+        let (prover_polynomials, partial_solutions): (Vec<_>, Vec<_>) = cfg_iter!(prover_solutions)
+            .filter_map(|solution| {
+                if solution.proof().is_hiding() {
+                    return None;
+                }
+                let polynomial = solution.to_prover_polynomial(epoch_challenge).unwrap();
```
ljedrz: is this `unwrap` guaranteed to not trigger?

vm/compiler/src/coinbase_puzzle/mod.rs
```diff
@@ -0,0 +1,363 @@
...
+
+        // Construct the provers polynomial.
+        let accumulated_prover_polynomial = cfg_into_iter!(prover_polynomials)
+            .zip_eq(challenges)
+            .fold(DensePolynomial::zero, |mut accumulator, (mut prover_polynomial, challenge)| {
+                prover_polynomial *= challenge;
+                accumulator += &prover_polynomial;
+                accumulator
+            })
+            .sum::<DensePolynomial<_>>();
```
ljedrz, line 208: is this `sum` even needed? the earlier `fold` should already return a single value, and it looks like it's doing the sum

vm/compiler/src/coinbase_puzzle/mod.rs
```diff
@@ -0,0 +1,363 @@
...
+            None => bail!("Missing the accumulator challenge point"),
+        };
+
+        // Compute the accumulator evaluation.
+        let mut accumulator_evaluation = cfg_iter!(prover_polynomials)
+            .zip_eq(&challenge_points)
+            .fold(<N::PairingCurve as PairingEngine>::Fr::zero, |accumulator, (prover_polynomial, challenge_point)| {
+                accumulator + (prover_polynomial.evaluate(accumulator_point) * challenge_point)
+            })
+            .sum();
```
ljedrz, line 301: same question as in https://github.com/AleoHQ/snarkVM/pull/1100/files#r996828432

vm/compiler/src/coinbase_puzzle/mod.rs
```diff
@@ -0,0 +1,363 @@
...
+            coinbase_verifying_key,
+            &accumulator_commitment,
+            accumulator_point,
+            accumulator_evaluation,
+            coinbase_solution.proof(),
+        )?)
+    }
+
+    /// Returns the coinbase proving key.
+    pub fn coinbase_proving_key(&self) -> Result<&CoinbaseProvingKey<N>> {
```
ljedrz, line 328: note: the `CoinbaseProvingKey` is `Arc`ed; returning a reference here is fine, just pointing it out in case it's troublesome elsewhere

vm/compiler/src/coinbase_puzzle/mod.rs
```diff
@@ -0,0 +1,363 @@
...
+        match self {
+            Self::Prover(coinbase_proving_key) => Ok(coinbase_proving_key),
+            Self::Verifier(_) => bail!("Cannot fetch the coinbase proving key with a verifier"),
+        }
+    }
+
+    /// Returns the coinbase verifying key.
+    pub fn coinbase_verifying_key(&self) -> Result<&CoinbaseVerifyingKey<N>> {
```
ljedrz, line 336: ditto, just wrt. `CoinbaseVerifyingKey`

vm/compiler/src/coinbase_puzzle/mod.rs
```diff
@@ -0,0 +1,363 @@
...
+            Self::Verifier(coinbase_verifying_key) => Ok(coinbase_verifying_key),
+        }
+    }
+}
+
+impl<N: Network> CoinbasePuzzle<N> {
+    /// Returns the prover polynomial for the coinbase puzzle.
+    fn prover_polynomial(
+        epoch_challenge: &EpochChallenge<N>,
+        address: &Address<N>,
```
ljedrz:
```suggestion
        address: Address<N>,
```

vm/compiler/src/coinbase_puzzle/mod.rs
```diff
@@ -0,0 +1,363 @@
...
+        nonce: u64,
+    ) -> Result<DensePolynomial<<N::PairingCurve as PairingEngine>::Fr>> {
+        let input = {
+            let mut bytes = [0u8; 80];
+            bytes[..40].copy_from_slice(&epoch_challenge.to_bytes_le()?);
```
ljedrz: in the absence of a related `const` it might make sense to at least preface this operation with a `debug_assert` that the length of the serialized `epoch_challenge` is `40`

vm/compiler/src/coinbase_puzzle/tests.rs
```diff
@@ -0,0 +1,53 @@
...
+use super::*;
+use console::{account::*, network::Testnet3};
+use snarkvm_utilities::Uniform;
+
+use rand::RngCore;
+
+#[test]
+fn test_coinbase_puzzle() {
+    let max_degree = 1 << 15;
+    let mut rng = rand::thread_rng();
```
ljedrz: this can be `TestRng::default()` instead

vm/compiler/src/ledger/block/bytes.rs
```diff
@@ -32,10 +32,22 @@ impl<N: Network> FromBytes for Block<N> {
...
         let header = FromBytes::read_le(&mut reader)?;
         let transactions = FromBytes::read_le(&mut reader)?;
+
+        // Write the coinbase proof.
+        let coinbase_variant = u8::read_le(&mut reader)?;
+        let coinbase_proof = match coinbase_variant {
+            0 => None,
+            1 => Some(FromBytes::read_le(&mut reader)?),
+            _ => return Err(error("Invalid coinbase proof variant")),
+        };
```
ljedrz, line 42: it might make sense to introduce an
```
#[repr(u8)]
enum CoinbaseVariant {
    Zero,
    One,
}
```
which implements `TryFrom<u8>`

vm/compiler/src/ledger/block/header/mod.rs
```diff
@@ -83,6 +90,11 @@ impl<N: Network> Header<N> {
         &self.transactions_root
     }
 
+    /// Returns the coinbase puzzle accumulator point in the block header.
+    pub const fn coinbase_accumulator_point(&self) -> &Field<N> {
+        &self.coinbase_accumulator_point
```
ljedrz:
```suggestion
    pub const fn coinbase_accumulator_point(&self) -> Field<N> {
        self.coinbase_accumulator_point
```
`Field` is `Copy`

vm/compiler/src/ledger/block/header/serialize.rs
```diff
@@ -41,6 +42,7 @@ impl<'de, N: Network> Deserialize<'de> for Header<N> {
                 Ok(Self::from(
                     serde_json::from_value(header["previous_state_root"].clone()).map_err(de::Error::custom)?,
                     serde_json::from_value(header["transactions_root"].clone()).map_err(de::Error::custom)?,
+                    serde_json::from_value(header["coinbase_accumulator_point"].clone()).map_err(de::Error::custom)?,
```
ljedrz: same as https://github.com/AleoHQ/snarkVM/pull/1100/files#r996780312, applies to all the clones

vm/compiler/src/ledger/block/mod.rs
```diff
@@ -104,6 +116,11 @@ impl<N: Network> Block<N> {
         self.previous_hash
     }
 
+    /// Returns the coinbase proof.
+    pub const fn coinbase_proof(&self) -> &Option<CoinbaseSolution<N>> {
+        &self.coinbase_proof
```
ljedrz:
```suggestion
    pub const fn coinbase_proof(&self) -> Option<&CoinbaseSolution<N>> {
        self.coinbase_proof.as_ref()
```
unless `CoinbaseSolution` is `Copy`, in which case just `self.coinbase_proof`

vm/compiler/src/ledger/block/serialize.rs
```diff
@@ -21,11 +21,16 @@ impl<N: Network> Serialize for Block<N> {
...
         match serializer.is_human_readable() {
             true => {
-                let mut block = serializer.serialize_struct("Block", 5)?;
+                let mut block = serializer.serialize_struct("Block", 6)?;
                 block.serialize_field("block_hash", &self.block_hash)?;
                 block.serialize_field("previous_hash", &self.previous_hash)?;
                 block.serialize_field("header", &self.header)?;
                 block.serialize_field("transactions", &self.transactions)?;
+
+                if let Some(coinbase_proof) = &self.coinbase_proof {
```
ljedrz:
```suggestion
                if let Some(coinbase_proof) = self.coinbase_proof() {
```

vm/compiler/src/ledger/helpers/mod.rs
```diff
@@ -0,0 +1,401 @@
...
+    // Compute the denominator.
+    let denominator = anchor_height_at_year_1 * (anchor_height_at_year_1 + 1);
+    // Return the anchor reward.
+    numerator / denominator
+}
+
+/// Returns the anchor block height after a given number of years for a specific anchor time.
+pub const fn anchor_block_height(anchor_time: u16, num_years: u32) -> u32 {
+    // Calculate the number of seconds in a year.
+    const SECONDS_IN_A_YEAR: u32 = 60 * 60 * 24 * 365;
```
ljedrz, line 78: (I'm assuming that leap years can be ignored)

vm/compiler/src/ledger/helpers/mod.rs
```diff
@@ -0,0 +1,401 @@
...
+    use snarkvm_utilities::TestRng;
+
+    use rand::Rng;
+
+    const ITERATIONS: usize = 1000;
+
+    #[test]
+    fn test_anchor_reward() {
+        let reward = anchor_reward::<STARTING_SUPPLY, ANCHOR_TIME>();
+        assert_eq!(reward, 8);
```
ljedrz: is there some `const` we can check against here?

vm/compiler/src/ledger/helpers/mod.rs
```diff
@@ -0,0 +1,401 @@
...
+
+        // Decreasing the anchor time will decrease the reward.
+        let smaller_reward = anchor_reward::<STARTING_SUPPLY, { ANCHOR_TIME - 1 }>();
+        assert!(reward > smaller_reward);
+    }
+
+    #[test]
+    fn test_staking_reward() {
+        let reward = staking_reward::<STARTING_SUPPLY, ANCHOR_TIME>();
+        assert_eq!(reward, 17_440_385);
```
ljedrz: ditto

vm/compiler/src/ledger/helpers/mod.rs
```diff
@@ -0,0 +1,401 @@
...
+
+    #[test]
+    fn test_coinbase_reward() {
+        let reward = coinbase_reward::<STARTING_SUPPLY, ANCHOR_TIME, NUM_BLOCKS_PER_EPOCH>(
+            GENESIS_TIMESTAMP,
+            GENESIS_TIMESTAMP + ANCHOR_TIME as i64,
+            1,
+        )
+        .unwrap();
+        assert_eq!(reward, 126_143_992);
```
ljedrz: ditto

vm/compiler/src/ledger/helpers/mod.rs
```diff
@@ -0,0 +1,401 @@
...
+    }
+}
+
+/// Calculates the anchor reward.
+///     R_anchor = floor((2 * S) / (H_Y10 * (H_Y10 + 1))).
+///     S = Starting supply.
+///     H_Y10 = Expected block height at year 10.
+const fn anchor_reward<const STARTING_SUPPLY: u64, const ANCHOR_TIME: u16>() -> u64 {
+    // Calculate the anchor block height at year 10.
+    let anchor_height_at_year_1 = anchor_block_height(ANCHOR_TIME, 10) as u64;
```
yhio: anchor_height_at_year_1 => anchor_height_at_year_10

ljedrz (commented):
Left some comments.

vm/compiler/src/coinbase_puzzle/helpers/coinbase_solution/serialize.rs
ljedrz: I filed https://github.com/AleoHQ/snarkVM/pull/1181 for the existing implementations; these changes are rather trivial, so it's low priority wrt. any conflicts

vm/compiler/src/ledger/helpers/mod.rs
howardwu: This logic is used for the first decade (approximately), thus it should witness at most 2 leap days, which is acceptable. (For context, the retarget algorithm is a fixed-point approximation with an error bound greater than 2 / (365 * 10) * 100 %)

vm/compiler/src/coinbase_puzzle/hash.rs
Pratyush: +1. We should also change it to `[u8; 96]`:
```suggestion
    let mut bytes = [0u8; 96];
```

vm/compiler/src/coinbase_puzzle/helpers/epoch_challenge/mod.rs
Pratyush: Good point; we should take a `NonZeroU32` as input.

vm/compiler/src/coinbase_puzzle/mod.rs
Pratyush: Good catch. It is potentially triggerable; the current limitation is because there's no `try_filter_map` in either `rayon` or `itertools`.

vm/compiler/src/coinbase_puzzle/mod.rs
Pratyush: AFAIK the `rayon` `fold` does require a sum at the end: https://docs.rs/rayon/latest/rayon/iter/trait.ParallelIterator.html#method.fold

ljedrz: Ah yes, I was thinking of the single-threaded one 👍

ljedrz: However, this means that this code might not compile without the `parallel` feature.

vm/compiler/src/coinbase_puzzle/mod.rs
raychu86: Great catch, seemed like we were using extra bytes (which was still safe, since the `epoch_polynomial` is deterministically derived). I now made the objects more explicit with the following: 

```
let input = {
            let mut bytes = [0u8; 76];
            bytes[..4].copy_from_slice(&epoch_challenge.epoch_number().to_bytes_le()?);
            bytes[4..36].copy_from_slice(&epoch_challenge.epoch_block_hash().to_bytes_le()?);
            bytes[36..68].copy_from_slice(&address.to_bytes_le()?);
            bytes[68..].copy_from_slice(&nonce.to_le_bytes());
            bytes
        };
```

vm/compiler/src/coinbase_puzzle/mod.rs
Pratyush: Yeah, I was thinking about that; not sure of any easy way to unify the two. Maybe we should introduce a `cfg_fold`, where the `cfg_fold` returns `[result_of_fold].into_iter()`?

vm/compiler/src/ledger/block/bytes.rs
Pratyush: We can even be more explicit:
```rust
#[repr(u8)]
enum CoinbaseVariant {
	Zero = 0,
	One = 1,
}
```

vm/compiler/src/coinbase_puzzle/hash.rs
```diff
@@ -0,0 +1,81 @@
...
+    let coefficients = hash_to_coefficients(input, degree + 1)?;
+    // Construct the polynomial from the coefficients.
+    Ok(DensePolynomial::from_coefficients_vec(coefficients))
+}
+
+pub fn hash_commitment<E: PairingEngine>(commitment: KZGCommitment<E>) -> Result<E::Fr> {
+    // Convert the commitment into bytes.
+    let mut bytes = Vec::with_capacity(96);
+    commitment.serialize_uncompressed(&mut bytes)?;
+    ensure!(bytes.len() == 96, "Invalid commitment byte length for hashing");
```
niklaslong, line 53: ditto here for the assertion (96 should be a `const`).

vm/compiler/src/coinbase_puzzle/mod.rs
ljedrz: the result of a fold would usually not be iterable in the single-threaded mode; `cfg_fold` would probably need to already return the base type of the fold

vm/compiler/src/coinbase_puzzle/helpers/partial_solution/mod.rs
```diff
@@ -0,0 +1,68 @@
...
+    pub fn to_prover_polynomial(
+        &self,
+        epoch_challenge: &EpochChallenge<N>,
+    ) -> Result<DensePolynomial<<N::PairingCurve as PairingEngine>::Fr>> {
+        CoinbasePuzzle::prover_polynomial(epoch_challenge, self.address(), self.nonce())
+    }
+
+    /// Returns the target of the solution.
+    pub fn to_target(&self) -> Result<u64> {
+        Ok(u64::MAX / sha256d_to_u64(&self.commitment.to_bytes_le()?))
```
niklaslong: Is `sha256d_to_u64` guaranteed to be non-zero?

vm/compiler/src/coinbase_puzzle/helpers/mod.rs
Pratyush: Isn't being explicit better?

vm/compiler/src/coinbase_puzzle/mod.rs
Pratyush: I mean the follow definition: 
```rust
macro_rules! cfg_fold {
    ($e: expr, $fold_body: expr) => {
        #[cfg(feature = parallel)]
        $e.fold($fold_body).collect::<Vec<_>>()
        #[cfg(not(feature = parallel))]
        [$e.fold($fold_body)]
    }
}
```

algorithms/src/polycommit/kzg10/mod.rs
Pratyush: Good catch. `E::Fr` is a field, so arithmetic is wrapping, which means there's no danger of overflows. However `e - point` must not be zero; added a check for that case in [`a2f44b5` (#1100)](https://github.com/AleoHQ/snarkVM/pull/1100/commits/a2f44b5d4ea8d5120009f808ec77a853cab3daa8).

algorithms/src/polycommit/kzg10/mod.rs
Pratyush: Changed to `zip_eq` in [`a2f44b5` (#1100)](https://github.com/AleoHQ/snarkVM/pull/1100/commits/a2f44b5d4ea8d5120009f808ec77a853cab3daa8)

vm/compiler/benches/coinbase_puzzle.rs
Pratyush: It's probably a good idea to replicate the production environment in benches, no?

vm/compiler/src/coinbase_puzzle/helpers/epoch_challenge/mod.rs
Pratyush: I've handled this by `ensure`-ing that `degree` non-zero in [`ef950de` (#1100)](https://github.com/AleoHQ/snarkVM/pull/1100/commits/ef950dea9ab89461d8f6e1768b5d303ee7cd66f7)

vm/compiler/src/coinbase_puzzle/helpers/partial_solution/mod.rs
raychu86: Good catch! It is not guaranteed to be non-zero

vm/compiler/benches/coinbase_puzzle.rs
ljedrz: that depends; in terms of marker traits it doesn't make a difference, but it could be helpful if some API needs to be adhered to; as for the source of randomness, `thread_rng` might provide more representative results, but it will reduce the usefulness of the benchmark as a means to find potential performance improvements (which are unlikely to be introduced on the randomness side in production)

vm/compiler/src/coinbase_puzzle/mod.rs
ljedrz: I guess that could work; maybe even without the `collect` in the `parallel` case (but I'm not 100% sure about that)

vm/compiler/src/coinbase_puzzle/helpers/mod.rs
ljedrz: `Default::default()` is more future-proof if the underlying object were to change at some point in time; other than that I have no issues with it being explicit if it's deemed practical

howardwu (approved):

PR #1101, opened 2022-09-23
Split out the compiler crate

ljedrz:
Many users will likely only be interested in the `snarkvm-compiler` crate, so it makes sense to isolate it.

An alternative would be to introduce a `compiler` feature for `snarkvm` (in a fashion similar to the other features that are currently there), but it results in a much larger lockfile (in an empty project using just that dependency): `2561` lines, compared to `1548` when importing an isolated `snarkvm-compiler` crate.

In addition, the `dev-dependencies` of `snarkvm-parameters` are also made more granular, making its compilation faster.


ljedrz:

Cc https://github.com/AleoHQ/snarkVM/pull/1092 for the `clippy` one.

howardwu:
@ljedrz let's wait on this PR until #1100 is merged, which introduces time-sensitive changes for snarkVM

ljedrz:
Sounds good; it's simple enough to redo 👌

howardwu:
Once #1100 is merged, let's move to the `testnet3.2` branch and pursue the approach in #1165.

ljedrz:
Done in https://github.com/AleoHQ/snarkVM/pull/1201.

PR #1105, opened 2022-09-26
Guard against overflow when using next_power_of_two

### LINKED ISSUE #1103: Use a checked power of two method for added safety
We have many instances of `.next_power_of_two()` which should be switched to the checked variant as per this PR.

_Originally posted by @howardwu in https://github.com/AleoHQ/snarkVM/issues/1091#issuecomment-1256969379_

ljedrz:
This PR changes all uses of `next_power_of_two` to `checked_next_power_of_two` outside of tests and benchmarks. @howardwu it's possible that one or more of these changes is overzealous, and I'm happy to adjust if need be.

Fixes https://github.com/AleoHQ/snarkVM/issues/1103.


ljedrz:

howardwu (approved):

PR #1108, opened 2022-09-26
profiling in snarkVM fields with an optional feature

pablodeymo:
## Motivation

SnarkVM fields has some calls to profiling functions that are being compiled even in release mode.
This PR proposes the use of the feature `profiler` to enable conditional compilation for these calls of profiling macros.

## Test Plan

(Write your test plan here)

## Related PRs

(Link your related PRs here)

### DIFF
--- a/fields/Cargo.toml
+++ b/fields/Cargo.toml
@@ -31,6 +31,7 @@ default-features = false
 [dependencies.aleo-std]
 version = "0.1.14"
 default-features = false
+optional = true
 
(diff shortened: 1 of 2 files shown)


ljedrz:
This will not be needed with https://github.com/AleoHQ/aleo-std/pull/4, if accepted.

howardwu:
Closing as https://github.com/AleoHQ/aleo-std/pull/4 was merged.

PR #1136, opened 2022-10-04
Updates equality check on string; temporarily fix CI

howardwu:
## Motivation

Updates equality check on string.

This PR temporarily fixes CI (as per #986), however it comes as a temporary solution as we need to rearchitect the string design.

### DIFF
--- a/circuit/types/string/src/helpers/from_bits.rs
+++ b/circuit/types/string/src/helpers/from_bits.rs
@@ -25,7 +25,7 @@ impl<E: Environment> FromBits for StringType<E> {
         let mode = Mode::combine(bits_le[0].eject_mode(), bits_le.iter().skip(1).map(|bit| bit.eject_mode()));
 
         // Inject the string size in bytes.
-        let size_in_bytes = Self::inject_size_in_bytes(bits_le, mode);
+        let size_in_bytes = Self::inject_size_in_bytes(mode, bits_le);
 
         // Ensure the list of booleans is byte-aligned.
         let num_bits = bits_le.len();
@@ -45,7 +45,7 @@ impl<E: Environment> FromBits for StringType<E> {
         let mode = Mode::combine(bits_be[0].eject_mode(), bits_be.iter().skip(1).map(|bit| bit.eject_mode()));
 
         // Inject the string size in bytes.
-        let size_in_bytes = Self::inject_size_in_bytes(bits_be, mode);
+        let size_in_bytes = Self::inject_size_in_bytes(mode, bits_be);
 
         // Ensure the list of booleans is byte-aligned.
         let num_bits = bits_be.len();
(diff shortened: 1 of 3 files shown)


circuit/types/string/src/helpers/from_bits.rs
```diff
@@ -21,25 +21,73 @@ impl<E: Environment> FromBits for StringType<E> {
...
+
+        // Inject the string size in bytes.
+        let size_in_bytes = Self::inject_size_in_bytes(mode, bits_be);
+
         // Ensure the list of booleans is byte-aligned.
         let num_bits = bits_be.len();
         match num_bits % 8 == 0 {
-            true => StringType { mode: bits_be.eject_mode(), bytes: bits_be.chunks(8).map(U8::from_bits_be).collect() },
+            true => StringType {
+                mode: bits_be.eject_mode(),
```
Pratyush: We already computed `mode` above
```suggestion
                mode,
```

circuit/types/string/src/helpers/from_bits.rs
```diff
@@ -21,25 +21,73 @@ impl<E: Environment> FromBits for StringType<E> {
...
+
+        // Inject the string size in bytes.
+        let size_in_bytes = Self::inject_size_in_bytes(mode, bits_le);
+
         // Ensure the list of booleans is byte-aligned.
         let num_bits = bits_le.len();
         match num_bits % 8 == 0 {
-            true => StringType { mode: bits_le.eject_mode(), bytes: bits_le.chunks(8).map(U8::from_bits_le).collect() },
+            true => StringType {
+                mode: bits_le.eject_mode(),
```
Pratyush: We already computed `mode above
```suggestion
                mode,
```

circuit/types/string/src/helpers/from_bits.rs
```diff
@@ -21,25 +21,73 @@ impl<E: Environment> FromBits for StringType<E> {
 
     /// Initializes a new string from a list of little-endian bits *with* trailing zeros (to byte-alignment).
     fn from_bits_le(bits_le: &[Self::Boolean]) -> Self {
+        // Determine the mode.
+        let mode = Mode::combine(bits_le[0].eject_mode(), bits_le.iter().skip(1).map(|bit| bit.eject_mode()));
+
+        // Inject the string size in bytes.
+        let size_in_bytes = Self::inject_size_in_bytes(mode, bits_le);
+
         // Ensure the list of booleans is byte-aligned.
```
Pratyush: We can push the construction of the struct, and the computation of the `mode` also to the helper function. The only part that depends on endianness is the call to `U8::from_bits_{le,be}`, but we can pass that as a closure.

circuit/types/string/src/helpers/from_bits.rs
howardwu: Great catch, updating.

circuit/types/string/src/helpers/from_bits.rs
howardwu: Great catch, updating.

raychu86 (approved):

PR #1139, opened 2022-10-04
[Update] Package Deployment

ilitteri:
## Motivation

- `DeploymentRequest` should contain a deployment transaction instead of a deployment because the client should be responsible for executing the transaction that involves the additional fee execution and then broadcast it.
- `DeploymentResponse` should contain the same deployment transaction that was sent in the `DeploymentRequest`.
- In order to execute the deployment transaction the client must provide the private key and an unspent record.

In another iteration, this function should be responsible only for executing the transaction and sending it to the Aleo CLI for it to broadcast it.
## Test Plan

The tests were updated.

## Related PRs/Issues

- https://github.com/AleoHQ/aleo/issues/364


vm/package/deploy.rs
```diff
@@ -14,31 +14,34 @@
...
     address: Address<N>,
     program_id: ProgramID<N>,
 }
 
 impl<N: Network> DeployRequest<N> {
     /// Sends the request to the given endpoint.
-    pub fn new(deployment: Deployment<N>, address: Address<N>, program_id: ProgramID<N>) -> Self {
-        Self { deployment, address, program_id }
+    pub fn new(transaction: Transaction<N>, address: Address<N>, program_id: ProgramID<N>) -> Self {
+        Self { transaction, address, program_id }
```
raychu86: Add a check here that the Transaction is a deployment transaction and not an execution.

vm/package/deploy.rs
```diff
@@ -84,26 +87,26 @@ impl<'de, N: Network> Deserialize<'de> for DeployRequest<N> {
...
-    deployment: Deployment<N>,
+    transaction: Transaction<N>,
 }
 
 impl<N: Network> DeployResponse<N> {
     /// Initializes a new deploy response.
-    pub const fn new(deployment: Deployment<N>) -> Self {
-        Self { deployment }
+    pub const fn new(transaction: Transaction<N>) -> Self {
+        Self { transaction }
```
raychu86: Same as above, add a check here that the Transaction is a deployment transaction and not an execution.

raychu86 (changes requested):

vm/package/deploy.rs
ilitteri: Done! I'll do the same for #1140.

vm/package/deploy.rs
```diff
@@ -113,19 +128,25 @@ impl<'de, N: Network> Deserialize<'de> for DeployResponse<N> {
     fn deserialize<D: Deserializer<'de>>(deserializer: D) -> Result<Self, D::Error> {
```
howardwu, line 128: We should also check that the deserialized transaction here is a `Transaction::Deploy`

vm/package/deploy.rs
```diff
@@ -71,39 +78,47 @@ impl<'de, N: Network> Deserialize<'de> for DeployRequest<N> {
...
```
howardwu, line 78: We should also check that the deserialized transaction here is a `Transaction::Deploy`

vm/package/deploy.rs
```diff
@@ -158,23 +179,21 @@ impl<N: Network> Package<N> {
         // Initialize the RNG.
         let rng = &mut rand::thread_rng();
         // Compute the deployment.
-        let deployment = process.deploy::<A, _>(program, rng).unwrap();
+        let deployment = process.deploy::<A, _>(program, rng)?;
+        // Compute the additional fee.
+        let (_, additional_fee) = process.execute_additional_fee::<A, _>(private_key, credits, 1, rng)?;
```
howardwu: Can we pass in a `fee: u64` instead of hardcoding `1`?

vm/package/deploy.rs
```diff
@@ -14,31 +14,38 @@
...
 }
 
 impl<N: Network> DeployRequest<N> {
     /// Sends the request to the given endpoint.
-    pub fn new(deployment: Deployment<N>, address: Address<N>, program_id: ProgramID<N>) -> Self {
-        Self { deployment, address, program_id }
+    pub fn new(transaction: Transaction<N>, address: Address<N>, program_id: ProgramID<N>) -> Result<Self> {
+        ensure!(
+            matches!(transaction, Transaction::Deploy(_, _, _)),
+            "Cannot create a deploy request with an execution transaction"
```
howardwu:
```suggestion
            "Cannot create a deploy request with a non-deployment transaction"
```

vm/package/deploy.rs
```diff
@@ -71,39 +78,47 @@ impl<'de, N: Network> Deserialize<'de> for DeployRequest<N> {
...
 }
 
 impl<N: Network> DeployResponse<N> {
     /// Initializes a new deploy response.
-    pub const fn new(deployment: Deployment<N>) -> Self {
-        Self { deployment }
+    pub fn new(transaction: Transaction<N>) -> Result<Self> {
+        ensure!(
+            matches!(transaction, Transaction::Deploy(_, _, _)),
+            "Cannot create a deploy response with an execution transaction"
```
howardwu:
```suggestion
            "Cannot create a deploy response with a non-deployment transaction"
```

vm/package/deploy.rs
howardwu: ^We intend to support other transaction types in the future, hence the greater specificity.

vm/package/deploy.rs
howardwu: ^We intend to support other transaction types in the future, hence the greater specificity.

vm/package/deploy.rs
ilitteri: Sure! I'll do the same for #1140.

PR #1141, opened 2022-10-04
Update REST Endpoints

ilitteri:
### Updated
-  `GET /testnet3/records/all` is now `POST` and returns a list of records.
-  `GET /testnet3/records/unspent` is now `POST` and returns a list of records.
-  `GET /testnet3/records/spent` is now `POST` and returns a list of records.

### Created
- `POST /testnet3/ciphertexts/unspent` - Retrieves a vector of unspent record ciphertexts given a view key. 
- `GET /testnet3/record/ciphertext/{commitment}` - Retrieves the record ciphertext given its commitment. This endpoint is needed because the Aleo CLI should be able to retrieve the record/s associated with their commitments which can be inputs of the execute command.
- `POST /testnet3/transfer` - Creates a transfer from account A given its private key to account B given its address, for a given amount. This endpoint is needed to make transfers between accounts and to make a faucet of a sort.

#### `/testnet3/transfer` request example

```json
{
    "from": "APrivateKey1zkp6ka4UZu9JfMFDBqzKBffX7HQpsneKtzCTwnHb5GXVDpL",
    "to": "aleo1ttf46heagh0e50esnd0ju29xjka6jd9ccsq7hqg9n4wv57mx4uzsm4frgf",
    "amount": 1
}
```

## Related PRs/Issues

This PR addresses:

- AleoHQ/snarkVM#1056 
- AleoHQ/snarkOS#2010


rest/src/helpers/or_reject.rs
```diff
@@ -31,3 +31,10 @@ impl<T> OrReject<T> for anyhow::Result<T> {
         self.map_err(|e| reject::custom(RestError::Request(e.to_string())))
     }
 }
+
+impl<T> OrReject<T> for Option<T> {
+    /// Returns the result if it is successful, otherwise returns a rejection.
+    fn or_reject(self) -> Result<T, Rejection> {
```
ljedrz, line 37: maybe there could be a `&str` param here indicating the reason behind the rejection?

rest/src/routes.rs
```diff
@@ -267,8 +335,28 @@ impl<N: Network, B: BlockStorage<N>, P: ProgramStorage<N>> Server<N, B, P> {
...
+        Ok(reply::with_status(reply::json(&records), StatusCode::OK))
+    }
+
+    /// Returns the unspent records for the given view key.
+    async fn ciphertexts_unspent(
+        view_key: ViewKey<N>,
+        ledger: Arc<RwLock<Ledger<N, B, P>>>,
+    ) -> Result<impl Reply, Rejection> {
+        // Fetch the records using the view key.
+        let ledger_reader = ledger.read();
```
ljedrz: this read access should happen inline (like in the previous methods) otherwise this lock is held longer than necessary

rest/src/routes.rs
```diff
@@ -288,4 +376,60 @@ impl<N: Network, B: BlockStorage<N>, P: ProgramStorage<N>> Server<N, B, P> {
...
+    ) -> Result<impl Reply, Rejection> {
+        let from = query.from;
+        let to = query.to;
+        let amount = query.amount;
+        let view_key = ViewKey::try_from(from).or_reject()?;
+
+        // TODO: This solution does not broadcast the transaction just to
+        // ensure that the record retrieval, the transaction creation and
+        // addition to the mempool are atomic.
+        let mut ledger_writer = ledger.write();
```
ljedrz, line 400: unless no other reads should happen in the meantime, due to the amount of time it takes to perform the following operations, I'd procure **read** access here instead, drop the read guard after `Transaction::execute`, and acquire write access inline to alter the memory pool

ljedrz:
The `compiler` job failure seems transient, but the `clippy` one isn't.

ilitteri:
> The `compiler` job failure seems transient, but the `clippy` one isn't.

I saw that the clippy warning was added in another PR. Do you think that we should address it in this PR, or should be do that in another?

ljedrz:
> I saw that the clippy warning was added in another PR.

If it was caused by a different change, I think it should be a separate PR, so it could possibly be merged sooner.

howardwu:
@ilitteri could you take a look at the merge conflicts for this PR? We made a few bug + clippy fixes to the REST endpoints in PRs today.

rest/src/routes.rs
```diff
@@ -92,38 +125,46 @@ impl<N: Network, B: BlockStorage<N>, P: ProgramStorage<N>> Server<N, B, P> {
             .and(with(self.ledger.clone()))
             .and_then(Self::get_validators);
 
-        // GET /testnet3/statePath/{commitment}
-        let get_state_path = warp::get()
+        // POST /testnet3/statePath/{commitment}
+        let get_state_path = warp::post()
```
howardwu: This should be a GET endpoint as per recent changes

ilitteri: Yes you're right!

rest/src/helpers/or_reject.rs
ilitteri: This suggestion deserves a separate PR because it implies updating the implementation of `OrReject` (and all the places it's called).

rest/src/routes.rs
ilitteri: I see your point, but doing so exposes a race condition: if we drop the read guard after `Transaction::execute` there could possibly be another thread that gets the same record and executes the same transaction.

In addition, I don't think time could be won, the operations are going to be performed anyways (together or separately).

rest/src/routes.rs
ilitteri: Done!

PR #1151, opened 2022-10-05
Tighten checks on identifiers built from strings.

acoglio:
This aligns the checks to the nom parser for identifiers:
- Non-emtpy.
- Start with ASCII letter.
- Only contains ASCII letters, ASCII digits, and underscores.

Specifically, this tightens the checks for Unicode alphabetic and numeric to ASCII alphabetic and numeric. It also removes the ability for an underscore to start with underscore. The resulting code is made a little simpler than before.

Adds some tests to demonstrate the new restrictions, as well as a test for an empty identifier.


howardwu (approved):
LGTM

d0cd (approved):

PR #1154, opened 2022-10-06
Fix generation of random test strings.

acoglio:
The details are in the comments for the newly added code.

This fixes an observed failure of these tests.


circuit/types/string/src/helpers/to_bits.rs
```diff
@@ -58,7 +59,11 @@ mod tests {
 
         for i in 0..ITERATIONS {
             // Sample a random string. Take 1/4th to ensure we fit for all code points.
-            let expected: String = (0..(Circuit::MAX_STRING_BYTES - i) / 4).map(|_| rng.gen::<char>()).collect();
+            let expected: String = (0..(Circuit::MAX_STRING_BYTES - i) / 4)
```
ljedrz: this doesn't come from this PR, but is there a reason why we subtract `i` in this one case?

circuit/program/src/data/literal/from_bits.rs
```diff
@@ -132,7 +133,11 @@ mod tests {
             // String
             // Sample a random string. Take 1/4th to ensure we fit for all code points.
             let range = 0..rng.gen_range(0..Circuit::MAX_STRING_BYTES / 4);
-            let string: String = range.map(|_| rng.gen::<char>()).collect();
+            let string: String = range
+                .map(|_| rng.gen::<char>())
+                .map(adjust_unsafe_char)
+                .map(adjust_backslash_and_doublequote)
+                .collect();
```
ljedrz: since we call this quite a few times, we might want to turn this into a new function in `snarkvm_utilities` so that it can be used to generate random compliant strings when supplied with the desired length and an `Rng`

circuit/types/string/src/helpers/to_bits.rs
howardwu: It's to try different size strings.

circuit/program/src/data/literal/from_bits.rs
howardwu: I can handle this, thanks for flagging @ljedrz

howardwu: I've added this suggestion in my latest commit! :)

circuit/types/string/src/helpers/to_bits.rs
howardwu: I've added a boolean toggle for whether to randomly sample a fixed-length or variable-length string

howardwu (approved):

PR #1127, opened 2022-10-01
Implement new retargeting algorithm

raychu86:
## Motivation

This PR implements the new coinbase and proof retargeting algorithm as follows:

```
target = u64::MAX / HashToU64(cm)

delta = time_block_i - time_block_{i-1}
factor = (delta - duration_round) / # rounds per epoch
target_coinbase' = target_coinbase * 2^-factor
target_proof' = target_coinbase' >> 10
```

PR #1157, opened 2022-10-06
Add `CoinbaseSolution` accumulator point to block header

raychu86:
## Motivation

This PR adds the coinbase accumulator point to the block header. The genesis block and subsequent blocks with no coinbase solution will have a `Field::zero` as the coinbase accumulator point (this is enforced in consensus).

Built on #1127


howardwu (approved):

howardwu:
LGTM

PR #1159, opened 2022-10-07
Renamings

acoglio:
These commits carry out the following renamings at the user level (i.e. in parser and error messages):
- interfaces to structs
- functions to transitions
- closures to functions

As discussed, type names, variable names, comments, etc. will be renamed separately.

### DIFF
--- a/synthesizer/src/program/mod.rs
+++ b/synthesizer/src/program/mod.rs
@@ -224,8 +224,7 @@ transition fee:
     /// Returns the interface with the given name.
     pub fn get_interface(&self, name: &Identifier<N>) -> Result<Interface<N>> {
         // Attempt to retrieve the interface.
-        let interface =
-            self.interfaces.get(name).cloned().ok_or_else(|| anyhow!("Struct '{name}' is not defined."))?;
+        let interface = self.interfaces.get(name).cloned().ok_or_else(|| anyhow!("Struct '{name}' is not defined."))?;
         // Ensure the interface name matches.
         ensure!(interface.name() == name, "Expected interface '{name}', but found struct '{}'", interface.name());
         // Ensure the interface contains members.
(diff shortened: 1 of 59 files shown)


synthesizer/src/program/mod.rs
```diff
@@ -224,12 +224,11 @@ function fee:
     /// Returns the interface with the given name.
     pub fn get_interface(&self, name: &Identifier<N>) -> Result<Interface<N>> {
         // Attempt to retrieve the interface.
-        let interface =
-            self.interfaces.get(name).cloned().ok_or_else(|| anyhow!("Interface '{name}' is not defined."))?;
+        let interface = self.interfaces.get(name).cloned().ok_or_else(|| anyhow!("Struct '{name}' is not defined."))?;
         // Ensure the interface name matches.
-        ensure!(interface.name() == name, "Expected interface '{name}', but found interface '{}'", interface.name());
+        ensure!(interface.name() == name, "Expected interface '{name}', but found struct '{}'", interface.name());
```
d0cd: Rename `interface` to `struct`.

d0cd:
```suggestion
        ensure!(interface.name() == name, "Expected struct '{name}', but found struct '{}'", interface.name());
```

howardwu:
After reviewing the codebase further, I think it is safe to pursue the `interface` to `struct` change, however, because `Transition` is already a struct in our codebase, renaming `function` to `transition` would likely become easily confusing. Therefore, I suggest we hold off on changing `function` and `closure`.

acoglio:
@howardwu That's a valid point. I wonder if we should consider going full `TransitionFunction` in the code and `transition function` in the syntax, with just `Function` and `function` for helper functions. It was also my understanding that in the future we may need functions for transactions, in which case the scheme would naturally extend to `TransactionFunction` and `transaction function`. I believe that the Leo codebase actually has one struct `Function`, with a component that says what kind of function it is is (transition, helper, ...). @d0cd Is that the case? So that might be a possibility for Aleo instruction too.

collinc97:
Closing this PR since the first part of the renaming was merged in https://github.com/AleoHQ/snarkVM/pull/1246 and this PR is now pretty outdated.

> After reviewing the codebase further, I think it is safe to pursue the interface to struct change, however, because Transition is already a struct in our codebase, renaming function to transition would likely become easily confusing. Therefore, I suggest we hold off on changing function and closure.

PR #1171, opened 2022-10-11
Fix Arc unwraps in Process::execute

ljedrz:
I'd like to make sure that the `Arc` used for the `execution` in `Process::execute` isn't "leaky".


ljedrz:
Nope, looks fine, so we should apply the correct unwrap in order to avoid deep-cloning.

howardwu:
Could you provide a little more context on this change? Would love to learn :)

ljedrz:
Due to the way those values were being taken out of their `Arc`s, I thought that maybe `Arc::try_unwrap` was initially failing, and that's why it was not being used. Such failure would mean the existence of more than a single strong reference to the `Arc`, which would indicate a likely `Arc`-related memory leak (some strong reference not being dropped) happening somewhere in the execution.

ljedrz:
Rebased against `testnet3.2`.

It should also be noted that switching the deep clones to unwraps makes that operation much faster: it takes it down from ~`23µs` to ~`70ns`, which is a **99.7%** speedup, which can easily become meaningful when a lot of executions are carried out.

howardwu:
Oddly, I had to add `Debug` on the `Inclusion` struct, but not the `Execution` struct, despite them being similar.

howardwu (approved):
LGTM

PR #1182, opened 2022-10-17
Update temp dir handle on pertinent tests

ilitteri:
## Motivation

[`tempfile::tempdir`](https://docs.rs/tempfile/3.3.0/tempfile/fn.tempdir.html) (used in `test_helpers::sample_package`) automatically removes the temporary directory on `Drop`.

### DIFF
--- a/vm/package/build.rs
+++ b/vm/package/build.rs
@@ -290,7 +290,7 @@ mod tests {
     #[test]
     fn test_build() {
         // Samples a new package at a temporary directory.
-        let (directory, package) = crate::package::test_helpers::sample_package();
+        let (_, package) = crate::package::test_helpers::sample_package();
 
         // Ensure the build directory does *not* exist.
         assert!(!package.build_directory().exists());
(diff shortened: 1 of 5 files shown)


ljedrz:
There's an extra thing that needs addressing here. `tempdir` does remove the temporary directory _unless_ [into_path](https://docs.rs/tempfile/latest/tempfile/struct.TempDir.html#method.into_path) is called in the meantime. In order to take advantage of the deletion on `Drop`, we'll need to:
1. make `test_helpers::temp_dir` return [TempDir](https://docs.rs/tempfile/latest/tempfile/struct.TempDir.html) (i.e. not use `into_path` there)
2. make the users of `temp_dir` (`sample_package` and `sample_package_with_import`) return `TempDir` instead of `PathBuf` in the returned tuples, and only call [TempDir::path](https://docs.rs/tempfile/latest/tempfile/struct.TempDir.html#method.path) to obtain the path of the temporary directory
3. not bind to `_`, as it makes the bound object immediately go out of scope (this is how I noticed the issue in the first place: `_` was being used, which should have caused the related test cases to fail)

ljedrz (changes requested):
(as per my comment above)

PR #1187, opened 2022-10-18
Add record helpers for serial number and tag

howardwu:
## Motivation

Add record helpers for serial number and tag


raychu86 (approved):

Alegro24:
good projekt

PR #1190, opened 2022-10-19
Implement `get`, `get.or_init`, and `set` commands for `mapping` in `finalize`

d0cd:
This PR,
- implements `get`, `get.or_init`, `set` operations for `mappings`. 
- supports instruction execution in a finalize block.
- restricts `finalize` contexts from using records.

The proposed syntax is:
- `get accounts[r0] into r1;` - Note that this operation fails if the key does not exist in the mapping.
- `get.or_init accounts[r0] r1 into r2;` - Note that this operation initializes `accounts[r0]` with `r1` if the key does not exist in the mapping.
-  `set r0 into accounts[r1]`;

A mapping's keys and values are restricted to plaintext types (literals and structs).


vm/compiler/src/program/finalize/command/mod.rs
```diff
@@ -67,8 +79,12 @@ impl<N: Network> FromBytes for Command<N> {
             1 => Ok(Self::Instruction(Instruction::read_le(&mut reader)?)),
             // Read the increment.
             2 => Ok(Self::Increment(Increment::read_le(&mut reader)?)),
+            // Read the load.
+            3 => Ok(Self::Load(Load::read_le(&mut reader)?)),
+            // Read the store.
+            4 => Ok(Self::Store(Store::read_le(&mut reader)?)),
             // Invalid variant.
-            3.. => Err(error(format!("Invalid command variant: {}", variant))),
+            5.. => Err(error(format!("Invalid command variant: {}", variant))),
```
ljedrz:
```suggestion
            _ => Err(error(format!("Invalid command variant: {}", variant))),
```

d0cd: Isn't the range syntax more suited for catching missing cases?

ljedrz: in this case there is no difference, as we're dealing with a `u8`, and the numbers in the range `0..=4` have their own branches, so `_` is exactly the same as `5..`; I'd say that it's technically faster, but the difference is negligible

howardwu (approved):

vm/compiler/src/program/finalize/command/load.rs
```diff
@@ -0,0 +1,222 @@
...
+        // Load the operand as a plaintext.
+        let key = registers.load_plaintext(stack, &self.key)?;
+
+        // Retrieve the value from storage as a literal.
+        let value = match store.get_value(stack.program_id(), &self.mapping, &key)? {
+            Some(Value::Plaintext(plaintext)) => Value::Plaintext(plaintext),
+            Some(Value::Record(..)) => bail!("Cannot 'load' a 'record'"),
+            // If the key does not exist, use a default value.
+            // TODO: Infer type.
+            None => todo!(),
```
howardwu: When attempting to load a value from an uninitialized `key`, we should be bailing as the entry in the `mapping` does not exist.

synthesizer/src/process/stack/finalize_registers/load.rs
```diff
@@ -76,28 +47,17 @@ impl<N: Network> FinalizeRegisters<N> {
...
-                    // Retrieve the plaintext member from the path.
-                    Value::Plaintext(plaintext) => Value::Plaintext(plaintext.find(path)?),
-                    // Retrieve the record entry from the path.
-                    Value::Record(record) => match record.find(path)? {
-                        Entry::Constant(plaintext) | Entry::Public(plaintext) | Entry::Private(plaintext) => {
-                            Value::Plaintext(plaintext)
-                        }
-                    },
-                }
-            }
```
howardwu, line 90: Shouldn't we be keeping this logic unchanged?

synthesizer/src/process/stack/finalize_registers/load.rs
```diff
@@ -76,28 +47,17 @@ impl<N: Network> FinalizeRegisters<N> {
...
+            Register::Member(_, ref path) => stack_value.find(path)?,
         };
 
         // Retrieve the register type.
         match self.finalize_types.get_type(stack, register) {
             // Ensure the stack value matches the register type.
-            Ok(register_type) => stack.matches_register_type(&stack_value, &register_type)?,
```
howardwu, line 96: For safety, does it make sense to enforce the record case here?

howardwu:
```rs
        // Ensure that the stack value is a plaintext value.
        let stack_value = match stack_value {
            Value::Plaintext(plaintext) => plaintext,
            Value::Record(_) => bail!("Cannot load a record to a finalize register"),
        };
```

synthesizer/src/program/finalize/mod.rs
```diff
@@ -124,17 +125,34 @@ impl<N: Network> Finalize<N> {
...
+                ensure!(matches!(get.destination(), Register::Locator(..)), "Destination register must be a locator");
+            }
+            Command::GetOrInit(get_or_init) => {
+                // Ensure the destination register is a locator.
+                ensure!(
+                    matches!(get_or_init.destination(), Register::Locator(..)),
+                    "Destination register must be a locator"
+                );
+            }
+            _ => {}
```
howardwu: nit: can you make it explicit which cases are no-ops?

synthesizer/src/program/mapping/value/parse.rs
```diff
@@ -17,7 +17,7 @@
 use super::*;
 
 impl<N: Network> Parser for MapValue<N> {
-    /// Parses a string into the value statement of the form `value {name} as {register_type};`.
+    /// Parses a string into the value statement of the form `value {name} as {plaintext_type}.public;`.
```
howardwu, line 20: Some files say `{plaintext_type}.public` and most files say `{register_type}` now. We should be consistent.

synthesizer/src/program/instruction/mod.rs
```diff
@@ -359,7 +359,13 @@ impl<N: Network> Instruction<N> {
         stack: &Stack<N>,
         registers: &mut Registers<N, A>,
     ) -> Result<()> {
-        instruction!(self, |instruction| instruction.evaluate::<A>(stack, registers))
+        instruction!(self, |instruction| instruction.evaluate(stack, registers))
+    }
+
+    /// Evaluates the instruction in the context of the finalize block.
+    #[inline]
+    pub fn evaluate_finalize(&self, stack: &Stack<N>, registers: &mut FinalizeRegisters<N>) -> Result<()> {
```
howardwu: nit: rename to `finalize` and move all instances of this method impl to be after `fn execute`

synthesizer/src/program/instruction/operation/cast.rs
```diff
@@ -185,6 +185,55 @@ impl<N: Network> Cast<N> {
...
+            RegisterType::Plaintext(PlaintextType::Struct(struct_name)) => {
+                // Ensure the operands is not empty.
+                ensure!(!inputs.is_empty(), "Casting to a struct requires at least one operand");
+
+                // Retrieve the struct and ensure it is defined in the program.
+                let struct_ = stack.program().get_struct(&struct_name)?;
+
+                // Initialize the struct members.
+                let mut members = IndexMap::new();
+                for (member, (member_name, member_type)) in inputs.iter().zip_eq(struct_.members()) {
```
howardwu: We should check `inputs.len() == struct_.members().len()` to prevent a panic. The same vulnerability is present in the `evaluate` method above.

synthesizer/src/process/stack/finalize_registers/load.rs
```diff
@@ -76,28 +47,17 @@ impl<N: Network> FinalizeRegisters<N> {
             // If the register is a locator, then return the stack value.
             Register::Locator(..) => stack_value.clone(),
```
howardwu: Can we rename all instances of `stack_value` to `plaintext_value` for readability, now that the type has changed?

synthesizer/src/process/stack/finalize_registers/load.rs
d0cd: I don't think we need to since:
- L54 always produces a plaintext type.
- `stack_value` is a plaintext value.

Will rename `stack_value` to `plaintext_value` and `register_type` to `plaintext_type` for clarity.

synthesizer/src/process/stack/finalize_registers/load.rs
```diff
@@ -38,26 +38,26 @@ impl<N: Network> Load<N> for FinalizeRegisters<N> {
...
-            Register::Member(_, ref path) => stack_value.find(path)?,
+        let plaintext_value = match register {
+            // If the register is a locator, then return the plaintext value.
+            Register::Locator(..) => plaintext_value.clone(),
+            // If the register is a register member, then load the specific plaintext value.
+            Register::Member(_, ref path) => plaintext_value.find(path)?,
         };
 
-        // Retrieve the register type.
+        // Retrieve the type pf the register.
```
howardwu: nit: typo

synthesizer/src/process/stack/finalize_registers/store.rs
```diff
@@ -39,16 +39,16 @@ impl<N: Network> Store<N> for FinalizeRegisters<N> {
                 // Ensure the register does not already exist.
                 ensure!(!self.registers.contains_key(locator), "Cannot write to occupied register '{register}'");
 
-                // Ensure the register type is valid.
+                // Ensure the type of the is valid.
```
howardwu: nit: typo

synthesizer/src/process/stack/finalize_registers/load.rs
```diff
@@ -67,37 +38,26 @@ impl<N: Network> FinalizeRegisters<N> {
...
-            }
+        let plaintext_value = match register {
+            // If the register is a locator, then return the plaintext value.
+            Register::Locator(..) => plaintext_value.clone(),
+            // If the register is a register member, then load the specific plaintext value.
+            Register::Member(_, ref path) => plaintext_value.find(path)?,
         };
 
-        // Retrieve the register type.
+        // Retrieve the type pf the register.
```
howardwu:
```suggestion
        // Retrieve the type of the register.
```

synthesizer/src/process/stack/finalize_registers/store.rs
howardwu:
```suggestion
                // Ensure the type of the register is valid.
```

synthesizer/src/process/stack/finalize_registers/store.rs
```diff
@@ -16,26 +16,21 @@
...
     /// This method will halt if the given register is a register member.
     /// This method will halt if the given register is an input register.
     /// This method will halt if the register is already used.
     #[inline]
-    pub fn store(&mut self, stack: &Stack<N>, register: &Register<N>, stack_value: Value<N>) -> Result<()> {
+    fn store(&mut self, stack: &Stack<N>, register: &Register<N>, stack_value: Value<N>) -> Result<()> {
+        // Ensure that the stack value is a plaintext value.
+        let plaintext_value = match stack_value {
+            Value::Plaintext(plaintext) => plaintext,
+            Value::Record(_) => bail!("Cannot store a record to a finalize register"),
```
howardwu:
```suggestion
            Value::Record(_) => bail!("Cannot store a record in a finalize register"),
```

howardwu (approved):
LGTM

synthesizer/src/program/finalize/command/get.rs
```diff
@@ -0,0 +1,220 @@
...
+// along with the snarkVM library. If not, see <https://www.gnu.org/licenses/>.
+
+use crate::{Load as LoadTrait, Opcode, Operand, ProgramStorage, ProgramStore, Stack, Store};
+use console::{
+    network::prelude::*,
+    program::{Identifier, Register, Value},
+};
+
+/// A get command, e.g. `get accounts[r0] into r1;`.
+/// Gets the value stored at `operand` in `mapping` and stores the result in`destination`.
```
kpp, line 24: Fix comment "in\`destination\`"

synthesizer/src/program/finalize/command/get.rs
```diff
@@ -0,0 +1,220 @@
...
+            bail!("Mapping '{}/{}' does not exist in storage", stack.program_id(), self.mapping);
+        }
+
+        // Load the operand as a plaintext.
+        let key = registers.load_plaintext(stack, &self.key)?;
+
+        // Retrieve the value from storage as a literal.
+        let value = match store.get_value(stack.program_id(), &self.mapping, &key)? {
+            Some(Value::Plaintext(plaintext)) => Value::Plaintext(plaintext),
+            Some(Value::Record(..)) => bail!("Cannot 'get' a 'record'"),
```
kpp, line 87: Why is that?

PR #1191, opened 2022-10-20
Add header authorization checks for REST endpoints.

raychu86:
## Motivation

This PR adds a primitive header authorization check for the REST API endpoints related to records.

The authorized token is specified before initializing the `routes` and supports Bearer tokens and Basic base64 encoded `username:password` formats.


howardwu:
Noting that this approach while straightforward is insecure. This is sufficient for a temporary fix. After snarkOS resets, lets update this to a correct design.

howardwu (approved):

ljedrz (approved):

PR #1192, opened 2022-10-20
Fix test for integer division

d0cd:
This PR fixes the `panic_on_halt` condition for the integer division instruction.

### DIFF
--- a/synthesizer/src/program/instruction/operation/macros.rs
+++ b/synthesizer/src/program/instruction/operation/macros.rs
@@ -719,6 +719,10 @@ mod tests {
                             3.. => ($input_a::<CurrentNetwork>::rand(&mut rng), $input_b::<CurrentNetwork>::rand(&mut rng))
                         };
 
+                        // This flag is used to determine halting conditions.
+                        #[allow(deprecated)]
+                        let is_rhs_zero = (*b) == *$input_b::<CurrentNetwork>::zero();
+
                         // Initialize an indicator whether the operation should succeed or not.
                         #[allow(unused_mut)]
                         let mut should_succeed = true;
@@ -785,8 +789,11 @@ mod tests {
                                 let mut should_panic_on_halt = false;
                                 // If the operation is a shift operator, check if the mode of the RHS is a constant and if the shift amount exceeds the bitwidth.
                                 should_panic_on_halt |= is_shift_operator && shift_exceeds_bitwidth && mode_b.is_constant();
-                                // If the operation is a division operator, check if the mode of the RHS is a constant.
-                                should_panic_on_halt |= is_division_operator && mode_b.is_constant();
+                                // If the operation is a division operator, check if both operands are constant or if the RHS is a constant and zero.
+                                should_panic_on_halt |= is_division_operator && (
+                                    (mode_a.is_constant() && mode_b.is_constant()) ||
+                                    (mode_b.is_constant() && is_rhs_zero)
+                                );
 
                                 // If this iteration should succeed, ensure the evaluated and executed outputs match the expected output.
                                 if should_succeed {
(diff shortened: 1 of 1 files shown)


vm/compiler/src/program/instruction/operation/macros.rs
```diff
@@ -785,8 +789,11 @@ mod tests {
                                 let mut should_panic_on_halt = false;
                                 // If the operation is a shift operator, check if the mode of the RHS is a constant and if the shift amount exceeds the bitwidth.
                                 should_panic_on_halt |= is_shift_operator && shift_exceeds_bitwidth && mode_b.is_constant();
-                                // If the operation is a division operator, check if the mode of the RHS is a constant.
-                                should_panic_on_halt |= is_division_operator && mode_b.is_constant();
+                                // If the operation is a division operator, check if both operands are constant or if the RHS is a constant and zero.
+                                should_panic_on_halt |= is_division_operator && (
+                                    (mode_a.is_constant() && mode_b.is_constant()) ||
+                                    (mode_b.is_constant() && is_rhs_zero)
```
ljedrz:
```suggestion
                                    mode_b.is_constant() && (mode_a.is_constant() || is_rhs_zero)
```

ljedrz (commented):
I'm not too familiar with this particular logic, but what I can recommend is to extend the test in question with an additional division operation that falls under these extended conditions.

howardwu:
@d0cd can we add an explicit test case to check for the failure case that was detected?

vm/compiler/src/program/instruction/operation/macros.rs
```diff
@@ -710,14 +710,36 @@ mod tests {
 
                     // Check the operation on randomly-sampled values.
                     for i in 0..num_iterations {
+                        macro_rules! sample_value {
```
ljedrz: this macro should be defined outside of this loop

vm/compiler/src/program/instruction/operation/macros.rs
```diff
@@ -710,14 +710,36 @@ mod tests {
...
+                            (I8, I8) => { sample_value!(I128, I128) };
+                            (I16, I16) => { sample_value!(I128, I128) };
+                            (I32, I32) => { sample_value!(I128, I128) };
+                            (I64, I64) => { sample_value!(I128, I128) };
```
ljedrz: nit: these branches can probably be compacted to `(I8, I8) | (I16, I16) | ... => { sample_value!(I128, I128) };`

ljedrz (commented):
Left 2 suggestions.

vm/compiler/src/program/instruction/operation/macros.rs
d0cd: It doesn't seem to work.
The compiler errors with `error: no rules expected the token "|"`

vm/compiler/src/program/instruction/operation/macros.rs
d0cd: The macro uses `i` in the loop's scope.
I can move it outside and have the macro take `i` as an argument, it just didn't feel as clean to me.
Happy to go with whichever you feel is better.

howardwu (approved):
LGTM

vm/compiler/src/program/instruction/operation/macros.rs
ljedrz: ah true, I thought I saw a `match` there, but these are macro branches, so this doesn't apply :+1:

vm/compiler/src/program/instruction/operation/macros.rs
ljedrz: nah, it's ok; I usually declare macros outside of loops, but since this doesn't result in them being expanded any different and this one is only used inside of it, there's no issue

ljedrz (approved):
LGTM :+1:

PR #1194, opened 2022-10-20
Don't persist the test ledger across tests

ljedrz:
I've noticed that an error can be triggered if all the `compiler` tests are run together, and found the root cause to be the fact that some of the tests reuse a single ledger, leading to duplicate (and thus faulty) entries. The solution is to not reuse the ledger, which makes the tests run a bit longer, but not "cross-contaminate" one another.

In addition, the specific error (a duplicate transaction or its component) was erased by a higher-level error, which was also corrected.


ljedrz:
It doesn't seem like the runtime of the `compiler` job was affected to a meaningful degree :+1:.

howardwu (approved):
LGTM

PR #1210, opened 2022-10-24
Allow storage queries for items that are still pending

### LINKED ISSUE #1203: [Bug] Atomic speculation in `Map`
## 🐛 Bug Report

We need the ability to "atomically speculate".

The basic problem is this:
```
1. start_atomic
2. insert(k, 0)
3. let v = get(k) + 1
4. insert(k, v)
5. let v2 = get(k)
6. assert_eq!(v, v2)
```

Step 6 will fail because `v2 == 0 != v`

ljedrz:
I've considered alternative approaches to this, and in the end decided to propose 2 new methods for `MapRead`:
- a required `get_batched`, which works much like a `MapRead::get`, but on the pending atomic write batch
- a provided `get_speculative` method which first performs a plain `get`, and only falls back to `get_batched` if the former doesn't return a value

One of the existing related tests was extended with the new calls.

As for alternatives, the existing `get` could be extended with a `bool` parameter indicating whether the search should include the pending batch, but that would require a lot of changes to the existing calls. Making it the default is also not necessarily desirable, as it is a performance penalty.

Optional tweaks: the new `get_speculative` method could first check the pending batch and only then fall back to the "finalized" records; it's a performance consideration that depends on the expected use cases. The new method names could be changed too.

Closes https://github.com/AleoHQ/snarkVM/issues/1203.


howardwu:
This is great @ljedrz. I've been thinking about this problem some more.

`BlockStorage`, `ConsensusStorage`, `TransactionStorage` and `TransitionStorage` do not necessarily need this feature, however `ProgramStorage` always needs this feature.

Perhaps to minimize performance penalties for the former 4 storage, we could introduce two new traits (i.e. `SpeculativeWrite`/`SpeculativeRead` to mirror `MapWrite`/`MapRead`), and introduce these new traits only to the later trait (`ProgramStorage`) without giving it access to the original `MapWrite`/`MapRead`.

howardwu:
> Optional tweaks: the new get_speculative method could first check the pending batch and only then fall back to the "finalized" records; it's a performance consideration that depends on the expected use cases. The new method names could be changed too.

Would you mind reordering that `get_speculative` in this order? This is the desired functionality of that endpoint.

howardwu (approved):
Made necessary fixes to speculation. LGTM

PR #1211, opened 2022-10-24
A few performance tweaks for the slower tests

ljedrz:
This is a new instance of https://github.com/AleoHQ/snarkVM/pull/1199 rebased on top of `testnet3.2`.

This PR includes the following adjustments:
- a few allocations are avoided or reduced in size
- the number of iterations in serialization tests for `Transactions` is reduced, but at the same time the range of tested values is increased; there is little benefit from deserializing the same values in larger collections, as 2 of them (or even just one) deserialized within the context of a collection are deserialized in the same fashion as 100


howardwu:
Apologies, it appears recent merges caused merge conflicts here. Would you mind rebasing?

PR #1212, opened 2022-10-25
Improve the range of serialization tests for Transactions

ljedrz:
This is a new instance of https://github.com/AleoHQ/snarkVM/pull/1211.

This PR includes the following adjustment:
- the number of iterations in serialization tests for `Transactions` is reduced, but at the same time the range of tested values is increased; there is little benefit from deserializing the same values in larger collections, as 2 of them (or even just one) deserialized within the context of a collection are deserialized in the same fashion as 100

### DIFF
--- a/synthesizer/src/block/transactions/serialize.rs
+++ b/synthesizer/src/block/transactions/serialize.rs
@@ -72,35 +72,36 @@ mod tests {
 
     type CurrentNetwork = Testnet3;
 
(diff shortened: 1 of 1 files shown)


ljedrz:
@howardwu The PRs no longer seem to update once I force-push to my branch, nor can I reopen it. Were there recently any related changes to the repo settings?

PR #1218, opened 2022-10-26
Adds the last coinbase timestamp to the block metadata

howardwu:
## Motivation

- Adds the last coinbase timestamp to the block metadata


ljedrz (approved):
LGTM

PR #1219, opened 2022-10-26
Remove old parameter-related plumbing

ljedrz:
This used to be part of [an optimization](https://github.com/AleoHQ/snarkOS/pull/510) which is no longer in use, so it can be removed now.

### DIFF
--- a/curves/src/templates/macros.rs
+++ b/curves/src/templates/macros.rs
@@ -137,13 +137,8 @@ macro_rules! impl_sw_curve_serializer {
                     let (y, flags) = P::BaseField::deserialize_with_flags::<_, SWFlags>(&mut reader)?;
                     Affine::<P>::new(x, y, flags.is_infinity())
                 };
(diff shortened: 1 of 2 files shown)


howardwu:
Discussed offline, will green light after checking for performance regressions on object deserialization throughout snarkVM

ljedrz:
I ran the following test:
```
#[test]
fn genesis_s11n() {
    let genesis = Block::<CurrentNetwork>::from_bytes_le(CurrentNetwork::genesis_bytes()).unwrap();
    let iterations = 100u32;

    let now = std::time::Instant::now();
    for _ in 0..iterations {
        let serialized = genesis.to_bytes_le().unwrap();
        let deserialized = Block::<CurrentNetwork>::from_bytes_le(&serialized).unwrap();

        assert!(deserialized == genesis);
    }
    println!("avg. iteration time: {:?}", now.elapsed() / iterations);
}
```
and saw no difference in performance between `testnet3.2` and this PR; the same applies to a single deserialization (in case some objects were initialized just once during the runtime).

howardwu (approved):
LGTM

PR #1220, opened 2022-10-26
Optimize Record::is_owner

zvolin:
## Motivation

There seems to exist a possibility to reduce time cost of `Record::is_owner` by about 30%.

To verify it I've generated some records and tried to check `is_owner`'s result for a set of view keys. Results there are for verifying 4000 of records:

```
before:
ViewKey 1 is_owner - took 1470ms
ViewKey 2 is_owner - took 1460ms
ViewKey 3 is_owner - took 1426ms
ViewKey 4 is_owner - took 1459ms

after:
ViewKey 1 is_owner - took 1016ms
ViewKey 2 is_owner - took 983ms
ViewKey 3 is_owner - took 979ms
ViewKey 4 is_owner - took 995ms

delta: -31.7%
```

## Tldr

The optimization idea is to not create an `Address` from computed x coordinate

## Reasoning

I'm not even a crypto padawan so sorry if I confuse some terminology or I'm unclear, my reasoning about this change is as follows.

Address is a wrapper for an Affine which is a valid point on Environment's curve. Previously we computed the 'x' coordinate from the Record's ciphertext and the ViewKey and tried to create a valid Address from it. This basically tries to compute the point on the curve which all parameters are fixed (only 'x' and 'y' are varying). That means that 'y' is kinda 'purely' a function of x and for the same 'x's there always be the same 'y's.

In `is_owner` we're computing the `Address` to compare it to the provided one. We could instead just check if the computed 'x' coordinate is the same as the one from provided address. If the 'x' coordinates are equal then this implies that we will be able to compute the same `Address` because of the fact that this operation is just a function of 'x' so we can skip doing this ahead.

## Plan B

In case it is crucial to compute the `Affine` from the computed `x` coordinate we can still skip computation of `Address` itself. Doing so is costly because it includes afterwards checks whether the computed point lies on curve.

This was the first solution tried and it is included in this PR as the first of two commits.

The performance change of this attempt is like 2 or 3 % slower than the proposed solution.

## Test Plan

Was comparing the results from before this change to after this change on a few thousands of records for 30 random view keys and there was no difference but the computation time.

### DIFF
--- a/circuit/algorithms/src/bhp/commit.rs
+++ b/circuit/algorithms/src/bhp/commit.rs
@@ -80,7 +80,7 @@ mod tests {
 
     #[test]
     fn test_commit_constant() -> Result<()> {
-        check_commit::<32, 48>(Mode::Constant, 8200, 0, 0, 0)
+        check_commit::<32, 48>(Mode::Constant, 8250, 0, 0, 0)
     }
 
     #[test]
--- a/console/program/src/data/record/is_owner.rs
+++ b/console/program/src/data/record/is_owner.rs
@@ -29,12 +29,9 @@ impl<N: Network> Record<N, Ciphertext<N>> {
                 // Compute the 0th randomizer.
                 let randomizer = N::hash_many_psd8(&[N::encryption_domain(), record_view_key], 1);
                 // Decrypt the owner.
-                let x_coordinate = ciphertext[0] - randomizer[0];
-                if let Some((point1, point2)) = N::Affine::from_x_coordinate_variants(*x_coordinate) {
-                    point1 == ***address || point2 == ***address
-                } else {
-                    false
-                }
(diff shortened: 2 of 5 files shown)


Pratyush:
This is a good catch! The reason this works is because we check in the transaction SNARK that the ciphertext was constructed correctly, and hence it contains a valid `Address`. Validity for `Address`es means that the underlying group element is a member of the prime-order group, which in turn means that only one `(x, y)` pair constitutes a valid address. It would be good to add a comment to this effect in the updated code.

cc @howardwu

zvolin:
Could you form the comment to be added? I don't feel like the right person for this :)

Pratyush:
Done!

howardwu:
^Fixed a tiny flaky test count

howardwu (approved):
LGTM

PR #1225, opened 2022-10-29
Introduce `StatePath` verification to Transitions.

raychu86:
## Motivation

This PR introduces `inclusion_proof`s into `Transition`s - `executions` and `additional_fees`.

This proof is constructed by batching the state path verification for each input record in the transition; note that this proof must be included in each transaction if there is at least one input record. In the case that a `Transaction` does not have any input records, the `inclusion_proof` can be set to `None`.

PR #1227, opened 2022-10-31
Remove some allocations

ljedrz:
This PR contains a few self-contained changes which are aimed at reducing the number of allocations.


circuit/program/src/data/record/to_bits.rs
```diff
@@ -25,7 +25,7 @@ impl<A: Aleo> ToBits for Record<A, Plaintext<A>> {
         let data_bits_le = self
             .data
             .iter()
-            .flat_map(|(identifier, entry)| vec![identifier.to_bits_le(), entry.to_bits_le()])
+            .flat_map(|(identifier, entry)| [identifier.to_bits_le(), entry.to_bits_le()])
             .flatten()
```
zvolin, line 29: hard to say without benchmarks but wouldn't not introducing an array be faster?
```suggestion
            .flat_map(|(identifier, entry)| identifier.to_bits_le().chain(entry.to_bits_le()))
```

console/program/src/data/record/to_bits.rs
```diff
@@ -23,7 +23,7 @@ impl<N: Network> ToBits for Record<N, Plaintext<N>> {
         let data_bits_le = self
             .data
             .iter()
-            .flat_map(|(identifier, entry)| vec![identifier.to_bits_le(), entry.to_bits_le()])
+            .flat_map(|(identifier, entry)| [identifier.to_bits_le(), entry.to_bits_le()])
```
zvolin, line 26: saa

circuit/program/src/data/record/to_bits.rs
ljedrz: it's unlikely that there would be a difference in performance, but there is a more elegant way to do this; I'll add a new commit

howardwu:
These are great changes. Note that #1225 will cause some breakages here. We also have some tech debt on `function ID` incoming from #1225. We need to abstract it into a method in `console` and `circuit`.

howardwu (approved):
LGTM, really nice.

PR #1229, opened 2022-11-01
Unifies state path checks into 2 cases: execution and fee

howardwu:
## Motivation

Unifies state path checks into 2 cases: execution and fee.

This PR will be early-merged as it builds on-top of #1225


howardwu:
Early merging into #1225 for further review and testing

PR #1239, opened 2022-11-04
Adds `E::assert_neq`, update `Field::div`, update `Group::new`

howardwu:
## Motivation

- Adds `E::assert_neq` for more efficient comparison
- Updates `Field::div` to be more efficient
- Adds `Field::div_unchecked` for BHP hashing
- Updates `Group::new` to be checked by default

## Testing

All tests are updated to new constraint counts, and enforced to be passing.


circuit/types/group/src/helpers/from_xy_coordinates.rs
```diff
@@ -20,28 +20,8 @@ impl<E: Environment> Group<E> {
...
-            let second = (d * &x2) - &Field::one();
-            let third = (a * x2) - Field::one();
-
-            // Ensure y^2 * (dx^2 - 1) = (ax^2 - 1).
-            E::enforce(|| (first, second, third));
-        }
-
-        Self { x, y }
+        // Recover point from the `(x, y)` coordinates as a witness.
+        witness!(|x, y| console::Group::from_xy_coordinates(x, y))
```
howardwu, line 24: This function does the following:
1. It ejects the `mode` of `x` and `y`, and determines whether to allocate the recovered `point` as `Mode::Constant` or `Mode::Private`
2. It then calls `circuit::Group::new(mode, point)`

circuit/algorithms/src/bhp/hasher/hash_uncompressed.rs
```diff
@@ -155,9 +157,9 @@ impl<E: Environment, const NUM_WINDOWS: u8, const WINDOW_SIZE: u8> HashUncompres
                 match &sum {
                     Some((sum_x, sum_y)) => {
                         // Convert the accumulated sum into a point on the twisted Edwards curve.
-                        // TODO: If we know that sum_y is nonzero, then we can replace with `sum_x * sum_y.inverse()`.
-                        let edwards_x = sum_x / sum_y; // 5 constraints;
-                        Group::from_x_coordinate(edwards_x) // 3 constraints
+                        let edwards_x = sum_x * sum_y.inverse(); // 1 constraint (`sum_y` is never 0)
```
Pratyush: This should be `let edwards_x = sum_x.div_unchecked(sum_y)`

circuit/algorithms/src/bhp/hasher/hash_uncompressed.rs
```diff
@@ -155,9 +157,9 @@ impl<E: Environment, const NUM_WINDOWS: u8, const WINDOW_SIZE: u8> HashUncompres
...
+                        let edwards_y = (sum_x - &one) / (sum_x + &one); // 1 constraint (`sum_x` is never -1)
```
Pratyush: This should be `let edwards_y = (sum_x - &one).div_unchecked(sum_x + &one)`.

circuit/algorithms/src/bhp/hasher/hash_uncompressed.rs
Pratyush: Here `div_unchecked` omits the check that the numerator is zero.

Pratyush: i.e. it witnesses `result = sum_x.value() * sum_y.value().inverse()`, asserts that `result * sum_y == sum_x`, and returns `sum_x`

circuit/algorithms/Cargo.toml
```diff
@@ -15,6 +15,7 @@ optional = true
 [dependencies.snarkvm-circuit-types]
 path = "../types"
 version = "0.9.0"
+features = [ "unchecked" ]
```
Pratyush: Why this feature?

circuit/algorithms/src/elligator2/encode.rs
```diff
@@ -97,7 +97,7 @@ impl<E: Environment> Elligator2<E> {
         // Convert the Montgomery element (u, v) to the twisted Edwards element (x, y).
         let x = &u / v;
         let y = (&u - &one) / (u + &one);
-        let encoding = Group::from_xy_coordinates(x, y);
+        let encoding = Group::from_xy_coordinates_unchecked(x, y); // This is safe.
```
Pratyush: We have to additionally perform an on-curve check here.

```suggestion
         // This is safe because we perform the
         // on-curve and subgroup checks right below.
        let encoding = Group::from_xy_coordinates_unchecked(x, y);
```

circuit/types/group/src/helpers/from_unchecked.rs
```diff
@@ -0,0 +1,75 @@
...
+// You should have received a copy of the GNU General Public License
+// along with the snarkVM library. If not, see <https://www.gnu.org/licenses/>.
+
+use super::*;
+
+impl<E: Environment> Group<E> {
+    /// Initializes an affine group element from a given x- and y-coordinate field element.
+    /// Note: The resulting point is **not** enforced to be on the curve with constraints.
+    #[doc(hidden)]
+    #[cfg(any(test, feature = "unchecked"))]
```
Pratyush: Why are we feature-gating this? Isn't it always on (in our code base), since `algorithms` requires it?

circuit/types/group/src/helpers/mod.rs
```diff
@@ -17,6 +17,7 @@
 use super::*;
 
 pub mod from_bits;
+pub mod from_unchecked;
```
Pratyush: If we're going to feature gate, we should just put `#[cfg(feature = "unchecked")]` on this module directly.

Pratyush (changes requested):
Looks good, but there are a couple of efficiency and security changes that need to be made

circuit/types/group/src/helpers/from_unchecked.rs
howardwu: I've removed the "unchecked" feature flag.

circuit/algorithms/Cargo.toml
howardwu: I've removed the "unchecked" feature flag.

circuit/types/group/src/helpers/mod.rs
howardwu: I've removed the "unchecked" feature flag.

circuit/algorithms/src/elligator2/encode.rs
howardwu: I have switched this to the checked variant.

circuit/algorithms/src/bhp/hasher/hash_uncompressed.rs
howardwu: Switched both to use `div_unchecked`.

circuit/algorithms/src/bhp/hasher/hash_uncompressed.rs
howardwu: Switched both to use `div_unchecked`.

Pratyush (approved):

Pratyush (approved):

console/types/group/src/lib.rs
```diff
@@ -91,10 +91,6 @@ impl<E: Environment> Group<E> {
 
     /// Returns `self / COFACTOR`.
     pub fn div_by_cofactor(&self) -> Self {
-        // (For advanced users) The cofactor for this curve is `4`. Thus doubling is used to be performant.
-        // See unit tests below, which sanity check that this condition holds.
-        debug_assert!(E::Affine::cofactor().len() == 1 && E::Affine::cofactor()[0] == 4);
```
howardwu, line 96: This check was in both `mul_by_cofactor` and `div_by_cofactor`. However, only `mul_by_cofactor` is using this `double().double()` trick. This was a copypasta.

PR #1244, opened 2022-11-04
Enforce value exists in transition, adds `Transaction::fee` & `Transaction::find_transition` methods

howardwu:
## Motivation

- Enforce the input and output value exists in each transition
- Adds a `Transaction::fee` method
- Adds a `Transaction::find_transition` method
- Adds a maximum number of functions rule check


howardwu:
CircleCI instances are having issues, ran locally and tests passes.

raychu86 (approved):

PR #1249, opened 2022-11-07
perf: optimize MemoryMap::get_batched

zvolin:
This reduces the cost of MemoryMap::get_batched from O(n) to O(1).

The cost of whole `start_atomic` + insertions + removals + `finish_atomic` combined should stay similar to the one before the change


niklaslong (approved):

zvolin:
This will also improve finish_atomic times when multiple operations for the same key are in batch

howardwu (approved):
LGTM. Amazing work!

PR #1250, opened 2022-11-07
Add more checks for Universal SRS

Pratyush:
## Motivation

Forgot to push this commit in the last PR 
## Test Plan

(Write your test plan here)

## Related PRs

(Link your related PRs here)

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -5,6 +5,9 @@
 **storage_db
 **/*.prover.*
 **/*.verifier.*
-**universal.srs.*.*
 *.usrs
-!**universal.srs.*.metadata
+!**/powers-of-beta-15.usrs*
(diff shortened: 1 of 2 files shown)


.gitignore
```diff
@@ -6,4 +6,5 @@
 **/*.prover.*
 **/*.verifier.*
 **universal.srs.*.*
+*.usrs
```
howardwu, line 8: I see there are usrs files that are required, we should be pushing those on change and not ignoring them.

Pratyush: These are local to the repo, not to `.aleo/resources`, no?

howardwu: For 2^15 they're in the parameters folder

howardwu (approved):
LGTM

PR #1255, opened 2022-11-10
Cache Inclusion proving and verifying keys

raychu86:
## Motivation

This PR caches the `Inclusion` Proving key and Verifying key in the `Network` trait. This will reduce execution time significantly, because it won't have to deserialize the entire proving key for every Inclusion proof.

For reference, on a Threadripper 3960x, this change reduces execution time from ~6.3 seconds to ~5.1 seconds.


howardwu (approved):
LGTM

PR #1260, opened 2022-11-11
Adds finder methods on block, transaction, and transition

howardwu:
## Motivation

- Adds finder methods on block, transaction, and transition
- Adds `IntoIterator` on `Transactions` struct


raychu86 (approved):
LGTM

PR #1261, opened 2022-11-12
Fixes cast instruction to allow `MAX_DATA_ENTRIES` operands.

d0cd:
This PR,
- Fixes parsing for the cast instruction to allow `MAX_DATA_ENTRIES` operands. Note that the check is conditional on the `RegisterType` of the instruction.
- Adds parser tests for `Record` and `Struct`.


howardwu (approved):
LGTM, thanks for spotting this

PR #1263, opened 2022-11-12
Updates record `is_owner` for better performance

howardwu:
## Motivation

This PR removes the call to convert the address to affine format in each call to `is_owner` by passing the x-coordinate of the address in directly.

## Performance

I observed a 10% speedup after making this change.


howardwu:
Early merging this into #1262

PR #1280, opened 2022-11-30
Fix possible issue in integer bit serialization.

acoglio:
The call of `to_le` is a no-op on little-endian machines (i.e. pretty much all the machines we normally use), but it swaps bytes on big-endian machines. Thus, for instance, `0x01ff` would be swapped to `0xff01` by `to_le` on a big-endian machine, and then deserialized to the bits `1000 0000 1111 1111` by the loop that follows the `to_le` call, which would not be the correct deserialization, which should be `1111 1111 1000 0000`.

Note that the deserialization code does nothing to invert the `to_le`, so on a big-endian machine serializing and then deserializing would not be an identity.


howardwu (approved):

howardwu:
Closing as this change already landed in https://github.com/AleoHQ/snarkVM/pull/1564

PR #1281, opened 2022-11-30
fix use local beacon rest path error

Linuxpizi:
## Motivation

now, when i use local beacon rest service test to get records and transafer, first i create a client like

```rust
let client = Client::<N>::new("http://127.0.0.1:3030")?;
```

but the internal use **Url::parse** function will append **/** in suffix, it will be **http://127.0.0.1:3030/**.

when send a request to beacon rest service the url will become **http://127.0.0.1:3030//testnet3/latest/height** it is a invalid url

## Test Plan

- [x] use [origin public api](https://vm.aleo.org/api) test ok, all test done
- [x] use [local snarkos beacon rest](http://127.0.0.1:3030) test done

## Related PRs

(Link your related PRs here)


Linuxpizi:
#1284

raychu86:
feat/client is not longer used. This can be closed.

Feel free to reopen against the latest `staging` branch if the issue this PR addresses is still relevant.

PR #1285, opened 2022-12-01
Add `global_state_root` for transitions with no `InclusionProof`

raychu86:
## Motivation

This PR adds the `global_state_root` to executions even if they do not have an `InclusionProof` (input record). This is done by updating the `global_state_root` with in `Inclusion::prepare_execution` using the `Query`. 

## Test Plan

A test (`test_verify_deploy_and_execute`) has been added to enforce that the execution of a test program function with no input record will still have a unique state root.


howardwu (approved):
LGTM

PR #1296, opened 2022-12-05
Remove no-op loop.

acoglio:
Casting `false` to `u8` yields 0, which shifted by any amount is still 0, so the loop just adds 0 to `result`. Perhaps a remnant from old code.

### DIFF
--- a/utilities/src/bytes.rs
+++ b/utilities/src/bytes.rs
@@ -431,13 +431,6 @@ pub fn bytes_from_bits_le(bits: &[bool]) -> Vec<u8> {
             result += bit_value << i as u8;
         }
 
-        // Pad the bits if their number doesn't correspond to full bytes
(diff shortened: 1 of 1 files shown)


raychu86:
I believe this is correct and that this noop can be removed. 

This change originates from - https://github.com/AleoHQ/snarkVM/pull/433/files#diff-bdc9b21556b97252abcd19100cb9e616ad9879dfda1bd46d352c7cf799baa82eR47 as a way to reduce Vec allocations, but I don't believe it's doing anything, and uneeded in the `bits_le` case.

raychu86 (approved):

Pratyush:
Yeah this looks good. It would be even better to replace the loop with an iterator-based impl, but that can be done in a separate PR.

Pratyush (approved):

PR #1297, opened 2022-12-09
Adds metrics to stack executions

howardwu:
## Motivation

Adds metrics to stack executions. `Package::run` now has access to these metrics.

### DIFF
--- a/synthesizer/src/process/stack/mod.rs
+++ b/synthesizer/src/process/stack/mod.rs
@@ -93,13 +93,23 @@ use std::sync::Arc;
 
 pub type Assignments<N> = Arc<RwLock<Vec<circuit::Assignment<<N as Environment>::Field>>>>;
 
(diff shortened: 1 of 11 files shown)


collinc97 (approved):

synthesizer/src/process/stack/mod.rs
```diff
@@ -113,8 +123,9 @@ impl<N: Network> CallStack<N> {
         authorization: Authorization<N>,
         execution: Arc<RwLock<Execution<N>>>,
         inclusion: Arc<RwLock<Inclusion<N>>>,
+        metrics: Arc<RwLock<Vec<CallMetrics<N>>>>,
```
ljedrz, line 126: do we expect multiple simultaneous readers for `metrics`? if not, it would be a bit faster to use a `Mutex` here

PR #1316, opened 2023-01-09
Add support for spending deployment fees

raychu86:
## Motivation

This PR makes 2 primary changes to allow for records created by deployment fees to be spent. 

_Both of these solutions are temporary and will be addressed formally before Phase 3 is launched. (The formal solution will require resetting the network)_

1. Add iterator to fetch the transaction id from a transition id in the deployment store.
    - This is inefficient, as it requires iterating through all program deployments.
    - Formal solution: Add an additional map to the DeploymentStore
2. Update  `TransactionLeaf` for deployment transactions to be compatible with the `Inclusion` circuit.
    - The inclusion circuit will need to be updated to support both execution and deployment.
    - Formal solution: Update the inclusion circuit to support either variant of transactions.
    
## Test Plan

A test has been updated to test spending with a fee record included in a program deployment.

### DIFF
--- a/console/program/src/state_path/transaction_leaf/mod.rs
+++ b/console/program/src/state_path/transaction_leaf/mod.rs
@@ -34,10 +34,13 @@ pub struct TransactionLeaf<N: Network> {
 }
 
 impl<N: Network> TransactionLeaf<N> {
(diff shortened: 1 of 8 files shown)


howardwu (approved):
LGTM

PR #1330, opened 2023-01-24
Fixes `Field::to_bits_le` gadget.

d0cd:
This PR:
- Fixes the `Field::to_bits_le` gadget by adding a check that enforces that the resulting bits represent a number less than the base field modulus.
- Fixes the `Scalar::to_bits_le` gadget by adding a check that enforces that the resulting bits represent a number less than the scalar field modulus.
- Cleans up some related circuits.
- Updates constraint counts.

NOTE: CI fails because parameters need to be resampled.


circuit/types/scalar/src/helpers/to_bits.rs
```diff
@@ -45,7 +45,25 @@ impl<E: Environment> ToBits for &Scalar<E> {
...
+                        Boolean::constant(false),
+                        |rest_is_less, (this, that)| {
+                            if *this { that.bitand(&rest_is_less) } else { that.bitor(&rest_is_less) }
+                        },
+                    );
+
+                    // Ensure that the non-unique bit-representation is less than `ScalarField::MODULUS`.
+                    E::assert(is_less_than_modulus);
+                }
+                bits_le
```
Pratyush: Might be nice to share this logic with the base field modulus check, as we can pass in the modulus and `modulus_minus_one`, and the rest of the code is the same.

circuit/types/scalar/src/helpers/from_bits.rs
```diff
@@ -45,41 +45,38 @@ impl<E: Environment> FromBits for Scalar<E> {
...
-                accumulator += Field::from_boolean(bit) * &coefficient;
-                coefficient = coefficient.double();
-            }
-
-            // Return the scalar.
-            Scalar { field: accumulator, bits_le: OnceCell::with_value(bits_le) }
+            let bits_le = &bits_le[..size_in_bits];
+
+            // Compute `!((ScalarField::MODULUS - 1) < bits_le)`, which is equivalent to `((ScalarField::MODULUS - 1) >= bits_le)`, which is equivalent to `bits_le < ScalarField::MODULUS`.
+            let is_less_than_modulus = !modulus_minus_one.to_bits_le().iter().zip_eq(bits_le.iter()).fold(
```
Pratyush:
```suggestion
            let is_less_than_modulus = !modulus_minus_one.to_bits_le().iter().zip_eq(bits_le).fold(
```

circuit/types/scalar/src/compare.rs
```diff
@@ -29,7 +30,7 @@ impl<E: Environment> Compare<Scalar<E>> for Scalar<E> {
         else {
             // If all scalar field elements are less than (MODULUS - 1)/2 on the base field,
             // we can perform an optimized check for `is_less_than` by casting the scalars onto the base field.
-            debug_assert!(E::ScalarField::modulus() < E::BaseField::modulus_minus_one_div_two());
+            debug_assert!(E::ScalarField::modulus() <= E::BaseField::modulus_minus_one_div_two());
```
Pratyush: We should be able to leave this as is, no?

* `r = E::ScalarField::modulus() = 2111115437357092606062206234695386632838870926408408195193685246394721360383`, 
* `q = E::BaseField::modulus() = 8444461749428370424248824938781546531375899335154063827935233455917409239041`

Plugging these values into python tells me that that `(q - 1)/2 > r`

Pratyush (changes requested):
Thanks for the PR! Left a couple of comments

circuit/types/scalar/src/compare.rs
d0cd: Yes, the original check is sufficient for our instantiation of `r` and `q`.

It turns out that for the optimized implementation of `is_less_than`, we require all scalar field elements need to be `<= (BaseField::MODULUS - 1)/2`.
In that case, would the more "correct" check be 
```
debug_assert!(E::ScalarField::modulus() <= E::BaseField::modulus_minus_one_div_two() + 1);
```
Do we need to cover this edge case?

circuit/types/scalar/src/helpers/to_bits.rs
acoglio: I agree. In ACL2, we have defined this as its own gadget, parameterized by the sequence of constant bits (which we require to have at least one 0 and to have 1 in the most significant position).

circuit/types/field/src/helpers/to_bits.rs
```diff
@@ -33,44 +33,71 @@ impl<E: Environment> ToBits for Field<E> {
...
     fn to_bits_be(&self) -> Vec<Self::Boolean> {
         let mut bits_le = self.to_bits_le();
         bits_le.reverse();
         bits_le
     }
 }
 
+impl<E: Environment> Field<E> {
+    /// Outputs a non-unique little-endian bit representation of `self` *without* trailing zeros.
+    fn to_non_unique_bits_le(&self) -> Vec<Boolean<E>> {
```
acoglio, line 68: Factoring this into its own function is nice. It mirrors one of the gadgets we have defined in ACL2.

Pratyush (approved):

Pratyush:
I'm guessing CI is failing since the circuit changed.

circuit/types/scalar/src/compare.rs
d0cd: Updated check.

Pratyush (approved):

howardwu (approved):
LGTM

howardwu:
This PR needs resampling.

howardwu:
Resampled parameters

PR #1335, opened 2023-01-25
Fix the ordering of object serialization

### LINKED ISSUE #1334: [Bug] Deployment serialization breaks verification
## 🐛 Bug Report

Serializing and deserializing a `Deployment` transaction breaks the verification of the transaction because it breaks the ordering of the `verifying_keys`. This mismatch is then caught in the `VM::verify_deployment` call.

## Steps to Reproduce

1. Generate a deployment transcation.
2. Serialize the deployment transaction (i.e. to_string)
3. Deserialize the deployment transaction (i.e. serde_json::from_str)
4. Verification of the deserialized deployment will fail.

raychu86:
## Motivation

This PR addresses #1334, by enforcing that all objects maintain the same ordering after serialization and deserialization. This was necessary due to the requirements of Deployment transaction verification.

The fix was to add the `preserve_order` feature for `serde_json` dependencies.

## Test Plan

Added test to check that serialization/deserialization of deployments and executions don't break the transaction verification.

### DIFF
--- a/synthesizer/src/vm/verify.rs
+++ b/synthesizer/src/vm/verify.rs
@@ -242,6 +242,11 @@ mod tests {
 
         // Ensure the deployment is valid.
         assert!(vm.verify_deployment(&deployment));
+
+        // Ensure that deserialization doesn't break the transaction verification.
+        let serialized_deployment = deployment.to_string();
+        let deployment_transaction: Deployment<CurrentNetwork> = serde_json::from_str(&serialized_deployment).unwrap();
+        assert!(vm.verify_deployment(&deployment_transaction));
     }
 
     #[test]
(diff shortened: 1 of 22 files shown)


synthesizer/src/vm/verify.rs
```diff
@@ -260,6 +265,12 @@ mod tests {
                 assert!(Inclusion::verify_execution(&execution).is_ok());
                 // Verify the execution.
                 assert!(vm.verify_execution(&execution));
+
+                // Ensure that deserialization doesn't break the transaction verification.
+                let serialized_execution = execution.to_string();
+                let execution_transaction: Execution<CurrentNetwork> =
+                    serde_json::from_str(&serialized_execution).unwrap();
+                assert!(vm.verify_execution(&execution_transaction));
```
iamalwaysuncomfortable, line 273: The solution in this PR seems like it does the trick just fine, however I wonder if this test vector is robust enough to verify that ordering would be preserved for a program execution with a larger set of features represented. What are your thoughts on that?

iamalwaysuncomfortable (commented):
This solution looks like it works! However, left a quick comment about potentially testing this out (and potentially benching it?) with a more complex test vector.

howardwu (approved):
LGTM

PR #1338, opened 2023-01-27
Expose snarkvm synthesizer types to wasm

iamalwaysuncomfortable:
## Motivation

Types from the Synthesizer crate are necessary to enable functionality in the Aleo WebAssembly SDK for things like transaction creation, record selection, etc. This PR aims to expose the ability to access that functionality from the snarkvm-wasm crate.

## Changelog
* Blocking calls from `reqwests` were preventing the compilation of the `snarkvm-synthesizer` crate into `wasm` targets. Conditional compilation based upon a new `wasm` features in the `snarkvm-synthesizer` crate was added to address this.
* Added a `snarkvm-synthesizer` feature to the `snarkvm-wasm` crate which provides access to `synthesizer` types

## Test Plan
The types imported within the Aleo SDK will have a subset of features from synthesizer. That subset will be tested there.


synthesizer/src/process/stack/inclusion/mod.rs
```diff
@@ -95,36 +94,46 @@ impl<N: Network, B: BlockStorage<N>> Query<N, B> {
...
     pub fn current_state_root(&self) -> Result<N::StateRoot> {
         match self {
             Self::VM(block_store) => Ok(block_store.current_state_root()),
+            #[cfg(not(feature = "wasm"))]
             Self::REST(url) => match N::ID {
                 3 => Ok(Self::get_request(&format!("{url}/testnet3/latest/stateRoot"))?.json()?),
                 _ => bail!("Unsupported network ID in inclusion query"),
             },
+            #[cfg(feature = "wasm")]
+            _ => bail!("External API calls not supported from WASM"),
```
iamalwaysuncomfortable, line 117: Originally when working with the WASM feature enabled this will lead to a situation where circuits can't be built, but it's also more or less required currently to find an approach to build a working circuit in WASM

wasm/Cargo.toml
```diff
@@ -72,13 +79,15 @@ version = "0.3.33"
 
 [features]
 default = [ "full", "parallel" ]
-full = [ "console", "curves", "fields", "utilities" ]
+full = [ "console", "curves", "fields", "synthesizer", "utilities" ]
```
iamalwaysuncomfortable, line 82: Even if we can't yet build a circuit in wasm - exporting these types will immensely helpful to application developers

synthesizer/src/process/stack/inclusion/mod.rs
```diff
@@ -95,36 +94,46 @@ impl<N: Network, B: BlockStorage<N>> Query<N, B> {
...
                 3 => Ok(Self::get_request(&format!("{url}/testnet3/statePath/{commitment}"))?.json()?),
                 _ => bail!("Unsupported network ID in inclusion query"),
             },
+            #[cfg(feature = "wasm")]
+            _ => bail!("External API calls not supported from WASM"),
         }
     }
 
     /// Performs a GET request to the given URL.
+    #[cfg(not(feature = "wasm"))]
```
iamalwaysuncomfortable, line 136: The reqwest::blocking feature is what prevents wasm from compiling (wasm does not allow thread blocking) and thus the reason for this change. An associated macro was designed for wasm network calls to download the parameters but unfortunately it isn't viable due to the fact that the method called relies in javascript microtasks which make the required calls only after all the code specified in the macro runs, leading to an empty set of parameters (and the way its currently designed, small memory leaks). https://github.com/AleoHQ/snarkVM/blob/testnet3/parameters/src/macros.rs#L175

iamalwaysuncomfortable (commented):
Added clarifying comments to the PR

howardwu (approved):

PR #1339, opened 2023-01-30
Remove records from closure outputs

### LINKED ISSUE #1319: [Bug] Records constructed by closures may construct nonces incorrectly.
## 🐛 Bug Report

A program execution can't be properly generated if the program function outputs a record generated by an internal call to a closure and the (program and closure) register indexes are misaligned. 

If the closure and function have different register indexes for the record, then the record nonce will be generated differently, which will be caught in the circuit record encryption.

## Steps to Reproduce

Try to generated a transition with the following program: 

```
program token_with_cast.aleo;

record token:
    owner as address.private;
    gates as u64.private;
    token_amount as u64.private;

closure mint:
   input r0 as u64;
   input r1 as u64;
   cast self.caller r0 r1 into r2 as token.record;
   output r2 as token.record;

function compute:
    input r0 as token.record;
    call mint r0.gates r0.token_amount into r1;
    output r1 as token.record;",
```

(A full test is a bit too lengthy to post in this issue, as it requires simulating the deployment as well. Will provide on request if necessary)

Calling the `compute` function on the program above will fail to produces a valid transition because the record nonce generated in the closure will not match the record nonce generated for the function. The exact line is found in `circuit/program/src/data/record/encrypt.rs` in the `encrypt` call  - `A::assert_eq(&self.nonce,  A::g_scalar_multiply(randomizer));`

As you can see for the record in question, the register index in the closure is 2, while the register index in the function is 1. If we update the program to have the register index match, then the program execution is generated without error, only due to coincidence that the indexes were the same.

## Expected Behavior

A user should be able to create a proper transaction for the `compute` function call OR be notified that the `compute` function is not a valid function for whatever reason.

## Possible solutions

A couple possible solutions come to mind:

- Enforce that records can't be constructed in closures, and enforce their construction in functions only.
- Handle the index discrepancy by defaulting to the outside function index or the inside closure index.
- Enforce that records constructed by closures are not output directly without a proper cast inside the function.

raychu86:
## Motivation

This PR addresses #1319 by disallowing records from being closure outputs.

### DIFF
--- a/circuit/environment/src/environment.rs
+++ b/circuit/environment/src/environment.rs
@@ -24,10 +24,10 @@ pub trait Environment: 'static + Copy + Clone + fmt::Debug + fmt::Display + Eq +
     type Network: console::Network<Affine = Self::Affine, Field = Self::BaseField, Scalar = Self::ScalarField>;
 
     type Affine: AffineCurve<
-        BaseField = Self::BaseField,
-        ScalarField = Self::ScalarField,
-        Coordinates = (Self::BaseField, Self::BaseField),
-    >;
+            BaseField = Self::BaseField,
+            ScalarField = Self::ScalarField,
+            Coordinates = (Self::BaseField, Self::BaseField),
+        >;
     type BaseField: PrimeField + SquareRootField + Copy;
     type ScalarField: PrimeField<BigInteger = <Self::BaseField as PrimeField>::BigInteger> + Copy;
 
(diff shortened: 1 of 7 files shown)


ljedrz (commented):
I'm not familiar enough with this specific logic, but LGTM engineering-wise.

vicsn:
This solution seems best to me out of the three you mentioned in the issue, maybe you also want to [update the documentation](https://developer.aleo.org/aleo/language/#closure) (I wanted to but can't find any website repo)

howardwu (approved):
LGTM

PR #1340, opened 2023-01-30
Prototype for `flagged_division`.

d0cd:
This PR,
- implements a prototype for the `flagged_division` circuit and opcodes.
- adds supports in opcode macros for binary operations with two outputs.


circuit/types/field/src/div_flagged.rs
```diff
@@ -0,0 +1,170 @@
...
+
+impl<E: Environment> DivFlagged<Field<E>> for Field<E> {
+    type Output = (Field<E>, Boolean<E>);
+
+    /// Performs flagged division of two field elements.
+    #[doc(hidden)]
+    fn div_flagged(&self, other: &Field<E>) -> Self::Output {
+        match (self.is_constant(), other.is_constant()) {
+            // If `other` is a constant and zero, return the zero element and set the flag.
+            (_, true) if other.eject_value().is_zero() => (Field::zero(), Boolean::constant(true)),
```
iamalwaysuncomfortable: I take it this is being done because any nontrivial ring/field's zero element doesn't have a multiplicative inverse? Is it possible to document here with a sentence or to why this is implemented - namely what the consequences of allowing a finite field arithmetic violation like this to happen without a flag?

circuit/types/field/src/div_flagged.rs
```diff
@@ -0,0 +1,170 @@
...
+                    assert_eq!(result.eject_value(), console::Field::zero());
+                    assert!(flag.eject_value());
+                });
+                Circuit::reset();
+
+                // Check that `div` produces the correct result.
+                match mode_b.is_constant() {
+                    true => {
+                        // Check that `div` halts.
+                        let result = std::panic::catch_unwind(|| a.div(b));
```
iamalwaysuncomfortable, line 70: Why does that initially panic instead of !bailing into an Error monad? I know panicking is often the right decision in many contexts (like within SGX, etc.) - just wondering why specifically here.

iamalwaysuncomfortable (commented):
Couple introductory questions. It looks like if I'm not mistaken this PR is enforcing a lot of rules which govern finite field algebra.

circuit/types/field/src/div_flagged.rs
```diff
@@ -0,0 +1,170 @@
...
+
+use super::*;
+
+impl<E: Environment> DivFlagged<Field<E>> for Field<E> {
+    type Output = (Field<E>, Boolean<E>);
+
+    /// Performs flagged division of two field elements.
+    #[doc(hidden)]
+    fn div_flagged(&self, other: &Field<E>) -> Self::Output {
+        match (self.is_constant(), other.is_constant()) {
```
ljedrz:
```suggestion
        match other.is_constant() {
```
since the first one is never matched on, and there's no mutability involved

circuit/types/field/src/div_flagged.rs
ljedrz: I'm guessing it's due to halting; to change that one to not contain a panic would be a huge undertaking

synthesizer/src/program/instruction/operation/literals.rs
```diff
@@ -84,19 +105,32 @@ impl<N: Network, O: Operation<N, Literal<N>, LiteralType, NUM_OPERANDS>, const N
...
 
         // Compute the operation.
-        let output = O::evaluate(&inputs.try_into().map_err(|_| anyhow!("Failed to prepare operands in evaluate"))?)?;
+        let outputs = O::evaluate(&inputs.try_into().map_err(|_| anyhow!("Failed to prepare operands in evaluate"))?)?;
         // Compute the output type.
-        let output_type = RegisterType::Plaintext(PlaintextType::from(output.to_type()));
+        let output_types = outputs
+            .iter()
+            .map(|output| RegisterType::Plaintext(PlaintextType::from(output.to_type())))
+            .collect::<Vec<_>>();
```
ljedrz, line 113:
```suggestion
```
seems like we only need to iterate over them later, so this allocation is avoidable

synthesizer/src/program/instruction/operation/literals.rs
```diff
@@ -119,19 +161,32 @@ impl<N: Network, O: Operation<N, Literal<N>, LiteralType, NUM_OPERANDS>, const N
...
 
         // Compute the operation.
-        let output = O::execute(&inputs.try_into().map_err(|_| anyhow!("Failed to prepare operands in evaluate"))?)?;
+        let outputs = O::execute(&inputs.try_into().map_err(|_| anyhow!("Failed to prepare operands in evaluate"))?)?;
         // Compute the output type.
-        let output_type = RegisterType::Plaintext(PlaintextType::from(output.to_type()));
+        let output_types = outputs
+            .iter()
+            .map(|output| RegisterType::Plaintext(PlaintextType::from(output.to_type())))
+            .collect::<Vec<_>>();
```
ljedrz:
```suggestion
            .map(|output| RegisterType::Plaintext(PlaintextType::from(output.to_type())));
```
ditto

synthesizer/src/program/instruction/operation/literals.rs
ljedrz:
```suggestion
            .map(|output| RegisterType::Plaintext(PlaintextType::from(output.to_type())));
```
seems like we only need to iterate over them later, so this allocation is avoidable

synthesizer/src/program/instruction/operation/macros.rs
```diff
@@ -815,32 +1009,150 @@ mod tests {
...
+                                Some(vec![console::program::Literal::$output_a(expected_a), console::program::Literal::$output_b(expected_b)])
+                            }
+                            false => None
+                        };
+
+                        for mode_a in &[circuit::Mode::Constant, circuit::Mode::Public, circuit::Mode::Private] {
+                            for mode_b in &[circuit::Mode::Constant, circuit::Mode::Public, circuit::Mode::Private] {
+                                // Initialize the operands.
+                                let a = console::program::Literal::from_str(&format!("{a}"))?;
+                                let b = console::program::Literal::from_str(&format!("{b}"))?;
```
ljedrz:
```suggestion
                                let a = console::program::Literal::from_str(&a.to_string()))?;
                                let b = console::program::Literal::from_str(&b.to_string()))?;
```
nit

ljedrz (commented):
While I'm not knowledgeable enough about this code to approve or reject it, I left some comments.

circuit/types/integers/src/div_flagged.rs
```diff
@@ -0,0 +1,241 @@
...
+    /// Returns the quotient of `self` and `other`.
+    /// This method does not check that `other` is not zero.
+    /// This method uses the flag `divisor_is_zero` to conditionally check that the results of the division are well formed, as long as the divisor is not zero.
+    fn flagged_wrapped_division(&self, other: &Self, divisor_is_zero: &Boolean<E>) -> Self {
+        if I::is_signed() {
+            // Divide the absolute value of `self` and `other` in the base field.
+            let unsigned_dividend = self.abs_wrapped().cast_as_dual();
+            // Note that `unsigned_divisor` is zero iff `other` is zero.
+            let unsigned_divisor = other.abs_wrapped().cast_as_dual();
+            let unsigned_quotient = unsigned_dividend.flagged_wrapped_division(&unsigned_divisor, divisor_is_zero);
```
iamalwaysuncomfortable, line 66: Would this ever possibly lead to a stack overflow in some edge cases?

See: https://play.rust-lang.org/?version=stable&mode=debug&edition=2021&gist=35e106c9b123c7c29d2f3772d897c7dd

circuit/types/integers/src/div_flagged.rs
```diff
@@ -0,0 +1,241 @@
...
+
+            //  Note that quotient <= |console::Integer::MIN|, since the dividend <= |console::Integer::MIN| and 0 <= quotient <= dividend.
+            let signed_quotient = Self { bits_le: unsigned_quotient.bits_le, phantom: Default::default() };
+            let operands_same_sign = &self.msb().is_equal(other.msb());
+
+            // Note that this expression handles the wrapping case, where the dividend is `I::MIN` and the divisor is `-1` and the result should be `I::MIN`.
+            Self::ternary(operands_same_sign, &signed_quotient, &Self::zero().sub_wrapped(&signed_quotient))
```
iamalwaysuncomfortable, line 73: Should there be an `assert(divisor_is_zero)` for the signed case?

circuit/types/integers/src/div_flagged.rs
```diff
@@ -0,0 +1,241 @@
...
+            // If the product of two unsigned integers can fit in the base field, then we can perform an optimized division operation.
+            if 2 * I::BITS < E::BaseField::size_in_data_bits() as u64 {
+                let enforce_flagged_division = |first: &Self, second: &Self, quotient: &Self, remainder: &Self| {
+                    let first_equals_quotient_times_other_plus_remainder =
+                        first.to_field().is_equal(&(quotient.to_field() * second.to_field() + remainder.to_field()));
+
+                    // If the divisor is not zero, ensure that Euclidean division holds for these values in the base field.
+                    // This is equivalent to `!divisor_is_zero ==> (self == quotient * other + remainder)`.
+                    // Which is equivalent to `divisor_is_zero || (self == quotient * other + remainder)`.
+                    E::assert((divisor_is_zero).bitor(&first_equals_quotient_times_other_plus_remainder));
```
iamalwaysuncomfortable, line 84: Neat!

iamalwaysuncomfortable (commented):
Further comments

circuit/types/integers/src/div_flagged.rs
d0cd: We don't want to enforce that the divisor is zero, rather we want the `div_flagged` operation to be totally defined.

circuit/types/integers/src/div_flagged.rs
d0cd: It shouldn't, the call to `cast_as_dual` ensures that `unsigned_divided` and `unsigned_divisor` are unsigned integers. The unsigned case shouldn't recurse.

Did I miss something here?

circuit/types/field/src/div_flagged.rs
d0cd: We're doing this to ensure that the division operation is totally defined.
This will be used to create an instruction called `DivFlagged`, likely `div.f`, which is used to implement non-strict semantics in Leo.
In the standard field division circuit, a division be zero results in a failed execution.
I believe that it will be sufficient to document these new instructions in our opcode reference.

synthesizer/src/program/instruction/operation/macros.rs
```diff
@@ -164,6 +164,116 @@ macro_rules! operation {
             }
         }
     };
+    // K-ary operation with at least 1 output.
+    ($vis:vis struct $name:ident<$operator:path, $circuit_operator:path, $operate:ident, $opcode:tt, $num_inputs:tt, $num_outputs:tt> { $( ( $($input:ident),+ ) => ( $($output:ident),+ ) $( ($($condition:tt),+) )?, )+ }) => {
+        /// The implementation of the operation.
+        #[derive(Clone, PartialEq, Eq, Hash)]
```
ljedrz, line 170: are the equality and hash impls meaningful here? the `struct` has no contents after all, so hashing is a noop, and equality checks always return `true`

synthesizer/src/program/instruction/operation/macros.rs
```diff
@@ -164,6 +164,116 @@ macro_rules! operation {
...
+                type Operation = $name::<CurrentNetwork>;
+                // Execute the test cases for the operation.
+                $crate::test_execute!(Operator::$operate == Operation::execute { $( ( $($input),+ ) => ( $($output),+ ) $( ($($condition),+) )?, )+ });
+            }
+        }
+    };
+    // K-ary operation with question mark (?) with at least one output.
+    ($vis:vis struct $name:ident<$operator:path, $circuit_operator:path, $operate:ident?, $opcode:tt, $num_inputs:tt, $num_outputs:tt> { $( ( $($input:ident),+ ) => ( $($output:ident),+ ) $( ($($condition:tt),+) )?, )+ }) => {
+        /// The implementation of the binary operation.
+        #[derive(Clone, PartialEq, Eq, Hash)]
```
ljedrz, line 225: ditto

synthesizer/src/program/instruction/operation/macros.rs
```diff
@@ -38,35 +38,35 @@
...
 
-        impl<N: Network> $crate::Operation<N, console::program::Literal<N>, console::program::LiteralType, $num_inputs> for $name<N> {
+        impl<N: Network> $crate::Operation<N, console::program::Literal<N>, console::program::LiteralType, $num_inputs, 1> for $name<N> {
             /// The opcode of the operation.
             const OPCODE: $crate::Opcode = Opcode::Literal($opcode);
 
             /// Returns the result of evaluating the operation on the given inputs.
             #[inline]
-            fn evaluate(inputs: &[console::program::Literal<N>; $num_inputs]) -> Result<console::program::Literal<N>> {
+            fn evaluate(inputs: &[console::program::Literal<N>; $num_inputs]) -> Result<Vec<console::program::Literal<N>>> {
```
ljedrz, line 69: as for the `Result`s here, is an `Err` return value possible? just making sure, as I'm only seeing `Ok(<return_value>)`, suggesting that they might not be needed

synthesizer/src/program/instruction/operation/macros.rs
```diff
@@ -641,32 +835,32 @@ mod tests {
...
+                                let candidate_a = <$operation as $crate::Operation<_, _, _, 1, 1>>::evaluate(&[a])?;
                                 // Compute the executed output.
-                                let candidate_b = <$operation as $crate::Operation<_, _, _, 1>>::$execute::<CurrentAleo>(&[first])?;
+                                let candidate_b = <$operation as $crate::Operation<_, _, _, 1, 1>>::$execute::<CurrentAleo>(&[first])?;
 
                                 // Ensure the outputs match.
                                 assert_eq!(expected, Some(candidate_a));
                                 // Ensure the outputs match.
-                                assert_eq!(expected, Some(circuit::Eject::eject_value(&candidate_b)));
+                                assert_eq!(expected, Some(candidate_b.iter().map(|candidate| circuit::Eject::eject_value(candidate)).collect()));
```
ljedrz, line 845:
```suggestion
                                assert_eq!(expected, Some(candidate_b.iter().map(circuit::Eject::eject_value).collect()));
```
nit

synthesizer/src/program/instruction/operation/macros.rs
```diff
@@ -815,32 +1009,150 @@ mod tests {
...
+                                    let candidate_a = <$operation as $crate::Operation<_, _, _, 2, 1>>::evaluate(&[a, b])?;
                                     // Compute the executed output.
-                                    let candidate_b = <$operation as $crate::Operation<_, _, _, 2>>::$execute::<CurrentAleo>(&[first, second])?;
+                                    let candidate_b = <$operation as $crate::Operation<_, _, _, 2, 1>>::$execute::<CurrentAleo>(&[first, second])?;
 
                                     // Ensure the outputs match.
                                     assert_eq!(expected, Some(candidate_a));
                                     // Ensure the outputs match.
-                                    assert_eq!(expected, Some(circuit::Eject::eject_value(&candidate_b)));
+                                    assert_eq!(expected, Some(candidate_b.iter().map(|candidate| circuit::Eject::eject_value(candidate)).collect()));
```
ljedrz, line 1019:
```suggestion
                                    assert_eq!(expected, Some(candidate_b.iter().map(circuit::Eject::eject_value).collect()));
```
nit

synthesizer/src/program/instruction/operation/macros.rs
```diff
@@ -815,32 +1009,150 @@ mod tests {
...
+                                if should_succeed {
+                                    // Compute the evaluated output.
+                                    let candidate_a = <$operation as $crate::Operation<_, _, _, 2, 2>>::evaluate(&[a, b])?;
+                                    // Compute the executed output.
+                                    let candidate_b = <$operation as $crate::Operation<_, _, _, 2, 2>>::$execute::<CurrentAleo>(&[first, second])?;
+
+                                    // Ensure the outputs match.
+                                    assert_eq!(expected, Some(candidate_a));
+                                    // Ensure the outputs match.
+                                    assert_eq!(expected, Some(candidate_b.iter().map(|candidate| circuit::Eject::eject_value(candidate)).collect()));
```
ljedrz, line 1138:
```suggestion
                                    assert_eq!(expected, Some(candidate_b.iter().map(circuit::Eject::eject_value).collect()));
```
nit

synthesizer/src/program/instruction/operation/macros.rs
```diff
@@ -923,31 +1228,31 @@ mod tests {
...
+                                        let candidate_a = <$operation as $crate::Operation<_, _, _, 3, 1>>::evaluate(&[a, b, c])?;
                                         // Compute the executed output.
-                                        let candidate_b = <$operation as $crate::Operation<_, _, _, 3>>::$execute::<CurrentAleo>(&[first, second, third])?;
+                                        let candidate_b = <$operation as $crate::Operation<_, _, _, 3, 1>>::$execute::<CurrentAleo>(&[first, second, third])?;
 
                                         // Ensure the outputs match.
                                         assert_eq!(expected, Some(candidate_a));
                                         // Ensure the outputs match.
-                                        assert_eq!(expected, Some(circuit::Eject::eject_value(&candidate_b)));
+                                        assert_eq!(expected, Some(candidate_b.iter().map(|candidate| circuit::Eject::eject_value(candidate)).collect()));
```
ljedrz, line 1238:
```suggestion
                                        assert_eq!(expected, Some(candidate_b.iter().map(circuit::Eject::eject_value).collect()));
```
nit

ljedrz (commented):
Left a few comments and suggestions.

PR #1342, opened 2023-01-30
Remove allocations in EvaluationDomain::mul_polynomials_in_eval…

ljedrz:
These allocations amount to ~8% of all snarkOS allocations when running a prover, and they can all be avoided by taking the first param by value instead of by reference.


ljedrz:
The `fmt` issue is unrelated and seems to already be addressed in https://github.com/AleoHQ/snarkVM/pull/1339.

Pratyush (dismissed):

Pratyush (dismissed):

raychu86 (approved):

PR #1343, opened 2023-01-31
Reduce SonicPC::VeriferKey serialized size by avoiding prepared G2 elems

Pratyush:
## Motivation

Fixes large vk size in deploy transactions

## Test Plan

Added benchmarks to test the performance impact of preparing every time

<details>
<summary>Serialization</summary>

```
snark_vk_serialize/compressed
                        time:   [4.1759 µs 4.1847 µs 4.1987 µs]
                        change: [-98.787% -98.781% -98.772%] (p = 0.00 < 0.05)
                        Performance has improved.

snark_vk_serialize/uncompressed
                        time:   [4.1772 µs 4.1796 µs 4.1827 µs]
                        change: [-98.787% -98.785% -98.783%] (p = 0.00 < 0.05)
                        Performance has improved.

snark_vk_deserialize/compressed_checked
                        time:   [12.797 ms 13.400 ms 13.798 ms]
                        change: [-62.848% -52.206% -35.096%] (p = 0.00 < 0.05)
                        Performance has improved.

snark_vk_deserialize/compressed_unchecked
                        time:   [8.1131 ms 8.1277 ms 8.1519 ms]
                        change: [+52.637% +54.673% +56.332%] (p = 0.00 < 0.05)
                        Performance has regressed.

snark_vk_deserialize/uncompressed_checked
                        time:   [8.0383 ms 8.0922 ms 8.1540 ms]
                        change: [+50.400% +51.974% +53.400%] (p = 0.00 < 0.05)
                        Performance has regressed.

snark_vk_deserialize/uncompressed_unchecked
                        time:   [3.4660 ms 3.4822 ms 3.4982 ms]
                        change: [+483.19% +490.57% +499.35%] (p = 0.00 < 0.05)
                        Performance has regressed.
```
</details>

TL;DR: Our `ToBytes` and `FromBytes` implementations invoke `serialize_compressed` and `deserialize_compressed` (which is checked by default), respectively. For these code paths, the new approach improves time by 98% and ~50% respectively. So overall this approach is a win wrt both space and time complexity.

In the future we avoid all regressions by just caching the prepared versions of all the G2 elements (these are only roughly ~35 G2 elements that are ever used across all `vk`s, so we can just prepare and store these once and for all)


algorithms/src/polycommit/kzg10/data_structures.rs
```diff
@@ -206,6 +206,61 @@ pub struct VerifierKey<E: PairingEngine> {
...
+        Ok(())
+    }
+
+    fn serialized_size(&self, compress: Compress) -> usize {
+        self.g.serialized_size(compress)
+            + self.gamma_g.serialized_size(compress)
+            + self.h.serialized_size(compress)
+            + self.beta_h.serialized_size(compress)
+    }
+}
```
iamalwaysuncomfortable, line 224: Is it low hanging fruit to generate a proc macro for struct members at some point that would be similar to `#[serde(skip_serializing)]`. A proc macro  (something like `[canonical(skip_serializing)]`) that can apply to struct members would potentially allow the team to quickly create size savings like this.

Just a minor thought for the future.

iamalwaysuncomfortable (dismissed):

algorithms/src/polycommit/kzg10/data_structures.rs
Pratyush: Yes, that would be a great contribution and time-saver!

synthesizer/src/vm/verify.rs
```diff
@@ -239,6 +239,7 @@ mod tests {
 
         // Deploy the program.
         let deployment = vm.deploy(&program, rng).unwrap();
+        dbg!(deployment.to_bytes_le().unwrap().len());
```
ljedrz: this should probably be removed before merging?

algorithms/src/polycommit/kzg10/data_structures.rs
```diff
@@ -206,6 +206,61 @@ pub struct VerifierKey<E: PairingEngine> {
...
+impl<E: PairingEngine> Valid for VerifierKey<E> {
+    fn check(&self) -> Result<(), SerializationError> {
+        Valid::check(&self.g)?;
+        Valid::check(&self.gamma_g)?;
+        Valid::check(&self.h)?;
+        Valid::check(&self.beta_h)?;
+        Ok(())
+    }
+
+    fn batch_check<'a>(batch: impl Iterator<Item = &'a Self> + Send) -> Result<(), SerializationError>
```
ljedrz, line 251:
```suggestion
    fn batch_check(batch: Vec<Self>) -> Result<(), SerializationError>
```
if we're immediately going to allocate into a vector, it would be best to make it clear to the caller by expecting a vector to begin with

Pratyush: Unfortunately sometimes the input actually is an iterator (e.g. `BTreeMap::iter`)

synthesizer/src/vm/verify.rs
Pratyush: Yes, good catch!

```suggestion
```

algorithms/src/polycommit/kzg10/data_structures.rs
ljedrz: it could still be collected into a `Vec` beforehand; it's only to not conceal possibly large allocations from the caller, unless that's not a concern here

algorithms/src/polycommit/sonic_pc/data_structures.rs
```diff
@@ -387,6 +387,68 @@ pub struct VerifierKey<E: PairingEngine> {
...
+        compress: Compress,
+        validate: Validate,
+    ) -> Result<Self, SerializationError> {
+        let vk = CanonicalDeserialize::deserialize_with_mode(&mut reader, compress, validate)?;
+        let degree_bounds_and_neg_powers_of_h: Option<Vec<(usize, E::G2Affine)>> =
+            CanonicalDeserialize::deserialize_with_mode(&mut reader, compress, validate)?;
+        let supported_degree = CanonicalDeserialize::deserialize_with_mode(&mut reader, compress, validate)?;
+        let max_degree = CanonicalDeserialize::deserialize_with_mode(&mut reader, compress, validate)?;
+        let degree_bounds_and_prepared_neg_powers_of_h =
+            degree_bounds_and_neg_powers_of_h.as_ref().map(|v| v.iter().map(|(b, pow)| (*b, pow.prepare())).collect());
```
iamalwaysuncomfortable, line 419: This is saves a lot of space, but is worth benching the deserializations to explicitly measure the cost of recomputing the G2 preparation? If it ends up being particularly heavy, what effect might this have on verifying nodes or application implementors?

algorithms/src/polycommit/kzg10/data_structures.rs
```diff
@@ -206,6 +206,61 @@ pub struct VerifierKey<E: PairingEngine> {
...
+        compress: Compress,
+        validate: Validate,
+    ) -> Result<Self, SerializationError> {
+        let g = CanonicalDeserialize::deserialize_with_mode(&mut reader, compress, validate)?;
+        let gamma_g = CanonicalDeserialize::deserialize_with_mode(&mut reader, compress, validate)?;
+        let h: E::G2Affine = CanonicalDeserialize::deserialize_with_mode(&mut reader, compress, validate)?;
+        let beta_h: E::G2Affine = CanonicalDeserialize::deserialize_with_mode(&mut reader, compress, validate)?;
+        let prepared_h = h.prepare();
+        let prepared_beta_h = beta_h.prepare();
+        Ok(VerifierKey { g, gamma_g, h, beta_h, prepared_h, prepared_beta_h })
```
iamalwaysuncomfortable, line 238: Is it worth explicit benching the re-computation before committing to this change?

iamalwaysuncomfortable (commented):
Further comments on Computation vs. Serialized Size tradeoff.

algorithms/src/polycommit/kzg10/data_structures.rs
Pratyush: This would also require changing the trait definition, which is out-of-scope for this PR, but something we should consider in a future PR.

algorithms/src/polycommit/sonic_pc/data_structures.rs
Pratyush: Good idea. I've added benchmarks in the latest commit, and updated the PR description to report some preliminary numbers

algorithms/src/polycommit/kzg10/data_structures.rs
Pratyush: Same.

iamalwaysuncomfortable:
If I'm not mistaken the SRS is the same between program runs, so if you cache the pairing-prepared G2 elements - they're able to be used across all executions of the same program? 

Also I am not yet familiar enough to answer this so I'll ask, given the wide variety of platforms that may be operating a verifier key - what strategies might we use for caching of the prepared G2 elements?

Pratyush:
@raychu86 is there a blocker on merging this? I think it's good to go.

howardwu (approved):
LGTM

PR #1345, opened 2023-01-31
Fix fetching execution fees

raychu86:
## Motivation

This PR fixes the usage of `fee_map` for fetching the `global_state_root` and `inclusion_proof`. This will not break any existing database structure because the initial storage of the objects was correct.

## Test Plan

Tests have been added to check that inserting and getting executions with fees work properly.

### DIFF
--- a/synthesizer/src/vm/mod.rs
+++ b/synthesizer/src/vm/mod.rs
@@ -332,6 +332,65 @@ function compute:
             .clone()
     }
 
+    pub(crate) fn sample_execution_transaction_with_fee(rng: &mut TestRng) -> Transaction<CurrentNetwork> {
+        static INSTANCE: OnceCell<Transaction<CurrentNetwork>> = OnceCell::new();
+        INSTANCE
+            .get_or_init(|| {
+                // Initialize a new caller.
+                let caller_private_key = crate::vm::test_helpers::sample_genesis_private_key(rng);
+                let caller_view_key = ViewKey::try_from(&caller_private_key).unwrap();
+                let address = Address::try_from(&caller_private_key).unwrap();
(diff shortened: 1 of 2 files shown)


synthesizer/src/vm/mod.rs
```diff
@@ -332,6 +332,65 @@ function compute:
             .clone()
     }
 
+    pub(crate) fn sample_execution_transaction_with_fee(rng: &mut TestRng) -> Transaction<CurrentNetwork> {
```
iamalwaysuncomfortable, line 335: It's advisable potentially to make this module only available in test so it isn't available to production methods. Also potentially it should be moved to `vm/helpers/mod.rs` or a new file like maybe `vm/helpers/test_utils.rs`?

```suggestion
    #[cfg(test)]
    pub(crate) fn sample_execution_transaction_with_fee(rng: &mut TestRng) -> Transaction<CurrentNetwork> {
```

iamalwaysuncomfortable (approved):
Looks good modulo the organization of the test method

ljedrz (approved):
LGTM :+1:

synthesizer/src/vm/mod.rs
raychu86: This function exists within the `test_helpers` module in `pub(crate) mod test_helpers {`, so it shouldn't be accessible from outside of the test context

iamalwaysuncomfortable: Ah yes I see, my bad. Perhaps moving to helpers is still an apt choice? Either way not a blocker.

howardwu (approved):
LGTM

PR #1348, opened 2023-02-02
Optimize field-from-bits gadgets.

acoglio:
This is the part that checks whether the integer value denoted by a sequence of variable bits is less than or equal to the integer value denoted by a sequence of known constant bits (specifically, the bits whose value is the prime minus one). This less-than-or-equal-to check involves, among other things, checking that any variable bit in excess of the constant bits is zero.

Before this commit, this check on the excess bits was done by `or`ing the excess bits and then checking that the result is 0. However, this creates additional variables, in the R1CS constraints, to hold the intermediate results of `or`. This commit optimizes this check by simply enforcing constraints saying that each excess bit is 0. This is the same number of constraints as before, but uses fewer variables, in fact no extra variables.

This commit also adjusts the test expectations on the number of generated private variables, confirming that no additional variable is created.

In a bit more detail:

The gadget takes as input a bit sequence of any length, e.g.
```
w0 w1 w2 ... w252 w253 w254 w255      -- 256 bits, for example
```
and checks that the value denoted by those bits in integer arithmetic is less than or equal to `prime - 1`, whose decomposition into constant bits is
```
c0 c1 c2 ... c252      -- 253 bits, for the current prime
```

First, the gadget checks that any bits higher than the 253rd are 0, i.e. that `w253 = w254 = w255 = 0`.

It does so by `or`’ing them together and checking that the final result is 0. The resulting R1CS constraints are thus
```
(1 - w253) (1 - w254) = (1 - w256)      -- i.e. w256 = w253 OR w254
(1 - w256) (1 - w255) = (1 - w257)      -- i.e. w257 = w256 OR w255
(0) (1) = (w257)                        -- i.e. w257 = 0
```
which creates the extra variables `w256` and `w257`.

This commit instead makes the code generate the following constraints:
```
(w253) (1) = (0)      -- i.e. w253 = 0
(w254) (1) = (0)      -- i.e. w254 = 0
(w255) (1) = (0)      -- i.e. w255 = 0
```


d0cd (approved):
LGTM!

iamalwaysuncomfortable (approved):

Pratyush (approved):

iamalwaysuncomfortable:
Outside of the failing rustfmt, this can be merged

acoglio:
The failing fmt is actually in some unrelated code, but I can give a shot at fixing it, unless someone else prefers to do that.

raychu86 (approved):

PR #1358, opened 2023-02-06
Fixes `Integer::shr_wrapped`

d0cd:
This PR uses binary long division as a subroutine for `shr` operations over `u128` and `i128`.
It formerly used "witnessed" binary long division, which is not safe if `2 * SIZE_OF_INT >= SIZE_OF_FIELD`.

### DIFF
--- a/circuit/types/integers/src/div_wrapped.rs
+++ b/circuit/types/integers/src/div_wrapped.rs
@@ -127,18 +127,21 @@ impl<E: Environment, I: IntegerType> Metrics<dyn DivWrapped<Integer<E, I>, Outpu
     type Case = (Mode, Mode);
 
     fn count(case: &Self::Case) -> Count {
-        match I::is_signed() {
-            true => match (case.0, case.1) {
-                (Mode::Constant, Mode::Constant) => Count::is(2 * I::BITS, 0, 0, 0),
-                (Mode::Constant, _) | (_, Mode::Constant) => {
-                    Count::less_than(9 * I::BITS, 0, (8 * I::BITS) + 2, (8 * I::BITS) + 12)
+        match (case.0, case.1) {
+            (Mode::Constant, Mode::Constant) => Count::is(I::BITS, 0, 0, 0),
+            (Mode::Constant, _) | (_, Mode::Constant) => {
+                match (I::is_signed(), 2 * I::BITS < E::BaseField::size_in_data_bits() as u64) {
+                    (true, true) => Count::less_than(6 * I::BITS + 1, 0, (9 * I::BITS) + 7, (9 * I::BITS) + 13),
+                    (true, false) => Count::less_than(6 * I::BITS + 1, 0, 136 * I::BITS + 5, 137 * I::BITS + 8),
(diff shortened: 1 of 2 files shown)


circuit/types/integers/src/div_wrapped.rs
```diff
@@ -63,6 +63,7 @@ impl<E: Environment, I: IntegerType> Integer<E, I> {
     /// Divides `self` by `other`, via witnesses, returning the quotient and remainder.
     /// This method does not check that `other` is non-zero.
     /// This method should only be used when 2 * I::BITS < E::BaseField::size_in_data_bits().
+    /// This method assumes the `self` and `other` are unsigned integers.
```
iamalwaysuncomfortable, line 66: Should there be an explicit assertion here? If this condition (or the other 2) is violated - would a circuit still be able to be built? I'm wondering if that runs the risk of the proof being about a subtly different statement.

d0cd: In the current usage of `unsigned_division_via_witness`, these conditions are satisfied or left intentionally unchecked.
If the preconditions are unsatisfied, the circuit would still be able to built but may be incorrect.

The first condition is not explicitly enforced because sometimes we do not want to perform it (when we know that it is always satisfied, like in `shr_wrapped`)
The other two could be checked, though the third may require some modification of the traits.
And we could always add debug assertions throughout the codebase.

Broadly speaking, we would benefit from an audit of our circuit libraries and a following discussion on user-safety.

iamalwaysuncomfortable: Yeah that is all understandable. Trait modification is less than simple and often opens up its own vulnerabilities in cryptographic applications. I think given its not a public method it *should* be safe - especially when the developers using it are familiar with the assumptions behind it. Perhaps we just add it to an audit tracking issue.

iamalwaysuncomfortable (approved):

howardwu (approved):
LGTM

howardwu:
Waiting on test case for the untested case befor emerging.

circuit/types/integers/src/shr_wrapped.rs
```diff
@@ -75,30 +76,20 @@ impl<E: Environment, I: IntegerType, M: Magnitude> ShrWrapped<Integer<E, M>> for
...
-                    let remainder_field = remainder_unsigned.to_field();
-                    E::assert_eq(
-                        dividend_unsigned.to_field(),
-                        quotient_unsigned.to_field() * divisor_unsigned.to_field() + &remainder_field,
-                    );
-
-                    // Ensure that the remainder is less than the divisor.
-                    E::assert(remainder_unsigned.is_less_than(&divisor_unsigned));
+                    // Compute the quotient and remainder using wrapped, unsigned division.
+                    // Note that do not invoke `div_wrapped` since we need the quotient AND the remainder.
```
howardwu:
```suggestion
                    // Note that we do not invoke `div_wrapped` since we need the quotient AND the remainder.
```

PR #1364, opened 2023-02-10
Add `check_*` methods for transactions

raychu86:
## Motivation

This PR adds the `VM::check_transaction`, `VM::check_deployment`, and `VM::check_execution` functions, each of which returns a `Result`. The `VM::verify_transaction`, `VM::verify_deployment` and `VM::verify_execution` functions have been updated to call the corresponding `VM::check_*` function and logging the error if one exists.

The purpose of this distinction is to improve the clarify of transaction verification using explicit errors instead of just `warn!` logs.


howardwu (approved):
LGTM

PR #1367, opened 2023-02-15
Support proof generation from the wasm

evanmarshall:
1. Refactor Cargo.toml to support non parallel wasm build
2. Force cannonical (de)serialization of usize to 8 bytes regardless of target
3. Add synthesizer and console to wasm to enable transaction, program, & circuit types

## Motivation

The goal is to enable the wasm to generate proofs.

## Test Plan

The related Aleo PR contains tests and a UI through the website to test implementation

## Related PRs
https://github.com/AleoHQ/aleo/pull/472


evanmarshall:
A lot of the tests are failing because the tests require a `parallel` feature flag. 

Would the preferred solution be to fix the ci to run: `cargo test --features parallel` or to fix the tests to run serially (may be slow) or to refactor `parallel` -> `serial` so that everything is parallel by default?

synthesizer/src/block/transactions/mod.rs
```diff
@@ -63,88 +63,125 @@ impl<'a, N: Network> FromIterator<&'a Transaction<N>> for Transactions<N> {
...
-        match cfg!(feature = "parallel") {
-            true => self.par_values().any(|tx| tx.contains_transition(transition_id)),
-            false => self.values().any(|tx| tx.contains_transition(transition_id)),
-        }
+        self.par_values().any(|tx| tx.contains_transition(transition_id))
+    }
+
+    #[cfg(not(feature = "parallel"))]
+    pub fn contains_transition(&self, transition_id: &N::TransitionID) -> bool {
+        self.values().any(|tx| tx.contains_transition(transition_id))
```
Pratyush: Might be worth making a `cfg_values` macro like the existing `cfg_iter`, `cfg_iter_mut`, etc. macros

console/Cargo.toml
```diff
@@ -38,15 +38,17 @@ optional = true
 
 [features]
 default = [
-  "parallel",
```
iamalwaysuncomfortable, line 41: Re: for most of these - I'm not sure that it's desirable to remove `parallel` for the `default` feature set. A large number of the consumers of these libraries will be nodes in the network and it's desirable to have `parallel` in the `default` feature set for them. If there are any cases of nested dependencies, they wouldn't be able to access parallel features and this would likely negatively impact overall network health.

iamalwaysuncomfortable: In our discussion you explained how the Parallel feature wasn't working within most browsers. I remember we were speaking about issues you had found, can you explain what you were running into there? I think pointing out what the issues were will help point to a solution.

A few ideas on the above:
1. Has it been looked into possibly using webworker extensions to Rayon with something like https://github.com/GoogleChromeLabs/wasm-bindgen-rayon ?
2. Are the currently nested `SnarkVM` dependencies using `parallel`? If the situation is fortunate enough to only be using top level crate dependencies, it can be selected for manually when consuming the libraries.

console/algorithms/Cargo.toml
```diff
@@ -53,3 +53,7 @@ version = "1.0"
 [dev-dependencies.serde_json]
 version = "1.0"
 features = [ "preserve_order" ]
+
+[features]
+default = [ ]
+parallel = [ "snarkvm-fields/parallel" ]
```
iamalwaysuncomfortable, line 59: Commenting again to re-enforce the above comment, I doubt it'll be passable to remove parallel from the default features, but potentially there is another solution.

iamalwaysuncomfortable (commented):
Comments on "parallel"

circuit/Cargo.toml
```diff
@@ -50,3 +50,14 @@ version = "0.9.12"
 [dependencies.snarkvm-circuit-types]
 path = "./types"
 version = "0.9.12"
+
+[features]
+default = [ ]
```
Pratyush, line 55: I think enabling `parallel` by default makes sense, so that we benefit by default from the efficiency benefits of parallelism, but still preserve the ability to turn off `parallel`.

```suggestion
default = [ "parallel" ]
```

circuit/account/Cargo.toml
```diff
@@ -33,3 +33,7 @@ version = "1.0.69"
 [features]
 default = [ "enable_console" ]
```
Pratyush, line 34:
```suggestion
default = [ "enable_console", "parallel" ]
```

circuit/algorithms/Cargo.toml
```diff
@@ -34,3 +34,4 @@ version = "1.0.69"
 [features]
 default = [ "enable_console" ]
```
Pratyush, line 35:
```suggestion
default = [ "enable_console", "parallel" ]
```

circuit/environment/Cargo.toml
```diff
@@ -72,3 +72,7 @@ version = "0.4"
 
 [features]
 default = [ "snarkvm-curves/default", "snarkvm-r1cs/default" ]
```
Pratyush, line 74:
```suggestion
default = [ "snarkvm-curves/default", "snarkvm-r1cs/default", "parallel" ]
```

circuit/network/Cargo.toml
```diff
@@ -30,3 +30,7 @@ path = "../../console/types"
 [features]
 default = [ "enable_console" ]
```
Pratyush, line 31:
```suggestion
default = [ "enable_console", "parallel" ]
```

circuit/program/Cargo.toml
```diff
@@ -49,3 +49,8 @@ version = "0.8"
 [features]
 default = [ "enable_console" ]
```
Pratyush, line 50:
```suggestion
default = [ "enable_console", "parallel" ]
```

console/account/Cargo.toml
```diff
@@ -42,6 +42,7 @@ default = [
   "signature",
   "view_key"
```
Pratyush, line 43:
```suggestion
  "view_key",
  "parallel"
```

console/algorithms/Cargo.toml
```diff
@@ -53,3 +53,7 @@ version = "1.0"
 [dev-dependencies.serde_json]
 version = "1.0"
 features = [ "preserve_order" ]
+
+[features]
+default = [ ]
```
Pratyush, line 58:
```suggestion
default = [ "parallel" ]
```

console/collections/Cargo.toml
```diff
@@ -31,11 +31,12 @@ optional = true
...
 [dev-dependencies.snarkvm-console-network]
 path = "../network"
+features = [ "parallel" ]
 
 [dev-dependencies.criterion]
 version = "0.4.0"
 
 [features]
-default = [ "parallel" ]
+default = [ ]
```
Pratyush, line 40:
```suggestion
default = [ "parallel" ]
```

console/network/environment/Cargo.toml
```diff
@@ -41,3 +41,7 @@ default-features = false
 
 [dependencies.serde]
 version = "1.0"
+
+[features]
+default = [ ]
```
Pratyush, line 46:
```suggestion
default = [ "parallel" ]
```

console/Cargo.toml
evanmarshall: It's still on the default for snarkVM crate so as long as most consumers use that directly it should work. I agree though if parallel is the right default, we should switch the feature flag to from `parallel` to `serial`. It doesn't make sense to have a `parallel` feature flag if it's the default.

It probably makes sense to move most of the `aleo-wasm` logic into here too. The `aleo` if I understand correctly is mostly a compile from aleo instructions to byte code so the wasm stuff there doesn't really have anything to do with it.

As for why it wasn't working, it's very much related to `rayon` and the https://github.com/GoogleChromeLabs/wasm-bindgen-rayon could be the solution. I haven't investigated it yet but I'm starting to. I'll report back what I find.

Either way it will add extra complexity. Not all web browser support web workers so it'll probably require the build to produce both a serial and parallel version along with the code to detect features and switch between them. I don't think it'll get us out of the problem of cleanly separating serial and parallel versions of this codebase.

console/program/Cargo.toml
```diff
@@ -9,6 +9,10 @@ edition = "2021"
 [features]
 default = [ ]
```
Pratyush, line 10:
```suggestion
default = [ "parallel" ]
```

curves/Cargo.toml
```diff
@@ -70,3 +70,4 @@ rustc_version = "0.4"
 
 [features]
 default = [ "snarkvm-fields/default", "snarkvm-utilities/default" ]
```
Pratyush, line 72:
```suggestion
default = [ "snarkvm-fields/default", "snarkvm-utilities/default", "parallel" ]
```

fields/Cargo.toml
```diff
@@ -61,6 +61,6 @@ features = [ "derive" ]
 version = "1.0"
 
 [features]
-default = [ "parallel", "snarkvm-utilities/default" ]
+default = [ "snarkvm-utilities/default" ]
```
Pratyush, line 64:
```suggestion
default = [ "snarkvm-utilities/default", "parallel" ]
```

r1cs/Cargo.toml
```diff
@@ -57,3 +57,4 @@ version = "1.0"
 
 [features]
 default = [ ]
```
Pratyush, line 59:
```suggestion
default = [ "parallel" ]
```

synthesizer/Cargo.toml
```diff
@@ -35,16 +35,20 @@ harness = false
 required-features = [ "setup" ]
 
 [features]
-default = [ "parallel" ]
+default = [ ]
```
Pratyush, line 38:
```suggestion
default = [ "parallell" ]
```

synthesizer/src/coinbase_puzzle/mod.rs
```diff
@@ -209,6 +211,16 @@ impl<N: Network> CoinbasePuzzle<N> {
...
+            .fold(DensePolynomial::zero(), |mut accumulator, (mut prover_polynomial, challenge)| {
+                prover_polynomial *= challenge;
+                accumulator += &prover_polynomial;
+                accumulator
+            });
+        
+        #[cfg(feature = "parallel")]
         let accumulated_prover_polynomial = cfg_into_iter!(prover_polynomials)
             .zip_eq(challenges)
             .fold(DensePolynomial::zero, |mut accumulator, (mut prover_polynomial, challenge)| {
```
Pratyush, line 226:
```suggestion
        let zipped = cfg_into_iter!(prover_polynomials).zip_eq(challenges);
        let folder = |mut accumulator, (mut prover_polynomial, challenge)| {
            prover_polynomial *= challenge;
            accumulator += &prover_polynomial;
            accumulator
        };
        #[cfg(not(feature = "parallel"))]
        let accumulated_prover_polynomial = zipped.fold(DensePolynomial::zero(), folder);
        
        #[cfg(feature = "parallel")]
        let accumulated_prover_polynomial = zipped.fold(DensePolynomial::zero, folder);
```

utilities/src/serialize/impls.rs
```diff
@@ -148,7 +148,57 @@ impl_canonical_serialization_uint!(u8);
...
+    ) -> Result<Self, SerializationError> {
+        let mut bytes = [0u8; 8];
+        reader.read_exact(&mut bytes)?;
+        let u64_value = u64::from_le_bytes(bytes);
+        if u64_value <= usize::MAX as u64 {
+            let usize_value = u64_value as usize;
+            return Ok(usize_value)
+        } else {
+            return Err(SerializationError::IncompatibleTarget);
+        }
```
Pratyush, line 199:
```suggestion
        let u64_value = u64::deserialize_compressed(&mut reader)?;
        usize::try_from(u64_value).map_err(|_| SerializationError::IncompatibleTarget)
      
```

console/network/src/lib.rs
```diff
@@ -60,8 +60,8 @@ pub type FiatShamir<N> = PoseidonSponge<Fq<N>, 2, 1>;
 pub type FiatShamirParameters<N> = <FiatShamir<N> as AlgebraicSponge<Fq<N>, 2>>::Parameters;
 
 /// Helper types for the Marlin proving and verifying key.
-pub(crate) type MarlinProvingKey<N> = CircuitProvingKey<<N as Environment>::PairingCurve, MarlinHidingMode>;
-pub(crate) type MarlinVerifyingKey<N> = CircuitVerifyingKey<<N as Environment>::PairingCurve, MarlinHidingMode>;
+pub type MarlinProvingKey<N> = CircuitProvingKey<<N as Environment>::PairingCurve, MarlinHidingMode>;
+pub type MarlinVerifyingKey<N> = CircuitVerifyingKey<<N as Environment>::PairingCurve, MarlinHidingMode>;
```
iamalwaysuncomfortable, line 64: The types these alias are public, so I think this should be fine? Unless anyone else knows a reason not to make these public? If there is an issue, the `CircuitProvingKey` and `CircuitVerifyingKey` types are public and can be NewTyped in `wasm` pretty easily.

utilities/src/serialize/impls.rs
```diff
@@ -148,7 +148,57 @@ impl_canonical_serialization_uint!(u8);
 impl_canonical_serialization_uint!(u16);
 impl_canonical_serialization_uint!(u32);
 impl_canonical_serialization_uint!(u64);
-impl_canonical_serialization_uint!(usize);
+
+impl CanonicalSerialize for usize {
```
iamalwaysuncomfortable, line 152: I think likely instead of deleting the original serialization path, a `wasm` feature flag might be better suited since the `wasm` feature already exists in SnarkVM utilities.

i.e. something like

```suggestion
#[cfg(feature = "wasm")]
impl_canonical_serialization_uint!(usize);
#[cfg(not(feature = "wasm"))]
impl CanonicalSerialize for usize {
```

iamalwaysuncomfortable (commented):
Some further comments

console/Cargo.toml
iamalwaysuncomfortable: These are points to consider, will revisit this tomorrow. Thank you for proposing your thoughts on the tradeoffs.

circuit/Cargo.toml
evanmarshall: I took a stab at implementing your suggestions and I suggest we don't continue in that direction: https://github.com/demox-labs/snarkVM/commit/7ead9e29e2af932d53f5d65e93bbb0febfb907b8

This is what a serial feature would look like:
https://github.com/demox-labs/snarkVM/commit/ebf9c6234199e7ca5e849a762a72b2e145fbc793

utilities/src/serialize/impls.rs
evanmarshall: I think it makes sense to leave for all targets as it also applies to other 32-bit & 128-bit targets.

raychu86:
Removed in favor of #1388

PR #1369, opened 2023-02-15
Introduce the concept of a program `owner`

raychu86:
## Motivation

This PR introduces the concept of an `Owner` for deployment transactions. An `Owner` struct is simply the address of the program deployer and a signature of the deployment transaction id. 

Every deployment transaction will contain a new `Owner` struct, which can be referenced and verified with the corresponding deployment transaction id. 

In addition, a new mapping was created in the `DeploymentStore` called  `owner_map` that maps a `N::TransactionID` to an `Owner<N>`.


synthesizer/src/block/transaction/bytes.rs
```diff
@@ -38,8 +38,16 @@ impl<N: Network> FromBytes for Transaction<N> {
                 let deployment = Deployment::read_le(&mut reader)?;
                 // Read the fee.
                 let fee = Fee::read_le(&mut reader)?;
+                // Read the admin.
+                let admin = Admin::read_le(&mut reader)?;
+
+                // Ensure the admin is valid for the deployment transaction.
+                if !admin.verify(id) {
```
howardwu: I think we can skip this check here in favor of a check in the `VM::verify_deployment` logic. (since the `Transaction::deserialize` doesn't do this check and we might not want the perf hit)

synthesizer/src/store/transaction/deployment.rs
```diff
@@ -56,6 +56,9 @@ pub trait DeploymentStorage<N: Network>: Clone + Send + Sync {
     type FeeMap: for<'a> Map<'a, N::TransactionID, (N::TransitionID, N::StateRoot, Option<Proof<N>>)>;
     /// The mapping of `fee transition ID` to `transaction ID`.
     type ReverseFeeMap: for<'a> Map<'a, N::TransitionID, N::TransactionID>;
+    /// The mapping of `transaction ID` to `Admin`.
+    type AdminMap: for<'a> Map<'a, N::TransactionID, Admin<N>>;
```
howardwu: Discussed offline. We'll change this layout to `(program ID, edition) => Admin<N>`.

raychu86: Done, though we may want to consider program upgrades and edition bumps.

synthesizer/src/block/transaction/mod.rs
```diff
@@ -92,8 +96,13 @@ impl<N: Network> Transaction<N> {
...
         // Compute the fee.
         let (_, fee, _) = vm.execute_fee(private_key, credits, fee_in_gates, query, rng)?;
+
+        // Construct the owner.
+        let id = *Self::deployment_tree(&deployment, &fee)?.root();
+        let owner = Owner::new(private_key, id.into(), rng)?;
+
         // Initialize the transaction.
-        Self::from_deployment(deployment, fee)
+        Self::from_deployment(deployment, fee, owner)
```
iamalwaysuncomfortable: Is it necessary to have this computation here? This adds a re-computation of the deployment Merkle Tree when in this function, it's assured to be computed correctly. A few cycles can be saved by just creating the deployment directly.

```suggestion
        Ok(Self::Deploy(id.into(), Box::new(deployment), fee, owner))
```

synthesizer/src/block/transaction/mod.rs
```diff
@@ -92,8 +96,13 @@ impl<N: Network> Transaction<N> {
         let deployment = vm.deploy(program, rng)?;
         // Compute the fee.
         let (_, fee, _) = vm.execute_fee(private_key, credits, fee_in_gates, query, rng)?;
+
```
iamalwaysuncomfortable: In line with the suggestion below:

```suggestion
        ensure!(!deployment.program().functions().is_empty(), "Attempted to create an empty transaction deployment");

```

iamalwaysuncomfortable (approved):
One minor nit, but this looks good

synthesizer/src/block/transaction/mod.rs
howardwu: The purpose is to guarantee corrupt dbs immediately detect failure upon deserialization.

synthesizer/src/process/stack/deployment/owner/bytes.rs
```diff
@@ -0,0 +1,60 @@
...
+#[cfg(test)]
+mod tests {
+    use super::*;
+    use console::network::Testnet3;
+
+    type CurrentNetwork = Testnet3;
+
+    #[test]
+    fn test_bytes() -> Result<()> {
+        // Construct a new deployment.
```
howardwu:
```suggestion
        // Construct a new owner.
```

synthesizer/src/process/stack/deployment/owner/bytes.rs
```diff
@@ -0,0 +1,60 @@
...
+        // Read the signature.
+        let signature = Signature::read_le(&mut reader)?;
+
+        // Return the owner.
+        Ok(Self::from(address, signature))
+    }
+}
+
+impl<N: Network> ToBytes for Owner<N> {
+    /// Writes the deployment to a buffer.
```
howardwu:
```suggestion
    /// Writes the owner to a buffer.
```

synthesizer/src/process/stack/deployment/owner/bytes.rs
```diff
@@ -0,0 +1,60 @@
...
+
+        // Return the owner.
+        Ok(Self::from(address, signature))
+    }
+}
+
+impl<N: Network> ToBytes for Owner<N> {
+    /// Writes the owner to a buffer.
+    fn write_le<W: Write>(&self, mut writer: W) -> IoResult<()> {
+        // Write the owner address.
```
howardwu:
```suggestion
        // Write the address.
```

synthesizer/src/block/transaction/mod.rs
howardwu: Added suggested change in ff8049abc

howardwu (approved):
LGTM

PR #1383, opened 2023-03-04
Fix `Map` leniency

raychu86:
## Motivation

This PR handles the `Map` leniency described here - https://github.com/AleoHQ/snarkVM/issues/1353. This is done by removing `Borrow<F>` trait requirement from `Bech32ID` and `Bech32Object`.

This leniency allowed map functions to take in generic fields by input instead of their wrapper, Bech32ID. In other words, a map would consider `N::StateRoot`, `N::BlockHash`, `N::TransactionID`, and `N::TransitionID` all the same (as fields), instead of their wrapper types. This leads to possible errors in the map operations like the one fixed in https://github.com/AleoHQ/snarkVM/pull/1345.

Alternative solutions are welcome, as this adds (minor) restrictions to the updated traits.

### DIFF
--- a/console/network/src/helpers/id.rs
+++ b/console/network/src/helpers/id.rs
@@ -23,7 +23,6 @@ use std::borrow::Borrow;
 
 pub trait Bech32ID<F: FieldTrait>:
     From<F>
-    + Borrow<F>
     + Deref<Target = F>
     + Into<Vec<F>>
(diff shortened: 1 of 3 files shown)


howardwu (approved):

howardwu:
LGTM

PR #1388, opened 2023-03-07
Replace Parallel Feature with Serial Feature

evanmarshall:
## Motivation

The goal is to cleanly separate parallel & serial. Currently, the codebase heavily uses a parallel feature flag but only actually works when running in parallel which prevents compatibility with wasm32 targets for core features like proof generation.

We make parallel the clear default with no feature flags. We will eventually have to pipe back in `serial` for the wasm but this will make it much safer to do by avoiding the `default-features = false` & `default = [ "parallel" ]` anti-pattern there currently is. Feature flags should be additive.

## Benchmarks
All benchmarks done on the following hardware:
```
System Version: macOS 12.6.3 (21G419)
Chip: Apple M1 Max
Cores: 10 (8 performance and 2 efficiency)
Memory: 64 GB
```
[Testnet3 Baseline Benchmarks](https://github.com/iamalwaysuncomfortable/imagestorage/blob/main/testnet3_baseline_benchmarks.txt) 

[Regression Benchmark of Parallel Functionality](https://github.com/iamalwaysuncomfortable/imagestorage/blob/main/snarkvm_rayon_regression_test.txt) (No Benchmark Regressions Found)

[Benchmark With Serial Flags Applied](https://github.com/iamalwaysuncomfortable/imagestorage/blob/main/serial_flag_benchmarks.txt) (Significantly slower as expected - but mainly in the area of proving mathematics)

## Test Plan
* In order to avoid ci bloat, a ci test for the serial flag was not added but below, a result of testing all repos that have a serial flag are shown. This could optionally become a ci test.
[Serial Flag Tests](https://github.com/iamalwaysuncomfortable/imagestorage/blob/main/serial_flag_feature_test_suites.txt)

## Related PRs
This is the first step for proof generation in the wasm: 
https://github.com/AleoHQ/snarkVM/pull/1367


console/algorithms/src/bhp/hasher/mod.rs
```diff
@@ -81,7 +81,8 @@ impl<E: Environment, const NUM_WINDOWS: u8, const WINDOW_SIZE: u8> BHPHasher<E,
         }
 
         // Compute the bases lookup.
-        let bases_lookup = cfg_iter!(bases)
+        let bases_lookup = bases
```
evanmarshall, line 84: This was a subtle issue. I'm not sure if this can be run in parallel, but due to the previous mix of default & parallel features, this was previously always being run serially so I thought to make it explicit.

synthesizer/src/coinbase_puzzle/mod.rs
```diff
@@ -209,6 +211,17 @@ impl<N: Network> CoinbasePuzzle<N> {
         };
 
         // Construct the provers polynomial.
+        #[cfg(feature = "serial")]
+        let accumulated_prover_polynomial = cfg_into_iter!(prover_polynomials).zip_eq(challenges).fold(
```
evanmarshall: This was always being run in parallel so it would break if you tried to use this with default features so now we can handle both serial and parallel executions.

raychu86:
Looks fine! Would appreciate some benchmarks comparing (before the change, serial, and without serial) to ensure that this change doesn't affect performance.

Cargo.toml
```diff
@@ -82,8 +82,8 @@ default = [
   "console",
   "synthesizer",
   "cli",
-  "parallel",
-  "parameters"
+  "parameters",
+  "utilities"
```
iamalwaysuncomfortable, line 86: Why the extra addition of utilities?

synthesizer/src/block/transactions/mod.rs
```diff
@@ -64,87 +66,63 @@ impl<'a, N: Network> FromIterator<&'a Transaction<N>> for Transactions<N> {
...
     }
 
     /// Returns `true` if the transactions contains the given commitment.
     pub fn contains_commitment(&self, commitment: &Field<N>) -> bool {
-        match cfg!(feature = "parallel") {
-            true => self.par_values().any(|tx| tx.contains_commitment(commitment)),
-            false => self.values().any(|tx| tx.contains_commitment(commitment)),
+        match cfg!(feature = "serial") {
+            false => self.par_values().any(|tx| tx.contains_commitment(commitment)),
+            true => self.values().any(|tx| tx.contains_commitment(commitment)),
```
raychu86: Why not use the `cfg_values` macro for these instances?

fields/Cargo.toml
```diff
@@ -61,6 +59,5 @@ features = [ "derive" ]
 version = "1.0"
 
 [features]
-default = [ "parallel", "snarkvm-utilities/default" ]
-parallel = [ "rayon", "snarkvm-utilities/parallel" ]
+default = [ ]
```
iamalwaysuncomfortable: While the parallel flag here should can be removed `snarkvm-utilities/default` has default features unrelated to parallelization, so it should stay.

```suggestion
default = [ "snarkvm-utilities/default" ]
```

curves/Cargo.toml
```diff
@@ -69,4 +70,4 @@ version = "0.8"
 rustc_version = "0.4"
 
 [features]
-default = [ "snarkvm-fields/default", "snarkvm-utilities/default" ]
+default = [ ]
```
iamalwaysuncomfortable: In general these two features don't have much to do with parallelization after the removal of the `parallel` flags from them. The remaining features have to do with `no-std` environments and byte serialization so, in lieu of rigorous testing for side effects removal of these would cause, these should be left in place

```suggestion
default = [ "snarkvm-fields/default", "snarkvm-utilities/default" ]
```

algorithms/Cargo.toml
```diff
@@ -46,34 +46,28 @@ harness = false
 [dependencies.snarkvm-curves]
 path = "../curves"
 version = "0.9.13"
-default-features = false
 
 [dependencies.snarkvm-fields]
 path = "../fields"
 version = "0.9.13"
-default-features = false
```
iamalwaysuncomfortable: These two have default features unrelated to parallelization and deal with some intricate flags that could have side effects, so we should these two as is.

algorithms/Cargo.toml
```diff
@@ -46,34 +46,28 @@ harness = false
...
 
 [dependencies.snarkvm-parameters]
 path = "../parameters"
 version = "0.9.13"
 optional = true
-default-features = false
```
iamalwaysuncomfortable: Although the `parameter` `default` flags aren't parallelization related, they are seemingly not used anywhere in that crate so likely can be turned off.

algorithms/Cargo.toml
```diff
@@ -46,34 +46,28 @@ harness = false
...
 
 [dependencies.snarkvm-r1cs]
 path = "../r1cs"
 version = "0.9.13"
-default-features = false
 
 [dependencies.snarkvm-utilities]
 path = "../utilities"
 version = "0.9.13"
-default-features = false
```
iamalwaysuncomfortable: Default features should be left on for `utilities` for now as this crate is likely to have the most side effects currently if default is disabled. On more careful examination, this might be fine, but our focus is getting what's needed for the `wasm` to function in browser, so we can exclude disabling the `default` features here.

circuit/algorithms/Cargo.toml
```diff
@@ -19,11 +19,9 @@ version = "0.9.13"
 [dependencies.snarkvm-fields]
 path = "../../fields"
 version = "0.9.13"
-default-features = false
 
 [dev-dependencies.snarkvm-curves]
 path = "../../curves"
-default-features = false
```
iamalwaysuncomfortable: Same nit: Fields & curves should retain default features outside of parallel flags.

circuit/environment/Cargo.toml
```diff
@@ -19,27 +19,22 @@ version = "0.9.13"
 [dependencies.snarkvm-circuit-environment-witness]
 path = "./witness"
 version = "0.9.13"
-default-features = false
```
iamalwaysuncomfortable: There are no default features - this can be removed

circuit/environment/Cargo.toml
```diff
@@ -19,27 +19,22 @@ version = "0.9.13"
...
 
 [dependencies.snarkvm-curves]
 path = "../../curves"
 version = "0.9.13"
-default-features = false
 
 [dependencies.snarkvm-fields]
 path = "../../fields"
 version = "0.9.13"
-default-features = false
```
iamalwaysuncomfortable: Same nit: Fields & curves should retain default features outside of parallel flags.

console/account/Cargo.toml
```diff
@@ -18,7 +18,6 @@ version = "0.9.13"
 [dependencies.snarkvm-console-types]
 path = "../types"
 version = "0.9.13"
-default-features = false
```
iamalwaysuncomfortable: This flag is unnecessary here if features are explicitly specified as they are below.

circuit/types/group/Cargo.toml
```diff
@@ -30,7 +30,6 @@ version = "0.9.13"
 
 [dev-dependencies.snarkvm-utilities]
 path = "../../../utilities"
-default-features = false
```
iamalwaysuncomfortable: This should be reversed as mentioned above

circuit/environment/Cargo.toml
```diff
@@ -19,27 +19,22 @@ version = "0.9.13"
...
 
 [dependencies.snarkvm-r1cs]
 path = "../../r1cs"
 version = "0.9.13"
-default-features = false
 
 [dependencies.snarkvm-utilities]
 path = "../../utilities"
 version = "0.9.13"
-default-features = false
```
iamalwaysuncomfortable: This should be reversed as mentioned above

circuit/environment/Cargo.toml
```diff
@@ -58,7 +53,6 @@ version = "1.17.1"
 
 [dev-dependencies.snarkvm-algorithms]
 path = "../../algorithms"
-default-features = false
```
iamalwaysuncomfortable, line 61: This flag is unnecessary here because the features are specified below

circuit/environment/Cargo.toml
```diff
@@ -69,6 +63,3 @@ path = "../../console/types"
 
 [dev-dependencies.criterion]
 version = "0.4"
-
-[features]
-default = [ "snarkvm-curves/default", "snarkvm-r1cs/default" ]
```
iamalwaysuncomfortable, line 74: I'd suggest to leave the `snarkvm-curves/default` and remove the `snarkvm-r1cs/default` perhaps

console/algorithms/Cargo.toml
```diff
@@ -14,13 +14,11 @@ harness = false
 [dependencies.snarkvm-console-types]
 path = "../types"
 version = "0.9.13"
-default-features = false
```
iamalwaysuncomfortable: This can be removed because it's automatically not default because there are features specified below

console/algorithms/Cargo.toml
```diff
@@ -14,13 +14,11 @@ harness = false
...
 features = [ "field", "group", "integers", "scalar" ]
 
 [dependencies.snarkvm-fields]
 path = "../../fields"
 version = "0.9.13"
-default-features = false
```
iamalwaysuncomfortable: `fields` `default-features` should be left alone (except for the parallel flags)

console/algorithms/Cargo.toml
```diff
@@ -29,14 +27,16 @@ version = "0.9.13"
...
+version = "1"
+
 [dependencies.smallvec]
 version = "1.10"
 default-features = false
 features = [ "const_generics", "const_new" ]
 
 [dev-dependencies.snarkvm-curves]
 path = "../../curves"
-default-features = false
```
iamalwaysuncomfortable: Same nit as above

console/algorithms/src/bhp/hasher/mod.rs
iamalwaysuncomfortable: I'd actually suggest just leaving it alone. I understand you're going for code hygiene and this is good, but unless we're certain it can't be run in `parallel` it should be left as-is to minimize side effects.

console/collections/Cargo.toml
```diff
@@ -18,7 +18,6 @@ version = "0.9.13"
 [dependencies.snarkvm-console-types]
 path = "../types"
 version = "0.9.13"
-default-features = false
```
iamalwaysuncomfortable: Unnecessary because features are specified below, can be removed.

console/network/Cargo.toml
```diff
@@ -34,18 +33,15 @@ version = "0.9.13"
 [dependencies.snarkvm-console-types]
 path = "../types"
 version = "0.9.13"
-default-features = false
 features = [ "field", "group", "scalar" ]
```
iamalwaysuncomfortable: These two above changes can be removed as the specifying flags automatically removes the `default` feature

console/network/Cargo.toml
```diff
@@ -34,18 +33,15 @@ version = "0.9.13"
...
 
 [dependencies.snarkvm-curves]
 path = "../../curves"
 version = "0.9.13"
-default-features = false
 
 [dependencies.snarkvm-fields]
 path = "../../fields"
 version = "0.9.13"
-default-features = false
```
iamalwaysuncomfortable: Curves and fields should keep `default` features minus parallel flags as mentioned above

console/network/environment/Cargo.toml
```diff
@@ -9,12 +9,10 @@ edition = "2021"
 [dependencies.snarkvm-curves]
 path = "../../../curves"
 version = "0.9.13"
-default-features = false
 
 [dependencies.snarkvm-fields]
 path = "../../../fields"
 version = "0.9.13"
-default-features = false
```
iamalwaysuncomfortable: Same nit as above

curves/Cargo.toml
```diff
@@ -37,17 +37,18 @@ harness = false
 [dependencies.snarkvm-fields]
 path = "../fields"
 version = "0.9.13"
-default-features = false
 
 [dependencies.snarkvm-utilities]
 path = "../utilities"
 version = "0.9.13"
-default-features = false
```
iamalwaysuncomfortable: Same nit as above

curves/Cargo.toml
```diff
@@ -37,17 +37,18 @@ harness = false
...
 
 [dependencies.rand]
 version = "0.8"
 default-features = false
 
+[dependencies.rayon]
+version = "1"
+
```
iamalwaysuncomfortable, line 53: What does adding rayon here add to the crate?

fields/Cargo.toml
```diff
@@ -26,7 +26,6 @@ edition = "2021"
 [dependencies.snarkvm-utilities]
 path = "../utilities"
 version = "0.9.13"
-default-features = false
```
iamalwaysuncomfortable: Default should be kept for `utilities`

parameters/Cargo.toml
```diff
@@ -26,7 +26,6 @@ edition = "2021"
 [dependencies.snarkvm-curves]
 path = "../curves"
 version = "0.9.13"
-default-features = false
```
iamalwaysuncomfortable: Same nit as above

r1cs/Cargo.toml
```diff
@@ -26,12 +26,10 @@ edition = "2021"
 [dependencies.snarkvm-curves]
 path = "../curves"
 version = "0.9.13"
-default-features = false
 
 [dependencies.snarkvm-fields]
 path = "../fields"
 version = "0.9.13"
-default-features = false
```
iamalwaysuncomfortable: Same nit as above

synthesizer/Cargo.toml
```diff
@@ -72,7 +67,6 @@ version = "0.9.13"
 [dependencies.snarkvm-utilities]
 path = "../utilities"
 version = "0.9.13"
-default-features = false
```
iamalwaysuncomfortable: Same nit as above

synthesizer/src/block/transactions/mod.rs
iamalwaysuncomfortable: Yeah this tracks

synthesizer/src/coinbase_puzzle/mod.rs
iamalwaysuncomfortable: Nice

iamalwaysuncomfortable: Perhaps we should write some tests to verify this? I can help with this.

utilities/src/parallel.rs
```diff
@@ -32,26 +32,26 @@ impl<'a, T> ExecutionPool<'a, T> {
...
-    #[cfg(feature = "parallel")]
+    #[cfg(not(feature = "serial"))]
     pub fn add_job<F: 'a + FnOnce() -> T + Send>(&mut self, f: F) {
         self.jobs.push(Box::new(f));
     }
 
-    #[cfg(not(feature = "parallel"))]
-    pub fn add_job<F: 'a + FnOnce() -> T>(&mut self, f: F) {
+    #[cfg(feature = "serial")]
+    pub fn add_job<F: 'a + FnOnce() -> T + Send>(&mut self, f: F) {
```
iamalwaysuncomfortable, line 32: Why add the send marker trait? Not against it, just curious about the motivation.

iamalwaysuncomfortable (dismissed):
Really nice work. Thanks for putting in such hard work Evan, this is great.

The main nits I had were
* There were some crates that should retain the default features that have nothing to do with parallelization namely, `curves`, `fields` and `utilities`. They all ultimately have `utilities` crate defaults on them which affect `no-std` environments and byte serialization so we should tread very carefully with those and ultimately leave their `default-features` in place. 
* There are some places where `default-features = false` is redundant, I suggested removal above
* A couple of questions about adding `rayon` in the curves crate + `send` flag on one of the functions you introduced
* We should definitely write up some unit tests and benchmark tests to test these out. Overall as you said, these are largely cosmetic changes but nonetheless should be benchmarked to ensure we're not missing any side effects. I can work with you to do create these tests

console/algorithms/src/bhp/hasher/mod.rs
iamalwaysuncomfortable: On second look, agreed!

synthesizer/src/block/transactions/mod.rs
iamalwaysuncomfortable: Applied it here, but had to remove it from the lines below given that some of the `iter` methods between `rayon` and Rust's `iter` were not the same.

fields/Cargo.toml
```diff
@@ -61,6 +60,6 @@ features = [ "derive" ]
 version = "1.0"
 
 [features]
-default = [ "parallel", "snarkvm-utilities/default" ]
-parallel = [ "rayon", "snarkvm-utilities/parallel" ]
+default = [ "snarkvm-utilities/default" ]
 profiler = [ "aleo-std/profiler" ]
+serial = [ ]
```
raychu86: Minor nit: New at the end

synthesizer/src/block/transactions/mod.rs
```diff
@@ -64,87 +66,78 @@ impl<'a, N: Network> FromIterator<&'a Transaction<N>> for Transactions<N> {
...
     /// Returns the transaction with the given transition ID, if it exists.
     pub fn find_transaction_for_transition_id(&self, transition_id: &N::TransitionID) -> Option<&Transaction<N>> {
-        match cfg!(feature = "parallel") {
-            true => self.par_values().find_any(|tx| tx.contains_transition(transition_id)),
-            false => self.values().find(|tx| tx.contains_transition(transition_id)),
-        }
+        #[cfg(not(feature = "serial"))]
+        return self.par_values().find_any(|tx| tx.contains_transition(transition_id));
+        #[cfg(feature = "serial")]
+        self.values().find(|tx| tx.contains_transition(transition_id))
```
raychu86: I prefer the styling of the match case here instead of an explicit return case. Is there a reason this format was chosen/preferred?

raychu86 (changes requested):
Great changes and thanks for the benchmarks! Left a couple comments/nits.

One question about the benchmarks, regarding `snarkvm_serial_benches.txt` - 
I believe the benches here were run with rayon, so the naming seems a bit unclear. Also would be interesting to see what the benches look like when actually running with the `serial` feature.

synthesizer/src/block/transactions/mod.rs
iamalwaysuncomfortable: The match case didn't work when the serial feature was compiled because at the top of this file, rayon is conditionally disabled. Thus, when using the match case here, it doesn't disable the code in the match statement leading the compiler complaining about methods not being available. I'll give it another pass to beautify it.

iamalwaysuncomfortable:
> Great changes and thanks for the benchmarks! Left a couple comments/nits.
> 
> One question about the benchmarks, regarding `snarkvm_serial_benches.txt` - I believe the benches here were run with rayon, so the naming seems a bit unclear. Also would be interesting to see what the benches look like when actually running with the `serial` feature.

They were indeed run with rayon, currently running the benches with the serial flag and will change the above file to `snarkvm_rayon_regression_test.txt` and post the other bench as `snarkvm_serial_flag_test.txt`

synthesizer/src/block/transactions/mod.rs
iamalwaysuncomfortable: Refactored it into macros

circuit/environment/Cargo.toml
```diff
@@ -16,6 +16,10 @@ package = "snarkvm-console-network"
 path = "../../console/network"
 version = "0.9.14"
 
+[dev-dependencies.snarkvm-algorithms]
+path = "../../algorithms"
+features = [ "polycommit_full", "snark" ]
+
```
howardwu: nit: move `dev-dependencies` to the appropriate section (below)

curves/Cargo.toml
howardwu: What's the reason `rayon` is added?

synthesizer/src/coinbase_puzzle/mod.rs
```diff
@@ -209,6 +211,17 @@ impl<N: Network> CoinbasePuzzle<N> {
...
+            DensePolynomial::zero(),
+            |mut accumulator, (mut prover_polynomial, challenge)| {
+                prover_polynomial *= challenge;
+                accumulator += &prover_polynomial;
+                accumulator
+            },
+        );
+
+        #[cfg(not(feature = "serial"))]
```
howardwu: @Pratyush can you check if there is a more elegant/integrated approach here?

synthesizer/src/coinbase_puzzle/mod.rs
```diff
@@ -310,6 +323,15 @@ impl<N: Network> CoinbasePuzzle<N> {
...
         // Compute the accumulator evaluation.
+        #[cfg(feature = "serial")]
+        let mut accumulator_evaluation = cfg_iter!(prover_polynomials).zip_eq(&challenge_points).fold(
+            <N::PairingCurve as PairingEngine>::Fr::zero(),
+            |accumulator, (prover_polynomial, challenge_point)| {
+                accumulator + (prover_polynomial.evaluate(accumulator_point) * challenge_point)
+            },
+        );
+
+        #[cfg(not(feature = "serial"))]
```
howardwu: @Pratyush can you check if there is a more elegant/integrated approach here?

console/algorithms/src/bhp/hasher/mod.rs
howardwu: We should preserve `cfg_iter!` on this.

@Pratyush can you confirm?

synthesizer/src/coinbase_puzzle/mod.rs
iamalwaysuncomfortable: `elegant/integrated` in terms of style? Or in terms of reducing the number of ops? I can certainly take a try at this.

iamalwaysuncomfortable: Made a suggested change which reduces boilerplate

synthesizer/src/coinbase_puzzle/mod.rs
iamalwaysuncomfortable: Made a suggested change which reduces boilerplate

synthesizer/src/coinbase_puzzle/mod.rs
```diff
@@ -310,13 +317,18 @@ impl<N: Network> CoinbasePuzzle<N> {
...
+            |accumulator, (prover_polynomial, challenge_point)| {
                 accumulator + (prover_polynomial.evaluate(accumulator_point) * challenge_point)
-            })
-            .sum();
-        accumulator_evaluation *= &epoch_challenge.epoch_polynomial().evaluate(accumulator_point);
+            },
+        );
+        #[cfg(not(feature = "serial"))]
+        let evaluation = evaluation.sum::<<<N as console::prelude::Environment>::PairingCurve as PairingEngine>::Fr>();
+        let accumulator_evaluation = evaluation * epoch_challenge.epoch_polynomial().evaluate(accumulator_point);
```
raychu86: A `map` + `sum` would likely be cleaner than `fold` due to differences in `rayon` and `core::iter` APIs.

Example:
```
            let mut accumulator_evaluation = cfg_iter!(prover_polynomials)
                .zip(&challenge_points)
                .map(|(prover_polynomial, challenge_point)| {
                    prover_polynomial.evaluate(accumulator_point) * challenge_point
                })
                .sum::<<<N as console::prelude::Environment>::PairingCurve as PairingEngine>::Fr>();
            accumulator_evaluation *= &epoch_challenge.epoch_polynomial().evaluate(accumulator_point);

File: synthesizer/src/coinbase_puzzle/mod.rs [hunk 38]
```diff
@@ -208,15 +210,20 @@ impl<N: Network> CoinbasePuzzle<N> {
...
+            |mut accumulator, (mut prover_polynomial, challenge)| {
                 prover_polynomial *= challenge;
                 accumulator += &prover_polynomial;
                 accumulator
-            })
-            .sum::<DensePolynomial<_>>();
+            },
+        );
+        #[cfg(not(feature = "serial"))]
+        let accumulated_prover_polynomial = accumulated_prover_polynomial.sum::<DensePolynomial<_>>();
```
raychu86: Ditto w/ below. `map` + `sum` would likely be cleaner.

curves/Cargo.toml
iamalwaysuncomfortable: I've spotted it, if you look below there's some conditionally disabled code that imports rayon. I think everything compiled because that feature perhaps was simply never used. @evanmarshall Caught that bug here.

synthesizer/src/coinbase_puzzle/mod.rs
iamalwaysuncomfortable: Made this change + several others to cleanup the flag code duplication

synthesizer/src/coinbase_puzzle/mod.rs
iamalwaysuncomfortable: Done!

console/algorithms/src/bhp/hasher/mod.rs
howardwu: We should preserve cfg_iter! on this.

utilities/src/parallel.rs
ljedrz, line 32: the signatures for both these methods seem to be identical now, why not collapse them into a single one without a `cfg` attrib?

utilities/src/parallel.rs
```diff
@@ -169,12 +169,68 @@ macro_rules! cfg_chunks_mut {
...
+        result
+    }};
+}
+
+/// Applies a function and returns the first value that is not None
+#[macro_export]
+macro_rules! cfg_find_map {
+    ($self:ident, $object:expr, $func:ident) => {{
+        #[cfg(not(feature = "serial"))]
+        let result = $self.par_values().filter_map(|tx| tx.$func($object)).find_any(|_| true);
```
ljedrz, line 220:
```suggestion
        let result = $self.par_values().filter_map(|tx| tx.$func($object)).take_any(1);
```
I think this would make it a bit more clear

ljedrz (approved):
Left one or two comments, the direction makes sense in general.

utilities/src/parallel.rs
iamalwaysuncomfortable: %100

console/algorithms/src/bhp/hasher/mod.rs
Pratyush: Hm I would benchmark and see whether parallelization is beneficial here; it might be the case that parallelization introduces too much overhead.

synthesizer/src/block/transactions/mod.rs
```diff
@@ -64,87 +66,63 @@ impl<'a, N: Network> FromIterator<&'a Transaction<N>> for Transactions<N> {
 impl<N: Network> Transactions<N> {
     /// Returns `true` if the transactions contains the given transition ID.
     pub fn contains_transition(&self, transition_id: &N::TransitionID) -> bool {
-        match cfg!(feature = "parallel") {
-            true => self.par_values().any(|tx| tx.contains_transition(transition_id)),
-            false => self.values().any(|tx| tx.contains_transition(transition_id)),
+        match cfg!(feature = "serial") {
+            false => self.par_values().any(|tx| tx.contains_transition(transition_id)),
+            true => self.values().any(|tx| tx.contains_transition(transition_id)),
         }
```
Pratyush: Should make a `cfg_values` macro for this?

Pratyush:
```suggestion
        cfg_values!(self).any(|tx| tx.contains_transition(transition_id))
```

synthesizer/src/block/transactions/mod.rs
```diff
@@ -64,87 +66,63 @@ impl<'a, N: Network> FromIterator<&'a Transaction<N>> for Transactions<N> {
...
 
     /// Returns `true` if the transactions contains the given serial number.
     pub fn contains_serial_number(&self, serial_number: &Field<N>) -> bool {
-        match cfg!(feature = "parallel") {
-            true => self.par_values().any(|tx| tx.contains_serial_number(serial_number)),
-            false => self.values().any(|tx| tx.contains_serial_number(serial_number)),
+        match cfg!(feature = "serial") {
+            false => self.par_values().any(|tx| tx.contains_serial_number(serial_number)),
+            true => self.values().any(|tx| tx.contains_serial_number(serial_number)),
         }
```
Pratyush:
```suggestion
       cfg_values!(self).any(|tx| tx.contains_serial_number(serial_number))
```

synthesizer/src/block/transactions/mod.rs
```diff
@@ -64,87 +66,63 @@ impl<'a, N: Network> FromIterator<&'a Transaction<N>> for Transactions<N> {
...
 
     /// Returns `true` if the transactions contains the given commitment.
     pub fn contains_commitment(&self, commitment: &Field<N>) -> bool {
-        match cfg!(feature = "parallel") {
-            true => self.par_values().any(|tx| tx.contains_commitment(commitment)),
-            false => self.values().any(|tx| tx.contains_commitment(commitment)),
+        match cfg!(feature = "serial") {
+            false => self.par_values().any(|tx| tx.contains_commitment(commitment)),
+            true => self.values().any(|tx| tx.contains_commitment(commitment)),
         }
```
Pratyush:
```suggestion
       cfg_values!(self).any(|tx| tx.contains_commitment(commitment))
```

synthesizer/src/coinbase_puzzle/mod.rs
Pratyush: Since the only diff between `rayon::ParallelIterator::fold` and `std::Iterator::fold` is in the initialization value (the former takes a closure while the latter directly takes a value), a suggestion to deduplicate logic would be to do the following:

```rust
#[cfg(serial)]
let init = DensePolynomial::zero();
#[cfg(not(feature = "serial")]
let init = DensePolynomial::zero;

let thing = cfg_iter!(...).fold(init, stuff);
```

synthesizer/src/coinbase_puzzle/mod.rs
Pratyush: Same suggestion as above.

utilities/src/parallel.rs
```diff
@@ -17,9 +17,9 @@
 use crate::{boxed::Box, vec::Vec};
 
 pub struct ExecutionPool<'a, T> {
-    #[cfg(feature = "parallel")]
+    #[cfg(not(feature = "serial"))]
     jobs: Vec<Box<dyn 'a + FnOnce() -> T + Send>>,
-    #[cfg(not(feature = "parallel"))]
+    #[cfg(feature = "serial")]
```
Pratyush, line 22: If we're making `add_job` take in `Send` closures, then this can be deduplicated:

```suggestion
    jobs: Vec<Box<dyn 'a + FnOnce() -> T + Send>>,
```

utilities/src/parallel.rs
```diff
@@ -17,9 +17,9 @@
...
     jobs: Vec<Box<dyn 'a + FnOnce() -> T>>,
```
Pratyush: If we're making `add_job` take in `Send` closures, then this can be deduplicated:

```suggestion
    jobs: Vec<Box<dyn 'a + FnOnce() -> T + Send>>,
```

Pratyush:
Looks like I reviewed an old diff :facepalm:. Deleting some stale suggestions.

synthesizer/src/coinbase_puzzle/mod.rs
```diff
@@ -208,14 +210,10 @@ impl<N: Network> CoinbasePuzzle<N> {
...
+        // Accumulate the prover polynomial.
         let accumulated_prover_polynomial = cfg_into_iter!(prover_polynomials)
             .zip_eq(challenges)
-            .fold(DensePolynomial::zero, |mut accumulator, (mut prover_polynomial, challenge)| {
-                prover_polynomial *= challenge;
-                accumulator += &prover_polynomial;
-                accumulator
-            })
+            .map(|(prover_polynomial, challenge)| prover_polynomial * challenge)
             .sum::<DensePolynomial<_>>();
```
Pratyush: Since the only diff between `rayon::ParallelIterator::fold` and `std::Iterator::fold` is in the initialization value (the former takes a closure while the latter directly takes a value), a suggestion to deduplicate logic would be to do the following:

```rust
#[cfg(serial)]
let init = DensePolynomial::zero();
#[cfg(not(feature = "serial")]
let init = DensePolynomial::zero;

let thing = cfg_iter!(...).fold(init, stuff);
```

We want to preserve the old behaviour as it reduces memory consumption

synthesizer/src/coinbase_puzzle/mod.rs
```diff
@@ -312,10 +310,8 @@ impl<N: Network> CoinbasePuzzle<N> {
         // Compute the accumulator evaluation.
         let mut accumulator_evaluation = cfg_iter!(prover_polynomials)
             .zip_eq(&challenge_points)
-            .fold(<N::PairingCurve as PairingEngine>::Fr::zero, |accumulator, (prover_polynomial, challenge_point)| {
-                accumulator + (prover_polynomial.evaluate(accumulator_point) * challenge_point)
-            })
-            .sum();
+            .map(|(prover_polynomial, challenge_point)| prover_polynomial.evaluate(accumulator_point) * challenge_point)
+            .sum::<<<N as console::prelude::Environment>::PairingCurve as PairingEngine>::Fr>();
```
Pratyush: Same as above.

synthesizer/src/coinbase_puzzle/mod.rs
Pratyush: We could also introduce a `cfg_fold!` macro analogously to `cfg_reduce`.

utilities/src/parallel.rs
iamalwaysuncomfortable: The type signature isn't the same and trying to make that work complicates the iterator a bit more

Pratyush:
btw to compare runs of the same benchmarks, you can use critcmp: https://github.com/BurntSushi/critcmp

Pratyush:
Also re: CI, we can make it a `check` run, so at least we know things compile. This should be lightweight enough, and our `cfg_xyz` infrastructure is simple enough that there should be any behavioural differences between the `serial` and normal versions.

synthesizer/src/coinbase_puzzle/mod.rs
raychu86: We may want to just do the "dumb" approach of having a full separation of the `serial` and non-`serial` logic, because the 
`rayon::ParallelIterator::fold` and `std::Iterator::fold` differs in the initalizer and final `sum` operation. 

The parallel iterator requires the `sum` operation because the objects are [disjointly folded](https://docs.rs/rayon/latest/rayon/iter/trait.ParallelIterator.html#method.fold) into breakpoints, whereas `std::Iterator::fold` already outputs the final sum.


Example:
```
        #[cfg(feature = "serial")]
        let mut accumulator_evaluation = cfg_iter!(prover_polynomials).zip_eq(&challenge_points).fold(
            <N::PairingCurve as PairingEngine>::fr::zero(),
            |accumulator, (prover_polynomial, challenge_point)| {
                accumulator + (prover_polynomial.evaluate(accumulator_point) * challenge_point)
            },
        );

        let mut accumulator_evaluation = cfg_iter!(prover_polynomials)
            .zip_eq(&challenge_points)
            .fold(<N::PairingCurve as PairingEngine>::fr::zero, |accumulator, (prover_polynomial, challenge_point)| {
                accumulator + (prover_polynomial.evaluate(accumulator_point) * challenge_point)
            })
            .sum();
```

This is more verbose, however more clear because it doesn't require an additional `cfg` flag for the final sum operation:
``` 
let accumulator_evaluation = accumulator_evaluation.sum::<DensePolynomial<_>>();
```

iamalwaysuncomfortable: I pushed up a macro for the `fold` operation as Pratyush suggested, but I think there's an argument for the dumb way. Lots of rust macros often save a few lines but they also create levels of indirection when trying to understand the code. I'd personally pick the "dumb" approach here for readability, but I used a macro as suggested!

utilities/src/parallel.rs
iamalwaysuncomfortable: Done!

synthesizer/src/coinbase_puzzle/mod.rs
iamalwaysuncomfortable: I encapsulated that in a single macro in the latest commit.

circuit/algorithms/Cargo.toml
```diff
@@ -27,10 +27,11 @@ default-features = false
...
 path = "../../utilities"
+default-features = false
 
 [dev-dependencies.anyhow]
 version = "1.0.69"
 
 [features]
 default = [ "enable_console" ]
-enable_console = [ "console" ]
+enable_console = [ "console" ]
```
raychu86, line 37: Nit: Add a new-line here

raychu86 (approved):

raychu86:
`cfg_iter` change in the bhp hasher will be addressed and further benchmarked in #1395

PR #1389, opened 2023-03-08
Test Framework for `synthesizer`.

d0cd:
This PR introduces a test framework for the `synthesizer` crate.
The goal of this framework is to define and run a large number integration tests.
The framework logs and checks against the expected output to ensure stability.

See `test_parser` for two examples of `parser` tests.
- `tests/parser/program` defines tests for `Program::parse` at the file granularity.
- `tests/parser/instruction` defines tests for `Instruction::parse` at the line granularity.


synthesizer/tests/utilities/traits.rs
```diff
@@ -0,0 +1,40 @@
...
+    type Test;
+    type Output;
+
+    /// Loads the expectation from the given path.
+    fn load<P: AsRef<Path>>(input: P) -> Result<Self>;
+
+    /// Checks the expectation against the given output.
+    fn check(&self, test: &Self::Test, output: &Self::Output) -> Result<()>;
+
+    fn save(&self, output: &Self::Output) -> Result<()>;
```
iamalwaysuncomfortable:
```suggestion

    /// Saves the expectation into a given output type
    fn save(&self, output: &Self::Output) -> Result<()>;
```

synthesizer/tests/expectations/parser/instruction/instruction_pass.out
```diff
@@ -0,0 +1 @@
+- Parsing was successful.
```
iamalwaysuncomfortable, line 1: Is this included to provide an example of outfile format?

synthesizer/tests/utilities/expectations.rs
```diff
@@ -0,0 +1,155 @@
...
+pub fn get_expectation_path<P: AsRef<Path>>(test_path: P) -> PathBuf {
+    // Get the current directory.
+    let current_dir = current_dir().expect("Failed to get current directory.");
+    // Construct the path to the expectations directory.
+    let expectations_dir = current_dir.join("tests/expectations");
+    // Construct the path to the test directory.
+    let tests_dir = current_dir.join("tests/tests");
+    // Construct the path to the expectation file.
+    expectations_dir.join(test_path.as_ref().to_path_buf().strip_prefix(tests_dir).unwrap().with_extension("out"))
+}
```
iamalwaysuncomfortable, line 119: This is expecting a specific structure in the test directory. Perhaps there should be an associated setup (and maybe teardown/cleanup) function that constructs this automatically?

synthesizer/tests/utilities/traits.rs
```diff
@@ -0,0 +1,40 @@
...
+    fn run(&self);
+}
+
+/// A general interface for an expectation.
+pub trait Expectation: Sized {
+    type Test;
+    type Output;
+
+    /// Loads the expectation from the given path.
+    fn load<P: AsRef<Path>>(input: P) -> Result<Self>;
```
iamalwaysuncomfortable: Env vars are sometimes hard to work with. This seems maybe be the beginning of a linting system, so a potential idea to make it more flexible is to allow a config object to be specified in the method directly.

```suggestion
    fn load<P: AsRef<Path>, T>(input: P, config: Option<&T>) -> Result<Self>;
```

synthesizer/tests/utilities/traits.rs
```diff
@@ -0,0 +1,40 @@
...
+
+// You should have received a copy of the GNU General Public License
+// along with the snarkVM library. If not, see <https://www.gnu.org/licenses/>.
+
+use super::*;
+
+/// A general interface for a test.
+pub trait Test: Sized {
+    /// Loads the test from the given path.
+    fn load<P: AsRef<Path>>(input: P) -> Result<Self>;
```
iamalwaysuncomfortable: Env vars are sometimes hard to work with. This seems maybe be the beginning of a linting system, so a potential idea to make it more flexible is to allow a config object to be specified in the method directly.

```suggestion
    fn load<P: AsRef<Path>, T>(input: P, config: Option<&T>) -> Result<Self>;
```

iamalwaysuncomfortable (commented):
Main points
* A suggestion about putting a config option into the trait
* The directory structure for this seems to need to be constructed manually? So made a quicky suggestion of potentially putting an auto-setup & cleanup/teardown

synthesizer/tests/expectations/parser/instruction/instruction_pass.out
d0cd: Yes, the test framework checks against this file in each run.
This helps ensure that the test output (what's reported by the parser) is consistent.

synthesizer/tests/utilities/traits.rs
d0cd: That's a fair point, however adding a generic `config` is challenging to work into the test `Runner`.
The intention was that an implementation of `Test` or `Expectation` specifies how to extract its configuration.
Usually, this would be from reading the test or expectation file.

That said, it would be good to revisit this idea if we have a need to extend the test framework.
Also for context, I hadn't considered how this would function as a linting framework.

synthesizer/tests/utilities/traits.rs
d0cd: See discussion below.

synthesizer/tests/utilities/traits.rs
iamalwaysuncomfortable: Yeah agreed, we don't want to make it took complicated before it needs to be. Just a thought for the future :)

console/network/environment/src/traits/parse.rs
```diff
@@ -14,11 +14,26 @@
...
+use nom::{
+    error::{convert_error, VerboseError},
+    Err as NomErr,
+    IResult,
+};
 
 /// The `nom`-compatible parser return type.
 pub type ParserResult<'a, O> = IResult<&'a str, O, VerboseError<&'a str>>;
 
+/// Convert's a parser result into a useful message.
```
howardwu:
```suggestion
/// Converts a `ParserResult` into a human-readable message.
```

howardwu (approved):
LGTM

PR #1395, opened 2023-03-14
Add additional benchmarks

raychu86:
## Motivation

This PR adds additional benchmarks for -

- BHP256, BHP512, BHP768, BHP1024
- Execution transactions
- Deployment transactions

### DIFF
--- a/console/algorithms/benches/bhp.rs
+++ b/console/algorithms/benches/bhp.rs
@@ -69,7 +69,7 @@ fn bhp1024(c: &mut Criterion) {
 
 criterion_group! {
     name = bhp;
-    config = Criterion::default().sample_size(50);
+    config = Criterion::default().sample_size(1000);
     targets = bhp256, bhp512, bhp768, bhp1024
 }
 
--- a/console/algorithms/benches/bhp.rs
+++ b/console/algorithms/benches/bhp.rs
@@ -1,0 +1,76 @@
+// Copyright (C) 2019-2023 Aleo Systems Inc.
+// This file is part of the snarkVM library.
+
+// The snarkVM library is free software: you can redistribute it and/or modify
+// it under the terms of the GNU General Public License as published by
+// the Free Software Foundation, either version 3 of the License, or
+// (at your option) any later version.
+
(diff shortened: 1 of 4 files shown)


console/algorithms/benches/bhp.rs
```diff
@@ -0,0 +1,76 @@
...
+
+fn bhp256(c: &mut Criterion) {
+    let rng = &mut TestRng::default();
+    let hash = BHP256::<Console>::setup("BHP256").unwrap();
+
+    let input = (0..hash.window_size() as u64 * hash.num_windows() as u64).map(|_| bool::rand(rng)).collect::<Vec<_>>();
+    c.bench_function(&format!("BHP256 Hash - input size {}", input.len()), |b| b.iter(|| hash.hash(&input)));
+
+    let input = (0..256).map(|_| bool::rand(rng)).collect::<Vec<_>>();
+    c.bench_function(&format!("BHP256 Hash - input size {}", input.len()), |b| b.iter(|| hash.hash(&input)));
```
iamalwaysuncomfortable, line 34: Does it make sense to separate these two into separate functions?

synthesizer/benches/transaction.rs
```diff
@@ -0,0 +1,152 @@
...
+program helloworld.aleo;
+
+function hello:
+    input r0 as u32.private;
+    input r1 as u32.private;
+    add r0 r1 into r2;
+    output r2 as u32.private;
+",
+    )
+    .unwrap();
```
iamalwaysuncomfortable, line 83: More benches for longer programs and different opcodes can probably added to this eventually.

I'd imagine we can think out a distribution of programs that we expect to see on chain (things like swaps, nft minters, utility programs, etc.) and get a couple representative instruction sets from those and bench them.

Also are there any opcodes in particular we want to see benched?

iamalwaysuncomfortable (commented):
Couple of questions about splitting up the BHP hash benches + potentially adding benches of longer programs.

If we're looking at longer benches it's likely we can put those on a manual trigger in CI so that they can be run at points at which the team needs them.

howardwu (approved):

PR #1397, opened 2023-03-15
Update `credits.aleo` program

raychu86:
## Motivation

This PR updates the `credits.aleo` program to support public credits via mappings.

List of changes:

1. `mint` is renamed to `mint_private`
2. `transfer` is renamed to `transfer_private`

List of new functions:

1. `mint_public` - Mints credits to the public mapping 
2. `transfer_public` - Transfer credits from public between addresses on the public mapping
3. `transfer_public_to_private` - Transfers credits from the public mapping to a private record.
4. `transfer_private_to_public` - Transfers credits from a private record to a public mapping.


synthesizer/src/program/mod.rs
```diff
@@ -124,6 +139,47 @@ function transfer:
...
+    cast r0 r1 into r2 as credits.record;
+    output r2 as credits.record;
+
+    finalize self.caller r1;
+
+finalize transfer_public_to_private:
+    input r0 as address.public;
+    input r1 as u64.public;
+    decrement account[r0] by r1;
+
```
iamalwaysuncomfortable, line 182: I like what's being done here. The counter-argument would be to "keep `credits.aleo` simple", but I think that the simplicity counter-argument is the wrong argument to focus on.  Since credits.aleo is the only program that can do official Aleo credit transfers, keeping it simple only pushes complexity to the end users when they want to do things like:
* Transfer to multiple users
* Transfer privately or publicly (or transition between the two)
* Make multi-input or multi-output transfers

And I think pushing it to the users is definitely the wrong thing to do as it makes our chain much more complicated to work with. I wholly support this addition and would support further adding more utilities. In particular I'd suggesting making multi-input and multi-output functions - as this even currently is a bottleneck for the faucet and has acted as a bottleneck for some of our partners. Solving the harsh reality of the knapsack problem present in UTXO-style blockchains is very difficult if you have to do it 1 input and 1 output at a time.


```suggestion
function transfer_private_2_inputs:
    input r0 as credits.record;
    input r1 as credits.record;
    input r2 as address.private;
    input r3 as u64.private;
    add r0.gates r1.gates into r4;
    sub r4 r3 into r5
    cast r2 r3 into r6 as credits.record;
    cast r0.owner r5 into r7 as credits.record;
    output r6 as credits.record;
    output r7 as credits.record;

function transfer_private_3_inputs:
    input r0 as credits.record;
    input r1 as credits.record;
    input r2 as credits.record;
    input r3 as address.private;
    input r4 as u64.private;
    add r0.gates r1.gates into r5;
    add r5 r2.gates into r6; 
    sub r6 r4 into r7;
    cast r3 r4 into r8 as credits.record;
    cast r0.owner r7 into r9 as credits.record;
    output r8 as credits.record;
    output r9 as credits.record;

function transfer_private_8_inputs:
    input r0 as credits.record;
    input r1 as credits.record;
    input r2 as credits.record;
    input r3 as credits.record;
    input r4 as credits.record;
    input r5 as credits.record;
    input r6 as credits.record;
    input r7 as credits.record;
    input r8 as address.private;
    input r9 as u64.private;
    add r0.gates r1.gates into r10;
    add r10 r2.gates into r11;
    add r11 r3.gates into r12;
    add r12 r4.gates into r13;
    add r13 r5.gates into r14;
    add r14 r6.gates into r15;
    add r15 r7.gates into r16;
    sub r16 r9 into r17;
    cast r8 r9 into r18 as credits.record;
    cast r0.owner r17 into r19 as credits.record;
    output r18 as credits.record;
    output r19 as credits.record;

function transfer_private_2_outputs:
    input r0 as credits.record;
    input r1 as address.private;
    input r2 as address.private;
    input r3 as u64.private;
    input r4 as u64.private;
    sub r0.gates r3 into r5;
    sub r5 r4 into r6;
    cast r1 r3 into r7 as credits.record;
    cast r2 r4 into r8 as credits.record;
    cast r0.owner r6 into r9 as credits.record;
    output r7 as credits.record;
    output r8 as credits.record;
    output r9 as credits.record;

function transfer_private_3_outputs:
    input r0 as credits.record;
    input r1 as address.private;
    input r2 as address.private;
    input r3 as address.private;
    input r4 as u64.private;
    input r5 as u64.private;
    input r6 as u64.private;
    sub r0.gates r4 into r7;
    sub r7 r5 into r8;
    sub r8 r6 into r9;
    cast r1 r4 into r10 as credits.record;
    cast r2 r5 into r11 as credits.record;
    cast r3 r6 into r12 as credits.record;
    cast r0.owner r9 into r13 as credits.record;
    output r10 as credits.record;
    output r11 as credits.record;
    output r12 as credits.record;
    output r13 as credits.record;

function transfer_private_7_outputs:
    input r0 as credits.record;
    input r1 as address.private;
    input r2 as address.private;
    input r3 as address.private;
    input r4 as address.private;
    input r5 as address.private;
    input r6 as address.private;
    input r7 as address.private;
    input r8 as u64.private;
    input r9 as u64.private;
    input r10 as u64.private;
    input r11 as u64.private;
    input r12 as u64.private;
    input r13 as u64.private;
    input r14 as u64.private;
    sub r0.gates r8 into r15;
    sub r15 r9 into r16;
    sub r16 r10 into r17;
    sub r17 r11 into r18;
    sub r18 r12 into r19;
    sub r19 r13 into r20;
    sub r20 r14 into r21;
    cast r1 r8 into r22 as credits.record;
    cast r2 r9 into r23 as credits.record;
    cast r3 r10 into r24 as credits.record;
    cast r4 r11 into r25 as credits.record;
    cast r5 r12 into r26 as credits.record;
    cast r6 r13 into r27 as credits.record;
    cast r7 r14 into r28 as credits.record;
    cast r0.owner r21 into r29 as credits.record;
    output r22 as credits.record;
    output r23 as credits.record;
    output r24 as credits.record;
    output r25 as credits.record;
    output r26 as credits.record;
    output r27 as credits.record;
    output r28 as credits.record;
    output r29 as credits.record;
```

The above would add what seems like boilerplate, but these functions, combined with multiple transitions would allow for high throughput. When working on other layer 1 blockchains, I found the following:
* Exchanges needed extremely high throughput during high trading volumes, which we should probably attempt to offer
* Making consumer applications that were hit heavily by end users also required high utxo throughput
* Allow application developers some convenience in managing their balances.

I realize public transfers may be harder due to finalize, but we should consider private transfers that are constructed offline have the above utilities. Secondly, I believe in the changes Ray's and think we should probably have a discussion quick about how to change credits.aleo before the next launch of testnet3 as it would allow us to fail out our assumptions with real traffic.

raychu86: We're aware of the limitations of the current model of individual transfers, however our approach is to address it via Multi-execution transactions, where we can compose a transaction with multiple state transitions. In my opinion it is more elegant than having monolith function circuits (though we still may want to consider possible updates here)

iamalwaysuncomfortable: Yeah that makes sense, I think potentially maybe just having one private multi-input and one private multi-output with the maximum number of outputs (which I think is 8) be available for situations where high performance is needed.

synthesizer/src/program/mod.rs
```diff
@@ -108,13 +108,28 @@ record credits:
     owner as address.private;
     gates as u64.private;
 
-function mint:
+mapping account:
```
evanmarshall, line 111: I would strongly recommend adding approvals to the default credits program: https://github.com/AleoHQ/aleo/pull/304/files#diff-34ed4dabecf71a6a2db55da718ffbd17767d9c7ec0245300e4da5a348ab0114eR30

Honestly a token standard should probably be designed and worked through before finalizing this program. If it differs from the later implemented token standard, it's going to lead to every program needing to "wrap" microcredits in order for it be used.

evanmarshall: It may also be time to consider adding `self.parent`. In order for there to be a composable ecosystem of contracts, contracts will need to be able to own assets through mappings.

PR #1398, opened 2023-03-17
[Feature] Merkle Tree Updates

d0cd:
This PR introduces "single" and "batched" updates to `MerkleTree`.

### Evaluation
We ran the benchmarks in `benches/merkle_tree.rs` to evaluate the performance of the update operations.
The evaluation was performed on an M1 Max w/ 32 GB RAM.

#### MerkleTree::new
| # leaves | Time (ms)
| -------- | -------- | 
| 1        | 1.9390 |
| 10       | 2.9340 |
| 100      | 12.827 |
| 1000     | 23.497 |
| 10000    | 178.71 |
| 100000   | 1427.2 |
| 1000000  | 12439 |

#### `append` vs `update` vs `batch_update` 
Note that the Merkle Tree is initialized with `1,000,000` leaves.
Note that `batch_update` is performed over **unique** leaves.
| # leaves updated/added | `append` time (ms) | `update` time (ms) |  `batch_update` time (ms) |
| ---------------------- | ------------------ | ------------------ | ------------------------- |
| 1       | 16.262 | 2.5093  | 2.6949 |
| 10      | 17.797 | 19.643  | 3.3254 |
| 100     | 25.940 | 195.45  | 13.110 |
| 1000    | 36.054 | 1926.1  | 27.064 |
| 10000   | 149.39 | 19214   | 153.66 |
| 100000  | 7912.0 | x       | 1366.6 |
| 1000000 | 12790  | x       | _(~864813 unique leaves)_ 11549 |

#### Update a single leaf (`update` vs `batch_update`) 

We evaluated `MerkleTree::update` and `MerkleTree::batch_update` for a single leaf on trees of varying (initialized) depths. Concretely, we tested depths `d = 1, ..., 15`. For all `d`, we found no significant difference in the execution time between the standard and batched implementations: `~1.9ms`. Interestingly, the execution time did not vary across depths. This could be due to execution of the benchmarks being dominated by setup and teardown of the Merkle Tree.


console/collections/benches/merkle_tree.rs
```diff
@@ -39,62 +40,144 @@ macro_rules! generate_leaves {
...
+        // Generate all of the updates that will be applied to the tree.
+        // Note that we generate 2 * MAX_UPDATE_SIZE leaves to increase the number of unique of updates.
+        let mut updates = generate_leaves!(2 * *UPDATE_SIZES.last().unwrap(), &mut rng)
+            .into_iter()
+            .map(|leaf| {
+                let index: usize = Uniform::rand(&mut rng);
+                (index % num_leaves, leaf)
+            })
+            .collect::<Vec<_>>();
+        updates.sort_by(|(a, _), (b, _)| a.cmp(b));
```
ljedrz:
```suggestion
        updates.sort_by_key(|(a, _)| a);
```

ljedrz: haven't tested it, might require dereferencing; also, since it is reversed right below, [`cmp::Reverse`](https://doc.rust-lang.org/std/cmp/struct.Reverse.html) could be applied here instead

console/collections/src/merkle_tree/mod.rs
```diff
@@ -208,6 +208,206 @@ impl<E: Environment, LH: LeafHash<Hash = PH::Hash>, PH: PathHash<Hash = Field<E>
...
+        lap!(timer, "Hashed {} padding levels", padding_depth);
+
+        // Update the root hash.
+        self.root = root_hash;
+
+        // Update the rest of the tree with the new path hashes.
+        let mut index = Some(start + leaf_index);
+        for path_hash in path_hashes {
+            self.tree[index.unwrap()] = path_hash;
+            index = parent(index.unwrap());
```
ljedrz, line 267:
```suggestion
        let mut index = start + leaf_index;
        for path_hash in path_hashes {
            self.tree[index] = path_hash;
            index = parent(index).unwrap();
```

console/collections/src/merkle_tree/mod.rs
```diff
@@ -208,6 +208,206 @@ impl<E: Environment, LH: LeafHash<Hash = PH::Hash>, PH: PathHash<Hash = Field<E>
...
+                .map(|(index, (left, right))| self.path_hasher.hash_children(left, right).map(|hash| (*index, hash)))
+                .collect::<Result<Vec<_>>>(),
+        };
+
+        // Compute the depth of the tree. This corresponds to the number of levels of hashes in the tree.
+        let tree_depth = tree_depth::<DEPTH>(self.tree.len())?;
+        // For each level in the tree, compute the path hashes.
+        for level in 0..tree_depth as usize {
+            // Prepare the inputs to path hasher.
+            let mut inputs = Vec::with_capacity(updated_hashes[level].len());
```
ljedrz: this vector could be allocated once before the `for` loop, `clear`ed after its each iteration, and reused

console/collections/src/merkle_tree/tests/batch.rs
```diff
@@ -0,0 +1,675 @@
...
+    updates: &[(usize, LH::Leaf)],
+) -> Result<()> {
+    // Construct the Merkle tree for the given leaves.
+    let mut merkle_tree = MerkleTree::<E, LH, PH, DEPTH>::new(leaf_hasher, path_hasher, leaves)?;
+    assert_eq!(leaves.len(), merkle_tree.number_of_leaves);
+
+    let mut rng = TestRng::default();
+
+    // Check each leaf in the Merkle tree.
+    if !leaves.is_empty() {
```
ljedrz: nit: this check isn't needed

console/collections/src/merkle_tree/tests/batch.rs
```diff
@@ -0,0 +1,675 @@
...
+                )?;
+            }
+        }
+        Ok(())
+    }
+
+    let mut rng = TestRng::default();
+
+    // Ensure DEPTH = 0 fails.
+    assert!(run_test::<0>(&mut rng).is_err());
```
ljedrz: [`seq_macro`](https://docs.rs/seq-macro/0.3.3/seq_macro/) could be used to reduce some of this boilerplate; it would look more or less like this:
```
seq!(N in 1..=10 {
    assert!(run_test::<N>(&mut rng).is_ok());
});
```
alternatively (or in addition), a macro like this could be used:
```
macro_rules! run_tests {
    ($rng:expr, [$($i:expr),*]) => {
        $( assert!(run_test::<$i>($rng).is_ok()); )*
    };
}

run_tests!(&mut rng, [1, 2, 10]); // just listing a few
```

console/collections/src/merkle_tree/tests/batch.rs
```diff
@@ -0,0 +1,675 @@
...
+            let num_leaves = 2u128.pow(DEPTH as u32);
+
+            // Construct the random updates.
+            let mut updates = (0..num_leaves)
+                .map(|_| {
+                    let index: u128 = Uniform::rand(rng);
+                    ((index % num_leaves) as usize, Field::<CurrentEnvironment>::rand(rng).to_bits_le())
+                })
+                .collect::<Vec<(usize, Vec<bool>)>>();
+            updates.sort_by(|(a, _), (b, _)| a.cmp(b));
```
ljedrz:
```suggestion
            updates.sort_by_key(|(a, _)| a));
```
plus the same `cmp::Reverse` note as before

console/collections/src/merkle_tree/tests/batch.rs
```diff
@@ -0,0 +1,675 @@
...
+            let num_leaves = 2u128.pow(DEPTH as u32);
+
+            // Construct the random updates.
+            let mut updates = (0..num_leaves)
+                .map(|_| {
+                    let index: u128 = Uniform::rand(rng);
+                    ((index % num_leaves) as usize, vec![Uniform::rand(rng)])
+                })
+                .collect::<Vec<(usize, Vec<_>)>>();
+            updates.sort_by(|(a, _), (b, _)| a.cmp(b));
```
ljedrz: ditto

console/collections/src/merkle_tree/tests/batch.rs
```diff
@@ -0,0 +1,675 @@
...
+            let single_updates = (0..num_leaves)
+                .map(|_| {
+                    let index: u128 = Uniform::rand(rng);
+                    ((index % num_leaves) as usize, Field::<CurrentEnvironment>::rand(rng).to_bits_le())
+                })
+                .collect::<Vec<(usize, Vec<bool>)>>();
+
+            // Construct the batch updates for the single updates.
+            let mut batch_updates = single_updates.clone();
+            batch_updates.sort_by(|(a, _), (b, _)| a.cmp(b));
```
ljedrz: ditto

console/collections/src/merkle_tree/tests/batch.rs
```diff
@@ -0,0 +1,675 @@
...
+            let single_updates = (0..num_leaves)
+                .map(|_| {
+                    let index: u128 = Uniform::rand(rng);
+                    ((index % num_leaves) as usize, vec![Uniform::rand(rng)])
+                })
+                .collect::<Vec<(usize, Vec<_>)>>();
+
+            // Construct the batch updates for the single updates.
+            let mut batch_updates = single_updates.clone();
+            batch_updates.sort_by(|(a, _), (b, _)| a.cmp(b));
```
ljedrz: ditto

console/collections/src/merkle_tree/tests/batch.rs
```diff
@@ -0,0 +1,675 @@
...
+
+    // Check the depth-3 Merkle tree.
+    check_merkle_tree_depth_3_single_update::<CurrentEnvironment, LH, PH>(
+        &leaf_hasher,
+        &path_hasher,
+        &(0..4).map(|_| Field::<CurrentEnvironment>::rand(&mut rng).to_bits_le()).collect::<Vec<Vec<bool>>>(),
+        &(0..1)
+            .rev()
+            .map(|i| (i, Field::<CurrentEnvironment>::rand(&mut rng).to_bits_le()))
+            .collect::<Vec<(usize, Vec<bool>)>>(),
```
ljedrz:
```suggestion
        &[(0, Field::<CurrentEnvironment>::rand(&mut rng).to_bits_le())]
```

console/collections/src/merkle_tree/tests/batch.rs
```diff
@@ -0,0 +1,675 @@
...
+    let path_hasher = PH::setup("AleoMerkleTreeTest1")?;
+
+    let mut rng = TestRng::default();
+
+    // Check the depth-3 Merkle tree.
+    check_merkle_tree_depth_3_single_update::<CurrentEnvironment, LH, PH>(
+        &leaf_hasher,
+        &path_hasher,
+        &(0..4).map(|_| vec![Uniform::rand(&mut rng)]).collect::<Vec<_>>(),
+        &(0..1).rev().map(|i| (i, vec![Uniform::rand(&mut rng)])).collect::<Vec<_>>(),
```
ljedrz:
```suggestion
        &[(0, vec![Uniform::rand(&mut rng)])],
```

console/collections/src/merkle_tree/tests/update.rs
```diff
@@ -0,0 +1,520 @@
...
+    updates: &[(usize, LH::Leaf)],
+) -> Result<()> {
+    // Construct the Merkle tree for the given leaves.
+    let mut merkle_tree = MerkleTree::<E, LH, PH, DEPTH>::new(leaf_hasher, path_hasher, leaves)?;
+    assert_eq!(leaves.len(), merkle_tree.number_of_leaves);
+
+    let mut rng = TestRng::default();
+
+    // Check each leaf in the Merkle tree.
+    if !leaves.is_empty() {
```
ljedrz: nit: this is not needed

console/collections/src/merkle_tree/tests/update.rs
```diff
@@ -0,0 +1,520 @@
...
+            assert!(proof.verify(leaf_hasher, path_hasher, merkle_tree.root(), leaf));
+            // Verify the Merkle proof **fails** on an invalid root.
+            assert!(!proof.verify(leaf_hasher, path_hasher, &PH::Hash::zero(), leaf));
+            assert!(!proof.verify(leaf_hasher, path_hasher, &PH::Hash::one(), leaf));
+            assert!(!proof.verify(leaf_hasher, path_hasher, &PH::Hash::rand(&mut rng), leaf));
+        }
+    }
+
+    // If additional leaves are provided, check that the Merkle tree is consistent with them.
+    if !updates.is_empty() {
```
ljedrz: nit: technically neither is this, but id does group things nicely

console/collections/src/merkle_tree/tests/update.rs
```diff
@@ -0,0 +1,520 @@
...
+        merkle_tree.update(*index, leaf)?;
+    }
+
+    // Add the updated leaves to the index map.
+    for (index, leaf) in updates {
+        leaf_map.insert(index, leaf);
+    }
+
+    // Get the updated leaves.
+    let updated_leaves = leaf_map.values().cloned().collect::<Vec<LH::Leaf>>();
```
ljedrz, line 109: since it immediately gets collected into a vector, is the `IndexMap` for the `leaf_map` needed at all?

console/collections/src/merkle_tree/tests/update.rs
```diff
@@ -0,0 +1,520 @@
...
+    let mut rng = TestRng::default();
+
+    // Check the depth-3 Merkle tree.
+    check_merkle_tree_depth_3_single_update::<CurrentEnvironment, LH, PH>(
+        &leaf_hasher,
+        &path_hasher,
+        &(0..4).map(|_| Field::<CurrentEnvironment>::rand(&mut rng).to_bits_le()).collect::<Vec<Vec<bool>>>(),
+        &(0..1)
+            .map(|i| (i, Field::<CurrentEnvironment>::rand(&mut rng).to_bits_le()))
+            .collect::<Vec<(usize, Vec<bool>)>>(),
```
ljedrz:
```suggestion
        &[(0, Field::<CurrentEnvironment>::rand(&mut rng).to_bits_le())],
```

console/collections/src/merkle_tree/tests/update.rs
```diff
@@ -0,0 +1,520 @@
...
+    let path_hasher = PH::setup("AleoMerkleTreeTest1")?;
+
+    let mut rng = TestRng::default();
+
+    // Check the depth-3 Merkle tree.
+    check_merkle_tree_depth_3_single_update::<CurrentEnvironment, LH, PH>(
+        &leaf_hasher,
+        &path_hasher,
+        &(0..4).map(|_| vec![Uniform::rand(&mut rng)]).collect::<Vec<_>>(),
+        &(0..1).map(|i| (i, vec![Uniform::rand(&mut rng)])).collect::<Vec<_>>(),
```
ljedrz:
```suggestion
        &[(0, vec![Uniform::rand(&mut rng)])],
```

ljedrz (commented):
Left some suggestions.

console/collections/benches/merkle_tree.rs
d0cd: Sound good! Using `cmp::Reverse` seems to complicate the logic since we want to keep the most recent update.

console/collections/src/merkle_tree/mod.rs
d0cd: In the last iteration of the loop, `parent(index)` results in a `None` value since we are at the root. 
By wrapping the index in the `Option`, we avoid the panic.

console/collections/src/merkle_tree/tests/update.rs
d0cd: The updates are done at random indices.
The `IndexMap` helps us keep track of them and get the unique updates.

ljedrz: I meant as opposed to a plain `Vec` instead of an `IndexMap`

console/collections/src/merkle_tree/tests/batch.rs
d0cd: Opting for the second solution.

console/collections/src/merkle_tree/tests/update.rs
d0cd: Ah yes, we could use a `Vec` and dedup the keys.
I figured using the `IndexMap` would emulate the right sort of abstraction.

ljedrz: As long as it's the best way to emulate the real world scenario, it's totally fine :+1:.

console/collections/benches/merkle_tree.rs
ljedrz: In order to get rid of the call to `reverse` below, the following can be done instead as well: `updates.sort_by(|(a, _), (b, _)| b.cmp(a));`

d0cd: We need it so that `updates` is in descending order of index with the last update made to that index.
`updates.sort_by(|(a, _), (b, _)| b.cmp(a));` results in descending order of index with the first update made to that index.
Is there a different sequence of operations that works? 
I'm having a tough time coming up with one

ljedrz: ah, ok; in that case I'd just add a comment on this so that it's not missed later on

console/collections/src/merkle_tree/tests/append.rs
```diff
@@ -604,8 +604,6 @@ fn test_profiler() -> Result<()> {
     let mut rng = TestRng::default();
 
     for num_leaves in NUM_LEAVES {
-        println!("Generating Merkle tree with {num_leaves} leaves, and appending 1 leaf...");
-
```
howardwu, line 608: revert

howardwu (approved):
LGTM. Note we will need to change this impl later because we will need `prepare_update` and `prepare_update_many` for our impl of finalize in the `Store` trait.

PR #1401, opened 2023-03-17
Introduce Continuous Benchmarking

raychu86:
## Motivation

This PR adds a GIthub actions workflow to run snarkVM benchmarks for each incoming PR. This will also track the past benchmark runtimes over a history of commits.


unbalancedparentheses:
Criterion usage is highly discouraged in CI environments. The results will have big deviations depending on the load of CI virtual machine. 

[iai](https://github.com/bheisler/iai) is highly recommended instead. iai uses cachegrind to count instructions and memory accesses instead of criterion's statistical analysis based on cpu running time.

raychu86:
> Criterion usage is highly discouraged in CI environments. The results will have big deviations depending on the load of CI virtual machine.
> 
> [iai](https://github.com/bheisler/iai) is highly recommended instead. iai uses cachegrind to count instructions and memory accesses instead of criterion's statistical analysis based on cpu running time.

Agreed, Criteron is definitely less than ideal for reaching consistent benchmarks. My worry about Iai is that it's "experimental" and lacks support. They intended to integrate `Iai` into `criterion`, but it doesn't seem ready. 

Iai also "cannot exclude setup code from the measurements"

.github/workflows/benchmarks.yml
```diff
@@ -0,0 +1,106 @@
...
+            ~/.cargo/git
+            target
+          key: ${{ runner.os }}-cargo-bench-${{ hashFiles('**/Cargo.lock') }}
+
+      - name: Benchmark algorithms
+        run: |
+          cd algorithms
+          cargo bench --bench variable_base -- --output-format bencher | tee -a ../output.txt
+          cargo bench --bench poseidon_sponge -- --output-format bencher | tee -a ../output.txt
+          cargo bench --bench marlin -- --output-format bencher | tee -a ../output.txt
```
iamalwaysuncomfortable, line 42: The FFT benches are quite important to not regress on, but I understand the running time is almost a non-starter for CI. However, the following approaches might work

1. *Create a reduced set of benches for CI wherein we sample a little bit less on each CI pull*: Our benches are quite exhaustive. For instance the merkle tree bench is O(number_of_starting_leaves*added_number_of_leaves) and our FFT benches bench  BENCHMARK_MIN_DEGREE..BENCHMARK_MAX_DEGREE. However instead of being `exhaustively` confident in the benches we could get `statistical confidence` by maybe just running 2 degrees within that range and a similar approach for the merkle tree & marlin benches.

2. Add the ability to run bench CI on a manual trigger and require the benchmark is run before an actual merge. This could cut the number of CI runs a lot.

iamalwaysuncomfortable (commented):
The FFT benches are quite important to not regress on, but I understand the running time is almost a non-starter for CI. However, the following approaches might work

Create a reduced set of benches for CI wherein we sample a little bit less on each CI pull: Our benches are quite exhaustive. For instance the merkle tree bench is O(number_of_starting_leaves*added_number_of_leaves) and our FFT benches bench BENCHMARK_MIN_DEGREE..BENCHMARK_MAX_DEGREE. However instead of being exhaustively confident in the benches we could get statistical confidence by maybe just running 2 degrees within that range and a similar approach for the merkle tree & marlin benches.

Add the ability to run bench CI on a manual trigger and require the benchmark is run before an actual merge. This could cut the number of CI runs a lot.

Pratyush:
> Iai also "cannot exclude setup code from the measurements"

We can handle this in two ways:
(a) For each benchmark, write a iai helper function which runs the benchmark code twice: once with only the setup code, and once with the setup and  benchmark code. Subtract the latter from the former, and we have a benchmark for the main code.
(b) If the purpose is primarily to detect regressions, then we only care about diffs between benchmarks, and not about the absolute values of the benchmarks.

However, I am overall reluctant to use Iai as it will require having two versions of each benchmark. Maybe instead we can use something like: https://github.com/jbreitbart/criterion-perf-events

iamalwaysuncomfortable (approved):
> > Iai also "cannot exclude setup code from the measurements"
>  (b) If the purpose is primarily to detect regressions, then we only care about diffs between benchmarks, and not about the absolute values of the benchmarks. 

^^ Yeah exactly this. We're using predictable hardware resources, even if the conditions CI is running under isn't consistent. There's some predictable random variable of CI runner conditions we're drawing from in each benchmark and the resulting sample distribution of benchmarks should reflect that (we should be able to spot it visually, but some simple statistics would be able to give the variance of CI runs if we ever need to suss out a performance issue)

> However, I am overall reluctant to use Iai as it will require having two versions of each benchmark. Maybe instead we can use something like: https://github.com/jbreitbart/criterion-perf-events

I don't think this is necessary - a deviation in performance caused by a software error would very likely skew the entire set of CI benchmark samples beyond the regular variance of CI runs. Also I think on CI runners some of the perf events we'd measure might not be entirely accurate because of the isolation from the actual OS.

I think this PR is good to merge as is

What I'd say could be done follow-in PRs are:
* Create hook to manually run the skipped benches & require them prior to merge
* Potentially collect the bench stats into a simple database (beyond what this action is doing) that we can query as needed for statistics

epompeii:
If you wanted to use more advanced statistical thresholds, which should help with the noisy CI runner problem, then you may want to check out the continuous benchmarking tool I've been working on, Bencher: https://github.com/bencherdev/bencher

Support of `iai` is also in the works: https://github.com/bencherdev/bencher/pull/98
And Bencher also supports multiple metric kinds per benchmark, which is required for `iai` and `github-action-benchmark` does not support it.

epompeii:
@howardwu just to follow up here, `iai` support has been added to Bencher if that is still of interest.

PR #1407, opened 2023-03-20
[Feature] Remove restriction on the number of records.

d0cd:
This PR removes restrictions on the number of records that can be defined in a program.
The current implementation allows users to define up to two records.


d0cd:
Opening this PR to start discussion on this change.
Is there anything else that needs to be done to support multiple records?

howardwu (approved):

PR #1413, opened 2023-03-23
Ensure usize serialization in cryptographic contexts is always 64 bits

iamalwaysuncomfortable:
## Motivation

There are 2 places in the code where `usize` integers are used in cryptographic contexts. 

1. [Absorption into the Poseidon Sponge during Marlin proofs](https://github.com/AleoHQ/snarkVM/blob/testnet3/algorithms/src/snark/marlin/marlin.rs#L153)
2. [Deserialization of the proving key](https://github.com/AleoHQ/snarkVM/blob/testnet3/utilities/src/serialize/impls.rs#L151)

In the sponge context, the proof only works if `usize` integers are represented as 8 bytes. If the Snark is run in a context where `usize` integers are represented without 8 bytes (such as a `wasm` context), the snark will generate an invalid proof.

Similarly, the deserialization within contexts where `usize` integers (again, in a `wasm` context) are represented as anything but 8 bytes leads to an incorrect proving key.

To ensure proof math computes correctly when using `usize` integers:
* `usize` integers are explicitly cast into 8 bytes before absorption into the sponge in the Marlin proofs
* `usize` integer serialization and deserialization is done explicitly in 8 bytes

This code was largely authored by Evan Marshall @ demox labs and modified by @iamalwaysuncomfortable - these serialization changes are currently in production enabling value transfers in the Leo wallet.

## Related PRs
This PR is a series of PRs to enable proving within `wasm` (and thus, within the browser). The following PRs are related to this effort:
Serial Flags: #1388 
[Enable Wasm Proofs in SnarkVM](https://github.com/demox-labs/snarkVM/pull/4)

### DIFF
--- a/algorithms/src/snark/marlin/marlin.rs
+++ b/algorithms/src/snark/marlin/marlin.rs
@@ -150,7 +150,7 @@ impl<E: PairingEngine, FS: AlgebraicSponge<E::Fq, 2>, MM: MarlinMode> MarlinSNAR
     ) -> FS {
         let mut sponge = FS::new_with_parameters(fs_parameters);
         sponge.absorb_bytes(&to_bytes_le![&Self::PROTOCOL_NAME].unwrap());
-        sponge.absorb_bytes(&batch_size.to_le_bytes());
+        sponge.absorb_bytes(&(batch_size as u64).to_le_bytes());
         sponge.absorb_native_field_elements(circuit_commitments);
         for input in inputs {
             sponge.absorb_nonnative_field_elements(input.iter().copied());
--- a/algorithms/src/snark/marlin/marlin.rs
+++ b/algorithms/src/snark/marlin/marlin.rs
@@ -150,7 +150,7 @@ impl<E: PairingEngine, FS: AlgebraicSponge<E::Fq, 2>, MM: MarlinMode> MarlinSNAR
     ) -> FS {
         let mut sponge = FS::new_with_parameters(fs_parameters);
         sponge.absorb_bytes(&to_bytes_le![&Self::PROTOCOL_NAME].unwrap());
-        sponge.absorb_bytes(&(batch_size as u64).to_le_bytes());
+        sponge.absorb_bytes(&(u64::try_from(batch_size).unwrap()).to_le_bytes());
         sponge.absorb_native_field_elements(circuit_commitments);
         for input in inputs {
             sponge.absorb_nonnative_field_elements(input.iter().copied());
--- a/utilities/src/serialize/impls.rs
+++ b/utilities/src/serialize/impls.rs
@@ -152,7 +152,7 @@ impl_canonical_serialization_uint!(u64);
 impl CanonicalSerialize for usize {
     #[inline]
     fn serialize_with_mode<W: Write>(&self, mut writer: W, _compress: Compress) -> Result<(), SerializationError> {
-        let u64_value = *self as u64;
+        let u64_value = u64::try_from(*self).map_err(|_| SerializationError::IncompatibleTarget)?;
         Ok(writer.write_all(&u64_value.to_le_bytes())?)
     }
 
--- a/utilities/src/serialize/impls.rs
+++ b/utilities/src/serialize/impls.rs
@@ -170,8 +170,8 @@ impl Valid for usize {
 
     #[inline]
     fn batch_check<'a>(_batch: impl Iterator<Item = &'a Self>) -> Result<(), SerializationError>
-        where
-            Self: 'a,
+    where
+        Self: 'a,
     {
         Ok(())
     }
--- a/utilities/src/serialize/impls.rs
+++ b/utilities/src/serialize/impls.rs
@@ -148,7 +148,46 @@ impl_canonical_serialization_uint!(u8);
 impl_canonical_serialization_uint!(u16);
 impl_canonical_serialization_uint!(u32);
 impl_canonical_serialization_uint!(u64);
-impl_canonical_serialization_uint!(usize);
+
+impl CanonicalSerialize for usize {
+    #[inline]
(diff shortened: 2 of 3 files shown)


vicsn (approved):

utilities/src/serialize/impls.rs
```diff
@@ -148,7 +148,46 @@ impl_canonical_serialization_uint!(u8);
 impl_canonical_serialization_uint!(u16);
 impl_canonical_serialization_uint!(u32);
 impl_canonical_serialization_uint!(u64);
-impl_canonical_serialization_uint!(usize);
+
+impl CanonicalSerialize for usize {
+    #[inline]
+    fn serialize_with_mode<W: Write>(&self, mut writer: W, _compress: Compress) -> Result<(), SerializationError> {
+        let u64_value = *self as u64;
+        Ok(writer.write_all(&u64_value.to_le_bytes())?)
```
raychu86, line 156: Why not 1-line this function?

Something like:
Ok(writer.write_all(&(*self as u64).to_le_bytes())?)

algorithms/src/snark/marlin/marlin.rs
```diff
@@ -150,7 +150,7 @@ impl<E: PairingEngine, FS: AlgebraicSponge<E::Fq, 2>, MM: MarlinMode> MarlinSNAR
     ) -> FS {
         let mut sponge = FS::new_with_parameters(fs_parameters);
         sponge.absorb_bytes(&to_bytes_le![&Self::PROTOCOL_NAME].unwrap());
-        sponge.absorb_bytes(&batch_size.to_le_bytes());
+        sponge.absorb_bytes(&(batch_size as u64).to_le_bytes());
```
howardwu: @Pratyush is the batch size unsafe cast safe here?

utilities/src/serialize/impls.rs
```diff
@@ -148,7 +148,46 @@ impl_canonical_serialization_uint!(u8);
 impl_canonical_serialization_uint!(u16);
 impl_canonical_serialization_uint!(u32);
 impl_canonical_serialization_uint!(u64);
-impl_canonical_serialization_uint!(usize);
+
+impl CanonicalSerialize for usize {
```
howardwu, line 152: @Pratyush can you review this file?

algorithms/src/snark/marlin/marlin.rs
ljedrz: I don't see an issue here, it will not truncate

ljedrz (approved):
LGTM, though the best solution would probably be to not use `usize` at all in the first place (i.e. to not even have to cast it to `u64`)

algorithms/src/snark/marlin/marlin.rs
Pratyush: Agreed, this is fine

Pratyush: Though I would maybe use try_from to catch any edge cases.

ljedrz: are any possible? I don't think so, unless we're thinking of some 128 bit architecture :thinking:?

utilities/src/serialize/impls.rs
```diff
@@ -148,7 +148,46 @@ impl_canonical_serialization_uint!(u8);
...
+    #[inline]
+    fn serialize_with_mode<W: Write>(&self, mut writer: W, _compress: Compress) -> Result<(), SerializationError> {
+        let u64_value = *self as u64;
```
Pratyush: I'm unwrapping, but maybe we can add a conversion error variant to the serialization enum and use that instead

```suggestion
        let u64_value = u64::try_from(*self).unwrap();
```

algorithms/src/snark/marlin/marlin.rs
Pratyush: I think CHERI is one such architecture (though there are no production machines with CHERI enabled yet, I think)

ljedrz: in that case it's fair to use `try_from` + `unwrap` as a futureproofing measure :+1:

utilities/src/serialize/impls.rs
ljedrz: I think an `unwrap` is fine here, as extending to higher-bit architectures is likely to require additional work targeting `usize` (i.e. we'd prefer to break here rather than miss some other spot)

algorithms/src/snark/marlin/marlin.rs
iamalwaysuncomfortable: Basically what happens in the `as` case is that it wraps so it's basically `usize%u64::MAX` (or whatever bit size you're working with). The tricky thing with usize is that sometimes address spaces and register sizes are different (like in wasm, registers are 64-bit but the address space is 32), so usizes can do some strange things overall. 

[Rust playground link](https://play.rust-lang.org/?version=stable&mode=debug&edition=2021&gist=551ac75776f196267b5b39e2b3195c00)

I believe this function runs on the main thread so it should crash the proving process allowing us to address it if anyone starts using it.

In general `usize` issues in cryptographic systems happen a lot given a lot of the convenience methods like `.len()` produce them. 

Ultimately the prudent thing to do would be to go through and replace them with `u64` but that will indeed involve casting earlier in the stack and we'll need to go through and carefully do this and consider how to handle the errors.

However, for a proving-in-wasm MVP, I'd like to get this change through to pipe it up to users and I think the `try_from` + `unwrap` is a satisfactory strategy for now. 

I will push through a `usize` --> `u64` PR at some point that finds all the locations in which this happens soon to help future proof us.

Pratyush (approved):

raychu86 (approved):

howardwu (approved):
LGTM

PR #1427, opened 2023-03-29
Introduce `StorageTree` for checkpointing mapping state

raychu86:
## Motivation

This PR introduces and implements 3 tree types for managing program state.

1. `StorageTree` 
   - Depth 32 
   - The leaves are the roots of each `ProgramTree`
   - This represents the state of the `finalize` mapping contexts for every program
   
2. `ProgramTree` 
   - Depth 10 
   - The leaves are the roots of each associated `MappingTree`
   - Represents the state of the `finalize` mapping contexts for a specific program

4. `MappingTree`
    - Depth 32 (Looking to optimize this. Maybe dynamic sized depths based on the mapping key type)
    - The leaves are the `ValueID`s for the particular mapping.
    - Represents the state of the `finalize` mapping contexts for a specific mapping in a specific program.

On initialization of the `ProgramStorage`, a new `StorageTree` will be initialized from scratch based on the stored key-value pairs. This tree will be kept in memory, and updated as new state transitions come along, similar to the current `BlockTree` model. 

Only the `StorageTree` is kept in memory because keeping every `ProgramTree` and `MappingTree` in memory is not feasibly as the on-chain state grows. This means that in order to update the `StorageTree`, each `ProgramTree` and `MappingTree` affected by the state transition will need to be resampled. 

Note: This model can be optimized by serializing the `ProgramTree`s and `MappingTree`s directly into storage, which will speed up the `StorageTree` updates, however will incur double the state-space growth because you'll need to store the intermediate/parent nodes of the tree.

## Test Plan

The existing `insert`, `remove`, `initialize`, and `update` tests now check that the merkle trees are updated properly.

## Related PRs

This PR was built off of #1398.


console/collections/src/merkle_tree/mod.rs
```diff
@@ -257,18 +271,28 @@ impl<E: Environment, LH: LeafHash<Hash = PH::Hash>, PH: PathHash<Hash = Field<E>
         }
         lap!(timer, "Hashed {} padding levels", padding_depth);
 
-        // Update the root hash.
-        self.root = root_hash;
+        // Initialize the Merkle tree.
+        let mut tree = Vec::with_capacity(self.tree.len());
+        // Extend the new Merkle tree with the existing leaf hashes.
+        tree.extend(&self.tree);
```
howardwu, line 277: @ljedrz is `.clone()` faster or is the current logic better?

console/collections/src/merkle_tree/mod.rs
```diff
@@ -213,6 +213,20 @@ impl<E: Environment, LH: LeafHash<Hash = PH::Hash>, PH: PathHash<Hash = Field<E>
     pub fn update(&mut self, leaf_index: usize, new_leaf: &LH::Leaf) -> Result<()> {
         let timer = timer!("MerkleTree::update");
 
+        // Compute the updated Merkle tree with the new leaves.
+        let updated_tree = self.prepare_update(leaf_index, new_leaf)?;
+        // Update the tree at the very end, so the original tree is not altered in case of failure.
+        *self = updated_tree;
```
howardwu, line 219: @ljedrz can you check efficiency and safety here? We want this to first build a candidate tree, fail if it needs to, then update in place if it is successful.

console/collections/src/merkle_tree/mod.rs
ljedrz: There should be little difference performance-wise; a `clone` would be simpler, the capacity might just differ slightly.

console/collections/src/merkle_tree/mod.rs
ljedrz: It's fine in terms of avoiding mutation in case anything goes wrong and only replacing the original object in case of success.
As for efficiency, technically the cloning of the original tree doesn't need to happen - note that the operations starting from the `// Initialize the Merkle tree.` comment are all infallible, so at that point the original tree might as well be updated in place; since that would require a mutable borrow on `self`, the function would have exclusive access to the tree, so there couldn't be any outside reads happening in the meantime.

howardwu:
Closing in favor of #1445

PR #1432, opened 2023-03-31
[Feature] Allow multiple record definitions.

d0cd:
This PR removes restrictions on the number of records that can be defined in a program.
The current implementation allows users to define up to two records.


howardwu (approved):
LGTM

PR #1433, opened 2023-03-31
[Fix] Check for well-formed booleans.

d0cd:
This PR fixes a check that a boolean - encoded as a linear combination - is well-formed.


howardwu (approved):
LGTM

PR #1436, opened 2023-04-02
[Design] Renames `load` and `store` instructions to `get` and `put` respectively.

d0cd:
This PR, renames `load` and `store` instructions to `get` and `put` respectively.
The motivation is to reserve `load` and `store`for array operations.

Overloading loads/stores for array and mapping operations can minimize the instructions (but not by much).
However, this comes at the expense of clarity. 
A user is forced to disambiguate between a load/store on a mapping and one on an array.
Furthermore, loads/stores on mappings are not allowed in the off-chain component of the program. This requires a context-dependent check on each use of load/store.

### DIFF
--- a/synthesizer/src/process/mod.rs
+++ b/synthesizer/src/process/mod.rs
@@ -1172,7 +1172,7 @@ function transfer:
     }
 
     #[test]
-    fn test_process_execute_and_finalize_get_add_set() {
+    fn test_process_execute_and_finalize_get_add_put() {
         // Initialize a new program.
         let (string, program) = Program::<CurrentNetwork>::parse(
             r"
@@ -1194,7 +1194,7 @@ finalize compute:
     input r1 as u64.public;
     get_or account[r0] 0u64 into r2;
     add r2 r1 into r3;
-    set r3 into account[r0];
+    put r3 into account[r0];
 ",
         )
         .unwrap();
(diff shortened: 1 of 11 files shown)


howardwu:
I think `load` and `store` are far more clear than `get` and `put`. It's less of an adoption curve to unify several operators for loading and storing values under a single operand name, with subvariants. My recommendation is to keep the current design.

d0cd:
In considering a learning/adoption curve, we should also consider a developer’s background knowledge. 
Many languages have map-like data types with interfaces that are more consistent with `get` and `put`.
 In [Javascript](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Map), you have `get` and `set`.
  In [Rust](https://doc.rust-lang.org/std/collections/struct.HashMap.html), you have `get` and `insert`.
   In [Python](https://docs.python.org/3/library/stdtypes.html#typesmapping), you have `get` and `update`. 
   These map-like data types behave similarly to our own and using a familiar syntax can be conceptually easier on developers.  
   
   Syntactic differences can also help developers understand that arrays and mappings are distinct data types with different contexts.
    Arrays operations should be supported in off-chain and on-chain contexts while mapping operations are only supported on-chain.  
    While subvariants accomplish this, e.g `load.mapping` and `load.array`, they are not as syntactically distinct as `get` and `load`.

d0cd:
Closing in favor of #1460.

PR #1437, opened 2023-04-03
[Design] Use canonical identifiers for opcodes.

d0cd:
Currently, our opcodes use "dot" syntax to denote variants, e.g `add.w`.
This PR, renames opcodes such that each opcode is a canonical identifier (`[a-zA-Z][_a-zA-Z0-9]*`).
This simplifies the formal grammar, making it easier to support inline assembly in Leo.
While this changes the syntax for Aleo instructions, the difference between using `.` and `_` is minimal. We can also help users migrate existing programs with a script.

### DIFF
--- a/synthesizer/src/process/mod.rs
+++ b/synthesizer/src/process/mod.rs
@@ -930,7 +930,7 @@ closure execute:
 closure check_not_equal:
     input r0 as field;
     input r1 as field;
-    assert.neq r0 r1;
+    assert_neq r0 r1;
 
 function compute:
     input r0 as field.private;
@@ -1056,7 +1056,7 @@ function produce_magic_number:
 
 function check_magic_number:
     input r0 as u64.private;
-    assert.eq r0 1234u64;
+    assert_eq r0 1234u64;
 
 function noop:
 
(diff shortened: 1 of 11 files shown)


acoglio (approved):

howardwu:
I think this approach is far less clear to the reader. Consider the ambiguous cases that a `.` would help to clarify:
```
verify_merkle_tree vs. verify.merkle_tree
verify_inclusion_proof vs. verify_inclusion.proof vs. verify.inclusion_proof
verify_inclusion_path vs. verify_inclusion.path vs. verify.inclusion_path
```
My suggestion is to keep with our current syntax.

d0cd:
That's a fair point!
AFAIK, Aleo instructions does not have the above ambiguities. 
In other words, `.`s can be replaced with `_`s in opcodes without any ambiguities.  

Moreover, introducing such cases would actually lead to a significant source of confusion for developers.
 For example, suppose you have two opcodes `verify_inclusion.proof` and `verify.inclusion_proof`. 
  If these opcodes have different semantics, a mistyped `.` results in a completely different program.
  It should be the case that any new opcode is distinct enough to minimize confusion with existing opcodes, thus mitigating the risk of ambiguities.
  
An added bonus of proceeding with the `_` syntax is that Leo parser for inline assembly becomes significantly more simple.

PR #1440, opened 2023-04-04
Use the atomic_write_batch macro more

ljedrz:
While looking into memory use, I noticed that there are a few spots where the contents of the `atomic_write_batch` macro are used, as opposed to just calling the macro. Unless there's some specific reason for it that I missed, this PR is a readability and maintainability improvement.


howardwu (approved):
LGTM

PR #1441, opened 2023-04-05
Enable Insertion of Proving Keys from External Contexts

iamalwaysuncomfortable:
## Motivation

Webassembly does not allow blocking operations to be performed. Network calls can be performed via the `fetch` operation but currently only on javascript microtasks which wait for the event loop to complete before making the network calls resulting in parameter fetching happening after webassembly code completes.  

Addtionally, asynchronous network calls aren't easily possible given Aleo's fetching of key material happens almost entirely in a synchronous context.

This PR add the ability for key material to be fetched from an external context and inserted in Aleo's inclusion proof flow.

This approach was initially researched by @evanmarshall and several other developers at Demox Labs. It includes some modification of approach to stay with the existing query model and make the code more concise.

## Test Plan
* Test to ensure executions & inclusion proving work with both externall inserted values & via downloads
* Test mock flow of proving within the web

### DIFF
--- a/synthesizer/src/process/stack/inclusion/mod.rs
+++ b/synthesizer/src/process/stack/inclusion/mod.rs
@@ -44,17 +44,17 @@ pub enum Query<N: Network, B: BlockStorage<N>> {
     /// Base URL of the node to fetch inclusion proof data
     REST(String),
     /// Inclusion proof data fetched from an outside environment
-    External(N::StateRoot, HashMap<Field<N>, StatePath<N>>),
+    LOCAL(N::StateRoot, HashMap<Field<N>, StatePath<N>>),
 }
 
 impl<N: Network, B: BlockStorage<N>> TryFrom<(&str, &str)> for Query<N, B> {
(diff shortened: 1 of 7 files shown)


synthesizer/src/process/stack/inclusion/mod.rs
```diff
@@ -671,13 +727,178 @@ mod tests {
...
+            state_root,
+            commitment_map,
+            credits_proving_key,
+            inclusion_proving_key,
+        )
+        .unwrap();
+
+        // ENTER WASM: Verify the execution
+        assert!(Inclusion::verify_execution(&execution, Some(inclusion_verifying_key)).is_ok());
+    }
```
iamalwaysuncomfortable, line 903: This is a sample test that mocks how it would work in a webassembly environment where namely:
* The program proving keys, inclusion keys are inserted into the proving process
* In a webassembly context, the data needed to make the inclusion proof is called from javascript and inserted back into another function.

PR #1442, opened 2023-04-06
[Feature] Implement CLI commands part one

collinc97:
## Completed in this PR
- [x] new - creates a new aleo project
- [x] build - compiles an aleo program
- [x] run - executes an aleo program locally
- [x] clean - cleans the aleo package build directory
- [x] update - updates snarkvm to the latest version

## Not Completed
account
deploy
execute
transfer


iamalwaysuncomfortable (commented):
Is it possible that we create methods wrapping the subroutines within these commands so that they can be exported in a library?

I think Package could likely also be moved up to a different crate. Perhaps there can be an extra `develop` crate within SnarkVM that exports development tools?

vm/cli/commands/run.rs
```diff
@@ -0,0 +1,113 @@
...
+
+impl Run {
+    /// Compiles an Aleo program function with the specified name.
+    #[allow(clippy::format_in_format_args)]
+    pub fn parse(self) -> Result<String> {
+        // Derive the program directory path.
+        let path = std::env::current_dir()?;
+
+        // Load the package.
+        let package = Package::open(&path)?;
```
iamalwaysuncomfortable, line 46: I think the design of the package's implementation of `program.json` is currently flawed. It has a `PrivateKey` in the .json file and people are using are CLIs in production with this private key sitting in plaintext to build custodial solutions which is a security nightmare. 

A plaintext `PrivateKey` in the `program.json` should be removed in favor of a more secure solution before any public release of this CLI is made.

iamalwaysuncomfortable (commented):
Additional comment about the Package

vm/cli/commands/run.rs
howardwu: This private key is only meant to be used for development (as its name implies), it shouldn't be for production. I agree we should patch this, let's prioritize this during Phase 3 (read: before mainnet).

howardwu (approved):
LGTM

PR #1444, opened 2023-04-07
Remove `gates` from `record`

howardwu:
## Motivation

This PR includes the following changes:
- Removes `gates` from each `record`
- Renames `gates` to `microcredits` in the `credits.aleo` program
- Updates the input `address` and `u64` in `credits.aleo/mint` to be public (for auditability)


howardwu:
This PR is ready for review. There are some minor tweaks (10-20 LoC) I'll add today but can start reviewing.

synthesizer/src/vm/mod.rs
```diff
@@ -357,32 +350,22 @@ function compute:
                 // Update the VM.
                 vm.add_next_block(&genesis).unwrap();
 
+                // Prepare the inputs.
+                let inputs = [
+                    Value::<CurrentNetwork>::from_str(&address.to_string()).unwrap(),
+                    Value::<CurrentNetwork>::from_str("1u64").unwrap(),
+                ]
+                .into_iter();
```
iamalwaysuncomfortable, line 358: It'd be an eventual nice re-work the Into<ExactSizedIterator> trait in a way that keeps those desired properties but allows the Transaction methods to take arrays or Vecs directly

synthesizer/src/program/mod.rs
```diff
@@ -127,23 +127,24 @@ function transfer:
...
     cast r0.owner r1 into r3 as credits.record;
     cast r0.owner r2 into r4 as credits.record;
     output r3 as credits.record;
     output r4 as credits.record;
 
 function fee:
     input r0 as credits.record;
-    input r1 as u64.private;
-    sub r0.gates r1 into r2;
+    input r1 as u64.public;
```
iamalwaysuncomfortable, line 145: Just for explanation sake - why does this need to be public now?

synthesizer/src/block/transition/serialize.rs
```diff
@@ -74,8 +73,6 @@ impl<'de, N: Network> Deserialize<'de> for Transition<N> {
                     DeserializeExt::take_from_value::<D>(&mut transition, "tpk")?,
                     // Retrieve the `tcm`.
                     DeserializeExt::take_from_value::<D>(&mut transition, "tcm")?,
-                    // Retrieve the fee.
-                    DeserializeExt::take_from_value::<D>(&mut transition, "fee")?,
```
iamalwaysuncomfortable, line 78: Where are fees & credit spends tracked now?

synthesizer/src/block/transaction/mod.rs
```diff
@@ -84,69 +84,91 @@ impl<N: Network> Transaction<N> {
...
-    /// Initializes a new execution transaction from an authorization, and an optional fee.
-    pub fn execute_authorization_with_additional_fee<C: ConsensusStorage<N>, R: Rng + CryptoRng>(
+    /// Initializes a new fee.
+    pub fn execute_fee<C: ConsensusStorage<N>, R: Rng + CryptoRng>(
         vm: &VM<N, C>,
         private_key: &PrivateKey<N>,
-        authorization: Authorization<N>,
-        additional_fee: Option<(Record<N, Plaintext<N>>, u64)>,
+        credits: Record<N, Plaintext<N>>,
+        fee_in_microcredits: u64,
```
iamalwaysuncomfortable, line 129: For clarity can it be explained how credits are going to be associated with records in general? I think the explanation for this should probably go in the top level documentation for the method.

Also naming-wise, this method is a little opaque on first look. I believe this is saying "the credits associated with this record are being used to pay the fee" 

Thus might be something like the following:
```suggestion
        unspent_funds: Record<N, Plaintext<N>>,
        fee_in_microcredits: u64,
```

synthesizer/src/block/transaction/mod.rs
```diff
@@ -84,69 +84,91 @@ impl<N: Network> Transaction<N> {
...
     }
 
-    /// Initializes a new execution transaction from an authorization, and an optional fee.
-    pub fn execute_authorization<C: ConsensusStorage<N>, R: Rng + CryptoRng>(
+    /// Initializes a new execution transaction.
+    pub fn execute<C: ConsensusStorage<N>, R: Rng + CryptoRng>(
         vm: &VM<N, C>,
-        authorization: Authorization<N>,
+        private_key: &PrivateKey<N>,
+        (program_id, function_name): (impl TryInto<ProgramID<N>>, impl TryInto<Identifier<N>>),
```
iamalwaysuncomfortable, line 103: Seems like this maybe was done to emphasize the grouping (and I didn't even know you could group parameter names like this 😹), but I notice it does require you have to supply these as a tuple which is feels a little ergonomically odd and might look off in Rustdoc, maybe we can change back to a list of params?

```suggestion
        program_id, function_name: impl TryInto<ProgramID<N>>
        function_name: impl TryInto<Identifier<N>>),
```

iamalwaysuncomfortable (commented):
This seems good (although I think we should all do one more pass in ensure all validations are in order) - I put in a few minor nits talking about the ergonomics of this going forward.

There are some considerations I'd like to announce here though:
1.  It's unclear looking at this on first pass exactly how credits are now associated with records. I remember we had talked about how this is done, but could it be explained here or if there's a branch or PR implementing it, could it be linked?
2. There likely should be some kind of helper functionality which allows people to look up the amount credits associated with records they own for both accounting and pre-emptive error handling (i.e. a method author should be able to check the balance to see if it's enough to perform given tasks)
3. The top level `record`, `execution` and `fee` method documentation should include some indication of how to associate balances with records

And a final question - with the new method of tracking credit transfers - does this currently mean there would now be a global ledger of value transfers?

synthesizer/src/program/mod.rs
howardwu: The rationale is that this is now the only way we can actually see the fee.

synthesizer/src/block/transition/serialize.rs
howardwu: They are tracked in the `Transaction` map.

synthesizer/src/block/transaction/mod.rs
howardwu: The intention is to remove the clippy complaint for too many arguments. Otherwise we need to add a clippy bypass lint onto the function.

howardwu: This convention is commonly used in the `algorithms` crate

synthesizer/src/vm/mod.rs
howardwu: The original request was from Rust folks to use iterators to prevent accumulating the cost of calling `.collect()`.

raychu86 (approved):

PR #1445, opened 2023-04-08
Implement `finalize`

raychu86:
This PR introduces the concept of `finalize`, which will perform speculative execution of `finalize` scopes in transactions. This will allow us to filter out transactions that will cause double spends and/or general operations that will invalidate on-chain state. 

Speculate iteratively executes transactions so we can determine a final ordering of transactions for a block. This is done by cache-ing the operational transforms and intermediate state changes, s.t. the `FinalizeStore` doesn't see changes until we are ready to commit to a final transaction ordering.

TODO :
- [x] Handle instruction panics (Currently instruction panic instead of being caught as an error, so speculate can't catch them)
- [x] Integrate `finalize_root` changes from #1162 
- [ ] Consider transaction fees during speculative executions. (Can be ignored now because `credits.aleo` doesn't have mappings yet)
- [x] Add additional tests that handle multiple programs that touch on-chain state.

## Test Plan

Tests have been added to ensure that finalize operations are correct and that `finalize` doesn't break the state changes that occur in `FinalizeStore`.

## Related PRs

This PR is build on (and contains all the changes of):
 - #1190 
 - #1427


synthesizer/src/store/helpers/rocksdb/internal/map.rs
```diff
@@ -204,32 +206,39 @@ impl<
...
         Q: PartialEq + Eq + Hash + Serialize + ?Sized,
     {
         if self.batch_in_progress.load(Ordering::Acquire) { self.atomic_batch.lock().get(key).cloned() } else { None }
     }
 
+    ///
+    /// Returns an iterator visiting each key-value pair in the atomic batch.
+    ///
+    fn iter_pending(&'a self) -> Self::PendingIterator {
+        self.atomic_batch.lock().clone().into_iter().map(|(k, v)| (Cow::Owned(k), v.map(|v| Cow::Owned(v))))
```
howardwu, line 275: @ljedrz is there a more efficient way to do this?

howardwu (approved):
LGTM

PR #1449, opened 2023-04-11
Speed up the happy path for MapRead::get_speculative

ljedrz:
Since the preference for results in `MapRead::get_speculative` has changed, we should perform storage accesses only if the atomic write batch lookup is a miss.

This will provide us with a considerable speedup for batch hits, and cause no difference for misses.


synthesizer/src/store/helpers/mod.rs
```diff
@@ -111,17 +111,14 @@ pub trait MapRead<
         K: Borrow<Q>,
         Q: PartialEq + Eq + Hash + Serialize + ?Sized,
     {
-        // Return early in case of errors in order to not conceal them.
```
howardwu: The comment states to return early.

ljedrz: Yes, which was significant in the older logic, where the map value was the one to use if there was one (instead of ignoring the error in case there was a batched one available); however, now that the batch value is the one to be returned if there is one, there is no need to look up the map value in all the calls.

howardwu: Understood, that makes sense. Let me review Ray's PR on finalize (which uses this code heavily) first, and assuming all is good, I'll come back and get this landed.

howardwu:
@ljedrz I think it's appropriate for us to rename `get_speculative` as `get`.
1. `get_speculative` is more all-encompassing than `get` now
2. we should be deferring to using `get_speculative` so that we do not accidentally forget to call it

howardwu (approved):

PR #1450, opened 2023-04-11
Catch instruction panics for `Speculate`

raychu86:
## Motivation

Built on #1445.

This PR catches instruction panics caused by `O::evaluate` in `Literals::evaluate` and returns an error. This is used to catch the `E::halt` panic calls from Literal operations - example [here](https://github.com/AleoHQ/snarkVM/blob/8695df3717a54cd8f7b29989171f005da0342eab/console/types/integers/src/arithmetic.rs#L72). 

This catch is required because `Speculate` needs to determine if a transaction execution will fail without dropping the entire process with a panic. 

This is a preliminary solution to the `panic` problem, however I am unsure if the `catch_unwind` approach is safe for production.

## Test Plan

Running `test_speculate_many` will run into the cases where panic handling is necessary.

### DIFF
--- a/console/network/src/lib.rs
+++ b/console/network/src/lib.rs
@@ -47,7 +47,10 @@ use snarkvm_curves::PairingEngine;
 
 use indexmap::IndexMap;
 use once_cell::sync::OnceCell;
-use std::sync::Arc;
+use std::{
+    panic::{RefUnwindSafe, UnwindSafe},
+    sync::Arc,
+};
 
 /// A helper type for the BHP Merkle tree.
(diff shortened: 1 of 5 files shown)


raychu86:
Closed in favor of spawning an individual thread to catch the panic - e.g.
```
        // Compute the operation.
        let inputs: [Literal<N>; NUM_OPERANDS] =
            inputs.try_into().map_err(|_| anyhow!("Failed to prepare operands in evaluate"))?;
        let operation = format!("{} {inputs:?}", O::OPCODE);
        let output: Literal<N> = std::thread::spawn(move || O::evaluate(&inputs))
            .join()
            .map_err(|_| anyhow!("Failed to evaluate operation: {operation}"))??;
```

PR #1459, opened 2023-04-14
Introduce index for program deployments via `ProgramIndexMap`

raychu86:
## Motivation

This PR introduces an index for programs. Each programID will be associated with an index (which is determined by the order of deployment). 

This is necessary for deterministic construction of the `StorageTree` when we need to checkpoint on-chain state. Introducing this preemptively, to prevent database issues when `StorageTrees` are released to the main network


howardwu (approved):
LGTM

PR #1462, opened 2023-04-14
Updates the number of functions (31), number of transitions (31), transition inputs (16), and transition outputs (16)

howardwu:
## Motivation

Updates the number of functions, number of transitions, transition inputs, and transition outputs.
- Supports 31 functions (up from 15)
- Supports 31 transitions (up from 15)
- Supports 16 transition inputs (up from 8)
- Supports 16 transition outputs (up from 8)


raychu86 (approved):
lgtm

PR #1453, opened 2023-04-13
Fix a few imports

ljedrz:
I encountered this when running some tests. With these tweaks in place, one can run `cargo check --release --workspace --all-targets` for the whole repo without errors or warnings.

PR #1463, opened 2023-04-14
Add `cargo check` in CI

raychu86:
## Motivation

This PR adds a `cargo check` CI job (This should fail until we get #1453 in)

Additionally, the rust version in CI was bumped from 1.67 to 1.68.


howardwu (approved):
LGTM

PR #1466, opened 2023-04-15
Use a BTreeMap instead of IndexMap in MemoryMap

ljedrz:
This purpose of this PR is to make the test conditions in snarkVM more aligned with those typical for snarkOS.

As stated in the newly added comment:
> The reason for using BTreeMap with binary keys is for the order of items to be the same as the one in the RocksDB-backed DataMap in snarkOS; if not for that, it could be any map with fast lookups and the keys could be typed (i.e. just `K` instead of `Vec<u8>`).

### DIFF
--- a/synthesizer/src/store/helpers/memory_map.rs
+++ b/synthesizer/src/store/helpers/memory_map.rs
@@ -19,10 +19,10 @@ use console::network::prelude::*;
 use indexmap::IndexMap;
 
 use core::{borrow::Borrow, hash::Hash};
-use indexmap::map;
 use parking_lot::{Mutex, RwLock};
 use std::{
     borrow::Cow,
+    collections::{btree_map, BTreeMap},
     sync::{
         atomic::{AtomicBool, Ordering},
         Arc,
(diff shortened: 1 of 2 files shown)


synthesizer/src/store/helpers/memory_map.rs
```diff
@@ -34,7 +34,10 @@ pub struct MemoryMap<
     K: Copy + Clone + PartialEq + Eq + Hash + Serialize + for<'de> Deserialize<'de> + Send + Sync,
     V: Clone + PartialEq + Eq + Serialize + for<'de> Deserialize<'de> + Send + Sync,
 > {
-    map: Arc<RwLock<IndexMap<K, V>>>,
+    // The reason for using BTreeMap with binary keys is for the order of items to be the same as
+    // the one in the RocksDB-backed DataMap in snarkOS; if not for that, it could be any map
```
howardwu:
```suggestion
    // the one in the RocksDB-backed DataMap; if not for that, it could be any map
```

howardwu (approved):
LGTM

PR #1467, opened 2023-04-15
[Design] Remove `output` statements from `finalize` blocks.

d0cd:
This PR, removes `output` statements from `finalize` blocks. 
Currently, finalize outputs are not stored, this change makes the behavior explicit.
Users can store results in mappings and query them directly.


howardwu (approved):
LGTM

PR #1471, opened 2023-04-16
Add `finalize_root` to block `Header` 

raychu86:
## Motivation

This PR adds the `finalize_root` (checkpointing of the on-chain finalize state) to each block header.

The genesis `finalize_root` is stored as Field::zero().


synthesizer/src/block/header/genesis.rs
```diff
@@ -53,7 +57,7 @@ mod tests {
     /// Update this method if the contents of a block header have changed.
     fn get_expected_size<N: Network>() -> usize {
         // Previous state root, transactions root, and accumulator point size.
```
howardwu:
```suggestion
        // Previous state root, transactions root, finalize root, and accumulator point size.
```

synthesizer/src/block/header/serialize.rs
```diff
@@ -26,6 +26,7 @@ impl<N: Network> Serialize for Header<N> {
                 let mut header = serializer.serialize_struct("Header", 4)?;
```
howardwu:
```suggestion
                let mut header = serializer.serialize_struct("Header", 5)?;
```

howardwu (approved):
LGTM

PR #1481, opened 2023-04-18
[Test] Check that the number of mappings does not exceed maximum.

d0cd:
This PR,
- adds tests checking that the number of mappings in a program does not exceed the maximum.
- enforces the check in `Program::add_mapping`


d0cd:
Closing since `speculate-tx` has been redesigned.

PR #1482, opened 2023-04-19
Update `feat/client` to v0.10.2

raychu86:
## Motivation

Update to match new v0.10.2 syntax and apis.


howardwu:
What happened here?

raychu86:
> What happened here?

If you mean the change, the url sanitization prevents the url from having double slashes (which invalidates the url).

We don't run into this issue when setting the url to `https://vm.aleo.org/api` because of how `reqwest::Url` handles the final `/` their `to_string` method.


For example:
```
reqwest::Url::parse("https://vm.aleo.org/api").unwrap().to_string() => "https://vm.aleo.org/api"

reqwest::Url::parse("http://localhost:3000").unwrap().to_string() => "http://localhost:3000/"
```

The second example would produce something like this - `http:/localhost:3000//testnet3/latest/height` when trying to query.

PR #1486, opened 2023-04-24
Remove fee check for `credits.aleo/split` calls

raychu86:
## Motivation

This PR removes the fee requirement from `credits.aleo/split` calls. The intention for this is so that users can start creating transactions without having to initially own 2+ records. 

This rule will be removed once batched executions are introduced, because fees can be baked into the Tx.


howardwu (approved):
LGTM

PR #1487, opened 2023-04-25
Wasm parameter downloads

iamalwaysuncomfortable:
## Motivation

When compiling Rust into WebAssembly meant to be used within a JavaScript context, the execution model differs from a `std` environment. It ends up as WebAssembly executing in a single thread (unless web workers are explicitly used) sharing the `Javascript` `event loop` with Javascript code. Key synthesis and function execution within Aleo is done within a largely synchronous codebase, and when compiled into WebAssembly and called from javascript, this blocks the entire event loop during the execution of key synthesis.

The [following lines](https://github.com/AleoHQ/snarkVM/blob/testnet3/parameters/src/macros.rs#L92-L109) (macros.rs 92-109) were the previous approach used to download the parameters within a `wasm` environment. It attempted to spawn a local async task to run alongside the synchronous that would download the parameters and store them within a buffer passed into the function.
```rust
        #[cfg(feature = "wasm")]
        fn remote_fetch(
            buffer: alloc::sync::Weak<parking_lot::RwLock<Vec<u8>>>,
            url: &'static str,
        ) -> Result<(), $crate::errors::ParameterError> {
            // NOTE(julesdesmit): We spawn a local thread here in order to be
            // able to accommodate the async syntax from reqwest.
            wasm_bindgen_futures::spawn_local(async move {
                let content = reqwest::get(url).await.unwrap().text().await.unwrap();

                let buffer = buffer.upgrade().unwrap();
                buffer.write().extend_from_slice(content.as_bytes());
                drop(buffer);
            });
            Ok(())
        }
    };
}
```

Unfortunately, this fails because the `spawn_local` function schedules a task on the event loop which can only be executed when the key synthesis process has finished. This means that http call won't run until the synthesis function yields back to the event loop, and thus it cannot download the parameters needed for key synthesis with the normal key synthesis flow. 

This leads to a situation where alternative approaches are needed.

The following approaches are viable:
1. Pass in `proving` & `verifying` keys externally obtained from a javascript (or other) environment. This is possible, but presents some privacy/security considerations and we did not want this to be the default way of doing web-based proving. It is however, something we should potentially consider allowing in the future (with a checksum and maybe proofs of key validity perhaps) to allow developers flexibility in creating caching strategies for their apps.
2. Create a `wasm` specific `asynchronous` flow. The idea here would be to create wasm-specific functions which break up the existing key synthesis and function execution functions to `yield` at specific points so that the `event loop` can schedule an `http call` and then continue execution once it completes. This is a potentially viable approach in which we'd be able to do the same level of encapsulation, but likely to be more messy/error-prone.
3. Use existing block browser apis. Modern browsers purposely discourage synchronous calls in the browser to discourage un-performant web design, however a few still exist (even if they have a downgraded feature set) and enable us to keep existing proof flow.

Option #3 is the considered approach here. `XmlHttpRequest` via `web_sys` is introduced in order to make the necessary blocking web calls. Although these will block web pages, the ordinary web proving process will also do so, and so this design is done in the hope that web workers are to be used when possible (which should be almost always given most browsers now support them). 

It is worth of note here that the approach will not work in `nodejs` as designed - for `nodejs` alternative approaches will tested. This likely will require us to introduce a `web` feature on the parameters crate to delineate between browser & nodejs usage.

### Steps to mark this PR as ready
This PR is in draft. Before this PR is marked ready the steps in the test plan below will be followed.

## Test Plan
1. Test a sample function execution to ensure that this that this works in major browsers and provide a working demo online that can be accessed by several browsers
2. Introduce a nodejs pathway into this PR and fully test it.

## Related PRs
The PR in which we attempted to allow parameter injection is here. It's encouraged to reference this to compare and contrast differences in the approach laid out within this PR: https://github.com/AleoHQ/snarkVM/pull/1441


evanmarshall:
@iamalwaysuncomfortable Thank you for putting in this work.

Option 1 is my preferred path. I think we should have a conversation with @howardwu & @raychu86 before committing too much to this path. If we don't go down the road of a stateless SDK, our wallet will have to maintain a fork indefinitely. This seems like poor design as the wallet should be a primary consumer of the SDK.

The key problems are:
1. Download file source? By forcing it to be the Aleo maintained endpoints, this increases centralization & risk. Anyone running the SDK should get to choose where their parameter files are stored.
2. Tight API coupling. Anyone should be able to implement their own API and integrate it with the SDK. This approach leads to an unnecessary tight coupling of the API and the SDK when making requests for things like state paths for inclusion proofs.
3. Storage/caching for browser environments? This road seems to lead to a ecosystem where each website / DApp downloads and stores 1GB. 
4. Using outdated / deprecated APIs to make fast rapidly browser APIs work within WASM. 

Any security concerns can sufficiently addressed as the integrity of digitally transmitted files is a pretty well studied and solved problem.

Honestly, I'm not sure what the hesitation is and I'd appreciate at least some explanation. That being said, we will maintain a fork until we can get this SDK to be completely stateless.

parameters/src/testnet3/mod.rs
```diff
@@ -20,7 +20,7 @@ pub use genesis::*;
 pub mod powers;
 pub use powers::*;
 
-const REMOTE_URL: &str = "https://testnet3.parameters.aleo.org";
+const REMOTE_URL: &str = "https://aleotestcors.s3.us-east-2.amazonaws.com";
```
iamalwaysuncomfortable: Note: This is only temporary to enforce CORS. Although the original bucket has CORS we need to enforce it also at the CDN termination as well

synthesizer/src/process/mod.rs
```diff
@@ -149,6 +149,26 @@ impl<N: Network> Process<N> {
         Ok(process)
     }
 
+    /// Initializes a new process without downloading the credits.aleo proving keys for web contexts
+    #[inline]
+    #[cfg(feature = "wasm")]
+    pub fn load_web() -> Result<Self> {
```
iamalwaysuncomfortable, line 165: Not *strictly* necessary but highly desired. If we desire to do key synthesis in the web resources & speed are at a premium and we can't rely on the same resources or file based caching as we do in a `std` environment. So in the cases we're deploying or executing a program not related to credits, we should avoid spending the browser resources on it.

parameters/src/macros.rs
```diff
@@ -90,20 +90,62 @@ macro_rules! impl_store_and_remote_fetch {
...
-            // able to accommodate the async syntax from reqwest.
-            wasm_bindgen_futures::spawn_local(async move {
-                let content = reqwest::get(url).await.unwrap().text().await.unwrap();
-
-                let buffer = buffer.upgrade().unwrap();
-                buffer.write().extend_from_slice(content.as_bytes());
-                drop(buffer);
-            });
-            Ok(())
+        fn remote_fetch(url: &str) -> Result<Vec<u8>, $crate::errors::ParameterError> {
```
iamalwaysuncomfortable, line 93: This is the core of this PR. There are other approaches (like breaking up synthesizer functions so we can yield back to the event loop) but XmlHttp has strong official support in most all modern browsers and IMHO is a strong choice.

iamalwaysuncomfortable (commented):
Some minor comments

iamalwaysuncomfortable:
> Option 1 is my preferred path. I think we should have a conversation with @howardwu & @raychu86 before committing too much to this path. If we don't go down the road of a stateless SDK, our wallet will have to maintain a fork indefinitely. This seems like poor design as the wallet should be a primary consumer of the SDK.

My opinion at the moment is that we can provide both options because you're right that application developers will want to manage their own caching strategies likely. We just need to be quite mindful where we'd be inserting pre-downloaded parameters or keys. I'd like to make another PR to address this specifically, but it would be good for us all to have a conversation about tradeoffs soon.

> Download file source? By forcing it to be the Aleo maintained endpoints, this increases centralization & risk. Anyone running the SDK should get to choose where their parameter files are stored.
Tight API coupling. Anyone should be able to implement their own API and integrate it with the SDK. This approach leads to an unnecessary tight coupling of the API and the SDK when making requests for things like state paths for inclusion proofs.

There's a security argument for making the Aleo defined API the strong default, and also a complexity argument (i.e. keeping the Aleo official API reduces complexity of having to think about where to store it for developers making lighter use of the system). However, there's also the argument for providing the option for custom storage options. How to do this safely and while preserving all the privacy goals is the point of discussion which we'll have soon.

synthesizer/src/process/stack/inclusion/wasm.rs
```diff
@@ -0,0 +1,192 @@
...
+// GNU General Public License for more details.
+
+// You should have received a copy of the GNU General Public License
+// along with the snarkVM library. If not, see <https://www.gnu.org/licenses/>.
+
+use super::*;
+
+impl<N: Network> Inclusion<N> {
+    /// Returns the inclusion assignments for the given execution.
+    pub async fn prepare_execution_wasm(
```
iamalwaysuncomfortable: If we would like enforce getting inclusion proof data from the web, we need an async version of these functions to allow the event loop to be able to create HTTP requests. There could be a more elegant way to do this perhaps (like an `asyncify` macro which conditionally makes specific functions async based on feature flags) instead of making new async functions, but this is at the very least, straightforward and doesn't modify existing code. Open to modifying the way this is done.

(Note: this will work both in node & browsers. This is partially due to the fact that the `reqwest` library is maintained by the maker of `wasm-bindgen` and thus they maintain the interface really nicely)

iamalwaysuncomfortable: The asyncify & awaitify idea would be like so, but it seems like trying too hard.

If we do go for making more things async for the browser though it might be a path we want to take

```rust
/// Create both a synchronous and asynchronous version of a function based on a feature flag
#[macro_export]
macro_rules! asyncify {
    ($feature:expr, $visibility:vis, $($function:tt)*) => {
        #[cfg(feature = $feature)]
        $visibility async $($function)*

        #[cfg(not(feature = $feature))]
        $visibility $($function)*
    };
}

/// Await or don't await an expression based on a feature flag
#[macro_export]
macro_rules! awaitify {
    ($feature:expr, $expr:expr) => {
        #[cfg(feature = $feature)]
        {
            $expr.await
        }

        #[cfg(not(feature = $feature))]
        {
            $expr
        }
    };
}
```

And then used like so:

```
asyncify { "wasm", pub, fn prepare_execution_wasm( .. ) { .. } }
```

utilities/src/parallel.rs
```diff
@@ -67,18 +67,18 @@ pub fn max_available_threads() -> usize {
...
 
 #[inline(always)]
-#[cfg(feature = "serial")]
+#[cfg(any(feature = "serial", feature = "wasm"))]
 pub fn execute_with_max_available_threads<T>(f: impl FnOnce() -> T + Send) -> T {
     f()
 }
 
-#[cfg(not(feature = "serial"))]
+#[cfg(not(any(feature = "serial", feature = "wasm")))]
```
iamalwaysuncomfortable, line 81: Outside of wasm, `execute_with_threads` starts up an isolated thread pool to churn through all the rayon annotated paralleism (i.e. most parallel iterators). However within the `wasm` context using `wasm_bindgen_rayon` a global thread pool is spun up using webworkers and the feature flagged version of `execute_with_max_available_threads` above takes will run the function using the available global web-worker thread pool. So this flag isn't actually taking away parallelism in the case of the non-serial wasm flag.

parameters/Cargo.toml
```diff
@@ -50,6 +50,9 @@ version = "1.0"
 version = "2"
 optional = true
 
+[dependencies]
+encoding = "0.2"
```
howardwu: Reformat and make optional

parameters/src/testnet3/mod.rs
howardwu: Fix URL

howardwu (approved):
LGTM

PR #1494, opened 2023-05-01
Add test for multiple deployments in `Speculate`

raychu86:
## Motivation

This PR adds tests for deploying multiple programs in a single block and ensuring the speculative state is correct. 

This PR also adds a fix to how the program index is initialized via a new `MapRead` function that fetches the batched iter changes.


howardwu (approved):
LGTM

PR #1499, opened 2023-05-02
Catch usize conversion and Err more consistently

vicsn:
## Motivation

See for a motivation [this comment](https://github.com/AleoHQ/snarkVM/pull/1382/files#r1181041614).

However, my initial intuition was wrong, we can't reduce developer confusion with a "use `usize` as little as possible" because Rust's library uses `usize` all over the place. We instead increase clarity and safety by being more rigorous about our usage of usize. See the CONTRIBUTING doc for more details.

## Test Plan

Given that the main concern for this is consensus issues, I suggest to take extra care to stress test snark{VM,OS} on both 32-bit and 64-bit machines, generating test input from the different machine or constructing it manually. Let me know if you want me to help setting that up already, otherwise I'll make a TODO for once we start testing AleoBFT more extensively.

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -31,7 +31,9 @@ snarkVM is a big project, so (non-)adherence to best practices related to perfor
 - try to keep the sizes of `enum` variants uniform; use `Box<T>` on ones that are large
 
 ### Cross-platform consistency
-- types which contain consensus- or cryptographic logic should have a consistent size across platforms. Their serialized output should not contain `usize`, and at times it may be worth it to avoid using `usize` in the types themselves for clarity.
+- First and foremost, types which contain consensus- or cryptographic logic should have a consistent size across platforms. Their serialized output should not contain `usize`. For defense in depth, we serialize `usize` as `u64`.
+- For clarity, use `u32` and `u64` as much or long as possible, especially in type definitions. 
+- Given that we only target 32- and 64-bit systems, casting `usize` as `u64` and casting `u32` as `usize` will always be safe and doesn't need a `try_from::`. In serialization code, for defense in depth it is still encouraged to use `try_from::`.
 
 ### Misc. performance
 
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -30,6 +30,9 @@ snarkVM is a big project, so (non-)adherence to best practices related to perfor
 - if possible, reuse collections; an example would be a loop that needs a clean vector on each iteration: instead of creating and allocating it over and over, create it _before_ the loop and use `.clear()` on every iteration instead
 - try to keep the sizes of `enum` variants uniform; use `Box<T>` on ones that are large
 
+### Cross-platform consistency
+- types which contain consensus- or cryptographic logic should have a consistent size across platforms. Their serialized output should not contain `usize`, and at times it may be worth it to avoid using `usize` in the types themselves for clarity.
+
 ### Misc. performance
 
 - avoid the `format!()` macro; if it is used only to convert a single value to a `String`, use `.to_string()` instead, which is also available to all the implementors of `Display`
--- a/algorithms/src/crypto_hash/poseidon.rs
+++ b/algorithms/src/crypto_hash/poseidon.rs
@@ -385,6 +385,7 @@ impl<F: PrimeField, const RATE: usize> PoseidonSponge<F, RATE, 1> {
         optimization_type: OptimizationType,
     ) -> SmallVec<[F; 10]> {
         let params = get_params(TargetField::size_in_bits(), F::size_in_bits(), optimization_type);
+        assert!(params.bits_per_limb <= u32::MAX as usize);
 
         // Push the lower limbs first
         let mut limbs: SmallVec<[F; 10]> = SmallVec::new();
--- a/algorithms/src/crypto_hash/poseidon.rs
+++ b/algorithms/src/crypto_hash/poseidon.rs
@@ -322,7 +322,7 @@ impl<F: PrimeField, const RATE: usize> PoseidonSponge<F, RATE, 1> {
         let capacity = F::size_in_bits() - 1;
         let mut dest_limbs = Vec::<F>::new();
 
-        let params = get_params(TargetField::size_in_bits(), F::size_in_bits(), ty);
+        let params = get_params(TargetField::size_in_bits_u32(), F::size_in_bits_u32(), ty);
 
         let adjustment_factor_lookup_table = {
             let mut table = Vec::<F>::new();
@@ -342,9 +342,9 @@ impl<F: PrimeField, const RATE: usize> PoseidonSponge<F, RATE, 1> {
             let first = &src_limbs[i];
             let second = if i + 1 < src_len { Some(&src_limbs[i + 1]) } else { None };
 
-            let first_max_bits_per_limb = params.bits_per_limb + crate::overhead!(first.1 + F::one());
+            let first_max_bits_per_limb = params.bits_per_limb as usize + crate::overhead!(first.1 + F::one());
             let second_max_bits_per_limb = if let Some(second) = second {
-                params.bits_per_limb + crate::overhead!(second.1 + F::one())
+                params.bits_per_limb as usize + crate::overhead!(second.1 + F::one())
             } else {
                 0
             };
@@ -377,26 +377,24 @@ impl<F: PrimeField, const RATE: usize> PoseidonSponge<F, RATE, 1> {
         Self::get_limbs_representations_from_big_integer::<TargetField>(&elem.to_bigint(), optimization_type)
     }
 
-    // params.bits_per_limb is safe to cast as u32 because it will be less than `max_limb_size`
-    #[allow(clippy::cast_possible_truncation)]
     /// Obtain the limbs directly from a big int
     pub fn get_limbs_representations_from_big_integer<TargetField: PrimeField>(
         elem: &<TargetField as PrimeField>::BigInteger,
         optimization_type: OptimizationType,
     ) -> SmallVec<[F; 10]> {
-        let params = get_params(TargetField::size_in_bits(), F::size_in_bits(), optimization_type);
-        assert!(params.bits_per_limb <= u32::MAX as usize);
+        let params = get_params(TargetField::size_in_bits_u32(), F::size_in_bits_u32(), optimization_type);
 
         // Push the lower limbs first
         let mut limbs: SmallVec<[F; 10]> = SmallVec::new();
         let mut cur = *elem;
         for _ in 0..params.num_limbs {
             let cur_bits = cur.to_bits_be(); // `to_bits` is big endian
-            let cur_mod_r =
-                <F as PrimeField>::BigInteger::from_bits_be(&cur_bits[cur_bits.len() - params.bits_per_limb..])
-                    .unwrap(); // therefore, the lowest `bits_per_non_top_limb` bits is what we want.
+            let cur_mod_r = <F as PrimeField>::BigInteger::from_bits_be(
+                &cur_bits[cur_bits.len() - params.bits_per_limb as usize..],
+            )
+            .unwrap(); // therefore, the lowest `bits_per_non_top_limb` bits is what we want.
             limbs.push(F::from_bigint(cur_mod_r).unwrap());
-            cur.divn(params.bits_per_limb as u32);
+            cur.divn(params.bits_per_limb);
         }
 
         // then we reserve, so that the limbs are ``big limb first''
--- a/algorithms/src/crypto_hash/poseidon.rs
+++ b/algorithms/src/crypto_hash/poseidon.rs
@@ -377,6 +377,8 @@ impl<F: PrimeField, const RATE: usize> PoseidonSponge<F, RATE, 1> {
         Self::get_limbs_representations_from_big_integer::<TargetField>(&elem.to_bigint(), optimization_type)
     }
 
+    // params.bits_per_limb is safe to cast as u32 because it will be less than `max_limb_size`
+    #[allow(clippy::cast_possible_truncation)]
     /// Obtain the limbs directly from a big int
     pub fn get_limbs_representations_from_big_integer<TargetField: PrimeField>(
         elem: &<TargetField as PrimeField>::BigInteger,
--- a/algorithms/src/fft/polynomial/sparse.rs
+++ b/algorithms/src/fft/polynomial/sparse.rs
@@ -16,12 +16,11 @@
 
 use crate::fft::{EvaluationDomain, Evaluations, Polynomial};
 use snarkvm_fields::{Field, PrimeField};
-use snarkvm_utilities::serialize::*;
 
 use std::{collections::BTreeMap, fmt};
 
 /// Stores a sparse polynomial in coefficient form.
-#[derive(Clone, PartialEq, Eq, Hash, Default, CanonicalSerialize, CanonicalDeserialize)]
+#[derive(Clone, PartialEq, Eq, Hash, Default)]
 #[must_use]
 pub struct SparsePolynomial<F: Field> {
     /// The coefficient a_i of `x^i` is stored as (i, a_i) in `self.coeffs`.
--- a/algorithms/src/lib.rs
+++ b/algorithms/src/lib.rs
@@ -15,6 +15,7 @@
 #![allow(clippy::module_inception)]
 #![allow(clippy::type_complexity)]
 #![cfg_attr(test, allow(clippy::assertions_on_result_states))]
+#![warn(clippy::cast_possible_truncation)]
 
 #[cfg(feature = "wasm")]
 #[macro_use]
--- a/algorithms/src/msm/variable_base/batched.rs
+++ b/algorithms/src/msm/variable_base/batched.rs
@@ -170,16 +170,13 @@ fn batch_add_write<G: AffineCurve>(
 }
 
 #[inline]
-// It is safe to cast num_buckets to u32 because it was small enough earlier in the callchain
-#[allow(clippy::cast_possible_truncation)]
 pub(super) fn batch_add<G: AffineCurve>(
-    num_buckets: usize,
+    num_buckets: u32,
     bases: &[G],
     bucket_positions: &mut [BucketPosition],
-) -> Vec<G> {
+) -> Result<Vec<G>, anyhow::Error> {
     // assert_eq!(bases.len(), bucket_positions.len());
     assert!(!bases.is_empty());
-    assert!(num_buckets <= u32::MAX as usize);
 
     // Fetch the ideal batch size for the number of bases.
     let batch_size = batch_size(bases.len());
@@ -205,7 +202,7 @@ pub(super) fn batch_add<G: AffineCurve>(
             global_counter += 1;
             local_counter += 1;
         }
-        if current_bucket >= num_buckets as u32 {
+        if current_bucket >= num_buckets {
             local_counter = 1;
         } else if local_counter > 1 {
             // all ones is false if next len is not 1
@@ -219,13 +216,17 @@ pub(super) fn batch_add<G: AffineCurve>(
                     bucket_positions[global_counter - (local_counter - 1) + 2 * i].scalar_index,
                     bucket_positions[global_counter - (local_counter - 1) + 2 * i + 1].scalar_index,
                 ));
-                bucket_positions[new_scalar_length + i] =
-                    BucketPosition { bucket_index: current_bucket, scalar_index: (new_scalar_length + i) as u32 };
+                bucket_positions[new_scalar_length + i] = BucketPosition {
+                    bucket_index: current_bucket,
+                    scalar_index: u32::try_from(new_scalar_length + i)?,
+                };
             }
             if is_odd {
                 instr.push((bucket_positions[global_counter].scalar_index, !0u32));
-                bucket_positions[new_scalar_length + half] =
-                    BucketPosition { bucket_index: current_bucket, scalar_index: (new_scalar_length + half) as u32 };
+                bucket_positions[new_scalar_length + half] = BucketPosition {
+                    bucket_index: current_bucket,
+                    scalar_index: u32::try_from(new_scalar_length + half)?,
+                };
             }
             // Reset the local_counter and update state
             new_scalar_length += half + (local_counter % 2);
@@ -244,7 +245,7 @@ pub(super) fn batch_add<G: AffineCurve>(
         } else {
             instr.push((bucket_positions[global_counter].scalar_index, !0u32));
             bucket_positions[new_scalar_length] =
-                BucketPosition { bucket_index: current_bucket, scalar_index: new_scalar_length as u32 };
+                BucketPosition { bucket_index: current_bucket, scalar_index: u32::try_from(new_scalar_length)? };
             new_scalar_length += 1;
         }
         global_counter += 1;
@@ -270,7 +271,7 @@ pub(super) fn batch_add<G: AffineCurve>(
                 global_counter += 1;
                 local_counter += 1;
             }
-            if current_bucket >= num_buckets as u32 {
+            if current_bucket >= num_buckets {
                 local_counter = 1;
             } else if local_counter > 1 {
                 // all ones is false if next len is not 1
@@ -318,24 +319,20 @@ pub(super) fn batch_add<G: AffineCurve>(
         new_scalar_length = 0;
     }
 
-    let mut res = vec![Zero::zero(); num_buckets];
+    let mut res = vec![Zero::zero(); num_buckets as usize];
     for bucket_position in bucket_positions.iter().take(num_scalars) {
         res[bucket_position.bucket_index as usize] = new_bases[bucket_position.scalar_index as usize];
     }
-    res
+    Ok(res)
 }
 
 #[inline]
-// It is safe to cast w_start and c to u32 because they were small enough earlier in the callchain
-#[allow(clippy::cast_possible_truncation)]
 fn batched_window<G: AffineCurve>(
     bases: &[G],
     scalars: &[<G::ScalarField as PrimeField>::BigInteger],
-    w_start: usize,
-    c: usize,
-) -> (G::Projective, usize) {
-    assert!(w_start <= u32::MAX as usize);
-    assert!(c <= u32::MAX as usize);
+    w_start: u32,
+    c: u32,
+) -> Result<(G::Projective, u32), anyhow::Error> {
     // We don't need the "zero" bucket, so we only have 2^c - 1 buckets
     let window_size = if (w_start % c) != 0 { w_start % c } else { c };
     let num_buckets = (1 << window_size) - 1;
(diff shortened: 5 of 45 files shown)


algorithms/src/snark/marlin/ahp/ahp.rs
```diff
@@ -237,19 +239,15 @@ impl<F: PrimeField, MM: MarlinMode> AHPForR1CS<F, MM> {
...
-            .collect::<BTreeMap<_, _>>();
+        let mut z_b_s = BTreeMap::new();
+        for (circuit_id, circuit_state) in state.circuit_specific_states.iter() {
+            let mut z_b_i = Vec::with_capacity(usize::try_from(circuit_state.batch_size)?);
+            for i in 0..circuit_state.batch_size {
+                let z_b = witness_label(*circuit_id, "z_b", usize::try_from(i)?);
+                z_b_i.push(LinearCombination::new(z_b.clone(), [(F::one(), z_b)]));
+            }
+            z_b_s.insert(circuit_id, z_b_i);
+        }
```
Pratyush: I think we can use `Itertools::try_collect` here, instead of rewriting as a for loop

algorithms/src/snark/marlin/ahp/ahp.rs
```diff
@@ -279,17 +277,18 @@ impl<F: PrimeField, MM: MarlinMode> AHPForR1CS<F, MM> {
...
-            .map(|(circuit_id, z_b_i)| {
-                let z_b_i_s = z_b_i.iter().map(|z_b| evals.get_lc_eval(z_b, beta).unwrap()).collect::<Vec<F>>();
-                (*circuit_id, z_b_i_s)
+            .values()
+            .map(|z_b_i| {
+                let z_b_i_s = z_b_i.iter().map(|z_b| evals.get_lc_eval(z_b, beta)).collect::<Result<Vec<F>, _>>();
+                z_b_i_s
             })
-            .collect::<BTreeMap<_, _>>();
+            .collect::<Result<Vec<_>, _>>()?;
```
Pratyush: Ditto here, I think we can use `try_collect` to still collect into a BTreeMap

Pratyush: Once we do that, the code that uses `z_b_s_at_beta` should go back to its old state.

algorithms/src/snark/marlin/ahp/prover/round_functions/first.rs
```diff
@@ -74,13 +74,14 @@ impl<F: PrimeField, MM: MarlinMode> AHPForR1CS<F, MM> {
     #[allow(clippy::type_complexity)]
     pub fn prover_first_round<'a, R: RngCore>(
         mut state: prover::State<'a, F, MM>,
+        batch_sizes: impl Iterator<Item = (&'a CircuitId, &'a usize)>,
```
Pratyush: Why are we passing this separately? Isn't it included in `state`?

algorithms/src/snark/marlin/ahp/prover/round_functions/first.rs
```diff
@@ -94,7 +95,7 @@ impl<F: PrimeField, MM: MarlinMode> AHPForR1CS<F, MM> {
             let c_domain = circuit_state.constraint_domain;
             let i_domain = circuit_state.input_domain;
 
-            let mut circuit_r_b_s = Vec::with_capacity(batch_size);
+            let mut circuit_r_b_s = Vec::with_capacity(z_a.len());
```
Pratyush: Why did this change? We do still have `batch_size` in scope

algorithms/src/snark/marlin/ahp/prover/round_functions/first.rs
```diff
@@ -123,20 +124,18 @@ impl<F: PrimeField, MM: MarlinMode> AHPForR1CS<F, MM> {
                 prover::SingleEntry { z_a, z_b, w_poly, z_a_poly, z_b_poly }
             })
             .collect::<Vec<_>>();
-        assert_eq!(batches.len(), state.total_instances);
+        assert_eq!(batches.len(), usize::try_from(state.total_instances)?);
```
Pratyush, line 124: Does this work?

```suggestion
        assert_eq!(batches.len().try_into()?, state.total_instances);
```

algorithms/src/snark/marlin/ahp/prover/round_functions/first.rs
```diff
@@ -74,13 +74,14 @@ impl<F: PrimeField, MM: MarlinMode> AHPForR1CS<F, MM> {
     #[allow(clippy::type_complexity)]
     pub fn prover_first_round<'a, R: RngCore>(
         mut state: prover::State<'a, F, MM>,
+        batch_sizes: impl Iterator<Item = (&'a CircuitId, &'a usize)>,
         rng: &mut R,
     ) -> Result<prover::State<'a, F, MM>, AHPError> {
         let round_time = start_timer!(|| "AHP::Prover::FirstRound");
         let mut r_b_s = Vec::with_capacity(state.circuit_specific_states.len());
-        let mut job_pool = snarkvm_utilities::ExecutionPool::with_capacity(3 * state.total_instances);
+        let mut job_pool = snarkvm_utilities::ExecutionPool::with_capacity(usize::try_from(3 * state.total_instances)?);
```
Pratyush: If the following works, I think we should switch to it, instead of using the more verbose `usize::try_from`

```suggestion
        let mut job_pool = snarkvm_utilities::ExecutionPool::with_capacity((3 * state.total_instances).try_into()?);
```

algorithms/src/snark/marlin/ahp/prover/state.rs
```diff
@@ -114,13 +114,13 @@ impl<'a, F: PrimeField, MM: MarlinMode> State<'a, F, MM> {
 
                 let first_padded_public_inputs = &variable_assignments[0].0;
                 let input_domain = EvaluationDomain::new(first_padded_public_inputs.len()).unwrap();
-                let batch_size = variable_assignments.len();
+                let batch_size = u32::try_from(variable_assignments.len())?;
```
Pratyush: Same, let's change this to `try_into()?`

algorithms/src/snark/marlin/ahp/prover/state.rs
```diff
@@ -114,13 +114,13 @@ impl<'a, F: PrimeField, MM: MarlinMode> State<'a, F, MM> {
...
                 total_instances += batch_size;
-                let mut z_as = Vec::with_capacity(batch_size);
-                let mut z_bs = Vec::with_capacity(batch_size);
-                let mut x_polys = Vec::with_capacity(batch_size);
-                let mut padded_public_variables = Vec::with_capacity(batch_size);
-                let mut private_variables = Vec::with_capacity(batch_size);
+                let mut z_as = Vec::with_capacity(variable_assignments.len());
```
Pratyush: Maybe we should make a `batch_size_usize` local variable and use that here?

algorithms/src/snark/marlin/ahp/verifier/messages.rs
```diff
@@ -93,8 +93,14 @@ impl<F: PrimeField> QuerySet<F> {
...
         // We also use an optimization: instead of explicitly calculating z_c, we
         // use the "virtual oracle" z_a * z_b
-        Self {
-            batch_sizes: state.circuit_specific_states.iter().map(|(c, s)| (*c, s.batch_size)).collect(),
+        let mut batch_sizes = BTreeMap::new();
+        for (circuit_id, batch_size) in
+            state.circuit_specific_states.iter().map(|(circuit_id, s)| (circuit_id, s.batch_size))
+        {
+            batch_sizes.insert(*circuit_id, usize::try_from(batch_size)?);
+        }
```
Pratyush: Same, `try_collect` and `try_into`.

algorithms/src/snark/marlin/ahp/ahp.rs
vicsn: I don't want to return the Result, I have to unwrap the result in order to use z_b on line 246. I got it to compile if we return a `Result` anyway for all paths, but I don't think we'd be happy with this (the compiler asks for those type annotations: 

```
        let z_b_s = state
        .circuit_specific_states
        .iter()
        .map(|(&circuit_id, circuit_state)| {
            let z_b_i = (0..circuit_state.batch_size)
                .map(|i| {
                    let z_b = witness_label(circuit_id, "z_b", usize::try_from(i)?);
                    Ok::<LinearCombination<F>, std::num::TryFromIntError>(LinearCombination::new(z_b.clone(), [(F::one(), z_b)]))
                }).collect::<Result<Vec<_>,_>>();
            Ok::<(CircuitId, Vec<LinearCombination<F>>), std::num::TryFromIntError>((circuit_id, z_b_i?))
        })
        .collect::<Result<BTreeMap<_, _>,_>>()?;
```

Meanwhile, online fora on [related](https://stackoverflow.com/questions/62687455/alternatives-for-using-the-question-mark-operator-inside-a-map-function-closure) issues keep suggesting to just use a forloop.

vicsn: Got it down to 4916a3ca0 but it's still more code than the for-loop

algorithms/src/snark/marlin/ahp/ahp.rs
vicsn: I can't use `try_collect` because I need to annotate the types, but I can use `?` within the map now and collect: 4916a3ca0

algorithms/src/snark/marlin/ahp/prover/round_functions/first.rs
vicsn: It is in state, but we created this iterator already earlier so I thought why not reuse it for line 138 below.

algorithms/src/snark/marlin/ahp/prover/round_functions/first.rs
vicsn: No:

```
127 |         assert_eq!(batches.len().try_into()?, state.total_instances);
    |         ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^ no implementation for `() == u32`
```

algorithms/src/snark/marlin/ahp/prover/round_functions/first.rs
vicsn: 5ebe7b258

algorithms/src/snark/marlin/ahp/prover/round_functions/first.rs
vicsn: 5ebe7b258

algorithms/src/snark/marlin/ahp/prover/state.rs
vicsn: 5ebe7b258

algorithms/src/snark/marlin/ahp/prover/round_functions/first.rs
vicsn: 5ebe7b258

algorithms/src/snark/marlin/ahp/prover/state.rs
vicsn: Indeed would be easier to read 5ebe7b258

algorithms/src/snark/marlin/ahp/verifier/messages.rs
vicsn: 5ebe7b258

algorithms/src/snark/marlin/ahp/prover/state.rs
vicsn: Indeed would be easier to read 5ebe7b258

CONTRIBUTING.md
```diff
@@ -30,6 +30,9 @@ snarkVM is a big project, so (non-)adherence to best practices related to perfor
 - if possible, reuse collections; an example would be a loop that needs a clean vector on each iteration: instead of creating and allocating it over and over, create it _before_ the loop and use `.clear()` on every iteration instead
 - try to keep the sizes of `enum` variants uniform; use `Box<T>` on ones that are large
 
+### Cross-platform consistency
+- types which contain consensus- or cryptographic logic should have a consistent size across platforms. Their serialized output should not contain `usize`, and at times it may be worth it to avoid using usize in the types themselves for clarity.
```
ljedrz:
```suggestion
- types which contain consensus- or cryptographic logic should have a consistent size across platforms. Their serialized output should not contain `usize`, and at times it may be worth it to avoid using `usize` in the types themselves for clarity.
```

algorithms/src/fft/polynomial/sparse.rs
```diff
@@ -18,12 +18,11 @@
 
 use crate::fft::{EvaluationDomain, Evaluations, Polynomial};
 use snarkvm_fields::{Field, PrimeField};
-use snarkvm_utilities::serialize::*;
 
 use std::{collections::BTreeMap, fmt};
 
 /// Stores a sparse polynomial in coefficient form.
-#[derive(Clone, PartialEq, Eq, Hash, Default, CanonicalSerialize, CanonicalDeserialize)]
+#[derive(Clone, PartialEq, Eq, Hash, Default)]
```
ljedrz, line 23: side note: it might be a good idea to check if any other objects can be "trimmed" in a similar fashion for faster compilation

algorithms/src/snark/marlin/data_structures/proof.rs
```diff
@@ -448,7 +451,7 @@ mod test {
             for j in 1..11 {
                 let test_with_none = i * j % 2 == 0;
                 let commitments = rand_commitments(i, j, test_with_none);
-                let batch_sizes = vec![i; j];
+                let batch_sizes = vec![i as u32; j];
```
ljedrz, line 453: nit: this (perfectly valid) cast could be avoided by iterating over `1u32..11`; applies to all such cases; I'm guessing it would lead to a larger diff, though

ljedrz (approved):
LGTM :+1:

ljedrz:
If we're concerned with verbosity, we could always introduce helper macros in `snarkvm_utilities` that would substitute calls like
```
usize::try_from(x)?
```
with sth like
```
to_usize!(x)
```
Such calls are cheap and they provide us with much more cross-platform safety, so even if it's an annoyance, it is IMO well worth it.

algorithms/src/snark/marlin/data_structures/proof.rs
vicsn: We need both usize and u32, so yeah either we cast or create more variables, as these are just tests I think we're fine

vicsn:
For the record, I cannot reproduce this CI error locally: ```thread 'package::run::tests::test_run_with_import' panicked at 'called `Result::unwrap()` on an `Err` value: marlin: PolynomialCommitmentError(AnyhowError(Expected size of 12582920, found size of 12075008))', vm/package/run.rs:145:44```

ljedrz:
> For the record, I cannot reproduce this CI error locally: thread 'package::run::tests::test_run_with_import' panicked at 'called `Result::unwrap()` on an `Err` value: marlin: PolynomialCommitmentError(AnyhowError(Expected size of 12582920, found size of 12075008))', vm/package/run.rs:145:44

some CI tests checking object sizes have been flaky recently

algorithms/src/fft/polynomial/sparse.rs
vicsn: Added a thumbsup! https://github.com/rust-lang/rust/issues/50927

algorithms/src/snark/marlin/ahp/prover/state.rs
```diff
@@ -114,13 +114,14 @@ impl<'a, F: PrimeField, MM: MarlinMode> State<'a, F, MM> {
 
                 let first_padded_public_inputs = &variable_assignments[0].0;
                 let input_domain = EvaluationDomain::new(first_padded_public_inputs.len()).unwrap();
-                let batch_size = variable_assignments.len();
+                let batch_size = variable_assignments.len().try_into()?;
                 total_instances += batch_size;
```
howardwu, line 116: these can overflow. we used `a.checked_add(b).ok_or_else(anyhow!("YOUR ERROR MSG"))?`, or alternative if you know with confidence that saturating is acceptable, then `a.saturating_add(b)`.

howardwu: it might be too much to do all of these in this PR :)

howardwu:
@vicsn the CI failures on `synthesizer` are concerning. Our serialization sizes appear to have genuinely changed. Those failures aren't flaky. (only the ParameterError ones are flaky, which are not present in `synthesizer`).

vicsn:
> @vicsn the CI failures on synthesizer are concerning. Our serialization sizes appear to have genuinely changed. Those failures aren't flaky. (only the ParameterError ones are flaky, which are not present in synthesizer).

I should have tackled this right away! The failing tests can be explained by the fact that we moved from `usize` (which was 64 bit on everyone's machine so far) to u32. So we can either (1) change the code to use u64 or (2) resample parameters yet again and update the tests. As a batch size of `u32` gives us enough space I vote for this option.

In more detail:
- The `block::` tests fail because they can't deserialize Proof - which is hardcoded. Resampling genesis resolved the issue for me.
- The `vm::` tests fail because they are 8 bytes (for two proof instances) or 4 bytes (for one proof instance) less than expected. 1690b3b9a

algorithms/src/snark/marlin/ahp/prover/state.rs
vicsn: Will scan and tackle such cases in a next PR!

howardwu:
@vicsn can you try adding the following line to the top of `snarkvm-algorithms/src/lib.rs` and see if the compiler warns?
```rs
#![warn(clippy::cast_possible_truncation)]
```
We use this flag in `circuit` and `console` to ensure there are no unsafe casts.

vicsn:
> ```rust
> warn(clippy::cast_possible_truncation)]
> ```

Added it. For the adjustments in the last commits I:
- let structs use fixed-size types instead of usize to make it clear to readers what size to expect for operations
- if the callchain doesn't use Results or requires a lot of casting, see if we can avoid the overhead of `try_from` and wrap the function in `#[allow(...)]`. I added comments about why casting is safe, if you think it is too ugly we can remove those but at least it can help your review. A big downside of this approach is that it can give a false sense of security in the future - just like all uses of `#[allow(...)]` in the codebase. It would be cool if `#[allow(...)]` codeblocks would have to be manually approved/enabled in CI whenever there is a commit which changes something in the respective codeblock. Perhaps for now it suffices to add a line to the PR template or CONTRIBUTING: "- [ ] I checked whether the functions I changed are still allowed to have it's #[allow(...)]". @ljedrz shall I add that?

ljedrz:
@vicsn I think it's always better to have more comments to avoid any misunderstanding/misuse :+1:. As for future-proofing, I think the safest course of action is to have more comprehensive tests, including extremes in terms of the number of items, but having some extra information in `CONTRIBUTING` is also valuable.

What we could also do is add `assert`ions in some of the places (the ones where we know the value is smaller than some other value) where we perform the "currently valid" casts - simple comparisons shouldn't be noticeable at all perf-wise.

algorithms/src/snark/marlin/data_structures/proof.rs
```diff
@@ -273,16 +277,17 @@ pub struct Proof<E: PairingEngine> {
...
         commitments: Commitments<E>,
         evaluations: Evaluations<E::Fr>,
         msg: ahp::prover::ThirdMessage<E::Fr>,
         pc_proof: sonic_pc::BatchLCProof<E>,
     ) -> Result<Self, SNARKError> {
         let mut total_instances = 0;
-        let batch_sizes: Vec<usize> = batch_sizes.into_values().collect();
-        for (z_b_evals, batch_size) in evaluations.z_b_evals.iter().zip(&batch_sizes) {
+        let mut batch_sizes = Vec::with_capacity(batch_sizes_in.len());
+        for (z_b_evals, batch_size) in evaluations.z_b_evals.iter().zip(batch_sizes_in.values()) {
```
ljedrz: nice allocation removal :ok_hand:

ljedrz (approved):
The further changes look fine to me :+1:.

algorithms/src/msm/variable_base/standard.rs
```diff
@@ -40,6 +41,7 @@ fn update_buckets<G: AffineCurve>(
     // If the scalar is non-zero, we update the corresponding bucket.
     // (Recall that `buckets` doesn't have a zero bucket.)
     if scalar != 0 {
+        assert!(scalar - 1 <= usize::MAX as u64);
```
ljedrz: `usize::MAX as u64` will truncate on architectures >64bit, is that fine?

vicsn: Good catch. Yes because `scalar` is a u64! But that also makes it useless. Removed it

vicsn:
@howardwu if you like the PR's logic, I suggest we merge and I will tackle this piece by piece in the coming weeks: https://github.com/AleoHQ/snarkVM/issues/1567

algorithms/src/snark/marlin/ahp/prover/round_functions/first.rs
howardwu: Why are we passing this in if it's in state?

algorithms/src/polycommit/sonic_pc/data_structures.rs
```diff
@@ -217,12 +217,14 @@ impl<E: PairingEngine> FromBytes for CommitterKey<E> {
...
             enforced_degree_bounds,
-            max_degree: max_degree as usize,
+            max_degree,
         })
     }
 }
 
 impl<E: PairingEngine> ToBytes for CommitterKey<E> {
+    // Values are safe to cast to u32 because they are read as u32.
+    #[allow(clippy::cast_possible_truncation)]
```
howardwu: This should be fixed, not allowed. See https://github.com/AleoHQ/snarkVM/blob/testnet3/synthesizer/coinbase/src/helpers/coinbase_solution/bytes.rs#L37 for how to proceed

algorithms/src/polycommit/kzg10/data_structures.rs
```diff
@@ -131,6 +131,8 @@ impl<E: PairingEngine> FromBytes for UniversalParams<E> {
     }
 }
 
+// It is safe to cast `supported_degree_bounds.len()` and individual bounds to a `u32` because they are read and used as u32
+#[allow(clippy::cast_possible_truncation)]
```
howardwu: This should be fixed, not allowed. See https://github.com/AleoHQ/snarkVM/blob/testnet3/synthesizer/coinbase/src/helpers/coinbase_solution/bytes.rs#L37 for how to proceed

algorithms/src/msm/variable_base/standard.rs
```diff
@@ -21,13 +21,17 @@ use snarkvm_utilities::{cfg_into_iter, BigInteger};
...
+// It is safe to cast scalar to usize because it is mod c, which is usize
+#[allow(clippy::cast_possible_truncation)]
 fn update_buckets<G: AffineCurve>(
     base: &G,
     mut scalar: <G::ScalarField as PrimeField>::BigInteger,
     w_start: usize,
     c: usize,
     buckets: &mut [G::Projective],
 ) {
+    assert!(w_start <= u32::MAX as usize);
```
howardwu: Why are we using assert?

algorithms/src/msm/variable_base/batched.rs
```diff
@@ -325,12 +328,16 @@ pub(super) fn batch_add<G: AffineCurve>(
...
 #[inline]
+// It is safe to cast w_start and c to u32 because they were small enough earlier in the callchain
+#[allow(clippy::cast_possible_truncation)]
 fn batched_window<G: AffineCurve>(
     bases: &[G],
     scalars: &[<G::ScalarField as PrimeField>::BigInteger],
     w_start: usize,
     c: usize,
 ) -> (G::Projective, usize) {
+    assert!(w_start <= u32::MAX as usize);
```
howardwu: Why are we using assert?

algorithms/src/msm/variable_base/batched.rs
```diff
@@ -172,13 +172,16 @@ fn batch_add_write<G: AffineCurve>(
...
+// It is safe to cast num_buckets to u32 because it was small enough earlier in the callchain
+#[allow(clippy::cast_possible_truncation)]
 pub(super) fn batch_add<G: AffineCurve>(
     num_buckets: usize,
     bases: &[G],
     bucket_positions: &mut [BucketPosition],
 ) -> Vec<G> {
     // assert_eq!(bases.len(), bucket_positions.len());
     assert!(!bases.is_empty());
+    assert!(num_buckets <= u32::MAX as usize);
```
howardwu: Why are we using assert?

algorithms/src/lib.rs
```diff
@@ -17,6 +17,7 @@
 #![allow(clippy::module_inception)]
 #![allow(clippy::type_complexity)]
 #![cfg_attr(test, allow(clippy::assertions_on_result_states))]
+#![warn(clippy::cast_possible_truncation)]
```
howardwu, line 18: This should be forbidden.

algorithms/src/crypto_hash/poseidon.rs
```diff
@@ -379,12 +379,15 @@ impl<F: PrimeField, const RATE: usize> PoseidonSponge<F, RATE, 1> {
...
 
+    // params.bits_per_limb is safe to cast as u32 because it will be less than `max_limb_size`
+    #[allow(clippy::cast_possible_truncation)]
     /// Obtain the limbs directly from a big int
     pub fn get_limbs_representations_from_big_integer<TargetField: PrimeField>(
         elem: &<TargetField as PrimeField>::BigInteger,
         optimization_type: OptimizationType,
     ) -> SmallVec<[F; 10]> {
         let params = get_params(TargetField::size_in_bits(), F::size_in_bits(), optimization_type);
+        assert!(params.bits_per_limb <= u32::MAX as usize);
```
howardwu: Why are we using assert?

Pratyush: In this case I think it's fine to error out, or we can have `params.bits_per_limb` be a `u32`. Generally when we have a malformed config it's better to assert and fail quickly. This avoid polluting function signatures with `Result`s

algorithms/src/lib.rs
vicsn: The rest of the codebase uses `#![warn(...)]`, do you want me to adjust at least the `cast_possible_truncation` to `deny`?

howardwu: My memory was off, thanks for pointing out, resolving.

algorithms/src/msm/variable_base/batched.rs
vicsn: So instead of panicking on a failed cast, we'd panic somewhat more clearly on this assert. 

But anyway, gave it more thought, removed the assert and now passing a u32 to this function so we only have to reason about safe casting one level higher.

algorithms/src/msm/variable_base/batched.rs
vicsn: Just for extra clarity, we'd already panic if the conversion would fail so I now removed the asserts.

algorithms/src/msm/variable_base/standard.rs
vicsn: Just for extra clarity, removed the assert now

algorithms/src/polycommit/sonic_pc/data_structures.rs
vicsn: Fixed: f339c4cbd

algorithms/src/polycommit/kzg10/data_structures.rs
vicsn: Fixed: f339c4cbd

algorithms/src/snark/marlin/ahp/prover/round_functions/first.rs
vicsn: See the comment right above yours... But since you both think it's weird, removed it: 7f2a6c800

vicsn:
Closing this, it's very outdated and validators are required to run on 64-bit architectures. Will consider making a new issue for potentially 32-bit wallets.

PR #1501, opened 2023-05-03
Use `BTreeMap` in `MerkleTree::update_many` to enforce uniqueness

raychu86:
## Motivation

This PR replaces `&[(usize, LH::Leaf)]` with `&BTreeMap<usize, LH::Leaf>` in `MerkleTree::update_many` to enforce ordering during the update. 

`update_many` requires the leaf indices to be sorted in descending order and be unique, so using a `BTreeMap` covers both cases.


howardwu (approved):
LGTM

PR #1503, opened 2023-05-03
[Feature] Increase limits for `Finalize`.

d0cd:
This PR,
- increases the number of commands allowed in a `Finalize` block.
- fixes off-by-one errors in allowed program size.


howardwu:
Why was `<= `changed to `<`?

d0cd:
This is an off-by-one error.
The methods check that the current amount is less than or equal to the maximum before adding another.

howardwu:
great catch, thanks

PR #1510, opened 2023-05-05
Multi circuit proof integration

vicsn:
## Motivation

This adds support for batch proving executions and their inclusions.

This is a very rough first draft. I suggest we do the review in two parts: first only important architecture comments on e.g. the callchain, function signatures and objects. There is a lot of repetition in the code, and existing TODOs of mine, so please save yourself time by mentioning your ideas only once.

Some notes on this PR:
- I removed the generic from `verify_execution<const VERIFY_INCLUSION: bool>` which was never actually used, because `verify_execution` would still check whether the execution had an inclusion proof.
- I separate preparation from execution inside of `vm.execute()` and `vm.execute_fee()` to avoid us having to `cast_ref!` parameters up and down the call chain.
- Ray asked: "How are we storing the batched function and fee proofs in the Transaction in a way that clearly references which assignments were used and which inputs are required for verification?" I now store a proof in the Execution and Fee objects. I think ordering is straightforward: just adhere to the fixed ordering of the Vector/Array of the Execution/Fee's Assignments/Inputs.
- I replaced the `Execute` mode of the `Callstack` by a `Prepare` mode - running execute_function(...) in Prepare mode doesn't create a proof so instead the caller can afterwards extract the filled registers.
- `inherent associated types are unstable` if we try sth like: type ExecutionAssignments<'a, N> = Vec<&Assignment<N::Field>>;

## Test Plan
- You'll have to resample parameters.
- It seems the original code never verified inclusion proofs. The likely reason for this is because this requires two input records, which was slightly more effort to program. Instead we had e.g. two test cases:
  - `sample_execution_transaction_without_fee` calls transfer() but fails as expected before verifying the inclusion
  - `sample_execution_transaction_with_fee` calls mint() which doesn't require an input record and corresponding 
inclusion proof. I adjusted the test now to call transfer(), there are still other tests which call mint().

## Related PRs and design document

https://github.com/AleoHQ/snarkVM/pull/1382
https://docs.google.com/document/d/10f_VfrcL9POyBPdFWVov5vVieVL-YNW6gBCxidEoNq8/edit#heading=h.9e2ahww0jjqr


synthesizer/src/process/execute_fee.rs
```diff
@@ -15,73 +15,48 @@
...
-        // Extract the metrics.
-        let metrics = Arc::try_unwrap(metrics).unwrap().into_inner();
+        let fee_function = Identifier::from_str("fee")?;
+        ensure!(*fee_transition.function_name() == fee_function, "Incorrect function name for fee");
 
-        finish!(timer);
+        // Ensure the assignments are not empty.
+        if assignments.is_empty() {
+            bail!("Inclusion expected the assignments for the fee to *not* be empty")
+        }
```
ljedrz:
```suggestion
        ensure!(!assignments.is_empty(), "Inclusion expected the assignments for the fee to *not* be empty");
```
nit

synthesizer/src/process/mod.rs
```diff
@@ -469,12 +477,20 @@ mod tests {
         assert_eq!(authorization.len(), 1);
 
         // Execute the request.
-        let (response, execution, _inclusion, _metrics) =
-            process.execute::<CurrentAleo, _>(authorization, rng).unwrap();
+        let mut all_assignments = Vec::with_capacity(1);
+        let (_response, mut execution, _inclusion, _metrics, function_assignments) =
+            process.prepare_function::<CurrentAleo, _>(authorization, rng).unwrap();
+        let function_assignments = Arc::try_unwrap(function_assignments).unwrap_or_default().into_inner(); // TODO: return Err if there is one
```
ljedrz: IIRC an existence of `Arc` clones here would indicate a leak/logic error

synthesizer/src/process/mod.rs
```diff
@@ -616,15 +632,22 @@ function hello_world:
...
```
ljedrz: ditto; also, this is just a test; the same applies to the other instances below

synthesizer/src/process/mod.rs
vicsn: Yes in a next iteration I'm following the same approach we use for e.g. execution/inclusion, just try_unwrap

synthesizer/src/snark/batch/bytes.rs
```diff
@@ -0,0 +1,77 @@
...
+        // Read the mode.
+        let modeint = u8::read_le(&mut reader)?;
+        let mode = match modeint {
+            0 => KeyMode::Proving,
+            1 => KeyMode::Verifying,
+            _ => return Err(error("Invalid key mode")),
+        };
+        // Read the number of keys.
+        // TODO: instead of unwrap, convert the error to an IoError
+        let batch_len = usize::try_from(u32::read_le(&mut reader)?).unwrap();
```
ljedrz: is there a plan to support any 16bit architectures? if not, casting `u32` to `usize` should never fail

synthesizer/src/snark/batch/bytes.rs
```diff
@@ -0,0 +1,77 @@
...
+
+use super::*;
+
+impl<N: Network> FromBytes for KeyBatch<N> {
+    /// Reads the proving key from a buffer.
+    fn read_le<R: Read>(mut reader: R) -> IoResult<Self> {
+        // Read the version.
+        let version = u16::read_le(&mut reader)?;
+        // Ensure the version is valid.
+        if version != 0 {
```
ljedrz: this should point to some `const`

synthesizer/src/snark/batch/bytes.rs
```diff
@@ -0,0 +1,77 @@
...
+        }
+        Ok(Self { batch, mode })
+    }
+}
+
+impl<N: Network> ToBytes for KeyBatch<N> {
+    /// Writes the proving key to a buffer.
+    fn write_le<W: Write>(&self, mut writer: W) -> IoResult<()> {
+        // Write the version.
+        0u16.write_le(&mut writer)?;
```
ljedrz: same `const` point here

synthesizer/src/snark/batch/bytes.rs
```diff
@@ -0,0 +1,77 @@
...
+            KeyMode::Proving => 0u8.write_le(&mut writer)?,
+            KeyMode::Verifying => 1u8.write_le(&mut writer)?,
+        }
+        // Write the number of Keys.
+        // TODO: instead of unwrap, convert the error to an IoError
+        u32::try_from(self.batch.len()).unwrap().write_le(&mut writer)?;
+        // Write the bytes.
+        for i in 0..self.batch.len() {
+            // TODO: I could add yet another check that we're not mixing pk and vk, but I think it should be fine?
+            match &self.batch[i] {
```
ljedrz:
```suggestion
        for key in &self.batch {
            // TODO: I could add yet another check that we're not mixing pk and vk, but I think it should be fine?
            match key {
```

synthesizer/src/snark/batch/mod.rs
```diff
@@ -0,0 +1,141 @@
...
+
+use std::collections::BTreeMap;
+
+#[derive(Clone, PartialEq, Eq)]
+pub enum Key<N: Network> {
+    ProvingKey(ProvingKey<N>),
+    VerifyingKey(VerifyingKey<N>),
+}
+
+#[derive(Clone)]
```
ljedrz:
```suggestion
#[derive(Clone, Copy, PartialEq, Eq)]
```
seems reasonable to extend these for such a simple enum

synthesizer/src/snark/batch/mod.rs
```diff
@@ -0,0 +1,141 @@
...
+
+    /// Adds keys to the batch.
+    pub(crate) fn add(
+        &mut self,
+        key: Key<N>,
+    ) -> Result<(), anyhow::Error> {
+        // TODO: consider creating custom error type. We use error() in this folder elsewhere
+        match key {
+            Key::VerifyingKey(_) => if matches!(self.mode,  KeyMode::Proving) {
+                return Err(anyhow!("Cannot add a verifying key to a proving batch"))
```
ljedrz:
```suggestion
                bail!("Cannot add a verifying key to a proving batch");
```
nit; applies to other spots like this

synthesizer/src/block/transition/mod.rs
```diff
@@ -430,3 +430,31 @@ impl<N: Network> Transition<N> {
...
+        let self_function_name_str = self.function_name.to_string();
+        let other_function_name_str = other.function_name.to_string();
+        let self_program_id_str = self.program_id.to_string();
+        let other_program_id_str = other.program_id.to_string();
+
+        // TODO: we should avoid the function_name from leaking into the program_id:
+        // option 1) '_' separator must be unique and not present in either string
+        // option 2) we can hash the function_name's and program_id's
+        (self_function_name_str + "_" + &self_program_id_str)
+            .cmp(&(other_function_name_str + "_" + &other_program_id_str))
```
ljedrz: won't this be rather slow? I'd recommend relying on `Ord` of the members, or at least comparing bytes (as opposed to strings)

vicsn: Updated! 84a844611

synthesizer/src/process/execute_fee.rs
vicsn: Done!

synthesizer/src/process/mod.rs
vicsn: Done

synthesizer/src/snark/batch/bytes.rs
vicsn: I think with that assumption in hand this PR would also be a lot simpler: https://github.com/AleoHQ/snarkVM/pull/1499
But we spent so much time on it already that I think not assuming that is fine for now.

synthesizer/src/snark/batch/bytes.rs
vicsn: Done

synthesizer/src/snark/batch/bytes.rs
vicsn: Done

synthesizer/src/snark/batch/bytes.rs
vicsn: Done thx

synthesizer/src/snark/batch/mod.rs
vicsn: Done

synthesizer/src/snark/batch/mod.rs
vicsn: Done

synthesizer/src/block/transition/mod.rs
```diff
@@ -440,3 +430,25 @@ impl<N: Network> Transition<N> {
...
+            false => self.program_id.cmp(&other.program_id),
+        }
+    }
+}
+
+impl<N: Network> PartialOrd for ProvingKeyId<N> {
+    fn partial_cmp(&self, other: &Self) -> Option<std::cmp::Ordering> {
+        Some(self.cmp(other))
+    }
+}
```
ljedrz, line 454: this should be identical to an automatic `#[derive(PartialOrd, Ord)]`: the `program_id` is compared first, and if it's the same, the `function_name` is compared ([source](https://doc.rust-lang.org/std/cmp/trait.PartialOrd.html#derivable)).

synthesizer/src/process/prepare.rs
```diff
@@ -0,0 +1,148 @@
...
+        // Extract the inclusion.
+        let inclusion = Arc::try_unwrap(inclusion).unwrap().into_inner();
+        // Extract the metrics.
+        let metrics = Arc::try_unwrap(metrics).unwrap().into_inner();
+        // Extract the assignments
+        let assignments = Arc::try_unwrap(assignments).unwrap_or_default().into_inner();
+
+        finish!(timer);
+
+        Ok((response, execution.peek()?.clone(), inclusion, assignments, metrics))
```
ljedrz: why not extract (move) the value from `execution` instead of cloning it?

synthesizer/src/process/stack/execution/mod.rs
```diff
@@ -60,13 +65,54 @@ impl<N: Network> Execution<N> {
...
+        assignments: &[&[Assignment<N::Field>]],
+        function_names: &[&Identifier<N>],
+        rng: &mut R,
+    ) -> Result<()> {
+        ensure!(self.proof.is_none(), "Proof already exists!");
+        ensure!(assignments.len() == function_names.len(), "Need to have equal assignments and names");
+        ensure!(!assignments.is_empty(), "Number of assignments must be greater than zero!");
+
+        let inclusion_name = Identifier::<N>::from_str(N::INCLUSION_FUNCTION_NAME)?;
+        let proves_inclusion = *function_names[function_names.len() - 1] == inclusion_name;
```
ljedrz:
```suggestion
        let proves_inclusion = *function_names.last().unwrap() == inclusion_name;
```
(same safety, easier to read)

synthesizer/src/process/stack/execution/mod.rs
```diff
@@ -133,4 +179,9 @@ impl<N: Network> Execution<N> {
     pub fn commitments(&self) -> impl '_ + Iterator<Item = &Field<N>> {
         self.transitions.values().flat_map(Transition::commitments)
     }
+
+    /// Inserts the global_state_root.
+    pub fn add_global_state_root(&mut self, global_state_root: N::StateRoot) {
```
ljedrz: nit: "set" or "update" might be more fitting than "add" here

synthesizer/src/process/stack/fee/mod.rs
```diff
@@ -76,9 +78,50 @@ impl<N: Network> Fee<N> {
         self.global_state_root
     }
 
-    /// Returns the inclusion proof.
-    pub const fn inclusion_proof(&self) -> Option<&Proof<N>> {
-        self.inclusion_proof.as_ref()
+    /// Returns whether the fee proves inclusion
+    pub fn proves_inclusion(&self) -> bool {
+        if self.proof.is_some() { self.proof.as_ref().unwrap().proves_inclusion() } else { false }
```
ljedrz: `if let Some(proof) = &self.proof (...)` would be tidier

synthesizer/src/process/stack/fee/mod.rs
```diff
@@ -76,9 +78,50 @@ impl<N: Network> Fee<N> {
...
+        rng: &mut R,
+    ) -> Result<()> {
+        if assignments.is_empty() || assignments.len() > 2 {
+            bail!("Expected 1 or 2 assignments, got {}", assignments.len())
+        }
+        let inclusion_name = Identifier::<N>::from_str(N::INCLUSION_FUNCTION_NAME)?;
+        if assignments.len() == 2 && *function_names[1] != inclusion_name {
+            bail!("Expected 2nd assignment to belong to inclusion")
+        }
+        let proves_inclusion = *function_names[function_names.len() - 1] == inclusion_name;
```
ljedrz:
```suggestion
        let proves_inclusion = *function_names.last().unwrap() == inclusion_name;
```

synthesizer/src/store/helpers/rocksdb/internal/id.rs
```diff
@@ -177,7 +177,7 @@ enum DataID {
     // Execution
     ExecutionIDMap,
     ExecutionReverseIDMap,
-    ExecutionInclusionMap,
+    ExecutionExecutionMap,
```
ljedrz, line 179: I know this can make sense, just double-checking that it's supposed to say `ExecutionExecution` :smiley:

synthesizer/src/store/helpers/memory/transition.rs
```diff
@@ -66,7 +67,6 @@ impl<N: Network> TransitionStorage<N> for TransitionMemory<N> {
             input_store: InputStore::open(dev)?,
             output_store: OutputStore::open(dev)?,
             finalize_map: MemoryMap::default(),
-            proof_map: MemoryMap::default(),
```
ljedrz, line 69: should we mark the corresponding `DataID` as [`deprecated`](https://doc.rust-lang.org/reference/attributes/diagnostics.html#the-deprecated-attribute)? if we don't intend to use it anymore, we can remove it before the next network reset

ljedrz (commented):
Left a few more comments.

synthesizer/src/block/transition/mod.rs
vicsn: Due to the way the derived Ord/PartialOrd works, the compiler also demands the networktype (Testnet3) to have those derivations, and to adjust a large amount of trait bounds to be updated to N: `Network+Ord+PartialOrd`. I think keep the code as is would be better.

For reference, a non-compiling partial implementation:
```
diff --git a/console/network/src/testnet3.rs b/console/network/src/testnet3.rs
index b29dcf99e..2ebd3435d 100644
--- a/console/network/src/testnet3.rs
+++ b/console/network/src/testnet3.rs
@@ -75,7 +75,7 @@ lazy_static! {
     };
 }
 
-#[derive(Copy, Clone, Debug, PartialEq, Eq, Hash, Serialize, Deserialize)]
+#[derive(Copy, Clone, Debug, PartialEq, Eq, Hash, Serialize, Deserialize, Ord, PartialOrd)]
 pub struct Testnet3;
 
 impl Testnet3 {
diff --git a/synthesizer/src/block/transition/mod.rs b/synthesizer/src/block/transition/mod.rs
index 1ac051851..863409eb9 100644
--- a/synthesizer/src/block/transition/mod.rs
+++ b/synthesizer/src/block/transition/mod.rs
@@ -431,24 +431,8 @@ impl<N: Network> Transition<N> {
     }
 }
 
-#[derive(Clone, Copy, PartialEq, Eq)]
-pub struct ProvingKeyId<N: Network> {
+#[derive(Clone, Copy, PartialEq, Eq, Ord, PartialOrd)]
+pub struct ProvingKeyId<N: Network + Ord> {
     pub program_id: ProgramID<N>,
     pub function_name: Identifier<N>,
-}
-
-impl<N: Network> Ord for ProvingKeyId<N> {
-    /// Ordering is determined by the program_id first, then the function_name second.
-    fn cmp(&self, other: &Self) -> Ordering {
-        match self.program_id == other.program_id {
-            true => self.function_name.cmp(&other.function_name),
-            false => self.program_id.cmp(&other.program_id),
-        }
-    }
-}
-
-impl<N: Network> PartialOrd for ProvingKeyId<N> {
-    fn partial_cmp(&self, other: &Self) -> Option<std::cmp::Ordering> {
-        Some(self.cmp(other))
-    }
-}
+}
\ No newline at end of file
diff --git a/synthesizer/src/process/execute.rs b/synthesizer/src/process/execute.rs
index eebdd5249..d761acc06 100644
--- a/synthesizer/src/process/execute.rs
+++ b/synthesizer/src/process/execute.rs
@@ -18,7 +18,7 @@ use super::*;
 
 use crate::{Key, KeyBatch, KeyMode, ProvingKeyId};
 
-impl<N: Network> Process<N> {
+impl<N: Network + Ord + PartialOrd> Process<N> {

```

synthesizer/src/process/prepare.rs
vicsn: Done: a5fb2b7be

synthesizer/src/process/stack/execution/mod.rs
vicsn: a5fb2b7be

synthesizer/src/store/helpers/rocksdb/internal/id.rs
vicsn: It is painfully ugly, but I followed the example of `FeeFeeMap`, so keeping that example for now :D

synthesizer/src/store/helpers/memory/transition.rs
vicsn: Hmm good question, implemented it here: 50ab2d307
@raychu86 do we have a process in place for updating database objects? Should I document it in CONTRIBUTING?

synthesizer/src/process/stack/execution/mod.rs
vicsn: Done: 79f705e58

synthesizer/src/process/stack/fee/mod.rs
vicsn: Done: 1ff86a0db

synthesizer/src/process/stack/fee/mod.rs
vicsn: Done

synthesizer/src/store/helpers/memory/transition.rs
ljedrz: related: https://github.com/AleoHQ/snarkVM/pull/1560; also, only the `DataID` variant needs to remain, so as to not perturb the order (and values) of the others (unless we explicitly assign values to the `DataID` variants)

vicsn: On advice from Howard, we're merging this PR into staging, allowing us to just remove the DataIDs

PR #1511, opened 2023-05-05
[Fix] Reorder evaluation of `finalize` transitions.

d0cd:
This PR reorders the evaluation of transitions when finalizing an `Execution`.
This is necessary to ensure correct semantics when finalize transactions that contain external calls.

Suppose that we have the following programs:
- `C`, with transition `c`.
- `B`, with transition `b`.
- `A`, which imports `B` and `C`, and contains transition `a` which invokes `B/b` and `C/c`.
Suppose that the call graph looks like:
```
A::a
  --> B::b
  --> C::c
```
Then the finalize blocks need to evaluated in the following order: `
```
B::b, C::c, A::a
```
In other words, **finalize blocks are evaluated in order of the transition that finished first**.

For a more complex example,
Suppose we have the following call graph, where each call is an external call.
```
A::a
--> B::b
       --> C::c
       --> D::d
--> E::e
       --> B::b
              --> C::c
              --> D::d"
       --> D::d
       --> C::c
```
The finalize blocks need to be evaluated in the following order: 
```
C::c, D::d, B::b, C::c, D::d, B::b, D::d, C::c, E::e, A::a
```

### DIFF
--- a/synthesizer/src/process/finalize.rs
+++ b/synthesizer/src/process/finalize.rs
@@ -138,7 +138,7 @@ impl<N: Network> Process<N> {
             // TODO (howardwu): This is a temporary approach. We should create a "CallStack" and recurse through the stack.
             //  Currently this loop assumes a linearly execution stack.
             // Finalize each transition, starting from the last one.
-            for transition in execution.transitions().rev() {
+            for transition in execution.transitions() {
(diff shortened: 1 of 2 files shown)


d0cd:
Closed. Tests included in #1524.

PR #1513, opened 2023-05-06
Update version types to be consistent - u8

raychu86:
## Motivation

This PR updates the version types for objects to be consistent - u8. Previously it was a mix of u8 and u16.


howardwu (approved):
LGTM, pending resampling.

PR #1515, opened 2023-05-06
A benchmarking framework for user-defined workloads.

d0cd:
## Motivation

As we roll out on-chain execution via `finalize`, we need to benchmark its performance on "real" workloads. While the definition of a "real" workload is somewhat arbitrary, we do know that we want to bench a variety of programs (deployments and executions). With that in mind, we desire a benchmarking framework with the following characterisitcs:
1. Easy/ergonomic to define benchmarks
2. Minimal overhead in setting up and running benchmarks
3. Reproducibility

## Design
This framework takes an opinionated stance in defining and running benchmarks.

### A `Benchmark` 
Users can specify new benchmarks by instantiating objects that implement the `Benchmark` trait.
```rust
pub enum Operation<N: Network> {
    Deploy(Box<Program<N>>),
    Execute(String, String, Vec<Value<N>>),
}

pub trait Benchmark<N: Network> {

    fn name(&self) -> String;

    fn setup_operations(&mut self) -> Vec<Vec<Operation<N>>>;

    fn benchmark_operations(&mut self) -> Vec<Operation<N>>;
}
```
In creating a new `Benchmark`, the user specifies:
- A unique name.
- A sequence of "setup" operations that are run before actually running the benchmark. The setup operations are specified in batches, where each batch is included in a single block.
- A sequnece of "benchmark" operations that are to be timed.

See `synthesizer/benches/benchmarks` for some examples.

### A `Workload`

A `Workload` is a sequence of benchmarks that are run on a VM. 
```rust
pub struct Workload {
    name: String,
    benchmarks: Vec<Box<dyn Benchmark<Testnet3>>>,
    object_store: ObjectStore,
}

impl Workload {
    pub fn new(name: String, benchmarks: Vec<Box<dyn Benchmark<Testnet3>>>) -> Result<Self> {...}

    pub fn add(&mut self, benchmark: Box<dyn Benchmark<Testnet3>>) {...}

    pub fn name(&self) -> &String {...}

    pub fn setup<C: ConsensusStorage<Testnet3>>(&mut self) -> (VM<Testnet3, C>, PrivateKey<Testnet3>, BenchmarkBatch, TestRng) {...}
}
```
Users construct a `Workload` with a **unique** name with the relevant benchmarks using `new`.

A workload is run by first invoking the `setup` function. `setup` will check the `object_store` whether or not precomputed workload objects (blocks containing setup transactions and the benchmark transactions) are stored. 

If they are not stored, they are initialized **NOTE: initialization may take a long time depending on the size of the workload.** 

Once the workload objects are initialized, an RNG is instantiated from a stored seed, a VM is set up with the blocks containing the setup transactions, and a triple of the VM, benchmark transactions, and RNG are returned for further use.

### The `ObjectStore`
The `ObjectStore` is a key component that enables caching workload objects for efficient setup and reproducibility. 
```rust
pub struct ObjectStore {
    root: PathBuf,
}

impl ObjectStore {
    pub fn new(root: PathBuf) -> Result<Self> {...}

    pub fn get<O: FromBytes, P: AsRef<Path>>(&mut self, path: P) -> Result<O> {...}

    pub fn insert<O: ToBytes + FromBytes, P: AsRef<Path>>(&mut self, path: P, object: &O) -> Result<()> {...}

    pub fn contains<P: AsRef<Path>>(&self, path: P) -> bool {...}

    pub fn clear(&mut self) -> Result<()> {...}
}
```
The `ObjectStore` acts as a key-value store using the filesystem. **NOTE: the `ObjectStore` can arbitrarily change the contents of the `root` directory.**

### Considerations

Although this PR is designed with benchmarking in mind, a `Workload` and and `ObjectStore` can be used to construct sets of blocks and transactions for other forms of testing. For example, one can define a large number of transactions to fire at the network for load testing. Another is to pre-construct a number of blocks to initialize the VM into a complex state.

## Usage

This PR introduces benchmarks for `VM::finalize` and `VM::add_next_block`. 
To run these benchmarks, run
```
cargo bench --features=testing
```
To run these benchmarks with the RocksDB backend, run
```
cargo bench --features=testing,rocks
```
If you'd like to avoid the time it takes to initialize the workloads (~6hrs), feel free to contact me directly for `.zip`s of the workload. Though this method isn't fully tested, I've been able to avoid setup on my secondary machine by copying the files over.

## Results

The results of benchmarking finalize can be found [here](https://hackmd.io/@_kjs44DbTQKhMY9aJJiTxw/rkyueU4En).
Note the benchmark names indicate what is tested, but can be hard to parse. Please feel free to ask for clarification.

## Future Work
- Reduce time it takes to produce fee records by increasing split width (use external call to `credits.aleo`).
- Add "accounts". Each operation is associated with a different private key. Allows specifying more complex interactions in a benchmark.
- Support dependent program interactions in a benchmark.
- Improve `ObjectStore` interface for other uses.


synthesizer/Cargo.toml
```diff
@@ -34,6 +40,24 @@ path = "benches/coinbase_puzzle.rs"
...
+name = "finalize"
+path = "benches/finalize.rs"
+harness = false
+required-features = [ "testing"]
+
+[[bench]]
+name = "speculate"
+path = "benches/speculate.rs"
+harness = false
+required-features = [ "testing"]
```
ljedrz, line 59:
```suggestion
required-features = [ "testing" ]
```

synthesizer/Cargo.toml
```diff
@@ -34,6 +40,24 @@ path = "benches/coinbase_puzzle.rs"
...
+name = "deployment"
+path = "benches/deployment.rs"
+harness = false
+required-features = [ "testing" ]
+
+[[bench]]
+name = "finalize"
+path = "benches/finalize.rs"
+harness = false
+required-features = [ "testing"]
```
ljedrz, line 53:
```suggestion
required-features = [ "testing" ]
```

synthesizer/Cargo.toml
```diff
@@ -43,6 +67,7 @@ harness = false
 default = [ ]
 aleo-cli = [ ]
 cuda = [ "snarkvm-algorithms/cuda" ]
+long-benchmarks = [ ]
```
ljedrz, line 70: shouldn't this one include `"testing"`?

synthesizer/benches/add_next_block.rs
```diff
@@ -0,0 +1,132 @@
...
+
+        let mut num_transactions = 0f64;
+        let mut num_rejected = 0f64;
+
+        c.bench_function(&format!("{header}/{name}/add_next_block"), |b| {
+            b.iter_batched(
+                // TODO: Do the blocks need to be removed on each iteration?
+                || {
+                    let block = construct_next_block(&vm, &private_key, &transactions, &mut rng).unwrap();
+                    num_transactions += block.transactions().iter().count() as f64;
```
ljedrz, line 65: side note: `transactions` could probably easily return `impl ExactSizeIterator` instead of `impl Iterator`, which would allow us to just use `len` here

synthesizer/benches/add_next_block.rs
```diff
@@ -0,0 +1,132 @@
...
+        );
+    }
+}
+
+fn bench_one_operation(c: &mut Criterion) {
+    // Initialize the workload.
+    let workload = one_execution_workload(NUM_COMMANDS);
+
+    #[cfg(not(any(feature = "rocks")))]
+    bench_add_next_block::<ConsensusMemory<Testnet3>>(c, "memory", workload);
```
ljedrz, line 89: this point will apply to all the uses of `Testnet3`: it is typically aliased as `CurrentNetwork` for a smaller diff whenever it changes; also, while that one is currently declared in many places, it should probably be a part of the `snarkvm` prelude (and be declared only once and reused everywhere)

synthesizer/benches/add_next_block.rs
```diff
@@ -0,0 +1,132 @@
...
+            (num_rejected / num_transactions) * 100.0
+        );
+    }
+}
+
+fn bench_one_operation(c: &mut Criterion) {
+    // Initialize the workload.
+    let workload = one_execution_workload(NUM_COMMANDS);
+
+    #[cfg(not(any(feature = "rocks")))]
```
ljedrz, line 88:
```suggestion
    #[cfg(not(feature = "rocks"))]
```

synthesizer/benches/add_next_block.rs
```diff
@@ -0,0 +1,132 @@
...
+    bench_add_next_block::<ConsensusMemory<Testnet3>>(c, "memory", workload);
+    #[cfg(any(feature = "rocks"))]
```
ljedrz, line 90:
```suggestion
    #[cfg(feature = "rocks")]
```

ljedrz: the same applies to all the other instances of this

synthesizer/benches/utilities/benchmark.rs
```diff
@@ -0,0 +1,78 @@
...
+/// Note that `name` must be unique for each benchmark.
+/// Note that `setup_operations` and `benchmark_operations` must be deterministic.
+/// Note that each operation must be independent of the others.
+pub trait Benchmark<N: Network> {
+    /// The name of the benchmark.
+    fn name(&self) -> String;
+    /// Batches of operations to be run to setup the benchmark.
+    fn setup_operations(&mut self) -> Vec<Vec<Operation<N>>>;
+    /// Operations to be run for the benchmark.
+    fn benchmark_operations(&mut self) -> Vec<Operation<N>>;
```
ljedrz, line 77: do these methods require mutability? I haven't checked every instance, but the ones I've gone through would work with just `&self`

synthesizer/benches/utilities/workload.rs
```diff
@@ -0,0 +1,182 @@
...
+        root.push(&name);
+        // Construct the object store.
+        let object_store = ObjectStore::new(root)?;
+        // Construct the workload.
+        Ok(Self { name, benchmarks, object_store })
+    }
+
+    /// Adds a benchmark to the workload.
+    pub fn add(&mut self, benchmark: Box<dyn Benchmark<Testnet3>>) {
+        self.benchmarks.push(benchmark);
```
ljedrz, line 90:
```suggestion
    pub fn add<T: Benchmark<Testnet3>>(&mut self, benchmark: T) {
        self.benchmarks.push(Box::new(benchmark) as Box<dyn Benchmark<Testnet3>>);
```
I think this should work; it would remove plenty of boilerplate at the callsites

synthesizer/benches/deployment.rs
```diff
@@ -0,0 +1,174 @@
...
+                    for (mapping_name, _) in program.mappings() {
+                        vm.finalize_store().remove_mapping(program.id(), mapping_name).unwrap();
+                        vm.process().write().remove_program(program.id()).unwrap();
+                    }
+                    vm.finalize_store().remove_program(program.id()).unwrap();
+                }
+            }
+            _ => unreachable!("The benchmark should only contain deployment transactions."),
+        }
+    }
```
ljedrz, line 54: it might be a good idea to add some assertion(s) here that would ensure that the finalize store is clean

synthesizer/benches/finalize.rs
```diff
@@ -0,0 +1,123 @@
...
+
+    // Benchmark each of the programs.
+    for (name, transactions) in benchmark_transactions {
+        if transactions.is_empty() {
+            println!("Skipping benchmark {} because it has no transactions.", name);
+            continue;
+        }
+
+        let mut num_transactions = 0f64;
+        let mut num_rejected = 0f64;
```
ljedrz, line 52: nit: I'd rather keep them as integers and only cast to `f64` when summarizing

synthesizer/benches/deployment.rs
```diff
@@ -0,0 +1,174 @@
...
+        // Construct the `Transactions` object.
+        let txns = vm.speculate(transactions.iter()).unwrap();
+
+        // Benchmark `finalize`.
+        c.bench_function(&format!("{header}/{name}/finalize"), |b| {
+            b.iter_batched(
+                || {
+                    vm.finalize_store().start_atomic();
+                    clean_finalize_state(&vm, &transactions);
+                    vm.finalize_store().finish_atomic().unwrap();
```
ljedrz, line 90: it might make sense to put `start_atomic` and `finish_atomic` inside `clean_finalize_state`

synthesizer/benches/utilities/benchmark.rs
```diff
@@ -0,0 +1,78 @@
...
+/// This is a wrapper around a vector of transactions, used to implement `FromBytes` and `ToBytes`.
+pub struct BenchmarkTransactions(pub Vec<Transaction<Testnet3>>);
+
+impl FromBytes for BenchmarkTransactions {
+    fn read_le<R: Read>(mut reader: R) -> Result<Self>
+    where
+        Self: Sized,
+    {
+        // Read the number of transactions.
+        let num_transactions = u64::read_le(&mut reader)? as usize;
```
ljedrz, line 44: `u64` seems like much, `u32` should probably suffice?

synthesizer/benches/utilities/mod.rs
```diff
@@ -0,0 +1,145 @@
...
+
+    // Initialize the genesis block.
+    let genesis = Block::genesis(&vm, private_key, rng).unwrap();
+
+    // Fetch the unspent records.
+    let records = genesis.transitions().cloned().flat_map(Transition::into_records).collect::<IndexMap<_, _>>();
+
+    // Select a record to spend.
+    let view_key = ViewKey::try_from(private_key).unwrap();
+    let record = records.values().next().unwrap().decrypt(&view_key).unwrap();
```
ljedrz, line 67:
```suggestion
    let record = records[0].decrypt(&view_key).unwrap();
```

synthesizer/benches/utilities/mod.rs
```diff
@@ -0,0 +1,145 @@
...
+    // Get the most recent block.
+    let block_hash =
+        vm.block_store().get_block_hash(*vm.block_store().heights().max().unwrap().borrow()).unwrap().unwrap();
+    let previous_block = vm.block_store().get_block(&block_hash).unwrap().unwrap();
+
+    // Construct the new block header.
+    let transactions = vm.speculate(transactions.iter())?;
+    // Construct the metadata associated with the block.
+    let metadata = Metadata::new(
+        Testnet3::ID,
```
ljedrz, line 92: side note: I wonder why `Metadata` carries `network` (and possibly the other `Testnet3::` values) at all - it can be derived ad-hoc from its type; I'm assuming for the simplicity of serialization

synthesizer/benches/utilities/object_store.rs
```diff
@@ -0,0 +1,217 @@
...
+        let bytes = object.to_bytes_le()?;
+        std::fs::write(&full_path, bytes)?;
+        Ok(())
+    }
+
+    pub fn contains<P: AsRef<Path>>(&self, path: P) -> bool {
+        self.root.join(path).exists()
+    }
+
+    pub fn clear(&mut self) -> Result<()> {
```
ljedrz, line 67:
```suggestion
    pub fn clear(&self) -> Result<()> {
```

ljedrz: I only noticed it now, but the same aplies to the other methods; unless, perhaps, they can be called concurrently and it's to ensure exclusive access? in that case this should ideally be documented

synthesizer/benches/utilities/object_store.rs
```diff
@@ -0,0 +1,217 @@
...
+    // Helper function to get the balance of a `credits.aleo` record.
+    let get_balance = |record: &Record<Testnet3, Plaintext<Testnet3>>| -> u64 {
+        match record.data().get(&Identifier::from_str("microcredits").unwrap()).unwrap() {
+            Entry::Private(Plaintext::Literal(Literal::U64(amount), ..)) => **amount,
+            _ => unreachable!("Invalid entry type for credits.aleo."),
+        }
+    };
+
+    // Initialize a counter for each block added to the VM.
+    let mut block_counter = 0u64;
```
ljedrz, line 111: since block height is `u32`, this doesn't need to be any larger, no?

synthesizer/benches/utilities/object_store.rs
```diff
@@ -0,0 +1,217 @@
...
+
+    println!("Splitting the initial fee record into {} fee records.", num_fee_records);
+
+    // Construct fee records for the workload, storing the relevant data in the object store.
+    let balance = get_balance(&record);
+    let mut fee_records = vec![(record, balance)];
+    let mut fee_counter = 1;
+    for _ in 0..num_levels_of_splits {
+        let mut transactions = Vec::with_capacity(fee_records.len());
+        for (fee_record, balance) in fee_records.drain(..).collect_vec() {
```
ljedrz, line 121:
```suggestion
        for (fee_record, balance) in fee_records.drain(..) {
```

ljedrz, line 121:
```suggestion
        for (fee_record, balance) in fee_records.drain(..) {
```

synthesizer/benches/utilities/object_store.rs
```diff
@@ -0,0 +1,217 @@
...
+        object_store.insert(format!("block_{}", block_counter), &block).unwrap();
+        vm.add_next_block(&block).unwrap();
+        block_counter += 1;
+    }
+
+    println!("Constructed fee records for the workload.");
+
+    // A helper to construct transactions from an operation.
+    let mut construct_transaction = |operation: &Operation<Testnet3>, rng: &mut TestRng| {
+        match &operation {
```
ljedrz, line 146:
```suggestion
        match operation {
```

synthesizer/benches/utilities/workload.rs
```diff
@@ -0,0 +1,182 @@
...
+    }
+
+    /// Returns the name of the workload.
+    pub fn name(&self) -> &String {
```
ljedrz, line 94:
```suggestion
    pub fn name(&self) -> &str {
```

synthesizer/benches/utilities/object_store.rs
```diff
@@ -0,0 +1,217 @@
...
+impl ObjectStore {
+    pub fn new(root: PathBuf) -> Result<Self> {
+        // Create the root directory if it does not exist.
+        if !root.try_exists()? {
+            std::fs::create_dir_all(&root)?;
+        }
+        Ok(Self { root })
+    }
+
+    pub fn get<O: FromBytes, P: AsRef<Path>>(&mut self, path: P) -> Result<O> {
```
ljedrz, line 47: ideally this should return a `Result<Option<O>>` to distinguish errors from misses

synthesizer/benches/utilities/object_store.rs
```diff
@@ -0,0 +1,217 @@
...
+        if !full_path.parent().expect("parent directory exists").exists() {
+            std::fs::create_dir_all(full_path.parent().expect("parent directory exists"))?;
+        }
+        let bytes = object.to_bytes_le()?;
+        std::fs::write(&full_path, bytes)?;
+        Ok(())
+    }
+
+    pub fn contains<P: AsRef<Path>>(&self, path: P) -> bool {
+        self.root.join(path).exists()
```
ljedrz, line 64: I wonder if an interrupted run couldn't result in a file existing, but having mangled contents; that would cause `contains` to return `true`, but a call to `get` would fail; even though it's slower, it would be safer for this method to call `get` instead

synthesizer/benches/utilities/workload.rs
```diff
@@ -0,0 +1,182 @@
...
+            for benchmark in &mut self.benchmarks {
+                let setup_ops = benchmark.setup_operations();
+                let benchmark_ops = benchmark.benchmark_operations();
+                setup_operations.push(setup_ops);
+                benchmark_operations.push((benchmark.name(), benchmark_ops));
+            }
+
+            // Fold the batches of setup operations across all benchmarks into a single sequence of batches.
+            let max_num_batches = setup_operations.iter().map(|operations| operations.len()).max().unwrap_or(0);
+            let mut aggregated_setup_operations = iter::repeat_with(Vec::new).take(max_num_batches).collect_vec();
```
ljedrz, line 135:
```suggestion
            let mut aggregated_setup_operations = vec![vec![]; max_num_batches];
```
nit: simplicity

synthesizer/benches/utilities/workload.rs
```diff
@@ -0,0 +1,182 @@
...
+            for setup_ops in setup_operations {
+                for (i, operations) in setup_ops.into_iter().enumerate() {
+                    aggregated_setup_operations[i].extend(operations);
+                }
+            }
```
ljedrz, line 140: am I misreading this, or is `aggregated_setup_operations` the same as `setup_operations`? even if it's a different collection, `let aggregated_setup_operations = setup_ops.into_iter().collect();` would suffice here, no?

synthesizer/benches/utilities/workload.rs
```diff
@@ -0,0 +1,182 @@
...
+            // Sample the private key.
+            let private_key = PrivateKey::<Testnet3>::new(&mut rng).unwrap();
+
+            // Initialize the VM.
+            let (vm, _) = initialize_vm::<C, _>(&private_key, &mut rng);
+
+            // Load the blocks.
+            let num_blocks: u64 = self.object_store.get("num_blocks").unwrap();
+            let blocks: Vec<Block<Testnet3>> =
+                (0..num_blocks).map(|i| self.object_store.get(&format!("block_{}", i)).unwrap()).collect_vec();
```
ljedrz, line 161:
```suggestion
            let blocks = (0..num_blocks).map(|i| self.object_store.get(&format!("block_{}", i)).unwrap());
```
there's no need to collect here

synthesizer/benches/utilities/workload.rs
```diff
@@ -0,0 +1,182 @@
...
+
+            // Add the blocks to the VM.
+            for (i, block) in blocks.iter().enumerate() {
```
ljedrz, line 164:
```suggestion
            for block in blocks {
```

synthesizer/src/process/mod.rs
```diff
@@ -102,6 +102,14 @@ impl<N: Network> Process<N> {
         Ok(())
     }
 
+    /// Removes a program from the process.
+    #[cfg(feature = "testing")]
+    #[inline]
+    pub fn remove_program(&mut self, program_id: &ProgramID<N>) -> Result<()> {
+        self.stacks.remove(program_id);
+        Ok(())
+    }
```
ljedrz, line 111:
```suggestion
    pub fn remove_program(&mut self, program_id: &ProgramID<N>) {
        self.stacks.remove(program_id);
    }
```
no need for the `Result`

ljedrz, line 111:
```suggestion
    pub fn remove_program(&mut self, program_id: &ProgramID<N>) {
        self.stacks.remove(program_id);
    }
```
no need for the `Result`

ljedrz: github seems to have glitched and duplicated my comment

synthesizer/benches/utilities/object_store.rs
ljedrz: github seems to have glitched and duplicated my comment

Issue #1528, opened 2023-05-11
[Bug] Change of version number size is not reflected in previously generated parameters and loader


weikengchen:
## 🐛 Bug Report

Recall, that in https://github.com/AleoHQ/snarkVM/commit/f52ea0bcd7f407f39dbf581b9bb3868ae1a91b7b, the version size is changed from u16 to u8.

As a result, all the new parameters generated, for example, from "credits.sh", are one byte shorter. Previously, the verifying key is 3787, now it is all 3786.

Nevertheless, the remote parameters are still using the old format. This is okay since the loader is still removing the first two bytes before deserialization.  

See parameters/src/testnet3/mod.rs:99

```
#[macro_export]
macro_rules! insert_key {
    ($map:ident, $string:tt, $type:ident<$network:ident>, ($name:tt, $circuit_key:expr)) => {{
        // Load the circuit key bytes.
        let key_bytes: Vec<u8> = $circuit_key.expect(&format!("Failed to load {} bytes", $string));
        // Recover the circuit key.
        let key = $type::<$network>::from_bytes_le(&key_bytes[2..]).expect(&format!("Failed to recover {}", $string));
        // Insert the circuit key.
        $map.insert($name.to_string(), std::sync::Arc::new(key));
    }};
}
```

This, however, causes a problem if a user is resampling the parameters.

The bug is located after checking a few files. We are fixing this bug because we are benchmarking Aleo.

## Steps to Reproduce

- Remove local parameter copies in ~/.aleo/*.prover.* and ~/.aleo/*.verifier.*
- Rerun `credits.sh`

## Expected Behavior

New parameters should fail to be loaded, saying invalid field element.

## Your Environment

snarkVM Version at 1442057 [1442057](https://github.com/zkspeedtest/snarkVM/commit/1442057c51e74e4eb0c41faf6242eeda6fc2bc58)

howardwu:
Fixed in #1529

PR #1529, opened 2023-05-11
Fix serialization of parameter version

raychu86:
## Motivation

This PR addresses #1528 by updating the serialization of the stored `credis.aleo` and `inclusion` keys. This change is required to match the unification of version types (u8). In this particular case, an extra byte was removed from each of the stored parameters, because we went from u16 to u8.

Thanks to @weikengchen and @vicsn for finding this discrepancy.


howardwu (approved):
LGTM

PR #1531, opened 2023-05-12
[Feature] Allow passing a `struct` as arguments to `finalize`.

d0cd:
This PR removes a restriction, allowing users to pass registers containing `structs` into `finalize` command.


synthesizer/src/process/stack/execute.rs
```diff
@@ -341,13 +341,7 @@ impl<N: Network> Stack<N> {
                     // Ensure the value is a literal (for now).
```
howardwu:
```suggestion
                    // Ensure the value is a literal or a struct.
```

howardwu (approved):
LGTM

PR #1535, opened 2023-05-12
[Test] Equivalence tests for `assert`.

d0cd:
This PR,
- generalizes `sample_finalize_registers` to accept any function name.
-  adds equivalence tests (between `evaluate`, `execute`, and `finalize`) for the `assert` operation.

Depends on #1534.

PR #1536, opened 2023-05-12
[Test] Equivalence tests for `is`.

d0cd:
This PR, adds equivalence tests (between `evaluate`, `execute`, and `finalize`) for the `is` operation.

Depends on #1535.


howardwu (approved):
LGTM

PR #1537, opened 2023-05-12
[Refactor] `cast`.

d0cd:
This PR, dedups code in the `cast` operation.

Depends on #1536.


howardwu (approved):
LGTM. I think we could even have a unique method for each `type` that we `cast` to (and one for the `console` and one for the `circuit` variant).

PR #1539, opened 2023-05-12
Use fixed version for snarkVM imports

raychu86:
## Motivation

This enforces the use of a fixed version for each imported snarkVM crate.


howardwu (approved):
LGTM

PR #1540, opened 2023-05-13
Adds method `to_ed25519` on Aleo private key

howardwu:
## Motivation

This PR adds method `to_ed25519` on Aleo private key.

The approach taken in principle is as follows:
```
Ed25519SecretKey( Hash( [ DOMAIN || sk_sig || r_sig || counter ] ) )
```
where `sk_sig` and `r_sig` compose the Aleo private key.


joske (approved):

niklaslong:
Excited to see this moving forwards, thank you @howardwu! 

I would recommend avoiding a dependency on `fastcrypto` for a few reasons. It's huge (as is visible in the lockfile diff) and it can't be trimmed down by depending on specific features. Besides, we'll be able to remove it completely from NB once this PR is in and there's no other reason to keep it. 

There are a few lighter replacements that I can see at the moment:
1. https://crates.io/crates/ed25519-dalek
2. https://crates.io/crates/ed25519 (probably not needed if we use the first option)
3. https://crates.io/crates/ed25519-consensus (as I understand it, developped for Zcash because of the signature verification concerns outlined in this post: https://hdevalence.ca/blog/2020-10-04-its-25519am)

howardwu:
@niklaslong I'm all for that. Just to check, does using `ed25519-dalek` introduce complexities for type conversions into your consensus work for snarkOS?

joske:
AFAIK fastcrypto uses `ed25519-dalek` internally, so I don't think it will make it any more complex.

@niklaslong correct me if I'm wrong.

niklaslong:
> does using ed25519-dalek introduce complexities for type conversions into your consensus work for snarkOS?

It shouldn't. Whatever we choose to use here, we'll use throughout. The only soft requirement I can think of is a conversion to and from bytes but that's included in `ed25519-dalek`.

console/account/src/private_key/to_ed25519.rs
```diff
@@ -0,0 +1,37 @@
...
+        // Prepare the domain separator.
+        let domain = Field::new_domain_separator(ACCOUNT_ED25519_DOMAIN);
+        // Prepare the counter.
+        let counter = Field::from_u32(counter);
+        // Construct the preimage.
+        let preimage = [domain, self.sk_sig.to_field()?, self.r_sig.to_field()?, counter];
+        // Hash the preimage.
+        let hash = N::hash_to_scalar_psd4(&preimage)?;
+        // Construct the private key.
+        Ok(fastcrypto::ed25519::Ed25519KeyPair::from_bytes(&hash.to_bytes_le()?)?)
```
Pratyush: What is the expected input domain of `from_bytes` here? The best way to perform this derivation would be to get *2* field elements in the output of Poseidon, convert these to ~512 bits of key information, and have `from_bytes` reduce that to construct the key pair

howardwu: It seems the `SecretKey` in `ed25519-dalek` is 32 bytes: https://github.com/dalek-cryptography/ed25519-dalek/blob/main/src/signing.rs#L48-L58

howardwu:
Updated this implementation to use `ed25519-dalek` and return a `Keypair`.

Pratyush (approved):

PR #1542, opened 2023-05-13
Add logic to clean missized parameter

howardwu:
## Motivation

Add logic to clean missized parameter

### DIFF
--- a/parameters/src/macros.rs
+++ b/parameters/src/macros.rs
@@ -29,6 +29,20 @@ macro_rules! checksum_error {
     };
 }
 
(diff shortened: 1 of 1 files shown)


d0cd (approved):
LGTM!
Related question, should the file also be removed if the checksum doesn't match?

howardwu:
Yeah, technically it should.

I wanted to start with just removing on size mismatches, to verify stability.

(technically, the odds a file 1) having the right size AND 2) having the wrong checksum, is very unlikely in our architecture)

PR #1545, opened 2023-05-14
Moves `Transaction` executors to `VM`

howardwu:
## Motivation

This PR builds upon #1544.

This PR moves the `Transaction` executors to `VM`, unifying their contexts and preparing the `block` folder for separation into a subcrate.

## Migration

The following is the change to syntax for construction a `Transaction`.
```rs
Before: Transaction::deploy(vm, ...)
After: vm.deploy(...)
```
```rs
Before: Transaction::execute(vm, ...)
After: vm.execute(...)
```
```rs
Before: Transaction::execute_fee(vm, ...)
After: vm.execute_fee(...)
```

PR #1546, opened 2023-05-14
Migrates finalize operation to the `block` folder

howardwu:
## Motivation

This PR builds upon #1545.

This PR migrates finalize operation to the `block` folder and updates the genesis method to be in `VM`.


d0cd (approved):
LGTM!

PR #1547, opened 2023-05-14
Introduce complete abstractions for `Stack` and `Registers`

howardwu:
## Motivation

This PR builds upon #1546.

This PR introduces complete abstractions for `Stack` and `Registers`. By using `impl Trait` on types that the `Program` cannot access, we abstract the underlying types so that `Stack` and `Registers` can be introduced at a higher level.


d0cd (approved):
LGTM!

PR #1557, opened 2023-05-17
add inject_tuple!

zhiqiangxu:
So that the coding style is more consistent with [`eject_tuple!`](https://github.com/AleoHQ/snarkVM/blob/e8466f87e4ed6f9709524df700a1279e34d59972/circuit/environment/src/traits/eject.rs#L147)

### DIFF
--- a/circuit/environment/src/traits/inject.rs
+++ b/circuit/environment/src/traits/inject.rs
@@ -70,7 +70,7 @@ impl<C: Inject<Primitive = P>, P> Inject for Vec<C> {
 
 /// A helper macro to implement `Inject` for a tuple of `Inject` circuits.
 macro_rules! inject_tuple {
-    (($t0:ident, $i0:expr), $(($ty:ident, $idx:tt)),*) => {
+    (($t0:ident, 0), $(($ty:ident, $idx:tt)),*) => {
         impl<$t0: Inject, $($ty: Inject),*> Inject for ($t0, $($ty),*) {
             type Primitive = ($t0::Primitive, $( $ty::Primitive ),*);
 
--- a/circuit/environment/src/traits/inject.rs
+++ b/circuit/environment/src/traits/inject.rs
@@ -68,44 +68,21 @@ impl<C: Inject<Primitive = P>, P> Inject for Vec<C> {
 /****** Tuples ******/
 /********************/
 
-impl<C0: Inject, C1: Inject> Inject for (C0, C1) {
-    type Primitive = (C0::Primitive, C1::Primitive);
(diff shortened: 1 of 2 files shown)


circuit/environment/src/traits/inject.rs
```diff
@@ -68,44 +68,21 @@ impl<C: Inject<Primitive = P>, P> Inject for Vec<C> {
...
-        (C0::new(mode, value.0), C1::new(mode, value.1), C2::new(mode, value.2), C3::new(mode, value.3))
+/// A helper macro to implement `Inject` for a tuple of `Inject` circuits.
+macro_rules! inject_tuple {
+    (($t0:ident, $i0:expr), $(($ty:ident, $idx:tt)),*) => {
+        impl<$t0: Inject, $($ty: Inject),*> Inject for ($t0, $($ty),*) {
+            type Primitive = ($t0::Primitive, $( $ty::Primitive ),*);
+
+            #[inline]
+            fn new(mode: Mode, value: Self::Primitive) -> Self {
+                ($t0::new(mode, value.0), $($ty::new(mode, value.$idx)),*)
```
ljedrz:
```suggestion
    (($t0:ident, $i0:tt), $(($ty:ident, $idx:tt)),*) => {
        impl<$t0: Inject, $($ty: Inject),*> Inject for ($t0, $($ty),*) {
            type Primitive = ($t0::Primitive, $( $ty::Primitive ),*);

            #[inline]
            fn new(mode: Mode, value: Self::Primitive) -> Self {
                ($t0::new(mode, value.$i0), $($ty::new(mode, value.$idx)),*)
```
I'd suggest to either do this (i.e. align and use `$i0`) or to change the first param from `($t0:ident, $i0:tt)` to `(C0, 0)` in order to be consistent in its use

ljedrz (approved):
LGTM with one comment.

circuit/environment/src/traits/inject.rs
zhiqiangxu: But `$t0::Primitive` is always the first element in the tuple whether `$i0` is `0` or not, so I'd prefer to stick to it, which is consistent with `eject_tuple`.

ljedrz: I see how this is aligned with `eject_tuple` - `$i0` unused in both; this is a good opportunity to fix it in both places, by at least changing `$i0:expr` to `0`

zhiqiangxu: Good point! Changed accordingly.

howardwu (approved):
LGTM

PR #1558, opened 2023-05-17
minor optimization of `Mode.combine`

zhiqiangxu:
Quit the loop if `current_mode` is already private, also makes the code a bit more readable.


circuit/environment/src/helpers/mode.rs
```diff
@@ -65,8 +65,11 @@ impl Mode {
         //   - If `current_mode == Mode::Public` && `next_mode == Mode::Private`, then `current_mode = next_mode`.
         //   - Otherwise, do nothing.
         for next_mode in modes {
+            if current_mode.is_private() {
```
howardwu, line 69:
```suggestion
            // If the current mode is `Mode::Private`, we can return early.
            if current_mode.is_private() {
```

howardwu (approved):
LGTM

PR #1561, opened 2023-05-17
fix comment about number of constraints

zhiqiangxu:
It seems there're only 2 constraints in [`Field.is_not_equal`](https://github.com/AleoHQ/snarkVM/blob/e8466f87e4ed6f9709524df700a1279e34d59972/circuit/types/field/src/equal.rs#L39)?


zhiqiangxu:
Oops, `let is_neq: Boolean<E> = witness!(|self, other| self != other);` is the 3rd constraint I missed...

PR #1562, opened 2023-05-18
fix sqrt_field_test

zhiqiangxu:
Should not allow `sqrt.square() == -elem`


Pratyush:
Is this because we're fixing `sqrt` to return only the positive square root?

zhiqiangxu:
> Is this because we're fixing `sqrt` to return only the positive square root?

For a quadratic residue `elem`, `elem.sqrt().square()` should always be equal to `elem` no matter whether `sqrt` returns the positive or negative result, right?

bendyarm:
@zhiqiangxu  I agree with your reply.  @Pratyush  I think this was just a typo.  I will submit an issue for positive/even square root if @d0cd hasn't already.

Pratyush:
Oh oops, I misread. Completely agree that this fix makes sense.

Pratyush (approved):

howardwu (approved):
LGTM, thanks for catching

PR #1566, opened 2023-05-19
[Test] Test framework for program execution.

d0cd:
This PR extends the test framework to test program execution.
An example of a program execution test is:
```
/*
- [hello, 0u32, 1u32]
- [hello, 1u32, 2u32]
*/

program hello.aleo;
function hello:
    input r0 as u32.public;
    input r1 as u32.private;
    add r0 r1 into r2;
    output r2 as u32.private;
```
The framework will test `hello` with `[0u32, 1u32]` and `[1u32, 2u32]` and check the output against the corresponding expectation file.

To regenerate expectations, set the `REWRITE_EXPECTATIONS` flag.
To filter tests, set the `FILTER_TEST` flag with the substring to filter by.


howardwu (approved):
LGTM

PR #1577, opened 2023-05-24
Fix atomic checkpointing

raychu86:
## Motivation

This PR fixes three issues found in storage map checkpointing: 

1. Incorrect checkpoint truncation and state reverting
2. Incorrect number of rewinds (we weren't rewinding to the proper point)
3. Atomic rewinds incorrectly sets the `batch_in_progress` flag to false.

The first issue was solved by updating the `atomic_batch` from an IndexMap to a `Vec`. This Vec incurs additional cost during lookups/iterators, however it lets us properly revert state. 

The second issue was solved by introducing the concept of a `checkpoint_milestone` and `checkpoint_depth`, that will `atomic_rewind` checkpoints until the declared checkpoint.

The third issue was solved very simply, by removing the incorrect logic in `atomic_rewind`

In the VM finalize calls, a new `atomic_wrapped_batch_scope` is being used to utilize the checkpoint milestones for individual transactions within `vm.speculate` and `vm.finalize`

## Test Plan

Individual tests have been added to the `MemoryMap` and `DataMap` to test that the checkpoints were set correctly and that the new `atomic_wrapped_batch_scope` macro works correctly.

An integration test was also added to ensure that the VM state is still consistent when processing accepted and rejected transactions.


synthesizer/src/store/helpers/memory/internal/mod.rs
```diff
@@ -37,8 +37,9 @@ pub struct MemoryMap<
     // with fast lookups and the keys could be typed (i.e. just `K` instead of `Vec<u8>`).
     map: Arc<RwLock<BTreeMap<Vec<u8>, V>>>,
     batch_in_progress: Arc<AtomicBool>,
-    atomic_batch: Arc<Mutex<IndexMap<K, Option<V>>>>,
+    atomic_batch: Arc<Mutex<Vec<(K, Option<V>)>>>,
     checkpoint: Arc<Mutex<VecDeque<usize>>>,
```
ljedrz, line 41: this is not in this PR, but it seems to me that `checkpoint` could be a plain `Vec` instead of `VecDeque`

synthesizer/src/store/helpers/memory/internal/mod.rs
```diff
@@ -37,8 +37,9 @@ pub struct MemoryMap<
...
+    checkpoint_depth: Arc<Mutex<Option<usize>>>,
```
ljedrz, line 42: could we offset the depths by `1` and use an `Arc<AtomicUsize>` instead? `0` would indicate the current `None`

synthesizer/src/store/helpers/memory/internal/mod.rs
```diff
@@ -163,10 +172,40 @@ impl<
...
 
-        // If the atomic batch is now empty, set the atomic batch flag to `false`.
-        if atomic_batch.is_empty() {
-            self.batch_in_progress.store(false, Ordering::SeqCst);
+        // Decrement the checkpoint depth, if it is set.
+        let mut checkpoint_depth = self.checkpoint_depth.lock();
+        if let Some(depth) = *checkpoint_depth {
+            // Decrement the checkpoint depth, if it is greater than zero.
+            if depth > 0 {
+                *checkpoint_depth = Some(depth.saturating_sub(1));
```
ljedrz, line 180:
```suggestion
                *checkpoint_depth = Some(depth - 1);
```
we know we're `> 0`

synthesizer/src/store/helpers/memory/internal/mod.rs
```diff
@@ -147,6 +150,12 @@ impl<
     fn atomic_checkpoint(&self) {
         // Push the current length of the atomic batch to the checkpoint stack.
         self.checkpoint.lock().push_back(self.atomic_batch.lock().len());
+
+        // Increment the checkpoint depth, if it is set.
+        let mut checkpoint_depth = self.checkpoint_depth.lock();
+        if let Some(depth) = *checkpoint_depth {
+            *checkpoint_depth = Some(depth.saturating_add(1));
```
ljedrz, line 157: what's the relationship between the length of `checkpoint` and `checkpoint_depth`? is it not the same?

synthesizer/src/store/helpers/memory/internal/mod.rs
```diff
@@ -947,4 +998,149 @@ mod tests {
...
+        assert_eq!(map.checkpoint.lock().back(), None);
+
+        // Ensure that the map value is correct.
+        assert_eq!(*map.iter_confirmed().next().unwrap().1, "1");
+    }
+
+    #[test]
+    fn test_atomic_wrapped_batch_scope() -> Result<()> {
+        // Initialize a map.
+        let map: MemoryMap<usize, String> = Default::default();
```
ljedrz, line 1056: we could use a test with multiple maps, which would bring us closer to real-life conditions; however, this is more applicable to the `rocksdb` impl

synthesizer/src/store/helpers/rocksdb/internal/map.rs
```diff
@@ -124,12 +132,44 @@ impl<
...
     }
 
+    ///
+    /// Removes pending operations by calling `atomic_rewind`
+    /// until the map returns to the start of the milestone.
+    ///
+    fn atomic_rewind_milestone(&self) {
+        // Rewind the atomic batch until the checkpoint depth is zero.
+        let x = self.checkpoint_depth.lock().unwrap_or(0);
+        println!("CHECKPOINT DEPTH: {:?}-------------------------------------------------------------", x);
```
ljedrz:
```suggestion
```

ljedrz (commented):
This setup is becoming rather complex. My suggestion, other than the comments that I left, would be to only have a single checkpointing mechanism, as we now have 2 of them - do we even utilize the first one (the one that performs a checkpoint whenever `atomic_batch_scope` is called)? I think all the checkpoints should be manual instead, like in this PR.

raychu86:
Closed in favor of #1592

PR #1578, opened 2023-05-24
[nit] avoid duplicate check

zhiqiangxu:
Just stumble upon this, the subsequent `match case` already has a default case handling the mismatch.

### DIFF
--- a/synthesizer/src/process/stack/finalize_types/initialize.rs
+++ b/synthesizer/src/process/stack/finalize_types/initialize.rs
@@ -303,10 +303,6 @@ impl<N: Network> FinalizeTypes<N> {
                 );
             }
             Opcode::Assert(opcode) => {
-                // Ensure the instruction belongs to the defined set.
-                if !["assert.eq", "assert.neq"].contains(&opcode) {
-                    bail!("Instruction '{instruction}' is not for opcode '{opcode}'.");
-                }
                 // Ensure the instruction is the correct one.
                 match opcode {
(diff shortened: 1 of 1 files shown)


howardwu:
Thanks for flagging this @zhiqiangxu. I suppose there is a subtlety in this check.

The initial `if` checks the `opcode` is correct, while the latter `match case` checks the `Instruction::{Variant}` is correct.

There could be a case (in theory) where the `Instruction::opcode` does not match `Instruction::{Variant}` (due to a coding error).

A few options in my mind to address this:
1. We can make the `if` check `cfg(debug_assertions)` so it only triggers in development mode.
2. We can manually implement unit tests that enforce the `opcode` and `enum variant` match correctly.

zhiqiangxu:
Option 1 may be overkill as it's more about coding style, duplicating is not good :)

d0cd:
Another option is to fold the check for the opcode and the variant into one match arm.
Something like:
```
  Opcode::Assert(opcode) => {
      // Ensure the instruction is the correct one.
      match (opcode, instruction) {
          ("assert.eq", Instruction::AssertEq(..))
          | ("assert.neq", Instruction::AssertNeq(..)) => Ok(()),
          _ => bail!("Instruction '{instruction}' is not for opcode '{opcode}'."),
      }
  }
 ```

howardwu:
Closing as these ended up being removed in a different PR, thanks!

PR #1583, opened 2023-05-24
Update clap to 4.3

ljedrz:
This PR updates `clap` to the current stable release.

note: this currently requires the use of a feature named `unstable-styles`, suggesting that it might require further changes in the future.

### DIFF
--- a/vm/cli/commands/run.rs
+++ b/vm/cli/commands/run.rs
@@ -20,10 +20,10 @@ pub const LOCALE: &num_format::Locale = &num_format::Locale::en;
 #[derive(Debug, Parser)]
 pub struct Run {
     /// The function name.
-    #[clap(parse(try_from_str))]
+    #[clap(long)]
     function: Identifier<CurrentNetwork>,
     /// The function inputs.
-    #[clap(parse(try_from_str))]
+    #[clap(long)]
     inputs: Vec<Value<CurrentNetwork>>,
     /// Uses the specified endpoint.
     #[clap(long)]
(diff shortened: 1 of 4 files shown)


vm/cli/commands/run.rs
```diff
@@ -20,10 +20,10 @@ pub const LOCALE: &num_format::Locale = &num_format::Locale::en;
 #[derive(Debug, Parser)]
 pub struct Run {
     /// The function name.
-    #[clap(parse(try_from_str))]
+    #[clap(long)]
```
howardwu, line 23: Why did we remove `parse(try_from_str)`?

The function and inputs fields appear to be command line arguments that specify the function name and its inputs, respectively. Before the change, the values of these arguments were parsed from a string using the FromStr trait. Now, with the change to #[clap(long)], these arguments can be specified using long flags.

However, this change will likely alter the behavior of the command line interface. Previously, the function and inputs arguments probably accepted values directly after the flag (e.g., --function some_function and --inputs some_input). With the change to #[clap(long)], these arguments will likely just act as flags and not accept any values. This could lead to incorrect behavior if the program still expects function and inputs to have values.

For instance, if function and inputs are expected to be provided by the user, the appropriate change might be to use #[clap(long, parse(try_from_str))], which would both allow them to be specified with a long flag and parse their values from a string. Without seeing more of the context or the rest of the program, it's difficult to say for certain whether this change is correct.

ljedrz: The reason for removing `try_from_str` is that it no longer exists; as for feeding further arguments, I'd tested this on snarkOS, and it seemed to work just fine; you can also check the updated clap [derive tutorial](https://docs.rs/clap/latest/clap/_derive/_tutorial/index.html) and [derive reference](https://docs.rs/clap/latest/clap/_derive/index.html) to see examples of such use.

ljedrz:
The CI failure is unrelated.

howardwu (approved):
LGTM

PR #1592, opened 2023-05-25
Fix atomic checkpointing and rewinds

### LINKED ISSUE #1548: [Bug] Rejected transitions are changing the finalize storage
## 🐛 Bug Report

(This is a security bug, but as we are still on testnet, I'm reporting here)

snarkVM doesn't rollback changes when trying to execute the finalize operations of rejected executions.

## Steps to Reproduce

Already reproduced by someone on the live testnet with [this program](https://explorer.hamp.app/program?id=testing123.aleo). The first rejected transition changed the storage.

## Expected Behavior

Rejected transitions should not change the storage.

## Your Environment

Current testnet 3

## Details

I'm trying to write an interpreter of Aleo instructions to get the mapping values, until I encountered some issues on [this transition](https://explorer.hamp.app/transition?id=as1n9h7j45z239hjjgp6y0fgt24rzpva8ngkgxh35q0tqkpwv0yquys2rgnay) (I've skipped the [rejected transition](https://explorer.hamp.app/transition?id=as1mrh9nffvgd6e4jzyu4asvrlq84qgjwplpvsc7w5hvfkzwj8t5czq0eu4ac) as it should not affect the mapping storage).

Note the rejected transition has input of `10u8, 100u8`, so the second set of `add` operations will fail (tried to do `200u8 + 100u8`). However at this point, the last `set` command has stored the value `{count: 20u8, data: 200u8}` in the mapping, and the rejection of the transition didn't revert the storage changes.

In theory, the next accepted transition would start at a clean mapping storage, but there are only 2 finalize commands in that transaction, instead of 3 (come from `get.or_init` and two `set`s) (didn't show them on my explorer, so [raw block](https://vm.aleo.org/api/testnet3/block/14931) here). The `value_id` of the first command should verify it, as the value is `{count: 21u8, data: 202u8}`:

```rust
use snarkvm_console_program::{Field, Identifier, Literal, Plaintext, Value};
use snarkvm_console_program::FromStr;
use indexmap::IndexMap;
use snarkvm_console_network::{Network, Testnet3, ToBits};

fn main() {
    let value = Value::Plaintext(Plaintext::Struct(
        IndexMap::from_iter(
            vec![
                (
                    Identifier::from_str("count").unwrap(),
                    Plaintext::<Testnet3>::Literal(
                        Literal::<Testnet3>::from_str("21u8").unwrap(),
                        Default::default()
                    )
                ),
                (
                    Identifier::from_str("data").unwrap(),
                    Plaintext::<Testnet3>::Literal(
                        Literal::<Testnet3>::from_str("202u8").unwrap(),
                        Default::default()
                    )
                ),
            ]
                .into_iter(),
        ),
        Default::default(),
    ));
    println!("{:?}", value);
    let key_id = Field::<Testnet3>::from_str("6679816225070869277642545706246168867612829149463851804023197862921272508769field").unwrap();
    let value_id = <Testnet3 as Network>::hash_bhp1024(&(key_id, <Testnet3 as Network>::hash_bhp1024(&value.to_bits_le()).unwrap()).to_bits_le()).unwrap();
    println!("{:?}", value_id);
}
```
prints
```
{
  count: 21u8,
  data: 202u8
}
8429669485387850725534994500954642953635506875168032440768836063535078757651field
```
which matches the value in the block.

It seems the atomic batch is not working correctly, but I didn't figure out why it didn't rewind on failed `finalize()` calls.

raychu86:
## Motivation

This is an alternative and cleaner approach to #1577.

This PR fixes three issues found in storage map checkpointing: 

1. Incorrect checkpoint truncation and state reverting
2. Incorrect number of rewinds (we weren't rewinding to the proper point)
3. Atomic rewinds incorrectly sets the `batch_in_progress` flag to false.

The first issue was solved by updating the `atomic_batch` from an IndexMap to a `Vec`. This Vec incurs additional cost during lookups/iterators, however it lets us properly revert state. 

The second issue was solved by clearing the current checkpoint in `atomic_batch_scope` once a value is returned. This ensures that in the nested case, we rewind to the start of your current `atomic_batch_scope` no matter what happens in your internal operations.

The third issue was solved very simply, by removing the incorrect logic in `atomic_rewind`

## Test Plan

Individual tests have been added to the `MemoryMap` and `DataMap` to test that the checkpoints were set correctly and that the new checkpoint logic works as expected.

An integration test was also added to ensure that the VM state is still consistent when processing accepted and rejected transactions.


ljedrz (approved):
Looks reasonable; note: `checkpoint` doesn't need to be a `VecDeque`, seems like a `Vec` would suffice (and make it clear that checkpoints work like a stack).

synthesizer/src/store/helpers/memory/internal/mod.rs
```diff
@@ -162,11 +170,6 @@ impl<
 
         // Remove all operations after the checkpoint.
         atomic_batch.truncate(checkpoint);
-
-        // If the atomic batch is now empty, set the atomic batch flag to `false`.
-        if atomic_batch.is_empty() {
-            self.batch_in_progress.store(false, Ordering::SeqCst);
-        }
```
ljedrz, line 169: I'm only concerned about this: I realize that this is because a single rewind might not mean that the atomic batch is over; however, if the calls of `atomic_batch_scope` end with `?`, cascading to the outer scopes (due to errors in `run_atomic_ops` resulting in `atomic_rewind` as opposed to `atomic_abort`), this could cause a situation where we end up with `batch_in_progress` set to `true` while being outside of an atomic batch. Technically we wouldn't panic on that, but it could potentially cause issues.

synthesizer/src/store/helpers/rocksdb/internal/map.rs
```diff
@@ -123,11 +131,6 @@ impl<
 
         // Remove all operations after the checkpoint.
         atomic_batch.truncate(checkpoint);
-
-        // If the atomic batch is now empty, set the atomic batch flag to `false`.
-        if atomic_batch.is_empty() {
-            self.batch_in_progress.store(false, Ordering::SeqCst);
-        }
```
ljedrz, line 130: Same comment as https://github.com/AleoHQ/snarkVM/pull/1592/files#r1206517525.

synthesizer/src/store/helpers/memory/internal/mod.rs
ljedrz: We might want to consider having `run_atomic_ops` return e.g. `Result<Option<_>>`, where the outer `Result` would indicate a storage error (in which case it would be more reasonable to call `atomic_abort` ), and the inner `Option` would indicate (in case of a `None`) that an operation didn't encounter a fatal error, but only needs to be skipped (rewound).

synthesizer/src/store/helpers/rocksdb/internal/map.rs
```diff
@@ -959,4 +973,151 @@ mod tests {
...
+
+                // Insert the key.
+                map.insert(0, "1".to_string()).unwrap();
+
+                Ok(())
+            });
+
+            // The atomic finalize should have succeeded.
+            assert!(result.is_ok());
+            // The map should still contain no items.
```
howardwu:
```suggestion
            // The map should contain 1 item.
```

synthesizer/src/store/helpers/memory/internal/mod.rs
```diff
@@ -947,4 +960,149 @@ mod tests {
...
+
+                // Update the key.
+                map.insert(0, "2".to_string()).unwrap();
+
+                bail!("This batch scope should fail.")
+            });
+
+            // Ensure that the batch scope failed.
+            assert!(result.is_err());
+            // The map should still contain no items.
```
howardwu:
```suggestion
            // The map should contain 1 item.
```

synthesizer/src/store/helpers/memory/internal/mod.rs
```diff
@@ -947,4 +960,149 @@ mod tests {
...
+
+                // Insert the key.
+                map.insert(0, "1".to_string()).unwrap();
+
+                Ok(())
+            });
+
+            // The atomic finalize should have succeeded.
+            assert!(result.is_ok());
+            // The map should still contain no items.
```
howardwu:
```suggestion
            // The map should contain 1 item.
```

synthesizer/src/store/helpers/memory/internal/mod.rs
```diff
@@ -947,4 +960,149 @@ mod tests {
...
+                    Ok(())
+                });
+                assert!(result.is_ok());
+
+                bail!("This batch scope should fail.")
+            });
+
+            // Ensure that the batch scope failed.
+            assert!(result.is_err());
+            // The map should still contain no items.
```
howardwu:
```suggestion
            // The map should contain 1 item.
```

synthesizer/src/store/helpers/rocksdb/internal/map.rs
```diff
@@ -908,8 +922,8 @@ mod tests {
             assert!(map.iter_confirmed().next().is_none());
             // The pending batch should contain NUM_ITEMS items.
```
howardwu:
```suggestion
            // The pending batch should contain NUM_ITEMS / 2 items.
```

synthesizer/src/store/helpers/memory/internal/mod.rs
```diff
@@ -897,8 +910,8 @@ mod tests {
             assert!(map.iter_confirmed().next().is_none());
             // The pending batch should contain NUM_ITEMS items.
```
howardwu:
```suggestion
            // The pending batch should contain NUM_ITEMS / 2 items.
```

synthesizer/src/store/helpers/rocksdb/internal/map.rs
```diff
@@ -959,4 +973,151 @@ mod tests {
...
+
+                // Update the key.
+                map.insert(0, "2".to_string()).unwrap();
+
+                bail!("This batch scope should fail.")
+            });
+
+            // Ensure that the batch scope failed.
+            assert!(result.is_err());
+            // The map should still contain no items.
```
howardwu:
```suggestion
            // The map should contain 1 item.
```

synthesizer/src/store/helpers/rocksdb/internal/map.rs
```diff
@@ -959,4 +973,151 @@ mod tests {
...
+                    Ok(())
+                });
+                assert!(result.is_ok());
+
+                bail!("This batch scope should fail.")
+            });
+
+            // Ensure that the batch scope failed.
+            assert!(result.is_err());
+            // The map should still contain no items.
```
howardwu:
```suggestion
            // The map should contain 1 item.
```

howardwu (approved):
LGTM

howardwu:
> Looks reasonable; note: `checkpoint` doesn't need to be a `VecDeque`, seems like a `Vec` would suffice (and make it clear that checkpoints work like a stack).

I've added this change on top of this PR in #1598.

PR #1593, opened 2023-05-25
Unifies proving of `Transaction::execute` into two proofs

howardwu:
## Motivation

Unifies proving of `Transaction::execute` into two proofs


synthesizer/src/process/trace/mod.rs
```diff
@@ -0,0 +1,308 @@
...
+    }
+}
+
+impl<N: Network> Trace<N> {
+    /// Inserts the transition into the trace.
+    pub fn insert_transition(
+        &mut self,
+        input_ids: &[InputID<N>],
+        transition: &Transition<N>,
+        (proving_key, assignment): (ProvingKey<N>, Assignment<N::Field>),
```
vicsn, line 85: Nit: I would either pass them separately or define a new type

synthesizer/src/block/transaction/execution/mod.rs
```diff
@@ -27,25 +27,24 @@ pub struct Execution<N: Network> {
     transitions: IndexMap<N::TransitionID, Transition<N>>,
     /// The global state root.
     global_state_root: N::StateRoot,
-    /// The inclusion proof.
-    inclusion_proof: Option<Proof<N>>,
+    /// The proof.
+    proof: Option<Proof<N>>,
```
vicsn, line 31: Given that you now create a Trace instead of a proofless Execution, I think this does not need to be optional anymore. Just like Transition also used to have a non-optional proof. 

This will also simplify the serialization of Execution.

A reason to keep it optional is if we want to send Executions for delegated proving soon.

synthesizer/src/block/transaction/fee/mod.rs
```diff
@@ -29,15 +29,15 @@ pub struct Fee<N: Network> {
     transition: Transition<N>,
     /// The global state root.
     global_state_root: N::StateRoot,
-    /// The inclusion proof.
-    inclusion_proof: Option<Proof<N>>,
+    /// The proof.
+    proof: Option<Proof<N>>,
```
vicsn, line 33: See comment on `struct Execution`, this could be optional now.

synthesizer/src/store/transaction/execution.rs
```diff
@@ -37,7 +37,7 @@ pub trait ExecutionStorage<N: Network>: Clone + Send + Sync {
     type IDMap: for<'a> Map<'a, N::TransactionID, (Vec<N::TransitionID>, bool)>;
     /// The mapping of `transition ID` to `transaction ID`.
     type ReverseIDMap: for<'a> Map<'a, N::TransitionID, N::TransactionID>;
-    /// The mapping of `transaction ID` to `(global state root, (optional) inclusion proof)`.
+    /// The mapping of `transaction ID` to `(global state root, (optional) proof)`.
```
vicsn, line 40: Just in case you overlooked it, since there is not just an inclusion proof might want to rename: https://github.com/AleoHQ/snarkVM/pull/1510/files#diff-15129d118f13be5484e52c96b29f3e95e65f712150fed731d8ea8e02d1d31db5R43

synthesizer/src/process/trace/inclusion/mod.rs
```diff
@@ -91,44 +94,62 @@ impl<N: Network> Inclusion<N> {
...
+                // Construct the transaction leaf.
+                let transaction_leaf = TransactionLeaf::new_execution(transition_index as u16, **transition.id());
+                // Insert the leaf into the transaction tree.
+                transaction_tree.append(&[transaction_leaf.to_bits_le()])?;
+            }
         }
 
         // Ensure the global state root is not zero.
-        if *global_state_root == Field::zero() {
+        if batch_verifier_inputs.is_empty() && *global_state_root == Field::zero() {
```
vicsn, line 142: Why do we only want a non-zero `global_state_root` when `batch_verifier_inputs.is_empty()`?

synthesizer/src/process/tests.rs
```diff
@@ -225,14 +229,14 @@ fn test_process_multirecords() {
...
-    let (response, execution, _inclusion, _metrics) = process.execute::<CurrentAleo, _>(authorization, rng).unwrap();
+    let (response, _trace, _metrics) = process.execute::<CurrentAleo>(authorization).unwrap();
     let candidate = response.outputs();
     assert_eq!(3, candidate.len());
     assert_eq!(output_a, candidate[0]);
     assert_eq!(output_b, candidate[1]);
     assert_eq!(output_c, candidate[2]);
 
-    process.verify_execution::<false>(&execution).unwrap();
+    // process.verify_execution::<false>(&execution).unwrap();
```
vicsn: These should all be uncommented again

vicsn (changes requested):
I like the introduction of Trace, encapsulating the proving of inclusion and execution!

synthesizer/src/process/trace/inclusion/mod.rs
howardwu: To preserve privacy in the case that we introduce function privacy in the future.

synthesizer/src/process/tests.rs
howardwu: These can actually be removed. We weren't checking the proofs prior, hence the `false`. We merely verified that it was possible to construct the verifier inputs without fail. (there's not enough structure here to check a real proof)

synthesizer/src/block/transaction/execution/mod.rs
howardwu: It's a good point, will think about it. There used to be cases where it was necessary but I agree with the sentiment.

For delegation, we can always introduce other structs for that.

synthesizer/src/block/transaction/fee/mod.rs
howardwu: Unifying discussion to comment above.

synthesizer/src/store/transaction/execution.rs
howardwu: Yeah I debated making this change here, mainly because its quite a large changeset already.

Will address in a future PR.

synthesizer/src/process/trace/mod.rs
howardwu: Let's keep it this way for now until the `Trace` design is stabilized.

There is compelling evidence that this design is not final. (we can now propagate much of the execution logic to the authorization stage, and move the trace upstream as well)

vicsn (approved):
Approved in case you want to tackle the nits in future PRs.

synthesizer/src/process/execute.rs
```diff
@@ -208,15 +204,27 @@ impl<N: Network> Process<N> {
...
+                .entry(Locator::new(*stack.program_id(), *function.name()))
+                .or_insert((verifying_key, vec![]))
+                .1
+                .push(inputs);
+
+            lap!(timer, "Constructed the verifier inputs for a transition of {}", function.name());
         }
 
+        // Count the number of verifier instances.
+        let num_instances = verifier_inputs.values().flat_map(|(_, inputs)| inputs).count();
```
ljedrz:
```suggestion
        let num_instances = verifier_inputs.values().map(|(_, inputs)| inputs.len()).sum::<usize>();
```
this should be faster, as `count` has to consume the whole iterator one-by-one, while the lengths are immediately available

synthesizer/src/process/execute_fee.rs
```diff
@@ -56,30 +56,29 @@ impl<N: Network> Process<N> {
         #[cfg(feature = "aleo-cli")]
         println!("{}", format!(" • Calling '{}/{}'...", request.program_id(), request.function_name()).dimmed());
 
-        // Initialize the execution.
-        let execution = Arc::new(RwLock::new(Execution::new()));
-        // Initialize the inclusion.
-        let inclusion = Arc::new(RwLock::new(Inclusion::new()));
+        // Initialize the trace.
+        let trace = Arc::new(RwLock::new(Trace::new()));
```
ljedrz, line 63:
```suggestion
        let trace = Default::default();
```
(type inference should be able to handle this, same with `metrics`)

synthesizer/src/process/trace/mod.rs
```diff
@@ -0,0 +1,308 @@
...
+impl<N: Network> Trace<N> {
+    /// Initializes a new trace.
+    pub fn new() -> Self {
+        Self {
+            transitions: Vec::new(),
+            transition_tasks: HashMap::new(),
+            inclusion_tasks: Inclusion::new(),
+            inclusion_assignments: OnceCell::new(),
+            global_state_root: OnceCell::new(),
+        }
```
ljedrz:
```suggestion
        Self::default()
```

synthesizer/src/process/trace/mod.rs
```diff
@@ -0,0 +1,308 @@
...
+        // Ensure there is only 1 inclusion assignment.
+        ensure!(inclusion_assignments.len() == 1, "Expected 1 inclusion assignment for proving the fee");
+        // Retrieve the global state root.
+        let global_state_root =
+            self.global_state_root.get().ok_or_else(|| anyhow!("Global state root has not been set"))?;
+        // Retrieve the fee transition.
+        let fee_transition = match self.transitions.len() {
+            1 => &self.transitions[0],
+            _ => bail!("Expected 1 transition in the trace for proving the fee"),
+        };
```
ljedrz:
```suggestion
        let fee_transition = &self.transitions[0];
```
the the number of transitions is already guaranteed to be `1` with the initial call to `is_fee`

synthesizer/src/store/helpers/rocksdb/internal/id.rs
```diff
@@ -237,7 +236,6 @@ enum DataID {
     // Transition
     TransitionLocatorMap,
     TransitionFinalizeMap,
-    TransitionProofMap,
```
ljedrz, line 240: double-check note: this insta-breaks compatibility with current storage

ljedrz (approved):
Left a few comments, LGTM in general.

synthesizer/src/store/helpers/rocksdb/internal/id.rs
howardwu: Yup this is for the network reset.

synthesizer/src/process/trace/mod.rs
howardwu: It isn't able to be auto-derived due to `N`

synthesizer/src/block/transaction/execution/mod.rs
howardwu: https://github.com/AleoHQ/snarkVM/pull/1593/commits/c2f2e5762e45ee9aa8fb159d615cc1f5634ea340
Removed

synthesizer/src/process/tests.rs
howardwu: Removed in https://github.com/AleoHQ/snarkVM/pull/1593/commits/c2f2e5762e45ee9aa8fb159d615cc1f5634ea340

PR #1598, opened 2023-05-26
Switches out `VecDeque` for `Vec` in checkpointing

howardwu:
## Motivation

Switches out `VecDeque` for `Vec` in checkpointing


raychu86 (approved):
LGTM

PR #1599, opened 2023-05-27
[nit] remove impossible SynthesisError::AssignmentMissing

zhiqiangxu:
`a` and `b` have already been checked against `SynthesisError::AssignmentMissing` [in the beginning](https://github.com/AleoHQ/snarkVM/blob/80ccbe5161a1d5e8fa16881521cce14af8402394/algorithms/src/snark/marlin/data_structures/test_circuit.rs#L42), so lines after can just do `.unwrap()` safely.


howardwu:
I think it doesn't hurt to keep the errors as-is. I agree neither will be triggered, however if the logic is changed in the future, we have a clear error type prepared.

zhiqiangxu:
Yeah it's just a nit, I'm also OK with keeping it as is.

PR #1600, opened 2023-05-28
remove required cli flags, support multiple arguments, update readme

collinc97:
The CLI changes merged in https://github.com/AleoHQ/snarkVM/pull/1583 require that the `--function` and `--inputs` long flags be present when calling the `run` command of the CLI. In addition, clap does not support multiple arguments by default so you cannot pass multiple `inputs` arguments to the CLI at all right now. https://docs.rs/clap/latest/clap/_derive/_tutorial/index.html#options

The fix is very simple - remove the attributes. This supports the previous and intended syntax to run a program as well as multiple inputs.
```bash
snarkvm run hello 1u32 2u32
```


howardwu (approved):
LGTM

PR #1601, opened 2023-05-29
Add a test case for cli argument parsing

ljedrz:
This will help us ensure that any future breaking `clap` changes (some of which are not necessarily included in the CHANGELOG) don't result in unexpected behavior.

For now I only added one officially recommended test case and one for the following command:
```
snarkvm run hello 1u32 2u32 --endpoint ENDPOINT --offline
```

Happy to include more if desirable. I'll include a similar test case (or more) in the snarkOS clap update PR.

### DIFF
--- a/vm/cli/cli.rs
+++ b/vm/cli/cli.rs
@@ -62,3 +62,32 @@ impl Command {
         }
     }
 }
+
+#[cfg(test)]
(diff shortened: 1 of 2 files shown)


howardwu (approved):
LGTM

PR #1606, opened 2023-05-30
Speed up VM::from via parallelism

ljedrz:
While looking into ledger memory use, I noticed that it took plenty of time to load (~127s on my PC, with low CPU utilization), and that there was potential for a quick perf win via parallelism.

This PR reduces the current time needed to load the ledger by ~**80%** (down to ~25s on my PC) by parallelizing the loading of the deployments.

There may be other wins to apply to the loading process, but this simple drive-by is good enough on its own for now.

### DIFF
--- a/ledger/src/lib.rs
+++ b/ledger/src/lib.rs
@@ -144,6 +144,7 @@ impl<N: Network, C: ConsensusStorage<N>> Ledger<N, C> {
     pub fn load_unchecked(genesis_block: Block<N>, dev: Option<u16>) -> Result<Self> {
         let timer = timer!("Ledger::load_unchecked");
 
+        debug!("Loading the consensus store...");
         // Initialize the consensus store.
         let store = match ConsensusStore::<N, C>::open(dev) {
             Ok(store) => store,
--- a/ledger/src/lib.rs
+++ b/ledger/src/lib.rs
@@ -144,7 +144,7 @@ impl<N: Network, C: ConsensusStorage<N>> Ledger<N, C> {
     pub fn load_unchecked(genesis_block: Block<N>, dev: Option<u16>) -> Result<Self> {
         let timer = timer!("Ledger::load_unchecked");
 
-        debug!("Loading the ledger from storage...");
+        info!("Loading the ledger from storage...");
         // Initialize the consensus store.
         let store = match ConsensusStore::<N, C>::open(dev) {
             Ok(store) => store,
--- a/ledger/src/lib.rs
+++ b/ledger/src/lib.rs
@@ -144,7 +144,7 @@ impl<N: Network, C: ConsensusStorage<N>> Ledger<N, C> {
     pub fn load_unchecked(genesis_block: Block<N>, dev: Option<u16>) -> Result<Self> {
         let timer = timer!("Ledger::load_unchecked");
 
-        debug!("Loading the consensus store...");
+        debug!("Loading the ledger from storage...");
         // Initialize the consensus store.
         let store = match ConsensusStore::<N, C>::open(dev) {
             Ok(store) => store,
--- a/ledger/store/src/transaction/deployment.rs
+++ b/ledger/store/src/transaction/deployment.rs
@@ -30,6 +30,8 @@ use synthesizer_snark::{Certificate, VerifyingKey};
 
 use anyhow::Result;
 use core::marker::PhantomData;
+#[cfg(not(feature = "serial"))]
+use rayon::prelude::*;
 use std::borrow::Cow;
 
 /// A trait for deployment storage.
@@ -629,6 +631,12 @@ impl<N: Network, D: DeploymentStorage<N>> DeploymentStore<N, D> {
     /// Returns an iterator over the deployment transaction IDs, for all deployments.
     pub fn deployment_transaction_ids(&self) -> impl '_ + Iterator<Item = Cow<'_, N::TransactionID>> {
         self.storage.id_map().keys_confirmed()
+    }
+
+    /// Returns a parallel iterator over the deployment transaction IDs, for all deployments.
+    #[cfg(not(feature = "serial"))]
+    pub fn par_deployment_transaction_ids(&self) -> impl '_ + ParallelIterator<Item = Cow<'_, N::TransactionID>> {
+        self.storage.id_map().keys_confirmed().collect::<Vec<_>>().into_par_iter()
     }
 
     /// Returns an iterator over the program IDs, for all deployments.
--- a/ledger/store/src/transaction/deployment.rs
+++ b/ledger/store/src/transaction/deployment.rs
@@ -30,8 +30,6 @@ use synthesizer_snark::{Certificate, VerifyingKey};
 
 use anyhow::Result;
 use core::marker::PhantomData;
-#[cfg(not(feature = "serial"))]
-use rayon::prelude::*;
 use std::borrow::Cow;
 
 /// A trait for deployment storage.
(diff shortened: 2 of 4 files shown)


ljedrz:
The CI failure is unrelated.

synthesizer/src/vm/mod.rs
```diff
@@ -69,6 +71,42 @@ impl<N: Network, C: ConsensusStorage<N>> VM<N, C> {
...
+        Ok(Self { process: Arc::new(RwLock::new(process)), store })
+    }
+
+    #[cfg(not(feature = "serial"))]
+    #[inline]
+    fn load_deployments(process: &mut Process<N>, store: &ConsensusStore<N, C>) -> Result<()> {
+        // Retrieve the transaction store.
+        let transaction_store = store.transaction_store();
+        // Retrieve the stacks of the deployments from the store.
+        let stacks = transaction_store
```
Pratyush: Why not use the same code structure in both cases, by using `map` in both places? That way we can reuse the `cfg_iter` code everywhere.

ljedrz: The issue is that we don't have direct parallel iteration for database records - there's an intermediate collection into a vector. I don't how this could be achieved without introducing clunky APIs.

howardwu:
Can this be rebased, and can the implementation be cleaned up for improved readability?

ljedrz:
I've rebased the PR and collected new figures for `VM::from` for the current (testnet) ledger:

before: 19.18s
after: 3.95s (-80%)

I'm leaving it as a draft, as the question of how to tackle `serial` is still open - `cfg_iter` can't be used here (not unless we were to make some `Ledger` internals `pub`), which leaves us with either a small helper macro or `#[cfg(feature = ...))]` blocks/functions.

I'm pretty sure the performance improvement is worth it; let me know how you'd prefer to handle it and I'll finalize the PR.

synthesizer/src/vm/mod.rs
```diff
@@ -107,25 +112,37 @@ impl<N: Network, C: ConsensusStorage<N>> VM<N, C> {
...
-            load_deployment_and_imports(&mut process, transaction_store, *transaction_id)?;
+        let deployments = transaction_store
+            .par_deployment_transaction_ids()
+            .map(|transaction_id| {
+                // Load the deployment and its imports.
+                load_deployment_and_imports(&process, transaction_store, *transaction_id)
+            })
+            .collect::<Result<Vec<_>>>()?;
+
+        for (program_id, deployment) in deployments.iter().flatten() {
```
howardwu: I think this iterator doesn't need to collect.

This will consume a ton of memory (possibly overflowing memory) if we collect all of the deployments in one go.

ledger/store/src/transaction/mod.rs
```diff
@@ -37,6 +37,8 @@ use synthesizer_program::Program;
 use synthesizer_snark::{Certificate, VerifyingKey};
 
 use anyhow::Result;
+#[cfg(not(feature = "serial"))]
+use rayon::prelude::*;
```
howardwu, line 41: nit: move below to line 46

ledger/store/src/transaction/deployment.rs
```diff
@@ -30,6 +30,8 @@ use synthesizer_snark::{Certificate, VerifyingKey};
 
 use anyhow::Result;
 use core::marker::PhantomData;
+#[cfg(not(feature = "serial"))]
+use rayon::prelude::*;
 use std::borrow::Cow;
 
 /// A trait for deployment storage.
```
howardwu:
```suggestion
use std::borrow::Cow;

#[cfg(not(feature = "serial"))]
use rayon::prelude::*;

/// A trait for deployment storage.
```

ledger/store/src/transaction/mod.rs
```diff
@@ -37,6 +37,8 @@ use synthesizer_program::Program;
...
 use serde::{Deserialize, Serialize};
 use std::borrow::Cow;
```
howardwu:
```suggestion
use serde::{Deserialize, Serialize};
use std::borrow::Cow;

#[cfg(not(feature = "serial"))]
use rayon::prelude::*;

```

synthesizer/src/vm/mod.rs
```diff
@@ -48,6 +48,8 @@ use synthesizer_program::{FinalizeGlobalState, FinalizeStoreTrait, Program};
 use aleo_std::prelude::{finish, lap, timer};
 use indexmap::IndexMap;
 use parking_lot::RwLock;
+#[cfg(not(feature = "serial"))]
+use rayon::prelude::*;
 use std::sync::Arc;
```
howardwu:
```suggestion
use std::sync::Arc;

#[cfg(not(feature = "serial"))]
use rayon::prelude::*;

```

synthesizer/src/vm/mod.rs
ljedrz: the reason for the collect is that `rayon` processes those and they couldn't be loaded sequentially otherwise; perhaps we could look into chunking the list of deployment transaction ids in order to limit maximum memory use?

ljedrz:
Currently there are only 3 deployments in the network, but we can revisit this PR once there are more, in order to be able to measure how much memory is used when collecting the average deployment so we can consider some chunking strategy.

synthesizer/src/vm/mod.rs
ljedrz: updated the PR with a possible approach; it will need to be benchmarked again, so I'll mark the PR as a draft for now

ljedrz:
Even now, with a small number of deployments, this PR (in the current, chunked form) speeds up the deployment loading process by ~59%, which I only expect to improve with the increase in deployment count.

ledger/src/lib.rs
```diff
@@ -144,6 +144,7 @@ impl<N: Network, C: ConsensusStorage<N>> Ledger<N, C> {
     pub fn load_unchecked(genesis_block: Block<N>, dev: Option<u16>) -> Result<Self> {
         let timer = timer!("Ledger::load_unchecked");
 
+        debug!("Loading the consensus store...");
```
howardwu:
```suggestion
        debug!("Loading the ledger from storage...");
```

synthesizer/src/vm/mod.rs
```diff
@@ -61,6 +61,8 @@ use synthesizer_program::{FinalizeGlobalState, FinalizeOperation, FinalizeStoreT
 use aleo_std::prelude::{finish, lap, timer};
 use indexmap::{IndexMap, IndexSet};
 use parking_lot::RwLock;
+#[cfg(not(feature = "serial"))]
+use rayon::prelude::*;
 use std::sync::Arc;
```
howardwu:
```suggestion
use parking_lot::RwLock;
use std::sync::Arc;

#[cfg(not(feature = "serial"))]
use rayon::prelude::*;

```

raychu86 (approved):

ledger/src/lib.rs
```diff
@@ -144,6 +144,7 @@ impl<N: Network, C: ConsensusStorage<N>> Ledger<N, C> {
     pub fn load_unchecked(genesis_block: Block<N>, dev: Option<u16>) -> Result<Self> {
         let timer = timer!("Ledger::load_unchecked");
 
+        debug!("Loading the ledger from storage...");
```
howardwu:
```suggestion
        info!("Loading the ledger from storage...");
```

synthesizer/src/vm/mod.rs
```diff
@@ -120,25 +126,46 @@ impl<N: Network, C: ConsensusStorage<N>> VM<N, C> {
...
         let transaction_store = store.transaction_store();
+        // Retrieve the list of deployment transaction IDs.
+        let deployment_ids = transaction_store.deployment_transaction_ids().collect::<Vec<_>>();
         // Load the deployments from the store.
-        for transaction_id in transaction_store.deployment_transaction_ids() {
-            // Load the deployment and its imports.
-            load_deployment_and_imports(&mut process, transaction_store, *transaction_id)?;
+        for (i, chunk) in deployment_ids.chunks(256).enumerate() {
+            debug!(
+                "Loading deployments: {}-{}/{}...",
```
howardwu:
```suggestion
                "Loading deployments {}-{} (of {})...",
```

howardwu (approved):
LGTM

PR #1612, opened 2023-06-01
Peinlcy/fix create transfer transaction

peinlcy:
## Motivation

(Write your motivation here)

## Test Plan

(Write your test plan here)

## Related PRs

(Link your related PRs here)


raychu86:
Although this check is correct, it is insufficient to simply check that the record balance is greater than the priority fee (as there is a minimum required fee for executions). The following call to `vm.execute` will do this check internally.

PR #1616, opened 2023-06-02
feat: logging

joske:
Using `println!` for logging is annoying when snarkVM is used as library (like snarkOS). 
* Replace println and friends with logging
* install simple_logger impl in binaries where appropriate


howardwu:
The `aleo-cli` usage must be `println` currently, as we pass this to the CLI. A larger refactor with the Leo team will be needed to change these functionalities.

joske:
The log calls will use `println!` when used with the (now) provided SimplerLogger (see `utilities/src/simpler_logger.rs`). 

In `leo` repository, there weren't many instances of println!, PR here: https://github.com/AleoHQ/leo/pull/2405

When snarkOS uses snarkVM, all these log calls will use the logger that snarkOS installs (instead of just indiscriminately printing to stdout).

PR #1619, opened 2023-06-05
fix construct transaction path in inclusion

peinlcy:
append transaction leaf to transaction tree before prove its existance

### DIFF
--- a/synthesizer/src/process/stack/inclusion/execute.rs
+++ b/synthesizer/src/process/stack/inclusion/execute.rs
@@ -44,6 +44,9 @@ macro_rules! prepare_execution_impl {
             // Construct the transaction leaf.
             let transaction_leaf = TransactionLeaf::new_execution(transition_index as u16, **transition.id());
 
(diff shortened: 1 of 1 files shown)


howardwu:
Closing as this would introduce a vulnerability. One cannot use their own leaf to prove `local` path. The local path must exist beforehand.

howardwu:
Just noticed the report. Will review and follow up there.

peinlcy:
> Closing as this would introduce a vulnerability. One cannot use their own leaf to prove `local` path. The local path must exist beforehand.

Sorry i still have a question, if the transition with local input task is not appended to the transaction tree, how it can calculate the transaction path, won't it be an out-of-index error ?  (line 53 in inclusion/execute.rs)

peinlcy:
I write a simple test

peinlcy:
I add a unittest to cover the scene of local record, can u review it plz
https://github.com/AleoHQ/snarkVM/pull/1624

PR #1624, opened 2023-06-05
add unittest to test call external contract with local records

peinlcy:
## Motivation

(Write your motivation here)

## Test Plan

(Write your test plan here)

## Related PRs

(Link your related PRs here)


howardwu:
@peinlcy thanks for writing this!

Just a heads up there is a known bug for verifying nested external calls. @d0cd is working through that bug at the moment, and should have it fixed this week.

howardwu:
Closing as this test already exists now.

https://github.com/AleoHQ/snarkVM/blob/testnet3/synthesizer/src/vm/mod.rs#L834-L898

PR #1633, opened 2023-06-07
Reduces the size of verifying keys

howardwu:
## Motivation

Reduces the size of verifying keys, by introducing a `UniversalVerifier` key, containing globally static parameters.
- Verifying keys are now 665 bytes (82.4% reduction)

This PR resamples parameters.

### Changes to Marlin
- Removes `MM: MarlinMode` from `CircuitVerifyingKey`
- Removes `Marlin::verify_batch_prepared`
- Removes `PreparedCircuitVerifyingKey`
- Removes `PreparedCommitment`
- Removes `SNARK::UniversalSetupConfig` trait


Pratyush:
Instead of surgically removing parts of the vk related to the PC scheme, I would keep `SonicKZG10::VerifyingKey` as is, and pass that around. That is, `marlin::VerifyingKey` would no longer contain `SonicKZG10::VerifyingKey` as a subcomponent, but we would instead pass in the latter as another argument to `Marlin::verify`.

howardwu:
Agreed, moving this PR back to a draft.

ljedrz (commented):
Looks great, will come back to take a look after the aforementioned adjustment.

howardwu:
Given the size of this PR already, I will checkpoint my work here. `UniversalVerifier` is a new verifying key that contains all globally static parameters for verifcation.

In a follow-on PR, I'll tackle `UniversalProver`, which will be a new proving key that will contain all globally static parameters for proving (i.e. the `CommitterKey`).

howardwu:
VKs are now reduced by 82.4% :)

ljedrz (approved):
While the entire scope of these changes is beyond my knowledge of the snarkVM codebase, I didn't see anything that would stand out to me; in other words, LGTM :+1:.

vicsn (commented):
I'm late to the party but LGTM!

PR #1640, opened 2023-06-08
Limits the number of write commands, updates finalize root to merklization

howardwu:
## Motivation

Handles todos for finalize root:
- Limits the number of write commands to 10
- Updates finalize root to use merklization


d0cd (approved):
LGTM!

PR #1643, opened 2023-06-08
[Feature] `get.or_use`

d0cd:
This PR, removes `get.or_init` and introduces `get.or_use`. 
If the entry does not exist in the mapping, the default operand is loaded and stored into the destination register.

### DIFF
--- a/synthesizer/src/program/finalize/command/get_or_use.rs
+++ b/synthesizer/src/program/finalize/command/get_or_use.rs
@@ -28,7 +28,7 @@ use console::{
     program::{Identifier, Register, Value},
 };
 
-/// A get command that uses the provided operand in case of failure, e.g. `get.or_use accounts[r0] r1 into r2;`.
+/// A get command that uses the provided default in case of failure, e.g. `get.or_use accounts[r0] r1 into r2;`.
 /// Gets the value stored at `operand` in `mapping` and stores the result in `destination`.
 /// If the key is not present, `default` is stored in `destination`.
 #[derive(Clone, PartialEq, Eq, Hash)]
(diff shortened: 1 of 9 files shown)


synthesizer/src/program/finalize/mod.rs
```diff
@@ -110,11 +110,11 @@ impl<N: Network> Finalize<N> {
         // Ensure the number of write commands has not been exceeded.
         ensure!(
             self.num_writes < N::MAX_WRITES,
-            "Cannot add more than {} 'get.or_init' and 'set' commands",
+            "Cannot add more than {} 'get.or_use' and 'set' commands",
```
howardwu:
```suggestion
            "Cannot add more than {} 'set' commands",
```

synthesizer/src/program/finalize/mod.rs
```diff
@@ -110,11 +110,11 @@ impl<N: Network> Finalize<N> {
...
             N::MAX_WRITES
         );
 
-        // If the command is an instruction, `get` command, or `get.or_init` command, perform additional checks.
+        // If the command is an instruction, `get` command, or `get.or_use` command, perform additional checks.
```
howardwu:
```suggestion
        // Perform additional checks on the command.
```

synthesizer/src/program/finalize/command/get_or_use.rs
```diff
@@ -28,11 +28,11 @@ use console::{
     program::{Identifier, Register, Value},
 };
 
-/// A get command that initializes the mapping in case of failure, e.g. `get.or_init accounts[r0] r1 into r2;`.
+/// A get command that uses the provided operand in case of failure, e.g. `get.or_use accounts[r0] r1 into r2;`.
```
howardwu:
```suggestion
/// A get command that uses the provided default in case of failure, e.g. `get.or_use accounts[r0] r1 into r2;`.
```

synthesizer/src/program/finalize/command/get_or_use.rs
```diff
@@ -81,7 +81,7 @@ impl<N: Network> GetOrInit<N> {
     }
 }
 
-impl<N: Network> GetOrInit<N> {
+impl<N: Network> GetOrUse<N> {
     /// Finalizes the command.
     #[inline]
     pub fn finalize<P: FinalizeStorage<N>>(
```
howardwu, line 86: You can change this function to return `Result<()>`.

In the parent function that calls this function, you can add `.map(|_| None)` to the end of the call.

howardwu (approved):
LGTM

PR #1646, opened 2023-06-10
[Fix] Execute and verify nested function calls.

d0cd:
This PR,
- fixes execution for nested function calls.
- fixes verification for nested function calls.
- adds tests using nested external calls and checks their execution order (the order of transitions in an `Execution`).

**To: Reviewer**
This PR needs feedback on how to appropriately organize the call graph reconstruction logic.

### What is the correct order?
Suppose that we have the following programs:
- `C`, with transition `c`.
- `B`, with transition `b`.
- `A`, which imports `B` and `C`, and contains transition `a` which invokes `B/b` and `C/c`.
Suppose that the call graph looks like:
```
A::a
  --> B::b
  --> C::c
```
Then the finalize blocks need to evaluated in the following order: `
```
B::b, C::c, A::a
```
In other words, **finalize blocks are evaluated in order of the transition that finished first**.

For a more complex example,
Suppose we have the following call graph, where each call is an external call.
```
A::a
--> B::b
       --> C::c
       --> D::d
--> E::e
       --> B::b
              --> C::c
              --> D::d"
       --> D::d
       --> C::c
```
The finalize blocks need to be evaluated in the following order: 
```
C::c, D::d, B::b, C::c, D::d, B::b, D::d, C::c, E::e, A::a
```

### DIFF
--- a/synthesizer/src/process/execute.rs
+++ b/synthesizer/src/process/execute.rs
@@ -126,7 +126,10 @@ impl<N: Network> Process<N> {
         // Iterate over each transition in reverse post-order, and populate the call graph and IO cache.
         for transition in execution.transitions().rev() {
             // Add the transition inputs and outputs to the IO cache.
-            io_cache.insert(*transition.id(), (transition.inputs().to_vec(), transition.outputs().to_vec()));
+            io_cache.insert(
+                *transition.id(),
+                (transition.inputs().to_vec(), transition.output_ids().cloned().collect::<Vec<_>>()),
+            );
 
             // If the stack has entries and they are complete, then pop them.
             while let Some(entry) = traversal_stack.last() {
@@ -136,6 +139,7 @@ impl<N: Network> Process<N> {
                     break;
                 }
             }
+
             // Now process the current transition.
             match traversal_stack.last_mut() {
                 // If the stack is empty, then push the current transition.
@@ -154,7 +158,7 @@ impl<N: Network> Process<N> {
             }
             // Process the entry at the top of the stack.
             // If the entry is complete, then pop it.
-            // TODO: Unwrap safety
+            // Note this unwrap is safe, since we just pushed the entry.
             let top = traversal_stack.last().unwrap();
             if !top.children.is_empty() {
                 update_call_graph(traversal_stack.pop().unwrap(), &mut call_graph, &mut uid_to_tid);
@@ -171,7 +175,7 @@ impl<N: Network> Process<N> {
                         Instruction::Call(call) => Some(call.operator()),
                         _ => None,
                     })
-                    .collect_vec();
+                    .collect::<Vec<_>>();
 
                 // If the function has no children, then pop the entry.
                 if children.is_empty() {
@@ -181,8 +185,8 @@ impl<N: Network> Process<N> {
                         .into_iter()
                         .map(|call_operator| {
                             let (pid, fname) = match call_operator {
-                                crate::CallOperator::Locator(locator) => (locator.program_id(), locator.name()),
-                                crate::CallOperator::Resource(name) => (&top.pid, name),
+                                crate::CallOperator::Locator(locator) => (locator.program_id(), locator.resource()),
+                                crate::CallOperator::Resource(fname) => (&top.pid, fname),
                             };
                             TransitionMetadata::new(&mut counter, *pid, *fname, None)
                         })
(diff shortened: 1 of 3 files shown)


synthesizer/src/process/execute.rs
```diff
@@ -226,4 +220,153 @@ impl<N: Network> Process<N> {
...
+        }
+
+        impl<N: Network> TransitionMetadata<N> {
+            fn new(counter: &mut usize, pid: ProgramID<N>, fname: Identifier<N>, tid: Option<N::TransitionID>) -> Self {
+                let uid = *counter;
+                *counter += 1;
+                Self { uid, pid, fname, tid, children: None }
+            }
+
+            fn is_complete(&self) -> bool {
```
howardwu, line 262:
```suggestion
            /// Returns 'true' if the subgraph starting from this transition has been fully-indexed.
            fn is_complete(&self) -> bool {
```

synthesizer/src/process/execute.rs
```diff
@@ -226,4 +220,153 @@ impl<N: Network> Process<N> {
...
+            Ok(())
+        };
+
+        // Initialize a call graph, which is a map of transition IDs to the transition IDs it calls.
+        let mut call_graph = HashMap::new();
+        // Initialize a mapping from UIDs to transition IDs.
+        let mut uid_to_tid = HashMap::new();
+
+        // Initialize a stack to track transition metadata, while traversing the call graph.
+        let mut traversal_stack: Vec<TransitionMetadata<N>> = Vec::new();
```
howardwu, line 289: nit: rename to `metadata_stack` to improve readability.

synthesizer/src/process/execute.rs
```diff
@@ -226,4 +220,153 @@ impl<N: Network> Process<N> {
...
+    // In other words, we do not know which transitions were invoked by which transitions.
+    // Note that transition names are insufficient to reconstruct the call graph, since the same function can be invoked multiple times, in different ways.
+    //
+    // In order to reconstruct the call graph, we:
+    // - Iterate over the call structure in reverse post-order. The ordering is maintained by the `traversal_stack`.
+    // - Process each transition in the `Execution` in reverse, assigning its transition ID to the corresponding function call.
+    fn construct_call_graph(&self, execution: &Execution<N>) -> Result<HashMap<N::TransitionID, Vec<N::TransitionID>>> {
+        // Metadata for each transition the execution.
+        struct TransitionMetadata<N: Network> {
+            uid: usize,
```
howardwu, line 247: nit: rename to `index` for consistency with rest of repo.

synthesizer/src/process/execute.rs
```diff
@@ -226,4 +220,153 @@ impl<N: Network> Process<N> {
...
+        // Iterate over each transition in reverse post-order, and populate the call graph.
+        for transition in execution.transitions().rev() {
+            // If the stack has complete metadata entries, then remove and add them to the call graph.
+            while let Some(metadata) = traversal_stack.last() {
+                if metadata.is_complete() {
+                    update_call_graph(traversal_stack.pop().unwrap(), &mut call_graph, &mut uid_to_tid)?;
+                } else {
+                    break;
+                }
+            }
```
howardwu: This can be moved to the end of the `for` loop. Then the last `while` loop at the end of this function can be removed (and an `ensure!(traversal_stack.is_empty, "...")` can be added at the end of the function).

synthesizer/src/process/execute.rs
```diff
@@ -226,4 +220,153 @@ impl<N: Network> Process<N> {
...
+                    })
+                    .collect::<Vec<_>>();
+
+                // Add the children UIDs to the metadata.
+                // Note this unwrap is safe, for the same reason as above.
+                let top = traversal_stack.last_mut().unwrap();
+                let child_uids = children.iter().map(|child| child.uid).collect::<Vec<_>>();
+                match top.children {
+                    None => top.children = Some(child_uids),
+                    Some(ref mut children) => children.extend(child_uids),
```
howardwu: To my knowledge, this `Some(..)` case is never hit. We should replace it with a `Some(..) => bail!("...")`.

howardwu (approved):
LGTM, pending minor safety changes

synthesizer/src/process/execute.rs
```diff
@@ -226,4 +220,153 @@ impl<N: Network> Process<N> {
...
+
+        // A helper function to update the call graph, given transition metadata.
+        let update_call_graph = |metadata: TransitionMetadata<N>,
+                                 call_graph: &mut HashMap<N::TransitionID, Vec<N::TransitionID>>,
+                                 uid_to_tid: &mut HashMap<usize, N::TransitionID>|
+         -> Result<()> {
+            // Check that the transition metadata is complete.
+            ensure!(metadata.is_complete(), "The transition metadata is incomplete");
+            // Update the UID to TID mapping.
+            uid_to_tid.insert(metadata.uid, metadata.tid.unwrap());
```
howardwu: I would move this call to after `call_graph.insert()`. To my knowledge, the `call_graph.insert()`'s usage of `uid_to_tid[&uid]` on its `children` should fail if it ever needed this iteration's `uid_to_tid.insert()`.

synthesizer/src/process/execute.rs
```diff
@@ -226,4 +220,153 @@ impl<N: Network> Process<N> {
...
+            // Update the call graph.
+            call_graph.insert(
+                metadata.tid.unwrap(),
+                metadata.children.unwrap().into_iter().map(|uid| uid_to_tid[&uid]).collect::<Vec<_>>(),
```
howardwu:
```suggestion
                                metadata
                    .children // Safe to unwrap, since the metadata is complete.
                    .unwrap()
                    .into_iter()
                    .map(|uid| uid_to_tid.get(&uid).copied().ok_or(anyhow!("Invalid traversal - missing 'tid' for uid '{uid}'")))
                    .collect::<Result<Vec<_>, _>>()?,
```

howardwu (approved):
LGTM

PR #1651, opened 2023-06-11
Refactors the verify methods to their own files

howardwu:
## Motivation

Refactors the verify methods to their own files.

In addition, this PR adds a missing check to ensure Transition only contains finalize inputs if the function expects a finalize scope.


d0cd (approved):
LGTM!

PR #1652, opened 2023-06-12
Adds cost solvers for transactions

howardwu:
## Motivation

This PR introduces the following changes:
- Adds cost solvers for transactions: `deployment_cost()` and `execution_cost()`
- Adds a `namespace_cost` for deployments
    - `namespace_cost := 10^(10 - num_characters_in_program_id)`
    - In English, programs with IDs that are 10 characters or more incur 1 credit.
        - 10 or more characters -> 1 credit
        - 9 characters -> 10 credits
        - 8 characters -> 100 credits
        - 7 characters -> 1,000 credits
        - 6 characters -> 10,000 credits
        - etc.


synthesizer/src/vm/verify.rs
```diff
@@ -215,6 +215,9 @@ impl<N: Network, C: ConsensusStorage<N>> VM<N, C> {
     fn check_fee(&self, fee: &Fee<N>, deployment_or_execution_id: Field<N>) -> Result<()> {
         let timer = timer!("VM::verify_fee");
 
+        // Ensure the fee does not exceed the limit.
+        ensure!(*fee.amount()? < N::MAX_FEE, "Fee verification failed: fee exceeds the maximum limit");
```
d0cd, line 219: Should this be `<= N::MAX_FEE`, to check against exceeding the limit?

ledger/src/helpers/cost.rs
```diff
@@ -0,0 +1,87 @@
...
+    let program_id = deployment.program_id();
+    // Determine the number of characters in the program ID.
+    let num_characters = u32::try_from(program_id.name().to_string().len())?;
+
+    // Compute the storage cost in microcredits.
+    let storage_cost = size_in_bytes
+        .checked_mul(N::DEPLOYMENT_FEE_MULTIPLIER)
+        .ok_or(anyhow!("The storage cost computation overflowed for a deployment"))?;
+
+    // Compute the namespace cost in microcredits: 10^(10 - num_characters).
```
d0cd:
```suggestion
    // Compute the namespace cost in credits: 10^(10 - num_characters).
```

HarukaMa:
Considering the genesis supply is 1.5B and the cost of one letter program is 1B, would it be more sensible to add a cutoff (sth like 10m for 3-letter or shorter) or just set a minimal program name length?

howardwu:
We're considering introducing a minimal program length (at 2 or 3 characters).  I also think it is reasonable to draw a cutoff at 10m. Let's try this in this iteration of Testnet 3 and see how things go. More feedback will help us to guide our thinking.

raychu86 (approved):
LGTM

d0cd (commented):
Left some comments.

ledger/src/helpers/cost.rs
raychu86: I believe the comment is correct, as the `.saturating_mul(1_000_000)` call brings this number to microcredits

synthesizer/src/vm/verify.rs
howardwu: I think `<` is fine. We have a Byzantine problem if someone singlehandedly controls 1B credits.

d0cd (approved):
LGTM!

PR #1653, opened 2023-06-12
check serial_number if record is local

peinlcy:
## Motivation

(Write your motivation here)

## Test Plan

(Write your test plan here)

## Related PRs

(Link your related PRs here)


howardwu:
This logic would break the circuit by definition.

PR #1655, opened 2023-06-12
Fix breaking build

howardwu:
## Motivation

The `finalize-fees` PR merged and then the `hash` PR merged.

But the `hash` PR didn't know about a change from the `finalize_fees` PR.

This fixes it.


d0cd (approved):
LGTM!

PR #1658, opened 2023-06-12
Refactors the transitiion verifier inputs into its own internal method

howardwu:
## Motivation

This PR refactors the transitiion verifier inputs into its own internal method.

One small optimization was made: we no longer retrieve the same verifying key multiple times for an execution.

No new features.


raychu86 (approved):
LGTM

PR #1661, opened 2023-06-12
[Nit] Fix error message and Coinbase benchmarks

raychu86:
## Motivation

This PR is a quick fix to the error message checking the number of the prover solutions in a `CoinbaseSolution`. Also the benchmarks were fixed, as running accumulation on `500` prover solutions would fail.


howardwu (approved):
LGTM

PR #1662, opened 2023-06-12
[Fix] Disable `benches/instruction.rs`

d0cd:
This PR disables `benches/instruction.rs` because it significantly increases build/clippy times.


howardwu:
Does the bench file need to be commented out?

d0cd:
Yes, otherwise clippy still processes the code.

howardwu:
Can you add a comment to the file then?

howardwu:
Never mind. I really need this in, merging.

PR #1664, opened 2023-06-12
Adds additional checks to `check_next_block`

raychu86:
## Motivation

This PR adds additional `has_duplicate` checks to the block level in `check_next_block`. 

Additionally, redundant `self.contains_*` checks were removed from `check_next_block`, as they are already done further down in the `check_transaction_basic` calls.


howardwu (approved):
LGTM

PR #1665, opened 2023-06-13
[Feature] Cast group element into its coordinates.

d0cd:
This PR adds support for casting a group element into its coordinates.
For example, one can write:
- `cast r0 into r1 as group.x`
- `cast r0 into r1 as group.y`
- `cast r0 into r1 as field`, where `r0` is a `group`.


synthesizer/src/program/instruction/operation/cast.rs
```diff
@@ -46,15 +46,94 @@ use console::{
...
+    /// Prints the cast type as a string.
+    fn fmt(&self, f: &mut Formatter) -> fmt::Result {
+        Display::fmt(self, f)
+    }
+}
+
+impl<N: Network> FromStr for CastType<N> {
+    type Err = Error;
+
+    /// Returns a register type from a string literal.
```
howardwu, line 88:
```suggestion
    /// Returns a cast type from a string literal.
```

howardwu (approved):
LGTM

PR #1673, opened 2023-06-14
[Feature] `cast` and `cast.lossy` for literals.

d0cd:
This PR supports casting between different literals.
Users may use:
- `cast`, which checks that information are preserved between casts.
- `cast.lossy` which performs casting with lossy truncation

**Example Usage**
```aleo
input r0 as field.private;
cast r0 into r1 as group;
cast r0 into r2 as u8; // r0 must be [0field, 255field] for this to pass
cast.lossy r0 into r3 as u8; // The bottom 8 bits of the r0 are extracted into a u8 and placed into r3
```


d0cd:
Closing in favor of #1968

PR #1674, opened 2023-06-14
[Test] Update test framework to run `execute` and `finalize`.

d0cd:
This PR,
- updates the test configs to use YAML objects. This allows for more flexible definitions of test files.
- introduces `test_vm_execute_and_finalize`, which executes and finalizes transactions on the VM, storing the output.
- adds the `randomness` configuration option, allowing for reproducibility.
- adds the `private_key` configuration option, allowing test cases to be run using a specific key.


howardwu (approved):
LGTM

PR #1677, opened 2023-06-14
Proposal: reduce the size of the BlockCoinbasePuzzleCommitmentMap

ljedrz:
One of the "heaviest" data sets in the ledger is the `BlockCoinbasePuzzleCommitmentMap`, which holds the largest (by an order of magnitude) number of records of all the maps. Due to that, and due to its values not being used in any hot site (at least not in snarkVM), I propose that the type of its value be changed from the block hash to the block height, reducing the size of every record by 28B (reducing the disk size of the map by ~30%). This would still allow us to trace the block to which the puzzle commitment belongs, and to obtain the hash with an additional query (if needed).

This is a breaking change.

### DIFF
--- a/ledger/src/find.rs
+++ b/ledger/src/find.rs
@@ -25,7 +25,7 @@ impl<N: Network, C: ConsensusStorage<N>> Ledger<N, C> {
         self.vm.block_store().find_block_hash(transaction_id)
     }
 
-    /// Returns the block hash that contains the given `puzzle commitment`.
+    /// Returns the block height that contains the given `puzzle commitment`.
     pub fn find_block_height_from_puzzle_commitment(
         &self,
         puzzle_commitment: &PuzzleCommitment<N>,
--- a/ledger/src/find.rs
+++ b/ledger/src/find.rs
@@ -26,11 +26,11 @@ impl<N: Network, C: ConsensusStorage<N>> Ledger<N, C> {
     }
 
     /// Returns the block hash that contains the given `puzzle commitment`.
-    pub fn find_block_hash_from_puzzle_commitment(
+    pub fn find_block_height_from_puzzle_commitment(
         &self,
         puzzle_commitment: &PuzzleCommitment<N>,
-    ) -> Result<Option<N::BlockHash>> {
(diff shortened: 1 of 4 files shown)


synthesizer/src/store/block/mod.rs
```diff
@@ -844,11 +841,11 @@ impl<N: Network, B: BlockStorage<N>> BlockStore<N, B> {
     }
 
     /// Returns the block hash that contains the given `puzzle commitment`.
```
howardwu:
```suggestion
    /// Returns the block height that contains the given `puzzle commitment`.
```

ledger/src/find.rs
```diff
@@ -26,11 +26,11 @@ impl<N: Network, C: ConsensusStorage<N>> Ledger<N, C> {
     }
 
     /// Returns the block hash that contains the given `puzzle commitment`.
```
howardwu:
```suggestion
    /// Returns the block height that contains the given `puzzle commitment`.
```

synthesizer/src/store/block/mod.rs
```diff
@@ -142,7 +142,7 @@ pub trait BlockStorage<N: Network>: 'static + Clone + Send + Sync {
     /// The mapping of `block hash` to `block coinbase solution`.
     type CoinbaseSolutionMap: for<'a> Map<'a, N::BlockHash, Option<CoinbaseSolution<N>>>;
     /// The mapping of `puzzle commitment` to `block hash`.
```
howardwu:
```suggestion
    /// The mapping of `puzzle commitment` to `block height`.
```

synthesizer/src/store/block/mod.rs
```diff
@@ -437,12 +437,9 @@ pub trait BlockStorage<N: Network>: 'static + Clone + Send + Sync {
     }
 
     /// Returns the block hash that contains the given `puzzle commitment`.
```
howardwu:
```suggestion
    /// Returns the block height that contains the given `puzzle commitment`.
```

howardwu (approved):
LGTM. This is a great find. It also opens the question "do we need this map at all?" We can discuss this over the next month. For now, let's land this before network reset.

PR #1679, opened 2023-06-14
Adds a hotfix for `credits.aleo` in `store`

howardwu:
## Motivation

Adds a hotfix for `credits.aleo` in `store`


raychu86 (approved):
LGTM

PR #1680, opened 2023-06-14
[Test] Invoke `transfer_public` with insufficient fees.

d0cd:
This PR introduces a test where `transfer_public` is invoked with insufficient finalize fees and then with sufficient finalize fees.


raychu86:
Changes to how we set minimum fees proposed here - https://github.com/AleoHQ/snarkVM/pull/1687 will invalidate this test (since this test uses the priority_fee_in_microcredits as the finalize fee payment)

PR #1681, opened 2023-06-14
[Feature] Introduce `position` and `branch.<eq, neq>` commands.

d0cd:
This PR introduces:
- the `position` command, e.g `position exit`, which indicates a point to branch execution to.
- the `branch.eq` command, e.g `branch.eq r0 r1 to exit`, which branches execution to the position indicated by `exit` if `r0` and `r1` are equal.
- the `branch.neq` command, e.g `branch.neq r0 r1 to exit`, which branches execution to the position indicated by `exit` if `r0` and `r1` are **not** equal.

**Example Usage**
```
program test_branch.aleo;

function run_test:
    input r0 as u8.public;
    finalize r0;

finalize run_test:
    input r0 as u8.public;
    branch.eq r0 0u8 to exit;
    assert.eq true false;
    position exit;
```
The `finalize` block exits successfully if the input is `0u8` and fails otherwise.


howardwu (approved):
LGTM

PR #1682, opened 2023-06-15
Load program imports before the program itself

raychu86:
## Motivation

This PR addresses this [issue](https://github.com/AleoHQ/snarkOS/issues/2306) that prevented the VM from loading properly when loading programs from storage. The issue what that since the iterator is lexicographical, the node may try to load a program before its import. 

The solution here is to recursively load program imports prior to loading the program itself.

## Test Plan

A test has been added to check that the VM imports the program's imports before processing the program itself. This is done by creating three programs (each importing the previous) and checking that loading a new VM from storage works correctly.

### DIFF
--- a/synthesizer/src/vm/mod.rs
+++ b/synthesizer/src/vm/mod.rs
@@ -32,6 +32,7 @@ use crate::{
     process::{Authorization, FinalizeGlobalState, Process, Query, Trace},
     program::Program,
     store::{BlockStore, ConsensusStorage, ConsensusStore, FinalizeStore, TransactionStore, TransitionStore},
+    TransactionStorage,
 };
 use console::{
     account::{Address, PrivateKey},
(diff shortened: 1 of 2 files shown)


ghostant-1017:
Maybe a non-recursive topological sort would be better and safer than a recursive depth-first search in engineering

howardwu:
> Maybe a non-recursive topological sort would be better and safer than a recursive depth-first search in engineering

Hm, the topological sort would require performing 2 passes, and keep a dependency tree in memory. Assuming we store program IDs, for 1M programs, each with 1 dependency on average, that would be an 128MB allocated.

This design seems to guarantee we reach every deployment, without increasing memory requirements.

Also worth noting: We're planning to introduce a namespace layer to abstract program IDs from programs as part of the next network reset. Introducing this would deprecate the current recursive approach completely.

synthesizer/src/vm/mod.rs
```diff
@@ -69,16 +70,53 @@ impl<N: Network, C: ConsensusStorage<N>> VM<N, C> {
...
+            // Retrieve the deployment from the transaction id.
+            let deployment = match transaction_store.get_deployment(&transaction_id)? {
+                Some(deployment) => deployment,
+                None => bail!("Deployment transaction '{transaction_id}' is not found in storage."),
+            };
+
+            // Fetch the program from the deployment.
+            let program = deployment.program();
+            let program_id = program.id();
+
```
howardwu, line 88: You can add a check here:
```rs
if process.contains_program(program_id) {
    return Ok(());
}
```

howardwu (approved):
LGTM

PR #1686, opened 2023-06-15
Use a fixed rng seed for `test_load_deployments_with_imports`

raychu86:
## Motivation

This PR uses a fixed rng seed for `test_load_deployments_with_imports` in order to ensure that the `deployment_transaction_ids()` lexicographical ordering is different for the sequential deployment transactions. This allows us to test the recursive loading of import programs before parent programs.


howardwu (approved):
LGTM

PR #1687, opened 2023-06-15
Include execution/deployment minimum cost during generation

raychu86:
## Motivation

This PR migrates the minimum fee calculation from the `ledger` into `synthesizer`, so the minimum fee will be required upon generation of the deployment or execution transaction. Previously, we had to rely on the `additional_fee_in_microcredits` to pay for the `finalize` fees, which doesn't make sense.

## Test Plan

The `test_insufficient_finalize_fees` test has been updated to show that transactions can't be created if the minimum fee requirement isn't hit.


howardwu (approved):
LGTM

d0cd (commented):
LGTM!

PR #1689, opened 2023-06-16
Updates the inclusion path to use the local transition path

howardwu:
## Motivation

This PR updates the inclusion path to use the local transition path.

### DIFF
--- a/synthesizer/src/process/trace/inclusion/mod.rs
+++ b/synthesizer/src/process/trace/inclusion/mod.rs
@@ -65,7 +65,7 @@ impl<N: Network> Inclusion<N> {
         }
 
         // Retrieve the transition index.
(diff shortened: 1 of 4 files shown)


synthesizer/src/process/trace/inclusion/mod.rs
```diff
@@ -65,20 +64,22 @@ impl<N: Network> Inclusion<N> {
             bail!("Inclusion expected the same number of input IDs as transition inputs")
         }
 
+        // Retrieve the transition index.
+        let transition_index = self.input_tasks.len() as u8;
+
```
raychu86, line 69: Nit: This can be moved down into line 92.

raychu86 (approved):
LGTM

PR #1690, opened 2023-06-16
Switch this test to be fully-deterministic

howardwu:
## Motivation

Switch this test to be fully-deterministic.


raychu86 (approved):
lgtm

PR #1693, opened 2023-06-16
Refactors `snarkvm-ledger` testing, removing `once_cell` usage

howardwu:
## Motivation

This PR refactors `snarkvm-ledger` testing, removing `once_cell` usage. This means that each test will be responsible with sampling its own initial state, and may lead to performance overheads, but introduces better reproducibility via deterministic testing.


raychu86 (approved):
LGTM assuming CI passes

PR #1696, opened 2023-06-17
Abstracts program, function, finalize, and closures

howardwu:
## Motivation

This PR abstracts program, function, finalize, and closure into a generic form, in preparation for extracting `program` into its own subcrate of `synthesizer`.

As part of this, the implementation of `Finalize::add_command` needed to be redesigned, and trait methods were used to pass through the necessary property checks performed when adding the `command` to `finalize`.


d0cd (approved):
LGTM!

PR #1700, opened 2023-06-17
Splits synthesizer CI into lib and integration tests

howardwu:
## Motivation

This PR splits the `snarkvm-synthesizer` crate's CI into a `lib` and an `integration` portion for testing. This is due to recently increasingly CI runtimes, due to an increase in the number of tests in the crate.


d0cd (approved):
LGTM!

PR #1704, opened 2023-06-20
Moves target computation to commitment

howardwu:
## Motivation

This refactor PR:
- Moves target computation to commitment, so that it is accessible from a lower level.
- Adds a G1 affine constructor for the commitment


raychu86 (approved):
LGTM

PR #1719, opened 2023-06-25
Fix program fetching in cost function

HarukaMa:
## Motivation

In `snarkos developer` context, we don't really have a transaction storage; programs are added to `Process` directly. This will cause executing to fail as the `cost` function can't find the program in the non-existent storage. This PR makes the `cost` function read the program directly from the `Process` instead of the consensus storage to make executing work again.

## Test Plan

Well, there really should be a test case for this but I didn't write one...

## Related PRs

<sub>(What's that `protocol-docs` repo? Seems it's private.)</sub>


howardwu:
Thank you for doing this @HarukaMa ! I was planning to do it myself as well.

howardwu (approved):
LGTM

PR #1726, opened 2023-06-27
feat: use median timestamp

joske:
## Motivation
Use the median timestamp of the consensus output to have a deterministic timestamp

## Related PRs

https://github.com/AleoHQ/snarkOS/pull/2475

### DIFF
--- a/ledger/src/advance.rs
+++ b/ledger/src/advance.rs
@@ -22,6 +22,7 @@ impl<N: Network, C: ConsensusStorage<N>> Ledger<N, C> {
         candidate_transactions: Vec<Transaction<N>>,
         candidate_solutions: Option<Vec<ProverSolution<N>>>,
         rng: &mut R,
+        next_timestamp: i64,
     ) -> Result<Block<N>> {
         // Retrieve the latest state root.
         let latest_state_root = *self.latest_state_root();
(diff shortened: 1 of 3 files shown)


ledger/src/advance.rs
```diff
@@ -22,6 +22,7 @@ impl<N: Network, C: ConsensusStorage<N>> Ledger<N, C> {
         candidate_transactions: Vec<Transaction<N>>,
         candidate_solutions: Option<Vec<ProverSolution<N>>>,
         rng: &mut R,
+        median_timestamp: i64,
```
vvp: Perhaps rename this as `block_timestamp`, `next_timestamp` or just `timestamp`?  Talking about median timestamp here leaks unnecessary details of the current implementation.

ledger/src/tests.rs
```diff
@@ -128,7 +129,15 @@ finalize foo:
...
 
     // Construct the next block.
-    let block = ledger.prepare_advance_to_next_block(&private_key, vec![transaction], None, rng).unwrap();
+    let block = ledger
+        .prepare_advance_to_next_block(
+            &private_key,
+            vec![transaction],
+            None,
+            rng,
+            OffsetDateTime::now_utc().unix_timestamp(),
```
vvp: Add assertion that the `block.timestamp()`  equals to the given timestamp.

vvp (changes requested):
vn :+1:  left few comments

niklaslong (approved):

niklaslong:
lgtm (pending Vesku's suggestions ofc)

vvp (approved):

ljedrz (approved):

ledger/src/advance.rs
```diff
@@ -22,6 +22,7 @@ impl<N: Network, C: ConsensusStorage<N>> Ledger<N, C> {
         candidate_transactions: Vec<Transaction<N>>,
         candidate_solutions: Option<Vec<ProverSolution<N>>>,
         rng: &mut R,
+        next_timestamp: i64,
```
howardwu, line 25: We should be passing in `candidate_certificates` to calculate the median timestamp here.

Let's get those data structures in place, then revisit this.

PR #1727, opened 2023-06-28
[Feature] Introduce `ArrayType`.

d0cd:
This PR introduces the `ArrayType`.
This allows users to define single-dimensional static arrays, e.g `[u8; 14]`.
The maximum size of an array is artificially set to be `u32::MAX`.
Note that this PR does **not** integrate `ArrayType` into the rest of the codebase, just introduces it as an isolated module


howardwu (approved):
LGTM

PR #1731, opened 2023-06-30
Verify transition output limit using correct constant

vvp:
This PR changes the transition output limit verification to compare against correct constant. 

As both `MAX_INPUTS` and `MAX_OUTPUTS` are currently the same, this doesn't functionally change anything.


howardwu (approved):
LGTM, great catch!

Issue #1735, opened 2023-06-30
[Bug] Transactions that fail to execute the finalize function are not rejected


elderhammer:
## 🐛 Bug Report

Transactions that fail to execute the finalize function are not rejected.

## Steps to Reproduce

1. Create a transaction Tx whose finalize result is wrong (intentional).
2. Broadcast the transaction Tx.
3. In the logic of the node executing the produce_next_block() -> propose_next_block() -> prepare_advance_to_next_block() -> speculate() -> atomic_speculate() function, because the finalize execution of the transaction Tx fails, a ConfirmedTransaction will be created, we call it CTx, whose type is RejectedExecute, contains a Fee transaction. CTx will be returned as a transaction in the new block.
4. The node executes check_next_block() to check the new block, and executes the atomic_speculate() function again. Because there is a Fee transaction in the transactions of the new block, according to the logic of atomic_speculate(), the program reports an error "Cannot speculate on a fee transaction".

## Expected Behavior

This case should be rejected and the fee consumed, otherwise a malicious transaction can be sent to the network.

## Your Environment

- snarkOS: 796de94816d6b8aa238c51468f0398fe0170bc14
- snarkVM: 0.12.6
- rustc: 1.69.0
- OS: Ubuntu 18.04.5 LTS

howardwu:
Thank you @elderhammer . Our team also caught this yesterday afternoon while debugging a similar bug. We'll get a fix implemented quickly.

elderhammer:
Glad to hear this. :-)

raychu86:
Closed with https://github.com/AleoHQ/snarkVM/pull/1736

PR #1736, opened 2023-06-30
Introduce formal `Rejected` wrapper

raychu86:
## Motivation

This PR introduces a formalized enum `Rejected` for rejected deployments and executions.

This change addresses #1735, by modifying `check_next_block` to properly check `vm.speculate`. 

We now properly reproduce the original unconfirmed transaction from the confirmed transactions, so we don't run into the  `Cannot speculate on a fee transaction` errors in `check_next_block` like before.

_Note: This change is safe to include in `testnet3`, because no rejected executions or deployments have been included on-chain yet._
## Test Plan

Tests have been added to check that the new `check_next_block` logic correctly checks that transactions are rejected properly.


howardwu (approved):
LGTM

PR #1737, opened 2023-06-30
[Fix] Fixes `verify_execution` on functions that call closures.

d0cd:
This PR, fixes `verify_execution` on functions that call closures.
Prior to this PR, the call graph, used to verify transitions, would include calls to closures. 
This resulted in call graphs that were inconsistent with the transitions in an `Execution`.
The errors produced were of the form:
```
Invalid traversal - traversal stack is not empty',
```
This PR fixes the issue by excluding calls to closures from the call graph.


howardwu (approved):
LGTM

PR #1738, opened 2023-07-01
Improve some doc about multiplication gadgets.

acoglio:
Clarify the carry bits returned by `mul_with_carry`, and that we actually use Babbage and not Karatsuba.

Clarify that in `mul_wrapped` we use Babbage, which in this case consists of three multiplications, same as Karatsuba.


bendyarm (approved):
Looks good to me.  Even though the triple-slash region is long, the clarification
of the meaning of the returned carry bits could prevent a misunderstanding,
so I think it is good to have.

CynthiaCurtis:
ok

raychu86 (approved):

PR #1756, opened 2023-07-06
[Fix] Remove `Send` requirement from async_trait

iamalwaysuncomfortable:
## Motivation

The implementation of `async_trait` for the Query trait introduces a send bound on the futures which is incompatible with wasm-bindgen. This PR removes the explicit send requirement on the async trait allowing wasm to properly compile.

### DIFF
--- a/ledger/query/src/query.rs
+++ b/ledger/query/src/query.rs
@@ -59,7 +59,7 @@ impl<N: Network, B: BlockStorage<N>> From<&str> for Query<N, B> {
     }
 }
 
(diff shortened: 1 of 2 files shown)


ljedrz (approved):
LGTM :+1:

iamalwaysuncomfortable:
@ljedrz @howardwu @raychu86 Is it okay to merge this? Ultimately it'll likely require a new SnarkVM release to unbreak `wasm` target. Is a new release something we can do?

raychu86 (approved):
LGTM

PR #1762, opened 2023-07-06
Fully atomic rocksDB writes

ljedrz:
While we're already using rocksDB's atomic write batches, they are currently only atomic within a single `DataMap`. In order to have fully atomic writes, we need to adjust the current logic to keep track of the queued operations across all the maps and perform a write only once per atomic write batch.

Unlike with the typed `DataMap`s, this forces us to keep track of the untyped/serialized operations within the shared storage object, and to update it in tandem with operations on all the maps. This is made more complicated by the checkpointing logic, but it can be achieved by keeping track of the atomic scope depth (i.e. the checkpoint index) and using a "stack of stacks" instead of a single stack (like in the `DataMap`). This way we are able to keep track of the checkpoints across all the `DataMap`s, and gain improved tracking of nested storage operations.

In order to better reflect the complexity of our typical storage operations, 2 new tests containing nested `DataMap`s are also added.

As a drive-by, the `checkpoint` field in `DataMap` is renamed to `checkpoints` in order to better reflect that it contains a number of them as opposed to a single one at a time. Also, some of the occurrences of replacing a vector with a new empty one are replaced with a call to `Vec::clear` in order to preserve the earlier allocation(s).

### DIFF
--- a/ledger/store/src/helpers/rocksdb/internal/map.rs
+++ b/ledger/store/src/helpers/rocksdb/internal/map.rs
@@ -94,6 +94,8 @@ impl<
 
         // Ensure that the atomic batch is empty.
         assert!(self.atomic_batch.lock().is_empty());
+        // Ensure that the database atomic batch is empty.
+        assert!(self.database.atomic_batch.lock().is_empty());
     }
 
     ///
@@ -192,6 +194,10 @@ impl<
 
         // Subtract the atomic depth index.
         let previous_atomic_depth = self.database.atomic_depth.fetch_sub(1, Ordering::SeqCst);
+
+        // Ensure that the value of `atomic_depth` doesn't overflow, meaning that all the
+        // calls to `start_atomic` have corresponding calls to `finish_atomic`.
+        assert!(previous_atomic_depth != 0);
 
         // If we're at depth 0, it is the final call to `finish_atomic` and the
         // atomic write batch can be physically executed.
@@ -360,12 +366,12 @@ mod tests {
         helpers::rocksdb::{internal::tests::temp_dir, MapID, TestMap},
         FinalizeMode,
     };
-    use anyhow::anyhow;
     use console::{
         account::{Address, FromStr},
         network::Testnet3,
     };
 
+    use anyhow::anyhow;
     use serial_test::serial;
     use tracing_test::traced_test;
 
--- a/ledger/store/src/helpers/rocksdb/internal/map.rs
+++ b/ledger/store/src/helpers/rocksdb/internal/map.rs
@@ -43,19 +43,17 @@ impl<
     /// Inserts the given key-value pair into the map.
     ///
     fn insert(&self, key: K, value: V) -> Result<()> {
-        // Prepare the prefixed key and serialized value.
-        let raw_key = self.create_prefixed_key(&key)?;
-        let raw_value = bincode::serialize(&value)?;
-
         // Determine if an atomic batch is in progress.
         match self.is_atomic_in_progress() {
             // If a batch is in progress, add the key-value pair to the batch.
             true => {
                 self.atomic_batch.lock().push((key, Some(value)));
-                self.database.atomic_batch.lock().push((raw_key.into(), Some(raw_value.into())));
             }
             // Otherwise, insert the key-value pair directly into the map.
             false => {
+                // Prepare the prefixed key and serialized value.
+                let raw_key = self.create_prefixed_key(&key)?;
+                let raw_value = bincode::serialize(&value)?;
                 self.database.put(raw_key, raw_value)?;
             }
         }
@@ -67,18 +65,16 @@ impl<
     /// Removes the key-value pair for the given key from the map.
     ///
     fn remove(&self, key: &K) -> Result<()> {
-        // Prepare the prefixed key.
-        let raw_key = self.create_prefixed_key(key)?;
-
         // Determine if an atomic batch is in progress.
         match self.is_atomic_in_progress() {
             // If a batch is in progress, add the key to the batch.
             true => {
                 self.atomic_batch.lock().push((*key, None));
-                self.database.atomic_batch.lock().push((raw_key.into(), None));
             }
             // Otherwise, remove the key-value pair directly from the map.
             false => {
+                // Prepare the prefixed key.
+                let raw_key = self.create_prefixed_key(key)?;
                 self.database.delete(raw_key)?;
             }
         }
@@ -93,10 +89,11 @@ impl<
     fn start_atomic(&self) {
         // Set the atomic batch flag to `true`.
         self.batch_in_progress.store(true, Ordering::SeqCst);
+        // Increment the atomic depth index.
+        self.database.atomic_depth.fetch_add(1, Ordering::SeqCst);
 
         // Ensure that the atomic batch is empty.
         assert!(self.atomic_batch.lock().is_empty());
-        assert!(self.database.atomic_batch.lock().is_empty());
     }
 
     ///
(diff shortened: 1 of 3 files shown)


ledger/store/src/helpers/rocksdb/internal/map.rs
```diff
@@ -173,26 +175,33 @@ impl<
...
+        self.checkpoints.lock().clear();
         // Set the atomic batch flag to `false`.
         self.batch_in_progress.store(false, Ordering::SeqCst);
 
+        // Subtract the atomic depth index.
+        let previous_atomic_depth = self.database.atomic_depth.fetch_sub(1, Ordering::SeqCst);
+
+        // If we're at depth 0, it is the final call to `finish_atomic` and the
+        // atomic write batch can be physically executed.
+        if previous_atomic_depth == 1 {
```
raychu86, line 208: Do we need to check the edge case if `self.database.atomic_depth` is already zero? Since `fetch_sub` does a wrap if we subtract from 0

ledger/store/src/helpers/rocksdb/internal/map.rs
```diff
@@ -351,13 +360,178 @@ mod tests {
...
     };
-    use console::{account::Address, network::Testnet3, prelude::*};
+    use anyhow::anyhow;
+    use console::{
+        account::{Address, FromStr},
+        network::Testnet3,
+    };
 
     use serial_test::serial;
     use tracing_test::traced_test;
```
raychu86: Nit: Move the anyhow import lower.

```suggestion
    use console::{
        account::{Address, FromStr},
        network::Testnet3,
    };

    use anyhow::anyhow;
    use serial_test::serial;
    use tracing_test::traced_test;
```

ledger/store/src/helpers/rocksdb/internal/map.rs
```diff
@@ -109,55 +111,55 @@ impl<
...
+        self.atomic_batch.lock().clear();
         // Clear the checkpoint stack.
-        *self.checkpoint.lock() = Default::default();
+        self.checkpoints.lock().clear();
         // Set the atomic batch flag to `false`.
         self.batch_in_progress.store(false, Ordering::SeqCst);
+        // Clear the database-wide atomic batch.
+        self.database.atomic_batch.lock().clear();
+        // Reset the atomic batch depth.
+        self.database.atomic_depth.store(0, Ordering::SeqCst);
```
raychu86, line 155: I believe setting the `atomic_depth` to 0 here is correct, but is there ever a case where we should be decrementing by 1, and then ensuring on the last `abort_atomic` call, that we hit 0? Meaning we've properly aborted all maps?

ljedrz:
Pushed a new commit with a simpler setup, as suggested by @raychu86; instead of being a "parallel" database-wide plumbing, it heavily leverages the underlying `DataMap`s, and only requires the tracking of the current atomic batch scope depth in order to trigger with the very last call to `finish_atomic`. This solution is also faster and more lightweight.

raychu86 (changes requested):

ledger/store/src/helpers/rocksdb/internal/map.rs
ljedrz: I can add a relevant assertion; all the calls to `start_atomic` should have corresponding calls to `finish_atomic`, so the final value should always be 0.

ledger/store/src/helpers/rocksdb/internal/map.rs
```diff
@@ -109,55 +111,55 @@ impl<
...
+        self.atomic_batch.lock().truncate(checkpoint);
     }
 
     ///
     /// Aborts the current atomic operation.
     ///
     fn abort_atomic(&self) {
         // Clear the atomic batch.
-        *self.atomic_batch.lock() = Default::default();
+        self.atomic_batch.lock().clear();
```
raychu86, line 147: Just curious: Is there a benefit to using `clear` vs reassigning to `Default::default()`?

ljedrz: yes; this reuses the previous allocation, which is very likely of an appropriate size

ledger/store/src/helpers/rocksdb/internal/map.rs
ljedrz: `abort_atomic` used to "bubble up" in the past, but this has recently changed; now, instead of being called in case of errors, it is only called in `atomic_finalize`, after all the other operations; I _think_ we should be fine decrementing it as well, but I think it would require a multi-map test for `atomic_finalize`

ledger/store/src/helpers/rocksdb/internal/map.rs
```diff
@@ -90,6 +89,9 @@ impl<
     fn start_atomic(&self) {
         // Set the atomic batch flag to `true`.
         self.batch_in_progress.store(true, Ordering::SeqCst);
+        // Increment the atomic depth index.
+        self.database.atomic_depth.fetch_add(1, Ordering::SeqCst);
+
```
raychu86, line 94: We may want to check that the underlying `atomic_batch` in `self.database` is also empty.

```suggestion
// Ensure that the database atomic batch is empty.
assert!(self.database.atomic_batch.lock().is_empty());
```

ledger/store/src/helpers/rocksdb/internal/map.rs
```diff
@@ -109,55 +113,55 @@ impl<
...
-        // Acquire the write lock on the atomic batch.
-        let mut atomic_batch = self.atomic_batch.lock();
-
         // Retrieve the last checkpoint.
-        let checkpoint = self.checkpoint.lock().pop().unwrap_or(0);
-
+        let checkpoint = self.checkpoints.lock().pop().unwrap_or(0);
         // Remove all operations after the checkpoint.
-        atomic_batch.truncate(checkpoint);
+        self.atomic_batch.lock().truncate(checkpoint);
```
howardwu: TODO: Revert this change.

ledger/store/src/helpers/rocksdb/internal/map.rs
```diff
@@ -173,26 +177,37 @@ impl<
...
+        // calls to `start_atomic` have corresponding calls to `finish_atomic`.
+        assert!(previous_atomic_depth != 0);
+
+        // If we're at depth 0, it is the final call to `finish_atomic` and the
+        // atomic write batch can be physically executed.
+        if previous_atomic_depth == 1 {
+            // Empty the collection of pending operations.
+            let batch = mem::take(&mut *self.database.atomic_batch.lock());
+            // Execute all the operations atomically.
+            self.database.rocksdb.write(batch)?;
```
howardwu, line 212:
```suggestion
            self.database.rocksdb.write(batch)?;
            // Ensure that the database atomic batch is empty.
            assert!(self.database.atomic_batch.lock().is_empty());
```

ledger/store/src/helpers/rocksdb/internal/map.rs
```diff
@@ -1071,21 +1257,27 @@ mod tests {
...
+
                     // Update the key.
                     map.insert(0, "2".to_string()).unwrap();
 
                     Ok(())
                 });
                 assert!(result.is_ok());
 
+                // Make sure the checkpoint index is 1.
+                assert_eq!(map.checkpoints.lock().last(), Some(&1));
```
howardwu, line 1281:
```suggestion
                assert_eq!(map.checkpoints.lock().last(), Some(&1));
                // Ensure that the atomic batch is empty.
                assert!(self.atomic_batch.lock().is_empty());
                // Ensure that the database atomic batch size is 1.
                assert_eq!(self.database.atomic_batch.lock().len(), 1);
                // Ensure that the database atomic depth is 1.
                assert_eq!(self.database.atomic_depth.load(Ordering::SeqCst), 1);
```

howardwu (approved):
LGTM

raychu86 (approved):
LGTM

Issue #1743, opened 2023-07-04
[Bug] Nested `Struct` in `Record` have an extra closing bracket `}`


pa-long:
## 🐛 Bug Report

Whenever a program generates `Record` with that contains a `Struct`, an extra closing bracket will be added to the end of the `Struct`. Same for nested `Struct` recursively.

## Steps to Reproduce

All data to reproduce can be found in this transaction, executing a program transition :
https://vm.aleo.org/api/testnet3/transaction/at1d08qvu95lzlv92dsjrv8n4lvel8x9rann2t54py4h80vjy0nxcyqhxrpuq

`transition[0].outputs[0]` decoded with viewkey `AViewKey1hKNmqi83PpwjCTSSUCZT2rX3p3jcktNXzJZnk6rSqq4b` returns :
```
{
    owner: aleo1wamjqlka7d0gazlxdys6n8e8zeee3ymedwvw8elvh7529kwd45rq0plgax.private,
    id: {
        collection_number: 25371715469272213720224343178070736482u128.private
    },
    data: {
        privately_minted: {
        positive: 0u64.private,
        negative: 0u64.private
    }
    },
    royalty_fees: 500u16.private,
    publicizable: true.private,
    updatable: true.private,
    metadata_uri: {
        part0: 140152554740597502496525853903429913932u128.private,
        part1: 134617583084312214720973570367896241221u128.private,
        part2: 104133346941662617879306115050903128141u128.private,
        part3: 153398181356134783820482129834790617088u128.private
    }
    },
    _nonce: 4657518007057346011967333542044959903015797864849889452602278140358226817257group.public
}
```

## Expected Behavior

Return value should be:
```
{
    owner: aleo1wamjqlka7d0gazlxdys6n8e8zeee3ymedwvw8elvh7529kwd45rq0plgax.private,
    id: {
        collection_number: 25371715469272213720224343178070736482u128.private
    },
    data: {
    privately_minted: {
        positive: 0u64.private,
        negative: 0u64.private
    },
    royalty_fees: 500u16.private,
    publicizable: true.private,
    updatable: true.private,
    metadata_uri: {
        part0: 140152554740597502496525853903429913932u128.private,
        part1: 134617583084312214720973570367896241221u128.private,
        part2: 104133346941662617879306115050903128141u128.private,
        part3: 153398181356134783820482129834790617088u128.private
    },
    _nonce: 4657518007057346011967333542044959903015797864849889452602278140358226817257group.public
}
```

[...]

PR #1767, opened 2023-07-07
[Test] Check implementation of `Display` on `Entry`.

d0cd:
This PR adds tests checking that the implementation of `Display` on `Entry` is correct.
Note that it includes the test case from #1743.


raychu86 (approved):
LGTM

PR #1769, opened 2023-07-07
[Fix] Remove serial flag from wasm feature within synthesizer-process

iamalwaysuncomfortable:
## Motivation

One of the feature flags specified in the wasm flag `snarkvm-synthesizer-process` specified a `serial` flag. Unforutnately this creates a dependency cascade which ensures that the `serial` feature of snarkvm-algorithms is activated and thus Multi-Threaded `wasm` isn't possible. This would cause a significant regression in performance on `aleo.tools` and any other web property using Aleo wasm for web execution.

This PR removes the serial flag therefore re-enabling multi-threading in the wasm compilation target.


iamalwaysuncomfortable:
@howardwu @ljedrz @raychu86 - hate to ask again, but can this get a rubber stamp and subsequent release? It's another blocker.

synthesizer/process/Cargo.toml
```diff
@@ -38,7 +38,6 @@ serial = [
 wasm = [
   "console/wasm",
   "ledger-block/wasm",
-  "ledger-query/serial",
```
howardwu, line 41: That should read `ledger-query/wasm`

iamalwaysuncomfortable: @howardwu change made

iamalwaysuncomfortable:
Updated the dependency and verified it enables multi-threading, possible to merge this?

howardwu (approved):
LGTM

PR #1777, opened 2023-07-11
[Feature] Introduce support for `Plaintext::Array`

d0cd:
This PR introduces `Plaintext::Array`, allowing users to define plaintext arrays.

PR #1778, opened 2023-07-11
[Feature] Support `Access::Index` variant.

d0cd:
This PR supports the `Access::Index` variant, allowing users to index into `Plaintext::List` with constant indices.

Depends on #1777.


howardwu:
Because of the staleness, I had to manually port over the relevant code changes. They are now in #1777. Closing this PR in favor of a combined PR in #1777

PR #1779, opened 2023-07-12
Defer loading `credits.aleo` proving keys and inclusion proving keys until time of use.

howardwu:
## Motivation

This PR defers loading `credits.aleo` proving keys and inclusion proving keys until time of use.


raychu86 (approved):
LGTM, though this may impact the first run of the transaction benchmarks

iamalwaysuncomfortable (approved):
This can probably unify the load() and load_web() functions

PR #1780, opened 2023-07-12
[Feature] Enables the use of `ArrayType` in `synthesizer`.

d0cd:
This PR adds `ArrayType` to `PlaintextType`, allowing it to be used in `synthesizer` and consequently in programs.
This PR also extends the `cast` operation to arrays.

Depends on #1778.


AlexZhao6666:
is mean i can use ArrayType in my leo program? if not when i can use?

PR #1784, opened 2023-07-12
Parallelize `Deployment` verification

raychu86:
## Motivation

This PR speeds up `Deployment` verification (in `Stack::verify_deployment`) by synthesizing the function circuits and verifying their corresponding certificates in in parallel. 

This optimization is much more noticeable when verifying deployments of programs with many large functions, as each function needs to go through the same disjoint verification process.


howardwu (approved):
LGTM

PR #1795, opened 2023-07-15
Ensure Mul of LinearCombination doesnt create zero entries in the R1CS

vicsn:
## Motivation

This PR ensures only non-zero valued entries end up in our sparse R1CS matrices. 

Goals I kept in mind:
- semantic correctness: sparse R1CS matrices should not contain non-zero entries
- proper abstraction: functionality such as filtering out zero entries should happen at the same level of abstraction
- performance during normal workloads

Background: recall that these matrices are of the form `[(variable_index, value), ...]`. So if we have a single variable `1` and a single constraint `1*1=1`, we have matrices: `[(1,1)]*[(1,1)]=[(1,1)]`. Any value of `0` doesn't make sense.

The zero entries are caused by the way we track of variables and constraints. E.g. when multiplying a `Field<E>` of zero by another `Field<E>` with actual terms, we keep those zeroed terms around. This led to 1542 entries in the R1CS matrix of the form: `Public(0, 1) * 0` when running `credits.sh`. 

We already have logic for filtering out zero entries in `impl {From, Add} for LinearCombination<F>`, this PR adds it for `Mul`.

## Test Plan

The performance impact on running a large script (credits.sh) is insignificant on my machine.

### DIFF
--- a/algorithms/src/snark/marlin/ahp/errors.rs
+++ b/algorithms/src/snark/marlin/ahp/errors.rs
@@ -15,6 +15,8 @@
 /// Describes the failure modes of the AHP scheme.
 #[derive(Debug)]
 pub enum AHPError {
+    /// Anyhow error
+    Anyhow(anyhow::Error),
     /// The batch size is zero.
     BatchSizeIsZero,
     /// An error occurred during constraint generation.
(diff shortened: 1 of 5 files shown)


vicsn:
Closing for now, number of tests are still failing.

PR #1796, opened 2023-07-15
Ensure Mul of LinearCombination doesnt create zero entries in the R1CS 

vicsn:
## Motivation

This PR ensures only non-zero valued entries end up in our sparse R1CS matrices. 

Goals I kept in mind:
- semantic correctness: sparse R1CS matrices should not contain non-zero entries
- encapsulation and consistency: particular functionality should happen at one level of abstraction
- performance during normal workloads

Background: recall that these matrices are of the form `[(variable_index, value), ...]`. So if we have a single variable `1` and a single constraint `1*1=1`, we have matrices: `[(1,1)]*[(1,1)]=[(1,1)]`. An entry of `(1,0)` doesn't make sense.

The zero entries are caused by the way we track of variables and constraints. E.g. when multiplying a `Field<E>` of zero by another `Field<E>` with actual terms, we keep those zeroed terms around. This led to 1542 entries in the R1CS matrix of the form: `Public(0, 1) * 0` when running `credits.sh`. 

We already have logic for filtering out zero entries in `impl {From, Add} for LinearCombination<F>`, this PR adds it for `Mul`. 

## Open Question
The required adjustments to `fn output_mode` and `fn count` reflect that more operations output a `Constant`. I'm not sure this is sound, if none of the assigned reviewers know I'm happy to dive deeper.

## Test Plan

The performance impact of this PR on running a large script (credits.sh) is insignificant on my machine.

This PR requires resampling, because it changes the credits circuit (removing the, surprise, zero entries). The ledger and synthesizer tests succeeds for me locally.


vicsn:
@Pratyush would also be great to have a sanity check from you on this

circuit/environment/src/helpers/linear_combination.rs
```diff
@@ -424,6 +424,8 @@ impl<F: PrimeField> Mul<&F> for LinearCombination<F> {
         let mut output = self;
         output.constant *= coefficient;
         output.terms.iter_mut().for_each(|(_, current_coefficient)| *current_coefficient *= coefficient);
+        // If the coefficient of terms are now zero, remove the entries.
```
Pratyush: I think you can combine this with the previous line via a filter_map

vicsn: 194323a29

circuit/types/integers/src/mul_checked.rs
```diff
@@ -211,42 +211,52 @@ impl<E: Environment, I: IntegerType> Integer<E, I> {
...
 
     fn count(case: &Self::Case) -> Count {
         // Case 1 - 2 integers fit in 1 field element (u8, u16, u32, u64, i8, i16, i32, i64).
         if 2 * I::BITS < (E::BaseField::size_in_bits() - 1) as u64 {
             match I::is_signed() {
                 // Signed case
-                true => match (case.0, case.1) {
-                    (Mode::Constant, Mode::Constant) => Count::is(I::BITS, 0, 0, 0),
-                    (Mode::Constant, _) | (_, Mode::Constant) => {
+                true => match (case.0, case.1, case.2, case.3) {
```
howardwu, line 221: I think you can revert these code changes (and ignore `case.2, case.3`), right?

circuit/types/integers/src/mul_checked.rs
```diff
@@ -211,42 +211,52 @@ impl<E: Environment, I: IntegerType> Integer<E, I> {
...
         }
         // Case 2 - 1.5 integers fit in 1 field element (u128, i128).
         else if (I::BITS + I::BITS / 2) < (E::BaseField::size_in_bits() - 1) as u64 {
             match I::is_signed() {
                 // Signed case
-                true => match (case.0, case.1) {
-                    (Mode::Constant, Mode::Constant) => Count::is(I::BITS, 0, 0, 0),
-                    (Mode::Constant, _) | (_, Mode::Constant) => Count::is(4 * I::BITS, 0, 965, 1164),
-                    (_, _) => Count::is(3 * I::BITS, 0, 1227, 1427),
+                true => match (case.0, case.1, case.2, case.3) {
```
howardwu: I think you can revert these code changes (and ignore `case.2, case.3`), right?

circuit/types/integers/src/mul_checked.rs
```diff
@@ -211,42 +211,52 @@ impl<E: Environment, I: IntegerType> Integer<E, I> {
 }
 
 impl<E: Environment, I: IntegerType> Metrics<dyn MulChecked<Integer<E, I>, Output = Integer<E, I>>> for Integer<E, I> {
-    type Case = (Mode, Mode);
+    type Case = (Mode, Mode, bool, bool);
```
howardwu, line 214: Is the `bool, bool` necessary?

circuit/types/integers/src/mul_checked.rs
```diff
@@ -258,12 +268,13 @@ impl<E: Environment, I: IntegerType> Metrics<dyn MulChecked<Integer<E, I>, Outpu
 impl<E: Environment, I: IntegerType> OutputMode<dyn MulChecked<Integer<E, I>, Output = Integer<E, I>>>
     for Integer<E, I>
 {
-    type Case = (Mode, Mode);
+    type Case = (Mode, Mode, bool, bool);
```
howardwu, line 267: Is the `bool, bool` necessary?

circuit/types/integers/src/mul_wrapped.rs
```diff
@@ -58,28 +58,32 @@ impl<E: Environment, I: IntegerType> MulWrapped<Self> for Integer<E, I> {
 }
 
 impl<E: Environment, I: IntegerType> Metrics<dyn MulWrapped<Integer<E, I>, Output = Integer<E, I>>> for Integer<E, I> {
-    type Case = (Mode, Mode);
+    type Case = (Mode, Mode, bool, bool);
```
howardwu, line 61: ditto

circuit/types/integers/src/mul_wrapped.rs
```diff
@@ -58,28 +58,32 @@ impl<E: Environment, I: IntegerType> MulWrapped<Self> for Integer<E, I> {
...
+            (_, _, _, _) => Count::is(0, 0, I::BITS + (I::BITS / 2) + 4, I::BITS + (I::BITS / 2) + 5),
         }
     }
 }
 
 impl<E: Environment, I: IntegerType> OutputMode<dyn MulWrapped<Integer<E, I>, Output = Integer<E, I>>>
     for Integer<E, I>
 {
-    type Case = (Mode, Mode);
+    type Case = (Mode, Mode, bool, bool);
```
howardwu, line 80: ditto

circuit/types/integers/src/pow_checked.rs
```diff
@@ -106,7 +106,7 @@ impl<E: Environment, I: IntegerType, M: Magnitude> PowChecked<Integer<E, M>> for
 impl<E: Environment, I: IntegerType, M: Magnitude> Metrics<dyn PowChecked<Integer<E, M>, Output = Integer<E, I>>>
     for Integer<E, I>
 {
-    type Case = (Mode, Mode);
+    type Case = (Mode, Mode, bool, bool);
```
howardwu, line 109: ditto

circuit/types/integers/src/pow_wrapped.rs
```diff
@@ -38,7 +38,7 @@ impl<E: Environment, I: IntegerType, M: Magnitude> PowWrapped<Integer<E, M>> for
 impl<E: Environment, I: IntegerType, M: Magnitude> Metrics<dyn PowWrapped<Integer<E, M>, Output = Integer<E, I>>>
     for Integer<E, I>
 {
-    type Case = (Mode, Mode);
+    type Case = (Mode, Mode, bool, bool);
```
howardwu, line 41: ditto

circuit/types/integers/src/shl_checked.rs
```diff
@@ -116,7 +116,7 @@ impl<E: Environment, I: IntegerType> OutputMode<dyn Shl<Integer<E, I>, Output =
 impl<E: Environment, I: IntegerType, M: Magnitude> Metrics<dyn ShlChecked<Integer<E, M>, Output = Integer<E, I>>>
     for Integer<E, I>
 {
-    type Case = (Mode, Mode);
+    type Case = (Mode, Mode, bool, bool);
```
howardwu, line 119: ditto

circuit/types/integers/src/mul_checked.rs
vicsn: If we revert this file to what it was before this PR, these tests will fail. Same for your other comments. 

This is because of the way we aggregate `LinearCombination`s. When a `LinearCombination` is multiplied by 0, it will now only have a `constant: 0`, instead of e.g. `terms: [(0*Private(some_variable_index))]`.


For example, before this PR, here we had only one case:
```
(Mode::Constant, _) | (_, Mode::Constant) => {
                        Count::is(4 * I::BITS, 0, (7 * I::BITS) + 4, (8 * I::BITS) + 9)
                    }
```

After this PR, if either value in the test case is zero, we get less constraints and more constants:
```
(Mode::Constant, _, true, _) | (_, Mode::Constant, _, true) => {
                        Count::is(7 * I::BITS, 0, (2 * I::BITS) + 1, (2 * I::BITS) + 2)
                    }
 (Mode::Constant, _, false, _) | (_, Mode::Constant, _, false) => {
                        Count::is(4 * I::BITS, 0, (7 * I::BITS) + 4, (8 * I::BITS) + 9)
                    }
```

As mentioned in Slack, I realized some tests are still unaddressed, I will chat with Pranav and look into it again this week. I'm also going to document the way we aggregate the linear combinations for the auditors.

circuit/types/integers/src/pow_wrapped.rs
vicsn: Even though the specific test case is not adjusted in this file, we are passing a `Case` through the same `count!(...)` which requires a `(Mode, Mode, bool, bool)` type. I will make an `IsZero` type alias to make it clearer

circuit/types/integers/src/mul_checked.rs
vicsn: Removed this one ce85fac9c
the others don't seem as trivially removable.

vicsn:
Fixed the remaining tests on this PR. Given that some of the tests depend even not just on the Mode or IsZero, but another aspect, decided to use the more lenient `less_than` comparison method as used elsewhere already.

circuit/types/integers/src/mul_checked.rs
vicsn: FYI the test has `bool, bool)` again, I must have reintroduced that because now the test case counts áre different depending on whether the values are zero or not.

PR #1808, opened 2023-07-19
Test deployments of `credits.aleo`

raychu86:
## Motivation

This PR tests that external deployments of the `credits.aleo` program or programs with the same name will fail.


howardwu (approved):
LGTM

PR #1809, opened 2023-07-19
Optimize

ghostant-1017:


raychu86 (approved):
lgtm

ledger/block/src/transactions/confirmed/mod.rs
```diff
@@ -142,6 +142,15 @@ impl<N: Network> ConfirmedTransaction<N> {
...
     }
 
+    pub fn variant(&self) -> String {
+        match self {
+            Self::AcceptedDeploy(..) => "accepted deploy".to_string(),
+            Self::AcceptedExecute(..) => "accepted execute".to_string(),
+            Self::RejectedDeploy(..) => "rejected deploy".to_string(),
+            Self::RejectedExecute(..) => "rejected execute".to_string(),
+        }
+    }
```
howardwu: We can save 1 allocation by changing the return type to `&str` and avoid calling `.to_string()`.

Also, I realized there are many functions here (including this one) that can be changed from `pub fn` to `pub const fn`.

howardwu (approved):
LGTM

ledger/block/src/transactions/confirmed/mod.rs
```diff
@@ -142,8 +142,17 @@ impl<N: Network> ConfirmedTransaction<N> {
         }
     }
 
+    pub const fn variant(&self) -> &str {
```
howardwu, line 146:
```suggestion
    /// Returns the human-readable variant of the confirmed transaction. 
    pub const fn variant(&self) -> &str {
```

PR #1811, opened 2023-07-19
Introduce ToBits::to_bits_le_into and use it in Vec/[T]

ljedrz:
This PR targets the `feat/narwhal` branch, but it's quite likely it would result in a performance improvement in other use cases as well; it introduces a `ToBits::to_bits_le_into` method that allows us to avoid a lot of allocations when calling `Vec::to_bits` on large collections.

In a 1-minute run of `test_state_coherence` in `snarkOS/narwhal`, this PR reduces the number of allocations from ~4.2M to ~2M, and temporary allocations from ~2.2M to ~0.7M, as measured with `heaptrack`.

I've investigated the option of calling `to_bits_le` directly in `BatchHeader::compute_batch_id`, but that would require plenty of additional implementations of `ToBits`, and it wouldn't have the potential to reduce other allocations in snarkVM, so I decided against that.

I'm filing this as a draft until all the tests have run - it's possible that I might need to introduce another impl of `ToBits::to_bits_le_into`, which should be trivial.

before:
image
image
image

after:
image
image
image

### DIFF
--- a/fields/src/macros.rs
+++ b/fields/src/macros.rs
@@ -108,14 +108,12 @@ macro_rules! sqrt_impl {
                 l_s.resize(l_s.len() + k_1 as usize, l_minus_one);
                 l_s.resize(l_s.len() + k_2 as usize, l);
 
-                let mut x_s: Vec<$Self> = Vec::with_capacity(k as usize);
                 let mut l_sum = 0;
-                l_s.iter().take((k as usize) - 1).for_each(|l| {
+                let x_s = l_s.iter().take((k as usize) - 1).map(|l| {
                     l_sum += l;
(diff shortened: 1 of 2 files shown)


ljedrz:
The CI failures appear to be unrelated, so it's ready for review.

howardwu:
I am retargeting this PR to `testnet3` since the changes are applicable to all snarkVM constructs that use this.

ljedrz:
Rebased against `testnet3` and added one additional commit which provides a small further improvement; that one was also found via profiling.

ljedrz:
Superseded by https://github.com/AleoHQ/snarkVM/pull/1836; the drive-by commit will be filed as its own PR.

PR #1812, opened 2023-07-19
Updates ledger to use memory pool objects

howardwu:
## Motivation

This PR is to prepare snarkOS to begin usage of new memory pool objects for consensus.


raychu86 (approved):
LGTM

PR #1813, opened 2023-07-19
Update `Plaintext::from_bits_*` and introduce `decrypt_unchecked`

raychu86:
## Motivation

This PR adds the `next_bits` helper method to the `Plaintext::from_bits_le` and `Plaintext::from_bits_be` functions to prevent unwanted panics from direct indexing via the old counter method.  

In addition, the `decrypt` function in `Record<N, Ciphertext<N>>` now checks that the view key matches the decrypted record's owner. A `decrypt_unchecked` function was added to perform the old method of decrypting without the ownership check.

## Test Plan

The existing `test_encrypt_and_decrypt` test now also checks that calling `decrypt` from an incorrect view key will cause failures and calling `decrypt_unchecked` will either fail (due to serialization) or return a gibberish record.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -2805,6 +2805,7 @@ dependencies = [
  "bs58",
  "criterion",
  "ed25519-dalek",
+ "once_cell",
  "serde_json",
  "snarkvm-console-network",
  "snarkvm-console-types",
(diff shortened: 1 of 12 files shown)


ljedrz (approved):
Added 2 commits removing allocations and preserving safety. LGTM

PR #1820, opened 2023-07-21
A minimal prepare advance to next block using transmissions

howardwu:
## Motivation

A minimal prepare advance to next block using transmissions


raychu86 (approved):
LGTM

PR #1822, opened 2023-07-21
Switches signature to authority in block

howardwu:
## Motivation

This PR switches `signature` to `authority` in `Block`.

In addition, the `SignatureMap` has been removed, and an `AuthorityMap` has been introduced.

This PR resamples the genesis block.


raychu86 (approved):
LGTM, though you'll likely need to update the genesis block size test

PR #1825, opened 2023-07-21
Adds support for the quorum authority type

howardwu:
## Motivation

This PR adds support for the quorum authority type


raychu86 (approved):
LGTM

PR #1833, opened 2023-07-22
Add profiler for `CoinbasePuzzle::verify`

raychu86:
## Motivation

This PR adds a profiler for the `CoinbasePuzzle::verify` function. It profiles the verification of a CoinbasePuzzle with 10, 100, and `MAX_PROVER_SOLUTIONS` solutions.


iamalwaysuncomfortable (approved):
Would it make sense to somehow add an event publisher to the Timer (like an optional `info!` tracing call that can be picked up by an optionally configured logger)? It'd help with diagnostic information from live nodes.

howardwu (approved):
LGTM

PR #1836, opened 2023-07-24
Introduce ToBitsInto

ljedrz:
This PR introduces a `ToBitsInto` trait which does basically the same thing as `ToBits`, but doesn't require an allocation for every call - instead, it expects a mutable reference to a vector, so that it can extend an existing one instead of having to create a new one every time. This is especially beneficial when called on vectors of `ToBits` objects, but in some cases it improves the implementations for singular objects as well.

The `ToBits` trait is changed so that it extends `ToBitsInto` and provides a default implementation which works just like the current version. All the existing implementations of `ToBits` are changed to `ToBitsInto` ones, which is the bulk of the diff.

Supersedes https://github.com/AleoHQ/snarkVM/pull/1811 (specifically its first commit).


circuit/program/src/data/record/to_bits.rs
```diff
@@ -14,82 +14,74 @@
...
-            .collect::<Vec<_>>();
+        let mut data_bits_be = vec![];
+        for (identifier, entry) in &self.data {
+            identifier.to_bits_be_into(&mut data_bits_be);
+            entry.to_bits_be_into(&mut data_bits_be);
+        }
 
         // Construct the record bits.
-        let mut bits_be = self.owner.to_bits_be();
-        bits_be.extend(U32::constant(console::U32::new(data_bits_be.len() as u32)).to_bits_le());
```
ljedrz, line 50: note: the current implementation contains a call to `to_bits_le`, even though this is in `to_bits_be`

circuit/program/src/data/record/to_bits.rs
```diff
@@ -14,82 +14,74 @@
...
```
ljedrz, line 90: note: the current implementation contains a call to `to_bits_le`, even though this is in `to_bits_be`

ljedrz:
`heaptrack` (the same kind of run as in https://github.com/AleoHQ/snarkVM/pull/1811) confirms that these changes lead to an even greater reduction in the number of allocations, especially small ones:

image

This also confirms that these changes reduce the number of allocations in snarkVM across the board.

ljedrz:
Rebased and added a commit removing some allocations around the uses of `to_bytes_le`.

ljedrz:
Rebased.

SplittyDev (approved):
LGTM

ljedrz:
When running this simple snippet:
```
let genesis = Block::<Testnet3>::from_bytes_le(Testnet3::genesis_bytes()).unwrap();
genesis.to_bytes_le().unwrap().to_bits_le();
```
these changes also reduce the number of allocations by 68% and temporary allocations by 85%.

ljedrz:
Rebased.

acoglio:
@ljedrz At first sight, it looks like this does not change the generated R1CS constraints. Could you confirm that? (I'm asking to see whether we may need to revise the formal verification of the involved gadgets; it's not a problem to do that, but I just wanted to know whether those are affected.)

ljedrz:
No format should be affected - the intent is to just optimize the way we reach it.

The only exception is the 2 cases I commented on, where we currently use `to_bits_le` in a `to_bits_be` implementation (possibly a copy-paste error).

ljedrz:
Superseded by https://github.com/AleoHQ/snarkVM/pull/1877.

PR #1837, opened 2023-07-24
Remove an allocation in sqrt_impl

ljedrz:
Formerly a drive-by part of https://github.com/AleoHQ/snarkVM/pull/1811.

It's a simple removal of an avoidable allocation of an iterator in the `sqrt_impl` macro. It was detected via heap profiling and results in a detectable decrease in the number of allocations.


howardwu (approved):
LGTM

PR #1838, opened 2023-07-24
Update CR structure

raychu86:
## Motivation

This PR updates the formula for rewards to be based on the ratio/portion of the target being fulfilled in the block.


howardwu (approved):
LGTM

PR #1840, opened 2023-07-25
Adds safer logic to guard prepare

howardwu:
## Motivation

Adds safer logic to guard prepare

## Testing

Adds test cases to ensure rewards are empty on unsatisfiable cases.


raychu86 (approved):
LGTM

PR #1841, opened 2023-07-25
Minor refactor

howardwu:
## Motivation

This PR reorders to move logic that is now timestamp independent up in the advance method.


raychu86 (approved):
LGTM

PR #1844, opened 2023-07-26
Updates target methods to remove usage of timestamps and reference heights 

howardwu:
## Motivation

This PR updates target methods to remove usage of timestamps and reference heights.


raychu86 (approved):
LGTM

PR #1794, opened 2023-07-15
Add tests for `abs`, `abs.w`, `add`, and `add.w`.

acoglio:

PR #1849, opened 2023-07-29
[tests] Some tests for `add` and `add.w`.

acoglio:
This may need to be slightly modified in order to pass all the tests, since the added tests are only meant for process-execute.


howardwu:
Can you check the `synthesizer-integration` CI failure? I think it might be related.

d0cd:
Closing in favor of #1794

PR #1854, opened 2023-07-31
Improve API of committee storage, adds current round to get leader API

howardwu:
## Motivation

Improve API of committee storage, adds current round to get leader API


raychu86 (approved):
LGTM

PR #1856, opened 2023-07-31
Removes `current_committee` from the Ledger struct

howardwu:
## Motivation

Removes `current_committee` from the Ledger struct.

This PR prepares for the genesis committee feature.


raychu86 (approved):
LGTM

PR #1857, opened 2023-08-01
Adds committee logic, implements pre-ratify and post-ratify logic in finalize scope

howardwu:
## Motivation

This PR introduces the concept of pre-ratify and post-ratify logic, which adds support for the genesis committee and dynamic committee updates during consensus. As part of this, staking rewards and proving rewards are now being tallied into finalize storage.

As part of this PR, we now enforce that stakers who bond to validators with more than 25% of the total stake will *NOT* receive staking rewards. In addition, to reduce the accumulation of dust, stakers with less than 10 credit bonded will not receive staking rewards.

In addition, the minimum stake TODO has been handled, and the minimum stake for a validator has now been raised to 1M credits (as previously ratified).

## Test Plan

- Unit tests are added to check that `Ratify::Genesis` (de)serializes correctly.
- Unit tests are added to ensure committee helpers are performant and correct.
- Unit tests are introduced to ensure committee storage inserts and removes correctly.
- Unit tests are added to ensure staking rewards and proving rewards are computed correctly.
- Prop tests in committee are updated to reflect the new minimum stake and round rules.
- Integration tests for synthesizer have been updated to match the new genesis block.


raychu86 (approved):
lgtm

PR #1861, opened 2023-08-01
Optimize the ToBits impl for integers and their vectors

ljedrz:
This is a minimal version of the changes in https://github.com/AleoHQ/snarkVM/pull/1836 which doesn't require a new trait; instead, `ToBits` is extended with 2 new methods, `write_bits_le` and `write_bits_be`, which provide a default unoptimized implementation, while integers and their vectors (the most likely targets of `ToBits`) obtain optimized implementations. This aligns to `ToBits` with the `ToBytes` trait, which has a `write_le` method providing similar optimization opportunities.

Just these changes result in the following improvements in a 1 minute Narwhal run with 4 nodes:
- a ~55% reduction in allocations
- a ~76% reduction in temporary allocations
- a great reduction in the number of smallest (0-8B) allocations

These optimizations can later be extended to the rest of the codebase, providing an even greater reduction in the number of allocations and memory fragmentation.

### DIFF
--- a/utilities/src/bits.rs
+++ b/utilities/src/bits.rs
@@ -20,8 +20,24 @@ pub trait ToBits: Sized {
     /// Returns `self` as a boolean array in little-endian order.
     fn to_bits_le(&self) -> Vec<bool>;
 
+    /// Writes `self` into the given vector as a boolean array
+    /// in little-endian order.
+    fn write_bits_le(&self, vec: &mut Vec<bool>) {
+        // A default, unoptimized implementation.
+        let bits = self.to_bits_le();
+        vec.extend_from_slice(&bits);
+    }
+
     /// Returns `self` as a boolean array in big-endian order.
     fn to_bits_be(&self) -> Vec<bool>;
+
+    /// Writes `self` into the given vector as a boolean array
+    /// in big-endian order.
+    fn write_bits_be(&self, vec: &mut Vec<bool>) {
+        // A default, unoptimized implementation.
+        let bits = self.to_bits_be();
+        vec.extend_from_slice(&bits);
+    }
 }
 
 pub trait FromBits: Sized {
@@ -126,10 +142,26 @@ macro_rules! impl_bits_for_integer {
                 bits_le
             }
 
+            #[inline]
+            fn write_bits_le(&self, vec: &mut Vec<bool>) {
+                vec.reserve(<$int>::BITS as usize);
+                let mut value = *self;
+                for _ in 0..<$int>::BITS {
+                    vec.push(value & 1 == 1);
+                    value = value.wrapping_shr(1u32);
+                }
+            }
+
             /// Returns `self` as a boolean array in big-endian order.
             #[inline]
             fn to_bits_be(&self) -> Vec<bool> {
                 self.to_bits_le().into_iter().rev().collect()
+            }
+
+            #[inline]
+            fn write_bits_be(&self, vec: &mut Vec<bool>) {
+                let reversed = self.reverse_bits();
+                reversed.write_bits_le(vec);
             }
         }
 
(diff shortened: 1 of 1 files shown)


ljedrz:
The `ledger-committee` CI failure seems to be unrelated.

utilities/src/bits.rs
```diff
@@ -126,11 +142,27 @@ macro_rules! impl_bits_for_integer {
...
+                let mut value = *self;
+                for _ in 0..<$int>::BITS {
+                    vec.push(value & 1 == 1);
+                    value = value.wrapping_shr(1u32);
+                }
+            }
+
             /// Returns `self` as a boolean array in big-endian order.
             #[inline]
             fn to_bits_be(&self) -> Vec<bool> {
```
howardwu, line 157: Is there a way (without perf penalty) to have the `to_bits_le/be` use the `write_bits_le/be` impl?

ljedrz: yes; it's basically what https://github.com/AleoHQ/snarkVM/pull/1836 did (just with an additional trait); with a single trait the trick would be to have a default implementation of `ToBits::to_bits_le/be` (instead of `::write_bits_le/be` like in the current shape of this PR) implemented in terms of `::write_bits_le/be` and then to reimplement every existing instance of `ToBits::to_bits_le/be` so that they also use `::write_bits_le/be` internally; that way, every top-level/outermost call to `ToBits::to_bits_le` would explicitly allocate only once, as every inner call would extend an initially created vector.

ljedrz: note on the aforementioned perf penalty: this PR on its own also doesn't introduce any, because there are currently no other users of `ToBits::write_bits_le/be` and the ones that do use it reimplement it in a performant fashion.

ljedrz: this is how the trait would look like:
```
diff --git a/utilities/src/bits.rs b/utilities/src/bits.rs
index 64478054d..10fc29b00 100644
--- a/utilities/src/bits.rs
+++ b/utilities/src/bits.rs
@@ -18,26 +18,26 @@ use anyhow::Result;
 
 pub trait ToBits: Sized {
     /// Returns `self` as a boolean array in little-endian order.
-    fn to_bits_le(&self) -> Vec<bool>;
+    fn to_bits_le(&self) -> Vec<bool> {
+        let mut bits = vec![];
+        self.write_bits_le(&mut bits);
+        bits
+    }
 
     /// Writes `self` into the given vector as a boolean array
     /// in little-endian order.
-    fn write_bits_le(&self, vec: &mut Vec<bool>) {
-        // A default, unoptimized implementation.
-        let bits = self.to_bits_le();
-        vec.extend_from_slice(&bits);
-    }
+    fn write_bits_le(&self, vec: &mut Vec<bool>);
 
     /// Returns `self` as a boolean array in big-endian order.
-    fn to_bits_be(&self) -> Vec<bool>;
+    fn to_bits_be(&self) -> Vec<bool> {
+        let mut bits = vec![];
+        self.write_bits_be(&mut bits);
+        bits
+    }
 
     /// Writes `self` into the given vector as a boolean array
     /// in big-endian order.
-    fn write_bits_be(&self, vec: &mut Vec<bool>) {
-        // A default, unoptimized implementation.
-        let bits = self.to_bits_be();
-        vec.extend_from_slice(&bits);
-    }
+    fn write_bits_be(&self, vec: &mut Vec<bool>);
 }
 
 pub trait FromBits: Sized {
```
and this is an example of how one of its implementations (the changes to others would be very similar) would change:
```
diff --git a/utilities/src/biginteger/bigint_256.rs b/utilities/src/biginteger/bigint_256.rs
index 188f68353..6f9754dfa 100644
--- a/utilities/src/biginteger/bigint_256.rs
+++ b/utilities/src/biginteger/bigint_256.rs
@@ -227,13 +227,13 @@ impl crate::biginteger::BigInteger for BigInteger256 {
 
 impl ToBits for BigInteger256 {
     #[doc = " Returns `self` as a boolean array in little-endian order, with trailing zeros."]
-    fn to_bits_le(&self) -> Vec<bool> {
-        BitIteratorLE::new(self).collect::<Vec<_>>()
+    fn write_bits_le(&self, vec: &mut Vec<bool>) {
+        vec.extend(BitIteratorLE::new(self));
     }
 
     #[doc = " Returns `self` as a boolean array in big-endian order, with leading zeros."]
-    fn to_bits_be(&self) -> Vec<bool> {
-        BitIteratorBE::new(self).collect::<Vec<_>>()
+    fn write_bits_be(&self, vec: &mut Vec<bool>) {
+        vec.extend(BitIteratorBE::new(self));
     }
 }

```

ljedrz:
Superseded by https://github.com/AleoHQ/snarkVM/pull/1877.

PR #1869, opened 2023-08-07
Implement DoubleEndedIterator for IterWrap

ljedrz:
It seems that `ExactSizeIterator` is also something that can be implemented here if need be.


ljedrz:
The CI job failures seem unrelated.

howardwu (approved):
LGTM

PR #1873, opened 2023-08-08
Implement tests for staking operations

raychu86:
## Motivation

This PR adds tests for the new staking functions - `bond_public`, `unbond_public`, `unbond_delegator_as_validator`, `claim_unbond_public`, and `set_validator_state`. 

The `is_open` flag was also changed to default to `true` instead of false.


howardwu (approved):
LGTM

PR #1877, opened 2023-08-09
Optimize the implementations of ToBits across the codebase

ljedrz:
This PR supersedes both https://github.com/AleoHQ/snarkVM/pull/1836 and https://github.com/AleoHQ/snarkVM/pull/1861, but does basically the same thing - it optimizes the way we use `ToBits` across the codebase, providing us with a large improvement to the number and size of heap allocations, as confirmed with `heaptrack` in different scenarios.

Further improvements are still possible (more fine-tuned pre-allocations etc.), but the diff is large enough already, so it's best that they are filed as follow-ups later on.


circuit/program/src/data/record/to_bits.rs
```diff
@@ -18,78 +18,70 @@ impl<A: Aleo> ToBits for Record<A, Plaintext<A>> {
...
+        }
 
         // Construct the record bits.
-        let mut bits_be = self.owner.to_bits_be();
-        bits_be.extend(U32::constant(console::U32::new(data_bits_be.len() as u32)).to_bits_le());
-        bits_be.extend(data_bits_be);
-        bits_be.extend(self.nonce.to_bits_be());
-        bits_be
+        self.owner.write_bits_be(vec);
+        U32::constant(console::U32::new(data_bits_be.len() as u32)).write_bits_be(vec);
```
ljedrz, line 47: same note as [here](https://github.com/AleoHQ/snarkVM/pull/1836#discussion_r1272476644) - this is a change, as it seems like a copy-paste error in the current code

circuit/program/src/data/record/to_bits.rs
```diff
@@ -18,78 +18,70 @@ impl<A: Aleo> ToBits for Record<A, Plaintext<A>> {
...
```
ljedrz, line 83: same note as [here](https://github.com/AleoHQ/snarkVM/pull/1836#discussion_r1272477351) - this is a change, as it seems like a copy-paste error in the current code

ljedrz:
~Missed a small `clippy` lint, will update shortly; it doesn't affect the review.~

Fixed.

circuit/algorithms/src/bhp/hash_uncompressed.rs
```diff
@@ -44,12 +44,13 @@ impl<E: Environment, const NUM_WINDOWS: u8, const WINDOW_SIZE: u8> HashUncompres
...
                     preimage.extend(self.domain.clone());
-                    preimage.extend(U64::constant(console::U64::new(input.len() as u64)).to_bits_le());
+                    U64::constant(console::U64::new(input.len() as u64)).write_bits_le(&mut preimage);
                     preimage.extend_from_slice(input_bits);
                 }
                 // Construct the subsequent iterations as: [ PREVIOUS_HASH[0..DATA_BITS] || INPUT[I * BLOCK_SIZE..(I + 1) * BLOCK_SIZE] ].
                 false => {
-                    preimage.extend(digest.to_x_coordinate().to_bits_le().into_iter().take(num_data_bits));
+                    digest.to_x_coordinate().write_bits_le(&mut preimage);
+                    preimage.truncate(num_data_bits);
```
howardwu: Can we leave these unchanged? I don't believe they have any major implication since they are already slice extensions.

howardwu: The use of truncation is error-prone if we add something in front in the future.

circuit/program/src/data/identifier/to_bits.rs
```diff
@@ -18,31 +18,31 @@ impl<A: Aleo> ToBits for Identifier<A> {
...
 impl<A: Aleo> ToBits for &Identifier<A> {
     type Boolean = Boolean<A>;
 
     /// Returns the little-endian bits of the identifier.
-    fn to_bits_le(&self) -> Vec<Self::Boolean> {
-        self.0.to_bits_le()[..8 * self.1 as usize].to_vec()
+    fn write_bits_le(&self, vec: &mut Vec<Self::Boolean>) {
+        let initial_len = vec.len();
+        self.0.write_bits_le(vec);
+        vec.truncate(initial_len + 8 * self.1 as usize);
```
howardwu, line 38: I'm not following here.

circuit/program/src/data/plaintext/to_bits.rs
```diff
@@ -18,60 +18,52 @@ impl<A: Aleo> ToBits for Plaintext<A> {
...
+                for (identifier, value) in members {
+                    let value_bits = value.to_bits_be();
+                    identifier.size_in_bits().write_bits_be(&mut bits);
+                    identifier.write_bits_be(&mut bits);
+                    U16::constant(console::U16::new(value_bits.len() as u16)).write_bits_be(&mut bits);
+                    bits.extend_from_slice(&value_bits);
+                }
+                bits
+            })),
+        };
```
howardwu: These are impossible to read

ljedrz: `rustfmt` heavily altered these parts of the diff; I applied a commit that makes them easier on the eyes for now, I can later either mark them as `#[rustfmt::skip]` or allow `rustfmt` to do its thing after the review

circuit/algorithms/src/bhp/hash_uncompressed.rs
ljedrz: this change actually removes an allocation, because now we no longer create a new vector before extending the preimage; BHP hashing is a clear "hot spot" during profiling, so it would be best if we didn't miss out on it

ljedrz: would this be better?
```
false => {
    // Record the initial length of the vector in case anything is added before this point in the future.
    let initial_len = preimage.len();
    digest.to_x_coordinate().write_bits_le(&mut preimage);
    preimage.truncate(initial_len + num_data_bits);
    preimage.extend_from_slice(input_bits);
}
```

circuit/types/scalar/src/helpers/to_bits.rs
```diff
@@ -18,29 +18,27 @@ impl<E: Environment> ToBits for Scalar<E> {
...
+    fn write_bits_le(&self, vec: &mut Vec<Self::Boolean>) {
+        vec.extend_from_slice(self.bits_le
             .get_or_init(|| {
                 // Note: We are reconstituting the scalar field into a base field.
                 // This is safe as the scalar field modulus is less than the base field modulus,
                 // and thus will always fit within a single base field element.
-                debug_assert!(
-                    console::Scalar::<E::Network>::size_in_bits() < console::Field::<E::Network>::size_in_bits()
-                );
+                debug_assert!(console::Scalar::<E::Network>::size_in_bits() < console::Field::<E::Network>::size_in_bits());
```
ljedrz: (there are no changes here, it's just a bit `rustfmt` squeezed that I missed in my new "un-formatting" commit)

circuit/algorithms/src/bhp/hash_uncompressed.rs
ljedrz: adjusted in https://github.com/AleoHQ/snarkVM/pull/1877/commits/c3d6613747e91c7f60fd70319a3e19945adbbffe.

howardwu (approved):
LGTM

PR #1880, opened 2023-08-10
Handle unwrap in `verify_deployment`

raychu86:
## Motivation

This PR is a minor fix that handles an unwrap case in `Stack::verify_deployment`.


howardwu (approved):
LGTM

PR #1883, opened 2023-08-14
Introduces Block::verify to abstract checks

howardwu:
## Motivation

Introduces Block::verify to abstract checks


raychu86 (commented):
LGTM

PR #1884, opened 2023-08-14
Further ToBits/ToBytes-related improvements

ljedrz:
A much shorter diff this time; these changes are follow-ups to https://github.com/AleoHQ/snarkVM/pull/1877 and they also lead to significant improvements related to the size and number of heap allocations.

### DIFF
--- a/console/program/src/data/identifier/parse.rs
+++ b/console/program/src/data/identifier/parse.rs
@@ -92,7 +92,7 @@ impl<N: Network> Display for Identifier<N> {
 #[cfg(test)]
 mod tests {
     use super::*;
-    use crate::data::identifier::tests::sample_identifier_as_string;
+    use crate::data::identifier::tests::{sample_identifier, sample_identifier_as_string};
     use snarkvm_console_network::Testnet3;
 
     type CurrentNetwork = Testnet3;
@@ -218,6 +218,23 @@ mod tests {
     fn test_display() -> Result<()> {
         let identifier = Identifier::<CurrentNetwork>::from_str("foo_bar")?;
         assert_eq!("foo_bar", format!("{identifier}"));
+        Ok(())
+    }
+
+    #[test]
+    fn test_proxy_bits_equivalence() -> Result<()> {
+        let mut rng = TestRng::default();
+        let identifier: Identifier<CurrentNetwork> = sample_identifier(&mut rng)?;
+
+        // Direct conversion to bytes.
+        let bytes1 = identifier.0.to_bytes_le()?;
+
+        // Combined conversion via bits.
+        let bits_le = identifier.0.to_bits_le();
+        let bytes2 = bits_le.chunks(8).map(u8::from_bits_le).collect::<Result<Vec<u8>, _>>()?;
+
+        assert_eq!(bytes1, bytes2);
+
         Ok(())
     }
 }
--- a/console/program/src/data/identifier/parse.rs
+++ b/console/program/src/data/identifier/parse.rs
@@ -70,14 +70,8 @@ impl<N: Network> Debug for Identifier<N> {
 impl<N: Network> Display for Identifier<N> {
     /// Prints the identifier as a string.
     fn fmt(&self, f: &mut Formatter) -> fmt::Result {
-        // Convert the identifier to bits.
-        let bits_le = self.0.to_bits_le();
-
-        // Convert the bits to bytes.
-        let bytes = bits_le
-            .chunks(8)
-            .map(|byte| u8::from_bits_le(byte).map_err(|_| fmt::Error))
-            .collect::<Result<Vec<u8>, _>>()?;
+        // Convert the identifier to bytes.
+        let bytes = self.0.to_bytes_le().map_err(|_| fmt::Error)?;
 
         // Parse the bytes as a UTF-8 string.
         let string = String::from_utf8(bytes).map_err(|_| fmt::Error)?;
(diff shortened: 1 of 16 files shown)


console/program/src/data/identifier/parse.rs
```diff
@@ -70,14 +70,8 @@ impl<N: Network> Debug for Identifier<N> {
...
-        // Convert the identifier to bits.
-        let bits_le = self.0.to_bits_le();
-
-        // Convert the bits to bytes.
-        let bytes = bits_le
-            .chunks(8)
-            .map(|byte| u8::from_bits_le(byte).map_err(|_| fmt::Error))
-            .collect::<Result<Vec<u8>, _>>()?;
+        // Convert the identifier to bytes.
+        let bytes = self.0.to_bytes_le().map_err(|_| fmt::Error)?;
```
ljedrz, line 74: it didn't seem like there was a reason for the twofold conversion here, but feel free to revert if I missed something

fields/src/macros.rs
```diff
@@ -144,25 +144,28 @@ macro_rules! sqrt_impl {
...
                             }
-                            q_bits.iter().enumerate().take(*l as usize).for_each(|(k, bit)| {
-                                if *bit {
+                            let mut value = *q;
+                            (0..*l as usize).for_each(|k| {
+                                let bit = value & 1 == 1;
+                                if bit {
                                     gamma *= $Self($P::POWERS_OF_ROOTS_OF_UNITY[(kappa as usize) + k], PhantomData);
                                 }
+                                value = value.wrapping_shr(1u32);
```
ljedrz, line 161: this is a non-allocating conversion that's identical to the one in the `ToBits` impl for `u64`

utilities/src/biginteger/bigint_256.rs
```diff
@@ -265,7 +265,11 @@ impl FromBits for BigInteger256 {
 impl ToBytes for BigInteger256 {
     #[inline]
     fn write_le<W: Write>(&self, writer: W) -> IoResult<()> {
-        self.0.write_le(writer)
+        let mut arr = [0u8; 8 * 4];
+        for (i, num) in self.0.iter().enumerate() {
+            arr[i * 8..][..8].copy_from_slice(&num.to_le_bytes());
+        }
+        arr.write_le(writer)
```
ljedrz: reducing the number of writes with a proxy array was found to be beneficial via profiling; the same applies to the similar change for `BigInteger256`

utilities/src/bits.rs
```diff
@@ -41,14 +41,14 @@ pub trait ToBits: Sized {
 
     /// Returns `self` as a boolean array in little-endian order.
     fn to_bits_le(&self) -> Vec<bool> {
-        let mut bits = vec![];
+        let mut bits = Vec::with_capacity(32);
```
ljedrz, line 44: this value was chosen empirically; the same applies to the remaining pre-allocations

utilities/src/biginteger/bigint_256.rs
SplittyDev: I suppose you meant the similar change for `BigInteger384`?

console/program/src/data/identifier/parse.rs
howardwu: I think the concern may have been that they might not be equivalent? Can you confirm this? (Maybe add a test so we leave a sanity check in place)

utilities/src/biginteger/bigint_256.rs
ljedrz: indeed, good catch :slightly_smiling_face:

console/program/src/data/identifier/parse.rs
ljedrz: test case added in https://github.com/AleoHQ/snarkVM/pull/1884/commits/c618efbfcc59c62159f710319c2628bd1da40aad; as anticipated, both implementations are equivalent :+1:

ljedrz:

utilities/src/biginteger/bigint_256.rs
```diff
@@ -265,7 +265,11 @@ impl FromBits for BigInteger256 {
 impl ToBytes for BigInteger256 {
     #[inline]
     fn write_le<W: Write>(&self, writer: W) -> IoResult<()> {
-        self.0.write_le(writer)
+        let mut arr = [0u8; 8 * 4];
+        for (i, num) in self.0.iter().enumerate() {
+            arr[i * 8..][..8].copy_from_slice(&num.to_le_bytes());
```
howardwu: Instead of `arr[i * 8..][..8]`, can you write `arr[i * 8..(i + 1) * 8]`?

utilities/src/biginteger/bigint_384.rs
```diff
@@ -269,7 +269,11 @@ impl FromBits for BigInteger384 {
 impl ToBytes for BigInteger384 {
     #[inline]
     fn write_le<W: Write>(&self, writer: W) -> IoResult<()> {
-        self.0.write_le(writer)
+        let mut arr = [0u8; 8 * 6];
+        for (i, num) in self.0.iter().enumerate() {
+            arr[i * 8..][..8].copy_from_slice(&num.to_le_bytes());
```
howardwu: Instead of `arr[i * 8..][..8]`, can you write `arr[i * 8..(i + 1) * 8]`?

utilities/src/biginteger/bigint_256.rs
ljedrz: sure, done in https://github.com/AleoHQ/snarkVM/pull/1884/commits/b4f9aa2c3887f5b7b3e6ee068e3dea3eccbd98b1

howardwu (approved):
LGTM

PR #1887, opened 2023-08-14
Update input visibility in `transfer_public_to_private` 

raychu86:
## Motivation

This PR updates the `credits.aleo` `transfer_public_to_private` function by changing the recipient address from public to private.


howardwu (approved):
LGTM

PR #1888, opened 2023-08-14
Add `tcm` to the `FunctionTree` for `TransitionID`

raychu86:
## Motivation

This PR adds the `tcm` to the `FunctionTree` construction in the `Transition`. This allow 0 input and 0 output transitions to have unique transition ids.

NOTE: Genesis block needs to be resampled properly


raychu86:
Closed in favor of #1902

PR #1889, opened 2023-08-15
Add `unconfirmed_id` helper method for `ConfirmedTransaction`s

raychu86:
## Motivation

This PR adds a helper method to derive the `ConfirmedTransaction`'s original `unconfirmed_id` before it was confirmed. 

In the case of an accepted deploy or execute, the id remains the same. 
In the case of a rejected deploy or execute, we regenerate the original transaction id by rebuilding the deployment or execution trees.

Note: We may want to consider adding a DataMap that stores the reference of the original transaction ids to the confirmed transaction ids. This will be helpful for users to find their rejected transactions on chain.


howardwu (approved):
LGTM

PR #1893, opened 2023-08-15
Add profiler for deployment verification of a large program

raychu86:
## Motivation

This PR adds a profiler for `verify_deployment` for a large program defined in `synthesizer/process/src/resources/large_functions.aleo`. These functions perform many `calls` to `credits.aleo` functions, which greatly increase the `verify_deployment` time.


howardwu (approved):
LGTM

PR #1894, opened 2023-08-15
[Fix] Reset `Circuit` into a consistent state.

d0cd:
This PR, sets `in_witness` to `false` when invoking `Circuit::*reset`.
This allows recovery in a consistent state in the situation where `Circuit` is rolled back while in witness mode.


howardwu (approved):
LGTM

PR #1895, opened 2023-08-15
Removes deprecated account derivation to ed25519

howardwu:
## Motivation

Removes deprecated account derivation to ed25519


raychu86 (approved):
lgtm

PR #1899, opened 2023-08-16
Speed up PoseidonGrainLFSR::new

ljedrz:
Instead of updating 30 bits individually, update them all at once with `copy_from_slice`; this appeases `clippy` and should make the operation a lot times faster (>15x when I benchmarked it locally), at least in unoptimized builds.


howardwu (approved):
LGTM

PR #1902, opened 2023-08-16
Add `TCM` to `TransitionID` and update `StatePath`

raychu86:
## Motivation

This PR adds the TCM to the transition id generation by changing the transition id to `BHP512(function_tree.root(), tcm)`. This allow 0 input and 0 output transitions to have unique transition ids.

The console and circuit `StatePath` impls have also been updated to reflect this change.

NOTE: Inclusion keys and genesis block has been resampled


circuit/program/src/state_path/verify.rs
```diff
@@ -88,7 +92,8 @@ impl<A: Aleo> StatePath<A> {
             A::verify_merkle_path_bhp(&self.block_path, &self.global_state_root, &self.block_hash.to_bits_le());
 
         // Combine the transition and transaction path checks.
-        let check_transition_and_transaction_path = check_transition_path & check_transaction_path;
+        let check_transition_and_transaction_path =
+            check_transition_path & check_transaction_path & check_transaction_leaf;
```
raychu86, line 96: We may want `check_transaction_leaf` for the `check_global` case only

howardwu (approved):
LGTM

PR #1867, opened 2023-08-06
Introduces support for `fee_private` and `fee_public` in `credits.aleo`

howardwu:
## Motivation

- Renames `credits.aleo/fee` to `credits.aleo/fee_private`
- Adds `credits.aleo/fee_public`
- Adds `Process::authorize_fee_private` and `Process::authorize_fee_public`

## Test Plan

- Adds tests to property check executions of `fee_private` and `fee_public`
- Adds unit tests to ensure `Fee` (de)serialization behaves correctly for private and public variants
- Adds unit tests to ensure that `Fee` behaves correctly during verification

PR #1904, opened 2023-08-17
Add tests for public fees

raychu86:
## Motivation

This PR adds tests for public fees and transactions using public fees.


howardwu:
Merging early as this is a continuation of #1867.

PR #1905, opened 2023-08-17
Split out the Data object from the transmission module

ljedrz:
This will allow us to later reuse it in snarkOS.


howardwu (approved):
LGTM

PR #1918, opened 2023-08-21
Use an integral log2 function in tree_depth

ljedrz:
This PR both simplifies and speeds up the `tree_depth` function used in the Merkle tree methods by changing the use of `f64::log2` to `u64::ilog2` which was introduced in Rust 1.67.

While we don't have a dedicated benchmark for `tree_depth`, in my local tests `ilog2` was almost an order of magnitude faster than `log2`.

### DIFF
--- a/console/collections/src/merkle_tree/mod.rs
+++ b/console/collections/src/merkle_tree/mod.rs
@@ -721,28 +721,15 @@ impl<E: Environment, LH: LeafHash<Hash = PH::Hash>, PH: PathHash<Hash = Field<E>
 
 /// Returns the depth of the tree, given the size of the tree.
 #[inline]
-#[allow(clippy::cast_possible_truncation)]
 fn tree_depth<const DEPTH: u8>(tree_size: usize) -> Result<u8> {
     let tree_size = u64::try_from(tree_size)?;
-    // Ensure the tree size is less than 2^52 (for casting to an f64).
-    let tree_depth = match tree_size < 4503599627370496_u64 {
-        // Compute the log2 of the tree size.
-        true => (tree_size as f64).log2(),
-        false => bail!("Tree size must be less than 2^52"),
-    };
-    // Ensure the tree depth is within a u8 range.
-    match tree_depth <= u8::MAX as f64 {
-        true => {
-            // Convert the tree depth to a u8.
-            let tree_depth = tree_depth as u8;
-            // Ensure the tree depth is within the depth bound.
-            match tree_depth <= DEPTH {
-                // Return the tree depth.
-                true => Ok(tree_depth),
-                false => bail!("Merkle tree cannot exceed depth {DEPTH}: attempted to reach depth {tree_depth}"),
-            }
-        }
-        false => bail!("Merkle tree depth ({tree_depth}) exceeds maximum size ({})", u8::MAX),
+    // Since we only allow tree sizes up to u64::MAX, the maximum possible depth is 63.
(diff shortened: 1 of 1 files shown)


console/collections/src/merkle_tree/mod.rs
```diff
@@ -724,25 +724,13 @@ impl<E: Environment, LH: LeafHash<Hash = PH::Hash>, PH: PathHash<Hash = Field<E>
...
-            // Ensure the tree depth is within the depth bound.
-            match tree_depth <= DEPTH {
-                // Return the tree depth.
-                true => Ok(tree_depth),
-                false => bail!("Merkle tree cannot exceed depth {DEPTH}: attempted to reach depth {tree_depth}"),
-            }
-        }
-        false => bail!("Merkle tree depth ({tree_depth}) exceeds maximum size ({})", u8::MAX),
+    // Since we only allow tree sizes up to u64::MAX, the maximum possible depth is 63.
+    let tree_depth = u8::try_from(tree_size.ilog2()).unwrap();
```
howardwu:
```suggestion
    let tree_depth = u8::try_from(tree_size.ilog2())?;
```

console/collections/src/merkle_tree/mod.rs
```diff
@@ -724,25 +724,13 @@ impl<E: Environment, LH: LeafHash<Hash = PH::Hash>, PH: PathHash<Hash = Field<E>
 #[allow(clippy::cast_possible_truncation)]
```
howardwu:
```suggestion
```

console/collections/src/merkle_tree/mod.rs
ljedrz: I used `unwrap` there since it's infallible, but I don't mind changing it to `?` :+1:

ljedrz:
Rebased and applied the review comments.

console/collections/src/merkle_tree/mod.rs
```diff
@@ -721,28 +721,15 @@ impl<E: Environment, LH: LeafHash<Hash = PH::Hash>, PH: PathHash<Hash = Field<E>
...
-            // Ensure the tree depth is within the depth bound.
-            match tree_depth <= DEPTH {
-                // Return the tree depth.
-                true => Ok(tree_depth),
-                false => bail!("Merkle tree cannot exceed depth {DEPTH}: attempted to reach depth {tree_depth}"),
-            }
-        }
-        false => bail!("Merkle tree depth ({tree_depth}) exceeds maximum size ({})", u8::MAX),
+    // Since we only allow tree sizes up to u64::MAX, the maximum possible depth is 63.
+    let tree_depth = u8::try_from(tree_size.ilog2())?;
```
raychu86: `ilog2` will panic if `tree_size` is 0. https://doc.rust-lang.org/std/primitive.u64.html#method.ilog2

We may want to use `checked_ilog2`

raychu86: `ilog2` will panic if `tree_size` is 0. https://doc.rust-lang.org/std/primitive.u64.html#method.ilog2

We may want to use `checked_ilog2`

ljedrz: what should we do in case the `tree_size` is `0`? throw an error or determine that the `tree_depth` is `0`? I assume the latter, as it's closer to what we have right now (`-inf as u8` is `0`)

ljedrz:
Rebased and used `checked_ilog2` instead of plain `ilog2`.

raychu86 (approved):
LGTM

howardwu (approved):

PR #1919, opened 2023-08-21
[Feature] Allow `signature`s in `ternary` instructions.

d0cd:
This PR supports the use of signature literals as operands to the `ternary` instruction.

@reviewers
I needed to implement `Signature:{zero, rand}` to be compatible with the macros.
These are only intended to be used for testing and mirrors the what we do for `address`es.
Open to other alternatives.


raychu86 (approved):
LGTM

howardwu (approved):
LGTM

PR #1922, opened 2023-08-22
Improve the performance of R1CS

ljedrz:
I found this while working on improving the performance of Keccak introduced in https://github.com/AleoHQ/snarkVM/pull/1911; with this change and a benchmark I was using locally (similar to `check_hash`), the difference in performance is as follows:

runtime: -34%
allocations: -19%
max RAM: -60%

### DIFF
--- a/circuit/environment/src/helpers/counter.rs
+++ b/circuit/environment/src/helpers/counter.rs
@@ -15,15 +15,17 @@
 use crate::*;
 use snarkvm_fields::PrimeField;
 
+use std::rc::Rc;
+
 #[derive(Debug, Default)]
 pub(crate) struct Counter<F: PrimeField> {
     scope: Scope,
-    constraints: Vec<Constraint<F>>,
+    constraints: Vec<Rc<Constraint<F>>>,
     constants: u64,
     public: u64,
     private: u64,
     nonzeros: (u64, u64, u64),
-    parents: Vec<(Scope, Vec<Constraint<F>>, u64, u64, u64, (u64, u64, u64))>,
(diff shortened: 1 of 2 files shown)


joske (approved):

ljedrz:
Rebased and added an extra small commit that hasn't changed the performance of my Keccak benchmark, but is very likely to improve the performance where multiple scopes are involved.

ljedrz:
I added another commit which compacts the constraints; it increases the number of allocations, but reduces memory use and doesn't affect runtime, so I think it's a good tradeoff.

With it, the changes to performance (of the entire PR) are as follows:
- runtime: -34% (unchanged)
- allocations: -12%
- max RAM: -69%

ljedrz:
I rebased the PR; in the end I also dropped that last commit (compacting the constraints), as it regressed the `Transaction` benchmarks. Without it, those benchmarks improved with this PR:
```
Transaction::Deploy     time:   [4.9043 s 4.9498 s 4.9916 s]
                        change: [-4.2968% -3.2415% -2.0276%] (p = 0.00 < 0.05)
                        Performance has improved.

Transaction::Deploy - verify
                        time:   [243.46 ms 245.67 ms 247.92 ms]
                        change: [-7.4173% -6.3773% -5.2248%] (p = 0.00 < 0.05)
                        Performance has improved.

Transaction::Execute(transfer)
                        time:   [5.1603 s 5.1926 s 5.2276 s]
                        change: [-2.3483% -1.4832% -0.5746%] (p = 0.01 < 0.05)
```

howardwu (approved):
LGTM

PR #1925, opened 2023-08-22
Remove timestamp from `BatchHeader` signature 

raychu86:
## Motivation

This PR removes the `timestamp` field from the `BatchHeader` `Signature` message to prevent collision (and possible replay attacks) from the `BatchCertificate` signatures.


howardwu (approved):
LGTM

PR #1926, opened 2023-08-22
Fix `is_fee_required` check

raychu86:
## Motivation

This PR fixes the `is_fee_required` condition in `Ledger::check_transaction_basic`. 

We were incorrectly deriving the `is_fee_required` variable, which meant we were skipping the fee amount check in `Ledger`, but not the fee existence check in `vm.check_transaction`. Now, we will properly check the fee amounts for transactions and properly skipping the `split` executions.

Additionally, a test has been added to check that transactions (execution and deployments) with sufficient fees are valid and insufficient fees are invalid.


howardwu (approved):
LGTM

PR #1932, opened 2023-08-24
Introduce `supply` mapping

raychu86:
## Motivation

This PR introduces the `supply` mapping to `credits.aleo` and handles the updates in the `VM::speculate` and `VM::finalize` operations. 

Each of the following conditions will update the supply mapping:
 - Ratify::Genesis
 - Ratify::BlockReward
 - Ratify::PuzzleReward
 - Specific transactions
     - bond_public
     - unbond_public
     - transfer_private_to_public
     - transfer_public_to_private
     - fee_private
     - fee_public


ledger/block/src/transition/mod.rs
```diff
@@ -309,6 +309,44 @@ impl<N: Network> Transition<N> {
         self.program_id.to_string() == "credits.aleo" && self.function_name.to_string() == "unbond"
     }
 
+    /// Returns `true` if this is a `bond_public` transition.
```
kpp, line 312: Frankly speaking I have no idea why this function is marked as inline because there are so much code inside of them including at least 4 memory allocations.

PR #1942, opened 2023-08-28
Renames `Struct` to `StructType`

howardwu:
## Motivation

Renames `Struct` to `StructType`.

This is to align with the naming convention of all the other data type instantiations.


raychu86 (approved):
LGTM

PR #1948, opened 2023-08-30
Reduce proving key size

vicsn:
## Motivation

This PR reduces the size of our proving keys by ~35% (depending on the circuit size). More improvements follow in future PRs. In detail:
- There was no need for the prover to store polynomials representing the indexed circuit (`{row, col, rowcol, rowcolval}`), because their only purpose was to be used as part of the matrix sumcheck linear combination polynomial - for creating an opening/evaluation proof at the challenge `gamma`. However, we already store evaluations to those indexed polynomials, and during proving compute `a(X)` and `b(X)`. Those polynomials are enough for the prover to create the opening proof.
- I also removed some related dead code which confused me for hours. During the commitment opening phase, we do not need commitments, only the polynomials. 
- Moreover, we also never used the evaluation in the `BatchLCProof`, they are stored separately in Proof.

## Test Plan

Added tests for proving key and verifying key size.

## Documentation
This PR does not change the spec, but you can find it in https://github.com/AleoHQ/protocol-docs/pull/17/files#diff-062489691f6f4a9431ceb1873d411b037df6fa32112827197f94f451391284b8


algorithms/src/snark/varuna/ahp/ahp.rs
```diff
@@ -333,50 +326,44 @@ impl<F: PrimeField, MM: SNARKMode> AHPForR1CS<F, MM> {
...
-            let sum_c = sums_fourth_msg[i].sum_c;
-
-            Self::add_g_m_term(&mut matrix_sumcheck, id, "a", challenges, g_a_at_gamma, v_rc, d_a, sum_a, s_a, RC)?;
-            Self::add_g_m_term(&mut matrix_sumcheck, id, "b", challenges, g_b_at_gamma, v_rc, d_b, sum_b, s_b, RC)?;
-            Self::add_g_m_term(&mut matrix_sumcheck, id, "c", challenges, g_c_at_gamma, v_rc, d_c, sum_c, s_c, RC)?;
-
-            linear_combinations.insert(g_a.label.clone(), g_a);
-            linear_combinations.insert(g_b.label.clone(), g_b);
-            linear_combinations.insert(g_c.label.clone(), g_c);
+            let RC = state_i.constraint_domain.size_as_field_element * state_i.variable_domain.size_as_field_element;
```
Pratyush: let's use lowercase `rc` here

algorithms/src/snark/varuna/ahp/ahp.rs
```diff
@@ -333,50 +326,44 @@ impl<F: PrimeField, MM: SNARKMode> AHPForR1CS<F, MM> {
...
+                let g_m = LinearCombination::new(g_m_label.clone(), [(F::one(), g_m_label)]);
+                let g_m_at_gamma = evals.get_lc_eval(&g_m, gamma)?;
+
+                let label_a_poly = format!("circuit_{id}_a_poly_{m}");
+                let a_poly = LinearCombination::new(label_a_poly.clone(), [(F::one(), label_a_poly)]);
+                let a_poly_missing = evals.get_lc_eval(&a_poly, gamma).is_err();
+                let label_b_poly = format!("circuit_{id}_b_poly_{m}");
+                let b_poly = LinearCombination::new(label_b_poly.clone(), [(F::one(), label_b_poly)]);
+                let b_poly_missing = evals.get_lc_eval(&b_poly, gamma).is_err();
+                assert_eq!(a_poly_missing, b_poly_missing);
```
Pratyush: What logic is going on here? It could do with a comment, but why would `a_poly` and `b_poly` ever be missing at this point?

Pratyush: Ah I see now. Maybe we can rewrite this to call a "evaluate_a_and_polys` method, which we implement differently for the prover and the verifier.

algorithms/src/snark/varuna/ahp/indexer/circuit.rs
```diff
@@ -126,31 +133,22 @@ impl<F: PrimeField, MM: SNARKMode> Circuit<F, MM> {
...
-            &self.c_arith.row,
-            &self.a_arith.row_col,
-            &self.b_arith.row_col,
-            &self.c_arith.row_col,
-            &self.a_arith.row_col_val,
-            &self.b_arith.row_col_val,
-            &self.c_arith.row_col_val,
-        ]
-        .into_iter()
+    pub fn arithmetize_polys(&self) -> impl Iterator<Item = LabeledPolynomial<F>> {
```
Pratyush: Maybe we should find a better name for this; the polynomials are already the arithmetized versions

algorithms/src/snark/varuna/ahp/matrices.rs
```diff
@@ -189,30 +173,54 @@ pub(crate) fn matrix_evals<F: PrimeField>(
...
+    pub row_col: LabeledPolynomial<F>,
+    /// LDE of the vector containing entry-wise products of `row`, `col` and the non-zero entries of M.
+    pub row_col_val: LabeledPolynomial<F>,
+}
+
+impl<F: PrimeField> MatrixArithmetization<F> {
+    /// Create a new MatrixArithmetization
+    pub fn new(id: &CircuitId, label: &str, matrix_evals: &MatrixEvals<F>) -> Result<MatrixArithmetization<F>> {
+        let interpolate_time = start_timer!(|| "Interpolating on K");
+        let non_zero_domain = matrix_evals.row.domain();
```
Pratyush: This is a confusing name; let's add a `matrix_evals.domain()` method.

algorithms/src/snark/varuna/varuna.rs
```diff
@@ -525,50 +530,39 @@ where
...
             .chain(fourth_commitment_randomnesses)
             .chain(fifth_commitment_randomnesses)
             .collect();
 
+        let empty_randomness = Randomness::<E>::empty();
         if !MM::ZK {
-            let empty_randomness = Randomness::<E>::empty();
             assert!(commitment_randomnesses.iter().all(|r| r == &empty_randomness));
+        } else {
+            assert!(commitment_randomnesses.iter().any(|r| r != &empty_randomness));
```
Pratyush: Let's switch the cases so that we can say `if MM::ZK` (avoiding the negation)

algorithms/src/snark/varuna/ahp/prover/round_functions/fourth.rs
```diff
@@ -200,9 +206,7 @@ impl<F: PrimeField, MM: SNARKMode> AHPForR1CS<F, MM> {
         let new_constant = v_R_i_alpha_v_C_i_beta * constraint_domain.size_inv * variable_domain.size_inv;
         batch_inversion_and_mul(&mut inverses, &new_constant);
```
Pratyush: Let's give `new_constant` a more informative name

algorithms/src/snark/varuna/ahp/prover/round_functions/fourth.rs
```diff
@@ -146,35 +151,36 @@ impl<F: PrimeField, MM: SNARKMode> AHPForR1CS<F, MM> {
...
         job_pool.add_job(|| {
             let a_poly_time = start_timer!(|| format!("Computing a poly for {label}"));
             let a_poly = {
-                let coeffs = cfg_iter!(arithmetization.row_col_val.as_dense().unwrap().coeffs())
-                    .map(|v| v_R_i_alpha_v_C_i_beta * v)
-                    .collect();
-                DensePolynomial::from_coefficients_vec(coeffs)
+                let evals = cfg_iter!(row_col_val.evaluations).map(|v| v_R_i_alpha_v_C_i_beta * v).collect();
+                EvaluationsOnDomain::from_vec_and_domain(evals, non_zero_domain)
+                    .interpolate_with_pc(ifft_precomputation)
```
Pratyush, line 178: Hm looks like we're adding an extra IFFT here now, which is non-ideal...

If you have the time, it would be good to figure out the impact of this on the prover time

algorithms/src/snark/varuna/ahp/ahp.rs
vicsn: Done 68f294cea

algorithms/src/snark/varuna/ahp/ahp.rs
vicsn: Done 68f294cea

Note that I wanted to avoid passing a magic `prover: bool` to `construct_linear_combinations`, so instead I'm directly inferring whether a_poly and b_poly are present in `evals`.

algorithms/src/snark/varuna/ahp/indexer/circuit.rs
vicsn: Done 68f294cea

algorithms/src/snark/varuna/ahp/matrices.rs
vicsn: Done 68f294cea

algorithms/src/snark/varuna/varuna.rs
vicsn: Done 68f294cea

algorithms/src/snark/varuna/ahp/prover/round_functions/fourth.rs
vicsn: Done 68f294cea

algorithms/src/snark/varuna/ahp/prover/round_functions/fourth.rs
vicsn: Indeed, however fortunately overall time for a typical `test_transfer_public_transaction_size` even goes *down*. One sample (I ran multiple):
- proving time before: 98.84s user 12.05s system 766% cpu 14.475 total
- proving time after: 96.65s user 10.57s system 772% cpu 13.881 total

The reduction may come from:
- the prover is aggregating a smaller number of polynomials during opening (from 4 (row,col,rowcol,rowcolval) to 2 (a,b) indexing polynomials per circuit)
- we no longer uselessly aggregate commitments before opening: https://github.com/AleoHQ/snarkVM/pull/1948/files#diff-1c2269e15b2c9eb1b102866e39b60f7149b01d43503a3bdd02081fc7200a358cL478

Pratyush: Lovely! (Though I imagine the commitment aggregation step is dwarfed by other steps; it's a tiny MSM). I wonder if there's similar redundancies elsewhere. E.g., could we apply the same idea to the first sumcheck? We're already computing the random linear combinations of the polynomials there, so we shouldn't have to repeat work *again* inside the PC scheme.

vicsn: Yes interesting thought. But I don't see obvious contenders. Will make a TODO to investigate if we can re-use the entire "left hand side" in the lineval sumcheck.

Pratyush: I think the savings would happen here. 

In `construct_linear_combinations`, we take a linear combination of the `w_j`s: https://github.com/AleoHQ/snarkVM/blob/b87eb05a93703ddb8a9b4bff53a5d129d92741a2/algorithms/src/snark/varuna/ahp/ahp.rs#L298. This is called when we invoke `SonicPC::open`.

Meanwhile, in https://github.com/AleoHQ/snarkVM/blob/b87eb05a93703ddb8a9b4bff53a5d129d92741a2/algorithms/src/snark/varuna/ahp/prover/round_functions/third.rs#L313, we compute z_m, but we can already multiply in the random linear combiners into `assignment` before hand, and save this result.

(Also, separately, I think in the latter file we don't need to compute the individual `g_i`s and `h_i`s; we should just directly compute `g_1` and `h_1` by multiply the `z_m_i` with the vanishing poly for the `column_i` domain.)

Overall I suspect these changes should knock off another 5-10% off the time.

vicsn: Cool, thx for sharing the details, appreciate it. Will tackle in a next PR!

algorithms/src/polycommit/sonic_pc/polynomial.rs
```diff
@@ -97,6 +97,11 @@ impl<F: Field> LabeledPolynomial<F> {
         &self.info.label
     }
 
+    /// Return the label for `self`.
+    pub fn clone_label(&self) -> String {
```
howardwu: This is odd as a Rust paradigm

Pratyush: We can just call it `to_label`

Pratyush (approved):

algorithms/src/snark/varuna/ahp/prover/round_functions/fourth.rs
Pratyush: Let's extract this off into another issue (it's just a prover-side improvement, and so can be done at a later stage)

algorithms/src/polycommit/sonic_pc/polynomial.rs
vicsn: d03b543ef

howardwu:
Resampling parameters

howardwu:
Deployment generation improved by 2.5%
Transaction generation improved by 4.5%
Proving keys reduced by 35-40%

howardwu (approved):

PR #1949, opened 2023-08-31
Adjust clap features

ljedrz:
As of `4.4.1`, the `unstable-styles` feature is no longer needed for color support.

note: `clap`'s MSRV is `1.70` now, so this PR also updates it in the `TOML`.


howardwu (approved):
LGTM

PR #1951, opened 2023-08-31
Varuna Test Vectors

vicsn:
## Motivation

The Varuna proof system underlying the Aleo protocol is built upon previous work on the Marlin proof system. Since Aleo is the first concrete implementation of Varuna, test vectors should be provided in order to provide any other concrete implementations of the proof system an indication that the proof system has been implemented correctly.

A secondary implementation of the Varuna proof system in has been created in [Sagemath](https://github.com/AleoHQ/varuna-sage-impl) by @mdelle1 which lead to the same test vectors provided in this PR (given the same randomness)

This PR Implements:

1. Initial test vectors that include:
* A set of challenges produced by the Varuna Verifier (in `algorithms/src/snark/varuna/resources/challenges.input`)
* A circuit description over which the Varuna proof system runs (in `algorithms/src/snark/varuna/resources/instances.input`)
* The expected intermediate oracle polynomials produced by the Varuna Prover (in `algorithms/src/snark/varuna/resources/polynomials`)
* The expected `non_zero`, `variable`, and `constraint` domains (in `algorithms/src/snark/varuna/resources/domain`)

2. A unit test over the test vectors that runs the Varuna proof system over the expected test vectors that ensures all intermediate polynomials and domains generated by the Varuna proof system match the test vectors. This test will run every ci run and should catch any regressions introduced into the proof system. 

## Related PRs

https://github.com/AleoHQ/varuna-sage-impl

## Future work
- Add a readme to explain how to run the test vectors
- Add the same regression tests to the Sagemath implementation in ci
- Add test vectors for multiple circuits


algorithms/src/snark/varuna/data_structures/test_circuit.rs
```diff
@@ -96,6 +97,9 @@ impl<F: Field> TestCircuit<F> {
             public_inputs.push(new_var);
         }
 
+        println!("a: {}", a);
```
vicsn: Please remove all prints

utilities/src/rand.rs
```diff
@@ -140,3 +140,39 @@ impl rand::RngCore for TestRng {
...
+
+    fn fill_bytes(&mut self, dest: &mut [u8]) {
+        for byte in dest.iter_mut() {
+            *byte = self.state.pop().unwrap() as u8;
+        }
+    }
+
+    fn try_fill_bytes(&mut self, dest: &mut [u8]) -> Result<(), rand::Error> {
+        for byte in dest.iter_mut() {
+            *byte = self.state.pop().unwrap() as u8; // TODO: use try_from and convert the Error
```
vicsn: TODO left here. If it is too much work, we can also remove the TODO, this is just a test anyway

algorithms/src/snark/varuna/tests.rs
```diff
@@ -524,3 +524,293 @@ mod varuna_hiding {
...
+        if !path.exists() {
+            fs::create_dir_all(&path).unwrap_or_else(|_| panic!("Failed to create resources folder: {path:?}"));
+        }
+        // Output the path.
+        path
+    }
+
+    /// Loads the given `test_folder/test_file` and asserts the given `candidate` matches the expected values.
+    #[track_caller]
+    fn assert_snapshot<S1: Into<String>, S2: Into<String>>(test_folder: S1, test_file: S2, candidate: &str) {
```
vicsn: They are not snapshots (as we use the term elsewhere), so maybe rename `assert_test_vector` or something more generic

algorithms/src/snark/varuna/tests.rs
```diff
@@ -524,3 +524,293 @@ mod varuna_hiding {
...
+        fs::write(&path, data).unwrap_or_else(|_| panic!("Failed to write to file: {:?}", path));
+    }
+
+    #[test]
+    fn test_proving_system_with_test_vectors() {
+        run_prover_sage_vectors(false);
+    }
+
+    #[test]
+    #[ignore]
```
vicsn: Perhaps this is something we want to be even more careful about. Imagine a future developer runs `--include-ignored` and then accidentally commits potentially updated test_vectors. Perhaps we can comment it out with a comment: `in case we want to generate new test vectors for other implementations, run this`.

algorithms/src/snark/varuna/tests.rs
```diff
@@ -524,3 +524,293 @@ mod varuna_hiding {
...
+    fn create_prover_test_vectors() {
+        run_prover_sage_vectors(true);
+    }
+
+    // TODO: make macros to use different hiding modes
```
vicsn: I would remove the TODO, moved it here: https://github.com/orgs/AleoHQ/projects/13?pane=issue&itemId=40044078

algorithms/src/snark/varuna/tests.rs
```diff
@@ -524,3 +524,293 @@ mod varuna_hiding {
...
+            Fr::from_str(challenges[4]).unwrap(),
+            vec![Fr::from_str(challenges[5]).unwrap()],
+            vec![Fr::from_str(challenges[6]).unwrap()],
+            vec![Fr::from_str(challenges[7]).unwrap()],
+            Fr::from_str(challenges[8]).unwrap(),
+        );
+        let circuit_combiner = Fr::one();
+        let instance_combiners = vec![Fr::one()];
+
+        let mut random_state = vec![0u64; 100];
```
vicsn:
```suggestion
        let mut rng_state = vec![0u64; 100];
```

algorithms/src/snark/varuna/tests.rs
```diff
@@ -524,3 +524,293 @@ mod varuna_hiding {
...
+        let add_f_to_state = |s: &mut Vec<u64>, f: Fp256<FrParameters>| {
+            // Fp384 field elements sample 6 u64 values in total.
+            for i in 0..f.0.0.len() {
+                s.push(f.0.0[f.0.0.len() - 1 - i]);
+            }
+        };
+        for witness in instance[15].split(',') {
```
vicsn: I added a TODO that we can test the other instance values in the future: https://github.com/orgs/AleoHQ/projects/13?pane=issue&itemId=40044078

algorithms/src/snark/varuna/tests.rs
```diff
@@ -524,3 +524,293 @@ mod varuna_hiding {
...
+            if witness.trim() == "" {
+                continue;
+            }
+            add_f_to_state(&mut random_state, Fr::from_str(witness.trim()).unwrap());
+        }
+
+        let rng = &mut snarkvm_utilities::rand::TestMockRng::fixed(random_state);
+
+        let max_degree = AHPForR1CS::<Fr, MM>::max_degree(100, 25, 300).unwrap();
```
vicsn: Can you make these values a function of the instance file values? We just need a big enough setup.

algorithms/src/snark/varuna/tests.rs
```diff
@@ -524,3 +524,293 @@ mod varuna_hiding {
...
+        let universal_srs = VarunaSonicInst::universal_setup(max_degree).unwrap();
+        let mul_depth = 3;
+        let num_constraints = 7;
+        let num_variables = 7;
+
+        // TODO: pass the right randomness in
```
vicsn: I guess this TODO can be removed

algorithms/src/snark/varuna/tests.rs
```diff
@@ -524,3 +524,293 @@ mod varuna_hiding {
...
+        let (index_pk, _index_vk) = VarunaSonicInst::circuit_setup(&universal_srs, &circ).unwrap();
+        let mut keys_to_constraints = BTreeMap::new();
+        keys_to_constraints.insert(index_pk.circuit.deref(), std::slice::from_ref(&circ));
+
+        let prover_state = AHPForR1CS::<_, MM>::init_prover(&keys_to_constraints, rng).unwrap();
+        let mut prover_state = AHPForR1CS::<_, MM>::prover_first_round(prover_state, rng).unwrap();
+        let first_round_oracles = Arc::new(prover_state.first_round_oracles.as_ref().unwrap());
+
+        // Get private witness polynomial coefficients.
+        // TODO: Support batched circuits
```
vicsn: Remove these and the same TODOs below, added here https://github.com/orgs/AleoHQ/projects/13?pane=issue&itemId=40044078

vicsn (commented):
FYI just preparing you, the most "painful" rebase might be because of this one: https://github.com/AleoHQ/snarkVM/pull/1875

algorithms/src/snark/varuna/tests.rs
iamalwaysuncomfortable: Good point. I just removed the #[test] macro from it ultimately.

algorithms/src/snark/varuna/tests.rs
iamalwaysuncomfortable: Yeah actually, it should be a fixed randomness that other implementations are capable of replicating.

utilities/src/rand.rs
iamalwaysuncomfortable: Fixed it with an `ok_or`

algorithms/src/snark/varuna/tests.rs
iamalwaysuncomfortable: Changed!

vicsn:
I can't review because I made the PR, but LGTM (noting that mentioned improvements can be done in future PRs). I'd also be curious for either Pratyush' or @raychu86's opinion regarding the architecture - this tight coupling means that any change to the proof system will subsequently also require a change to our SageMath implementation.

utilities/src/rand.rs
```diff
@@ -140,3 +141,39 @@ impl rand::RngCore for TestRng {
 }
 
 impl rand::CryptoRng for TestRng {}
+
+/// A cache of numbers LARPing as an RNG used **solely** for testing and benchmarking, **not** for any real world purposes.
+pub struct TestMockRng {
```
howardwu: Are there any special features of this RNG that make it inadvisable to use `TestRng`?

vicsn: Yes, we load it with predetermined "random" numbers. This was a hack to import randomly generated challenges from the Sage script.

Thinking about it again, the other direction is likely way cleaner and by now very easy to adjust to. @iamalwaysuncomfortable if you agree you could adjust the code so:
1. let the test use `TestRng::fixed` with a hardcoded seed (per set of test vectors), and we get rid of `TestMockRng`
2. run it once with `create_test_vectors` to export test vectors.
3. adjust the Sage code to read the challenges instead of [randomly generating them](https://github.com/search?q=repo%3AAleoHQ%2Fvaruna-sage-impl%20random_element&type=code)

howardwu: Tagging @ljedrz here for visibility, as he helped design the original `TestRng`.

@ljedrz we have a SAGE program that generates the randomness for test vectors and would like to unify them with the Rust test's RNG usage so the test vectors will match.

ljedrz: The suggestion by @vicsn to generate test vectors using `TestRng::fixed` and to later reuse it sounds good to me :+1:; we should only make sure to comment on the fact why a fixed seed is used in those tests.

iamalwaysuncomfortable: > adjust the Sage code to read the challenges instead of [randomly generating them](https://github.com/search?q=repo%3AAleoHQ%2Fvaruna-sage-impl%20random_element&type=code)

As of a few weeks ago, the sagemath code already reads from the test vector challenges, so I'll update the rng now and just re-regenerate the test vectors.

iamalwaysuncomfortable: > This was a hack to import randomly generated challenges from the Sage script.

Also, yes we no longer need this because the Sage script already reads test vectors from file. It's been deleted.

howardwu (approved):

howardwu:
LGTM, this PR requires resolving a merge conflict.

iamalwaysuncomfortable:
Merge conflict resolved with a rebase

PR #1955, opened 2023-09-05
Use a single committer_key and fft_precomputation for batch of proofs

vicsn:
## Motivation

To reduce memory and computation for batch proofs, we should only have to use a single `committer_key` and `fft_precomputation` for the biggest circuit of a batch proof. Previously we were computing these for every circuit. For a large batch this reduces the size of the `ProvingKey` by around 40%. We also don't need `CommitterUnionKey` anymore which was collecting the union of the committer keys.

## Test Plan

The existing test suite seems to suffice.


algorithms/src/polycommit/kzg10/data_structures.rs
```diff
@@ -14,8 +14,10 @@
 
 use crate::{
     fft::{DensePolynomial, EvaluationDomain},
+    prelude::PCError,
     AlgebraicSponge,
 };
+use anyhow::bail;
```
raychu86: nit: Move this import down and group with the other `anyhow` import

algorithms/src/polycommit/kzg10/data_structures.rs
```diff
@@ -108,6 +134,118 @@ impl<E: PairingEngine> UniversalParams<E> {
             prepared_negative_powers_of_beta_h: self.powers.read().prepared_negative_powers_of_beta_h(),
         })
     }
+
+    pub fn to_committer_key(
```
Pratyush: Is this a port of the `CommitterUnionKey` code?

Pratyush: Seems like a combination of that + `trim`

algorithms/src/snark/varuna/ahp/indexer/circuit.rs
```diff
@@ -123,14 +110,19 @@ impl<F: PrimeField, MM: SNARKMode> Circuit<F, MM> {
...
-        crate::fft::EvaluationDomain::<F>::new(self.index_info.num_variables).unwrap().size()
+    /// The maximum domain size.
+    pub fn max_domain_size(&self) -> usize {
+        let size = [
+            2 * self.index_info.num_constraints,
+            2 * self.index_info.num_variables,
+            2 * self.index_info.num_non_zero_a,
+            2 * self.index_info.num_non_zero_b,
+            2 * self.index_info.num_non_zero_c,
+        ]
```
Pratyush: Let's document where these numbers come from

algorithms/src/snark/varuna/ahp/indexer/circuit_info.rs
```diff
@@ -42,6 +45,20 @@ impl CircuitInfo {
...
+        let num_variables = self.num_variables;
+        let num_non_zero_a = self.num_non_zero_a;
+        let num_non_zero_b = self.num_non_zero_b;
+        let num_non_zero_c = self.num_non_zero_c;
+        [
+            EvaluationDomain::<F>::compute_size_of_domain(num_variables).unwrap() - 2,
+            EvaluationDomain::<F>::compute_size_of_domain(num_non_zero_a).unwrap() - 2,
+            EvaluationDomain::<F>::compute_size_of_domain(num_non_zero_b).unwrap() - 2,
+            EvaluationDomain::<F>::compute_size_of_domain(num_non_zero_c).unwrap() - 2,
+        ]
```
Pratyush: Ditto

algorithms/src/snark/varuna/data_structures/circuit_proving_key.rs
```diff
@@ -34,16 +31,12 @@ pub struct CircuitProvingKey<E: PairingEngine, MM: SNARKMode> {
...
-    pub committer_key: Arc<sonic_pc::CommitterKey<E>>,
 }
 
 impl<E: PairingEngine, MM: SNARKMode> ToBytes for CircuitProvingKey<E, MM> {
     fn write_le<W: Write>(&self, mut writer: W) -> io::Result<()> {
         CanonicalSerialize::serialize_compressed(&self.circuit_verifying_key, &mut writer)?;
-        CanonicalSerialize::serialize_compressed(&self.circuit, &mut writer)?;
-
-        self.committer_key.write_le(&mut writer)
+        Ok(CanonicalSerialize::serialize_compressed(&self.circuit, &mut writer)?)
```
Pratyush, line 39: This is great!

Pratyush (approved):

algorithms/src/polycommit/kzg10/data_structures.rs
vicsn: It is a port of `trim`, the union structs and logic are complete superfluous in this new setup. We just collect our max needed sizes (and in a future PR the union of lagrange sizes), and then trim in one go accordingly.

algorithms/src/polycommit/kzg10/data_structures.rs
vicsn: Ah oh I thought cargo fmt did that for me, fixed and taking into account for future PRs: dd8eea044

algorithms/src/snark/varuna/ahp/indexer/circuit.rs
vicsn: 72b4d063a

algorithms/src/snark/varuna/ahp/indexer/circuit_info.rs
vicsn: 72b4d063a

algorithms/benches/snark/varuna.rs
```diff
@@ -66,13 +67,19 @@ fn snark_prove(c: &mut Criterion) {
...
 
         let (circuit, _) = TestCircuit::gen_rand(mul_depth, num_constraints, num_variables, rng);
 
-        let params = VarunaInst::circuit_setup(&universal_srs, &circuit).unwrap();
-        b.iter(|| VarunaInst::prove(universal_prover, &fs_parameters, &params.0, &circuit, rng).unwrap())
+        let (pk, _) = VarunaInst::circuit_setup(&universal_srs, &circuit).unwrap();
+        let max_degree = pk.circuit.max_degree();
+        let max_domain_size = pk.circuit.max_domain_size();
+        let coefficient_support = pk.circuit.index_info.get_degree_bounds::<Fr>();
+        let universal_prover = &universal_srs
```
Pratyush: Maybe this API should take as input a bunch of `CircuitInfo`s so that it can perform the above calculation internally?

Pratyush: That way we don't have to peer inside the contents of the `pk` to construct `universal_prover`.

algorithms/src/snark/varuna/ahp/indexer/circuit.rs
```diff
@@ -123,14 +110,19 @@ impl<F: PrimeField, MM: SNARKMode> Circuit<F, MM> {
...
+            2 * self.index_info.num_constraints, // zerocheck poly degree
+            2 * self.index_info.num_variables,   // lineval sumcheck poly degree
+            2 * self.index_info.num_non_zero_a,  // matrix sumcheck poly degree
+            2 * self.index_info.num_non_zero_b,  // matrix sumcheck poly degree
+            2 * self.index_info.num_non_zero_c,  // matrix sumcheck poly degree
+        ]
+        .into_iter()
+        .max()
+        .unwrap();
+        crate::fft::EvaluationDomain::<F>::compute_size_of_domain(size).unwrap()
```
Pratyush: Should this also live with `CircuitInfo`, just like `get_degree_bounds`?

Pratyush: (In general there are bunch of such utility functions which are kind of distributed between `AHPForR1CS`, `Circuit`, and `CircuitInfo`; these should be audited and we should clarify why each is in its current place (or move it if appropriate))

algorithms/src/snark/varuna/tests.rs
```diff
@@ -61,6 +65,17 @@ mod varuna {
                         let (index_pk, index_vk) = $snark_inst::circuit_setup(&universal_srs, &circ).unwrap();
                         println!("Called circuit setup");
 
+                        let max_domain_size = index_pk.circuit.max_domain_size();
+                        let max_degree = index_pk.circuit.max_degree();
+                        let coefficient_support = index_pk.circuit.index_info.get_degree_bounds::<Fr>();
```
Pratyush: Yeah this seems to be a code smell; we shouldn't have to extract these details just to pass them in again.

algorithms/benches/snark/varuna.rs
Pratyush: If not a `CircuitInfo`, maybe a `DegreeAndDomainInfo` or something like that.

Pratyush (changes requested):
I think we should change the API of `to_universal_prover` to take in a struct (maybe `DegreeInfo`?) containing the relevant information; as it stands we're just extracting some `usize`s from other structs and then passing these into `to_universal_prover`, which is creating a lot of repetition and room for errors stemming from accidentally switching the order of arguments.

vicsn:
> I think we should change the API of to_universal_prover to take in a struct (maybe DegreeInfo?) containing the relevant information; as it stands we're just extracting some usizes from other structs and then passing these into to_universal_prover, which is creating a lot of repetition and room for errors stemming from accidentally switching the order of arguments.

Agreed, in the past I also lost half an hour testing a bug caused by false ordering. [3383c94](https://github.com/AleoHQ/snarkVM/pull/1955/commits/3383c94abdbb491b2e9ab6f5874172b8f4b43bf7)

Also worth noting: I renamed `max_domain_size` to `max_fft_size` because that is the only thing we use that particular domain/degree for at the moment.

algorithms/benches/snark/varuna.rs
vicsn: > Maybe this API should take as input a bunch of CircuitInfos so that it can perform the above calculation internally?

Just so we're on the same page, this is a slighly different recommendation than you give below. We can pass either `pk`, `pk.circuit`, `pk.circuit.index_info` or `pk.circuit.index_info.degree_info()` into universal prover. I'm doing the latter now, which seems cleaner than passing the entire `pk`.

algorithms/src/snark/varuna/ahp/indexer/circuit.rs
vicsn: Very much agree, I left `max_degree(...)` within ahp.rs because it is also used in other tests to directly infer the size needed for the SRS.

[3383c94](https://github.com/AleoHQ/snarkVM/pull/1955/commits/3383c94abdbb491b2e9ab6f5874172b8f4b43bf7)

algorithms/benches/snark/varuna.rs
```diff
@@ -66,13 +67,15 @@ fn snark_prove(c: &mut Criterion) {
...
-        let universal_prover = &universal_srs.to_universal_prover().unwrap();
         let fs_parameters = FS::sample_parameters();
+        let hiding_bound = AHPForR1CS::<Fr, VarunaHidingMode>::zk_bound().unwrap_or(0);
 
         let (circuit, _) = TestCircuit::gen_rand(mul_depth, num_constraints, num_variables, rng);
 
-        let params = VarunaInst::circuit_setup(&universal_srs, &circuit).unwrap();
-        b.iter(|| VarunaInst::prove(universal_prover, &fs_parameters, &params.0, &circuit, rng).unwrap())
+        let (pk, _) = VarunaInst::circuit_setup(&universal_srs, &circuit).unwrap();
+        let degree_info = pk.circuit.index_info.degree_info::<Fr, VarunaHidingMode>();
```
iamalwaysuncomfortable, line 75: One small but significant consideration here is that if we support Wasm targets for the web (which I think at some point may constitute one of the most common places where people run Varuna proofs) the parameter downloads need to be fetched using asynchronous methods (they'll fail if called from a synchronous trim method).

This degree info is the key piece of information needed to download parameters, so I'd like to propose that either the `UniversalProver` (or an enclosing struct) has an async method for fetching these parameters (like an async version of `to_commiter_key` or a method like `preload_parameters`) OR at the very least, that the degree info be available so that the caller can handle the concern loading these parameters into the `UniversalProver`

algorithms/src/srs/universal_prover.rs
```diff
@@ -12,11 +12,21 @@
...
+    polycommit::sonic_pc::CommitterKey,
+};
 use snarkvm_curves::PairingEngine;
 
 /// `UniversalProver` is used to compute evaluation proofs for a given commitment.
-#[derive(Clone, Debug, Default, PartialEq, Eq)]
+#[derive(Clone, Debug, PartialEq, Eq)]
 pub struct UniversalProver<E: PairingEngine> {
+    /// The committer key for the underlying KZG10 scheme.
+    pub committer_key: CommitterKey<E>,
```
iamalwaysuncomfortable, line 25: There should be some method included in this struct that allows parameter fetching (or checked extension perhaps if developers store their own powers of tau - say for offline proofs on offline devices) in a web friendly way. This effectively means an asynchronous parameter fetching path. Ditto for any constructor methods for the `UniversalProver` within other structs that has this as a member.

iamalwaysuncomfortable (commented):
Made some comments about the web/wasm targets. The tl;dr of it is that I conjecture that web & wasm targets will make up an outsized portion of users proving with Aleo's implementation of Varuna (people like to use JavaScript, Rust is difficult for most application developers).

Therefore  that there should be some asynchronous parameter download methods baked into enclosing structs which construct the UniversalProver & Universal verifiers. An example method for which to fetch these parameters is proposed here: https://github.com/AleoHQ/snarkVM/pull/1987/files#diff-d907035b6432d9b2ef95015b23710393694a19b111dfffc195499ff3da02c3bcR112-R126

This PR provides  good decoupling and actually will make it easier to get parameters on the web, but we should ensure we include it so that web developers are fully supported prior to mainnet.

vicsn:
>  Made some comments about the web/wasm targets. The tl;dr of it is that I conjecture that web & wasm targets will make up an outsized portion of users proving with Aleo's implementation of Varuna (people like to use JavaScript, Rust is difficult for most application developers).

> Therefore that there should be some asynchronous parameter download methods baked into enclosing structs which construct the UniversalProver & Universal verifiers. An example method for which to fetch these parameters is proposed here: https://github.com/AleoHQ/snarkVM/pull/1987/files#diff-d907035b6432d9b2ef95015b23710393694a19b111dfffc195499ff3da02c3bcR112-R126

> This PR provides good decoupling and actually will make it easier to get parameters on the web, but we should ensure we include it so that web developers are fully supported prior to mainnet.

For other reviewers reading this, we talked about it elsewhere and for now this PR can stay as is - PR #1987 will be rebased on this one.

Pratyush:
I think we should roll supported Lagrange sizes and hiding_bound also into DegreeInfo

vicsn:
> I think we should roll supported Lagrange sizes and hiding_bound also into DegreeInfo

I would vote against that because we have tests which generate [lagrange_sizes/hiding_bound independently from CircuitInfo](https://github.com/AleoHQ/snarkVM/blob/testnet3/algorithms/src/polycommit/sonic_pc/mod.rs#L701). Both ways of doing does not seem obviously better and does not seem worth the (minor) refactor:
1. let `CircuitInfo::degree_info()` return an unfinished *mutable* `DegreeInfo`, which we then populate with `lagrange_sizes` and `hiding_bound`. This makes `DegreeInfo` less clean of an object and is not less prone to errors.
2. get `max_degree/max_fft/degree_bounds` separately from `CircuitInfo` and compose `DegreeInfo` outside of `CircuitInfo`. This regresses in terms of complexity, as the developer has to order the arguments correctly.

Pratyush:
In any real world use case, `CircuitInfo` completely determines `DegreeInfo`. In tests, we should be mutating `DegreeInfo` to change `lagrange_sizes` and `hiding_bound`. 

Also, the maximum supported degree of the SRS can depend on whether or not the committed polynomials are hiding. So logically `hiding_bound` affects `circuit_info` anyway.

Pratyush:
I would also be fine if `to_universal_prover` *always* output all the SRS elements that correspond to hiding (This is fine, as there's only a small number of them). I just don't want there to be a knob that the user can forget to turn that suddenly means that there's no hiding.

vicsn:
That makes sense, looks a lot cleaner now, also removed the `Default` which could lead to 0 hiding bound. f17ce31e9

algorithms/benches/snark/varuna.rs
```diff
@@ -191,7 +188,7 @@ fn snark_batch_verify(c: &mut Criterion) {
             all_circuits.push(circuits);
             all_inputs.push(inputs);
             let degree_info_i = pk.circuit.index_info.degree_info::<Fr, VarunaHidingMode>();
-            degree_info.union(&degree_info_i);
+            degree_info = degree_info.map(|i: DegreeInfo| i.union(&degree_info_i).unwrap()).or(Some(degree_info_i));
```
Pratyush, line 191:
```suggestion
            degree_info = degree_info.and_then(|i: DegreeInfo| i.union(&degree_info_i)).or_else(degree_info_i);
```

algorithms/src/polycommit/kzg10/degree_info.rs
```diff
@@ -0,0 +1,50 @@
...
+    pub degree_bounds: Option<HashSet<usize>>,
+    /// hiding bound for polynomial queries
+    pub hiding_bound: usize,
+    /// supported lagrange-basis SRS
+    pub lagrange_sizes: Option<HashSet<usize>>,
+}
+
+impl DegreeInfo {
+    pub fn union(self, other: &Self) -> Result<Self> {
+        ensure!(self.hiding_bound == other.hiding_bound);
```
Pratyush: We don't need this `ensure`, right? The union should just return the maximum of the two `hiding_bound`s. If we do that, then the output of this method can be a `Self`, which will clean up a bunch of other code.

algorithms/src/polycommit/kzg10/degree_info.rs
```diff
@@ -0,0 +1,50 @@
...
+        let hiding_bound = self.hiding_bound;
+        let max_degree = self.max_degree.max(other.max_degree);
+        let max_fft_size = self.max_fft_size.max(other.max_fft_size);
+        let mut new_bounds = self.degree_bounds.unwrap_or_default();
+        if let Some(other_bounds) = &other.degree_bounds {
+            new_bounds = new_bounds.union(other_bounds).copied().collect();
```
Pratyush:
```suggestion
            new_bounds = new_bounds | other_bounds;
```

algorithms/src/polycommit/kzg10/degree_info.rs
```diff
@@ -0,0 +1,50 @@
...
+        }
+        let degree_bounds = (!new_bounds.is_empty()).then_some(new_bounds);
+        let mut new_l_sizes = self.lagrange_sizes.unwrap_or_default();
+        if let Some(other_l_sizes) = &other.lagrange_sizes {
+            new_l_sizes = new_l_sizes.union(other_l_sizes).copied().collect();
```
Pratyush:
```suggestion
            new_l_sizes = new_l_sizes | other_l_sizes;
```

Pratyush (approved):
LGTM modulo couple of small changes

algorithms/src/polycommit/kzg10/degree_info.rs
vicsn: Are you suggested I impl a new operator? Cause otherwise: `no implementation for std::collections::HashSet<usize> | std::collections::HashSet<usize>`

algorithms/src/polycommit/kzg10/degree_info.rs
vicsn: [04d55d4](https://github.com/AleoHQ/snarkVM/pull/1955/commits/04d55d42c90ac9b4b5b0405a31e6949ef6926a26)

algorithms/src/polycommit/kzg10/degree_info.rs
Pratyush: Ah maybe you have to use references

vicsn: Sweet, I will check the Rust docs next time - 6acf4be72

console/network/src/testnet3.rs
```diff
@@ -199,14 +200,11 @@ impl Network for Testnet3 {
...
     /// Returns the Varuna universal prover.
-    fn varuna_universal_prover() -> &'static UniversalProver<Self::PairingCurve> {
-        static INSTANCE: OnceCell<UniversalProver<<Console as Environment>::PairingCurve>> = OnceCell::new();
-        INSTANCE.get_or_init(|| {
-            snarkvm_algorithms::polycommit::kzg10::UniversalParams::load()
-                .expect("Failed to load universal SRS (KZG10).")
-                .to_universal_prover()
-                .expect("Failed to convert universal SRS (KZG10) to the prover.")
-        })
+    fn varuna_universal_prover(degree_info: DegreeInfo) -> UniversalProver<Self::PairingCurve> {
```
ljedrz, line 205: I see that we can no longer keep `UniversalProver` as a `'static`, but what about the `UniversalParams`? those could likely be made `'static` so that we don't need to load them for every call to `ProvingKey::{prove, prove_batch}`

ljedrz (approved):
LGTM with a suggestion regarding `UniversalParams`

algorithms/src/polycommit/kzg10/degree_info.rs
```diff
@@ -0,0 +1,48 @@
...
+impl DegreeInfo {
+    pub fn union(self, other: &Self) -> Self {
+        let hiding_bound = self.hiding_bound.max(other.hiding_bound);
+        let max_degree = self.max_degree.max(other.max_degree);
+        let max_fft_size = self.max_fft_size.max(other.max_fft_size);
+        let mut new_bounds = self.degree_bounds.unwrap_or_default();
+        if let Some(other_bounds) = &other.degree_bounds {
+            new_bounds = &new_bounds | &other_bounds;
+        }
+        let degree_bounds = (!new_bounds.is_empty()).then_some(new_bounds);
```
Pratyush: Sorry for the nitpicking, I think this code is a bit easier to understand.

```suggestion
        let degree_bounds = match (&self.degree_bounds, &other.degree_bounds) {
            (Some(a), Some(b)) => Some(a | b),
            (Some(a), None) | (None, Some(a)) => Some(a),
            (None, None) => None,
        };
```

algorithms/src/polycommit/kzg10/degree_info.rs
```diff
@@ -0,0 +1,48 @@
...
+        let mut new_l_sizes = self.lagrange_sizes.unwrap_or_default();
+        if let Some(other_l_sizes) = &other.lagrange_sizes {
+            new_l_sizes = &new_l_sizes | &other_l_sizes;
+        }
+        let lagrange_sizes = (!new_l_sizes.is_empty()).then_some(new_l_sizes);
```
Pratyush:
```suggestion
        let lagrange_sizes = match (&self.lagrange_sizes, &other.lagrange_sizes) {
            (Some(a), Some(b)) => Some(a | b),
            (Some(a), None) | (None, Some(a)) => Some(a),
            (None, None) => None,
        };
```

Pratyush (approved):

console/network/src/testnet3.rs
vicsn: Good idea! This would be a somewhat involved change so not sure I can get it in before the code freeze, will tackle in a new PR.

Currently, every time we create a proof, we indeed create `UniversalParams` and grow this according to the circuit size. For a simple transfer it grows to ~ 50MB. But it could grow to gigabytes. If we were to make the object static, we'll also need to add a method to shrink the parameters to our minimum parameter size (~ 10MB) so they don't take up memory forever.

algorithms/src/polycommit/kzg10/degree_info.rs
vicsn: 7bfb1e669

algorithms/src/polycommit/kzg10/degree_info.rs
vicsn: 7bfb1e669

howardwu:
Due to some time-sensitive needs for these memory improvements, I'm going to take a stab at updating this PR to be mergable.

PR #1957, opened 2023-09-05
Document LinearCombination

vicsn:
## Motivation

This puzzled me and others for a while. Also adding the description to our protocol spec.

### DIFF
--- a/circuit/environment/src/helpers/linear_combination.rs
+++ b/circuit/environment/src/helpers/linear_combination.rs
@@ -21,6 +21,16 @@ use core::{
 };
 use indexmap::{map::Entry, IndexMap};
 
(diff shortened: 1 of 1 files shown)


circuit/environment/src/helpers/linear_combination.rs
```diff
@@ -21,6 +21,16 @@ use core::{
 };
 use indexmap::{map::Entry, IndexMap};
 
+// Before high level program operations are converted into constraints, they are first tracked as linear combinations.
+// Each linear combination corresponds to a single row of an R1CS matrix, and consists of:
```
acoglio:
```suggestion
// Each linear combination corresponds to a portion or all of a single row of an R1CS matrix, and consists of:
```

acoglio (approved):

PR #1958, opened 2023-09-05
[WIP] Initial metrics support

niklaslong:
This PR adds a metrics crate to snarkVM. I've chosen not to add a prometheus exporter here (meaning there's no need for a `tokio` dependency) but rather to register the metrics with snarkOS. Corresponding PR: https://github.com/AleoHQ/snarkOS/pull/2643. I haven't implemented any specific metrics yet but these will need to be feature-guarded.


niklaslong:
Superseded by #1974.

PR #1960, opened 2023-09-06
Assert TestCircuit is as expected

vicsn:
## Motivation

It is clearer for developers if the number of constraints and variables of the TestCircuit corresponds to what was pass into the test.


howardwu:
We should `cfg(feature = "test")` guard this struct.

You'll need to add:
```toml
[features]
test = [ ]
```
to the Cargo.toml.

vicsn:
If I wrap the struct and tests in `#[cfg(feature = "test")]`, which we don't use anywhere else in the code, the tests don't run anymore with `cargo test`.
So wrapped it in `#[cfg(any(test, feature = "test"))]`, this we do use elsewhere already.

howardwu (approved):

PR #1967, opened 2023-09-08
Cache the current committee in ledger

niklaslong:
Currently, the latest committee is retrieved from storage, caching it on the `Ledger` makes the data accessible with less overhead. Useful for https://github.com/AleoHQ/snarkOS/pull/2562. 

I _think_ all edge cases are handled i.e. `VM::from` will load or initialise the committee so it can simply be read from storage but I'm not perfectly familiar with all the intricacies involved here so any input would be appreciated.


ledger/src/lib.rs
```diff
@@ -161,13 +163,16 @@ impl<N: Network, C: ConsensusStorage<N>> Ledger<N, C> {
         let vm = VM::from(store)?;
         lap!(timer, "Initialize a new VM");
 
+        let current_committee = vm.finalize_store().committee_store().current_committee().ok();
```
raychu86, line 167: Not saying we do this, but technically we can get rid of the `Option` around the `Committee` cache since we ensure that the genesis block must be added to the ledger and already contains a `Committee`. This means we can directly extract the committee from the `Ratify::Genesis` for the case where Ledger starts as empty.

Something like 
```
let current_committee = ledger.vm.block_store().heights().max() {
  None => // Extract committee from genesis block.
  Some(_) => vm.finalize_store().committee_store().current_committee()?
};
```
Though, the current approach is likely cleaner.

raychu86 (approved):
LGTM

howardwu (approved):
LGTM

PR #1688, opened 2023-06-15
[Test] Snapshot Testing for Constraint Counts.

d0cd:
## Motivation

This PR:
- Introduces `UpdatableCount`, a struct that stores the desired number of constants, public variables, private variables, and constraints.

## Usage
1. Initialize an `UpdatableCount` using `count_is!` or `count_less_than!`.
2. Use `UpdatableCount::assert_matches` to check that the number of constants, variables, and constraints match.
3. Set an environment variable `UPDATE_COUNT=<query>` while running `cargo test` to update invocations of `count_is!` and `count_less_than!` with correct numbers (on failing cases) in the source files whose name contains the substring `<query>`.

Example. To update counts in `downcast.rs`, run:
```
UPDATE_COUNT=downcast cargo test
```

If you would like to see update statistics, enable the `nocapture` flag.
```
UPDATE_COUNT=downcast cargo test -- --nocapture
```

## Notes
The updated counts are only accurate for the concrete cases that are tested.
When using this feature, you may have to run your tests (with `UPDATE_COUNT` set) multiple times to reach a fixed point.

PR #1968, opened 2023-09-08
[Feature] Casts between literal types.

d0cd:
This PR 
- adds support for casting between the remaining unimplemented literal types.
- ensures that the `FromBits` implementations for native Rust integers check appropriate size

Depends on #1688


circuit/program/src/data/literal/cast.rs
```diff
@@ -0,0 +1,287 @@
...
+            Self::Signature(..) => bail!("Cannot cast a signature literal to another type."),
+            Self::String(..) => bail!("Cannot cast a string literal to another type."),
+        }
+    }
+}
+
+/// Casts a boolean literal to the given literal type.
+fn cast_boolean_to_type<A: Aleo>(boolean: &Boolean<A>, to_type: LiteralType) -> Result<Literal<A>> {
+    match to_type {
+        LiteralType::Address => Ok(Literal::Address(boolean.cast())),
```
raychu86: These `cast` and cast_lossy` calls seem easily macro-able. Might be better for cleanliness

console/types/src/lib.rs
```diff
@@ -16,8 +16,22 @@
 #![allow(clippy::too_many_arguments)]
 #![warn(clippy::cast_possible_truncation)]
 
+#[cfg(feature = "default")]
```
raychu86, line 19: Why are the `default` cfg flags necessary?

utilities/src/bits.rs
```diff
@@ -155,6 +163,14 @@ macro_rules! impl_bits_for_integer {
             /// Reads `Self` from a boolean array in big-endian order.
             #[inline]
             fn from_bits_be(bits: &[bool]) -> Result<Self> {
+                // If the number of bits exceeds the size of the integer, ensure that the upper bits are all zero.
+                // Note that because the input bits are big-endian, these are the bits at the beginning of slice.
+                if bits.len() > (<$int>::BITS as usize) {
+                    for bit in bits[0..(bits.len() - (<$int>::BITS as usize))].iter() {
+                        ensure!(!bit, "upper bits are not zero");
+                    }
+                }
```
raychu86: A `saturating_sub` would remove the need for the `if` check

```suggestion
                for bit in bits[0..(bits.len().saturating_sub(<$int>::BITS as usize))].iter() {
                    ensure!(!bit, "upper bits are not zero");
                }     
```

utilities/src/bits.rs
```diff
@@ -146,6 +146,14 @@ macro_rules! impl_bits_for_integer {
             /// Reads `Self` from a boolean array in little-endian order.
             #[inline]
             fn from_bits_le(bits: &[bool]) -> Result<Self> {
+                // If the number of bits exceeds the size of the integer, ensure that the upper bits are all zero.
+                // Note that because the input bits are little-endian, these are the bits at the end of slice.
+                if bits.len() > (<$int>::BITS as usize) {
+                    for bit in bits[<$int>::BITS as usize..].iter() {
+                        ensure!(!bit, "upper bits are not zero");
+                    }
+                }
```
raychu86: A `skip` should replicate this logic and get rid of the `if` statement.

```suggestion
                for bit in bits.iter().skip(<$int>::BITS as usize) {
                    ensure!(!bit, "upper bits are not zero");
                }
```

console/program/src/data/literal/cast.rs
```diff
@@ -0,0 +1,311 @@
...
+        + AsPrimitive<u64>
+        + AsPrimitive<u128>
+        + AsPrimitive<i8>
+        + AsPrimitive<i16>
+        + AsPrimitive<i32>
+        + AsPrimitive<i64>
+        + AsPrimitive<i128>,
+{
+    match to_type {
+        LiteralType::Address => Ok(Literal::Address(integer.cast_lossy()?)),
```
raychu86: Ditto above: Possibly use a macro for cleanliness

circuit/types/src/cast/integer.rs
```diff
@@ -0,0 +1,1305 @@
...
+                    }
+                    Integer::<E, I1>::from_bits_le(&bits_le[..(I1::BITS as usize)])
+                }
+            },
+            // If the two types are both signed, perform the required checks.
+            (true, true) => match I0::BITS <= I1::BITS {
+                // If the source type is smaller than or equal to the destination type, sign extend the source integer.
+                // Then instantiate the new integer from the bits.
+                true => {
+                    bits_le.extend(std::iter::repeat(self.msb().clone()).take((I1::BITS - I0::BITS) as usize));
```
raychu86: Use `saturating_sub` instead of `I1::BITS - I0::BITS` can likely get rid of the internal match case.

circuit/types/src/cast/integer.rs
```diff
@@ -0,0 +1,1305 @@
...
+            // If the two types are both unsigned, instantiate the new integer from the bits.
+            (false, false) => Integer::<E, I1>::from_bits_le(&bits_le),
+            // If the source type is unsigned and the destination type is signed, perform the required checks.
+            (false, true) => match I0::BITS < I1::BITS {
+                // If the source type is smaller than the destination type, instantiate the new integer from the bits.
+                true => Integer::<E, I1>::from_bits_le(&bits_le),
+                // If the source type is the same size or larger than the destination type, check that the most significant bits are zero.
+                // Then instantiate the new integer from the lower bits.
+                false => {
+                    for bit in &bits_le[((I1::BITS - 1) as usize)..] {
```
raychu86: Use `.iter().skip(..)` instead of indexing. Also use `saturating_sub` instead of `I1::BITS - 1`

circuit/types/src/cast_lossy/integer.rs
```diff
@@ -0,0 +1,1249 @@
...
+        if I0::BITS < I1::BITS {
+            match I0::is_signed() {
+                // If the source type is signed, then sign extend.
+                true => bits_le.extend(std::iter::repeat(self.msb().clone()).take((I1::BITS - I0::BITS) as usize)),
+                // Otherwise, zero extend.
+                false => {
+                    bits_le.extend(std::iter::repeat(Boolean::constant(false)).take((I1::BITS - I0::BITS) as usize))
+                }
+            }
+        }
```
raychu86: Can be cleaned up a bit like this.

```suggestion
        let padding = match I0::is_signed() {
            // If the source type is signed, then sign extend.
            true => self.msb.clone(),
            // Otherwise, zero extend.
            false => Boolean::constant(false),
        };
        bits_le.extend(std::iter::repeat(padding).take(I1::BITS.saturating_sub(I0::BITS) as usize))
```

console/types/src/lib.rs
d0cd: `default` is necessary because the other dependent crates (like `console_types_address`, `console_types_integers`, etc) are made available under that flag.

circuit/types/src/cast/integer.rs
d0cd: The match case is still needed as we need to perform additional checks when casting a larger signed integer into a smaller one.

howardwu (approved):

circuit/types/src/cast/boolean.rs
```diff
@@ -0,0 +1,189 @@
...
+    #[inline]
+    fn cast(&self) -> Field<E> {
+        Field::from_boolean(self)
+    }
+}
+
+impl<E: Environment> Cast<Group<E>> for Boolean<E> {
+    /// Casts a `Boolean` to a `Group`.
+    #[inline]
+    fn cast(&self) -> Group<E> {
```
howardwu, line 45: I think this case will always fail on `true`

PR #1969, opened 2023-09-10
Add self.parent opcode & thus unlock composable programs

evanmarshall:
This is an updated version of: https://github.com/AleoHQ/snarkVM/pull/1178

## Motivation

Building the https://github.com/AleoHQ/ARCs/discussions/11 self.parent functionality will unlock a wide variety of applications to be built on top of Aleo, including anything that relates to DeFi and NFTs. As an application builder in the Aleo ecosystem, I want to be able to use self.parent functionality, so I can build robust smart contracts involving token swaps, escrow contracts, and NFTs. In particular, when programs are able to know which contract is calling them, access controls can be expanded beyond just users. This allows callee programs with public state in mappings to assign state to program addresses, and only allow calls by that program address to modify that state. The Aleo-equivalent of ERC-20 can leverage this feature to be a fully functional implementation of the ERC-20 contract.

### Reentrancy attacks
This shouldn't really be possible given that the `call` opcode needs to match against a `locator` ie `some_program.aleo/some_function`. Since `self.parent` and `self.caller` compile to an address and not a locator, it shouldn't be possible for a program to use: `call self.parent/some_function ...args` any more than it's possible to use `call self.caller/some_function ..args`

If this is still a concern, we can further restrict the usage of `self.parent` to only be used as an argument to a finalize statement. Note, it cannot be used within a finalize block, only as an argument from the transition to the finalize block. Restricting as such would still serve every significant use case I can imagine.

### Deriving the private key that leads to the address
If someone can recover the private key that derives a public key that matches the program, this would be a majority vulnerability. Someone could easily clear the balance held by a program.

This shouldn't be any easier than guessing a private key from a public key.
The program address is calculated as `HashToGroup(program_id)` and a normal address is calculated as `address := pk_sig + pr_sig + pk_prf`

### Programs owning records
This could lead to programs that create records with an owner of `self.parent` which would effectively make them unusable forever. I don't think this is a big consideration as it's already possible to randomly generate a public key and then transfer a record to it. This may perhaps just make it more likely someone does this. I wouldn't consider it a large consideration but leo should perhaps be updated to throw warnings if someone tries to set the owner of a record to `self.parent`

## Related PRs

* https://github.com/AleoHQ/snarkVM/pull/1178
* https://github.com/AleoHQ/welcome/pull/139
* https://github.com/AleoHQ/ARCs/discussions/11

### DIFF
--- a/circuit/program/src/request/mod.rs
+++ b/circuit/program/src/request/mod.rs
@@ -118,6 +118,8 @@ impl<A: Aleo> ToFields for InputID<A> {
 pub struct Request<A: Aleo> {
     /// The request caller.
     caller: Address<A>,
+    /// The request parent, i.e. the caller/program directly invoking this request.
+    parent: Address<A>,
     /// The network ID.
     network_id: U16<A>,
     /// The program ID.
@@ -211,6 +213,7 @@ impl<A: Aleo> Inject for Request<A> {
 
         Self {
             caller: Address::new(mode, *request.caller()),
+            parent: Address::new(mode, *request.parent()),
             network_id: U16::new(Mode::Constant, *request.network_id()),
             program_id: ProgramID::new(Mode::Constant, *request.program_id()),
             function_name: Identifier::new(Mode::Constant, *request.function_name()),
@@ -229,6 +232,11 @@ impl<A: Aleo> Request<A> {
     /// Returns the request caller.
     pub const fn caller(&self) -> &Address<A> {
         &self.caller
+    }
+
+    /// Returns the request parent.
+    pub const fn parent(&self) -> &Address<A> {
+        &self.parent
     }
 
     /// Returns the network ID.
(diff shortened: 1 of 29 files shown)


d0cd:
Thanks for the proposal @evanmarshall!
I believe that there is also another security consideration, that `self.parent` can be forged in the circuit,
In other words, a malicious prover sets `self.parent` to an arbitrary address.

evanmarshall:
> Thanks for the proposal @evanmarshall! I believe that there is also another security consideration, that `self.parent` can be forged in the circuit, In other words, a malicious prover sets `self.parent` to an arbitrary address.

That's a good point and one potential fix is that we could treat `self.parent` as a public input and it would be the responsibility of the validators verify proper execution.

Another potential solution would be to move `self.parent` to only the finalize blocks & similarly require the validators to verify.

In any case, I think it will be necessary to reveal the parent address publicly to enable verification.

What would your ideal solution look like? 

I'm happy to iterate on this.

d0cd:
Leaning towards the first solution, where `self.parent` is a public input to the circuit.
Currently prototyping a solution.

PR #1983, opened 2023-09-13
Prevent addition overflow

vicsn:
## Motivation

We should avoid panics, this PR adds checking if additions are allowed in snarkVM. I could not find problematic subtraction cases.

For this, we should agree on the attack surface. I can imagine the following attack vectors:

- an attacker lets different nodes misinterpret a serialized object, which can happen if our flags are serialized last. Fortunately I haven't seen this anywhere.
- an attacker's frontend lets an honest user serialize a giant object so their local node panics, I don't think this is a huge issue
- an attacker lets an honest node deserialize a giant object so it panics. This might kill the network so we should prevent this. Nodes already seem to have a [maximum message size](https://github.com/AleoHQ/snarkOS/blob/1ec292b8d3d71013f43084f663e4e7b926e8c170/node/messages/src/helpers/codec.rs#L26), but as defense in depth I use saturating_add in serialized_size. Using checked_add there is tricky as it will change the serialized_size trait function signature all over the codebase, including in macros and for types which we know have a fixed size.
    

## Test Plan

The existing tests should cover it.

## Related PRs

Succeeding: https://github.com/AleoHQ/snarkVM/pull/1584
Compared to that PR, here we *only* care about fixing addition overflow and we are rebased on testnet3

### DIFF
--- a/algorithms/src/snark/varuna/data_structures/proof.rs
+++ b/algorithms/src/snark/varuna/data_structures/proof.rs
@@ -71,17 +71,15 @@ impl<E: PairingEngine> Commitments<E> {
     }
 
     fn serialized_size(&self, compress: Compress) -> usize {
-        let mut size = 0;
-        size += serialized_vec_size_without_len(&self.witness_commitments, compress);
-        size += CanonicalSerialize::serialized_size(&self.mask_poly, compress);
-        size += CanonicalSerialize::serialized_size(&self.h_0, compress);
-        size += CanonicalSerialize::serialized_size(&self.g_1, compress);
-        size += CanonicalSerialize::serialized_size(&self.h_1, compress);
-        size += serialized_vec_size_without_len(&self.g_a_commitments, compress);
-        size += serialized_vec_size_without_len(&self.g_b_commitments, compress);
-        size += serialized_vec_size_without_len(&self.g_c_commitments, compress);
-        size += CanonicalSerialize::serialized_size(&self.h_2, compress);
-        size
+        serialized_vec_size_without_len(&self.witness_commitments, compress)
+            .saturating_add(CanonicalSerialize::serialized_size(&self.mask_poly, compress))
+            .saturating_add(CanonicalSerialize::serialized_size(&self.h_0, compress))
+            .saturating_add(CanonicalSerialize::serialized_size(&self.g_1, compress))
+            .saturating_add(CanonicalSerialize::serialized_size(&self.h_1, compress))
+            .saturating_add(serialized_vec_size_without_len(&self.g_a_commitments, compress))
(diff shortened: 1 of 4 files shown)


algorithms/src/snark/varuna/data_structures/proof.rs
```diff
@@ -140,12 +138,11 @@ impl<F: PrimeField> Evaluations<F> {
...
 
     fn serialized_size(&self, compress: Compress) -> usize {
-        let mut size = 0;
-        size += CanonicalSerialize::serialized_size(&self.g_1_eval, compress);
-        size += serialized_vec_size_without_len(&self.g_a_evals, compress);
-        size += serialized_vec_size_without_len(&self.g_b_evals, compress);
-        size += serialized_vec_size_without_len(&self.g_c_evals, compress);
-        size
+        let size = 0usize;
+        size.saturating_add(CanonicalSerialize::serialized_size(&self.g_1_eval, compress))
```
howardwu:
```suggestion
        CanonicalSerialize::serialized_size(&self.g_1_eval, compress)
```

algorithms/src/snark/varuna/data_structures/proof.rs
```diff
@@ -71,17 +71,15 @@ impl<E: PairingEngine> Commitments<E> {
...
-        size += CanonicalSerialize::serialized_size(&self.h_0, compress);
-        size += CanonicalSerialize::serialized_size(&self.g_1, compress);
-        size += CanonicalSerialize::serialized_size(&self.h_1, compress);
-        size += serialized_vec_size_without_len(&self.g_a_commitments, compress);
-        size += serialized_vec_size_without_len(&self.g_b_commitments, compress);
-        size += serialized_vec_size_without_len(&self.g_c_commitments, compress);
-        size += CanonicalSerialize::serialized_size(&self.h_2, compress);
-        size
+        0usize
+            .saturating_add(serialized_vec_size_without_len(&self.witness_commitments, compress))
```
howardwu:
```suggestion
        serialized_vec_size_without_len(&self.witness_commitments, compress)
```

howardwu (changes requested):
I think we can remove all usage of `checked_add` in these cases. It is unlikely we would wrap around a `usize` assuming a `32-bit` or `64-bit` architecture, because our SRS is only up to `2^27`.

howardwu:
I am open to changing our usage of `usize` to a `u64` to truly mitigate this issue as well.

vicsn:
Pushed and squashed the changes.

> I am open to changing our usage of usize to a u64 to truly mitigate this issue as well.

For that topic, I will review and rebase this PR later: https://github.com/AleoHQ/snarkVM/pull/1499

howardwu (approved):

PR #1987, opened 2023-09-15
[Feature] Async parameter preloading

iamalwaysuncomfortable:
## Motivation

Currently parameter downloads in SnarkVM in a `wasm` environment use a synchronous version of `XmlHttpRequest` to download  parameters. However this has been considered deprecated behavior for browsers for a long time and enforce suboptimal restrictions on them. Further, upcoming changes to `Chrome Manifest v3` are not allowing `XmlHttpRequest` to be used in chrome extensions whatsoever and are enforcing the move to fetch. 

Chrome manifest V3 changes are explained [here](https://developer.chrome.com/docs/extensions/migrating/to-service-workers/#replace-xmlhttprequest)

This has the following consequences for parameter & key downloads during program execution:
1. **Downloads have 3x Memory Footprint:** XmlHttpRequest in syncronous mode only allows text encodings (byte encodings are not allowed by modern browsers in synchronous mode). This has the effect of requiring the 3x the memory (i.e. a 250mb proving key requires 750mb memory reserved in `wasm` for the download) and degrading end user experience in both performance & time waiting for network i/o.
2. **Chrome Extension Support is currently limited:** Because XmlHttpRequest is deprecated for new chrome apps, chrome extensions which perform executions that involve parameter or key downloads will not be able to built without resorting to suboptimal techniques such as popup windows.
3. **Risk of deprecation:** If web browsers decide to eliminate support for synchronous XmlHttpRequest, parameter downloads will cease to work in the web.

To support modern web standards & enable future chrome apps to be built on Aleo, this PR introduces an option to download necessary powers asynchronously prior to program execution. The asynchronous downloads use the `reqwest` library which uses fetch under the hood in `wasm` environments (therefore eliminating reliance on` XmlHttpRequest`) and enables bytes to be downloaded directly without needing a conversion to text encoding beforehand. 

This enables both Chrome extensions & NodeJS to perform a normal synchronous execution flow without having to reach out to the web for parameters during execution.  Acceptance of the asynchronous download path in this PR would reduce `wasm` memory needed by 3x, enable support for modern chrome extensions, de-risk possible future deprecation of synchronous XmlHttpRequests in web browsers. Further, the option to use `fetch` (instead of XmlHttpRequest) for parameter downloads enables Aleo program key synthesis & Deployment within NodeJS.

## Test Plan

Tests to test changes made in order to support this have have been added as well as a test demonstrating program execution in wasm after parameter pre-downloads

## Related PRs

 https://github.com/AleoHQ/snarkVM/pull/1955 proposes a single committer key. In #1955  - ProvingKeys don't contain CommiterKeys (this is delegated to a `UniversalProver`). ProvingKeys will continue to contain Degree Information required to download parameters for the Universal Prover and the `preload_powers` method in this PR would be provide a Browser/Wasm friendly path for downloading these parameters.


algorithms/src/polycommit/kzg10/data_structures.rs
```diff
@@ -28,13 +29,17 @@ use snarkvm_utilities::{
...
 
-use crate::srs::{UniversalProver, UniversalVerifier};
-use anyhow::Result;
+use anyhow::{anyhow, ensure, Result};
 use core::ops::{Add, AddAssign};
 use parking_lot::RwLock;
 use rand_core::RngCore;
 use std::{collections::BTreeMap, io, ops::Range, sync::Arc};
 
+const MAX_POWER_OF_TWO: usize = 28;
```
vicsn: Any way that we can import these from parameters/src/testnet3/mod.rs?

parameters/src/testnet3/powers.rs
```diff
@@ -529,3 +716,140 @@ impl<E: PairingEngine> ToBytes for PowersOfBetaG<E> {
...
+        assert_eq!(powers.num_powers(), NUM_POWERS_16);
+
+        // Ensure powers can't be extended with powers of the wrong length.
+        let result = powers.extend_normal_powers_checked(&powers_17, NUM_POWERS_16);
+        assert!(result.is_err());
+        let result = powers.extend_normal_powers_checked(&powers_17, NUM_POWERS_18);
+        assert!(result.is_err());
+        assert_eq!(powers.num_powers(), NUM_POWERS_16);
+
+        // Ensure contiguous powers CAN be extended.
```
vicsn, line 751: Is it new behaviour in this PR that we can only extend by the next power?

parameters/src/testnet3/powers.rs
```diff
@@ -529,3 +716,140 @@ impl<E: PairingEngine> ToBytes for PowersOfBetaG<E> {
...
+        assert_eq!(shifted_powers_estimate, (vec![NUM_POWERS_18, NUM_POWERS_17, NUM_POWERS_16], true));
+
+        // Download parameters and ensure new range estimates are correct.
+        powers.download_powers_for(initial_range).unwrap();
+        powers.download_powers_for(initial_shifted_range.clone()).unwrap();
+
+        // Get extended estimates and ensure they are correct.
+        let normal_power_estimate = powers.estimate_powers_for(&extended_range).unwrap();
+        let extended_power_estimate = powers.estimate_powers_for(&extended_shifted_range).unwrap();
+        assert_eq!(normal_power_estimate, (vec![NUM_POWERS_19, NUM_POWERS_20], false));
```
vicsn, line 852: Is there any way to test this without creating a giant vector?

wasm/src/tests.rs
```diff
@@ -12,10 +12,18 @@
 // See the License for the specific language governing permissions and
 // limitations under the License.
 
+const NUM_POWERS_15: usize = 1 << 15;
```
vicsn: Can we import these from parameters?

iamalwaysuncomfortable: Yes it should be fine to import them just for the tests.

algorithms/src/polycommit/kzg10/data_structures.rs
iamalwaysuncomfortable: They were previously private, but I think it should be fine to make them public.

parameters/src/testnet3/powers.rs
iamalwaysuncomfortable: That's a question. There was previously no method for extending powers manually, and so this PR adds that specific behavior. However, at the moment is it legal/prudent to download non-contigous powers?

vicsn: I was just thinking that if we know we want to download `NUM_POWERS_20`, why iteratively download 18 to 19 to 20 (which the code does if I read and recall correctly). On the other hand, I can't imagine this to introduce a huge overhead so this can also be changed later if it ever becomes a performance issue.

iamalwaysuncomfortable: Yeah, I think you're right about this though if one needs to commit to polynomials with a mix of high and low degree terms and nothing in the middle it's non-sensical to download it in series all the way towards those terms. I'll change the logic to accomodate this.

vicsn: >  if one needs to commit to polynomials with a mix of high and low degree terms 
Then you need the powers adhering to the biggest polynomial.

The only reason to download the lower powers, is if just downloading the higher power doesn't include that lower power yet.

PR #1992, opened 2023-09-17
Move verifiers to local parameters, include 2^16 in local

howardwu:
## Motivation

Move verifiers to local parameters, include 2^16 in local


raychu86 (approved):
Lgtm

PR #1994, opened 2023-09-17
Adds support for `cast.lossy` opcode

howardwu:
## Motivation

This PR updates the `cast` opcode to align with the new design for the `cast.lossy` opcode.

The implementation of `cast.lossy` under the hood has been updated to ensure there is no fallibility when converting between types, at the cost of lossy truncation. To ensure conversions from `field` to `group`/`address` are infallible, `Elligator2` is used as a fallback, when recovering from the x-coordinate is invalid.

In addition, the following changes are introduced:
- Adds `circuit::Address::zero()`
- Adds `circuit::Group::generator()`
- Adds `circuit::Integer::to_scalar()` and `console::Integer::to_scalar()`
- A benchmark for `Elligator2` is added to `console/algorithms/benches`.

## Testing

- Extensive unit tests are added to each casting operator for `cast_lossy` in `console`
- All unit tests in `circuit` for `cast` are updated with the new counts

## Reviewers' Note

The most important change worth a careful review is that of the `Field -> Group` `cast_lossy` implementation in `circuit/program/src/data/literal/cast_lossy/field.rs`. You should compare this against the `console` version to understand what the intended logic is supposed to be. (The `console` version is human-readable).

In addition, the impl of [`circuit::Group::from_xy_coordinates()`](https://github.com/AleoHQ/snarkVM/pull/1994/files#diff-e3592a499cc78d2ca1fb836e88e09d17f7f65720adab28ee78788a9f21174795) has been changed to prevent triggering a halting case. This logic should also be carefully reviewed.


circuit/types/group/src/helpers/from_xy_coordinates.rs
```diff
@@ -16,10 +16,14 @@ use super::*;
 
 impl<E: Environment> Group<E> {
     /// Initializes an affine group element from a given x- and y-coordinate field element.
-    /// For safety, the resulting point is always enforced to be on the curve and in the subgroup.
+    /// For safety, the resulting point is always enforced to be on the curve.
```
howardwu:
```suggestion
    /// For safety, the resulting point is always enforced to be on the curve and in the subgroup.
```

circuit/program/src/data/literal/cast_lossy/scalar.rs
```diff
@@ -0,0 +1,260 @@
...
+    #[inline]
+    fn cast_lossy(&self) -> Address<E> {
+        let field: Field<E> = self.cast_lossy();
+        field.cast_lossy()
+    }
+}
+
+impl<E: Environment> CastLossy<Boolean<E>> for Scalar<E> {
+    /// Casts a `Scalar` to a `Boolean`, with lossy truncation.
+    /// This operation returns the least significant bit of the field.
```
acoglio:
```suggestion
    /// This operation returns the least significant bit of the scalar.
```

console/algorithms/Cargo.toml
```diff
@@ -12,10 +12,15 @@ path = "benches/bhp.rs"
 harness = false
 
 [[bench]]
-name = "poseidon_sponge"
+name = "poseidon"
```
raychu86, line 15: This will require a change to the benchmarks CI, need to update the benchmark being run from `poseidon_sponge` to `poseidon`.

circuit/program/src/data/literal/cast_lossy/field.rs
```diff
@@ -0,0 +1,302 @@
...
+        let initial_x = Field::ternary(&is_x_one, &generator.to_x_coordinate(), &elligator.to_x_coordinate());
+        // Determine the initial y-coordinate, if the given field element is one.
+        let initial_y = Field::ternary(&is_x_one, &generator.to_y_coordinate(), &elligator.to_y_coordinate());
+
+        // Determine the y-coordinate, if the x-coordinate is valid.
+        let possible_y: Field<E> = {
+            // Set the x-coordinate.
+            let x = self.clone();
+            // Derive the y-coordinate.
+            witness!(|x| match console::Group::from_x_coordinate(x) {
```
acoglio, line 90: I'm not sure whether `possible_y`, as a variable in the R1CS circuit, is adequately constrained here. If I understand `witness!` correctly, the check whether `x` is the abscissa of a point on the elliptic curve is performed by this Rust code on the specific value of `x` supplied when this Rust code is run, as opposed to generating R1CS constraints that represent that check on the generic `x` variable, no matter what its value is, producing (in the circuit) a boolean saying whether `x` is on the curve or not (a "flagged `from_x_coordinate`, to use terminology from past discussions). If that's the case, the synthesized circuit will correctly represent the computation for the specific witness value of `x` for which the ZK proof is generated and verified, but not necessarily for other values of `x`; maybe that's okay because we only need the circuit to be correct for the specific ZK proof, but I thought that the synthesized circuit shouldn't depend on the witness values. (I may be well missing something here though.)

acoglio:
Modulo the comment I left on the `witness!` issue, the console and circuit computations look consistent to me:
- If `x` is the abscissa of a point on the curve, the result is that point.
- Otherwise, if `x` is 1, the result is the generator point.
- Otherwise, the result is Elligator applied to `x`.
In the circuit, this rests on the fact that no point on the curve has ordinate `y = 0`, so that `possible_y` is 0 if and only if `x` is not on the curve — is that the case?

circuit/program/src/data/literal/cast/integer.rs
```diff
@@ -0,0 +1,1568 @@
...
+                    Integer::<E, I1>::from_bits_le(&bits_le[..(I1::BITS as usize)])
+                }
+            },
+            // If the two types are both signed, perform the required checks.
+            (true, true) => match I0::BITS <= I1::BITS {
+                // If the source type is smaller than or equal to the destination type, sign extend the source integer.
+                // Then instantiate the new integer from the bits.
+                true => {
+                    bits_le
+                        .extend(std::iter::repeat(self.msb().clone()).take(I1::BITS.saturating_sub(I0::BITS) as usize));
```
ljedrz, line 110:
```suggestion
                    bits_le.resize(I1::BITS as usize, self.msb().clone());
```

circuit/program/src/data/literal/cast_lossy/integer.rs
```diff
@@ -0,0 +1,1871 @@
...
+        // If the source type is smaller than the destination type, then perform the appropriate extension.
+        let padding = match I0::is_signed() {
+            // If the source type is signed, then sign extend.
+            true => self.msb().clone(),
+            // Otherwise, zero extend.
+            false => Boolean::constant(false),
+        };
+        bits_le.extend(std::iter::repeat(padding).take(I1::BITS.saturating_sub(I0::BITS) as usize));
+        // Construct the integer from the bits.
+        Integer::from_bits_le(&bits_le[..(I1::BITS as usize)])
```
ljedrz, line 76:
```suggestion
        bits_le.resize(I1::BITS as usize, padding);
        // Construct the integer from the bits.
        Integer::from_bits_le(&bits_le)
```

d0cd (approved):
LGTM!

PR #1995, opened 2023-09-17
[Keccak] Slighly simplify a line of code.

acoglio:
No functional change.


bendyarm (commented):
The change looks good.  I suggest adding an assertion that bitrate mod 8 is 0, since that is less obvious after the change.

acoglio:
> The change looks good. I suggest adding an assertion that bitrate mod 8 is 0, since that is less obvious after the change.

Good idea, I just pushed a commit that does that.

howardwu (approved):
LGTM

PR #1996, opened 2023-09-18
[Keccak] Improvements to documentation.

acoglio:
Individual commits provide more detailed explanations.


howardwu (approved):
LGTM, I was using the original Keccak team specification, however the FIPS documentation is likely the more relevant one.

PR #1997, opened 2023-09-18
[Keccak] Tweak code for rotation offsets.

acoglio:
No functional change, but the number of loop iterations in rho is always 24, regardless of the number of rounds. In our case the number of rounds is also 24, but in my opinion it's cleaner to use `24` for the rho loop, since that's what FIPS 202 uses.

### DIFF
--- a/circuit/algorithms/src/keccak/mod.rs
+++ b/circuit/algorithms/src/keccak/mod.rs
@@ -87,7 +87,7 @@ impl<E: Environment, const TYPE: u8, const VARIANT: usize> Keccak<E, TYPE, VARIA
     pub fn new() -> Self {
         Self {
             round_constants: Self::ROUND_CONSTANTS.into_iter().map(|e| U64::constant(console::U64::new(e))).collect(),
-            rotl: Self::rotl_offsets::<NUM_ROUNDS>(),
+            rotl: Self::rotl_offsets(),
         }
     }
 }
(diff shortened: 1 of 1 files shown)


howardwu:
I would suggest keeping the usage of `NUM_ROUNDS` which itself is derived from `12 + 2 * l` where `l = 6` as defined by the standard. It is worth noting that due to `NUM_ROUNDS` being a constant, it is compiled in every time, making it equivalent to the proposed change in this PR (i.e. there would also be no performance differences with hardcoding the number).

acoglio:
The proposed change is only for the loop in `rotl_offsets`, which is always 24 iterations, no matter what the number of rounds is. It's in Section 3.2.2 of FIPS 202, where t goes from 0 to 23, as part of the definition of rho:
image
The number of rounds `n_r` comes into play in Section 3.3, in the loop in Algorithm 7, which is instantiated to `12 + 2  * l` as you say (in Section 3.4):
image
The proposed commit doesn't modify `permutation_f`, which still uses `NUM_ROUNDS`.

howardwu:
Screenshot 2023-09-19 at 03 12 13
Agreed, on secondary inspection, the Keccak team specs also hardcode 23. In this case, let's proceed with this PR.

circuit/algorithms/src/keccak/mod.rs
```diff
@@ -130,18 +130,18 @@ impl<E: Environment, const TYPE: u8, const VARIANT: usize> Keccak<E, TYPE, VARIA
...
     ///
     /// This method transposes the offsets to match the access pattern (i.e. for y, then for x).
-    fn rotl_offsets<const NUM_ROUNDS: usize>() -> [usize; MODULO * MODULO] {
+    fn rotl_offsets() -> [usize; MODULO * MODULO] {
         let mut rotl = [0; MODULO * MODULO];
         let mut x: usize = 1;
         let mut y: usize = 0;
 
-        for t in 0..NUM_ROUNDS {
+        for t in 0..24 {
```
howardwu:
```suggestion
        for t in 0..=23 {
```

howardwu: For improved clarity to the reader

howardwu (approved):

PR #1998, opened 2023-09-19
cleanup timers

vicsn:
## Motivation

Our profiling code was panicking when running a batch proof.

## Test Plan

Tried to add a simple test in our CI to catch this by enabling `--features profiler`

## Related PRs

No need for this anymore: https://github.com/AleoHQ/aleo-std/pull/9/


.circleci/config.yml
```diff
@@ -150,6 +150,10 @@ jobs:
       - run_serial:
           workspace_member: algorithms
           cache_key: snarkvm-algorithms-cache
+      - run:
+          command: |
+            cd algorithms
+            cargo test snark::varuna::tests::varuna::prove_and_verify_with_square_matrix --features profiler
```
howardwu: What is this here for?

vicsn: Added a comment and cleaned it up a bit. None of our existing tests were run using `--features profiler`. Only with that feature enabled would we have triggered the panic resolved by this PR.

howardwu (approved):

PR #2001, opened 2023-09-22
Remove unused allows and derivations in algorithms and utilities

vicsn:
## Motivation

Removing these allows and derives seemed harmless to me.

## Test Plan

The only issue I can imagine is that we fail building or clippy on some combination of feature flags - but that combination should really be added into a CI then. ljedrz and iamalwaysuncomfortable should be able to reason about features used in snarkOS and SDK respectively.

### DIFF
--- a/algorithms/src/snark/varuna/ahp/indexer/circuit_info.rs
+++ b/algorithms/src/snark/varuna/ahp/indexer/circuit_info.rs
@@ -19,8 +19,7 @@ use snarkvm_utilities::{serialize::*, ToBytes};
 /// Information about the circuit, including the field of definition, the number of
 /// variables, the number of constraints, and the maximum number of non-zero
 /// entries in any of the constraint matrices.
-#[allow(clippy::derive_partial_eq_without_eq)]
-#[derive(Copy, Clone, Debug, PartialEq, Eq, Ord, PartialOrd, CanonicalSerialize, CanonicalDeserialize)]
+#[derive(Copy, Clone, Debug, PartialEq, Eq, CanonicalSerialize, CanonicalDeserialize)]
 pub struct CircuitInfo {
     /// The number of public inputs after padding.
     pub num_public_inputs: usize,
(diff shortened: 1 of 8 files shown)


ljedrz (approved):
LGTM; as long as `clippy` doesn't complain and we don't need the extra derives, that's perfectly fine; superfluous attributes can be misleading, and removing unused derives can speed up compilation.

algorithms/src/snark/varuna/ahp/indexer/circuit_info.rs
```diff
@@ -19,8 +19,7 @@ use snarkvm_utilities::{serialize::*, ToBytes};
 /// Information about the circuit, including the field of definition, the number of
 /// variables, the number of constraints, and the maximum number of non-zero
 /// entries in any of the constraint matrices.
-#[allow(clippy::derive_partial_eq_without_eq)]
-#[derive(Copy, Clone, Debug, PartialEq, Eq, Ord, PartialOrd, CanonicalSerialize, CanonicalDeserialize)]
+#[derive(Copy, Clone, Debug, PartialEq, Eq, CanonicalSerialize, CanonicalDeserialize)]
```
iamalwaysuncomfortable, line 22: %100 the ord wasn't needed here since all the members are public and directly comparable.

algorithms/src/snark/varuna/ahp/prover/round_functions/first.rs
```diff
@@ -57,7 +57,6 @@ impl<F: PrimeField, MM: SNARKMode> AHPForR1CS<F, MM> {
     }
 
     /// Output the first round message and the next state.
-    #[allow(clippy::type_complexity)]
```
iamalwaysuncomfortable, line 60: I imagine this and a lot of other lints are artifacts from a time when we didn't have the types encapsulated as they are now. There's no "redundant clippy lint" lint, so these can pile up. Thanks for removing them!

iamalwaysuncomfortable (approved):

howardwu (approved):

PR #2002, opened 2023-09-22
Adds support for SocketAddr in ToBytes/FromBytes

howardwu:
## Motivation

Adds support for SocketAddr in ToBytes/FromBytes

## Test Plan

A unit test to check random IP v4 and v6 addresses is included.


raychu86 (approved):
lgtm

PR #2004, opened 2023-09-23
Reduce duplication of universal provers

vicsn:
## Motivation

When creating circuit keys for a batch of N circuits, we are currently creating the universal prover N times. This PR reduces it to only once.

Side note: the way synthesizer works, making one proving key at a time, when proving N transitions we still end up creating a universal_prover N+1 times (N times during key generation and once before proving). I benchmarked it and the cost is too low to justify a big refactor now, but will clean this up in the future.

## Test Plan

I removed comments in 92f4ea67e because I believe we already take `MM` into account when calling `AHPForR1CS::<_, MM>::index`, and if the SRS is not big enough, we will gently fail when we `download_powers_for`.

### DIFF
--- a/algorithms/src/snark/varuna/varuna.rs
+++ b/algorithms/src/snark/varuna/varuna.rs
@@ -15,7 +15,10 @@
 use super::Certificate;
 use crate::{
     fft::EvaluationDomain,
-    polycommit::sonic_pc::{Commitment, Evaluations, LabeledCommitment, QuerySet, Randomness, SonicKZG10},
+    polycommit::{
+        kzg10::DegreeInfo,
+        sonic_pc::{Commitment, Evaluations, LabeledCommitment, QuerySet, Randomness, SonicKZG10},
+    },
     r1cs::{ConstraintSynthesizer, SynthesisError},
     snark::varuna::{
         ahp::{AHPError, AHPForR1CS, CircuitId, EvaluationsProvider},
@@ -38,7 +41,7 @@ use snarkvm_curves::PairingEngine;
 use snarkvm_fields::{One, PrimeField, ToConstraintField, Zero};
 use snarkvm_utilities::{to_bytes_le, ToBytes};
 
-use anyhow::{anyhow, Result};
+use anyhow::{anyhow, ensure, Result};
 use core::marker::PhantomData;
 use itertools::Itertools;
 use rand::{CryptoRng, Rng};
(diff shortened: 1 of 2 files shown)


algorithms/src/snark/varuna/varuna.rs
```diff
@@ -66,25 +69,35 @@ impl<E: PairingEngine, FS: AlgebraicSponge<E::Fq, 2>, MM: SNARKMode> VarunaSNARK
         universal_srs: &UniversalSRS<E>,
         circuits: &[&C],
     ) -> Result<Vec<(CircuitProvingKey<E, MM>, CircuitVerifyingKey<E>)>> {
+        ensure!(!circuits.is_empty());
         let index_time = start_timer!(|| "Varuna::CircuitSetup");
 
-        let mut circuit_keys = Vec::with_capacity(circuits.len());
+        let mut indexed_circuits = Vec::with_capacity(circuits.len());
+        let mut degree_info = None::<DegreeInfo>;
```
Pratyush, line 76:
```suggestion
        let mut degree_info = None;
```

Pratyush (approved):

algorithms/src/snark/varuna/varuna.rs
```diff
@@ -66,25 +69,35 @@ impl<E: PairingEngine, FS: AlgebraicSponge<E::Fq, 2>, MM: SNARKMode> VarunaSNARK
...
-                .map_err(|e| anyhow!("Failed to download powers for degree {}: {e}", degree_info.max_degree))?;
+            let indexed_circuit = AHPForR1CS::<_, MM>::index(*circuit)?;
+            let degree_info_i = indexed_circuit.index_info.degree_info::<E::Fr, MM>();
+            degree_info = match degree_info {
+                Some(degree_info) => Some(degree_info.union(&degree_info_i)),
+                None => Some(degree_info_i),
+            };
+            indexed_circuits.push(indexed_circuit);
+        }
+        let degree_info = degree_info.unwrap();
```
iamalwaysuncomfortable, line 86: I'd thought about proposing this change earlier. This subtle change will be a pretty big improvement to key parameter download i/o & memory allocation for powers in the case of multiple circuits. In a future PR We might consider making  lines 75-->86 a separate function so that callers estimate the powers they need to download for their circuit batches in case they need a separate download strategy (i.e. wasm)

iamalwaysuncomfortable (approved):

howardwu (approved):

PR #2008, opened 2023-09-24
Add `unconfirmed_transaction` helper method for `ConfirmedTransaction`s

raychu86:
## Motivation

This PR adds a helper method to derive the `ConfirmedTransaction`'s original transaction before it was confirmed.

In the case of an accepted deploy or execute, the transaction remains the same.
In the case of a rejected deploy or execute, we regenerate the original transaction by rebuilding the deployment or execution.


howardwu (approved):
LGTM

ntbinh1204:
flg

PR #2013, opened 2023-09-25
batch inverse for selectors at challenge

### LINKED ISSUE #1927: This performs one inversion per circuit; it might be worthwhile to consider a `batch_selector_evals` method which performs only a single inversion for the entire batch.
This performs one inversion per circuit; it might be worthwhile to consider a `batch_selector_evals` method which performs only a single inversion for the entire batch.

_Originally posted by @Pratyush in https://github.com/AleoHQ/snarkVM/pull/1701#discussion_r1297513433_

vicsn:
## Motivation

Uses batch inversion to more efficiently compute all selectors at the query point.

When batching e.g. 8 very small circuits (10k constraints), prover time reduces to a whopping 50%. At 8 circuits of 100k constraints, more akin to our transfer circuit, the speedup is ~5%.

Closes https://github.com/AleoHQ/snarkVM/issues/1927


algorithms/src/snark/varuna/ahp/selectors.rs
```diff
@@ -0,0 +1,113 @@
...
+    fn evaluate_selector_polynomial(&self, other: EvaluationDomain<F>, point: F) -> F;
+}
+
+impl<F: PrimeField> SelectorPolynomial<F> for EvaluationDomain<F> {
+    fn evaluate_selector_polynomial(&self, other: EvaluationDomain<F>, point: F) -> F {
+        let numerator = self.evaluate_vanishing_polynomial(point) * other.size_as_field_element;
+        let denominator = other.evaluate_vanishing_polynomial(point) * self.size_as_field_element;
+        numerator / denominator
+    }
+}
```
Pratyush: I'd recommend moving this to `EvaluationDomain`, so that we don't need a separate trait for it.

Pratyush (approved):
LGTM modulo nit.

algorithms/src/snark/varuna/ahp/selectors.rs
vicsn: Good idea. Actually its not used anymore so I moved it into the test, also moved the other free selector function here: 7bd8afc1f

Pratyush (approved):

PR #2014, opened 2023-09-25
Rename MM to SM, from MarlinMode to SnarkMode

vicsn:
## Motivation

Renaming as it might confuse future devs. This is a rename and does not change any logic.


howardwu (approved):
LGTM, love it. Thanks @vicsn!

PR #2015, opened 2023-09-26
[Refactor] Remove identifier from `MapKey` and `MapValue`.

d0cd:
This PR removes identifiers from `MapKey` and `MapValue` as they are not necessary.
Users can now declare mappings in the following way:
```aleo
mapping balances:
    key as address.public;
    value as u64.public;
```

### DIFF
--- a/synthesizer/program/src/mapping/key/mod.rs
+++ b/synthesizer/program/src/mapping/key/mod.rs
@@ -18,7 +18,7 @@ mod parse;
 use console::{network::prelude::*, program::PlaintextType};
 
 /// A key statement is of the form `key as {plaintext_type}.public`.
-#[derive(Clone, PartialEq, Eq, Hash)]
+#[derive(Clone, PartialEq, Eq)]
 pub struct MapKey<N: Network> {
     /// The key plaintext type.
     plaintext_type: PlaintextType<N>,
--- a/synthesizer/program/src/mapping/key/mod.rs
+++ b/synthesizer/program/src/mapping/key/mod.rs
@@ -15,27 +15,16 @@
 mod bytes;
 mod parse;
 
-use console::{
-    network::prelude::*,
-    program::{Identifier, PlaintextType},
-};
+use console::{network::prelude::*, program::PlaintextType};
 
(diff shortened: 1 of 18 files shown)


synthesizer/program/src/mapping/key/mod.rs
```diff
@@ -15,27 +15,16 @@
...
 
-use console::{
-    network::prelude::*,
-    program::{Identifier, PlaintextType},
-};
+use console::{network::prelude::*, program::PlaintextType};
 
-/// A key statement is of the form `key {name} as {plaintext_type}.public`.
+/// A key statement is of the form `key as {plaintext_type}.public`.
 #[derive(Clone, PartialEq, Eq, Hash)]
```
ljedrz: does `Hash` still hold without the `Identifier`? I'm not too familiar with the related objects, but the name `PlaintextType` suggests to me that they might be some overlap between them?

synthesizer/program/src/mapping/value/mod.rs
```diff
@@ -15,27 +15,16 @@
...
 
-use console::{
-    network::prelude::*,
-    program::{Identifier, PlaintextType},
-};
+use console::{network::prelude::*, program::PlaintextType};
 
-/// An value statement is of the form `value {name} as {plaintext_type}.public;`.
+/// An value statement is of the form `value as {plaintext_type}.public;`.
 #[derive(Clone, PartialEq, Eq, Hash)]
```
ljedrz: ditto

ljedrz (commented):
I'm not very familiar with this particular code, but I left some comments related to `Hash` impls.

acoglio (approved):

synthesizer/program/src/mapping/key/mod.rs
d0cd: It turns out we actually don't rely on it at all.

synthesizer/program/src/mapping/value/mod.rs
d0cd: As above.

howardwu (approved):
LGTM

PR #2017, opened 2023-09-26
[Aleo ins parser] Fix array type anomaly.

### LINKED ISSUE #2016: [Bug] Anomaly in array type parsing
Maybe not a (serious) bug, but at least anomalous behavior.

This works
```
cast r5 into r7 as /* comment */ [u8; 10u32];
```
while this does not work
```
cast r5 into r7 as /* comment */ u8;
```

The culprit is [this line](https://github.com/AleoHQ/snarkVM/blob/326e9d59befa0ccf02b45ecbdd6c3ebcf34e1f2c/console/program/src/data_types/array_type/parse.rs#L46), which parses comments and whitespace at the beginning of an array type. Array types are one of three kind of plaintext types, the other two being literal types and identifiers: neither the [parser of literal types](https://github.com/AleoHQ/snarkVM/blob/326e9d59befa0ccf02b45ecbdd6c3ebcf34e1f2c/console/program/src/data_types/literal_type/parse.rs#L20) nor the [parser for identifiers](https://github.com/AleoHQ/snarkVM/blob/326e9d59befa0ccf02b45ecbdd6c3ebcf34e1f2c/console/program/src/data/identifier/parse.rs#L27) parse comment and whitespace at the beginning; their callers parse whitespace (not comments). So the proposal is to remove the culprit line, and treat array types the same as the other plaintext types when parsing.

acoglio:
Fixes #2016.


d0cd (approved):
LGTM!

console/program/src/data_types/array_type/parse.rs
```diff
@@ -42,8 +42,6 @@ impl<N: Network> Parser for ArrayType<N> {
             Ok((string, length))
         }
 
-        // Parse the whitespace and comments from the string.
-        let (string, _) = Sanitizer::parse(string)?;
```
howardwu, line 46: I think this parse is justified. At the least we should have `parse_whitespaces`.

acoglio: This array type parser is called by the [plaintext type parser](https://github.com/AleoHQ/snarkVM/blob/6a81f8173b5627421dce72b9b04002ccf70f45a2/console/program/src/data_types/plaintext_type/parse.rs#L20), which expects whitespace to have been skipped over already: the other two alternative parsers for plaintext type, namely [literal type](https://github.com/AleoHQ/snarkVM/blob/6a81f8173b5627421dce72b9b04002ccf70f45a2/console/program/src/data_types/literal_type/parse.rs#L20) and [identifier](https://github.com/AleoHQ/snarkVM/blob/6a81f8173b5627421dce72b9b04002ccf70f45a2/console/program/src/data/identifier/parse.rs#L25), do not skip over whitespace.

This is an instance of a general pattern in the Aleo instructions parser:
1. Some (larger) constructs are parsed by first skipping over white space (and sometimes also comments).
2. Other (smaller) constructs are parsed without any such initial skipping, i.e. they are expected to start right there. Their callers skip white space already.

Plaintext types, including array types, belong to the second group of constructs above.

howardwu: I see your perspective, we should make sure its parent callers have some notion of `parse_whitespace` since we don't invoke `string.trim()` before calling the parser.

acoglio: Yes, definitely. I believe that to be the case. I don't have a formal proof yet, but here's an argument. As you may know, I've derived the [ABNF grammar of Aleo instructions](https://github.com/AleoHQ/grammars/blob/master/aleo.abnf) directly from the parser. So under the (good) assumption that the grammar is consistent with the parser, we can see the following in the grammar:
- Every occurrence of `plaintext-type` not at the start of the right side of a rule is preceded by `ws`, which corresponds to `Sanitizer::parse_whitespaces`.
- `plainttext-type` occurs at the start of the right sides of `value-type`, `mapping-type`, `finalize-type`, and `entry-type`.
- We repeat the same argument for `value-type`, ..., `entry-type`, recursively, and find that there is always a preceding `ws`.

(That was actually a sketch of a formal proof.)

PR #2019, opened 2023-09-26
Enforce subdag anchor rounds are even

raychu86:
## Motivation

This PR changes the subdag anchor round enforcement to even instead of odd.


howardwu (approved):
LGTM

PR #2022, opened 2023-09-27
Speed up OutputStorage::get

ljedrz:
This PR is opened against `feat/futures-2`.

While this logic doesn't allow us to ensure that there aren't multiple outputs, it should be much faster on average.


howardwu (approved):

howardwu (approved):

raychu86 (approved):
LGTM

PR #2025, opened 2023-09-27
[Aleo ins parser] Minor code and doc fixes.

acoglio:
This is for futures.

Update some doc comments.

The code fix is for parsing whitespace just before commas of future arguments. The current code allows comments and new lines there, which I don't think we want (not wrong per se, but anomalous, inconsistent with other parts of the syntax). So the fix restricts that to just whitespace (no comments) without new lines. We still allow comments and new lines after the comma, consistently with other parts of the syntax.

### DIFF
--- a/console/program/src/data/future/parse.rs
+++ b/console/program/src/data/future/parse.rs
@@ -28,7 +28,7 @@ impl<N: Network> Parser for Future<N> {
             let (string, _) = Sanitizer::parse(string)?;
             // Parse the members.
             let (string, arguments) = separated_list0(
-                pair(pair(Sanitizer::parse, tag(",")), Sanitizer::parse),
+                pair(pair(Sanitizer::parse_whitespaces, tag(",")), Sanitizer::parse),
                 alt((map(Future::parse, Argument::Future), map(Plaintext::parse, Argument::Plaintext))),
             )(string)?;
             // Parse the whitespace and comments from the string.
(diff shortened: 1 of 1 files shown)


howardwu (approved):
LGTM

console/program/src/data/future/parse.rs
```diff
@@ -28,7 +28,7 @@ impl<N: Network> Parser for Future<N> {
             let (string, _) = Sanitizer::parse(string)?;
             // Parse the members.
             let (string, arguments) = separated_list0(
-                pair(pair(Sanitizer::parse, tag(",")), Sanitizer::parse),
+                pair(pair(Sanitizer::parse_whitespaces, tag(",")), Sanitizer::parse),
```
howardwu, line 31: Actually, I would push back on this. People sometimes will comment/document what each member represents.

acoglio: I don't disagree. In fact, I think we should allow comments and new lines everywhere. But right now the Aleo instructions parser enforces certain whitespace and comment restrictions by selectively using `Sanitizer::parse` (which skips over comments and whitespace including new lines) and `Sanitizer::parse_whitespace` (which skips over just whitespace excluding new lines). So this PR is making this part of the parser more consistent with other parts. Note that here one can still put comments just after the comma:
```
member, /* comment */
member, // comment
```
This PR just disallows
```
member /*comment*/ ,
```
or
```
member
,
```

howardwu (approved):
LGTM

PR #2026, opened 2023-09-28
[Refactor] Rename caller to signer, then parent to caller.

d0cd:
This PR renames `caller` to `signer` and then `parent` to `caller` in `Request` and `Operand`.
Note that `self.signer` is the user that signed the original request, while `self.caller` is the address (user or program) that invoked the current request.
Note that existing programs that use `self.caller` are backed into the composability offered by `self.caller`.
If this is not the intended behavior, those programs would need to be adjusted to use `self.signer`.

### DIFF
--- a/synthesizer/process/src/stack/evaluate.rs
+++ b/synthesizer/process/src/stack/evaluate.rs
@@ -25,8 +25,8 @@ impl<N: Network> StackEvaluate<N> for Stack<N> {
         closure: &Closure<N>,
         inputs: &[Value<N>],
         call_stack: CallStack<N>,
+        signer: Address<N>,
         caller: Address<N>,
-        parent: Address<N>,
         tvk: Field<N>,
     ) -> Result<Vec<Value<N>>> {
         let timer = timer!("Stack::evaluate_closure");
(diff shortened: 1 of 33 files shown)


synthesizer/process/src/stack/evaluate.rs
```diff
@@ -76,10 +76,10 @@ impl<N: Network> StackEvaluate<N> for Stack<N> {
                     Operand::ProgramID(program_id) => {
                         Ok(Value::Plaintext(Plaintext::from(Literal::Address(program_id.to_address()?))))
                     }
-                    // If the operand is the caller, retrieve the caller from the registers.
+                    // If the operand is the signer, retrieve the caller from the registers.
```
howardwu:
```suggestion
                    // If the operand is the signer, retrieve the signer from the registers.
```

synthesizer/process/src/stack/evaluate.rs
```diff
@@ -76,10 +76,10 @@ impl<N: Network> StackEvaluate<N> for Stack<N> {
...
+                    Operand::Signer => Ok(Value::Plaintext(Plaintext::from(Literal::Address(registers.signer()?)))),
+                    // If the operand is the caller, retrieve the parent from the registers.
```
howardwu:
```suggestion
                    // If the operand is the caller, retrieve the caller from the registers.
```

PR #2027, opened 2023-09-28
Introduce subdag root to header

raychu86:
## Motivation

This PR adds the subdag root to the block header. If the block authority is quorum, then the `subdag_root` is the certificate id of the subdag leader certificate. If the block authority is beacon, then the `subdag_root` is `Field::zero()`.


howardwu (approved):
LGTM

PR #2032, opened 2023-09-28
[NCC-ADN] Prevent ignoring proving tasks

vicsn:
## Motivation

Cleanup based on NCC's finding, we should make sure that we don't silently ignore proving tasks.

Notes:
1. We were needlessly cloning the `Assignments` so I cleaned that up. Just for context: in `synthesizer` crate, we are passing the `proving_key` and `assignments` by value, while in the `algorithms` crate we are passing them by reference.
2. ~If two transitions with different locators compile to the same circuit, the previous design ignored them~ We inject program_id and function_name so this is impossible.
3. Incorporated the requirement that `prove_batch(...)` can be called multiple times on a `Trace`.

## Test Plan

Added a test which calls the same function twice. Apparently before this PR, we were already proving both executions. So this PR is just defense in depth.


synthesizer/process/src/trace/mod.rs
```diff
@@ -87,10 +84,13 @@ impl<N: Network> Trace<N> {
...
-        let locator = Locator::new(*transition.program_id(), *transition.function_name());
-        // Insert the assignment (and proving key if the entry does not exist), for the specified locator.
-        self.transition_tasks.entry(locator).or_insert((proving_key, vec![])).1.push(assignment);
+        // Insert the assignment (and proving key if the entry does not exist)
+        // We pass assignment by value because we are later also passing short-lived inclusion_assignment by value
+        if let Some(entry) = self.transition_tasks.get_mut(&proving_key) {
+            entry.push(assignment);
+        } else {
+            self.transition_tasks.insert(proving_key, vec![assignment]);
+        }
```
howardwu:
```suggestion
        self.transition_tasks.entry(proving_key).or_insert(vec![]).push(assignment);
```

synthesizer/process/src/trace/mod.rs
```diff
@@ -283,14 +284,17 @@ impl<N: Network> Trace<N> {
         }
 
         if !batch_inclusions.is_empty() {
-            // Fetch the inclusion proving key.
-            let proving_key = ProvingKey::<N>::new(N::inclusion_proving_key().clone());
+            let inclusion_proving_key = ProvingKey::<N>::new(N::inclusion_proving_key().clone());
```
howardwu:
```suggestion
            // Fetch the inclusion proving key.
            let inclusion_proving_key = ProvingKey::<N>::new(N::inclusion_proving_key().clone());
```

synthesizer/process/src/trace/mod.rs
```diff
@@ -283,14 +284,17 @@ impl<N: Network> Trace<N> {
...
+
             // Insert the inclusion proving key and assignments.
-            proving_tasks.push((proving_key, batch_inclusions));
+            if proving_tasks.insert(inclusion_proving_key, batch_inclusions).is_some() {
+                return Err(anyhow!("proving_key was already present"));
```
howardwu:
```suggestion
                bail!("The inclusion proving key (and its instances) have already been inserted");
```

synthesizer/process/src/trace/mod.rs
```diff
@@ -193,12 +196,10 @@ impl<N: Network> Trace<N> {
             self.global_state_root.get().ok_or_else(|| anyhow!("Global state root has not been set"))?;
         // Retrieve the fee transition.
         let fee_transition = &self.transitions[0];
-        // Construct the proving tasks.
-        let proving_tasks = self.transition_tasks.values().cloned().collect();
         // Compute the proof.
         let (global_state_root, proof) = Self::prove_batch::<A, R>(
             "credits.aleo/fee (private or public)",
-            proving_tasks,
+            &mut self.transition_tasks,
```
howardwu: Please revert this. We should not be mutating the proving_tasks. What if they call this twice?

synthesizer/process/src/trace/mod.rs
```diff
@@ -165,17 +165,20 @@ impl<N: Network> Trace<N> {
...
         let global_state_root =
             self.global_state_root.get().ok_or_else(|| anyhow!("Global state root has not been set"))?;
-        // Construct the proving tasks.
-        let proving_tasks = self.transition_tasks.values().cloned().collect();
         // Compute the proof.
-        let (global_state_root, proof) =
-            Self::prove_batch::<A, R>(locator, proving_tasks, inclusion_assignments, *global_state_root, rng)?;
+        let (global_state_root, proof) = Self::prove_batch::<A, R>(
+            locator,
+            &mut self.transition_tasks,
```
howardwu: Please revert, see other comment.

synthesizer/process/src/trace/mod.rs
vicsn: 8f1d96463

vicsn:
Incorporated the requirement that `prove_batch(...)` can be called multiple times on a `Trace`. I'll still pass `self.proving_tasks` by reference instead of creating a new struct like we did before, because to be extra clear: I'm now directly storing the the `BTreeMap<ProvingKey, Vec<Assignment>>` in `Trace`. 964db054d

synthesizer/snark/src/proving_key/mod.rs
```diff
@@ -20,7 +20,7 @@ mod serialize;
 
 use std::collections::BTreeMap;
 
-#[derive(Clone)]
+#[derive(Clone, PartialEq, Eq)]
```
Pratyush: We have to be a little careful here, as well as in `CircuitProvingKey`; right now the `Eq` and `PartialEq` impls on `CircuitProvingKey` compare the *entire* struct, whereas the corresponding `Ord` impls compare only the circuit IDs. The latter is obviously much more efficient, and if the circuit ID is computed honestly, the two checks are essentially equivalent. However, we should be careful about not accidentally invoking equality checks on the entire struct, as that would be very expensive.

synthesizer/process/src/trace/mod.rs
```diff
@@ -87,10 +84,9 @@ impl<N: Network> Trace<N> {
         // Insert the transition into the inclusion tasks.
         self.inclusion_tasks.insert_transition(input_ids, transition)?;
 
-        // Construct the locator.
-        let locator = Locator::new(*transition.program_id(), *transition.function_name());
-        // Insert the assignment (and proving key if the entry does not exist), for the specified locator.
-        self.transition_tasks.entry(locator).or_insert((proving_key, vec![])).1.push(assignment);
+        // Insert the assignment (and proving key if the entry does not exist)
+        // We pass assignment by value because we are later also passing short-lived inclusion_assignment by value
+        self.transition_tasks.entry(proving_key).or_insert(vec![]).push(assignment);
```
ljedrz:
```suggestion
        self.transition_tasks.entry(proving_key).or_default().push(assignment);
```

vicsn: a58ab117e

synthesizer/snark/src/proving_key/mod.rs
vicsn: Good point, made it less ambiguous: b6010842a

synthesizer/snark/src/proving_key/mod.rs
```diff
@@ -91,3 +91,21 @@ impl<N: Network> Deref for ProvingKey<N> {
         &self.proving_key
     }
 }
+
+impl<N: Network> PartialEq for ProvingKey<N> {
+    fn eq(&self, other: &Self) -> bool {
+        self.deref() == other.deref()
+    }
+}
```
ljedrz, line 102: this is how `PartialEq` is implemented for `Arc<T>`; did the compiler complain when you tried to just derive it?

synthesizer/snark/src/proving_key/mod.rs
```diff
@@ -91,3 +91,21 @@ impl<N: Network> Deref for ProvingKey<N> {
...
+
+impl<N: Network> Ord for ProvingKey<N> {
+    fn cmp(&self, other: &Self) -> Ordering {
+        self.deref().cmp(other.deref())
```
ljedrz, line 106: ditto

vicsn: Perhaps I could keep one of the automatic derivation, but when I remove some of them:
```
error[E0277]: the trait bound `N: Ord` is not satisfied
    --> synthesizer/process/src/trace/mod.rs:89:31
     |
89   |         self.transition_tasks.entry(proving_key).or_default().push(assignment);
     |                               ^^^^^ the trait `Ord` is not implemented for `N`
     |
     = note: required for `ProvingKey<N>` to implement `Ord`
```

ljedrz: ugh, that's just typical :smile:

synthesizer/process/src/trace/mod.rs
```diff
@@ -87,10 +84,9 @@ impl<N: Network> Trace<N> {
         // Insert the transition into the inclusion tasks.
         self.inclusion_tasks.insert_transition(input_ids, transition)?;
 
-        // Construct the locator.
-        let locator = Locator::new(*transition.program_id(), *transition.function_name());
-        // Insert the assignment (and proving key if the entry does not exist), for the specified locator.
-        self.transition_tasks.entry(locator).or_insert((proving_key, vec![])).1.push(assignment);
+        // Insert the assignment (and proving key if the entry does not exist)
+        // We pass assignment by value because we are later also passing short-lived inclusion_assignment by value
```
howardwu:
```suggestion
        // Insert the assignment (and proving key if the entry does not exist).
```

synthesizer/process/src/trace/mod.rs
```diff
@@ -148,7 +143,7 @@ impl<N: Network> Trace<N> {
 
     /// Returns a new execution with a proof, for the current inclusion assignments and global state root.
     pub fn prove_execution<A: circuit::Aleo<Network = N>, R: Rng + CryptoRng>(
-        &self,
+        &mut self,
```
howardwu:
```suggestion
        &self,
```

synthesizer/process/src/trace/mod.rs
```diff
@@ -165,17 +160,15 @@ impl<N: Network> Trace<N> {
...
         let (global_state_root, proof) =
-            Self::prove_batch::<A, R>(locator, proving_tasks, inclusion_assignments, *global_state_root, rng)?;
+            Self::prove_batch::<A, R>(locator, &self.transition_tasks, inclusion_assignments, *global_state_root, rng)?;
         // Return the execution.
         Execution::from(self.transitions.iter().cloned(), global_state_root, Some(proof))
     }
 
     /// Returns a new fee with a proof, for the current inclusion assignment and global state root.
-    pub fn prove_fee<A: circuit::Aleo<Network = N>, R: Rng + CryptoRng>(&self, rng: &mut R) -> Result<Fee<N>> {
+    pub fn prove_fee<A: circuit::Aleo<Network = N>, R: Rng + CryptoRng>(&mut self, rng: &mut R) -> Result<Fee<N>> {
```
howardwu:
```suggestion
    pub fn prove_fee<A: circuit::Aleo<Network = N>, R: Rng + CryptoRng>(&self, rng: &mut R) -> Result<Fee<N>> {
```

synthesizer/process/src/trace/mod.rs
vicsn: Hmpf good catch 55e032331

synthesizer/process/src/trace/mod.rs
vicsn: Hmpf good catch 55e032331

synthesizer/process/src/trace/mod.rs
```diff
@@ -282,15 +276,19 @@ impl<N: Network> Trace<N> {
             batch_inclusions.push(assignment.to_circuit_assignment::<A>()?);
         }
 
+        // Fetch the inclusion proving key.
+        let inclusion_proving_key = ProvingKey::<N>::new(N::inclusion_proving_key().clone());
+
```
howardwu, line 278:
```suggestion
        // Fetch the inclusion proving key.
        let inclusion_proving_key = ProvingKey::<N>::new(N::inclusion_proving_key().clone());

```

This should be moved back into the `if` statement. Otherwise, we are needlessly loading this proving key.

synthesizer/process/src/tests/test_execute.rs
```diff
@@ -2477,3 +2477,108 @@ function {function_name}:
...
+        assert_eq!(transition.program_id(), *expected_program_id);
+        assert_eq!(transition.function_name(), expected_function_name);
+    }
+
+    // Initialize a new block store.
+    let block_store = BlockStore::<CurrentNetwork, BlockMemory<_>>::open(None).unwrap();
+    // Prepare the trace.
+    trace.prepare(Query::from(block_store)).unwrap();
+    // Prove the execution.
+    let execution = trace.prove_execution::<CurrentAleo, _>("one", rng).unwrap();
```
howardwu, line 2586: Where is the duplicate function call?

algorithms/src/snark/varuna/data_structures/circuit_proving_key.rs
```diff
@@ -27,7 +27,7 @@ use snarkvm_utilities::{
 use std::{cmp::Ordering, sync::Arc};
 
 /// Proving key for a specific circuit (i.e., R1CS matrices).
-#[derive(Clone, Debug, PartialEq, Eq)]
+#[derive(Clone, Debug, Eq)]
```
howardwu: I'd recommend adding:
```
impl<E: PairingEngine, SM: SNARKMode> Eq for CircuitProvingKey<E, SM> {}
```
next to `PartialEq` for better readability. Otherwise the reader may see only the `PartialEq` impl and not realize `Eq` is also impl'ed.

synthesizer/snark/src/proving_key/mod.rs
```diff
@@ -20,7 +20,7 @@ mod serialize;
 
 use std::collections::BTreeMap;
 
-#[derive(Clone)]
+#[derive(Clone, Eq)]
```
howardwu: I'd recommend adding:
```
impl<N: Network> Eq for ProvingKey<N> {}
```
next to `PartialEq` for better readability. Otherwise the reader may see only the `PartialEq` impl and not realize `Eq` is also impl'ed.

synthesizer/snark/src/proving_key/mod.rs
```diff
@@ -58,17 +58,17 @@ impl<N: Network> ProvingKey<N> {
     #[allow(clippy::type_complexity)]
     pub fn prove_batch<R: Rng + CryptoRng>(
         locator: &str,
-        assignments: &[(ProvingKey<N>, Vec<circuit::Assignment<N::Field>>)],
+        assignments: BTreeMap<&ProvingKey<N>, &[circuit::Assignment<N::Field>]>,
```
howardwu: I would suggest changing this into an `impl Iterator` so the caller's function does not need to implement `collect` anymore.

algorithms/src/snark/varuna/data_structures/circuit_proving_key.rs
```diff
@@ -58,6 +58,12 @@ impl<E: PairingEngine, SM: SNARKMode> FromBytes for CircuitProvingKey<E, SM> {
     }
 }
 
+impl<E: PairingEngine, SM: SNARKMode> PartialEq for CircuitProvingKey<E, SM> {
+    fn eq(&self, other: &Self) -> bool {
+        self.circuit.id == other.circuit.id
```
howardwu: I'm not convinced this is safe.

There is no check in the `FromBytes` impl that this is computed correctly, and corresponds to the objects involved.

I would suggest auto-deriving `PartialEq, Eq` on the entire struct.

synthesizer/snark/src/proving_key/mod.rs
```diff
@@ -91,3 +91,21 @@ impl<N: Network> Deref for ProvingKey<N> {
...
+    fn cmp(&self, other: &Self) -> Ordering {
+        self.deref().cmp(other.deref())
+    }
+}
+
+impl<N: Network> PartialOrd for ProvingKey<N> {
+    fn partial_cmp(&self, other: &Self) -> Option<Ordering> {
+        Some(self.cmp(other))
+    }
+}
```
howardwu, line 114: Why can't these be auto-derived?

vicsn: error[E0277]: the trait bound `N: Ord` is not satisfied
    --> synthesizer/process/src/trace/mod.rs:89:31
     |
89   |         self.transition_tasks.entry(proving_key).or_default().push(assignment);
     |                               ^^^^^ the trait `Ord` is not implemented for `N`
     |
     = note: required for `ProvingKey<N>` to implement `Ord`

howardwu: That's very odd, I'll give it a look tomorrow as well.

synthesizer/process/src/trace/mod.rs
vicsn: I'm defining it here to make sure it has the appropriate lifetime:
- we are inserting it in line 284 by **reference**.
- Which I'm doing because the proving keys are inserted by reference in line 265.
- Which I'm doing because `&self.transition_tasks` is directly passed by reference in line 165. 

Before this PR we made copies of the `transition_tasks`, which was fine for the provingkeys (which were Arcs) but not fine for the Assignments. Maybe there is a slightly cleaner way but it does not seem worth the time to refactor.

synthesizer/process/src/tests/test_execute.rs
vicsn: The intention was above at `call zero.aleo/b r0 r1 into r2;`

algorithms/src/snark/varuna/data_structures/circuit_proving_key.rs
Pratyush: This would be very expensive. In the `FromBytes` impl I believe we reconstruct the `id` by hashing.

synthesizer/snark/src/proving_key/mod.rs
vicsn: It's not the prettiest, but this removes a collection: 07055c3d3

synthesizer/snark/src/proving_key/mod.rs
vicsn: [bc9fc63](https://github.com/AleoHQ/snarkVM/pull/2032/commits/bc9fc63321c1ce3b8649c314043657a7c2f3cb44)

algorithms/src/snark/varuna/data_structures/circuit_proving_key.rs
vicsn: 1bc011e53

algorithms/src/snark/varuna/data_structures/circuit_proving_key.rs
vicsn: > I'm not convinced this is safe.

Not sure what the attack vector is. I'm not aware of scenario's where a user should accept a CircuitProvingKey from a third party.

But even if they did, as Pratyush mentions, the user computes the `circuit_id` themselves which the user and verifiers should always check against. In theory the third party could put bogus precomputed values (index polys, (i)fft_precomputation) in the proving key to make the proof fail to be generated - not to accidentally prove a different statement.

synthesizer/snark/src/proving_key/mod.rs
vicsn: I think this would do the job: https://github.com/AleoHQ/snarkVM/pull/2231

howardwu:
I will review this after it's clear this PR is ready for review.

vicsn:
Closing in favour of: https://github.com/AleoHQ/snarkVM/pull/2276

PR #2036, opened 2023-10-02
Avoid single element allocations, explicitly preallocate some vecs

ljedrz:
This PR consists of 2 commits: the first one avoids single-element vector allocations, and the other fully preallocates for the expected length of some those vectors.

### DIFF
--- a/algorithms/src/snark/varuna/varuna.rs
+++ b/algorithms/src/snark/varuna/varuna.rs
@@ -676,7 +676,8 @@ where
                     .iter()
                     .map(|input| {
                         let input = input.borrow().to_field_elements().unwrap();
-                        let mut new_input = vec![E::Fr::one()];
+                        let mut new_input = Vec::new();
+                        new_input.push(E::Fr::one());
                         new_input.extend_from_slice(&input);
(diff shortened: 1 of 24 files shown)


howardwu (approved):
LGTM

algorithms/src/snark/varuna/varuna.rs
```diff
@@ -676,7 +676,8 @@ where
                     .iter()
                     .map(|input| {
                         let input = input.borrow().to_field_elements().unwrap();
-                        let mut new_input = vec![E::Fr::one()];
+                        let mut new_input = Vec::with_capacity(input.len().max(input_domain.size()));
```
zvolin, line 679: as this tries to skip the allocs, it should allocate for the `1 + input.len()`

howardwu: Can you open a new PR for this?

ljedrz: > as this tries to skip the allocs, it should allocate for the 1 + input.len()

actually, this should be `(1 + input.len()).max(input_domain.size())`, as there's no telling whether resize will increase or decrease the length

PR #2038, opened 2023-10-02
Avoid some more vec allocations

ljedrz:
The first commit is quite trival, as it removes one redundant allocation and adds a full preallocation.

The other commit implements `from_bits_be` for `BigInteger256` and `BigInteger384` without relying on the `from_bits_le` impl, by doing basically the same thing, but in reverse (`chunks + iter + rev` > `rchunks + iter`).

Both were found via heap profiling, with especially the second one having a sizable impact.

### DIFF
--- a/algorithms/benches/crypto_hash/poseidon.rs
+++ b/algorithms/benches/crypto_hash/poseidon.rs
@@ -16,33 +16,43 @@
 extern crate criterion;
 
 use snarkvm_algorithms::{crypto_hash::PoseidonSponge, AlgebraicSponge};
-use snarkvm_curves::bls12_377::Fq;
+use snarkvm_curves::bls12_377::{Fq, FqParameters};
+use snarkvm_fields::Fp384;
 use snarkvm_utilities::{TestRng, Uniform};
 
 use criterion::Criterion;
+use std::time::Duration;
 
-fn sponge_2_1_absorb_4(c: &mut Criterion) {
+fn sponge_2_1_absorb_100_native(c: &mut Criterion) {
     let rng = &mut TestRng::default();
     let mut sponge = PoseidonSponge::<Fq, 2, 1>::new();
 
-    let input = vec![Fq::rand(rng), Fq::rand(rng), Fq::rand(rng), Fq::rand(rng)];
-    c.bench_function("PoseidonSponge<2, 1> Absorb 4", move |b| b.iter(|| sponge.absorb_native_field_elements(&input)));
+    let mut input = Vec::with_capacity(100);
+    for _ in 0..100 {
+        input.push(Fq::rand(rng));
+    }
+    c.bench_function("PoseidonSponge<2, 1> Absorb 100 native", move |b| {
+        b.iter(|| sponge.absorb_native_field_elements(&input))
+    });
 }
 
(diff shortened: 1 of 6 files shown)


ljedrz:
Added 2 extra commits with small changes; their messages are self-explanatory.

ljedrz:
Hmm I can reproduce the `console-program` failure in `testnet3`, so it's unrelated to this PR; filed as https://github.com/AleoHQ/snarkVM/issues/2043.

ljedrz:
Pushed 2 more commits that reduce the overall number of allocations (registered during the runtime of a typical snarkOS validator) by around 20%

howardwu:
Can you give some color on the Poseidon and Algebraic Sponge (which is used in Poseidon) runtime performance changes?

The change is fairly nuanced, with no docs, and adds complexity to the existing code. I'd like to make sure we don't have any speed regressions.

Separately, `snarkvm-console-algorithms` also has an impl of Poseidon. Can you check whether these findings apply there as well?

ljedrz:
> The change is fairly nuanced, with no docs, and adds complexity to the existing code. I'd like to make sure we don't have any speed regressions.

I'm assuming you mean the final commit; that change is simpler to digest when read individually, but I'm happy to extend its docs. I'll provide the numbers for the performance impact shortly.

ljedrz:
I enhanced the existing `PoseidonSponge` benchmarks to provide more reliable results, and for `absorb_nonnative_field_elements` as well; compared to `testnet3`, this PR yields the following:
```
PoseidonSponge<2, 1> Absorb 100 native
                        time:   [1.3376 ms 1.3443 ms 1.3499 ms]
                        change: [-3.8885% -3.5254% -3.1465%] (p = 0.00 < 0.05)
                        Performance has improved.

PoseidonSponge<2, 1> Absorb 100 nonnative
                        time:   [5.5231 ms 5.5259 ms 5.5289 ms]
                        change: [-7.9397% -7.8100% -7.6850%] (p = 0.00 < 0.05)
                        Performance has improved.
Found 4 outliers among 100 measurements (4.00%)
  1 (1.00%) high mild
  3 (3.00%) high severe
```
most of the improvement comes from https://github.com/AleoHQ/snarkVM/pull/2038/commits/27787b51617b3c42a1a80ad02265c4fc2eb2419a.

ljedrz:
As for the allocation counts (measured with a small binary running a single iteration of both benchmarks from https://github.com/AleoHQ/snarkVM/pull/2038/commits/3a3f94b5c7dc2a26ca344e4099d41505d00c8157), they are as follows:
```
testnet3

total allocations: 8,485
total reallocations (growth): 5,713
```

```
this PR w/o 27787b5

total allocations: 4,971 (-41%)
total reallocations (growth): 3,499 (-39%)
```

```
this PR (full)

total allocations: 1,478 (-82%)
total reallocations (growth): 705 (-88%)
```

ljedrz:
Added extra docs explaining why the vectors are being reused.

ljedrz:
> Separately, `snarkvm-console-algorithms` also has an impl of Poseidon. Can you check whether these findings apply there as well?

I haven't seen that one pop up in the profiles, but I'll look into it.

update: double-checked; these improvements don't apply there.

howardwu (approved):

PR #2039, opened 2023-10-03
[zksecurity 06] Explicitly check consistency of proof component lengths

vicsn:
## Motivation

Zksecurity informational finding as defense in depth to prevent amiguous encoding in Fiat-Shamir transcript.

> Recommendation. We recommend adding explicit length checks to all the vectors in the `Proof` (as is currently done
for the public input) to avoid accidentally introducing vulnerabilities in the future if the serialization code changes,
since the serialization/deserialization code might erroneously be considered security non-critical

### DIFF
--- a/algorithms/src/snark/varuna/data_structures/proof.rs
+++ b/algorithms/src/snark/varuna/data_structures/proof.rs
@@ -251,19 +251,44 @@ impl<E: PairingEngine> Proof<E> {
         pc_proof: sonic_pc::BatchLCProof<E>,
     ) -> Result<Self, SNARKError> {
         let batch_sizes: Vec<usize> = batch_sizes.into_values().collect();
-        let total_instances = batch_sizes.iter().sum::<usize>();
-        if commitments.witness_commitments.len() != total_instances {
-            return Err(SNARKError::BatchSizeMismatch);
-        }
         Ok(Self { batch_sizes, commitments, evaluations, third_msg, fourth_msg, pc_proof })
     }
 
-    pub fn batch_sizes(&self) -> Result<&[usize], SNARKError> {
+    pub fn batch_sizes(&self) -> &[usize] {
+        &self.batch_sizes
+    }
+
+    /// Check that the number of messages is consistent with our batch size
+    pub fn check_batch_sizes(&self) -> Result<(), SNARKError> {
         let total_instances = self.batch_sizes.iter().sum::<usize>();
         if self.commitments.witness_commitments.len() != total_instances {
             return Err(SNARKError::BatchSizeMismatch);
         }
-        Ok(&self.batch_sizes)
+        let g_comms =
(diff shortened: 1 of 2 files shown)


algorithms/src/snark/varuna/data_structures/proof.rs
```diff
@@ -251,19 +251,45 @@ impl<E: PairingEngine> Proof<E> {
...
         let batch_sizes: Vec<usize> = batch_sizes.into_values().collect();
-        let total_instances = batch_sizes.iter().sum::<usize>();
-        if commitments.witness_commitments.len() != total_instances {
-            return Err(SNARKError::BatchSizeMismatch);
-        }
         Ok(Self { batch_sizes, commitments, evaluations, third_msg, fourth_msg, pc_proof })
     }
 
     pub fn batch_sizes(&self) -> Result<&[usize], SNARKError> {
+        self.check_batch_sizes()?;
```
ljedrz: I think this should instead be called before the current callsite of `batch_sizes`; otherwise we may perform multiple redundant checks in the future (as `batch_sizes` doesn't suggest that any checks are being performed)

algorithms/src/snark/varuna/data_structures/proof.rs
```diff
@@ -251,19 +251,45 @@ impl<E: PairingEngine> Proof<E> {
...
+        for evals in g_evals {
+            if evals.len() != self.batch_sizes.len() {
+                return Err(SNARKError::BatchSizeMismatch);
+            }
+        }
+        if self.third_msg.sums.len() != self.batch_sizes.len() {
+            return Err(SNARKError::BatchSizeMismatch);
+        }
+        for (i, msg) in self.third_msg.sums.iter().enumerate() {
+            if msg.len() != self.batch_sizes[i] {
```
ljedrz:
```suggestion
        for (msg, batch_size) in self.third_msg.sums.iter().zip(self.batch_sizes.iter()) {
            if msg.len() != batch_size {
```
(untested, but should work)

vicsn: Pushed

algorithms/src/snark/varuna/data_structures/proof.rs
vicsn: Yeah was thinking the same. I guess the rationale for calling it from `batch_sizes()` is some sort of assurance that the check will be called... But I moved it into `varuna.rs` and future developers will not be expected to gratitiously remove that check.

ljedrz (approved):
LGTM :+1:

howardwu:
LGTM.

Note to self: In the future, we could just do the `check` in `Proof::new` and fix the `deserialize` to use `Proof::new`. It is much easier to do this in `To/FromBytes` than the `Canonical` infra.

howardwu (approved):

PR #2041, opened 2023-10-04
Have Keccak and SHA-3 operate on any bits.

acoglio:
As discussed on Slack and agreed with Howard, this commit removes the code that zero-extends the input bits of Keccak and SHA-3 to be byte-aligned. Consistently with FIPS 202, these hashes are bit-oriented, and operate on bit sequences of any lengths.

While typical implementations of these hashes often take byte (not bit) sequences as inputs, our implementation may be used to hash boolean arrays in the future.  In this case, we want different boolean arrays to hash to different values, but the zero-extension to byte would cause collisions in that case.

Our current tests compare our implementation with our console wrapper of an implementation that operates on byte sequences. So we need to zero-extend to byte in the tests before calling our implementation, as indicated by the added comments.

At some point we should probably add other tests with known vectors, for both byte-aligned and non-byte-aligned bit sequences.


acoglio:
I'm looking at the CI failures. It looks like other tests must be adapted to the changes in Keccak and SHA-3.

bendyarm (approved):
LGTM

acoglio:
@howardwu The new commits should take care of all the CI failures except for the discrepancy between evaluation (which zero-extends to byte before Keccak'ing) and execution (which doesn't zero-extend to byte, per this PR). I discussed this with @d0cd and @bendyarm and we came up with the following plan:
- See if there exists an alternative Keccak implementation that, unlike tiny-keccak, provides a bit interface (obviating the need for zero-extending to byte) and that we can use in evaluation.
- If none is found, we could re-purpose this PR to add a check that the input to our Keccak is byte-aligned, adapting our tests as needed, and documenting that. This avoids baking the zero-extension to byte into our Keccak, and leaves the door open to remove the restriction if we find a way to do bit-oriented Keccak in evaluation in the future.
- Or we could stay with the current Keccak implementation, and document that it zero-extends to byte. It's a bit of a pity I think, but it's not wrong (if we are okay with, in the future, arrays `[true]` and `[true, false]` hashing to the same value).

acoglio:
Given that we couldn't find an outside implementation that is bit-oriented to use for the console computation, we have no choice but to have our Keccak zero-extend to bits. So I'm closing this PR. I would still suggest a slight refactoring in which we have a Keccak that operates on bits and a thin wrapped that zero extends to byte and then calls the Keccak of bits.

acoglio:
Re-opening this due to new/changed requirements, as discussed today. @howardwu

howardwu:
Discussed offline, we will hold off for now.

PR #2044, opened 2023-10-05
Use an array in IsInstruction, AssertInstruction, CommitInstruction and SignVerify

ljedrz:
Since we only ever use and expect 2 operands in `IsInstruction`, it's both faster and safer to use a fixed-sized array, which is not only faster to clone, but also makes the length checks redundant (as they are enforced at compilation time).

### DIFF
--- a/synthesizer/program/src/logic/instruction/operation/assert.rs
+++ b/synthesizer/program/src/logic/instruction/operation/assert.rs
@@ -36,15 +36,13 @@ enum Variant {
 #[derive(Clone, PartialEq, Eq, Hash)]
 pub struct AssertInstruction<N: Network, const VARIANT: u8> {
     /// The operands.
-    operands: Vec<Operand<N>>,
+    operands: [Operand<N>; 2],
 }
 
 impl<N: Network, const VARIANT: u8> AssertInstruction<N, VARIANT> {
     /// Initializes a new `assert` instruction.
     #[inline]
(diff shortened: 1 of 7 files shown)


ljedrz:
I noticed that the same applies to `AssertInstruction`, `CommitInstruction` and `SignVerify`; pushed 3 commits more.

ljedrz:
Note: these changes are a win as long as we don't expect to have a lot of such objects in existence at once (due to limited stack space); if we don't have tests for the maximum allowed (or just large) number of instructions, we should consider adding some to be on the safe side.

howardwu:
What's the real impact perf of this PR?

The design was intentional in using Vec so that we can expand the number of operands in the future.

ljedrz:
I wouldn't expect much performance impact from these changes in usual scenarios (at least judging by typical runtime profiles); the bigger benefit is from added (compile-time) safety.

Fort those objects where the future (increased) number of operands will still be fixed, this is a desirable change, but for those where we want that number to eventually be variable, a vector is more future-proof indeed. If all of them are going to have a variable number of operands in the future, the PR can be closed :+1:.

howardwu:
Let's hold off on this PR, can leave it around for now. In the future, if it becomes clear that this interface will remain unchanged, then we can use this PR. Given there is minimal perf changes, and the safety checks are there currently, it doesn't seem like a significant tradeoff.

PR #2045, opened 2023-10-05
[Fix] Check constraints on `snarkvm run`.

d0cd:
This PR ensures that constraints are checked when invoking `snarkvm run`.

Considerations
- This required introducing the `CallStack::PackageRun`, since enforcing constraints in `CheckDeployment` would cause `verify_deployment` to fail spuriously.
- An alternative implementation was considered, where `package::run` instantiated a call stack in `Evaluate` mode, and invoked `evaluate_function`. This work can be found in `fix/run-command-0`.


howardwu (approved):
LGTM

PR #2047, opened 2023-10-06
[Optimization] Checked Multiplication

d0cd:
This PR optimizes the checked multiplication circuits in snarkVM by:
- Efficiently using witnesses to reduce constraints for integers smaller than 128 bits.
- Use Karatsuba's algorithm.
- Return flags for overflow, rather than a bit decomposition.

This depends on #2046.

PR #2048, opened 2023-10-06
[Optimize] Unsigned checked addition

d0cd:
This PR optimizes unsigned checked addition by directly witnessing the product and checking that the sum and witness match in the base field.

This PR depends on #2047

### DIFF
--- a/circuit/types/integers/src/add_checked.rs
+++ b/circuit/types/integers/src/add_checked.rs
@@ -73,37 +73,38 @@ impl<E: Environment, I: IntegerType> AddChecked<Self> for Integer<E, I> {
                 Some(value) => Integer::constant(console::Integer::new(value)),
                 None => E::halt("Integer overflow on addition of two constants"),
             }
-        } else {
-            // TODO: Optimize
+        } else if I::is_signed() {
             // Instead of adding the bits of `self` and `other` directly, the integers are
             // converted into a field elements, and summed, before converting back to integers.
             // Note: This is safe as the field is larger than the maximum integer type supported.
             let sum = self.to_field() + other.to_field();
 
-            // Extract the integer bits from the field element, with a carry bit.
-            let (sum, carry) = match sum.to_lower_bits_le(I::BITS as usize + 1).split_last() {
-                Some((carry, bits_le)) => (Integer::from_bits_le(bits_le), carry.clone()),
+            // Extract the integer bits from the field element, ignoring the carry bit as it is not relevant for signed addition.
+            let sum = match sum.to_lower_bits_le(I::BITS as usize + 1).split_last() {
+                Some((_, bits_le)) => Integer::from_bits_le(bits_le),
                 // Note: `E::halt` should never be invoked as `I::BITS as usize + 1` is greater than zero.
                 None => E::halt("Malformed sum detected during integer addition"),
             };
 
-            // Check for overflow.
-            match I::is_signed() {
-                // For signed addition, overflow and underflow conditions are:
-                //   - a > 0 && b > 0 && a + b < 0 (Overflow)
-                //   - a < 0 && b < 0 && a + b > 0 (Underflow)
-                //   - Note: if sign(a) != sign(b) then over/underflow is impossible.
-                //   - Note: the result of an overflow and underflow must be negative and positive, respectively.
-                true => {
-                    let is_same_sign = self.msb().is_equal(other.msb());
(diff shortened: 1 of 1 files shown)


circuit/types/integers/src/add_checked.rs
```diff
@@ -73,37 +73,38 @@ impl<E: Environment, I: IntegerType> AddChecked<Self> for Integer<E, I> {
...
             // Instead of adding the bits of `self` and `other` directly, the integers are
             // converted into a field elements, and summed, before converting back to integers.
             // Note: This is safe as the field is larger than the maximum integer type supported.
             let sum = self.to_field() + other.to_field();
 
-            // Extract the integer bits from the field element, with a carry bit.
-            let (sum, carry) = match sum.to_lower_bits_le(I::BITS as usize + 1).split_last() {
-                Some((carry, bits_le)) => (Integer::from_bits_le(bits_le), carry.clone()),
+            // Extract the integer bits from the field element, ignoring the carry bit as it is not relevant for signed addition.
+            let sum = match sum.to_lower_bits_le(I::BITS as usize + 1).split_last() {
```
acoglio, line 83:
```suggestion
            let sum = match sum.to_lower_bits_le(I::BITS as usize) {
```
I think that the call of `to_lower_bits_le` is still allocating an R1CS boolean variable for the carry, even though the pattern below `Some((_, ...))` ignores it at the Rust level. But if instead we pass `I::BITS` to `to_lower_bits_le` and adjust the pattern below, it should only allocate `I::BITS` for the result, and ensure that their weighted sum equals the field value of the sum.

circuit/types/integers/src/add_checked.rs
```diff
@@ -73,37 +73,38 @@ impl<E: Environment, I: IntegerType> AddChecked<Self> for Integer<E, I> {
...
+                Some((_, bits_le)) => Integer::from_bits_le(bits_le),
```
acoglio, line 84:
```suggestion
                Some(bits_le) => Integer::from_bits_le(bits_le),
```
To match suggested change above.

circuit/types/integers/src/add_checked.rs
```diff
@@ -73,37 +73,38 @@ impl<E: Environment, I: IntegerType> AddChecked<Self> for Integer<E, I> {
...
+            //   - Note: if sign(a) != sign(b) then over/underflow is impossible.
+            //   - Note: the result of an overflow and underflow must be negative and positive, respectively.
+            let is_same_sign = self.msb().is_equal(other.msb());
+            let is_overflow = is_same_sign & sum.msb().is_not_equal(self.msb());
+            E::assert_eq(is_overflow, E::zero());
+
+            sum
+        } else {
+            // Instead of adding the bits of `self` and `other` directly, witness the integer sum.
+            let sum: Integer<E, I> = witness!(|self, other| self.add_wrapped(&other));
```
acoglio, line 101: I think that `add_wrapped` generates an extra R1CS boolean variable for the carry. Since we do checked addition here, we can save that variable by doing something like what's done above for the signed case, allocating just `I::BITS` variables for the result.

circuit/types/integers/src/add_checked.rs
d0cd: This doesn't work because the addition uses the `I::BITS + 1` lower bits of the field element.
`to_lower_bits_le` requires that the you extract all of the nonzero bits from the field.
The interesting thing is that for signed integer addition, overflow is not determined by the carry bit, rather by the signs.
https://www.doc.ic.ac.uk/~eedwards/compsys/arithmetic/index.html

circuit/types/integers/src/add_checked.rs
d0cd: See comment above.

circuit/types/integers/src/add_checked.rs
d0cd: This is inside the `witness!` macro, so only `I::BITS` variables should get allocated no?

bendyarm (approved):
LGTM

circuit/types/integers/src/add_checked.rs
acoglio: Yes, I missed that it was done inside the `witness!`.

circuit/types/integers/src/add_checked.rs
acoglio: You are right, I forgot about the signed case somehow (I was just thinking unsigned).

acoglio (approved):

PR #2049, opened 2023-10-06
[Fix] Square root

d0cd:
This PR fixes a non-determinism issue in the square root circuit by ensuring that that result is less than (p - 1)/2, where p is the field modulus.

Note: this PR affects the `elligator2::encode` circuit. If determinism is not needed, we should consider adding a `square_root_unchecked` circuit.

This PR depends on #2072


acoglio (approved):

bendyarm (approved):
Looks good!

synthesizer/process/src/tests/test_credits.rs
```diff
@@ -1536,9 +1536,9 @@ mod sanity_checks {
         // Compute the assignment.
         let assignment = get_assignment::<_, CurrentAleo>(stack, &private_key, function_name, &[r0, r1, r2], rng);
         assert_eq!(15, assignment.num_public());
-        assert_eq!(54688, assignment.num_private());
-        assert_eq!(54747, assignment.num_constraints());
-        assert_eq!((88514, 130701, 83646), assignment.num_nonzeros());
+        assert_eq!(55698, assignment.num_private());
+        assert_eq!(55761, assignment.num_constraints());
+        assert_eq!((90368, 132045, 84980), assignment.num_nonzeros());
```
Pratyush: This pushes us to the next power of two here (2^18, as opposed to 2^17)

howardwu: Great catch, I didn't notice that, thank you!

howardwu: I've addressed the imbalance with this PR:
- #2087

PR #2050, opened 2023-10-06
[Optimize] Left shift.

d0cd:
This PR optimizes the wrapped/checked left shift implementations for all integer types **except** `i/u128`.

This PR depends on #2048.

PR #2052, opened 2023-10-07
[Optimize] Right shift.

d0cd:
This PR optimizes right shifts for all integers except for `i/u128`.

This PR depends on #2050

### DIFF
--- a/circuit/types/integers/src/shr_wrapped.rs
+++ b/circuit/types/integers/src/shr_wrapped.rs
@@ -67,7 +67,7 @@ impl<E: Environment, I: IntegerType, M: Magnitude> ShrWrapped<Integer<E, M>> for
 
                     // Calculate the result directly in the field.
                     // Since 2^{rhs} < Integer::MAX and 2 * I::BITS is less than E::BaseField::size in data bits,
-                    // we know that the operation will not overflow Integer::MAX or the field modulus.
+                    // we know that the operation will not overflow the field modulus.
                     for (i, bit) in rhs.bits_le[..first_upper_bit_index].iter().enumerate() {
                         // In each iteration, multiply the result by 2^(1<<i), if the bit is set.
                         // Note that instantiating the field from a u128 is safe since it is larger than all eligible integer types.
--- a/circuit/types/integers/src/shr_wrapped.rs
+++ b/circuit/types/integers/src/shr_wrapped.rs
@@ -73,8 +73,13 @@ impl<E: Environment, I: IntegerType, M: Magnitude> ShrWrapped<Integer<E, M>> for
                         // Note that instantiating the field from a u128 is safe since it is larger than all eligible integer types.
                         let constant = Field::constant(console::Field::from_u128(2u128.pow(1 << i)));
                         let product = &result * &constant;
+
+                        // If `self` is negative, mask the value with 2^{1<<i} - 1.
+                        // For example, in the first, second, and third iterations, the mask is 0b1, 0b11, and 0b111, respectively.
+                        // This serves to appropriately sign-extend the result.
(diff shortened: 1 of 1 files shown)


circuit/types/integers/src/shr_wrapped.rs
```diff
@@ -56,6 +56,54 @@ impl<E: Environment, I: IntegerType, M: Magnitude> ShrWrapped<Integer<E, M>> for
...
+                    // Initialize the msb of `self` as a field element.
+                    let msb_field = Field::from((**self.msb()).clone());
+
+                    // The signed right-shift is implemented as an unsigned right-shift followed by a sign-extension.
+                    // Initialize the result from the reversed bits of `self`.
+                    let mut result = Field::from_bits_be(&self.bits_le);
+
+                    // Calculate the result directly in the field.
+                    // Since 2^{rhs} < Integer::MAX and 2 * I::BITS is less than E::BaseField::size in data bits,
+                    // we know that the operation will not overflow Integer::MAX or the field modulus.
```
acoglio:
```suggestion
                    // we know that the operation will not overflow the field modulus.
```

circuit/types/integers/src/shr_wrapped.rs
```diff
@@ -56,6 +56,54 @@ impl<E: Environment, I: IntegerType, M: Magnitude> ShrWrapped<Integer<E, M>> for
...
+                    bits_le.reverse();
+                    // Initialize the integer, ignoring the carry bits.
+                    Self { bits_le, phantom: Default::default() }
+                } else {
+                    // The unsigned right-shift is implemented as a left-shift over the reversed bits of `self`.
+                    // Initialize the result from the reversed bits of `self`.
+                    let mut result = Field::from_bits_be(&self.bits_le);
+                    // Calculate the result directly in the field.
+                    // Since 2^{rhs} < Integer::MAX and 2 * I::BITS is less than E::BaseField::size in data bits,
+                    // we know that the operation will not overflow Integer::MAX or the field modulus.
```
acoglio:
```suggestion
                    // we know that the operation will not overflow the field modulus.
```

circuit/types/integers/src/shr_wrapped.rs
```diff
@@ -56,6 +56,54 @@ impl<E: Environment, I: IntegerType, M: Magnitude> ShrWrapped<Integer<E, M>> for
...
+
+                    // Calculate the result directly in the field.
+                    // Since 2^{rhs} < Integer::MAX and 2 * I::BITS is less than E::BaseField::size in data bits,
+                    // we know that the operation will not overflow Integer::MAX or the field modulus.
+                    for (i, bit) in rhs.bits_le[..first_upper_bit_index].iter().enumerate() {
+                        // In each iteration, multiply the result by 2^(1<<i), if the bit is set.
+                        // Note that instantiating the field from a u128 is safe since it is larger than all eligible integer types.
+                        let constant = Field::constant(console::Field::from_u128(2u128.pow(1 << i)));
+                        let product = &result * &constant;
+                        let mask = Field::constant(console::Field::from_u128(2u128.pow(1 << i) - 1));
```
acoglio, line 80: It would be good to add comments to explain this mask and its addition to the result when the operand being shifted is negative. For now we could say something like: "If the self operand being shifted is negative, in each iteration of the loop, we add to the result the binary value 1...1 consisting of i 1s, e.g. when i is 3, we add 111, i.e. 7. This serves to replicate the sign bit: we shift by 3 (when i is 3), so we need to fill in 3 sign 1 bits." Later we should explain all of this in more detail, but it's for another day.

acoglio (approved):
Looks good, modulo the comments given.

PR #2055, opened 2023-10-08
Optimize field (in)equality circuit.

acoglio:
To calculate `z = x != y`, instead of
```
(z) (1 - z) = (0)
(x - y) (w) = (z)
(x - y) (1 - z) = (0)
```
(where `w` is an internal variable) we just need
```
(x - y) (w) = (z)
(x - y) (1 - z) = (0)
```
i.e. just the 2nd and 3rd constraints, because they imply the 1st one. So we save one constraint for every field (in)equality.

This optimization has been already verified in the ACL2 theorem prover.

To avoid generating the boolean constraint, we need to avoid `Boolean::new`, and instead generate the `Boolean` directly, as in boolean AND for example. However, boolean AND has access to the private `LinearCombination` component of `Boolean`, while field (in)equality does not. So I've added a `from_variable` constructor to `Boolean`, but if there is a better or more desirable approach, we can revise this part.

This PR depends on #2052

### DIFF
--- a/circuit/types/field/src/equal.rs
+++ b/circuit/types/field/src/equal.rs
@@ -46,11 +46,35 @@ impl<E: Environment> Equal<Self> for Field<E> {
             // Otherwise, we introduce a private field variable is_neq for the result,
             // along with an auxiliary variable multiplier for the inverse of the difference between the operands,
             // and enforce the following constraints:
-            //   (self - other) (multiplier) = (is_neq)
-            //   (self - other) (1 - is_neq) = (0)
-            // These constraints imply that is_neq is boolean, i.e. either 0 or 1;
+            // 1.  (self - other) (multiplier) = (is_neq)
+            // 2.  (self - other) (1 - is_neq) = (0)
+            // These constraints imply that is_neq is boolean, i.e. either 0 or 1 (see below);
             // so we avoid creating is_neq as a Boolean, which would generate an unneeded boolean constraint.
-            // See the comments just after the code for more detailed explanations.
+            //
+            // The specification of this circuit is the calculation of field inequality:
+            //   is_neq = [IF self = other THEN 0 ELSE 1]
+            // i.e. the result is_neq is 0 if the two operands are equal, 1 if they are not equal.
+            //
+            // The correctness of the circuit, i.e. its equivalence to the specification, is proved as follows.
+            //
+            // Soundness: the constraints imply the specification,
+            // i.e. every solution of the constraints corresponds to a correct computation of field inequality.
+            // - If self = other, constraint 1 implies is_neq = 0, consistently with the specification,
+            // - If self != other, constraint 2 implies (1 - is_neq) = 0, i.e. is_neq = 1, consistently with the specification.
+            //
+            // Completeness: the specification implies the constraints,
+            // i.e. the constraints are satisfiable for every correct computation of field inequality.
+            // - If self = other, constraint 2 reduces to 0 = 0.
+            //   The specification implies is_neq = 0, and constraint 1 reduces to 0 = 0,
+            //   regardless of the choice of multiplier.
+            // - If self != other, the specification implies is_neq = 1,
+            //   which reduces constraint 2 to 0 = 0 and constraint 1 to (self - other) (multiplier) = 1,
+            //   which is satisfied by choosing multiplier = 1 / (self - other),
+            //   which is well-defined because (self - other) != 0.
+            //
+            // Thus the circuit is equivalent to the specification.
+            // Since the specification implies that is_neq is either 0 or 1,
+            // the circuit does not need a boolean constraint for is_neq, as mentioned above.
             _ => {
                 let is_neq = Boolean::from_variable(E::new_variable(Mode::Private, match is_neq_ejected {
                     true => E::BaseField::one(),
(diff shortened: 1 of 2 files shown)


acoglio:
This PR also streamlines and speeds up the code that handles the case of two constant operands. The use of `witness!` was checking again whether the operands are constants, so the double check is now avoided. Another use of `witness!` in this PR (which was there before) could also be optimized.

circuit/types/field/src/equal.rs
```diff
@@ -110,21 +128,6 @@ impl<E: Environment> Equal<Self> for Field<E> {
                 // Check 2:  (a - b) * not(1) = 0
                 //                          0 = 0
                 // => is_neq is trivially correct.
```
howardwu: Is this documentation still relevant/does it need to be updated?

howardwu: It looks like this is still relevant.

It may be advisable to revert the bottom code so that its documentation remains in place.

circuit/types/field/src/equal.rs
```diff
@@ -32,24 +33,41 @@ impl<E: Environment> Equal<Self> for Field<E> {
...
+            // so we avoid creating is_neq as a Boolean, which would generate an unneeded boolean constraint.
+            // See the comments just after the code for more detailed explanations.
             _ => {
-                // Compute a boolean that is `true` if `this` and `that` are not equivalent.
-                let is_neq: Boolean<E> = witness!(|self, other| self != other);
-
-                // Assign the expected multiplier.
-                let multiplier: Field<E> = witness!(|self, other| {
-                    match (self - other).inverse() {
+                let is_neq = Boolean::from_variable(E::new_variable(Mode::Private, match is_neq_ejected {
```
howardwu, line 82: Can you convince me this is safe?

Without a booleanity check, a malicious actor can allocate `is_neq` with any field value. (As we no longer enforce 0/1 field).

acoglio: Yes, good question. Here is a proof that the remaining constraints
```
1.   (self - other) (multiplier) = (is_neq)
2.   (self - other) (1 - is_neq) = (0)
```
imply that `is_neq` is boolean. There are two cases: either `self = other` or `self != other`. If they are equal, constraint 1 implies `is_neq = 0`. If they are not equal, constraint 2 implies `1 - is_neq = 0`, i.e. `is_neq = 1`. So `is_neq` is either 0 or 1. If a malicious actor tries to construct a solution to the constraints with `is_neq` neither 0 nor 1, the system of two constraints won't be satisfied.

Another way to come to the same conclusion is to prove the correctness of this circuit, i.e. that the two constraints are equivalent to the specification of the operation, which is
```
is_neq = [IF self = other THEN 0 ELSE 1]
```
and which shows that `is_neq` can only be 0 or 1. (I can spell out this correctness proof if you want.)

acoglio: This is similar to what we do in boolean operations like [and](https://github.com/AleoHQ/snarkVM/blob/9cd5a323c611b65d2042282d0398d9eec88303ff/circuit/types/boolean/src/and.rs#L83-L96): note the comment there about `output` being guaranteed boolean by the constraints. (Side note: the boolean operations require preconditions that the operands are boolean; field (in)equality has no preconditions, works on any field values `self` and `other`.)

acoglio: Just in case you are curious, the ACL2 proof of the optimized circuit in PFCS form is [here](https://github.com/AleoHQ/aleo-acl2/blob/master/circuits/pfcs/field-neq-opt.lisp). [This part](https://github.com/AleoHQ/aleo-acl2/blob/0cb83fc59daae31c8641c6370fd0bb735b17b5d7/circuits/pfcs/field-neq-opt.lisp#L66) is the specification of the circuit: the output `z` (i.e. `is_neq`) is 0 or 1 based on whether the inputs `x` (i.e. `self`) and `y` (i.e. `other`) are equal or not.

circuit/types/field/src/equal.rs
acoglio: I think it's still relevant and applies in its current form, but we could also consider shortening it instead (I can add a proposed block of comments with a shorter version; let me know).

acoglio: We could move some of the code relative to the block comment, but I've also refactored it slightly, so I'm not sure where exactly we want to move the block comment.

howardwu: It seems we can revert L66-69 (new) back to L118-127 (old).

howardwu: And agreed, do you want to proposed a simplified "proof" in the comments (for the cases)?

acoglio: I just pushed a commit with a shorter explanation of correctness. I also put it before the code, to address the code movement you mentioned (I moved the comment instead). But let me know if you'd like something different.

circuit/types/field/src/equal.rs
d0cd: This follows. In the form of the previous explanation,
```
// Inequality Enforcement
// ----------------------------------------------------------------
// Check 1:  (a - b) * multiplier = is_neq
// Check 2:  (a - b) * (1 - is_neq) = 0
//
//
// Case 1: a == b AND is_neq == 0 (honest)
// ----------------------------------------------------------------
// Check 1:  (a - b) * multiplier = 0
//                 0 * multiplier = 0
//                              0 = 0
// => The constraint is satisfied.
//
// Check 2:  (a - b) * (1 - 0) = 0
//                       0 * 1 = 0
//                           0 = 0
// => The constraint is satisfied.
//
//
// Case 2: a == b AND is_neq != 0 (dishonest)
// ----------------------------------------------------------------
// Check 1:  (a - b) * multiplier = is_neq
//                 0 * multiplier = is_neq
//                              0 = is_neq
// => As is_neq != 0, the constraint is not satisfied.
//
//
// Case 3: a != b AND is_neq != 1 (dishonest).
// ----------------------------------------------------------------
// Check 2:  (a - b) * (1 - is_neq) = 0
//                     (1 - is_neq) = 0
// => As is_neq != 1, the constraint is not satisfied.
//
//
// Case 4a: a != b AND is_neq == 1 AND multiplier = n [!= (a - b)^(-1)] (dishonest)
// ---------------------------------------------------------------------------------
// Check 1:  (a - b) * n = 1
// => As n != (a - b)^(-1), the constraint is not satisfied.
//
//
// Case 4b: a != b AND is_neq == 1 AND multiplier = (a - b)^(-1) (honest)
// ---------------------------------------------------------------------------------
// Check 1:  (a - b) * (a - b)^(-1) = 1
//                                1 = 1
// => The constraint is satisfied.
//
// Check 2:  (a - b) * (1 - 1) = 0
//                 (a - b) * 0 = 0
//                           0 = 0
// => The constraint is satisfied.
//
//
// Observe that in both of the honest cases, `is_neq` is always 0 or 1.
```

circuit/types/field/src/equal.rs
```diff
@@ -32,98 +33,63 @@ impl<E: Environment> Equal<Self> for Field<E> {
...
-
-                // Check 1: (a - b) * multiplier = is_neq
-                E::enforce(|| (&delta, &multiplier, &is_neq));
-
-                // Check 2: (a - b) * not(is_neq) = 0
-                E::enforce(|| (delta, is_eq, E::zero()));
-
+                let is_eq = !is_neq.clone(); // 1 - is_neq
+                E::enforce(|| (&delta, &multiplier, &is_neq)); // 1st constraint
+                E::enforce(|| (delta, is_eq, E::zero())); // 2nd constraint
```
d0cd: Nit. The comments should be in the same style.

circuit/types/field/src/equal.rs
```diff
@@ -32,98 +33,63 @@ impl<E: Environment> Equal<Self> for Field<E> {
...
-                // Assign the expected multiplier.
-                let multiplier: Field<E> = witness!(|self, other| {
-                    match (self - other).inverse() {
+                let is_neq = Boolean::from_variable(E::new_variable(Mode::Private, match is_neq_ejected {
+                    true => E::BaseField::one(),
+                    false => E::BaseField::zero(),
+                }));
+                let delta = self - other;
+                let multiplier: Field<E> = witness!(|delta| {
+                    match delta.inverse() {
```
d0cd: Nit. The comments should be in the same style.

d0cd (changes requested):

circuit/types/field/src/equal.rs
acoglio: I just pushed something: let me know if that adheres to the convention (which appears to be to put `//` at the beginning of the code block they describe, separating code blocks with blank lines).

PR #2056, opened 2023-10-08
Fix bug in parsing of `hash_many`.

acoglio:
This failed
```
hash_many.psd2 r1 r2 into r3 as field;
```
while this passed
```
hash_many.psd2 r1r2 into r3 as field;
```

The problem was the lack of a call to `Sanitizer::parse_whitespaces` between the two operands.

This PR fixes the problem by systematically skipping whitespace just before each operand, as done in other parts of the Aleo instructions parser.

Also add a test demonstrating that the previous failing example above now passes.


howardwu (approved):
LGTM, great find!

PR #2059, opened 2023-10-08
Staging into testnet3

howardwu:
## Changelog

- #2039


howardwu:
Early merging this since it is a non-breaking change.

PR #2061, opened 2023-10-08
[zksecurity 04] Proof Delegation Leaks User Signing Key

vicsn:
## Motivation

Proof Delegation Leaks User Signing Key, let's avoid this.

Requires resampling.

## Test Plan

It would be great to have a clean write-up in our spec of all the components we're verifying. Will support Michel on this. We might also be able to remove the tcm check in the verify function.


raychu86 (approved):
LGTM! Will need to resample parameters

howardwu (approved):

PR #2068, opened 2023-10-09
Merklize `subdag_root` derivation

raychu86:
## Motivation

Update the `subdag_root` derivation from the leader certificate ID to a merkle root, where the leaves are the certificate ids of the subdag.


howardwu (approved):
LGTM

PR #2074, opened 2023-10-10
Fix algorithms dependency version

raychu86:
## Motivation

This PR is a simple fix where we update the dependency version of `snarkvm-algorithms` to the latest version, `0.15.4`.


d0cd (approved):
LGTM!

PR #2076, opened 2023-10-11
[ZKS-05] Caller Is Not Fixed Throughout Function Execution

vicsn:
## Motivation

We should ensure the self.signer is the same throughout a call graph in a single Execution.

As recommended, this PR adds binding and hiding commitments to the signer in each transition, which needs a secret which is constant for the entire execution. We propose the following, which will leak the signer to everyone holding the root_tvk:
- `signer_commitment = hash(signer address || root_tvk)`

Notes:
- performance improvement: each request in an execution has the same signer and scm now, in the future we could refactor and perhaps store a single signer and scm in the `Execution`
- we could hypothetically check [if all signer commitments in an entire **transaction** are the same](https://github.com/AleoHQ/snarkVM/blob/testnet3/synthesizer/src/vm/verify.rs#L88), but perhaps this doesn't leave open the door for more flexible or multi-execute transactions.
- we cannot use the account view_key as secret because we'd leak it to the party we delegate to. It is also expensive to derive the address from the view_key within the circuit.

## Test Plan

- Requires resampling, including `./resources/block.genesis`


mdelle1:
Posting what I sent you in Slack with some updates on the commitment: 

**Caller not fixed in function execution:** 
First, we can add to the ‘request’ object a signer view key svk. The svk is included in the request signature.
I think the appropriate fix in CallTrait::execute  https://github.com/AleoHQ/snarkVM/blob/testnet3/synthesizer/process/src/stack/call/mod.rs#L125 would be to: 
1. Load the signer address (signer) for the caller circuit from the registers, through registers.signer() 
2. Retrieve the signer address (signer’) for the callee circuit from this request https://github.com/AleoHQ/snarkVM/blob/testnet3/synthesizer/process/src/stack/call/mod.rs#L261 
3. Compute scm’ := HASH(signer’, svk’), where svk’ is obtained from the callee request. 
4. Inject scm’ as a variable into the caller circuit in mode Public. 
5. Retrieve svk from the caller request stored in the registers. 
6. Add a constraint to the caller circuit that the injected variable in Step 4 equal HASH(svk, signer). 
7. The verifier will check that the injected (public) value in Step 4 is the same for the caller and callee circuit.

mdelle1:
Regarding the caller not being fixed in function execution flow, what are your thoughts about this? This pertains to obtaining the same randomness across the entire authorization. 

**On the user (delegator) side:** 
1. A randomizer is generated when the root request is constructed, which would occur in Stack.authorize before this line https://github.com/AleoHQ/snarkVM/blob/testnet3/synthesizer/process/src/stack/authorize.rs#L38 
2. Each authorization sent to the third-party prover is now a tuple: 
(Authorized requests, randomizer, signature on randomizer that authenticates to pk of user)**

**I think it actually may be necessary to sign (tx id, randomizer). 

**On the third-party prover side:** 
1. When the prover is executing the authorization in Stack.execute_function, if we are in the context of the root request which can be checked via the is_root flag, we will add a constraint that authenticates the signature of the randomizer to the user pk. 
2. In every transition circuit, we can expose the signer commitment, ’scm’, as a public input and constrain that scm := Commit(randomizer, signer). 
3. The verifier will check each scm is the same across all transitions in the execution. 

How does the user actually compute the randomizer? 
One option is compute a ‘signer view key’. First compute ssk := HASH(sk_sig, nonce) and then ‘svk’ := (signer^ssk).to_x_coordinate().

circuit/types/address/src/lib.rs
```diff
@@ -45,6 +45,12 @@ impl<E: Environment> Address<E> {
     pub fn zero() -> Self {
         Self(Group::zero())
     }
+
+    /// Outputs the x coordinate of the address
+    #[inline]
+    pub fn to_x_coordinate(&self) -> Field<E> {
```
howardwu: Why not use `to_field`?

ledger/block/src/verify.rs
```diff
@@ -475,6 +475,7 @@ impl<N: Network> Block<N> {
         if has_duplicates(self.transition_commitments()) {
             bail!("Found a duplicate transition commitment in block {height}");
         }
+
```
howardwu:
```suggestion
```

synthesizer/process/src/stack/registers/mod.rs
```diff
@@ -51,6 +51,10 @@ pub struct Registers<N: Network, A: circuit::Aleo<Network = N>> {
     signer: Option<Address<N>>,
     /// The transition signer, as a circuit.
     signer_circuit: Option<circuit::Address<A>>,
+    /// The root transition viewkey.
```
howardwu:
```suggestion
    /// The root transition view key.
```

synthesizer/process/src/stack/registers/mod.rs
```diff
@@ -51,6 +51,10 @@ pub struct Registers<N: Network, A: circuit::Aleo<Network = N>> {
...
+    root_tvk: Option<Field<N>>,
+    /// The root transition viewkey, as a circuit.
```
howardwu:
```suggestion
    /// The root transition view key, as a circuit.
```

synthesizer/process/src/stack/registers/caller.rs
```diff
@@ -65,6 +77,18 @@ impl<N: Network, A: circuit::Aleo<Network = N>> RegistersSignerCircuit<N, A> for
         self.signer_circuit = Some(signer_circuit);
     }
 
+    /// Returns the root transition viewkey, as a circuit.
+    #[inline]
+    fn root_tvk_circuit(&self) -> Result<circuit::Field<A>> {
+        self.root_tvk_circuit.clone().ok_or_else(|| anyhow!("Root tvk (circuit) is not set in the registers."))
+    }
+
+    /// Sets the root transition viewkey, as a circuit.
```
howardwu:
```suggestion
    /// Sets the root transition view key, as a circuit.
```

synthesizer/process/src/stack/registers/caller.rs
```diff
@@ -65,6 +77,18 @@ impl<N: Network, A: circuit::Aleo<Network = N>> RegistersSignerCircuit<N, A> for
         self.signer_circuit = Some(signer_circuit);
     }
 
+    /// Returns the root transition viewkey, as a circuit.
```
howardwu:
```suggestion
    /// Returns the root transition view key, as a circuit.
```

synthesizer/process/src/stack/registers/caller.rs
```diff
@@ -27,6 +27,18 @@ impl<N: Network, A: circuit::Aleo<Network = N>> RegistersSigner<N> for Registers
         self.signer = Some(signer);
     }
 
+    /// Returns the root transition viewkey.
+    #[inline]
+    fn root_tvk(&self) -> Result<Field<N>> {
+        self.root_tvk.ok_or_else(|| anyhow!("root tvk (console) is not set in the registers."))
+    }
+
+    /// Sets the root transition viewkey.
```
howardwu:
```suggestion
    /// Sets the root transition view key.
```

synthesizer/process/src/stack/registers/caller.rs
```diff
@@ -27,6 +27,18 @@ impl<N: Network, A: circuit::Aleo<Network = N>> RegistersSigner<N> for Registers
         self.signer = Some(signer);
     }
 
+    /// Returns the root transition viewkey.
```
howardwu:
```suggestion
    /// Returns the root transition view key.
```

synthesizer/process/src/stack/registers/caller.rs
```diff
@@ -27,6 +27,18 @@ impl<N: Network, A: circuit::Aleo<Network = N>> RegistersSigner<N> for Registers
...
+    #[inline]
+    fn root_tvk(&self) -> Result<Field<N>> {
+        self.root_tvk.ok_or_else(|| anyhow!("root tvk (console) is not set in the registers."))
```
howardwu:
```suggestion
        self.root_tvk.ok_or_else(|| anyhow!("Root tvk (console) is not set in the registers."))
```

synthesizer/process/src/stack/helpers/synthesize.rs
```diff
@@ -52,13 +52,20 @@ impl<N: Network> Stack<N> {
         // Sample 'is_root'.
         let is_root = true;
 
+        // No root_tvk when deploying an individual circuit.
```
howardwu:
```suggestion
        // The `root_tvk` is `None` when deploying an individual circuit.
```

synthesizer/process/src/stack/helpers/synthesize.rs
```diff
@@ -52,13 +52,20 @@ impl<N: Network> Stack<N> {
...
+        let root_tvk = None;
+
+        // No caller when deploying an individual circuit.
```
howardwu:
```suggestion
        // The `caller` is `None` when deploying an individual circuit.
```

synthesizer/process/src/stack/deploy.rs
```diff
@@ -76,6 +76,12 @@ impl<N: Network> Stack<N> {
         // Construct the call stacks and assignments used to verify the certificates.
         let mut call_stacks = Vec::with_capacity(deployment.verifying_keys().len());
 
+        // No root_tvk when verifying deployment of an individual circuit.
```
howardwu:
```suggestion
        // The `root_tvk` is `None` when verifying the deployment of an individual circuit.
```

synthesizer/process/src/stack/deploy.rs
```diff
@@ -76,6 +76,12 @@ impl<N: Network> Stack<N> {
...
+        let root_tvk = None;
+
+        // No caller when verifying deployment of an individual circuit.
```
howardwu:
```suggestion
        // The `caller` is `None` when verifying the deployment of an individual circuit.
```

circuit/types/address/src/lib.rs
vicsn: 6230cab23

ledger/block/src/verify.rs
vicsn: eca98a6e8

synthesizer/process/src/stack/registers/mod.rs
vicsn: eca98a6e8

synthesizer/process/src/stack/registers/mod.rs
vicsn: eca98a6e8

synthesizer/process/src/stack/registers/caller.rs
vicsn: eca98a6e8

synthesizer/process/src/stack/registers/caller.rs
vicsn: eca98a6e8

synthesizer/process/src/stack/registers/caller.rs
vicsn: eca98a6e8

synthesizer/process/src/stack/registers/caller.rs
vicsn: eca98a6e8

synthesizer/process/src/stack/helpers/synthesize.rs
vicsn: eca98a6e8

synthesizer/process/src/stack/helpers/synthesize.rs
vicsn: eca98a6e8

synthesizer/process/src/stack/deploy.rs
vicsn: eca98a6e8

synthesizer/process/src/stack/deploy.rs
vicsn: eca98a6e8

synthesizer/process/src/stack/registers/caller.rs
vicsn: eca98a6e8

synthesizer/process/src/stack/execute.rs
```diff
@@ -194,6 +197,20 @@ impl<N: Network> StackExecute<N> for Stack<N> {
...
+        // inject the `root_tvk` as `Mode::Private`.
+        if let Some(root_tvk) = root_tvk {
+            registers.set_root_tvk(root_tvk);
+            let root_tvk_circuit = circuit::Field::<A>::new(circuit::Mode::Private, root_tvk);
+            registers.set_root_tvk_circuit(root_tvk_circuit);
+        } else {
+            registers.set_root_tvk(*console_request.tvk());
+            let root_tvk_circuit = circuit::Field::<A>::new(circuit::Mode::Private, *console_request.tvk());
+            registers.set_root_tvk_circuit(root_tvk_circuit);
+        }
```
raychu86:
```suggestion
        if let Some(root_tvk) = root_tvk {
            registers.set_root_tvk(root_tvk);
            registers.set_root_tvk_circuit(circuit::Field::<A>::new(circuit::Mode::Private, root_tvk));
        } else {
            registers.set_root_tvk(*console_request.tvk());
            registers.set_root_tvk_circuit(circuit::Field::<A>::new(circuit::Mode::Private, *console_request.tvk()));
        }
```

synthesizer/process/src/verify_execution.rs
```diff
@@ -131,6 +131,14 @@ impl<N: Network> Process<N> {
         let num_instances = verifier_inputs.values().map(|(_, inputs)| inputs.len()).sum::<usize>();
         // Ensure the number of instances matches the number of transitions.
         ensure!(num_instances == execution.transitions().len(), "The number of verifier instances is incorrect");
+        // Ensure the same signer is used for all transitions.
+        if let Some(first_transition) = execution.transitions().next() {
```
raychu86: minor comment - Any cleaner way to do this? Maybe a `try_fold` or some sort of `.all` call without having to do the `.next()`

synthesizer/process/src/stack/execute.rs
vicsn: 3f0e2d0c9

synthesizer/process/src/verify_execution.rs
vicsn: 7ec8642b5

PR #2081, opened 2023-10-12
Capture the full set of finalize operations, track aborted transactions

### LINKED ISSUE #2084: [Bug] Rejected transactions don't have `finalize` object
## 🐛 Bug Report

`AcceptedExecute` and `AcceptedDeploy` have `finalize` which stores all `FinalizeOperations`, but `RejectedExecute` and `RejectedDeploy` currently don't have that field.

That field was indeed not required before; however, we now have `public_fee`, so even if a transaction is rejected, there might still be a finalize operation to execute to remove the fee from the `account` mapping.

By the way, finalize operations created by `public_fee` are stored in the `finalize` field for accepted transactions.

I assumed the fee is indeed paid. Didn't check if this is actually the case.

## Steps to Reproduce

Block 41140 has a rejected transaction which has paid the fee with `public_fee`.

## Expected Behavior

All finalize operations created by transactions should be recorded in the block.

## Your Environment

- 
- 
-

howardwu:
## Motivation

This PR introduces a number of safety improvements to snarkVM.

This PR captures the full set of finalize operations and tracks aborted transactions in each block now.

In addition, there was no concept of a ratification ID, which will be problematic for Narwhal when these are supported in the future. This PR introduces `ratification_id` with the prefix `ar` (See breaking changes below for the migration process).

## Breaking Changes

The following is a list of breaking changes in this PR:
- Serialization of `Block`
    - `coinbase` has been renamed to `solutions`
    - `transactions` has been moved from after `authority` to after `solutions`
    - The prefix for a state root has changed from `ar` to `sr`
        - This allows `ratification_id` to take on the `ar` prefix.
    - The prefix for a transition ID has changed from `as` to `au`
        - This is to prepare for the future `solution_id` to take on `as`.

Note: This PR requires resampling.

Closes #2084


ledger/block/src/transactions/merkle.rs
```diff
@@ -16,9 +16,9 @@ use super::*;
 
 impl<N: Network> Transactions<N> {
     /// Returns the finalize root of the transactions.
-    pub fn to_finalize_root(&self) -> Result<Field<N>> {
+    pub fn to_finalize_root(&self, ratify_finalize_operations: Vec<FinalizeOperation<N>>) -> Result<Field<N>> {
```
ljedrz:
```suggestion
    pub fn to_finalize_root(&self, ratify_finalize_operations: &[FinalizeOperation<N>]) -> Result<Field<N>> {
```
that, or even `I: Iterator<Item = &FinalizeOperation<N>>`

PR #2093, opened 2023-10-15
Optimize some group circuits.

acoglio:
Refactor `Group::new` and `Group::from_xy_coordinates` to save constraints and increase clarity. The enforcement of a point being in the (sub)group is now delegated to a new circuit that does just that, and uses another new circuit to enforce that a point is twice another (this differs from the circuit that doubles a point).

This code has more separation between witness computation (console world) and circuit synthesis (circuit world), which is necessary to achieve some optimizations in the circuits.

~~There is a TODO for the witness computation in `Group::new`: there must be already code to check whether a pair of coordinates is in the (sub)group. This TODO is just at the console level; the circuit enforcement is already done in this commit.~~


acoglio:
I got a pre-commit hook error that seems unrelated to my changes, so I'm marking this PR as ready.

acoglio:
This PR supersedes https://github.com/AleoHQ/snarkVM/pull/2072.

howardwu (approved):

PR #2094, opened 2023-10-15
[Fix] Prevent passing in too many public inputs to Varuna

howardwu:
## Motivation

```
cd synthesizer/snark
cargo test test_varuna_verify_public_input_size --lib
```


vicsn:
Some notes about the fixes:
- handling errors in nested closures is very annoying, so I replaced some with simple for-loops
~- we don't use `ensure!` anywhere yet in `varuna.rs`, so followed the existing setup when throwing a `SNARKError::PublicInputSizeMismatch`~

algorithms/src/snark/varuna/tests.rs
```diff
@@ -76,7 +76,7 @@ mod varuna {
                         assert!($snark_inst::verify(universal_verifier, &fs_parameters, &index_vk, public_inputs, &proof).unwrap());
                         println!("Called verifier");
                         eprintln!("\nShould not verify (i.e. verifier messages should print below):");
-                        assert!(!$snark_inst::verify(universal_verifier, &fs_parameters, &index_vk, [random, random], &proof).unwrap());
+                        assert!(!$snark_inst::verify(universal_verifier, &fs_parameters, &index_vk, [random], &proof).unwrap());
```
howardwu: Why was this changed?

algorithms/src/snark/varuna/varuna.rs
```diff
@@ -670,7 +670,13 @@ where
 
             let input_fields = public_inputs_i
                 .iter()
-                .map(|input| input.borrow().to_field_elements())
+                .map(|input| {
+                    let input = input.borrow().to_field_elements()?;
+                    if input.len() > input_domain.size() - 1 {
```
ljedrz: is input domain size guaranteed to be non-zero?

algorithms/src/snark/varuna/ahp/ahp.rs
```diff
@@ -180,23 +180,20 @@ impl<F: PrimeField, SM: SNARKMode> AHPForR1CS<F, SM> {
...
+            let input_domain = circuit_state.input_domain;
+            let public_inputs_i = public_inputs[circuit_id]
+                .iter()
+                .map(|p| {
+                    let public_input = prover::ConstraintSystem::format_public_input(p);
+                    Self::formatted_public_input_is_admissible(&public_input)?;
+                    Ok::<_, AHPError>(public_input)
+                })
+                .collect::<Result<Vec<_>, _>>()?;
+            assert_eq!(public_inputs_i[0].len(), input_domain.size());
```
ljedrz: shouldn't this remain `ensure`, since we already have error handling available here?

ljedrz (approved):
LGTM with 2 questions addressed :+1:

algorithms/src/snark/varuna/tests.rs
vicsn: The new test and code introduced in this PR succesfully prevents too many inputs from being passed in. We used to pass in two public inputs here, even though the circuit only expects one public input.

For extra context: this is a negative test, so we're passing a random public input and expect it to fail.

algorithms/src/snark/varuna/ahp/ahp.rs
vicsn: Good catch: [Use ensure instead of assert_eq](https://github.com/AleoHQ/snarkVM/pull/2094/commits/e0ea4593decc3f608e015c283222676210f63bad)

algorithms/src/snark/varuna/varuna.rs
vicsn: This let me down an interesting rabbit hole. Yes it is guaranteed to be non-zero because we're always passing in a public input. And the - 1 is no longer needed in this somewhat related PR: https://github.com/AleoHQ/snarkVM/pull/2268

howardwu:
@vicsn is this PR still relevant?

vicsn:
> @vicsn is this PR still relevant?

@howardwu yes it is, made the PR title more descriptive.

algorithms/src/snark/varuna/varuna.rs
howardwu: So is the -1 still needed here?

vicsn: Nope, merged in the `mainnet` branch and resolved the conflict.

howardwu:
LGTM

PR #2099, opened 2023-10-17
Enable deferred deserialization of Data

ljedrz:
I haven't noticed it when we switched the snarkOS `Data` to the one from snarkVM, but it appears that the latter had a slightly different implementation that by default doesn't enable deferred deserialization. This PR fixes it by dropping the variant (which would almost exclusively end up being `0`) and making `FromBytes::read_le` always return a buffer.


ljedrz:
The CI failure is non-critical, but we might want to remove that one check now that there is no variant.

howardwu:
Keep the variant

howardwu (approved):

PR #2100, opened 2023-10-17
[Optimize] Assert bits are zero.

d0cd:
This PR implements a bitwise zero check in 1 constraint.

### DIFF
--- a/circuit/types/boolean/src/helpers/bits_are_zero.rs
+++ b/circuit/types/boolean/src/helpers/bits_are_zero.rs
@@ -14,9 +14,14 @@
 
 use super::*;
 
-pub mod adder;
-pub mod bits_are_zero;
-pub mod comparator;
-pub mod from_bits;
-pub mod subtractor;
-pub mod to_bits;
+impl<E: Environment> Boolean<E> {
+    /// Asserts that all bits in `bits_le` are zero.
+    #[doc(hidden)]
+    pub fn assert_bits_are_zero(bits_le: &[Boolean<E>]) {
+        let mut sum = Self::constant(false).0;
+        for bit in bits_le {
+            sum += &**bit;
+        }
+        E::assert_eq(sum, E::zero());
+    }
+}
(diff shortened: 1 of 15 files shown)


acoglio (approved):
This works not only when the bits are variables
`(v1 + v2 + ...) (1) = (0)`
but also when they are linear combinations (satisfying certain invariants), e.g.
`((1 - x) + y) (1) = (0)`
forces `1 - x = 0` and `y = 0`, where `1 - x` could be the result of negating the bit `x` (so the constraint forces `x = 1`); this relies on the invariant that `x` is either 0 or 1 (otherwise things may wrap around in the field).

acoglio:
It occurs to me that this also relies on not there being too many bits, where 'too many' means p bits where p is the prime of the field, otherwise they could all be 1 and add up to 0 mod p. Not sure if we should check for this condition or not.

circuit/types/boolean/src/helpers/bits_are_zero.rs
```diff
@@ -0,0 +1,27 @@
...
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+use super::*;
+
+impl<E: Environment> Boolean<E> {
+    /// Asserts that all bits in `bits_le` are zero.
+    #[doc(hidden)]
+    pub fn assert_bits_are_zero(bits_le: &[Boolean<E>]) {
+        let mut sum = Self::constant(false).0;
```
bendyarm, line 21: I'm wondering what would happen if a smaller field prime were used, like for a different proving system.
Is there an easy way to check the size of `bits_le` outside the circuit to make sure it is less
than the field prime?   If not, just ignore this comment.

bendyarm (approved):
LGTM, and saves lots of constraints!

circuit/types/boolean/src/helpers/bits_are_zero.rs
bendyarm: It looks like @acoglio had the same idea!

PR #2101, opened 2023-10-17
Update transmission id comparison 

raychu86:
## Motivation

This PR updates the transmission id checks in `check_subdag_transmissions`.


ledger/block/src/verify.rs
```diff
@@ -92,7 +92,10 @@ impl<N: Network> Block<N> {
             expected_last_coinbase_timestamp,
             expected_timestamp,
             current_timestamp,
-        )
+        )?;
+
+        // Re
```
raychu86:
```suggestion
        // Return the expected existing transaction ids.
```

howardwu (approved):

PR #2108, opened 2023-10-19
Code to explore checking if x in group.

acoglio:


howardwu:
Resolved using a different PR with a direct check that is optimized and formally verified.

PR #2110, opened 2023-10-20
Only declare the public 1 variable once

vicsn:
## Motivation

Our constraint system has not just one but a superfluous second public 1 variable. Because the R1CS spits it out, and the ConstraintSystem subsequently already has one hardcoded. However this never materially impacted proving or verifying. 

[The prover hardcodes it here](https://github.com/AleoHQ/snarkVM/blob/testnet3/circuit/environment/src/helpers/r1cs.rs#L40).
[The verifier hardcodes it here](https://github.com/AleoHQ/snarkVM/blob/declare_public_once/algorithms/src/snark/varuna/varuna.rs#L682).
[The verifier also hardcodes it here](https://github.com/AleoHQ/snarkVM/blob/testnet3/synthesizer/process/src/verify_execution.rs#L176).

## Test Plan

Requires resampling

### DIFF
--- a/algorithms/src/snark/varuna/ahp/indexer/constraint_system.rs
+++ b/algorithms/src/snark/varuna/ahp/indexer/constraint_system.rs
@@ -38,7 +38,7 @@ impl<F: Field> ConstraintSystem<F> {
             a: Vec::new(),
             b: Vec::new(),
             c: Vec::new(),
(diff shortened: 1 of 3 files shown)


vicsn:
Tests are failing even after resampling. Taking another look at this at a later point.

PR #2111, opened 2023-10-20
R1CS is only passing when all constraints refer to the same variables

vicsn:
TL;DR: `is_satisfied()` was only checking local correctness of the constraints, not global correctness.

A satisfying assignment to R1CS consists of a list of variables which each have a specific value. However, our R1CS duplicates the tracking of variables:
1. full list of private/public variables
2. each constraint again indicates which variables they use

If a user makes non-black-box use of the snarkVM compiler and decides to remove some declared variables, inconsistencies can occur. The check added in this PR should catch that.

We could refactor `R1CS::enforce(...)` to check for each incoming constraint if its variables are consistent with declared variables, but this would be very inefficient (or requires us to refactor the R1CS struct to store its variables in a mapping instead of a vector). But again, given that black-box use of snarkVM won't hit this case I think we can leave this out.


circuit/environment/src/helpers/r1cs.rs
```diff
@@ -91,9 +91,22 @@ impl<F: PrimeField> R1CS<F> {
...
     }
 
-    /// Returns `true` if all constraints in the environment are satisfied.
+    /// Returns `true` if both
+    /// - all constraints in the environment are satisfied.
+    /// - all constraints use variables corresponding to the declared variables
     pub fn is_satisfied(&self) -> bool {
-        self.constraints.iter().all(|constraint| constraint.is_satisfied())
+        let constraints_satisfied = self.constraints.iter().all(|constraint| constraint.is_satisfied());
+        let vars_consistent = self.constraints.iter().all(|constraint| {
```
ljedrz:
```suggestion
        let constraints_satisfied = self.constraints.iter().all(|constraint| constraint.is_satisfied());
        if !constraints_satisfied {
            return false;
        }
        let vars_consistent = self.constraints.iter().all(|constraint| {
```
to avoid extra work if the first check is `false`

circuit/environment/src/helpers/r1cs.rs
```diff
@@ -91,9 +91,22 @@ impl<F: PrimeField> R1CS<F> {
...
+            let (a, b, c) = constraint.to_terms();
+            [a, b, c].into_iter().all(|m| {
+                m.to_terms().into_iter().all(|(var, _)| match var {
+                    Variable::Constant(_value) => false, // terms should only contain private/public
+                    Variable::Private(index, value) => self.private[*index as usize].value() == **value,
+                    Variable::Public(index, value) => self.public[*index as usize].value() == **value,
```
ljedrz: I'm assuming this indexing is infallible?

circuit/environment/src/helpers/r1cs.rs
```diff
@@ -91,9 +91,22 @@ impl<F: PrimeField> R1CS<F> {
...
+                })
+            })
+        });
+        constraints_satisfied && vars_consistent
```
ljedrz: if we go with the early return, there is no need for the `vars_consistent` binding or this final line

ljedrz (commented):
Reviewed for performance & basic safety :+1:.

vicsn:
> Reviewed for performance & basic safety 👍.

Great points, addressed and squashed

circuit/environment/src/helpers/r1cs.rs
```diff
@@ -91,9 +93,33 @@ impl<F: PrimeField> R1CS<F> {
         self.counter.add_constraint(constraint);
     }
 
-    /// Returns `true` if all constraints in the environment are satisfied.
-    pub fn is_satisfied(&self) -> bool {
-        self.constraints.iter().all(|constraint| constraint.is_satisfied())
+    /// Returns `true` if both
+    /// - all constraints in the environment are satisfied.
+    /// - all constraints use variables corresponding to the declared variables
+    pub fn is_satisfied(&self) -> Result<bool> {
```
howardwu: We should return `bool` and use `.all()` with a `.map_or_else` or equiv.

vicsn: 🙏 Fixed, squashed and pushed.

raychu86 (approved):

howardwu (approved):
LGTM

PR #2123, opened 2023-10-24
Introduce `BFTStore` and `TransmissionStore`

raychu86:
## Motivation

This PR implements a `TransmissionStore` to store a list of transmissions based on the current round. This will help replace the in-memory tracking of the transmissions, which could lead to BFT issues when nodes reboot. These transmissions can be GC'd later to reduce state bloat.

Note: This will break current the database structure, and we may need to move the `MapID` ordering for the `TransmissionMap` enum.

TODO:
- [ ] Add the Transmission store to the `Ledger` or `ConsensusStore`


raychu86:
Closing with https://github.com/AleoNet/snarkOS/pull/2838

PR #2124, opened 2023-10-24
Allow programs to import `credits.aleo` when using `snarkvm execute`

### LINKED ISSUE #2122: [Bug] Unable to use `snarkvm execute` on a program that imports `credits.aleo`
## 🐛 Bug Report
Each `Process` is setup to automatically include a `credits.aleo` program stack. This is problematic as users are unable to write programs importing `credits.aleo`:
- If they include `import credits.aleo` in their `.aleo` file and add a corresponding import file for `credits.aleo` then it will pass parsing, but fail during process creation as program stacks are enforced to be unique. 
- If they don't include `import credits.aleo` then parsing will fail. 

## Steps to Reproduce
```
import credits.aleo;
program explore.aleo;

function main:
    input r0 as u32.public;
    input r1 as u32.private;
    add r0 r1 into r2;
    output r2 as u32.private;
```

1. Create directory `imports` with copy of `credits.aleo`
2. `snarkvm execute main`

## Expected Behavior
Result is ``⚠️  Program 'credits.aleo' already exists``. But this is a bug as there should be some way for users to use `snarkvm execute` on programs that import `credits.aleo`.

## Environment
- snarkVM Version: 0.16.2
- Rust Version: 1.73.0
- Computer OS: MacOS 13.2.1

evan-schott:
## Motivation
Resolves #2122

## Test Plan
Created a test that imports `credits.aleo` and now successfully builds without causing a duplicate name error.

### DIFF
--- a/synthesizer/process/src/lib.rs
+++ b/synthesizer/process/src/lib.rs
@@ -115,8 +115,12 @@ impl<N: Network> Process<N> {
     /// If you intend to `execute` the program, use `deploy` and `finalize_deployment` instead.
     #[inline]
     pub fn add_program(&mut self, program: &Program<N>) -> Result<()> {
-        // Compute the program stack, and add it to the process.
-        self.add_stack(Stack::new(self, program)?);
+        // Initialize the 'credits.aleo' program ID.
+        let credits_program_id = ProgramID::<N>::from_str("credits.aleo")?;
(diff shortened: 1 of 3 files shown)


evan-schott:
The check-fmt test failure is not related to any file I changed

howardwu (approved):

raychu86 (approved):

PR #2127, opened 2023-10-25
Fix genesis sampling for non-debug_assertions scenarios

ljedrz:
In addition, the existing 2 checks for genesis ratifications are merged into 1 and the related error message is fixed.

If there's a better ratification and finalization to use, I'm happy to adjust the PR.


ljedrz:
Drafting, as it's tricky to set it up correctly across all the crates.

PR #2129, opened 2023-10-26
Fix `contains_transaction_id` check

raychu86:
## Motivation

This PR fixes the `contains_transaction_id` check, and cleans up the impl by adding an abstraction into `BlockStore`.


howardwu (approved):
LGTM

PR #2133, opened 2023-10-26
[Fix] Fix circuit for cast lossy field to group.

acoglio:
This resolves the underconstraining bug that was found in that circuit.

The fix involves the creation of two new circuits, which are more general and may have other uses:
- A circuit for square roots flagged nondeterministic, which return both square roots, in no specified order (hence 'nondeterministic'), if the input is a non-zero square, and returns an error flag set if the input is not a square.
- A circuit for turning an x coordinate into the (unique, if it exists) point in the subgroup with that x coordinate, with an error flag set if there is no such point in the subgroup. This circuit makes use of the previous one to solve the curve equation, i.e. to find whether and where x intersects the elliptic curve.

The new circuit for cast lossy field to group is realized by using the latter circuit to attempt to find the point in the subgroup with the given value as x coordinate. If none exists, we return the generator if the input is 0, otherwise the result of Elligator-2 is returned, which is always a subgroup point.

The use of the new circuits for flagged operations eliminate the underconstraining without eliminating desired solutions (which would happen by enforcing constraints without taking flags into account).

This PR is draft because I still need to add tests for the two new circuits, but the implementation can be already reviewed.


acoglio:
As additional "testing", I inspected part of the constraints in the samples (generated by @bendyarm) of these new circuits, and they look reasonable to me. I only inspected part of the constraints because there are literally thousands. Eventually we'll check all of them mechanically in the theorem prover, but I wanted to do a preliminary sanity checks on these new and slightly complicated circuits.

circuit/types/group/src/helpers/from_x_coordinate.rs
```diff
@@ -26,6 +26,76 @@ impl<E: Environment> Group<E> {
...
+        let order = E::ScalarField::modulus();
+        let order_bits_be = order.to_bits_be();
+        let mut order_bits_be_constants = Vec::new();
+        for bit in order_bits_be.iter() {
+            order_bits_be_constants.push(Boolean::constant(*bit));
+        }
+        let point1_times_order = order_bits_be_constants.mul(point1);
+        let point2_times_order = order_bits_be_constants.mul(point2);
+        let point1_is_in_subgroup = point1_times_order.is_equal(&Self::zero());
+        let point2_is_in_subgroup = point2_times_order.is_equal(&Self::zero());
```
howardwu: Can you use `is_zero()`?

circuit/types/group/src/helpers/from_x_coordinate.rs
```diff
@@ -26,6 +26,76 @@ impl<E: Environment> Group<E> {
...
+        // but ignored in the final selection of the results returned below.
+        // The criterion for membership in the subgroup is that
+        // multiplying the point by the subgroup order yields the zero point (0, 1).
+        // The group operation that we use here is for the type `Group<E>` of the subgroup,
+        // which as mentioned above it can be performed on points outside the subgroup as well.
+        // We turn the subgroup order into big endian bits,
+        // to get around the issue that the subgroup order is not of Scalar<E> type.
+        let order = E::ScalarField::modulus();
+        let order_bits_be = order.to_bits_be();
+        let mut order_bits_be_constants = Vec::new();
```
howardwu: This should be `Vec::with_capacity`.

@ljedrz is there any performance hit to doing it manually `for loop` vs `.iter().collect();`?

circuit/types/group/src/helpers/from_x_coordinate.rs
```diff
@@ -26,6 +26,76 @@ impl<E: Environment> Group<E> {
 
         Self::from_xy_coordinates(x, y)
     }
+
+    /// Initializes an affine group element from a given x-coordinate field element.
+    /// Also returns an error flag, set if there is no group element with the given x-coordinate;
+    /// in that case, the returned point is `(0, 0)`, but immaterial.
+    pub fn from_x_coordinate_flagged(x: Field<E>) -> (Boolean<E>, Self) {
```
howardwu: Please change the return type to `(Self, Boolean)`

circuit/types/field/src/square_root.rs
```diff
@@ -63,6 +63,69 @@ impl<E: Environment> Field<E> {
...
+    pub fn square_roots_flagged_nondeterministic(&self) -> (Boolean<E>, Self, Self) {
+        // Obtain (p-1)/2, as a constant field element.
+        let modulus_minus_one_div_two = match E::BaseField::from_bigint(E::BaseField::modulus_minus_one_div_two()) {
+            Some(modulus_minus_one_div_two) => Field::constant(console::Field::new(modulus_minus_one_div_two)),
+            None => E::halt("Failed to initialize (modulus - 1) / 2"),
+        };
+
+        // Use Euler's criterion: self is a non-zero square iff self^((p-1)/2) is 1.
+        let euler = self.pow(modulus_minus_one_div_two);
+        let is_nonzero_square = euler.is_equal(&Field::one());
```
howardwu: Use `is_one()`

circuit/types/field/src/square_root.rs
```diff
@@ -63,6 +63,69 @@ impl<E: Environment> Field<E> {
...
+        let first_root = Field::new(mode, root_witness);
+        E::enforce(|| (&first_root, &first_root, &square));
+
+        // The second root returned by this function is the negation of the first one.
+        // So if self is a non-zero square, this is always different from the first root,
+        // but in the circuit it can be either positive (and the other negative) or vice versa.
+        let second_root = first_root.clone().neg();
+
+        // The error flag is set iff self is a non-square, i.e. it is neither zero nor a non-zero square.
+        let is_nonzero = self.is_not_equal(&Field::zero());
```
howardwu: Use `!self.is_zero()`

circuit/types/field/src/square_root.rs
```diff
@@ -63,6 +63,69 @@ impl<E: Environment> Field<E> {
...
+    /// if `self` is 0, both field results are 0;
+    /// if `self` is not a square, both field results are 0, but immaterial.
+    ///
+    /// The 'nondeterministic' part of the function name refers to the synthesized circuit,
+    /// whose represented computation, unlike the console computation just described,
+    /// returns the two roots (if `self` is a non-zero square) in no specified order.
+    /// This nondeterminism saves constraints, but generally this circuit should be only used
+    /// as part of larger circuits for which the nondeterminism in the order of the two roots does not matter,
+    /// and where the larger circuits represent deterministic computations despite this internal nondeterminism.
+    pub fn square_roots_flagged_nondeterministic(&self) -> (Boolean<E>, Self, Self) {
```
howardwu: Change return type to `Self, Self, Boolean`

circuit/types/group/src/helpers/from_x_coordinate.rs
```diff
@@ -26,6 +26,76 @@ impl<E: Environment> Group<E> {
...
+        // We turn the subgroup order into big endian bits,
+        // to get around the issue that the subgroup order is not of Scalar<E> type.
+        let order = E::ScalarField::modulus();
+        let order_bits_be = order.to_bits_be();
+        let mut order_bits_be_constants = Vec::new();
+        for bit in order_bits_be.iter() {
+            order_bits_be_constants.push(Boolean::constant(*bit));
+        }
+        let point1_times_order = order_bits_be_constants.mul(point1);
+        let point2_times_order = order_bits_be_constants.mul(point2);
```
howardwu: Can we implement a version of `Group::enforce_in_group` as `Group::is_in_group` that returns a boolean?

circuit/types/group/src/helpers/from_x_coordinate.rs
acoglio: Changed.

circuit/types/group/src/helpers/from_x_coordinate.rs
acoglio: Not that I know. If we impose the enforce_in_group constraints directly on the point (x, y) to check, clearly we get unsatisfiable constraints when (x, y) is not in the group. If we postulate a new/separate point (x', y') in the group (using the enforce_in_group constraints), the trouble is how to enforce the right connection between (x, y), (x', y'), and the boolean result r. Just saying that (r = 1) <==> (x = x' & y = y') is underconstrained, because even if (x, y) is in the group, I can pick a different (x', y') also in the group (because the group has more than one point) and set r = 0 (another possible solution being x = x' and y = y' and r = 1 of course).

acoglio: I'm not saying it's impossible. Maybe there's something clever, but if there is it has not occurred to any of us yet.

acoglio:
@howardwu The CI failure may be due to a change in the constraint count:
```
thread 'transaction::serialize::tests::test_bincode' panicked at 'called `Result::unwrap()` on an `Err` value: Expected size of 12582920, found size of 573440', ledger/block/src/transaction/fee/mod.rs:256:76
```
But I'm not seeing explicit numbers to change in that code; presumably they are calculated, but I'm not familiar with that part of the code.

howardwu:
The CI failure is just an S3 failure. I've triggered a re-run of CI

circuit/types/group/src/helpers/from_x_coordinate.rs
howardwu: Looking at `enforce_in_group` it seems like it's possible to change the `enforce_*` into `is_*`, but I don't know what the issue is.

circuit/types/group/src/helpers/from_x_coordinate.rs
ljedrz: if we know the length in advance, a `Vec::with_capacity` and a `for` loop is a safer bet for performance than `collect` (but it should be just as fast most of the time)

acoglio:
Re-targeted to `mainnet` branch as requested by @d0cd.

circuit/types/group/src/helpers/from_x_coordinate.rs
acoglio: It looks like this could be marked resolved then.

circuit/types/group/src/helpers/from_x_coordinate.rs
acoglio: I looked this morning at https://github.com/AleoHQ/snarkVM/compare/fix/cast-lossy-field-to-group...feat/is-in-group (which does that) but I see an underconstraining problem similar to above. I'll DM-forward you the details.

d0cd (approved):
Looks good to me! We should make a note to investigate improvements to the number of constraint counts required for `is_in_group`.

circuit/types/field/src/square_root.rs
```diff
@@ -64,22 +64,23 @@ impl<E: Environment> Field<E> {
...
     ///
-    /// In the console computation:
-    /// if `self` is a non-zero square,
-    /// the first field result is the positive root (i.e. closer to 0)
-    /// and the second field result is the negative root (i.e. closer to the prime);
-    /// if `self` is 0, both field results are 0;
-    /// if `self` is not a square, both field results are 0, but immaterial.
+    /// If `self` is a non-zero square,
+    ///  - the first field result is the positive root (i.e. closer to 0)
+    ///  - the second field result is the negative root (i.e. closer to the prime)
```
bendyarm, line 71:
```suggestion
    ///  - one of the field results is the positive root (i.e. closer to 0)
    ///  - the other field result is the negative root (i.e. closer to the prime)

File: circuit/types/field/src/square_root.rs [hunk 9]
```diff
@@ -63,6 +63,65 @@ impl<E: Environment> Field<E> {
...
+    ///
+    /// If `self` is 0,
+    ///  - both field results are 0
+    ///  - the flag is 0
+    ///
+    /// If `self` is not a square,
+    ///  - both field results are 0
+    ///  - the flag is 1
+    ///
+    /// Note that the constraints do **not** impose an ordering on the two roots returned by this function;
```
acoglio, line 82:
```suggestion
    /// Note that the input/output behavior just described refers to this Rust function,
    /// but the generated constraints do **not** impose an ordering on the two roots;
```

circuit/types/field/src/square_root.rs
```diff
@@ -63,6 +63,65 @@ impl<E: Environment> Field<E> {
     }
 }
 
+impl<E: Environment> Field<E> {
+    /// Returns both square roots of `self` and a `Boolean` flag, which is set iff `self` is not a square.
+    ///
+    /// If `self` is a non-zero square,
+    ///  - the first field result is the positive root (i.e. closer to 0)
+    ///  - the second field result is the negative root (i.e. closer to the prime)
+    ///  - the flag is 0
```
bendyarm, line 72:
```suggestion
    ///  - the first field result is the witnessed and checked square root
    ///  - the second field result is the negation of the first result
    ///  - the flag is 0.
    /// Since the standard console square_root function returns the positive square root
    /// (i.e. closer to 0), normally the first result would be less than the second result.
    /// However, that is not checked here and it would be a mistake to rely on the ordering.
```

circuit/types/field/src/square_root.rs
```diff
@@ -63,6 +63,65 @@ impl<E: Environment> Field<E> {
     }
 }
 
+impl<E: Environment> Field<E> {
+    /// Returns both square roots of `self` and a `Boolean` flag, which is set iff `self` is not a square.
```
acoglio, line 67:
```suggestion
    /// Returns both square roots of `self` and a `Boolean` flag, which is set iff `self` is not a square.
    /// The roots are returned in the order specified below by this function,
    /// but the generated constraints do not impose an order
    /// (this is where the `nondeterministic` part of this function name refers to).
```

bendyarm (approved):

PR #2145, opened 2023-11-02
Carry along an updateable universal_prover during synthesis and execution

vicsn:
## Motivation

Howard noticed that [in the previous PR](https://github.com/AleoHQ/snarkVM/pull/1955) certain tests would end up loading the SRS twice: first during synthesis and then during execution. To avoid this, this PR proposes carrying around the trimmed SRS (`UniversalProver::committer_key `) and updating it as needed.

Notes:
- `trim(...)` is renamed `update(...)`, to reflect that we can shrink or grow the trimmed SRS as needed. The current implementation of this update function is not atomic. If it fails halfway through, we might have an inconsistent `UniversalProver`. It seems ok to me to always expect that we first need to succesfully update before creating a new proof. Though this might not play nicely with parallelization.
- To avoid generating the `universalProver` every time we synthesizes proving keys ánd every time we prove and execution, I turned the UniversalProver into a global object in `Process`. Individual threads can acquire a write lock to use it for their specific proof. Depending on the context, this might incur a significant slowdown if we want to synthesize and prove in parallel. However, I reckon parallel proving within a single `Process` is not a usecase we're aiming to support at the moment anyway.
- We take one power of two less for the SRS, it appeared that it was superfluous.
- For a large circuit (say, 2^25), `reduce_pk_size_{2,3}` reduce proving time by ~8% on a 16 CPU, 128GB RAM machine.

## Test Plan

I manually ran `test_multiple_deployments_and_multiple_executions` to see if we load the expected amount of powers.


algorithms/benches/snark/varuna.rs
```diff
@@ -53,7 +60,10 @@ fn snark_circuit_setup(c: &mut Criterion) {
         let mul_depth = 1;
         let (circuit, _) = TestCircuit::gen_rand(mul_depth, num_constraints, num_variables, rng);
         c.bench_function(&format!("snark_circuit_setup_{size}"), |b| {
-            b.iter(|| VarunaInst::circuit_setup(&universal_srs, &circuit).unwrap())
+            b.iter(|| {
+                let mut universal_prover = UniversalProver::default();
```
ljedrz, line 64: do these benchmark changes make their results more stable? https://github.com/AleoHQ/snarkVM/pull/2135 is related, but there is still more variance than what I'd like

algorithms/src/polycommit/kzg10/degree_info.rs
```diff
@@ -12,31 +12,30 @@
...
+#[derive(Clone, Debug, Default)]
 pub struct DegreeInfo {
     /// The maximum degree of the required SRS to commit to the polynomials.
     pub max_degree: usize,
     /// The maximum IOP poly degree used for (i)fft_precomputation.
     pub max_fft_size: usize,
-    /// TODO (howardwu): This should never be happening... Use a BTreeSet.
     /// The degree bounds on IOP polynomials.
-    pub degree_bounds: Option<HashSet<usize>>,
+    pub degree_bounds: Option<BTreeSet<usize>>,
```
ljedrz, line 24: do we take advantage of the element order now? we use many `BTree*` collections these days, but `Hash*` is preferable if order is of no significance

synthesizer/snark/src/certificate/mod.rs
```diff
@@ -41,9 +43,11 @@ impl<N: Network> Certificate<N> {
 
         // Retrieve the proving parameters.
         let degree_info = proving_key.circuit.index_info.degree_info::<N::Field, varuna::VarunaHidingMode>();
-        let universal_prover = N::varuna_universal_prover(degree_info);
         let fiat_shamir = N::varuna_fs_parameters();
 
+        let mut universal_prover = universal_prover.write();
+        universal_prover.update(srs, degree_info)?;
```
ljedrz:
```suggestion
        universal_prover.write().update(srs, degree_info)?;
```

ljedrz: unless it should not be written to OR read from until the end of the method

synthesizer/snark/src/proving_key/mod.rs
```diff
@@ -44,9 +47,13 @@ impl<N: Network> ProvingKey<N> {
 
         // Retrieve the proving parameters.
         let degree_info = self.circuit.index_info.degree_info::<N::Field, varuna::VarunaHidingMode>();
-        let universal_prover = N::varuna_universal_prover(degree_info);
+
         let fiat_shamir = N::varuna_fs_parameters();
 
+        // Update the universal_prover and prevent other threads from writing to it
+        let mut universal_prover = universal_prover.write();
+        universal_prover.update(srs, degree_info)?;
```
ljedrz: ditto

synthesizer/snark/src/universal_srs.rs
```diff
@@ -29,13 +29,16 @@ impl<N: Network> UniversalSRS<N> {
...
         function_name: &str,
         assignment: &circuit::Assignment<N::Field>,
     ) -> Result<(ProvingKey<N>, VerifyingKey<N>)> {
         #[cfg(feature = "aleo-cli")]
         let timer = std::time::Instant::now();
 
-        let (proving_key, verifying_key) = Varuna::<N>::circuit_setup(self, assignment)?;
+        // Put a write lock on the universal_prover
+        let mut universal_prover = universal_prover.write();
+        let (proving_key, verifying_key) = Varuna::<N>::circuit_setup(self.deref(), &mut universal_prover, assignment)?;
```
ljedrz: ditto

algorithms/src/polycommit/sonic_pc/data_structures.rs
```diff
@@ -265,6 +284,151 @@ impl<E: PairingEngine> ToBytes for CommitterKey<E> {
...
+                let mut shifted_powers_of_beta_times_gamma_g =
+                    self.shifted_powers_of_beta_times_gamma_g.take().unwrap_or_default();
+
+                let shifted_ck_time = start_timer!(|| "Constructing `shifted_powers_of_beta_g`");
+                match new_highest_degree_bound.cmp(&highest_degree_bound) {
+                    Ordering::Greater => {
+                        let lowest_shift_degree = max_degree - new_highest_degree_bound;
+                        let highest_shift_degree = max_degree + 1 - highest_degree_bound;
+                        let mut new_powers = srs.powers_of_beta_g(lowest_shift_degree, highest_shift_degree)?;
+                        new_powers.append(&mut shifted_powers_of_beta_g);
```
ljedrz:
```suggestion
                        new_powers.extend_from_slice(&shifted_powers_of_beta_g);
```
since we overwrite `shifted_powers_of_beta_g` right afterwards, there is no need to clear that vector; unless those elements are moved, in which case feel free to ignore this suggestion

ljedrz (commented):
Reviewed :ok_hand:.

algorithms/src/polycommit/kzg10/degree_info.rs
vicsn: Yes for this one we actually need to find the highest degree_bound to download a part of the SRS!

algorithms/benches/snark/varuna.rs
vicsn: This should not materially change this benchmark, other than that we might be doing less work (not downloading powers twice).

algorithms/src/polycommit/kzg10/degree_info.rs
ljedrz: if we only need the highest one, perhaps a `BinaryHeap` would suffice instead :detective:?

algorithms/src/polycommit/sonic_pc/data_structures.rs
vicsn: My rationale for using `append` was indeed that it moves elements, so I thought this would be faster + of course consumes less memory.

synthesizer/snark/src/certificate/mod.rs
vicsn: Good one 1d76c1aaa

synthesizer/snark/src/proving_key/mod.rs
vicsn: 1d76c1aaa

synthesizer/snark/src/universal_srs.rs
vicsn: 1d76c1aaa

algorithms/src/polycommit/kzg10/degree_info.rs
vicsn: TIL BinaryHeap.

We do need the other bounds as well - not necessarily in a specific order, so I guess BinaryHeap is slightly better. But the benefit would be marginal, under "normal" usage (proving a single function) this map has less than 10 elements.

ljedrz: cool; indeed, as long as the number of elements is small, it won't make a difference

algorithms/src/polycommit/sonic_pc/data_structures.rs
ljedrz: as long as the elements are not `Copy`, this is correct :+1:

vicsn: A good point. Just read that moves may be copies under the hood anyway, so here you go: f52cbd381

ljedrz: nice; also, a small addition to what I said before: even if the values were not `Copy`, `append` would not consume less memory, because the vector would still retain its allocated space unless `shrink_to_fit` was called afterwards

Lanckneht:
gm

PR #2147, opened 2023-11-03
[NCC GKE] Always trim trailing zeros after poly operation

vicsn:
## Motivation

We should be consistent about trimming trailing zeros. Such trailing zeros are essentially meaningless (`2x == 2x + 0x^2`) but could mess with other functions checking e.g. the degree of the polynomial based on the length of the coefficients vector.


algorithms/src/fft/polynomial/dense.rs
```diff
@@ -113,10 +112,16 @@ impl<F: Field> DensePolynomial<F> {
...
+    /// Outputs a univariate polynomial of degree `d` where each non-leading
+    /// coefficient is sampled uniformly at random from R and the leading
+    /// coefficient is sampled uniformly at random from among the non-zero
+    /// elements of R.
     pub fn rand<R: Rng>(d: usize, rng: &mut R) -> Self {
-        let random_coeffs = (0..(d + 1)).map(|_| F::rand(rng)).collect();
+        let mut random_coeffs = (0..(d + 1)).map(|_| F::rand(rng)).collect_vec();
+        while random_coeffs[d].is_zero() {
+            // In the extremely unlikely event, sample again.
+            random_coeffs = (0..(d + 1)).map(|_| F::rand(rng)).collect();
```
ljedrz: couldn't we just sample only element `d` here, if only that one is being checked?

ljedrz (commented):
Reviewed :ok_hand:.

algorithms/src/fft/polynomial/dense.rs
vicsn: Neat 9bffffde6

algorithms/src/fft/polynomial/dense.rs
```diff
@@ -197,6 +202,10 @@ impl<'a, 'b, F: Field> Add<&'a DensePolynomial<F>> for &'b DensePolynomial<F> {
             let mut result = self.clone();
             // Zip safety: `result` and `other` could have different lengths.
             cfg_iter_mut!(result.coeffs).zip(&other.coeffs).for_each(|(a, b)| *a += b);
+            // If the leading coefficient ends up being zero, pop it off.
+            while let Some(true) = self.coeffs.last().map(|c| c.is_zero()) {
```
howardwu: @vicsn this should have been `result.coeffs` not `self.coeffs`

howardwu: Fixed in 30aa354c4

howardwu (approved):

PR #2148, opened 2023-11-03
[NCC CDE] Add to reserved keywords

vicsn:
## Motivation

NCC noted we might be missing certain reserved keywords. Besides import, they also noticed  `bool`, `inline` and `transition` but those only exist in Leo. Let me know if you want me to add those.


howardwu (approved):

vicsn:
Double checking, did you read the PR message above?

howardwu:
We don't need to reserve `bool`, `inline`, or `transition` at this time.

PR #2151, opened 2023-11-03
[NCC FYM] Correctly panic when dividing by zero poly

vicsn:
## Motivation

We were allowing division by zero and returning zero. No bueno.


ljedrz (commented):
Could we add some test case that we could use to ensure that we maintain the expected behavior down the line?

vicsn:
> Could we add some test case that we could use to ensure that we maintain the expected behavior down the line?

Added

howardwu (approved):

PR #2152, opened 2023-11-03
[NCC 72V] Do not deserialize polys with trailing zero

vicsn:
## Motivation

Note that this might open up new attacks, but the best I can come up with is negligible: Imagine an app manages to let a user create a `ProvingKey`, which contains polys with trailing zeros, which can then not be deserialized anymore. But the user could just report the failing deserialization as a bug and trace how the trailing zero ended up there.

Crucially, verifiers never deal with polynomials, only commitments, so validators should be safe.


algorithms/src/fft/polynomial/mod.rs
```diff
@@ -67,7 +67,12 @@ impl<'a, F: Field> CanonicalSerialize for Polynomial<'a, F> {
 
 impl<'a, F: Field> Valid for Polynomial<'a, F> {
     fn check(&self) -> Result<(), SerializationError> {
-        Ok(())
+        // Fail if the trailing coefficient is zero
+        if let Some(false) = self.coeffs().last().map(|c| c.1.is_zero()) {
```
ljedrz:
```suggestion
        if let Some(true) = self.coeffs().last().map(|c| c.1.is_zero()) {
```
unless it's the bad sleep getting to me, shouldn't this be `true` instead?

vicsn: You are very right. Thank you for spotting.

howardwu (approved):

PR #2153, opened 2023-11-03
[NCC XL9] Make sure we loop independent of order in map

vicsn:
## Motivation

We need to always iterate over all values in the map. Previously, we were "lucky" that g_1 was the last element iterated over.


ljedrz (commented):
In a `BTreeMap`, `g_1` would always come before `g_a` etc., so unless we intended to always return early here, this is a desirable change :+1:.

howardwu (approved):

PR #2154, opened 2023-11-06
[NCC-Y7H] Create compute_function_id helper which enforces minimal TLV encoding

vicsn:
## Motivation

As defense in depth, in case we allow custom program_id's in the future, it is good practice to encode the length of each input. Some implementation notes:
- requires resampling
- we could also encode a "type" prefix as extra defense in depth, but the function_id code lives quite separate from other code. 
- I did not spot other invocations of `A::hash` which would warrant an extra TLV encoding. There might be a problem if end-users hash containers themselves (in Leo or Aleo instructions), but then it is up to them to implement TLV encoding.
- If this design looks good we could rename the `program/id` folder to `program/program_id` to distinguish it from `program/function_id`.
- Technically we don't have to pass in network_id when in console mode, as it can be derived from the type, but I do it anyway to be consistent with the Circuit implementation

## Test Plan

I did not add more tests.


synthesizer/process/src/verify_execution.rs
```diff
@@ -73,16 +73,10 @@ impl<N: Network> Process<N> {
...
-            let function_id = N::hash_bhp1024(
-                &(
-                    U16::<N>::new(N::ID),
-                    transition.program_id().name(),
-                    transition.program_id().network(),
-                    transition.function_name(),
-                )
-                    .to_bits_le(),
-            )?;
+            // Get the network id
```
howardwu:
```suggestion
            // Retrieve the network ID.
```

circuit/program/src/request/verify.rs
```diff
@@ -21,10 +21,8 @@ impl<A: Aleo> Request<A> {
     /// Verifies (challenge == challenge') && (address == address') && (serial_numbers == serial_numbers') where:
     ///     challenge' := HashToScalar(r * G, pk_sig, pr_sig, signer, \[tvk, tcm, function ID, input IDs\])
     pub fn verify(&self, input_types: &[console::ValueType<A::Network>], tpk: &Group<A>) -> Boolean<A> {
-        // Compute the function ID as `Hash(network_id, program_id, function_name)`.
-        let function_id = A::hash_bhp1024(
-            &(&self.network_id, self.program_id.name(), self.program_id.network(), &self.function_name).to_bits_le(),
-        );
+        // Compute the function ID
```
howardwu:
```suggestion
        // Compute the function ID.
```

console/program/src/request/sign.rs
```diff
@@ -64,10 +64,10 @@ impl<N: Network> Request<N> {
         // Compute the transition commitment `tcm` as `Hash(tvk)`.
         let tcm = N::hash_psd2(&[tvk])?;
 
-        // Compute the function ID as `Hash(network_id, program_id, function_name)`.
-        let function_id = N::hash_bhp1024(
-            &(U16::<N>::new(N::ID), program_id.name(), program_id.network(), function_name).to_bits_le(),
-        )?;
+        // Get the network id
+        let network_id = U16::new(N::ID);
+        // Compute the function id
```
howardwu:
```suggestion
        // Retrieve the network ID.
        let network_id = U16::new(N::ID);
        // Compute the function ID.
```

synthesizer/process/src/verify_fee.rs
```diff
@@ -76,11 +76,10 @@ impl<N: Network> Process<N> {
     fn verify_fee_private(&self, fee: &&Fee<N>) -> Result<()> {
         let timer = timer!("Process::verify_fee_private");
 
-        // Compute the function ID as `Hash(network_id, program_id, function_name)`.
-        let function_id = N::hash_bhp1024(
-            &(U16::<N>::new(N::ID), fee.program_id().name(), fee.program_id().network(), fee.function_name())
-                .to_bits_le(),
-        )?;
+        // Get the network id
```
howardwu:
```suggestion
        // Retrieve the network ID.
```

synthesizer/process/src/verify_fee.rs
```diff
@@ -146,11 +145,10 @@ impl<N: Network> Process<N> {
     fn verify_fee_public(&self, fee: &&Fee<N>) -> Result<()> {
         let timer = timer!("Process::verify_fee_public");
 
-        // Compute the function ID as `Hash(network_id, program_id, function_name)`.
-        let function_id = N::hash_bhp1024(
-            &(U16::<N>::new(N::ID), fee.program_id().name(), fee.program_id().network(), fee.function_name())
-                .to_bits_le(),
-        )?;
+        // Get the network id
```
howardwu:
```suggestion
        // Retrieve the network ID.
```

PR #2155, opened 2023-11-06
Simplify compress_elements

ljedrz:
This PR is a suggestion for further investigation, so I'm filing it as a draft. It builds on https://github.com/AleoHQ/snarkVM/pull/2150 with the final commit being new.

In `compress_elements`, which only ever gets called from `push_elements_to_sponge`, the second tuple members (`first.1` and `second.1`) are always `F::one`, which means that:
- we don't need to pass tuples from `push_elements_to_sponge` (only limbs)
- the `overhead` calculation only needs to be performed once

This is not a viable approach if we want `compress_elements` to be callable from outside of `push_elements_to_sponge` etc, but if this is not expected, it would be a simple win for perf and clarity.


ljedrz:
Superseded by https://github.com/ProvableHQ/snarkVM/pull/3351.

PR #2156, opened 2023-11-06
Make the tests more deterministic

ljedrz:
While I was comparing the number of allocations between different versions of some pieces of code, some of the results were a bit surprising, and upon closer inspection I noticed that there was some non-negligible variance between runs using the same code, even with a fixed main source of randomness. This led me to believe that we have some extra sources of randomness somewhere, which this PR addresses.

With these changes:
- more tests will be fully reproducible knowing a single seed; they should also be a bit faster
- benchmarks should be a bit faster too
- comparisons between profiles (both CPU and memory) will be easier

The one downside is that we would need to adjust bits of snarkOS as well, i.e. add one or more calls to `thread_rng` at the callsites that would now require the rng to be given. However, since we already do that for some of the commands, this shouldn't be controversial.


ljedrz:
The CI failure might be legitimate, I'll look into it tomorrow.

howardwu (approved):

PR #2158, opened 2023-11-06
[TOB] Guard overflow in finalize cost calculation

raychu86:
## Motivation

This PR updates the use of `.sum()` in the finalize cost calculator to a `try_fold` + `checked_add` approach that will catch overflows. 

This change was made to be extra cautious. Note that in the previous model, the maximum number of finalize operations at `u16::MAX` * most expensive finalize operation would never cause an overflow with `.sum()`.

Finding: TOB-ALEO-14

### DIFF
--- a/synthesizer/src/vm/helpers/cost.rs
+++ b/synthesizer/src/vm/helpers/cost.rs
@@ -202,6 +202,6 @@ pub fn cost_in_microcredits<N: Network>(finalize: &Finalize<N>) -> Result<u64> {
     finalize
         .commands()
         .iter()
-        .map(|command| cost(command))
+        .map(cost)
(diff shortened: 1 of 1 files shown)


synthesizer/src/vm/helpers/cost.rs
```diff
@@ -199,5 +199,9 @@ pub fn cost_in_microcredits<N: Network>(finalize: &Finalize<N>) -> Result<u64> {
         Command::BranchEq(_) | Command::BranchNeq(_) => Ok(5_000),
         Command::Position(_) => Ok(1_000),
     };
-    finalize.commands().iter().map(|command| cost(command)).sum()
+    finalize
+        .commands()
+        .iter()
+        .map(|command| cost(command))
```
howardwu:
```suggestion
        .map(cost)
```

howardwu (approved):

PR #2167, opened 2023-11-09
[TOB-1] Adds check for possible panics in FromBytes impls

raychu86:
## Motivation

This PR adds guards for possible denial of service vectors in `FromBytes` implementations.

Example:
```
let num_ratifications = u32::read_le(&mut reader)?;
let mut ratifications = Vec::with_capacity(num_ratifications as usize);
This will cause a panic in 32-bit architectures if num_ratifications exceeds i32::MAX.
```

Note: we are skipping some of the bound checks in `sonic_pc/data_structures` because it won't practically hit these larger values. 

Finding: TOB-ALEO-1

### DIFF
--- a/ledger/committee/src/bytes.rs
+++ b/ledger/committee/src/bytes.rs
@@ -28,6 +28,7 @@ impl<N: Network> FromBytes for Committee<N> {
         let starting_round = u64::read_le(&mut reader)?;
         // Read the number of members.
         let num_members = u16::read_le(&mut reader)?;
+        // Ensure the number of members is within the allowed limit.
         if num_members > Self::MAX_COMMITTEE_SIZE {
             return Err(error(format!(
                 "Committee cannot exceed {} members, found {num_members}",
--- a/ledger/committee/src/bytes.rs
+++ b/ledger/committee/src/bytes.rs
@@ -28,6 +28,12 @@ impl<N: Network> FromBytes for Committee<N> {
         let starting_round = u64::read_le(&mut reader)?;
         // Read the number of members.
         let num_members = u16::read_le(&mut reader)?;
+        if num_members > Self::MAX_COMMITTEE_SIZE {
+            return Err(error(format!(
+                "Committee cannot exceed {} members, found {num_members}",
+                Self::MAX_COMMITTEE_SIZE,
+            )));
+        }
         // Read the members.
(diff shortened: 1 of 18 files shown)


ledger/committee/src/bytes.rs
```diff
@@ -28,6 +28,12 @@ impl<N: Network> FromBytes for Committee<N> {
         let starting_round = u64::read_le(&mut reader)?;
         // Read the number of members.
         let num_members = u16::read_le(&mut reader)?;
+        if num_members > Self::MAX_COMMITTEE_SIZE {
```
howardwu, line 32:
```suggestion
        // Ensure the number of members is within the allowed limit.
        if num_members > Self::MAX_COMMITTEE_SIZE {
```

ledger/narwhal/batch-header/src/bytes.rs
```diff
@@ -34,18 +34,26 @@ impl<N: Network> FromBytes for BatchHeader<N> {
         let timestamp = i64::read_le(&mut reader)?;
 
         // Read the number of transmission IDs.
-        let num_transmissions = u32::read_le(&mut reader)?;
+        let num_transmission_ids = u32::read_le(&mut reader)?;
+        // Ensure the number of transmission ids is within bounds.
```
howardwu:
```suggestion
        // Ensure the number of transmission IDs is within bounds.
```

ledger/narwhal/batch-header/src/bytes.rs
```diff
@@ -34,18 +34,26 @@ impl<N: Network> FromBytes for BatchHeader<N> {
...
         let mut transmission_ids = IndexSet::new();
-        for _ in 0..num_transmissions {
+        for _ in 0..num_transmission_ids {
             // Insert the transmission ID.
             transmission_ids.insert(TransmissionID::read_le(&mut reader)?);
         }
 
         // Read the number of previous certificate IDs.
         let num_previous_certificate_ids = u32::read_le(&mut reader)?;
+        // Ensure the number of previous certificate ids is within bounds.
```
howardwu:
```suggestion
        // Ensure the number of previous certificate IDs is within bounds.
```

howardwu (approved):

PR #2172, opened 2023-11-15
Adds BFT map ID to storage

howardwu:
## Motivation

Adds BFT map ID to storage


raychu86 (approved):

Asbak21:
Adds BFT map ID to storage

PR #2176, opened 2023-11-16
Reduce overhead of programs with deep nested imports

vicsn:
## Motivation

When we deployed/executed a program with many nested imports, this requires *lots* of memory. Why did this happen for nested imports? Actually, it would have happened for any large amount of nested function calls, imports are the only way in which we support them.

Two optimizations:
- we were needlessly cloning trees of stacks, each of which contained all of the imported stacks as external_stacks.
- during CheckDeployment, the verifier was running `execute_function` instead of `evaluate_function` for each nested call, even though it only needs to compute the circuit for the top level circuit.

I would have assumed during Authorization/Synthesis we could similarly replace calls in Call to `execute_function` by `evaluate_function` too, but this fails in mysterious ways (after resampling, integer addition fails).

## Test Plan

Added a test deploying and executing a program with the maximum of 31 nested imports, which is our maximum.

Also manually tested four scenario's:
- proving execution: not killed by OOM
- proving deployment: not killed by OOM, for 23 nested deployments this went from 1550s to 60s
- verifying execution: 100ms for 31 nested deployments
- verifying deployment: now 300ms for 20 nested deployments, 400ms for 31 nested deployments

Fun fact, the synthesizer only lets me execute a function using 30 nested imports. Together with the base function and fee execution, this reaches our limit of 32 transitions per execution.

I did not extensively test `PackageRun`, I assume the existing test suite covers it.

### DIFF
--- a/synthesizer/process/src/lib.rs
+++ b/synthesizer/process/src/lib.rs
@@ -75,7 +75,7 @@ pub struct Process<N: Network> {
     /// The universal SRS.
     universal_srs: Arc<UniversalSRS<N>>,
     /// The mapping of program IDs to stacks.
-    stacks: IndexMap<ProgramID<N>, Stack<N>>,
+    stacks: IndexMap<ProgramID<N>, Arc<Stack<N>>>,
 }
 
 impl<N: Network> Process<N> {
@@ -129,7 +129,7 @@ impl<N: Network> Process<N> {
     #[inline]
     pub fn add_stack(&mut self, stack: Stack<N>) {
         // Add the stack to the process.
-        self.stacks.insert(*stack.program_id(), stack);
+        self.stacks.insert(*stack.program_id(), Arc::new(stack));
     }
 }
 
@@ -202,11 +202,24 @@ impl<N: Network> Process<N> {
 
     /// Returns the stack for the given program ID.
     #[inline]
-    pub fn get_stack(&self, program_id: impl TryInto<ProgramID<N>>) -> Result<&Stack<N>> {
+    pub fn get_stack(&self, program_id: impl TryInto<ProgramID<N>>) -> Result<Arc<Stack<N>>> {
         // Prepare the program ID.
         let program_id = program_id.try_into().map_err(|_| anyhow!("Invalid program ID"))?;
         // Retrieve the stack.
-        let stack = self.stacks.get(&program_id).ok_or_else(|| anyhow!("Program '{program_id}' does not exist"))?;
+        let stack = self.stacks.get(&program_id).cloned().ok_or_else(|| anyhow!("Program '{program_id}' does not exist"))?;
+        // Ensure the program ID matches.
+        ensure!(stack.program_id() == &program_id, "Expected program '{}', found '{program_id}'", stack.program_id());
+        // Return the stack.
+        Ok(stack)
+    }
+
+    /// Returns the stack for the given program ID.
+    #[inline]
+    pub fn get_stack_ref(&self, program_id: impl TryInto<ProgramID<N>>) -> Result<&Stack<N>> {
+        // Prepare the program ID.
+        let program_id = program_id.try_into().map_err(|_| anyhow!("Invalid program ID"))?;
+        // Retrieve the stack.
+        let stack = self.stacks.get(&program_id).ok_or_else(|| anyhow!("Program '{program_id}' does not exist"))?.deref();
         // Ensure the program ID matches.
         ensure!(stack.program_id() == &program_id, "Expected program '{}', found '{program_id}'", stack.program_id());
         // Return the stack.
@@ -216,7 +229,7 @@ impl<N: Network> Process<N> {
     /// Returns the program for the given program ID.
     #[inline]
     pub fn get_program(&self, program_id: impl TryInto<ProgramID<N>>) -> Result<&Program<N>> {
-        self.get_stack(program_id).map(Stack::program)
+        self.get_stack_ref(program_id).map(Stack::program)
     }
 
     /// Returns the proving key for the given program ID and function name.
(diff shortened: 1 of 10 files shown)


synthesizer/process/src/lib.rs
```diff
@@ -202,11 +202,26 @@ impl<N: Network> Process<N> {
...
+            self.stacks.get(&program_id).cloned().ok_or_else(|| anyhow!("Program '{program_id}' does not exist"))?;
+        // Ensure the program ID matches.
+        ensure!(stack.program_id() == &program_id, "Expected program '{}', found '{program_id}'", stack.program_id());
+        // Return the stack.
+        Ok(stack)
+    }
+
+    /// Returns the stack for the given program ID.
+    #[inline]
+    pub fn get_stack_ref(&self, program_id: impl TryInto<ProgramID<N>>) -> Result<&Stack<N>> {
```
d0cd: Could you use `get_stack_ref` in place of `get_stack` in other parts of the codebase?

synthesizer/process/src/stack/mod.rs
```diff
@@ -235,9 +235,25 @@ impl<N: Network> StackProgram<N> for Stack<N> {
...
-        self.external_stacks.get(program_id).ok_or_else(|| anyhow!("External program '{program_id}' does not exist."))
+        self.external_stacks
+            .get(program_id)
+            .cloned()
+            .ok_or_else(|| anyhow!("External program '{program_id}' does not exist."))
+    }
+
+    /// Returns the external stack ref for the given program ID.
+    #[inline]
+    fn get_external_stack_ref(&self, program_id: &ProgramID<N>) -> Result<&Stack<N>> {
```
d0cd: Could you use `get_external_stack_ref` in place of `get_external_stack` in other parts of the codebase?

synthesizer/process/src/stack/call/mod.rs
```diff
@@ -355,6 +373,9 @@ impl<N: Network> CallTrait<N> for Call<N> {
             // Assign the output to the register.
             registers.store_circuit(stack, register, output)?;
         }
+        lap!(timer, "assigned outputs to registers");
```
d0cd: Nit: Match formatting of timer messages with existing ones.

synthesizer/src/vm/mod.rs
```diff
@@ -605,7 +605,6 @@ function compute:
         // Construct the new block header.
         let (ratifications, transactions, aborted_transaction_ids, ratified_finalize_operations) =
             vm.speculate(sample_finalize_state(1), None, vec![], None, transactions.iter())?;
-        assert!(aborted_transaction_ids.is_empty());
```
d0cd:
```suggestion
        assert!(aborted_transaction_ids.is_empty());
```

d0cd: We can put this back in (I think I may have deleted it while writing the test)

synthesizer/process/src/lib.rs
```diff
@@ -202,11 +202,26 @@ impl<N: Network> Process<N> {
 
     /// Returns the stack for the given program ID.
     #[inline]
-    pub fn get_stack(&self, program_id: impl TryInto<ProgramID<N>>) -> Result<&Stack<N>> {
+    pub fn get_stack(&self, program_id: impl TryInto<ProgramID<N>>) -> Result<Arc<Stack<N>>> {
```
ljedrz: I think this could be reimplemented as the following one-liner:
```
Ok(Arc::clone(self.get_stack_ref(program_id)?))
```
or
```
self.get_stack_ref(program_id).map(Arc::clone)
```
(as long as `get_stack_ref` can return a `Result<&Arc<Stack<N>>>` instead of `Result<&Stack<N>>`, which should be fine)

synthesizer/process/src/stack/mod.rs
```diff
@@ -235,9 +235,25 @@ impl<N: Network> StackProgram<N> for Stack<N> {
 
     /// Returns the external stack for the given program ID.
     #[inline]
-    fn get_external_stack(&self, program_id: &ProgramID<N>) -> Result<&Stack<N>> {
+    fn get_external_stack(&self, program_id: &ProgramID<N>) -> Result<Arc<Stack<N>>> {
```
ljedrz: I'd suggest the same approach as with `get_stack`, if possible

ljedrz (approved):
LGTM perf-wise; I can reproduce the issue with the test on the `testnet3` branch, and can no longer reproduce it with this branch. Also, the new test case is a treasure trove of profiling data, I'll be looking into it soon :microscope:.

synthesizer/process/src/lib.rs
vicsn: Neat. Let me know what you think: a3f93f00a

Note, the reason I introduced the `_ref()` functions was to avoid littering the codebase with `as_ref()`'s or refactoring a bunch of function signatures to use `Arc`'s instead of refs (which I looked into again but is very involved).

ljedrz (approved):

synthesizer/process/src/lib.rs
ljedrz: that makes sense, and I think the end result is a good compromise; simple APIs are nice, but code duplication can end up resulting in future errors, so I try to avoid it whenever possible

synthesizer/process/src/lib.rs
vicsn: Indeed we can! 98da30277

synthesizer/process/src/stack/mod.rs
vicsn: a5a1898c6

synthesizer/process/src/stack/call/mod.rs
vicsn: b30d4f415

synthesizer/src/vm/mod.rs
vicsn: Thx for catching b30d4f415

howardwu (approved):

PR #2177, opened 2023-11-17
Fewer BHP allocations

ljedrz:
This PR reduces the number of allocations needed for BHP (both console and circuit) hashing. The differences - measured for the console version, for a single long (`102400` booleans) input - are as follows:
```
allocations:   199 -> 2
reallocations: 1 -> 0
max mem use:   621.63 KiB -> 620.70 KiB
```
And this improvement scales quite well, e.g. for 100x more booleans the total number of allocations is still only `2`, compared to `19,619` with the current code.


ljedrz:
Marking as a draft for now, as there seems to be some issue with the circuit version :microscope:.

edit: ~I think I got it, let's see the CI~ all fixed now

howardwu (approved):

PR #2185, opened 2023-11-22
feat: bond enough aleo but make a delegate address be a validator

swift-mx:
## Motivation

Bond enough aleo but make a delegate address be a validator. solve the problem of asset security for Validators.


swift-mx:
TODO: In this scenario, when A delegates 1 million to B for B to become a validator, if C also delegates 100,000 to B, and A withdraws 100,000 (assuming B's remaining amount is still greater than or equal to 1 million), it may result in C being unable to withdraw the delegated amount.

swift-mx:
Some solution:
```rust
// The committee_state struct tracks the total stake of the validator, and whether they are open to stakers.
struct committee_state:
// The amount of microcredits bonded to the validator, by the validator and its delegators.
microcredits as u64;
// The boolean flag indicating if the validator is open to stakers.
is_open as boolean;
// The delegate address of the validator bft.
delegate_address as address;
```
add a parameter delegate_address in committee_state, when A delegates 1 million to B for B to become a validator. A is delegate_address, check the microcredits gte 1 million when delegate_address unbond

howardwu:
Can you answer the following concerns:

- What happens if a validator has less than 1M credits initially? Doesn't this imply they can squat all 200 validator slots with no stake
- What happens if the delegator unbonds their full stake and the validator now has nothing at stake?

swift-mx:
Thanks for your response. 

The purpose of this Pull Request  is to split the validator address into owner and worker addresses. 
Here are answers to some questions you raised:

- During initialization, the validator must be provided with a single address that contributes more than 1 million points, this address will be recorded as the delegate address (i.e., the `owner address).

- The validator is required to execute the unbond_delegator_as_validator operation on all bond addresses before performing the unbond_validator operation on the delegate_address (i.e., owner address).

swift-mx:
There might be better ways to make modifications, but in order to keep code changes minimal, the original logic has been reused.

raychu86:
Closing because ARC-0037 resolves these concerns - https://github.com/AleoNet/snarkVM/pull/2385

PR #2189, opened 2023-11-23
Always derive sk_prf

vicsn:
Requires resampling. Total constraints for the credits.aleo functions don't increase.


howardwu:
This does not seem desirable.

vicsn:
Closing because there is no actionable security risk to resolve warranting the heavy migration

PR #2191, opened 2023-11-23
[NCC THN] Correctly sample random vectors

vicsn:
## Motivation

We were sampling a single random element and duplicating it 10  times.


ljedrz (approved):
Indeed, LGTM.

howardwu (approved):

PR #2192, opened 2023-11-23
Prevent hypothetical overflow

vicsn:
## Motivation

Closes https://github.com/AleoHQ/snarkVM/issues/2166

Note that with our current stack we would have never hit these cases. But they could be a hypothetical concern when we support delegated proving and someone wants to DoS the party willing to do the proving.


ljedrz (approved):
LGTM, though I'd make the error messages a bit more informative; since these are not panics, we may be missing some useful debugging context if this ever happens IRL

howardwu:
Moving back to draft, as CI is failing

vicsn:
Fixed the broken test. The two CI tests which are still failing are unrelated to this PR and succeed locally. Perhaps because reduce_pk_size_3 (which this merges into) requires resampling

howardwu (approved):

PR #2194, opened 2023-11-24
Prevent panics

vicsn:
## Motivation

Given that anyone in the network can freely pass proof objects (as Execution, containing transitions, containing proofs) into the network, we should not panic easily.

I search for uses of `assert` and `unwrap` in the algorithms crate, and for now skipped most of the prover (which is not done by validators) or lower level FFT code.

### DIFF
--- a/algorithms/src/snark/varuna/ahp/indexer/indexer.rs
+++ b/algorithms/src/snark/varuna/ahp/indexer/indexer.rs
@@ -27,12 +27,12 @@ use crate::{
         SNARKMode,
     },
 };
-use itertools::Itertools;
 use snarkvm_fields::PrimeField;
 use snarkvm_utilities::cfg_into_iter;
 
 use anyhow::{anyhow, Result};
 use core::marker::PhantomData;
+use itertools::Itertools;
 use std::collections::BTreeMap;
 
 #[cfg(not(feature = "serial"))]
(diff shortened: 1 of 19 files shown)


algorithms/src/snark/varuna/ahp/indexer/indexer.rs
```diff
@@ -233,16 +229,18 @@ impl<F: PrimeField, SM: SNARKMode> AHPForR1CS<F, SM> {
...
         ]
         .into_iter()
-        .flat_map(move |(evals, domain)| {
-            let labels = Self::index_polynomial_labels(&["a", "b", "c"], std::iter::once(id));
+        .zip_eq(&labels)
+        .flat_map(|((evals, domain), label)| {
+            let labels = Self::index_polynomial_labels(label, std::iter::once(id));
             let lagrange_coefficients_at_point = domain.evaluate_all_lagrange_coefficients(point);
-            labels.zip(evals.evaluate(&lagrange_coefficients_at_point).unwrap())
+            labels.zip_eq(evals.evaluate(&lagrange_coefficients_at_point).unwrap())
```
howardwu, line 243: The lengths must be checked first here, otherwise a panic is imminent

howardwu (approved):
LGTM

algorithms/src/snark/varuna/ahp/indexer/indexer.rs
howardwu: Left as a new Github issue to resolve #2195

PR #2203, opened 2023-11-26
Add zip_eq safeguards to prevent panic

### LINKED ISSUE #2195: The lengths must be checked first here, otherwise a panic is imminent
The lengths must be checked first here, otherwise a panic is imminent

_Originally posted by @howardwu in https://github.com/AleoHQ/snarkVM/pull/2194#discussion_r1404619149_

### LINKED ISSUE #2216: [Bug] Panic when `zip_eq` on two different size iterables when calling the `evaluate_with_coeffs()` function with a wrong size evaluation vector
# Panic when `zip_eq` on two different size iterables when calling the `evaluate_with_coeffs()` function with a wrong size evaluation vector

**Author(s):** Ronan Thoraval from [@Fuzzinglabs](https://github.com/FuzzingLabs/)
**Date:** 29/11/2023

## ****Executive Summary****

Calling the [`evaluate_with_coeffs()`](https://github.com/AleoHQ/snarkVM/blob/afb38c59bc6d42a8f69fbb2fa526b7226b152328/algorithms/src/fft/evaluations.rs#L87) function on parameters `self` (which is a `&Evaluations<F>`) and `lagrange_coefficients_at_point` with different size leads to `zip_eq` on two iterables with different lengths, and so panics.

## Vulnerability Details

- **Severity:** Medium

## Environment

- **Compiler Version:** cargo 1.73.0 (9c4383fb5 2023-08-26)
- **Distro Version:** Ubuntu 23.04

## Steps to Reproduce

``` rust
use ark_std::One;
use fuzz_lib::*;
use snarkvm_curves::bls12_377::Fq as AleoFq;
use snarkvm_algorithms::polycommit::kzg10 as aleo_KZG10;
use snarkvm_curves::bls12_377::Fr as aleo_field;
use snarkvm_algorithms::polycommit::sonic_pc as aleo_sonic;
use snarkvm_algorithms::crypto_hash::PoseidonSponge;
use snarkvm_fields::Zero;

type SonicKzg10Perso = aleo_sonic::SonicKZG10::<EAleo,PoseidonSponge<AleoFq, 2, 1>>;

fn main(){

   let evals = vec![aleo_field::one(),aleo_field::one(),aleo_field::one()];

    let domain = snarkvm_algorithms::fft::EvaluationDomain::<aleo_field>::new(evals.len()).unwrap();
    let evals = snarkvm_algorithms::fft::Evaluations::from_vec_and_domain(evals, domain);
    let aleo_poly = aleo_sonic::LabeledPolynomialWithBasis::new_lagrange_basis("".to_string(), evals, None);

    let poly = &aleo_poly.polynomial.get(0).unwrap().1;

    let evaluations = match poly{
        aleo_sonic::PolynomialWithBasis::Lagrange { evaluations } => {
            evaluations.clone().into_owned()
        }
        _ => return
    };
    
    let aleo_point = aleo_field::zero();

    // evaluate() calls the evaluate_with_coeffs() function
    let eval_poly_on_point = evaluations.evaluate(&aleo_point);
}
```

## Root Cause Analysis

We construct a Lagrange polynomial $p$ with $p(1)=p(\omega)=p(\omega^2)=1$, where $\omega$ is a generator of a subgroup of size 4. Note that the subgroup has more elements than the vector of evaluations of $p$ ($=[1,1,1]$).

Calling the [`evaluate()`](https://github.com/AleoHQ/snarkVM/blob/afb38c59bc6d42a8f69fbb2fa526b7226b152328/algorithms/src/fft/evaluations.rs#L82C11-L82C11) function leads to call the [`evaluate_with_coeffs()`](https://github.com/AleoHQ/snarkVM/blob/afb38c59bc6d42a8f69fbb2fa526b7226b152328/algorithms/src/fft/evaluations.rs#L87) function on `self`, of size 3, and `lagrange_coefficients_at_point`, of same size than the `domain`, *i.e* 4. 

[`evaluate_with_coeffs()`](https://github.com/AleoHQ/snarkVM/blob/afb38c59bc6d42a8f69fbb2fa526b7226b152328/algorithms/src/fft/evaluations.rs#L87) then calls `zip_eq` on the previous parameters (of different sizes), and so panics.

Certain functions give the impression of expecting a vector of evaluations of the same size as the domain.

## Detailed Behavior

``` sh
thread 'main' panicked at /home/ronan/Desktop/Projects/ALEO/snarkVM/algorithms/src/fft/evaluations.rs:95:37:
assertion `left == right` failed: iterators must have the same length
  left: 3
 right: 4
stack backtrace:
   0: rust_begin_unwind
             at /rustc/cc66ad468955717ab92600c770da8c1601a4ff33/library/std/src/panicking.rs:595:5
   1: core::panicking::panic_fmt
             at /rustc/cc66ad468955717ab92600c770da8c1601a4ff33/library/core/src/panicking.rs:67:14
   2: core::panicking::assert_failed_inner
   3: core::panicking::assert_failed
             at /rustc/cc66ad468955717ab92600c770da8c1601a4ff33/library/core/src/panicking.rs:229:5
   4: rayon::iter::IndexedParallelIterator::zip_eq
             at /home/ronan/.cargo/registry/src/index.crates.io-6f17d22bba15001f/rayon-1.8.0/src/iter/mod.rs:2498:9
   5: snarkvm_algorithms::fft::evaluations::Evaluations<F>::evaluate_with_coeffs
             at /home/ronan/Desktop/Projects/ALEO/snarkVM/algorithms/src/fft/evaluations.rs:95:9
   6: snarkvm_algorithms::fft::evaluations::Evaluations<F>::evaluate
             at /home/ronan/Desktop/Projects/ALEO/snarkVM/algorithms/src/fft/evaluations.rs:91:9
   7: kzg_lag_open_errors::error_zip_eq_open_lagrange
             at ./src/error_found.rs:45:30
   8: kzg_lag_open_errors::main
             at ./src/error_found.rs:58:5
   9: core::ops::function::FnOnce::call_once
             at /rustc/cc66ad468955717ab92600c770da8c1601a4ff33/library/core/src/ops/function.rs:250:5   
```

## Recommendations

When creating a `Evaluations<F>` element, fill the vector of evaluations with 0 until it has the same size as the domain (in [`from_vec_and_domain()`](https://github.com/AleoHQ/snarkVM/blob/afb38c59bc6d42a8f69fbb2fa526b7226b152328/algorithms/src/fft/evaluations.rs#L41C8-L41C8) function):

From
``` rust
 pub fn from_vec_and_domain(evaluations: Vec<F>, domain: EvaluationDomain<F>) -> Self {
        Self { evaluations, domain }
    }
```

To 
``` rust
 pub fn from_vec_and_domain(evaluations: Vec<F>, domain: EvaluationDomain<F>) -> Self {
        let mut evaluations = evaluations.clone();
        evaluations.resize(domain.size(), F::zero());
        Self { evaluations, domain }
    }
```

## Appendices

In the [`open_lagrange()`](https://github.com/AleoHQ/snarkVM/blob/afb38c59bc6d42a8f69fbb2fa526b7226b152328/algorithms/src/polycommit/kzg10/mod.rs#L273) function, the same problem is present. Suppose that we don't call the [`evaluate()`](https://github.com/AleoHQ/snarkVM/blob/afb38c59bc6d42a8f69fbb2fa526b7226b152328/algorithms/src/fft/evaluations.rs#L82C11-L82C11) function to get the evaluation of the polynomial, then, if the vector of evaluations is of size between `domain.size()` and `domain.size()/2`, then the first three checks of the function will pass, but a `zip_eq()` function will be call on prameters with differents length ([l.293](https://github.com/AleoHQ/snarkVM/blob/afb38c59bc6d42a8f69fbb2fa526b7226b152328/algorithms/src/polycommit/kzg10/mod.rs#L293)).

vicsn:
## Motivation

Closes https://github.com/AleoHQ/snarkVM/issues/2195
Closes https://github.com/AleoHQ/snarkVM/issues/2216

For cases where we `zip_eq` on values (derived) from user-input, it is prudent to add an extra check that the values are of equal length to prevent panic.
Note that we can't easily make a new generic `zip_eq` impl which `ensure!`'s the length because in various places in the code we work with iterators which don't know their length at compile-time. See also this interesting discussion: https://internals.rust-lang.org/t/should-we-implement-exactsizeiterator-for-chain/12556/7

### DIFF
--- a/algorithms/src/fft/evaluations.rs
+++ b/algorithms/src/fft/evaluations.rs
@@ -38,7 +38,9 @@ pub struct Evaluations<F: PrimeField> {
 
 impl<F: PrimeField> Evaluations<F> {
     /// Construct `Self` from evaluations and a domain.
-    pub fn from_vec_and_domain(evaluations: Vec<F>, domain: EvaluationDomain<F>) -> Self {
+    pub fn from_vec_and_domain(mut evaluations: Vec<F>, domain: EvaluationDomain<F>) -> Self {
+        // Pad evaluations to ensure we can always evaluate
+        evaluations.resize(domain.size(), F::zero());
         Self { evaluations, domain }
     }
 
--- a/algorithms/src/fft/evaluations.rs
+++ b/algorithms/src/fft/evaluations.rs
@@ -15,6 +15,7 @@
 //! A polynomial represented in evaluations form.
 
 use crate::fft::{DensePolynomial, EvaluationDomain};
+use anyhow::Result;
 #[cfg(feature = "serial")]
 use itertools::Itertools;
 #[cfg(not(feature = "serial"))]
--- a/algorithms/src/fft/evaluations.rs
+++ b/algorithms/src/fft/evaluations.rs
@@ -15,7 +15,6 @@
 //! A polynomial represented in evaluations form.
 
 use crate::fft::{DensePolynomial, EvaluationDomain};
-use anyhow::Result;
 #[cfg(feature = "serial")]
 use itertools::Itertools;
 #[cfg(not(feature = "serial"))]
--- a/algorithms/src/polycommit/kzg10/mod.rs
+++ b/algorithms/src/polycommit/kzg10/mod.rs
@@ -24,7 +24,7 @@ use crate::{
     msm::VariableBase,
     polycommit::PCError,
 };
-use anyhow::anyhow;
+use anyhow::{anyhow, ensure, Result};
 use snarkvm_curves::traits::{AffineCurve, PairingCurve, PairingEngine, ProjectiveCurve};
 use snarkvm_fields::{One, PrimeField, Zero};
 use snarkvm_utilities::{cfg_iter, cfg_iter_mut, rand::Uniform, BitIteratorBE};
@@ -276,7 +276,7 @@ impl<E: PairingEngine> KZG10<E> {
         evaluations: &[E::Fr],
         point: E::Fr,
         evaluation_at_point: E::Fr,
-    ) -> Result<KZGProof<E>, PCError> {
+    ) -> Result<KZGProof<E>> {
         Self::check_degree_is_too_large(evaluations.len() - 1, lagrange_basis.size())?;
         // Ensure that the point is not in the domain
         if lagrange_basis.domain.evaluate_vanishing_polynomial(point).is_zero() {
@@ -290,6 +290,7 @@ impl<E: PairingEngine> KZG10<E> {
 
         let mut divisor_evals = cfg_iter!(domain_elements).map(|&e| e - point).collect::<Vec<_>>();
         snarkvm_fields::batch_inversion(&mut divisor_evals);
+        ensure!(divisor_evals.len() == evaluations.len());
         cfg_iter_mut!(divisor_evals).zip_eq(evaluations).for_each(|(divisor_eval, &eval)| {
             *divisor_eval *= eval - evaluation_at_point;
         });
@@ -351,7 +352,7 @@ impl<E: PairingEngine> KZG10<E> {
         values: &[E::Fr],
         proofs: &[KZGProof<E>],
         rng: &mut R,
-    ) -> Result<bool, PCError> {
+    ) -> Result<bool> {
         let check_time = start_timer!(|| format!("Checking {} evaluation proofs", commitments.len()));
         let g = vk.g.to_projective();
         let gamma_g = vk.gamma_g.to_projective();
@@ -365,6 +366,9 @@ impl<E: PairingEngine> KZG10<E> {
         // their coefficients and perform a final multiplication at the end.
         let mut g_multiplier = E::Fr::zero();
         let mut gamma_g_multiplier = E::Fr::zero();
+        ensure!(commitments.len() == points.len());
+        ensure!(commitments.len() == values.len());
+        ensure!(commitments.len() == proofs.len());
         for (((c, z), v), proof) in commitments.iter().zip_eq(points).zip_eq(values).zip_eq(proofs) {
             let w = proof.w;
             let mut temp = w.mul(*z);
--- a/algorithms/src/polycommit/kzg10/mod.rs
+++ b/algorithms/src/polycommit/kzg10/mod.rs
@@ -24,11 +24,11 @@ use crate::{
     msm::VariableBase,
     polycommit::PCError,
 };
-use anyhow::{anyhow, ensure, Result};
 use snarkvm_curves::traits::{AffineCurve, PairingCurve, PairingEngine, ProjectiveCurve};
 use snarkvm_fields::{One, PrimeField, Zero};
 use snarkvm_utilities::{cfg_iter, cfg_iter_mut, rand::Uniform, BitIteratorBE};
 
+use anyhow::{anyhow, ensure, Result};
 use core::{marker::PhantomData, ops::Mul};
 use itertools::Itertools;
 use rand_core::RngCore;
--- a/algorithms/src/polycommit/sonic_pc/polynomial.rs
+++ b/algorithms/src/polycommit/sonic_pc/polynomial.rs
@@ -14,10 +14,10 @@
 
 use super::PolynomialLabel;
 use crate::fft::{DensePolynomial, EvaluationDomain, Evaluations as EvaluationsOnDomain, Polynomial, SparsePolynomial};
-use anyhow::{ensure, Result};
 use snarkvm_fields::{Field, PrimeField};
 use snarkvm_utilities::{cfg_iter, cfg_iter_mut, CanonicalDeserialize, CanonicalSerialize};
 
+use anyhow::{ensure, Result};
 use hashbrown::HashMap;
 use std::borrow::Cow;
 
--- a/algorithms/src/polycommit/sonic_pc/polynomial.rs
+++ b/algorithms/src/polycommit/sonic_pc/polynomial.rs
@@ -235,8 +235,9 @@ impl<'a, F: PrimeField> LabeledPolynomialWithBasis<'a, F> {
                         match polynomial.as_ref() {
                             Dense(p) => {
                                 if let Some(e) = dense_polys.get_mut(degree_bound) {
-                                    // Zip safety: `p` could be of smaller degree than `e` (or vice versa),
+                                    // Zip safety: `p` could be of smaller degree than `e`,
                                     // so it's okay to just use `zip` here.
+                                    ensure!(e.len() >= p.coeffs.len());
                                     cfg_iter_mut!(e).zip(&p.coeffs).for_each(|(e, f)| *e += *c * f)
                                 } else {
                                     let mut e: DensePolynomial<F> = p.clone().into_owned();
--- a/algorithms/src/polycommit/sonic_pc/polynomial.rs
+++ b/algorithms/src/polycommit/sonic_pc/polynomial.rs
@@ -14,6 +14,7 @@
 
 use super::PolynomialLabel;
 use crate::fft::{DensePolynomial, EvaluationDomain, Evaluations as EvaluationsOnDomain, Polynomial, SparsePolynomial};
+use anyhow::{ensure, Result};
 use snarkvm_fields::{Field, PrimeField};
 use snarkvm_utilities::{cfg_iter, cfg_iter_mut, CanonicalDeserialize, CanonicalSerialize};
 
@@ -215,9 +216,9 @@ impl<'a, F: PrimeField> LabeledPolynomialWithBasis<'a, F> {
 
     /// Compute a linear combination of the terms in `self.polynomial`, producing an iterator
     /// over polynomials of the same time.
-    pub fn sum(&self) -> impl Iterator<Item = PolynomialWithBasis<'a, F>> {
+    pub fn sum(&self) -> Result<impl Iterator<Item = PolynomialWithBasis<'a, F>>> {
         if self.polynomial.len() == 1 && self.polynomial[0].0.is_one() {
-            vec![self.polynomial[0].1.clone()].into_iter()
+            Ok(vec![self.polynomial[0].1.clone()].into_iter())
         } else {
             use PolynomialWithBasis::*;
             let mut lagrange_polys = HashMap::<usize, Vec<_>>::new();
@@ -249,6 +250,7 @@ impl<'a, F: PrimeField> LabeledPolynomialWithBasis<'a, F> {
                     Lagrange { evaluations } => {
                         let domain = evaluations.domain().size();
                         if let Some(e) = lagrange_polys.get_mut(&domain) {
+                            ensure!(e.len() == evaluations.evaluations.len());
                             cfg_iter_mut!(e).zip_eq(&evaluations.evaluations).for_each(|(e, f)| *e += *c * f)
                         } else {
                             let mut e = evaluations.clone().into_owned().evaluations;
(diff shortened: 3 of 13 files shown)


algorithms/src/polycommit/sonic_pc/polynomial.rs
```diff
@@ -14,6 +14,7 @@
 
 use super::PolynomialLabel;
 use crate::fft::{DensePolynomial, EvaluationDomain, Evaluations as EvaluationsOnDomain, Polynomial, SparsePolynomial};
+use anyhow::{ensure, Result};
```
howardwu: nit: move down to next import block

algorithms/src/polycommit/kzg10/mod.rs
```diff
@@ -24,7 +24,7 @@ use crate::{
     msm::VariableBase,
     polycommit::PCError,
 };
-use anyhow::anyhow;
+use anyhow::{anyhow, ensure, Result};
```
howardwu: nit: reorder to import block below

algorithms/src/snark/varuna/ahp/indexer/circuit.rs
```diff
@@ -139,12 +139,11 @@ impl<F: PrimeField, SM: SNARKMode> Circuit<F, SM> {
...
-            .into_iter()
-            .map(|(label, evals)| MatrixArithmetization::new(&self.id, label, evals))
-            .collect::<Result<Vec<_>, _>>()?
-            .into_iter()
-            .flat_map(|arith| arith.into_iter()))
+        let mut iters = Vec::with_capacity(3);
+        for (label, evals) in [("a", &self.a_arith), ("b", &self.b_arith), ("c", &self.c_arith)] {
+            iters.push(MatrixArithmetization::new(&self.id, label, evals)?.into_iter());
+        }
+        Ok(iters.into_iter().flatten())
```
howardwu, line 146: @ljedrz please review for performance hit. Call to allocated vector followed by constructing a flattened iterator needs performance review.

algorithms/src/snark/varuna/ahp/indexer/indexer.rs
```diff
@@ -105,7 +104,7 @@ impl<F: PrimeField, SM: SNARKMode> AHPForR1CS<F, SM> {
     pub fn index_polynomial_labels<'a>(
         matrices: &'a [&str],
         ids: impl Iterator<Item = &'a CircuitId> + 'a,
-    ) -> impl Iterator<Item = PolynomialLabel> + 'a {
+    ) -> Vec<PolynomialLabel> {
```
howardwu: @ljedrz performance hit likely. Can you review?

algorithms/src/snark/varuna/ahp/indexer/indexer.rs
```diff
@@ -228,23 +228,21 @@ impl<F: PrimeField, SM: SNARKMode> AHPForR1CS<F, SM> {
...
-        let mut evals = [
-            (state.a_arith, state.non_zero_a_domain),
-            (state.b_arith, state.non_zero_b_domain),
-            (state.c_arith, state.non_zero_c_domain),
-        ]
-        .into_iter()
-        .zip_eq(&labels)
-        .flat_map(|((evals, domain), label)| {
+    ) -> Result<impl ExactSizeIterator<Item = F>> {
+        let mut all_evals = Vec::with_capacity(12);
```
howardwu: @ljedrz allocation amount needs further review. Can you validate size?

algorithms/src/snark/varuna/varuna.rs
```diff
@@ -310,11 +310,14 @@ where
         let evaluations_at_point = AHPForR1CS::<E::Fr, SM>::evaluate_index_polynomials(state, circuit_id, point)?;
         let mut lc = crate::polycommit::sonic_pc::LinearCombination::empty("circuit_check");
         let mut evaluation = E::Fr::zero();
+        ensure!(poly_info.len() == challenges.len() + 1);
```
howardwu: Directly check against `linear_combination_challenges`

algorithms/src/snark/varuna/varuna.rs
```diff
@@ -843,10 +846,11 @@ where
         // correct degree.
 
         // Gather commitments in one vector.
+        let poly_infos = AHPForR1CS::<E::Fr, SM>::index_polynomial_info(circuit_ids.iter());
```
howardwu: Why was this moved? Needless change.

algorithms/src/polycommit/sonic_pc/polynomial.rs
```diff
@@ -249,6 +250,7 @@ impl<'a, F: PrimeField> LabeledPolynomialWithBasis<'a, F> {
                     Lagrange { evaluations } => {
                         let domain = evaluations.domain().size();
                         if let Some(e) = lagrange_polys.get_mut(&domain) {
+                            ensure!(e.len() == evaluations.evaluations.len());
```
howardwu, line 254: Does the monomial case above require the same `zip_eq` treatment? We currently only just `.zip` for it.

howardwu:
@vicsn there are times when `.zip` is required rather than `.zip_eq`, because the second iterator can be shorter than the first. Consider in an MSM where `bases.zip(scalars)` is the correct behavior, because if there are less scalars than bases, you only need to compute up to the scalars' length.

We should be confident that each usage of `.zip_eq` here adheres to this desired property of matching length.

algorithms/src/snark/varuna/ahp/indexer/indexer.rs
ljedrz: indeed, this one is not great, and I assume this is due to us wanting to know the length; however, that can be achieved using iterator methods, e.g. that both iterators are exhausted with `.next().is_none()` after having used them (paired with `.by_ref()` so that they don't get consumed), or checking the remaining length using `.count()`

algorithms/src/snark/varuna/ahp/indexer/circuit.rs
ljedrz: this one is most likely fine, as we've used `collect` before; I'll run some benches to double-check

algorithms/src/snark/varuna/ahp/indexer/indexer.rs
ljedrz: this one is most likely fine, as we've used `collect` before; I'll run some benches to double-check

ljedrz (commented):
I ran the `transaction` and `varuna` benches and there doesn't seem to be any degradation in performance, but I think [this](https://github.com/AleoHQ/snarkVM/pull/2203/files#r1405463879) new allocation can be avoided.

algorithms/src/polycommit/sonic_pc/polynomial.rs
vicsn: Hmm now that you mention, I think that zip only holds under a specific condition: e2674ba50
However, plot twist, this is dead code which we never use. We also needlessly clone all polys in this setup. Removing it all here: https://github.com/AleoHQ/snarkVM/pull/2207

In the meantime I'll continue the search for remaining panics to avoid.

algorithms/src/polycommit/sonic_pc/polynomial.rs
vicsn: cbfaeb0c1

algorithms/src/polycommit/kzg10/mod.rs
vicsn: cbfaeb0c1

algorithms/src/snark/varuna/ahp/indexer/indexer.rs
vicsn: Refactored so we avoid the allocations: fdfd68a9c

algorithms/src/snark/varuna/varuna.rs
vicsn: Refactored so we are assured the lengths are equal and we don't panic: fdfd68a9c

algorithms/src/snark/varuna/varuna.rs
vicsn: Moved it back. Sorry I just see I removed another comment along with it but I assume its ok: e23945a69

vicsn:
Please also review 11889cd3d - clippy started complaining about it but not sure why must_use was needed

algorithms/src/fft/evaluations.rs
```diff
@@ -38,7 +38,8 @@ pub struct Evaluations<F: PrimeField> {
 
 impl<F: PrimeField> Evaluations<F> {
     /// Construct `Self` from evaluations and a domain.
-    pub fn from_vec_and_domain(evaluations: Vec<F>, domain: EvaluationDomain<F>) -> Self {
+    pub fn from_vec_and_domain(mut evaluations: Vec<F>, domain: EvaluationDomain<F>) -> Self {
+        evaluations.resize(domain.size(), F::zero());
```
ljedrz, line 43: perhaps a comment would be useful here? I assume these extra zeros are not meaningful and only serve as padding?

algorithms/src/polycommit/sonic_pc/polynomial.rs
```diff
@@ -215,9 +216,9 @@ impl<'a, F: PrimeField> LabeledPolynomialWithBasis<'a, F> {
 
     /// Compute a linear combination of the terms in `self.polynomial`, producing an iterator
     /// over polynomials of the same time.
-    pub fn sum(&self) -> impl Iterator<Item = PolynomialWithBasis<'a, F>> {
+    pub fn sum(&self) -> Result<impl Iterator<Item = PolynomialWithBasis<'a, F>>> {
         if self.polynomial.len() == 1 && self.polynomial[0].0.is_one() {
-            vec![self.polynomial[0].1.clone()].into_iter()
+            Ok(vec![self.polynomial[0].1.clone()].into_iter())
```
ljedrz, line 221:
```suggestion
            Ok([self.polynomial[0].1.clone()].into_iter())
```

ljedrz (approved):
LGTM, and I especially like the greater reliance on `ExactSizeIterator`, which is very practical for preallocations.

algorithms/src/polycommit/sonic_pc/polynomial.rs
vicsn: Actually I wanted to do the same but because the else condition returns a Vec.into_iter(), which is a different type, we can't. Then decided to remove the entire else statement because it is dead code: https://github.com/AleoHQ/snarkVM/pull/2207

ljedrz: Ah, that makes sense, I didn't notice it; since it's going to be cut out, nothing to do here.

howardwu (approved):
LGTM, only reviewed for code quality, unable to review for correctness

howardwu:
Screenshot 2023-12-17 at 3 38 35 PM

Namely, I cannot vet `AHPForR1CS::<E::Fr, SM>::evaluate_index_polynomials` or `Varuna::verify_vk` changes in the PR.

Possible breakage risk will be on Deployments in the future.

Merging as the risk of panics outweighs correctness.

algorithms/src/snark/varuna/ahp/indexer/indexer.rs
```diff
@@ -224,27 +227,35 @@ impl<F: PrimeField, SM: SNARKMode> AHPForR1CS<F, SM> {
...
-        .collect::<Vec<_>>();
-        evals.sort_by(|(l1, _), (l2, _)| l1.cmp(l2));
-        Ok(evals.into_iter().map(|(_, eval)| eval))
+            let evals_at_point = evals.evaluate(&lagrange_coefficients_at_point)?;
+            ensure!(labels.len() == evals_at_point.len());
+            all_evals.push(labels.into_iter().zip_eq(evals_at_point.into_iter()));
+        }
+        let sorted_evals = all_evals.into_iter().flatten().sorted_unstable_by(|(l1, _), (l2, _)| l1.cmp(l2));
+        for (label, eval) in sorted_evals {
+            let combiner = combiners.next().ok_or(anyhow!("No combiner left"))?;
```
howardwu, line 254: Is it safe for `combiners` to be longer than `sorted_evals`?

If not, we should add an `ensure!` on the lengths.

PR #2207, opened 2023-11-27
Perf: Remove clone and dead code linear combinations of commitments

vicsn:
## Motivation

This PR removes cloning some very large polynomials.

We supported `LabeledPolynomialWithBasis` holding a linear combination of polynomials, even though in practice we only ever stored one polynomial in there. I suspect this was an old version of the code, in a time when we didn't have the higher level linear combination abstractions.

### DIFF
--- a/algorithms/src/polycommit/sonic_pc/mod.rs
+++ b/algorithms/src/polycommit/sonic_pc/mod.rs
@@ -209,43 +209,34 @@ impl<E: PairingEngine, S: AlgebraicSponge<E::Fq, 2>> SonicKZG10<E, S> {
                     hiding_bound,
                 ));
 
-                let (comm, rand) = p
-                    .sum()?
-                    .map(move |p| {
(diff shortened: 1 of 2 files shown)


Pratyush (commented):
I think we can even remove the enum, as we don't use the `Lagrange` basis anymore, and so everything is in Monomial basis, at which point there's no use of the enum.

vicsn:
> I think we can even remove the enum, as we don't use the Lagrange basis anymore, and so everything is in Monomial basis, at which point there's no use of the enum.

I agree, but we still have a couple of tests which use the Lagrange basis. So that'll be a lot more lines of codes and prefer to tackle in a new PR.

Pratyush (approved):
Sounds good, in that case this PR LGTM

howardwu:
@vicsn if you could resolve the merge conflicts, I'll be happy to merge it in.

vicsn:
Rebased

PR #2211, opened 2023-11-27
Enforce `ProgramID` uniqueness in a block

raychu86:
## Motivation

This PR adds checks to enforce that a block does not have duplicate program IDs.

In addition, we reject subsequent deployments with the same `ProgramID` during speculation.

### DIFF
--- a/ledger/block/src/verify.rs
+++ b/ledger/block/src/verify.rs
@@ -412,10 +412,7 @@ impl<N: Network> Block<N> {
 
         // Ensure there are not duplicate program IDs.
         if has_duplicates(
-            self.transactions()
-                .iter()
-                .filter_map(|tx| tx.transaction().deployment())
-                .map(|deployment| deployment.program_id()),
+            self.transactions().iter().filter_map(|tx| tx.transaction().deployment().map(|d| d.program_id())),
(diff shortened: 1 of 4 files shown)


ledger/block/src/verify.rs
```diff
@@ -410,6 +410,13 @@ impl<N: Network> Block<N> {
             bail!("Found a duplicate transition in block {height}");
         }
 
+        // Ensure there are not duplicate program IDs.
+        if has_duplicates(
+            self.transactions().iter().filter_map(|tx| tx.transaction().deployment().map(|d| d.program_id())),
```
howardwu, line 415: This call checks ALL accepted and rejected deployments are unique. I think this is possible to fail. We should only be checking the accepted no?

raychu86: I believe this check only checks the accepted deployments. 

`ConfirmedTransaction::transaction()` gives you the underlying transaction, which is a `Deployment`, `Execution`, or `Fee`. Then the `Transaction::deployment` method returns a Some(Deployment) only for the `Deployment` type, which excludes the rejected deployments.

howardwu (approved):

PR #2212, opened 2023-11-28
[Fix] Add check for the correct number of commands and inputs in a `closure` and `finalize`.

d0cd:
This PR adds:
- a check for nonzero commands in `FinalizeCore::write_le`
- a check for nonzero commands in `FinalizeCore::read_le`
- a check for nonzero inputs in `ClosureCore::parse`
- a check for nonzero inputs in `ClosureCore::write_le`
- a check for nonzero inputs in `ClosureCore::read_le`
- a check for nonzero instructions in `ClosureCore::write_le`
- a check for nonzero instructions in `ClosureCore::read_le`

### DIFF
--- a/synthesizer/program/src/closure/bytes.rs
+++ b/synthesizer/program/src/closure/bytes.rs
@@ -23,6 +23,9 @@ impl<N: Network, Instruction: InstructionTrait<N>> FromBytes for ClosureCore<N, 
 
         // Read the inputs.
         let num_inputs = u16::read_le(&mut reader)?;
+        if num_inputs.is_zero() {
+            return Err(error(format!("Failed to deserialize a closure: needs at least one input")));
+        }
         if num_inputs > u16::try_from(N::MAX_INPUTS).map_err(error)? {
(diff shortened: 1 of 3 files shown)


synthesizer/program/src/closure/bytes.rs
```diff
@@ -23,6 +23,9 @@ impl<N: Network, Instruction: InstructionTrait<N>> FromBytes for ClosureCore<N,
 
         // Read the inputs.
         let num_inputs = u16::read_le(&mut reader)?;
+        if num_inputs.is_zero() {
```
howardwu, line 26: I think it could be valid to write a closure without an input. Not sure why, but it seems like a possibly desired result?

d0cd: Yeah, maybe they'd like to reuse some fixed code a number of times.
We'd have to relax [this](https://github.com/AleoHQ/snarkVM/blob/afb38c59bc6d42a8f69fbb2fa526b7226b152328/synthesizer/program/src/lib.rs#L255) check.

howardwu (approved):

raychu86 (approved):

PR #2219, opened 2023-11-30
Add length methods to MapRead & NestedMapRead

ljedrz:
This PR adds the following new methods to the store-related traits:
- `MapRead::{len_confirmed, len_pending, is_empty_confirmed, is_empty_pending}`
- `NestedMapRead::{len_map_confirmed, len_map_pending, is_empty_map_confirmed, is_empty_map_pending}`

The logic of the new impls follows the existing `iter_{confirmed, pending}` implementations (e.g. it performs filtering for `MapRead::iter_pending` impls and doesn't for `NestedMapRead::iter_pending`); this means the new methods return the same values as `iter_{confirmed, pending}().count()`, **but** they don't perform any deserialization, which makes them much faster.

Some of the existing test cases were extended to utilize the new methods.

### DIFF
--- a/ledger/store/src/helpers/memory/internal/map.rs
+++ b/ledger/store/src/helpers/memory/internal/map.rs
@@ -243,6 +243,13 @@ impl<
     type Values = core::iter::Map<btree_map::IntoValues<Vec<u8>, V>, fn(V) -> Cow<'a, V>>;
 
     ///
+    /// Returns the number of confirmed entries in the map.
+    ///
+    fn len_confirmed(&self) -> usize {
+        self.map.read().len()
+    }
+
+    ///
     /// Returns `true` if the given key exists in the map.
     ///
     fn contains_key_confirmed<Q>(&self, key: &Q) -> Result<bool>
--- a/ledger/store/src/helpers/memory/internal/nested_map.rs
+++ b/ledger/store/src/helpers/memory/internal/nested_map.rs
@@ -264,6 +264,16 @@ impl<
     type Values = core::iter::Map<btree_map::IntoValues<Vec<u8>, V>, fn(V) -> Cow<'a, V>>;
 
     ///
+    /// Returns the number of confirmed entries in the map.
+    ///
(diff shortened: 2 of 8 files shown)


howardwu:
Can you perf this with 1M, 10M, 100M entries?

ljedrz:
I updated the PR to use the raw iterator which is faster.

While the results could be slightly different with other kinds of records, I did a perf run with 1M, 10M and 100M entries, and the results were more or less linear, and not particularly fast:
```
1M:   88.56ms
10M:  1.145s
100M: 14.062245032s
```
I don't see a more performant way of doing this, though, as the iterator-related APIs are rather limited. I attempted to reduce the number of comparisons, but it's the iteration itself that's the slow part.

ljedrz:
@howardwu BTW, I noticed that we encode the length of the nested map prefix (in `create_prefixed_map`) as a `u32`, but perhaps a `u16` would suffice instead?

ledger/store/src/helpers/memory/internal/map.rs
```diff
@@ -242,6 +242,21 @@ impl<
...
+    ///
+    fn len_confirmed(&self) -> usize {
+        self.map.read().len()
+    }
+
+    ///
+    /// Returns the number of pending entries in the map.
+    ///
+    fn len_pending(&self) -> usize {
+        let filtered_atomic_batch: HashSet<_> = self.atomic_batch.lock().clone().into_iter().map(|(k, _)| k).collect();
```
raychu86: I believe this is insufficient because it doesn't account for the case of removed keys (where the value in the `atomic_batch` is None).

howardwu: I dont think `len_pending` makes semantic sense.

Let's just remove all usage of `len_pending` for now.

howardwu:
@ljedrz can you remove `len_pending` and `is_empty_map_pending` for NestedMap too?

ljedrz:
Rebased and removed the pending-related methods.

howardwu (approved):

PR #2223, opened 2023-11-30
[NCC-EVJ] Use `Committee` stakes when determining the timestamp of a `Subdag`

raychu86:
## Motivation

This PR updates the `Subdag::timestamp` function from a simple median timestamp to a [weighed median](https://en.wikipedia.org/wiki/Weighted_median) timestamp that uses the committee member stakes as the weights.

Because quorum is determined by stake and not the number of validators, we must also use stake to determine the timestamp. 

Audit Finding: **[NCC EVJ] Timestamp Calculation Does Not Provide Byzantine Fault Tolerance**


ledger/narwhal/subdag/src/lib.rs
```diff
@@ -78,6 +79,36 @@ fn sanity_check_subdag_with_dfs<N: Network>(subdag: &BTreeMap<u64, IndexSet<Batc
...
 
+/// Returns the weighted median timestamp of the given timestamps and stakes.
+fn weighted_median(timestamps_and_stake: Vec<(i64, u64)>) -> i64 {
+    let mut timestamps_and_stake = timestamps_and_stake;
+
+    // Sort the timestamps.
+    #[cfg(not(feature = "serial"))]
+    timestamps_and_stake.par_sort_unstable_by_key(|(timestamp, _)| *timestamp);
+    #[cfg(feature = "serial")]
+    timestamps_and_stake.sort_unstable_by_key(|(timestamp, _)| *timestamp);
```
vicsn, line 90: You could make [another one of these macros](https://github.com/AleoHQ/snarkVM/blob/testnet3/utilities/src/parallel.rs#L145), more in line with existing code (I only know it from the algorithms crate)

ledger/narwhal/subdag/src/lib.rs
```diff
@@ -141,20 +172,20 @@ impl<N: Network> Subdag<N> {
...
-                timestamps.par_sort_unstable();
-                #[cfg(feature = "serial")]
-                timestamps.sort_unstable();
-                // Return the median timestamp.
-                timestamps[timestamps.len() / 2]
+                // Retrieve the timestamps and stakes of the certificates.
+                let timestamps_and_stakes = self
+                    .values()
+                    .flatten()
+                    .map(|certificate| (certificate.timestamp(), committee.get_stake(certificate.author())))
```
ljedrz, line 211: should we filter out the elements where the stake is equal to `0`? it would speed up sorting etc.

ledger/narwhal/subdag/src/lib.rs
```diff
@@ -78,6 +79,36 @@ fn sanity_check_subdag_with_dfs<N: Network>(subdag: &BTreeMap<u64, IndexSet<Batc
...
+
+    // Initialize the current timestamp and accumulated stake.
+    let mut current_timestamp: i64 = 0;
+    let mut accumulated_stake: u64 = 0;
+
+    // Find the weighted median timestamp.
+    for (timestamp, stake) in timestamps_and_stake.iter() {
+        accumulated_stake = accumulated_stake.saturating_add(*stake);
+        current_timestamp = *timestamp;
+        if accumulated_stake >= total_stake.saturating_div(2) {
```
randomsleep: This should be:

```rust
accumulated_stake.saturating_mul(2) >= total_stake
```

Please consider this case:

```rust
  let data = vec![(1,1), (2,1), (3,1), (4,1), (5,1)];
  assert_eq!(weighted_median(data), 3);
```

ledger/narwhal/subdag/src/lib.rs
```diff
@@ -141,20 +172,20 @@ impl<N: Network> Subdag<N> {
...
-                // Sort the timestamps.
-                #[cfg(not(feature = "serial"))]
-                timestamps.par_sort_unstable();
-                #[cfg(feature = "serial")]
-                timestamps.sort_unstable();
-                // Return the median timestamp.
-                timestamps[timestamps.len() / 2]
+                // Retrieve the timestamps and stakes of the certificates.
+                let timestamps_and_stakes = self
+                    .values()
```
randomsleep, line 208: Should we use the median of leader_certificate's parents or all certificates? It makes no sense to use the weighted median of all certificates.

ledger/narwhal/subdag/src/lib.rs
raychu86: Great catch, updated to avoid the possible truncation.

ledger/narwhal/subdag/src/lib.rs
raychu86: In theory, there can't be a committee member with 0 stake. This may be a needless filter

ledger/narwhal/subdag/src/lib.rs
```diff
@@ -165,20 +196,20 @@ impl<N: Network> Subdag<N> {
...
+                    .flatten()
+                    .map(|certificate| (certificate.timestamp(), committee.get_stake(certificate.author())))
+                    .collect::<Vec<_>>();
+
+                // Return the weighted median timestamp
```
howardwu:
```suggestion
                // Return the weighted median timestamp.
```

howardwu (approved):

PR #2224, opened 2023-12-01
[Hotfix] Use `execute_function` when checking a deployment.

d0cd:
^


howardwu (approved):
LGTM

ghostant-1017:
I'm confused that why this would cause the network outage? As we have already called `self.ledger.check_next_block` before we advance to the next block, the logic is same in the `BlockSync`.
Why the check in the `BlockSync` failed, but succeeded in `BFT Consensus`? 

Could you explain a bit for me? Thanks a lot! @d0cd @howardwu

PR #2229, opened 2023-12-04
Abort transactions with already spent `InputID`s

raychu86:
## Motivation

This PR adds logic to `VM::atomic_speculate` to track spent `InputID`s and abort transactions that attempt to spend an `InputID` that was already spent in this block. 

Note: We may want to use the `InputStore::get_speculative` instead of per block trackers in `atomic_speculate`.

Related to: https://github.com/AleoHQ/snarkOS/pull/2852

## Test Plan

A test has been added to check that the possible double spends are caught and aborted.

### DIFF
--- a/ledger/src/tests.rs
+++ b/ledger/src/tests.rs
@@ -690,9 +690,26 @@ fn test_execute_duplicate_input_ids() {
     let inputs = [Value::from_str(&format!("{address}")).unwrap(), Value::from_str("100u64").unwrap()];
     let transfer_3 = ledger
         .vm
-        .execute(&private_key, ("credits.aleo", "transfer_public"), inputs.into_iter(), Some(record_1), 0, None, rng)
+        .execute(
+            &private_key,
+            ("credits.aleo", "transfer_public"),
+            inputs.into_iter(),
+            Some(record_1.clone()),
+            0,
+            None,
+            rng,
+        )
         .unwrap();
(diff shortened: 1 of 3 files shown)


ghostant-1017:
It seems check at `atomic_speculate ` is needless, we've already checked at `check_next_block` https://github.com/AleoHQ/snarkVM/blob/testnet3/ledger/block/src/verify.rs#L416

raychu86:
> It seems check at `atomic_speculate ` is needless, we've already checked at `check_next_block` https://github.com/AleoHQ/snarkVM/blob/testnet3/ledger/block/src/verify.rs#L416

The logic change in `atomic_speculate` is not just a check, as it fully aborts subsequent transactions with spent `InputIDs` before creating a block. The check in `check_next_block` is to validate that the final block is well-formed.

ghostant-1017:
> > It seems check at `atomic_speculate ` is needless, we've already checked at `check_next_block` https://github.com/AleoHQ/snarkVM/blob/testnet3/ledger/block/src/verify.rs#L416
> 
> The logic change in `atomic_speculate` is not just a check, as it fully aborts subsequent transactions with spent `InputIDs` before creating a block. The check in `check_next_block` is to validate that the final block is well-formed.

Thank you for explanation! I misunderstood it.

howardwu (approved):

PR #2231, opened 2023-12-05
Derive Ord + PartialOrd to compare Address

vicsn:
## Motivation

I wanted to compare `Address<N>`. Which is not a typed, so we need to also be able to compare its type.

## Related PRs

We also hit this problem in the past when hand implementing `Ord/PartialOrd` for the ProvingKey. If this is approved, I may change it to auto derive them.

### DIFF
--- a/console/network/src/lib.rs
+++ b/console/network/src/lib.rs
@@ -76,6 +76,8 @@ pub trait Network:
     + for<'a> Deserialize<'a>
     + Send
     + Sync
+    + Ord
+    + PartialOrd
 {
     /// The network ID.
     const ID: u16;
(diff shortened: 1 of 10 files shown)


howardwu:
This doesn't make sense for most of the types being operated over I believe. A custom impl should be made.

vicsn:
> This doesn't make sense for most of the types being operated over I believe. A custom impl should be made.

I see.

  Looked into doing a custom `impl Ord for Address<N>` but I fear it wouldn't be very efficient. The Address' inner `Group` exposes e.g. [to_x_coordinate()](https://github.com/AleoHQ/snarkVM/blob/testnet3/console/types/group/src/to_x_coordinate.rs#L19) which would incur field operations and a clone *for every single comparison*.

For [my usecase](https://github.com/AleoHQ/snarkOS/pull/2897), I now decided instead to create and cache the x coordinate once. Will ask in that PR whether we want to merge that construction into snarkVM.

PR #2232, opened 2023-12-05
A handful of SonicKZG10 perf tweaks

ljedrz:
This PR contains a few uncontroversial perf tweaks I've found while browsing the `sonic_pc` code.


ljedrz:
~I wasn't expecting a difference in serialization, I'll double-check it :microscope:.~

edit: fixed, it was an arithmetic error

vicsn (approved):
Neat. The only downside I see is that `range` silently ignores indices which are too big, so we don't throw the error anymore. But as this is the prover-side, the worst this can do is hinder our internal debugging.

ljedrz:
I'm guessing you mean the second instance (as in the first one the indexing suggests that it can't fail)? If it's desirable, perhaps I can add some single check for length there?

vicsn:
> I'm guessing you mean the second instance (as in the first one the indexing suggests that it can't fail)? If it's desirable, perhaps I can add some single check for length there?

Indeed. What makes it even sillier is that our hiding bound is always 1, so someone has to do a big booboo for that to overflow. You can add the check anyway and then we don't have to think about it anymore.

ljedrz:
Done and rebased. [These](https://github.com/AleoHQ/snarkVM/pull/2232/files#diff-1c2269e15b2c9eb1b102866e39b60f7149b01d43503a3bdd02081fc7200a358cR126-R130) are the only new changes.

vicsn (approved):

howardwu (approved):
LGTM

PR #2234, opened 2023-12-05
Implement canonical (de)serialization for Arc<[T]>

ljedrz:
This is something we may want to use eventually; I haven't found a good spot for it just yet, but it will be practical then.

### DIFF
--- a/utilities/src/serialize/impls.rs
+++ b/utilities/src/serialize/impls.rs
@@ -345,6 +345,45 @@ impl<T: CanonicalDeserialize + ToOwned + Sync + Send> CanonicalDeserialize for A
     }
 }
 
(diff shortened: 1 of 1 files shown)


howardwu:
Is there anything immediate that requires this? I'm trying to deprecate usage of `Canonical*` given the safety issues that we've ran into several times with the design (primarily on skipped checks that are implicit in its flagged design, we've opted for explicit checks using To/FromBytes which are IMO safer).

ljedrz:
If the plan is to deprecate them, this PR can be closed and the next time I need to serialize `Arc<[T]>`, I'll propose that the related object  be turned into `ToBytes`/`FromBytes` instead 👍.

PR #2235, opened 2023-12-05
A handful of PowersOfG tweaks

ljedrz:
This PR contains 2 changes:
1. it adjusts the locking and `Arc`ing around `PowersOfG` so that the lock only applies to the actually mutable bits, and the `Arc`s to the objects that get cloned
2. there was a pre-allocation that was not actually utilized, which is now fixed

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -3371,6 +3371,7 @@ dependencies = [
  "itertools 0.11.0",
  "js-sys",
  "lazy_static",
+ "parking_lot",
  "paste",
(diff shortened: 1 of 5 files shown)


ljedrz:
~These were cherry-picked so I missed it, but these changes depend on https://github.com/AleoHQ/snarkVM/pull/2233.~

Ready for review.

howardwu (approved):
LGTM

raychu86 (approved):

PR #2240, opened 2023-12-07
Reduce max RAM when verifying deployment transactions

ljedrz:
Using a subset of the `sir_hashalot` program containing 4096 transitions and running up until (and including) verification, I get the following improvements with this PR:

baseline:
```
Verification took: 22.020414433s
number of allocations: 113,112,661
max RAM: 8.20 GiB
```

this PR:
```
Verification took: 21.518246351s
number of allocations: 101,757,273 (-10%)
max RAM: 6.60 GiB (-20%)
```

There is one issue, though; despite getting a consistent, small performance improvement when running the test program, I'm seeing a small (~4%) degradation in the `Transaction::Deploy - verify` benchmark, and I can't tie it to any of the commits belonging to this PR (with the only real potential candidate being https://github.com/AleoHQ/snarkVM/commit/0053471d84dffa91722a55c183e92aa23374c3dc). I'd appreciate if someone else could also give it a go to check if they observe it locally as well.

### DIFF
--- a/algorithms/src/snark/varuna/ahp/indexer/constraint_system.rs
+++ b/algorithms/src/snark/varuna/ahp/indexer/constraint_system.rs
@@ -12,10 +12,7 @@
 // See the License for the specific language governing permissions and
 // limitations under the License.
 
-use crate::{
-    r1cs::{errors::SynthesisError, ConstraintSystem as CS, Index as VarIndex, LinearCombination, Variable},
-    snark::varuna::ahp::matrices::to_matrix_helper,
-};
+use crate::r1cs::{errors::SynthesisError, ConstraintSystem as CS, Index as VarIndex, LinearCombination, Variable};
 use snarkvm_fields::Field;
 use snarkvm_utilities::serialize::*;
 
@@ -42,24 +39,6 @@ impl<F: Field> ConstraintSystem<F> {
             num_private_variables: 0,
             num_constraints: 0,
         }
-    }
-
-    #[inline]
-    /// Returns the sparse A matrix as Vec of rows, where each row is a Vec of assigned value and variable index
-    pub(crate) fn a_matrix(&self) -> Result<Vec<Vec<(F, usize)>>> {
-        to_matrix_helper(&self.a, self.num_public_variables)
-    }
-
-    #[inline]
-    /// Returns the sparse B matrix as Vec of rows, where each row is a Vec of assigned value and variable index
-    pub(crate) fn b_matrix(&self) -> Result<Vec<Vec<(F, usize)>>> {
-        to_matrix_helper(&self.b, self.num_public_variables)
-    }
-
-    #[inline]
-    /// Returns the sparse C matrix as Vec of rows, where each row is a Vec of assigned value and variable index
-    pub(crate) fn c_matrix(&self) -> Result<Vec<Vec<(F, usize)>>> {
-        to_matrix_helper(&self.c, self.num_public_variables)
     }
 
     #[inline]
--- a/algorithms/src/snark/varuna/ahp/indexer/indexer.rs
+++ b/algorithms/src/snark/varuna/ahp/indexer/indexer.rs
@@ -15,13 +15,13 @@
 use crate::{
     fft::EvaluationDomain,
     polycommit::sonic_pc::{LinearCombination, PolynomialInfo, PolynomialLabel},
-    r1cs::{errors::SynthesisError, ConstraintSynthesizer, ConstraintSystem},
+    r1cs::{errors::SynthesisError, ConstraintSynthesizer},
     snark::varuna::{
         ahp::{
             indexer::{Circuit, CircuitId, CircuitInfo, ConstraintSystem as IndexerConstraintSystem},
             AHPForR1CS,
         },
-        matrices::{matrix_evals, MatrixEvals},
+        matrices::{into_matrix_helper, matrix_evals, MatrixEvals},
         num_non_zero,
         SNARKMode,
     },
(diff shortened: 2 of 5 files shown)


algorithms/src/snark/varuna/ahp/matrices.rs
```diff
@@ -37,19 +36,25 @@ use rayon::prelude::*;
...
-                    VarIndex::Public(i) => *i,
+                    VarIndex::Public(i) => i,
                     VarIndex::Private(i) => num_input_variables + i,
                 };
-                *row_map.entry(column).or_insert_with(F::zero) += *val;
+                match row_map.binary_search_by_key(&column, |(c, _)| *c) {
+                    Ok(idx) => row_map[idx].1 += val,
+                    Err(idx) => {
+                        row_map.insert(idx, (column, F::zero()));
+                        row_map[idx].1 += val;
```
ljedrz:
```suggestion
                        row_map.insert(idx, (column, val));
```
question: could we do this instead?

d0cd:
No performance degradation for `Transaction::Deploy - verify` on a 2021 MBP M1 MAX 32 GB RAM

algorithms/src/snark/varuna/ahp/matrices.rs
vicsn: Yes that would be equivalent.

ljedrz: sweet, because it should be a bit faster 🔥

vicsn (approved):
20% huh, awesome. Looks good. can you explain why the Arc is needed in the (pen)ultimate commit?

ljedrz:
> can you explain why the Arc is needed in the (pen)ultimate commit?

That one is a smaller win (at least with this program), but there is one spot where we clone the `Assignment`, and since we don't modify its contents, we can do a shallow clone instead.

ljedrz:
The CI failure is unrelated, that error pops up randomly from time to time.

ljedrz:
I just realized that there's another easy win here: https://github.com/AleoHQ/snarkVM/pull/2240/commits/5cfa348b2e184b6b74bcf9c538efc92c4f5908b7.

vicsn (approved):
Weirdly, I do consistently get (statistically insignificant) performance degradations of e.g. +1.0293%.

d0cd (approved):

algorithms/src/snark/varuna/ahp/indexer/constraint_system.rs
```diff
@@ -46,20 +47,20 @@ impl<F: Field> ConstraintSystem<F> {
...
+    pub(crate) fn b_matrix(&mut self) -> Result<Vec<Vec<(F, usize)>>> {
+        to_matrix_helper(mem::take(&mut self.b), self.num_public_variables)
     }
 
     #[inline]
     /// Returns the sparse C matrix as Vec of rows, where each row is a Vec of assigned value and variable index
-    pub(crate) fn c_matrix(&self) -> Result<Vec<Vec<(F, usize)>>> {
-        to_matrix_helper(&self.c, self.num_public_variables)
+    pub(crate) fn c_matrix(&mut self) -> Result<Vec<Vec<(F, usize)>>> {
+        to_matrix_helper(mem::take(&mut self.c), self.num_public_variables)
```
howardwu: This should be reverted, because it is actually **taking** the items now.

While in theory this is "safe" from usage, it is not the desired behavior of this interface.

ljedrz: hmm, could we introduce and use a new method that allows us to take `self` by value and deconstruct it, something like a
```
pub(crate) fn into_abc_matrices(self) -> Result<(Matrix<F>, Matrix<F>, Matrix<F>)>
```
while leaving the existing methods intact?

vicsn: I see that after we call `ics.{a,b,c}_matrix()`, we still need to extract a few more members. Can you follow the same approach as we use for the prover? https://github.com/AleoHQ/snarkVM/blob/testnet3/algorithms/src/snark/varuna/ahp/prover/round_functions/mod.rs#L101

ljedrz: done and rebased

algorithms/src/snark/varuna/ahp/indexer/indexer.rs
```diff
@@ -140,15 +140,15 @@ impl<F: PrimeField, SM: SNARKMode> AHPForR1CS<F, SM> {
...
+
+        let a = into_matrix_helper(a, num_public_variables)?;
+        let b = into_matrix_helper(b, num_public_variables)?;
+        let c = into_matrix_helper(c, num_public_variables)?;
 
         end_timer!(padding_time);
 
-        let num_padded_public_variables = ics.num_public_variables();
-        let num_private_variables = ics.num_private_variables();
-        let num_constraints = ics.num_constraints();
```
howardwu, line 151: Move these lines to inside the `if` statement below, otherwise debug mode will not compile.

ljedrz: These were only removed because they are already bound with the same names in [L143](https://github.com/AleoHQ/snarkVM/pull/2240/files#diff-308e0d2507e765d9aa886da9e3edd2a70deef59ef5d794aa10b842f2974e5051R143); I checked in debug mode and it compiles just fine.

algorithms/src/snark/varuna/varuna.rs
```diff
@@ -113,7 +113,7 @@ impl<E: PairingEngine, FS: AlgebraicSponge<E::Fq, 2>, SM: SNARKMode> VarunaSNARK
             ensure!(commitment_randomnesses.iter().all(|r| r == &empty_randomness));
             end_timer!(commit_time);
 
-            circuit_commitments.sort_by(|c1, c2| c1.label().cmp(c2.label()));
```
howardwu: Convince me this is safe. We use the `circuit_commitments` ordering in `Poseidon` for hashing.

ljedrz: this was just a drive-by, so I'm happy to drop this one commit; however, changing the sort from the default (stable) to unstable only results in a difference if the order of the _equal_ elements matters, which is rare

ljedrz:
Rebased and dropped the sort adjustment drive-by.

howardwu (approved):

PR #2254, opened 2023-12-18
Update `Subdag` equality check

raychu86:
## Motivation

This PR implements PartialEq and Eq for `Subdag`. The new change enforces equality only on the subdag component, and skips the check for the `election_certificate_ids`.

### DIFF
--- a/ledger/narwhal/subdag/src/lib.rs
+++ b/ledger/narwhal/subdag/src/lib.rs
@@ -78,13 +78,21 @@ fn sanity_check_subdag_with_dfs<N: Network>(subdag: &BTreeMap<u64, IndexSet<Batc
     &commit == subdag
 }
 
-#[derive(Clone, PartialEq, Eq)]
+#[derive(Clone)]
 pub struct Subdag<N: Network> {
     /// The subdag of round certificates.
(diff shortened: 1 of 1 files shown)


ledger/narwhal/subdag/src/lib.rs
```diff
@@ -78,14 +78,22 @@ fn sanity_check_subdag_with_dfs<N: Network>(subdag: &BTreeMap<u64, IndexSet<Batc
...
 pub struct Subdag<N: Network> {
     /// The subdag of round certificates.
     subdag: BTreeMap<u64, IndexSet<BatchCertificate<N>>>,
     /// The election certificate IDs.
     election_certificate_ids: IndexSet<Field<N>>,
 }
 
+impl<N: Network> PartialEq for Subdag<N> {
+    fn eq(&self, other: &Self) -> bool {
+        self.subdag == other.subdag
```
howardwu, line 92:
```suggestion
        // Note: We do not check equality on `election_certificate_ids` as it would cause `Block::eq` to trigger false-positives.
        self.subdag == other.subdag
```

howardwu (approved):

PR #2255, opened 2023-12-18
Add check that combiners.len() == sorted_evals.len()

### LINKED ISSUE #2251: Check safety for `combiners` to be longer than `sorted_evals`?
Is it safe for `combiners` to be longer than `sorted_evals`?

If not, we should add an `ensure!` on the lengths.

_Originally posted by @howardwu in https://github.com/AleoHQ/snarkVM/pull/2203#discussion_r1429313975_

vicsn:
## Motivation

Closes https://github.com/AleoHQ/snarkVM/issues/2251
Because we length of the iterator is unknown, we can only check if it is empty after it has been used.


ljedrz (approved):
LGTM :ok_hand:

howardwu (approved):

PR #2256, opened 2023-12-18
Optimize the split into (un)verified solutions

ljedrz:
This PR is a follow-up to https://github.com/AleoHQ/snarkVM/pull/2250 that _limits_ the verification of further solutions once the protocol maximum is reached. In theory - since the solutions are being processed in parallel - this logic can still result in a few calls to `ProverSolution::verify` being performed beyond the `N::MAX_SOLUTIONS` limit, but since we can't known if the next candidate is valid in advance, we can't go around it. Nevertheless, this change should result in a tangible performance improvement whenever the number of candidate solutions is greater than `N::MAX_SOLUTIONS`.

### DIFF
--- a/ledger/src/advance.rs
+++ b/ledger/src/advance.rs
@@ -108,7 +108,7 @@ impl<N: Network, C: ConsensusStorage<N>> Ledger<N, C> {
 }
 
 /// Splits candidate solutions into a collection of accepted ones and aborted ones.
-fn split_candidate_solutions<T, F>(
+pub fn split_candidate_solutions<T, F>(
     mut candidate_solutions: Vec<T>,
     max_solutions: usize,
     verification_fn: F,
--- a/ledger/src/advance.rs
+++ b/ledger/src/advance.rs
@@ -115,7 +115,7 @@ impl<N: Network, C: ConsensusStorage<N>> Ledger<N, C> {
         previous_block: &Block<N>,
         subdag: Option<&Subdag<N>>,
         candidate_ratifications: Vec<Ratify<N>>,
-        candidate_solutions: Vec<ProverSolution<N>>,
+        mut candidate_solutions: Vec<ProverSolution<N>>,
         candidate_transactions: Vec<Transaction<N>>,
     ) -> Result<(Header<N>, Ratifications<N>, Option<CoinbaseSolution<N>>, Transactions<N>, Vec<N::TransactionID>)>
     {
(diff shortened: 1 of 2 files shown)


ledger/src/advance.rs
```diff
@@ -127,30 +129,46 @@ impl<N: Network, C: ConsensusStorage<N>> Ledger<N, C> {
...
+                    // If there is still room for solutions, verify the next candidate.
+                    let verified = solution
+                        .verify(coinbase_verifying_key, &latest_epoch_challenge, self.latest_proof_target())
+                        .unwrap_or(false);
+
+                    match verified {
+                        true => {
+                            // Since the solutions are processed in parallel, make sure that the
+                            // collection of the valid ones is still below limit before pushing.
+                            let mut valid_candidate_solutions = valid_candidate_solutions.write();
```
ljedrz: note: I used `write` instead of a `read` and a conditional `write`, as we're expecting to be below the limit due to the earlier ([L141](https://github.com/AleoHQ/snarkVM/pull/2256/files#diff-bfbe2e2f8e1e6a889781c83d0ae44ba3dde3604d9415aaf27921d7c94acd02c6R141)) length check

joske (approved):

howardwu (approved):

howardwu:
@ljedrz Moving this PR back to draft.

It occurred to me that this code is not deterministic. As such, it will cause a network partition if nodes differ in their Rayon configs.

ljedrz:
Rebased and changed the logic so that the end result is deterministic.

This approach is not as aggressive as the initial one, so there may be a few more calls to `verify` than needed, but it is still an improvement, and the potential extra work is restricted.

ledger/src/advance.rs
```diff
@@ -127,26 +127,48 @@ impl<N: Network, C: ConsensusStorage<N>> Ledger<N, C> {
...
-                    })
-                    .collect();
+                // TODO: For mainnet - Add `aborted_solution_ids` to the block.
                 // Separate the candidate solutions into valid and aborted solutions.
                 let mut valid_candidate_solutions = Vec::with_capacity(N::MAX_SOLUTIONS);
                 let mut aborted_candidate_solutions = Vec::new();
-                for (solution, is_valid) in verification_results.into_iter() {
-                    if is_valid && valid_candidate_solutions.len() < N::MAX_SOLUTIONS {
-                        valid_candidate_solutions.push(solution);
+                // Reverse the candidate solutions in order to be able to chunk them more efficiently.
```
howardwu, line 134:
```suggestion
                // Reverse the candidate solutions in order to be able to chunk them more efficiently.
                // Note: Reversing the order here preserves FIFO processing across rounds.
```
@ljedrz Can you confirm this has no impact to ordering from Narwhal across rounds?

ledger/src/advance.rs
```diff
@@ -127,26 +127,48 @@ impl<N: Network, C: ConsensusStorage<N>> Ledger<N, C> {
...
+                candidate_solutions.reverse();
+                // Verify the candidate solutions in chunks. This is done so that we can potentially
+                // perform these operations in parallel while keeping the end result deterministic.
+                let chunk_size = 16;
```
howardwu: How was 16 chosen?

howardwu (approved):

ledger/src/advance.rs
```diff
@@ -127,26 +127,48 @@ impl<N: Network, C: ConsensusStorage<N>> Ledger<N, C> {
...
+                while !candidate_solutions.is_empty() {
+                    // Check if the collection of valid solutions is full.
+                    if valid_candidate_solutions.len() == N::MAX_SOLUTIONS {
```
howardwu:
```suggestion
                    if valid_candidate_solutions.len() >= N::MAX_SOLUTIONS {
```
Out of an abundance of caution

ledger/src/advance.rs
ljedrz: this number should be smaller than `N::MAX_SOLUTIONS`, large enough to not impede parallelism, and small enough to not negatively affect performance due to redundant verifications

ljedrz:
Rebased and addressed the suggestion.

howardwu:
Should we abstract this logic in order to make it testable? If we abstract the logic with a generic `T`, then we can check for cases where the number of entries are:
1. empty
2. within "MAX_SOLUTIONS"
3. exceeds "MAX_SOLUTIONS"

ljedrz:
I'm not sure if abstracting this would be comparable, as some of the internals of verification are likely parallelized too :thinking:.

howardwu:
It's mainly to abstract away this optimization, so the reader can black box see there is a batch that we're checking for admission and aborting the remainder.

The benefit is to be able to test that the logic behaves correctly for different number of solutions.

ljedrz:
Done; the PR was also rebased.

raychu86 (approved):

PR #2268, opened 2023-12-28
[Fix] Declare public variable 1field once

### LINKED ISSUE #2266: [Bug] Only declare the public variable 1field once
## 🐛 Bug Report

See this PR for context: https://github.com/AleoHQ/snarkVM/pull/2110

vicsn:
## Motivation

Closes https://github.com/AleoHQ/snarkVM/issues/2266

We were declaring two hardcoded public One variables (in R1CS and ConstraintSystem), and adding them twice (in `to_transition_verifier_inputs` and in `verify_batch`. This PR:
- skips adding the hardcoded variable from R1CS onto the ConstraintSystem
- removes adding the hardcoded input in `verify_batch`
- adds logic to remove `LinearCombination` terms with value 0. Similar to https://github.com/AleoHQ/snarkVM/pull/1796 but for a different object doing the same thing.

## Test Plan

No new tests were added.

Requires resampling credits and inclusion.

## Related PRs

Previous attempt: https://github.com/AleoHQ/snarkVM/pull/2110/


Pratyush:
This looks good to me, modulo an additional check in `generate_constraints` that `ConstraintSystem` allocates a one variable.

vicsn:
> This looks good to me, modulo an additional check in generate_constraints that ConstraintSystem allocates a one variable.

Thank you for the suggestion! I looked into it again and think we have a good setup already (avoiding further refactors):
1. **Verifier**: Most importantly, [the verifier already checks that the first public variable is a one](https://github.com/AleoHQ/snarkVM/pull/2268/files#diff-4f77dc7c82f9259422fc4e03c61ed1542f14047d2b628a9106492686695988b5R676)
2. **Prover**: Indeed clearly [pushes the one input variable into the converter](https://github.com/AleoHQ/snarkVM/pull/2268/files#diff-4a35a59bc6a3e5d46ae4e15cc355d642e768e966b6ba998092e3ec42f75e6e87R169). Meanwhile the `cs: ConstraintSystem` trait doesn't give access to its variables barring other checks on that.

There is a lot of complexity, one day, I hope we can cut out some parts from the `R1CS->Assignment->Converter->ConstraintSystem->[Matrix;3]` chain.

howardwu (approved):

PR #2270, opened 2023-12-29
Cache transaction checks

### LINKED ISSUE #2855: [Feature] Reduce duplicate tx verification
## 🚀 Feature

Currently validators call `Ledger::<N, C>::check_transaction_basic(...)` twice when receiving executions via the P2P network or RPC endpoint and thrice when receiving deployments.

## Motivation

It would be great if we can cache whether deployment synthesis and execution verification succeeded. Only inclusion proofs may become stale and needs to be rechecked at the end.

ljedrz:
This PR aims to greatly reduce the probability that we perform a duplicate check on a transaction. I'm filing it as a draft, as there are several possible ways of approaching this, and I'm not 100% sure this is the best one.

Some considerations:
- the entire `Transaction` is cached (in addition to the `TransactionID`) in order to allow the new entries to be fully compared; this wouldn't be necessary if we could verify that a given `TransactionID` truly belongs to the given `Transaction`, i.e. _compute_ it instead of retrieving its pre-computed value; if we could do that, the cache could be much smaller
- only valid checks are cached; this is done for several reasons:
  - to prevent a maliciously crafted invalid transaction from causing a valid one to be rejected if it arrives later; I'd consider this very improbable, though (as someone would have to know the `TransactionID` of the targeted transaction in advance)
  - to avoid a coordinated, malicious influx of invalid transactions from evicting the valid ones from the cache (which would cause us to perform duplicate checks as if we didn't have one)
  - to avoid situations where a transaction that would have been deemed valid if it arrived a second later instead becomes "blacklisted" until it's evicted from the cache

Fixes https://github.com/AleoHQ/snarkOS/issues/2855.


ledger/src/lib.rs
```diff
@@ -105,6 +109,8 @@ pub struct Ledger<N: Network, C: ConsensusStorage<N>> {
     current_committee: Arc<RwLock<Option<Committee<N>>>>,
     /// The current block.
     current_block: Arc<RwLock<Block<N>>>,
+    /// A cache containing the list of recent succesfully checked transactions.
+    checked_transactions: Arc<RwLock<LruCache<N::TransactionID, Transaction<N>>>>,
```
raychu86: We may want to purge the cached transaction once they are included in a block.

ledger/src/lib.rs
```diff
@@ -67,14 +67,18 @@ use aleo_std::prelude::{finish, lap, timer};
...
 use rand::{prelude::IteratorRandom, rngs::OsRng};
-use std::{borrow::Cow, sync::Arc};
+use std::{borrow::Cow, num::NonZeroUsize, sync::Arc};
 use time::OffsetDateTime;
 
 #[cfg(not(feature = "serial"))]
 use rayon::prelude::*;
 
+/// The number of transactions to keep cached in order to not perform redundant checks.
+const NUM_CACHED_TRANSACTIONS: usize = 128;
```
raychu86: How was the number of cached transactions determined?

ledger/src/lib.rs
raychu86: We also likely don't need to cache verification of execution transactions. Because it's a simple proof check (~20ms), which is likely not worth storing another tx in memory and taking up a cache spot. 

In addition, if a node receives 1 deployment and then 128 executions, the deployment verification (which is the important one) gets purged from the cache.

vicsn: I think it is safe to only store the transaction_id in the cache. It should be deterministic from the Transaction. We check this on deserialization and it is the second check inside `check_transaction`.

As for whether this is desired, let's assume the cache grows as big as five blocks worth of transactions. In theory blocks can have at most (`MAX_TRANSACTIONS`) 65536 transactions. In practice this might be up to 600 executions, or maybe a few deployments, due to slow processing times. So we're looking at a size of between 5x600 and 5x65536 transactions. If we only store IDs (32 bytes) this is between 100KB and 10MB. If we store the transactions (e.g. 1.5 KB execution), that would be between 4MB and 491 MB. 

4MB actually seems pretty doable, so if we get a big performance gain and this is a "hot path", we could choose to store transactions.

I would however advocate for also storing Executions, but this can also be in a future PR.

acoglio: Yes, based on a recent Slack discussion, my understanding is that the ID of a transaction is cryptographically calculated as part of deserializing the transaction (https://github.com/AleoHQ/snarkVM/blob/0f324520e22d9af9d855be41893355f41a1efc63/ledger/block/src/transaction/bytes.rs#L63), as soon as the transaction enters snarkOS (e.g. via a message). In other words, it should be an invariant of the transaction data structure that the ID is cryptographically tied to the rest of the transaction, and therefore that transaction IDs are (practically) unique. This is a critical property to justify the handling of transactions by their IDs in various parts of snarkOS.

ledger/src/lib.rs
ljedrz: it was a placeholder; in any case, as long as we only need to keep the transaction IDs, it could be as large as the maximum number of transactions per block.

ljedrz:
I addressed the comments:
- only the `TransactionId`s are cached
- the cache size is now equal to `MAX_TRANSACTIONS`
- advancing to the next block removes the corresponding transaction IDs from the cache

vicsn:
Great progress! To keep things simple, I think the following checks only have to be done once:
- [Stack::verify_deployment(...)](https://github.com/AleoHQ/snarkVM/blob/testnet3/synthesizer/process/src/verify_deployment.rs#L36). It would be even cleaner if we move the [process.contains_program(program_id)](https://github.com/AleoHQ/snarkVM/blob/testnet3/synthesizer/process/src/verify_deployment.rs#L29) check up and then we would be able to skip the entire process.verify_deployment or even [check_deployment_internal(...)](https://github.com/AleoHQ/snarkVM/blob/testnet3/synthesizer/src/vm/verify.rs#L116).
- [process.verify_execution](https://github.com/AleoHQ/snarkVM/tree/testnet3/synthesizer/src/vm/verify.rs#L236)

I can also imagine it would be nice if we have some kind of symmetry and we skip in both paths at the same level of abstraction.

synthesizer/src/vm/verify.rs
```diff
@@ -95,6 +96,12 @@ impl<N: Network, C: ConsensusStorage<N>> VM<N, C> {
         // First, verify the fee.
         self.check_fee(transaction, rejected_id)?;
 
+        // If the transaction has already been verified (as indicated by the transaction cache), skip verification.
+        if already_verified {
```
vicsn: For deployments, I think the following similar checks cannot be cached / can grow stale:
- `self.transaction_store().contains_program_id(deployment.program_id())` (in synthesizer/src/vm/verify.rs)
- `self.contains_program(program_id)` (in synthesizer/process/src/verify_deployment)

For executions, I think the following cannot be cached / can grow stale:
- `self.block_store().contains_rejected_deployment_or_execution_id(&execution_id)` (in synthesizer/src/vm/verify.rs)
- `self.block_store().contains_state_root(&execution.global_state_root())` (in synthesizer/src/vm/verify.rs)

So I think we should pass the `already_verified` flag down into `check_{deployment, execution}_internal`

synthesizer/src/vm/verify.rs
```diff
@@ -39,6 +39,7 @@ impl<N: Network, C: ConsensusStorage<N>> VM<N, C> {
         transaction: &Transaction<N>,
         rejected_id: Option<Field<N>>,
         rng: &mut R,
+        already_verified: bool,
```
raychu86: The addition of the `already_verified` flag feels unsafe. This make it incredibly easy to accidentally or intentionally skip a transaction verification.

raychu86: I feel like this conditional logic should only exist in `check_transaction_basic`, and not low level in the VM.

ljedrz: @raychu86 hmm, how about using type safety instead? i.e. have a wrapper struct like `VerifiedTransaction { transaction }` or using generics like `Transaction<N: Network, V: VerificationStatus>`?

raychu86:
@ljedrz  It's not ideal, but we could move the cache inside the VM rather than the Ledger, since only the VM needs to reason about it. This would remove the need to pass through variable flags/wrapper structs and leave the Ledger api mostly untouched.

ledger/src/advance.rs
```diff
@@ -87,6 +87,13 @@ impl<N: Network, C: ConsensusStorage<N>> Ledger<N, C> {
         let mut current_block = self.current_block.write();
         // Update the VM.
         self.vm.add_next_block(block)?;
+        // Clear the verified transactions cache.
+        let verified_transactions = self.vm().verified_transactions();
+        let mut transaction_cache = verified_transactions.write();
+        for transaction in block.transactions().iter() {
```
vicsn: I think we'll also need to pop the aborted_tx_ids in the block

synthesizer/src/vm/verify.rs
```diff
@@ -229,11 +242,11 @@ impl<N: Network, C: ConsensusStorage<N>> VM<N, C> {
...
     /// use `VM::check_transaction` instead.
     #[inline]
-    fn check_execution_internal(&self, execution: &Execution<N>) -> Result<()> {
+    fn check_execution_internal(&self, execution: &Execution<N>, already_verified: bool) -> Result<()> {
         let timer = timer!("VM::check_execution");
 
-        // Verify the execution.
-        let verification = self.process.read().verify_execution(execution);
+        // Verify the execution if it hasn't been done yet.
+        let verification = if already_verified { Ok(()) } else { self.process.read().verify_execution(execution) };
```
vicsn: This means we will skip the state root check if `already_verified`, please check all cases in this comment again: https://github.com/AleoHQ/snarkVM/pull/2270#discussion_r1440240416

vicsn: Nevermind I see we don't return

ledger/src/advance.rs
vicsn: Also, I now learned rejected transactions have a change in id. I think you may need to call `to_unconfirmed_transaction_id`/`to_unconfirmed_id` on the transaction first.

ljedrz: I'm not sure how to tackle those; the only txs that enter the cache are those that pass the transaction check, aren't aborted ones those that don't pass it (in which case they'd never enter the cache)?

vicsn: So to be clear, there is a difference between `rejected` and `aborted` transactions, and both are "created" just before a block is made in a call to `vm.speculate`

- `rejected` transactions are still included in the `transactions()` list, they failed during speculation and are turned into a Fee transaction. And we need to call `to_unconfirmed_transaction_id(...)/to_unconfirmed_id(...)` to get the original id. I think we can just call the function on all transactions.
- `aborted` transactions fail during speculation, but on top of that we also fail to turn them into a Fee. E.g. because the user doesn't have any funds left because previous transactions in that same Block already use up all of their balance. We can get the list via `block.aborted_transaction_ids(...)`.

ljedrz: done in 07ee4015327e08820000ee2e4d9ea5e181e3e458

vicsn (approved):
LGTM. The extra `already_verified` could still be avoided if we incur the performance hit of reading `self.verified_transactions` again inside `check_execution_internal`. Let's see what Ray prefers.

synthesizer/src/vm/mod.rs
```diff
@@ -76,6 +80,8 @@ pub struct VM<N: Network, C: ConsensusStorage<N>> {
     atomic_lock: Arc<Mutex<()>>,
     /// The lock for ensuring there is no concurrency when advancing blocks.
     block_lock: Arc<Mutex<()>>,
+    /// A cache containing the list of recent succesfully verified transactions.
+    verified_transactions: Arc<RwLock<LruCache<N::TransactionID, ()>>>,
```
howardwu: Can we use a lazy static instead?

ledger/src/advance.rs
```diff
@@ -87,6 +87,18 @@ impl<N: Network, C: ConsensusStorage<N>> Ledger<N, C> {
...
+        let mut transaction_cache = verified_transactions.write();
+        for transaction_id in block
+            .transactions()
+            .iter()
+            .filter_map(|tx| tx.to_unconfirmed_transaction_id().ok())
+            .chain(block.aborted_transaction_ids().iter().copied())
+        {
+            transaction_cache.pop(&transaction_id);
+        }
+        drop(transaction_cache);
```
howardwu: Can you abstract this into its own method below for readability please?

synthesizer/src/vm/mod.rs
ljedrz: Rust doesn't allow generic statics, and we're generic over the `Network` here

howardwu: Good callout, thanks for pointing that out.

ledger/src/advance.rs
```diff
@@ -105,6 +107,20 @@ impl<N: Network, C: ConsensusStorage<N>> Ledger<N, C> {
 
         Ok(())
     }
+
+    /// Updates the verified transactions cache based on the new block.
+    fn update_transaction_cache(&self, block: &Block<N>) {
```
howardwu: I am removing this, because it occurred to me we should not pop the admitted transaction IDs.

In Bullshark/Narwhal, it is expected that the transaction IDs will likely reappear a few more rounds due to the nature of the DAG. So we want to save on those costs by having checked this portion already.

howardwu: Removed in 62d01b9

synthesizer/process/src/verify_deployment.rs
```diff
@@ -23,10 +23,6 @@ impl<N: Network> Process<N> {
         rng: &mut R,
     ) -> Result<()> {
         let timer = timer!("Process::verify_deployment");
-        // Retrieve the program ID.
```
howardwu, line 26: This cannot be removed from here.

Local processes never use the VM to check, so this would skip a critical check.

howardwu: Reverted in [7660096](https://github.com/AleoHQ/snarkVM/pull/2270/commits/7660096acfbf4d47aa175a3ff502382b830f6f10)

ledger/src/advance.rs
howardwu: @ljedrz

synthesizer/process/src/verify_deployment.rs
howardwu: @ljedrz

howardwu (approved):
LGTM

PR #2271, opened 2023-12-29
[HackerOne-2300725] Limit the number of allowed constraints for deployments

vicsn:
## Motivation

Large deployments may take a lot of time to synthesize, so as an initial stopgap we can limit their size, as measurement by the number of constraints. 

Note that the current approach panics on failure, but because check_transaction_basic is called from snarkOS in a tokio task, this panic is gracefully handled. Returning a `Result` would cascade to 100 other functions using the `Circuit` api.

## Test Plan

- [x] Manually test deploying to a validator via RPC. Prevents large deployments succesfully. Even with a manipulated verifying_key.
- [x] Check that the synthesizer benchmarks don't take a big hit
- [x] Add a test deploying a program which is too big.

## Related PRs
https://github.com/AleoHQ/snarkOS/pull/2961


circuit/environment/src/circuit.rs
```diff
@@ -146,6 +147,13 @@ impl Environment for Circuit {
             // Ensure we are not in witness mode.
             if !in_witness.get() {
                 CIRCUIT.with(|circuit| {
+                    // Ensure we do not surpass maximum allowed number of constraints
+                    MAX_NUM_CONSTRAINTS.with(|max_constraints| {
+                        if circuit.borrow().num_constraints() > max_constraints.get() {
+                            Self::halt("Surpassing maximum allowed number of constraints")
+                        }
+                    });
+
```
ljedrz: shouldn't this check happen at the end of this function instead? I may be reading this wrong, but is seems to me that having this at the beginning could allow the following to happen:
- the check for max number of constraints passes
- new constraints are introduced, and they breach the limit
- the program is halted only if `enforce` is called again afterwards

circuit/environment/src/circuit.rs
```diff
@@ -305,11 +317,18 @@ impl Environment for Circuit {
         })
     }
 
+    /// Sets a maximum amount of allowed constraints
+    fn set_constraint_maximum(new_max_num_constraints: u64) {
```
ljedrz: maybe we could reuse this function within the other functions here? I think it would be more easy to read

circuit/environment/src/circuit.rs
```diff
@@ -305,11 +317,18 @@ impl Environment for Circuit {
         })
     }
 
+    /// Sets a maximum amount of allowed constraints
```
ljedrz:
```suggestion
    /// Sets a maximum number of allowed constraints.
```
(uber-nit)

ledger/block/src/transaction/deployment/mod.rs
```diff
@@ -124,6 +124,11 @@ impl<N: Network> Deployment<N> {
         &self.verifying_keys
     }
 
+    /// Returns the total number of constraints.
+    pub fn num_constraints(&self) -> u64 {
+        self.verifying_keys.iter().map(|(_, (vk, _))| vk.circuit_info.num_constraints).sum::<usize>() as u64
```
ljedrz:
```suggestion
        self.verifying_keys.iter().map(|(_, (vk, _))| vk.circuit_info.num_constraints as u64).sum::<u64>()
```
just to limit the possibility of an overflow in `u32` architectures

synthesizer/process/src/stack/deploy.rs
```diff
@@ -73,11 +73,17 @@ impl<N: Network> Stack<N> {
...
+        let total_num_constraints = deployment.num_constraints();
+        ensure!(total_num_constraints <= N::MAX_DEPLOYMENT_CONSTRAINTS);
+
         // Construct the call stacks and assignments used to verify the certificates.
         let mut call_stacks = Vec::with_capacity(deployment.verifying_keys().len());
 
         // Iterate through the program functions and construct the callstacks and corresponding assignments.
-        for function in deployment.program().functions().values() {
+        for (function, (_, (verifying_key, _))) in
+            deployment.program().functions().values().zip_eq(deployment.verifying_keys())
```
ljedrz, line 96: should we `ensure!` that these iterators are of the same length before calling `zip_eq`?

synthesizer/src/vm/helpers/cost.rs
```diff
@@ -43,12 +45,17 @@ pub fn deployment_cost<N: Network>(deployment: &Deployment<N>) -> Result<(u64, (
...
         .saturating_mul(1_000_000); // 1 microcredit = 1e-6 credits.
 
+    // Compute the synthesis cost in credits
+    let synthesis_cost = num_constraints * N::SYNTH_FEE_MULTIPLIER;
+
     // Compute the total cost in microcredits.
     let total_cost = storage_cost
         .checked_add(namespace_cost)
+        .ok_or(anyhow!("The total cost computation overflowed for a deployment"))?
+        .checked_add(synthesis_cost)
```
ljedrz: `Option::and_then` could be used here to avoid duplicating the error message

synthesizer/src/vm/mod.rs
```diff
@@ -1077,6 +1077,40 @@ function check:
...
+    hash.sha3_256 r0 into r1 as field;
+    output r1 as field.public;",
+        )
+        .unwrap();
+
+        // Create the Deployment Transaction
+        let deployment = vm.deploy(&private_key, &program, None, 0, None, rng).unwrap();
+
+        // Verify the Deployment Transaction. It should fail because there are too many constraints.
+        assert!(vm.check_transaction(&deployment, None, rng).is_err());
```
ljedrz, line 1157: nit: it would be best if we checked for the specific error too, just to avoid any other error from making this test pass

ljedrz (commented):
Left a few comments and suggestions.

synthesizer/src/vm/helpers/cost.rs
```diff
@@ -43,12 +45,17 @@ pub fn deployment_cost<N: Network>(deployment: &Deployment<N>) -> Result<(u64, (
         .ok_or(anyhow!("The namespace cost computation overflowed for a deployment"))?
         .saturating_mul(1_000_000); // 1 microcredit = 1e-6 credits.
 
+    // Compute the synthesis cost in credits
```
vicsn: microcredits

circuit/environment/src/circuit.rs
vicsn: New constraints are only introduced after this check. I think either at the beginning or end are both possible, it will halt the whole synthesis anyway as soon as it is breached.

In the context of this PR, one constraint more or less wouldn't be the end of the world. But I do see indeed we are now effectively allowing `max_constraints + 1`, thank you for spotting. d872a4231

synthesizer/src/vm/mod.rs
vicsn: Do you happen to have a preferred approach? With e.g. a specific Rust error type? Taking into account we don't want to introduce too many changes in this PR, we can also leave an error refactor to another PR.

circuit/environment/src/circuit.rs
vicsn: 0be995040

circuit/environment/src/circuit.rs
vicsn: 06ab66756

ledger/block/src/transaction/deployment/mod.rs
vicsn: 6b6fc407a

synthesizer/process/src/stack/deploy.rs
vicsn: yes 726a614d7

synthesizer/src/vm/helpers/cost.rs
vicsn: 457334845

synthesizer/src/vm/helpers/cost.rs
vicsn: 457334845

synthesizer/src/vm/mod.rs
ljedrz: right now it would indeed require additional changes (i.e. informative error `enum`s), and I agree it's material for its own PR

circuit/environment/src/circuit.rs
ljedrz: ah, I thought `enforce` could potentially introduce more than a single constraint, but since it's not the case, it's nice to avoid an off-by-one

vicsn:
Confirmed that even when manipulating the verifying_key, the parent process fails gracefully: `action 'at16f9za6lpaz32l..' to the memory pool - Invalid unconfirmed transaction 'at16f9za6lpaz32l..': [tokio::spawn_blocking] task 1526 pan
icked.`

howardwu:
Is 1M too low?

synthesizer/process/src/stack/mod.rs
```diff
@@ -81,7 +81,7 @@ pub type Assignments<N> = Arc<RwLock<Vec<(circuit::Assignment<<N as Environment>
 pub enum CallStack<N: Network> {
     Authorize(Vec<Request<N>>, PrivateKey<N>, Authorization<N>),
     Synthesize(Vec<Request<N>>, PrivateKey<N>, Authorization<N>),
-    CheckDeployment(Vec<Request<N>>, PrivateKey<N>, Assignments<N>),
+    CheckDeployment(Vec<Request<N>>, PrivateKey<N>, Assignments<N>, u64),
```
raychu86: There doesn't seem to be final verification that the deployment actually has this specified number of constraints. 

We just know that it has less than this number, because verification didn't halt. That might(?) be enought?

raychu86:
Changing base to `mainnet`, because this may break the current testnet (if there are programs with >1m constraints). 

In addition, the deployment fee structure has changed to include a synthesis multiplier.

synthesizer/process/src/stack/mod.rs
vicsn: There is a check before synthesis, and then subsequently for every constraint that's added, which should be sufficient.

console/network/src/lib.rs
```diff
@@ -98,6 +98,10 @@ pub trait Network:
     const STARTING_SUPPLY: u64 = 1_500_000_000_000_000; // 1.5B credits
     /// The cost in microcredits per byte for the deployment transaction.
     const DEPLOYMENT_FEE_MULTIPLIER: u64 = 1_000; // 1 millicredit per byte
+    /// The cost in microcredits per constraint for the deployment transaction.
+    const SYNTH_FEE_MULTIPLIER: u64 = 25; // 25 microcredit per constraint to synthesize
+    /// The maximum number of constraints in a deployment
+    const MAX_DEPLOYMENT_CONSTRAINTS: u64 = 1_000_000;
```
iamalwaysuncomfortable: As the network gets better at handling large deployments, I would guess that there will be a strong push from the network governance to increase this number, and probably rather quickly. As might be the case with any limiting constants we set such as this.

If these do change in a governance decision, upgrades are going to become quite the difficulty among the validator set - thus I think it would be wise to make at the very least some kind of initial version communication methods between nodes (perhaps for the entire stack, but perhaps also for individual actions such as this) to go along with these changes, and some kind of light protocol for how to coordinate handle inter-node communication with nodes of different versions.

iamalwaysuncomfortable: Also, would this at all block the coinbase puzzle from being checked if it increased beyond 1M constraints?

iamalwaysuncomfortable (commented):
The question of "is 1M too few constraints" was good because I think it brings up the issue that everyone involved in governance will have the same question. When governance DOES decide that this needs to change, upgrades to this would need to be managed very carefully, so we should consider how to communicate protocol versions between the validator set (at the very least) and how to handle inter-node communication with those who might be running different versions with different constants.

console/network/src/lib.rs
vicsn: It is possible that this limit imposes a problem, so we should indeed run tests to ensure nothing goes wrong. We could make an exception for it.

vicsn:
@evan-schott can you help finish the two tests in 01b95d213 and look at fixing any broken CI tests? We can discuss further in person.

evan-schott:
> @evan-schott can you help finish the two tests in [01b95d2](https://github.com/AleoHQ/snarkVM/commit/01b95d213e97514364387f23981f2e9814ac7506) and look at fixing any broken CI tests? We can discuss further in person.

I will investigate

raychu86:
Is there a reason more CI tests are failing? Do we need to resample anything?

synthesizer/process/src/stack/deploy.rs
```diff
@@ -124,12 +124,20 @@ impl<N: Network> Stack<N> {
             lap!(timer, "Compute the request for {}", function.name());
             // Initialize the assignments.
             let assignments = Assignments::<N>::default();
+            // Initialize the constraint limit. Account for the constraint added after synthesis that randomizes vars.
```
vicsn: // Initialize the constraint limit. Account for the constraint added after synthesis that makes the Varuna zerocheck hiding.

synthesizer/process/src/stack/execute.rs
```diff
@@ -421,7 +421,15 @@ impl<N: Network> StackExecute<N> for Stack<N> {
             lap!(timer, "Save the transition");
         }
         // If the circuit is in `CheckDeployment` mode, then save the assignment.
-        else if let CallStack::CheckDeployment(_, _, ref assignments, _) = registers.call_stack() {
+        else if let CallStack::CheckDeployment(_, _, ref assignments, constraint_limit) = registers.call_stack() {
```
vicsn: I thought we agreed we would remove this check because it is already in `verify_vk`

vicsn: The overreport test should still fail when the certificate is verified.

vicsn:
LGTM!

synthesizer/process/src/stack/call/mod.rs
```diff
@@ -382,6 +382,11 @@ impl<N: Network> CallTrait<N> for Call<N> {
             // Inject the existing circuit.
             A::inject_r1cs(r1cs);
 
+            // If in 'CheckDeployment' mode, set the expected constraint limit.
+            if let CallStack::CheckDeployment(_, _, _, constraint_limit) = &registers.call_stack() {
+                A::set_constraint_limit(*constraint_limit);
+            }
+
```
howardwu, line 389: I'm not convinced this logic is necessary.

When triggering `call`, no logic above this resets the constraint limit AFAIK.

howardwu (approved):

PR #2272, opened 2024-01-02
[Refactor] Adds some tests and contains some documentation suggestions for the `fix/cast-lossy-field-to-group` PR

d0cd:
^

### DIFF
--- a/circuit/types/field/src/square_root.rs
+++ b/circuit/types/field/src/square_root.rs
@@ -64,22 +64,22 @@ impl<E: Environment> Field<E> {
 }
 
 impl<E: Environment> Field<E> {
-    /// Returns both square roots of `self` (hence the plural 'roots' in the name of the function),
-    /// along with a boolean error flag, which is set iff `self` is not a square.
+    /// Returns both square roots of `self` and a `Boolean` flag, which is set iff `self` is not a square.
     ///
-    /// In the console computation:
-    /// if `self` is a non-zero square,
-    /// the first field result is the positive root (i.e. closer to 0)
-    /// and the second field result is the negative root (i.e. closer to the prime);
-    /// if `self` is 0, both field results are 0;
-    /// if `self` is not a square, both field results are 0, but immaterial.
+    /// If `self` is a non-zero square,
+    ///  - the first field result is the positive root (i.e. closer to 0)
+    ///  - the second field result is the negative root (i.e. closer to the prime)
+    ///  - the flag is 0
     ///
-    /// The 'nondeterministic' part of the function name refers to the synthesized circuit,
-    /// whose represented computation, unlike the console computation just described,
-    /// returns the two roots (if `self` is a non-zero square) in no specified order.
-    /// This nondeterminism saves constraints, but generally this circuit should be only used
-    /// as part of larger circuits for which the nondeterminism in the order of the two roots does not matter,
-    /// and where the larger circuits represent deterministic computations despite this internal nondeterminism.
+    /// If `self` is 0,
+    ///  - both field results are 0
+    ///  - the flag is 0
+    ///
+    /// If `self` is not a square,
+    ///  - both field results are 0
+    ///  - the flag is 1
+    ///
+    /// Note that there is no ordering on the two roots returned by this function.
     pub fn square_roots_flagged_nondeterministic(&self) -> (Self, Self, Boolean<E>) {
         // Obtain (p-1)/2, as a constant field element.
         let modulus_minus_one_div_two = match E::BaseField::from_bigint(E::BaseField::modulus_minus_one_div_two()) {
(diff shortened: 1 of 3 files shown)


circuit/types/field/src/square_root.rs
```diff
@@ -64,22 +64,22 @@ impl<E: Environment> Field<E> {
...
-    /// and where the larger circuits represent deterministic computations despite this internal nondeterminism.
+    /// If `self` is 0,
+    ///  - both field results are 0
+    ///  - the flag is 0
+    ///
+    /// If `self` is not a square,
+    ///  - both field results are 0
+    ///  - the flag is 1
+    ///
+    /// Note that there is no ordering on the two roots returned by this function.
```
acoglio: Actually, the function itself always returns the roots in a specified order. It is the computation represented by the circuit that is nondeterministic, and allowed to return the roots in any order. This is why I think it's important to explain how the console computation differs from the circuit computation, since they normally do not differ in other circuits.

circuit/types/field/src/square_root.rs
```diff
@@ -92,30 +92,24 @@ impl<E: Environment> Field<E> {
...
             Ok(root) => root,
             Err(_) => console::Field::zero(),
         };
 
-        // In order to avoid actually calculating the square root in the circuit,
-        // we would like to generate a constraint saying that squaring the root yields self.
-        // But this constraint would have no solutions if self is not a square.
-        // So we introduce a new variable that is either self (if square) or 0 (otherwise):
-        // either way, this new variable is a square.
+        // Initialize the square element, which is either `self` or 0, depending on whether `self` is a square.
```
acoglio, line 102: I recommend to leave some explanation of why we calculate this square this way, otherwise it may look as coming a bit "out of the blue".

circuit/types/group/src/helpers/from_x_coordinate.rs
```diff
@@ -28,8 +28,8 @@ impl<E: Environment> Group<E> {
     }
 
     /// Initializes an affine group element from a given x-coordinate field element.
-    /// Also returns an error flag, set if there is no group element with the given x-coordinate;
-    /// in that case, the returned point is `(0, 0)`, but immaterial.
+    /// Returns an error flag, indicating if there is a group element with the given x-coordinate;
```
acoglio: I would read 'indicating' as 'when set', in which case it should say 'there is no group element'. I think that if we use the 'set' terminology explicitly it's more clear.

circuit/types/group/src/helpers/from_x_coordinate.rs
```diff
@@ -48,50 +48,31 @@ impl<E: Environment> Group<E> {
         let yy: Field<E> = witness!(|a_xx_minus_1, d_xx_minus_1| { a_xx_minus_1 / d_xx_minus_1 });
         E::enforce(|| (&yy, &d_xx_minus_1, &a_xx_minus_1));
 
-        // Compute both square roots of y^2, in no specified order, with a flag saying whether y^2 is a square or not.
-        // That is, finish solving the curve equation for y.
-        // If the x-coordinate line does not intersect the elliptic curve, this returns (1, 0, 0).
+        // Compute both square roots of y^2, with a flag indicating whether y^2 is a square or not.
+        // Note that there is **no** ordering on the square roots.
```
acoglio:
```suggestion
        // Note that there is **no** ordering on the square roots in the circuit computation.
```

circuit/types/group/src/helpers/from_x_coordinate.rs
```diff
@@ -48,50 +48,31 @@ impl<E: Environment> Group<E> {
...
+        // Note that if the x-coordinate line does not intersect the elliptic curve, this returns (0, 0, false).
```
acoglio:
```suggestion
        // Note that if the x-coordinate line does not intersect the elliptic curve, this returns (0, 0, true).
```

circuit/types/group/src/lib.rs
```diff
@@ -129,6 +129,18 @@ impl<E: Environment> Group<E> {
         // i.e. that it is 4 (= cofactor) times the postulated point on the curve.
         double_point.enforce_double(self);
     }
+
+    /// Returns a `Boolean` indicating if `self` is in the largest prime-order subgroup.
+    pub fn is_in_group(&self) -> Boolean<E> {
```
acoglio, line 135: I suggest adding here some of the comments about the code that was moved here, namely that we can perform scalar multiplication also on (1) points not in the subgroup (the point may not be in the subgroup) and (2) scalars not in the scalar type (the order does not have scalar type). It wouldn't necessarily be the case for this to be possible, if the group type and operations were defined more restrictively.

d0cd: Added some comments, please advise if additional ones are needed for clarity.

circuit/types/field/src/square_root.rs
```diff
@@ -64,22 +64,22 @@ impl<E: Environment> Field<E> {
...
-    /// and where the larger circuits represent deterministic computations despite this internal nondeterminism.
+    /// If `self` is 0,
+    ///  - both field results are 0
+    ///  - the flag is 0
+    ///
+    /// If `self` is not a square,
+    ///  - both field results are 0
+    ///  - the flag is 1
+    ///
+    /// Note that the constraints do **not** impose an ordering on the two roots returned by this function.
```
acoglio:
```suggestion
    /// Note that the constraints do **not** impose an ordering on the two roots returned by this function;
    /// this is what the `nondeterministic` part of this function name refers to.
```

circuit/types/group/src/lib.rs
```diff
@@ -129,6 +129,21 @@ impl<E: Environment> Group<E> {
         // i.e. that it is 4 (= cofactor) times the postulated point on the curve.
         double_point.enforce_double(self);
     }
+
+    /// Returns a `Boolean` indicating if `self` is in the largest prime-order subgroup.
```
acoglio:
```suggestion
    /// Returns a `Boolean` indicating if `self` is in the largest prime-order subgroup,
    /// assuming that `self` is on the curve.
```

acoglio (approved):
Just a couple of additional suggestions.

PR #2275, opened 2024-01-04
Feature-gate the test constraint system

ljedrz:
This PR feature-gates 3 modules in `snarkvm-algorithms`, making them build only for tests.


ljedrz:
By the way, I don't think that the `ConstraintVariable` (whose module is next to the ones affected by this PR) is used anywhere.

howardwu (approved):

PR #2276, opened 2024-01-04
[NCC-ADN] Prevent ignoring proving/verifying tasks #2

vicsn:
## Motivation

Cleanup based on NCC's finding, we should make sure that we don't silently ignore proving/verifying tasks.

## Test Plan

No new tests were added because the current codebase would not have triggered this issue anyway, it is defense in depth.

## Related PRs

Successor to: https://github.com/AleoHQ/snarkVM/pull/2032 which had a bunch of unrelated and unclear changes.

### DIFF
--- a/synthesizer/process/src/trace/mod.rs
+++ b/synthesizer/process/src/trace/mod.rs
@@ -314,9 +314,6 @@ impl<N: Network> Trace<N> {
             verifier_inputs.push((verifying_key, batch_inclusion_inputs));
         }
         // Verify the proof.
-        match VerifyingKey::verify_batch(locator, verifier_inputs, proof) {
-            Ok(()) => Ok(()),
-            Err(_) => bail!("Failed to verify proof"),
-        }
+        VerifyingKey::verify_batch(locator, verifier_inputs, proof).map_err(|e| anyhow!("Failed to verify proof - {e}"))
     }
 }
--- a/synthesizer/process/src/trace/mod.rs
+++ b/synthesizer/process/src/trace/mod.rs
@@ -315,8 +315,8 @@ impl<N: Network> Trace<N> {
         }
         // Verify the proof.
         match VerifyingKey::verify_batch(locator, verifier_inputs, proof) {
-            true => Ok(()),
-            false => bail!("Failed to verify proof"),
+            Ok(()) => Ok(()),
+            Err(_) => bail!("Failed to verify proof"),
         }
     }
 }
--- a/synthesizer/snark/src/proving_key/mod.rs
+++ b/synthesizer/snark/src/proving_key/mod.rs
@@ -70,7 +70,7 @@ impl<N: Network> ProvingKey<N> {
             .iter()
             .map(|(proving_key, assignments)| (proving_key.deref(), assignments.as_slice()))
             .collect();
-        ensure!(instances.len() == num_expected_instances, "Found duplicate proving keys");
+        ensure!(instances.len() == num_expected_instances, "Incorrect number of proving keys for batch proof");
 
         // Retrieve the proving parameters.
         let universal_prover = N::varuna_universal_prover();
--- a/synthesizer/snark/src/proving_key/mod.rs
+++ b/synthesizer/snark/src/proving_key/mod.rs
@@ -70,7 +70,7 @@ impl<N: Network> ProvingKey<N> {
             .iter()
             .map(|(proving_key, assignments)| (proving_key.deref(), assignments.as_slice()))
             .collect();
-        ensure!(instances.len() == num_expected_instances, "duplicate proving keys found");
+        ensure!(instances.len() == num_expected_instances, "Found duplicate proving keys");
 
         // Retrieve the proving parameters.
         let universal_prover = N::varuna_universal_prover();
--- a/synthesizer/snark/src/proving_key/mod.rs
+++ b/synthesizer/snark/src/proving_key/mod.rs
@@ -65,10 +65,12 @@ impl<N: Network> ProvingKey<N> {
         let timer = std::time::Instant::now();
 
         // Prepare the instances.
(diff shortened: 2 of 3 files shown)


synthesizer/process/src/trace/mod.rs
```diff
@@ -315,8 +315,8 @@ impl<N: Network> Trace<N> {
         }
         // Verify the proof.
         match VerifyingKey::verify_batch(locator, verifier_inputs, proof) {
-            true => Ok(()),
-            false => bail!("Failed to verify proof"),
+            Ok(()) => Ok(()),
+            Err(_) => bail!("Failed to verify proof"),
```
ljedrz: it might be preferable to not ignore the reason for failing here; knowing if the proof was invalid or the inputs were duplicate could aid debugging

ljedrz (approved):
LGTM with one suggestion.

synthesizer/process/src/trace/mod.rs
vicsn: Indeed, though the original code didn't propagate the error either, and there is [already a println! where it originates](https://github.com/AleoHQ/snarkVM/pull/2276/files#diff-4536d0528c90ad243676c73b67ffc8ae30cb325065825c168d5e48ab34ff2717R94). 

It may or may not be because we prefer to e.g. not return raw errors to clients. Let's see what @howardwu  prefers.

ljedrz: in general, it'd be best if we had nice error `enum`s everywhere, not just for informing the user, but also for tighter test cases etc; perhaps it's a candidate for its own PR if we consider this viable

synthesizer/snark/src/verifying_key/mod.rs
```diff
@@ -61,13 +61,19 @@ impl<N: Network> VerifyingKey<N> {
...
+        proof: &Proof<N>,
+    ) -> Result<()> {
         #[cfg(feature = "aleo-cli")]
         let timer = std::time::Instant::now();
 
         // Convert the instances.
+        let num_expected_keys = inputs.len();
         let keys_to_inputs: BTreeMap<_, _> =
             inputs.iter().map(|(verifying_key, inputs)| (verifying_key.deref(), inputs.as_slice())).collect();
+        ensure!(keys_to_inputs.len() == num_expected_keys, "invalid number of keys for batch proof");
```
howardwu:
```suggestion
        ensure!(keys_to_inputs.len() == num_expected_keys, "Incorrect number of verifying keys for batch proof");
```

synthesizer/snark/src/proving_key/mod.rs
```diff
@@ -65,10 +65,12 @@ impl<N: Network> ProvingKey<N> {
         let timer = std::time::Instant::now();
 
         // Prepare the instances.
+        let num_expected_instances = assignments.len();
         let instances: BTreeMap<_, _> = assignments
             .iter()
             .map(|(proving_key, assignments)| (proving_key.deref(), assignments.as_slice()))
             .collect();
+        ensure!(instances.len() == num_expected_instances, "duplicate proving keys found");
```
howardwu:
```suggestion
        ensure!(instances.len() == num_expected_instances, "Found duplicate proving keys");
```

synthesizer/process/src/trace/mod.rs
howardwu: that [println](https://github.com/AleoHQ/snarkVM/pull/2276/files#diff-4536d0528c90ad243676c73b67ffc8ae30cb325065825c168d5e48ab34ff2717R94) is `cfg` guarded for CLI outputs only.

howardwu: see above examples - errors are always passed through
https://github.com/AleoHQ/snarkVM/pull/2276/files#diff-a35c55aacc5e19742943a6a3e97e6a7faabea519fe356ac30341b864627b7b71R252

synthesizer/snark/src/verifying_key/mod.rs
```diff
@@ -77,13 +83,16 @@ impl<N: Network> VerifyingKey<N> {
         match Varuna::<N>::verify_batch(universal_verifier, fiat_shamir, &keys_to_inputs, proof) {
             Ok(is_valid) => {
                 #[cfg(feature = "aleo-cli")]
-                println!("{}", format!(" • Verified '{locator}' (in {} ms)", timer.elapsed().as_millis()).dimmed());
-                is_valid
+                println!(
+                    "{}",
+                    format!(" • Verified '{locator}': {is_valid} (in {} ms)", timer.elapsed().as_millis()).dimmed()
+                );
+                if is_valid { Ok(()) } else { bail!("batch proof is invalid") }
```
howardwu:
```suggestion
                if is_valid { Ok(()) } else { bail!("'verify_batch' failed") }
```

synthesizer/process/src/trace/mod.rs
howardwu: resolved in c1d0898b5

synthesizer/snark/src/proving_key/mod.rs
```diff
@@ -65,10 +65,12 @@ impl<N: Network> ProvingKey<N> {
         let timer = std::time::Instant::now();
 
         // Prepare the instances.
+        let num_expected_instances = assignments.len();
         let instances: BTreeMap<_, _> = assignments
             .iter()
             .map(|(proving_key, assignments)| (proving_key.deref(), assignments.as_slice()))
             .collect();
+        ensure!(instances.len() == num_expected_instances, "Found duplicate proving keys");
```
howardwu:
```suggestion
        ensure!(instances.len() == num_expected_instances, "Incorrect number of proving keys for batch proof");
```

howardwu (approved):

PR #2277, opened 2024-01-04
perf: optimize console's group from_x_coordinate

bishopcheckmate:
Since the 'greatest' parameter is just a matter of trying 'y' and '-y' at the end and doesn't affect
the fallabality of computation, we can be sure that if it has failed to compute for 'y', it will also
fail for '-y', thus we can fail the group computation early.

## Motivation

I was learning the codebase and came across this idea. I'm not 100% sure if I'm correct tho, as this is a function from the trait, if there can be supported curves which do not hold to the simple y negation.

## Test Plan

regular tests run

## Related PRs

n/a


ljedrz (commented):
I agree with the general reasoning and about the caveat (the trait being public).

Another approach I've tinkered with would be to allow the caller to reuse the partial result from the calculation of the first point in order to reduce the number of operations if the second one is needed. You may want to try it out as an alternative that's safe even if there are other future users of the trait; otherwise I'll try to find and revive that old branch soon :broom:.

bishopcheckmate:
I was also thinking about that but more about another method in trait which just computes both and return a tuple, however to be honest what really stopped me is the naming for those. I don't think partial result works as the necessary steps to finish the computation may vary

ljedrz:
> I don't think partial result works as the necessary steps to finish the computation may vary

IIRC in my attempt I had an optional partial result that was returned alongside the point in a tuple; if there wasn't any meaningful partial result, that value would just be `None`.

bishopcheckmate:
so this is the alternative approach I've been thinking about. It should also optimize the case when the lexicographically greatest affine point is not in a correct subgroup and the smallest has to be checked

ljedrz (commented):
I'd like to review this, but I'll have to yield to someone with more cryptographic savvy. I do like the small diff, though, and believe that a perf improvement here could be meaningful, as I recognize `from_x_coordinate` from perf-related profiling.

vicsn:
Thank you for the contribution @bishopcheckmate ! Could you run some of the benchmarks (`cd <subdirectory> && cargo bench`) in the codebase and share how big the improvements are?

ljedrz:
@vicsn which benchmarks would you expect to be potentially affected?

bishopcheckmate:
So I've benched stuff in `console`, `circuit` and `ledger` but the results looked just like a regular noise. I've tried to read the benchmarks too and to be honest I couldn't see any that seemed to be relevant to this change.

I've written a simple bench for creating a group from field:

```rs
fn group_from_field(c: &mut Criterion) {

    c.bench_function("group_from_field", move |b| {
        b.iter(|| {
            let field: Field<CurrentNetwork> = rng.gen();
            let group = Group::from_field(&field);
            std::hint::black_box(group)
        })
    });
}
```

and this is the result:

```
group_from_field        time:   [75.370 µs 75.700 µs 76.106 µs]
                        change: [-13.170% -12.189% -11.205%] (p = 0.00 < 0.05)
                        Performance has improved.
Found 2 outliers among 20 measurements (10.00%)
  1 (5.00%) low mild
  1 (5.00%) high mild
```

Honestly I expected bigger gains there, but the bench isn't pure enough, there is a bias coming from generating bigints too, so likely the change is higher than reported

bishopcheckmate:
I've rewritten the bench to not include the sampling:

```rs
fn group_from_field(c: &mut Criterion) {

    c.bench_function("group_from_field", move |b| {
        let fields: Vec<Field<CurrentNetwork>> = (0..1000).map(|_| rng.gen()).collect();

        b.iter(|| {
            for field in &fields {
                let group = Group::from_field(field);
                let _ = std::hint::black_box(group);
            }
        })
    });
}
```

```
group_from_field        time:   [74.006 ms 74.841 ms 75.590 ms]
                        change: [-37.548% -30.986% -24.575%] (p = 0.00 < 0.05)
                        Performance has improved.
Found 1 outliers among 20 measurements (5.00%)
  1 (5.00%) low mild
```

bishopcheckmate:
do you want me to include the benchmark in this change?

bishopcheckmate:
I pushed the benchmarks. I've added two more, one for the case when the field is in correct subgroup and the second for when it's not. Here are the current results which seem pretty consistent. Previous -30% was likely an outlier, which can even be seen by the differences between min and max gain.

```
group_from_field        time:   [74.926 ms 75.784 ms 76.655 ms]
                        change: [-16.942% -15.758% -14.586%] (p = 0.00 < 0.05)
                        Performance has improved.
Found 2 outliers among 20 measurements (10.00%)
  1 (5.00%) low mild
  1 (5.00%) high mild


Initializing 'TestRng' with seed '7092101494189885881'

group_from_field_on_curve
                        time:   [124.83 ms 125.29 ms 125.78 ms]
                        change: [-7.1391% -6.5194% -5.9042%] (p = 0.00 < 0.05)
                        Performance has improved.


Initializing 'TestRng' with seed '2138409011234971088'

group_from_field_off_curve
                        time:   [59.320 ms 59.926 ms 60.568 ms]
                        change: [-17.393% -15.898% -14.401%] (p = 0.00 < 0.05)
                        Performance has improved.
Found 1 outliers among 20 measurements (5.00%)
  1 (5.00%) high mild
```

vicsn:
Thank you for the additions @bishopcheckmate ! This month is a very busy one with our coming mainnet launch, so I can't guarantee when we can finish the review. But really appreciate all the extra effort you put in the benchmarks, I genuinely look forward to diving into the proposed changes.

bishopcheckmate:
Another modification for benches, now it tests for the case when the `Affine::from_x_coordinate` is `None`, so now the last benchmark tests the fast-fail case.

```
group_from_field_off_curve
                        time:   [10.424 ms 10.430 ms 10.440 ms]
                        change: [-43.461% -43.378% -43.303%] (p = 0.00 < 0.05)
                        Performance has improved.

```

bishopcheckmate:
Sure, thanks @vicsn.

> I'd like to review this, but I'll have to yield to someone with more cryptographic savvy. 

Now with the addition of [tests](https://github.com/AleoHQ/snarkVM/pull/2277/commits/603a8f721d41dc505d7688c5ba65c0b0a642c6bc) which check correctness of the tiny math bit behind maybe there is no need for a big cryptographic dive

.github/workflows/benchmarks.yml
```diff
@@ -62,6 +62,12 @@ jobs:
           cargo bench --bench merkle_tree -- --output-format bencher | tee -a ../../output.txt
           cd ../..
           
+      - name: Benchmark console/types
+        run: |
+          cd console/types
+          cargo bench --bench merkle_tree -- --output-format bencher | tee -a ../../output.txt
```
ljedrz: seems like this one needs to be adjusted

bishopcheckmate: good catch

bishopcheckmate:
to make the dive into change simpler you can also check the [diff from first two commits](https://github.com/AleoHQ/snarkVM/pull/2277/files/29d25885556c622593a94f0d84177f45ae876397) as they contain the whole logic. The rest are tests and benches

bishopcheckmate:
Hey, kind bump. Any chance of getting this reviewed?

The math optimization behind this is really simple. Instead of computing affines one by one like:
```rust
let p1 = Affine::from_x_coordinate(x, greatest);
let p2 = Affine::from_x_coordinate(x, smallest);
```
We use the first point to compute the other with much less operations:
```rust
let p1 = Affine::from_x_coordinate(x, greatest);
let p2 = Affine::new(p1.x, -p1.y, -p1.t); // or (p1.x, -p1.y, false) for weierstrass
```

bishopcheckmate:
the CI failure doesn't seem related, this passe for me locally

ljedrz:
Judging by the results of my latest profiling run, this change has the potential to boost block sync speed.

bishopcheckmate:
sounds good. @vicsn any chance for getting your expertise on that? :pray:

ljedrz:
I benchmarked this change with `hyperfine` using a small binary to deserialize 50 blocks, and it yielded the following improvements:
- avg. runtime -2.55%
- std.dev -39.77% (a decrease from an already very small value, but still desirable)

howardwu (approved):

PR #2279, opened 2024-01-07
perf: optimize console field and scalar from_bits_le

bishopcheckmate:
## Motivation

Another idea for optimization. Instead `.iter().take().cloned().collect()` followed by resize, we can just allocate whole vec at once and copy data to it, as we are sure that it's length is less than `size_in_bits`.
This assumption is also utilized in the previous `if` branch.

I've made some messy benchmarking and it reports around 20% gain with this approach. I may polish and include them later

## Test Plan

regular test run

### DIFF
--- a/console/types/field/src/from_bits.rs
+++ b/console/types/field/src/from_bits.rs
@@ -47,7 +47,7 @@ impl<E: Environment> FromBits for Field<E> {
         } else {
             // Construct the sanitized list of bits padded with `false`
             let mut sanitized_bits = vec![false; size_in_bits];
-            // this is safe, because we just checked that the length of bits isn't bigger
+            // Note: This is safe, because we just checked that the length of bits isn't bigger
             // than `size_in_data_bits` which is equal to `size_in_bits - 1`.
             sanitized_bits[..num_bits].copy_from_slice(bits_le);
 
--- a/console/types/field/src/from_bits.rs
+++ b/console/types/field/src/from_bits.rs
@@ -45,12 +45,14 @@ impl<E: Environment> FromBits for Field<E> {
             // Return the field.
             Ok(Field { field: E::Field::from_bigint(field).ok_or_else(|| anyhow!("Invalid field from bits"))? })
         } else {
-            // Construct the sanitized list of bits, resizing up if necessary.
-            let mut bits_le = bits_le.iter().take(size_in_bits).cloned().collect::<Vec<_>>();
-            bits_le.resize(size_in_bits, false);
+            // Construct the sanitized list of bits padded with `false`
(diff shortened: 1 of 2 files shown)


console/types/field/src/from_bits.rs
```diff
@@ -46,11 +46,13 @@ impl<E: Environment> FromBits for Field<E> {
             Ok(Field { field: E::Field::from_bigint(field).ok_or_else(|| anyhow!("Invalid field from bits"))? })
         } else {
             // Construct the sanitized list of bits, resizing up if necessary.
```
ljedrz: since we're no longer resizing, perhaps this could now say that the list is padded with `false` instead? this applies to the other file too

ljedrz (approved):
LGTM :+1:

bishopcheckmate:
if you don't want me to include the benches then this is ready by me

console/types/scalar/src/from_bits.rs
```diff
@@ -48,12 +48,14 @@ impl<E: Environment> FromBits for Scalar<E> {
             // Return the scalar.
             Ok(Scalar { scalar: E::Scalar::from_bigint(scalar).ok_or_else(|| anyhow!("Invalid scalar from bits"))? })
         } else {
-            // Construct the sanitized list of bits, resizing up if necessary.
-            let mut bits_le = bits_le.iter().take(size_in_bits).cloned().collect::<Vec<_>>();
-            bits_le.resize(size_in_bits, false);
+            // Construct the sanitized list of bits padded with `false`
+            let mut sanitized_bits = vec![false; size_in_bits];
+            // this is safe, because we just checked that the length of bits isn't bigger
```
howardwu:
```suggestion
            // Note: This is safe, because we just checked that the length of bits isn't bigger
```

console/types/field/src/from_bits.rs
```diff
@@ -45,12 +45,14 @@ impl<E: Environment> FromBits for Field<E> {
             // Return the field.
             Ok(Field { field: E::Field::from_bigint(field).ok_or_else(|| anyhow!("Invalid field from bits"))? })
         } else {
-            // Construct the sanitized list of bits, resizing up if necessary.
-            let mut bits_le = bits_le.iter().take(size_in_bits).cloned().collect::<Vec<_>>();
-            bits_le.resize(size_in_bits, false);
+            // Construct the sanitized list of bits padded with `false`
+            let mut sanitized_bits = vec![false; size_in_bits];
+            // this is safe, because we just checked that the length of bits isn't bigger
```
howardwu:
```suggestion
            // Note: This is safe, because we just checked that the length of bits isn't bigger
```

howardwu (approved):

PR #2280, opened 2024-01-08
Make TransactionId Ord

vicsn:
## Motivation

Required to order transactions in our mempool queue in https://github.com/AleoHQ/snarkOS/pull/2977

### DIFF
--- a/console/network/src/helpers/id.rs
+++ b/console/network/src/helpers/id.rs
@@ -39,6 +39,7 @@ pub trait Bech32ID<F: FieldTrait>:
     + core::hash::Hash
     + Sync
     + Send
+    + Ord
 {
     fn prefix() -> String;
     fn size_in_bytes() -> usize;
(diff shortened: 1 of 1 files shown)


console/network/src/helpers/id.rs
```diff
@@ -210,3 +211,17 @@ impl<F: FieldTrait, const PREFIX: u16> Distribution<AleoID<F, PREFIX>> for Stand
...
+    #[inline]
+    fn cmp(&self, other: &Self) -> Ordering {
+        self.0.cmp(&other.0)
+    }
+}
+
+impl<F: FieldTrait, const PREFIX: u16> PartialOrd for AleoID<F, PREFIX> {
+    #[inline]
+    fn partial_cmp(&self, other: &Self) -> Option<Ordering> {
+        Some(self.0.cmp(&other.0))
```
ljedrz: I'm wondering if it wouldn't be more clear to add a `PartialOrd` bound to the `Bech32ID` trait and reuse its impl here instead (i.e. `self.0.partial_cmp(&other.0)`)

vicsn: Cool, is that some kind of default impl? 2454c874a

ljedrz: yep, looks good :ok_hand:

console/network/src/helpers/id.rs
```diff
@@ -210,3 +212,17 @@ impl<F: FieldTrait, const PREFIX: u16> Distribution<AleoID<F, PREFIX>> for Stand
         AleoID::<F, PREFIX>(Uniform::rand(rng))
     }
 }
+
+impl<F: FieldTrait, const PREFIX: u16> Ord for AleoID<F, PREFIX> {
+    #[inline]
+    fn cmp(&self, other: &Self) -> Ordering {
+        self.0.cmp(&other.0)
```
howardwu, line 219: Note: Ord wasn't implemented because it is unnatural to order `bech32m` **strings** by their `4829384field` **field** value.

howardwu: And I was concerned implementing a string comparator would be too expensive.

vicsn: I see, that makes the third time I made an unnatural Ord proposal. Closing this PR.

In the related snarkOS PR, I'm going ahead and order by the inner field anyway, just to distinguish transactions which come in at the same nanosecond. Which is also semantically weird, but it does the job.

PR #2294, opened 2024-01-10
Add fixed execution cost

vicsn:
## Motivation

Adds a fixed execution cost.

## Tests
Need to rewrite expectations, which can be done once we're sure when this will be merged.

## Related PRs

- Updates deployment costs: https://github.com/AleoHQ/snarkVM/pull/2271
- Updates finalize costs: https://github.com/AleoHQ/snarkVM/pull/2281
- Updates snarkOS to https://github.com/AleoHQ/snarkVM/pull/2271


synthesizer/src/vm/helpers/cost.rs
```diff
@@ -55,7 +55,7 @@ pub fn deployment_cost<N: Network>(deployment: &Deployment<N>) -> Result<(u64, (
 pub fn execution_cost<N: Network, C: ConsensusStorage<N>>(
     vm: &VM<N, C>,
     execution: &Execution<N>,
-) -> Result<(u64, (u64, u64))> {
+) -> Result<(u64, (u64, u64, u64))> {
```
ljedrz: I wonder if we shouldn't introduce some new `struct` to minimize the chance of mixing these values up?

ljedrz (commented):
Left one suggestion.

synthesizer/src/vm/helpers/cost.rs
vicsn: Or we don't return anything anymore: 6d0d4c986
(When creating a type for it the compiler complained it was never read)

iamalwaysuncomfortable: From a UI perspective it's actually prudent to return the cost breakdown so that people can plan their cost. The SDK uses all the values of this function for instances. It'd be wise to have a struct from the perspective of communication.

synthesizer/src/vm/helpers/cost.rs
```diff
@@ -52,10 +52,7 @@ pub fn deployment_cost<N: Network>(deployment: &Deployment<N>) -> Result<(u64, (
 }
 
 /// Returns the *minimum* cost in microcredits to publish the given execution (total cost, (storage cost, namespace cost)).
-pub fn execution_cost<N: Network, C: ConsensusStorage<N>>(
-    vm: &VM<N, C>,
-    execution: &Execution<N>,
-) -> Result<(u64, (u64, u64))> {
+pub fn execution_cost<N: Network, C: ConsensusStorage<N>>(vm: &VM<N, C>, execution: &Execution<N>) -> Result<u64> {
```
iamalwaysuncomfortable: I'd argue against condensing this into one cost. From a usability perspective, users should be able to understand the cost structure & plan their costs. 

From a perspective of usability I STRONGLY recommend returning to the original suggestion of having a cost struct to allow users to plan.

iamalwaysuncomfortable (changes requested):

synthesizer/src/vm/helpers/cost.rs
vicsn: Ah I was not aware, thx will change it back and will add a comment about its relevance.

synthesizer/src/vm/helpers/cost.rs
vicsn: Sounds good! 7b3c9bc6f

synthesizer/src/vm/helpers/cost.rs
```diff
@@ -23,6 +23,14 @@ use synthesizer_program::{Command, Finalize, Instruction};
...
 use std::collections::HashMap;
 
+#[allow(unused)]
+/// Helper struct for i.a. SDK to return costs for users.
+pub struct ExecutionCosts {
+    storage_cost: u64,
+    finalize_cost: u64,
+    execution_cost: u64,
+}
+
```
iamalwaysuncomfortable, line 33: Looks good, just needs accessor methods for end users, and perhaps a display method for printing it.

```suggestion
impl ExecutionCosts {
    /// Cost for storing the execution transaction on the Aleo Network
    pub fn storage_cost() -> u64 {
        self.storage_cost
    }
    
     /// Cost for executing all operations within the function's finalize scope on the Aleo Network
    pub fn finalize_cost() -> u64 {
        self.finalize_cost
    }
    
     /// Cost for verification of the function execution
    pub fn execution_cost() -> u64 {
        self.storage_cost
    }
}
```

iamalwaysuncomfortable (commented):
One further nit on providing accessor methods for end users

synthesizer/src/vm/helpers/cost.rs
vicsn: I feel like a rebel today: cc10cce5e

howardwu:
Closing as this is not necessary at this time.

PR #2296, opened 2024-01-10
[HackerOne-2239704] Introduce an override for atomic batch operations

ljedrz:
This PR introduces a database-level override flag that does not allow an atomic operation to be executed until it's flipped manually.

This is a relatively simple solution to make block insertion and finalization happen as a single atomic storage operation; the reason why they can't be carried out like that "normally" is that they are disjoint in the storage hierarchy.

This should fix the storage corruption that can currently happen if the node is shut down between those two operations.

Filing it as a draft, as I'd like to further improve the related docs, but it worked when I tested it locally.

This PR should address:
- https://github.com/AleoHQ/snarkVM/issues/2218
- https://github.com/AleoHQ/snarkOS/issues/2974


ljedrz:
I'm also going to check the failing test cases; it's likely that some of them are just using the low-level operations individually.

Edit: it was actually just due to the in-memory storage making assertions that only apply to the persistent one; fixed.

ledger/store/src/helpers/rocksdb/internal/mod.rs
```diff
@@ -189,6 +193,26 @@ impl Database for RocksDB {
...
+    /// operations get executed.
+    fn flip_atomic_override(&self) -> Result<bool> {
+        // https://github.com/rust-lang/rust/issues/98485
+        let previous_value = self.atomic_override.load(Ordering::SeqCst);
+
+        // A flip from enabled to disabled executes all pending operations
+        // as a single atomic batch.
+        if previous_value {
+            let batch = mem::take(&mut *self.atomic_batch.lock());
+            self.rocksdb.write(batch)?;
```
raychu86, line 245: This will write to DB even if the `previous_atomic_depth` is not 1. Is this safe?

Normally in `finish_atomic` we only write if we're at depth 0  `if previous_atomic_depth == 1`. But now we can get around that by abusing `flip_atomic_override`. In theory if we use the override flag correctly there will be no issues, but it is a bit concerning that this is a possibility.

ljedrz: I think it's reasonable to add an atomic depth safeguard here :+1:; it will restrict the applicability of the override, but we only want it used in one spot where this is a given.

ljedrz:
Rebased and addressed the review comment in a new commit.

ledger/store/src/block/mod.rs
```diff
@@ -392,6 +392,11 @@ pub trait BlockStorage<N: Network>: 'static + Clone + Send + Sync {
         self.transaction_store().finish_atomic()
     }
 
+    /// Either enables or disables the atomic override.
+    fn flip_atomic_override(&self) -> Result<bool> {
+        self.state_root_map().flip_atomic_override()
```
howardwu: Why are you only flipping one map?

ledger/store/src/block/mod.rs
```diff
@@ -1147,6 +1152,11 @@ impl<N: Network, B: BlockStorage<N>> BlockStore<N, B> {
     pub fn dev(&self) -> Option<u16> {
         self.storage.dev()
     }
+
+    /// Either enables or disables the atomic override.
+    pub fn flip_atomic_override(&self) -> Result<bool> {
+        self.storage.flip_atomic_override()
```
howardwu: Why are you only flipping one map?

ledger/store/src/helpers/rocksdb/internal/mod.rs
```diff
@@ -189,6 +193,30 @@ impl Database for RocksDB {
...
+        assert_eq!(self.atomic_depth.load(Ordering::SeqCst), 0);
+
+        // https://github.com/rust-lang/rust/issues/98485
+        let previous_value = self.atomic_override.load(Ordering::SeqCst);
+
+        // A flip from enabled to disabled executes all pending operations
+        // as a single atomic batch.
+        if previous_value {
+            let batch = mem::take(&mut *self.atomic_batch.lock());
+            self.rocksdb.write(batch)?;
```
howardwu, line 245: Why do we need to write all pending operations when flipping? Please provide commentary in the code for this.

ledger/store/src/block/mod.rs
ljedrz: the map is just a means to access the storage; since the override is only useful and effective with persistent storage, it is implemented at the lowest level that all the maps share and any of them can be used to toggle it

ledger/store/src/block/mod.rs
ljedrz: same as https://github.com/AleoHQ/snarkVM/pull/2296#discussion_r1460329080

ljedrz:
Rebased and addressed the latest review comments in a new commit.

I also noticed that the unhappy path for finalization needed to be adjusted, which was done in the last commit.

ledger/store/src/helpers/rocksdb/internal/mod.rs
ljedrz: I added some extra comments; there could be an additional operation responsible for performing them, but that would require a new method that would most likely be called at basically the same spot. Since the purpose of this solution is to merge 2 atomic batches into 1, it makes sense to do it automatically at the end. The overall order of operations is as follows
1. enable (flip from disabled to enabled) the atomic override, making the subsequent atomic batches only collect operations, but not execute them
2. execute block insertion and finalization
3. disable (flip from enabled to disabled) the atomic override, executing all the collected operations since it has been enabled (or discarding them in case of failure)

ledger/store/src/block/mod.rs
howardwu: I don't get it.

What is it about `state_root` and not anything else that impacts this?

`state_root` is not even touched by finalize scope.

ljedrz: The state root map (the choice of which is insignificant, it's just the first map listed in `BlockStorage`), like all the other instances of `DataMap`, contains a reference to a single, shared instance of `RocksDB`, so any of them (and either from `BlockStore` or `FinalizeStore`, as they are the ones available when inserting blocks) can be used to toggle the override, which is implemented within the `RocksDB` object. Once the override is toggled (from any `DataMap`), all the other maps are aware of it being in force, and when they call `finish_atomic`, they only extend the serialized atomic batch, but it doesn't get executed (which would normally happen with the final/outermost call) until the override is flipped again.

Note that `MemoryMap` - which doesn't have a single underlying instance of storage - is irrelevant here (as only persistent storage needs atomic operations), and only has a noop impl of the override.

synthesizer/src/vm/mod.rs
```diff
@@ -335,13 +335,34 @@ impl<N: Network, C: ConsensusStorage<N>> VM<N, C> {
...
+                // latest block.
+                #[cfg(feature = "rocks")]
+                self.block_store().atomic_abort();
+                #[cfg(feature = "rocks")]
+                self.finalize_store().atomic_abort();
+                // Disable the atomic batch override.
+                #[cfg(feature = "rocks")]
+                assert!(!self.block_store().flip_atomic_override()?);
+                // Rollback the block (only applies to in-memory storage).
+                #[cfg(not(feature = "rocks"))]
```
raychu86: This is not sufficient.

`self.block_store().insert(block)?;` updates the in-memory merkle tree. However since we are not calling `self.block_store().remove_last_n(1)` anymore in the rocksdb case, then the tree will not be correct for the given db state.

ljedrz: good catch :+1:, I reverted this bit; the `remove_last_n` code will still attempt to remove the non-existent blocks in case of persistent storage, but that shouldn't be noticeable

synthesizer/src/vm/mod.rs
```diff
@@ -335,13 +335,34 @@ impl<N: Network, C: ConsensusStorage<N>> VM<N, C> {
         // Attention: The following order is crucial because if 'finalize' fails, we can rollback the block.
         // If one first calls 'finalize', then calls 'insert(block)' and it fails, there is no way to rollback 'finalize'.
```
howardwu: This comment can be removed.

synthesizer/src/vm/mod.rs
```diff
@@ -335,12 +335,32 @@ impl<N: Network, C: ConsensusStorage<N>> VM<N, C> {
...
+                #[cfg(feature = "rocks")]
+                assert!(!self.block_store().flip_atomic_override()?);
+                Ok(())
+            }
             Err(finalize_error) => {
+                // Clear all the pending atomic operations so that disabling the atomic batch override
+                // doesn't execute any storage operation that has accumulated since it was enabled.
+                // This only applies to persistent storage - the in-memory one needs to roll back the
+                // latest block.
+                #[cfg(feature = "rocks")]
```
howardwu: nit: for readability, can you switch this syntax to:
```rs
#[cfg!(feature = "rocks)]
{
  // Do new stuff
}
```
or
```rs
if cfg!(feature = "rocks) {
  // Do new stuff
}
```

synthesizer/src/vm/mod.rs
```diff
@@ -335,12 +335,32 @@ impl<N: Network, C: ConsensusStorage<N>> VM<N, C> {
         // Attention: The following order is crucial because if 'finalize' fails, we can rollback the block.
         // If one first calls 'finalize', then calls 'insert(block)' and it fails, there is no way to rollback 'finalize'.
 
+        // Enable the atomic batch override, so that both the insertion and finalization belong to a single batch.
+        #[cfg(feature = "rocks")]
+        assert!(self.block_store().flip_atomic_override()?);
```
howardwu: Should this be `debug_assert!`?

ledger/store/src/helpers/rocksdb/internal/nested_map.rs
```diff
@@ -198,8 +198,11 @@ impl<
 
         // Ensure that the atomic batch is empty.
         assert!(self.atomic_batch.lock().is_empty());
-        // Ensure that the database atomic batch is empty.
-        assert!(self.database.atomic_batch.lock().is_empty());
+        // Ensure that the database atomic batch is empty; skip this check if the atomic
+        // override is enabled, as there may be pending storage operations.
+        if !self.database.atomic_override.load(Ordering::SeqCst) {
```
howardwu: We should introduce an abstraction for this call everywhere, so that it is used properly:
```rs
self.database.atomic_override.load(Ordering::SeqCst)
```

synthesizer/src/vm/mod.rs
howardwu: resolved in e2e207d37

synthesizer/src/vm/mod.rs
howardwu: resolved in e2e207d37

synthesizer/src/vm/mod.rs
```diff
@@ -332,15 +332,30 @@ impl<N: Network, C: ConsensusStorage<N>> VM<N, C> {
...
+                {
+                    // Clear all pending atomic operations so that disabling the atomic batch override
+                    // doesn't execute any storage operation that accumulated after it was enabled.
+                    self.block_store().atomic_abort();
+                    self.finalize_store().atomic_abort();
+                    // Disable the atomic batch override.
+                    assert!(!self.block_store().flip_atomic_override()?);
+                }
                 // Rollback the block.
                 self.block_store().remove_last_n(1).map_err(|removal_error| {
```
howardwu: This is still incorrect.

Calling `remove_last_n(1)` here will 1) correctly update the Merkle tree, however 2) it will also delete the penultimate block now from `self.block_store()`.

This is erroneous because we are using this concept of a `flip_atomic_override` to remove the pending last block, so the call to `remove_last_n(1)` will now remove the penultimate block erroneously.

synthesizer/src/vm/mod.rs
ljedrz: the plain `assert` was intentional, as this ensures that it is initially disabled - if this was not the case, it would indicate a logic bug, and other atomic operations could be affected

synthesizer/src/vm/mod.rs
ljedrz: Indeed; there are 2 ways I can see of tackling this; we can either:
- put the storage deletion within `remove_last_n` behind a feature flag (so that it doesn't happen with `rocks`); however, this would obscure the purpose of the method and affect its use outside of typical runtime (and probably break some tests)
- break the `remove_last_n` method into 2 distinct operations (isolated into new methods) - the tree update and the storage update - and only call the tree update here; I'll see if I can propose something "clean" for this

ljedrz: I proposed the 2nd approach in https://github.com/AleoHQ/snarkVM/pull/2296/commits/1a27305b7e0acb2c23874bdf1d11ce2b527746e1. The scope of the write lock over the Merkle tree is different, but since the entirety of block insertion is behind the `block_lock`, this shouldn't affect anything (I'm assuming that only block insertion/rollback can alter it).

ledger/store/src/helpers/rocksdb/internal/nested_map.rs
ljedrz: https://github.com/AleoHQ/snarkVM/pull/2296/commits/14a113609d221751f2081afc8465eeca944d6d31

ljedrz:
~Switching to draft so I can double-check the updated setup.~

Done, and tested locally.

raychu86 (approved):
LGTM

howardwu (approved):

PR #2297, opened 2024-01-10
perf: remove unnecessary allocation in check_next_block

bishopcheckmate:
## Motivation

It seems that rust wasn't able to resolve the correct `.zip(...)` because rayon has it under `IndexedParallelIterator` which luckily `.par_values(...)` implements. This removes the unnecessary reallocation of transactions.

## Test Plan

regular tests run


bishopcheckmate:
@ljedrz might be of interest to you

ljedrz (approved):
Cool, I totally missed this :+1:.

howardwu (approved):

PR #2303, opened 2024-01-15
fix varuna selector annotation

cx19981114:


algorithms/src/snark/varuna/ahp/selectors.rs
```diff
@@ -85,7 +85,7 @@ pub(crate) fn apply_randomized_selector<F: PrimeField>(
     // This removes a mul and div by v_H operation over each circuit's (target_domain - src_domain)
```
howardwu, line 85: @vicsn can you confirm the comment here remains correct?

howardwu (approved):

algorithms/src/snark/varuna/ahp/selectors.rs
vicsn: Confirmed

PR #2304, opened 2024-01-15
Store a bitset of transmissions in compact subdag.

vicsn:
## Motivation

Subdags can include duplicate transmission_ids which scales very badly at the moment (30MB w/ 50 validators). In this PRs, Subdags can track transmission_ids using a succinct bitset.

New types:
- CompactCertificate
- CompactHeader
- Subdag enum can hold either
- NarwhalCertificate trait provides an interface
- We use BitSet, which used to be part of standard library

Some considerations on the design:
- transmissions in BatchCertificates are now ordered lexicographically so we can reconstruct them. 
- It is useful for the Subdag to be an enum, so we know *all* of its certificates are *either* compact or full.
- I spent a while thinking about using a NarwhalCertificate trait, or enum wrapper. In the end, it seemed to me that the enum would require more wrapping operations and changes in more places in the codebase. In contrast, the trait turned out to nicely "enforce" me not to use deref() anymore from Subdag.
- Currently various to_compact() or to_full() conversion functions only work when called on the right opposite type, to have correct semantics.
- bitset seems a safe bet: https://stackoverflow.com/questions/47853504/how-do-i-use-stdcollectionsbitset-in-stable-rust

## Test Plan

- [x] Unit tests pass. (need to resample genesis block)
- [x] Local network passes.
- [ ] Discuss whether it is ok if BatchProposals order Transmissions lexicographically. To avoid malicious use, we could select them at random when forming the ordering into the Block.
- [ ] Integration with snarkOS on existing testnet passes. Or we just go for mainnet integration.

## Related PRs
https://github.com/AleoHQ/snarkOS/pull/3000


ledger/narwhal/compact-header/Cargo.toml
```diff
@@ -0,0 +1,76 @@
...
+features = [ "preserve_order" ]
+
+[dependencies.time]
+version = "0.3"
+optional = true
+
+[dev-dependencies.bincode]
+version = "1.3"
+
+
```
ljedrz:
```suggestion
version = "1.3"

```

ledger/narwhal/subdag/src/bytes.rs
```diff
@@ -21,18 +21,21 @@ impl<N: Network> FromBytes for Subdag<N> {
         let version = u8::read_le(&mut reader)?;
         // Ensure the version is valid.
         // TODO (howardwu): For mainnet - Change the version back to 1.
-        if version != 1 && version != 2 {
+        if version != 1 && version != 2 && version != 3 {
```
ljedrz, line 24:
```suggestion
        if !(1..=3).contains(&version) {
```

vicsn: Noting the comment just above it, I think it will be easier to revert if we have my ugly version checks.

ledger/block/src/lib.rs
```diff
@@ -216,6 +215,42 @@ impl<N: Network> Block<N> {
...
         })
     }
+
+    /// Consume the Block and return a Subdag with full batch certificates.
+    pub fn into_full_subdag(self) -> Result<Subdag<N>> {
+        let Block { ratifications, solutions, transactions, aborted_transaction_ids, authority, .. } = self;
+
+        // Check if Authority is a Quorum
+        let Authority::Quorum(subdag) = authority else {
+            bail!("Cannot convert block with Quorum Authority to compact subdag");
```
raychu86:
```suggestion
            bail!("Cannot convert block with Quorum Authority to full subdag");
```

ledger/block/src/lib.rs
```diff
@@ -216,6 +215,42 @@ impl<N: Network> Block<N> {
...
+        subdag.into_full(ratification_ids, solutions, transaction_ids, aborted_transaction_ids)
+    }
+
+    /// Borrow the Block and return a Subdag with full batch certificates.
+    pub fn to_full_subdag(&self) -> Result<Subdag<N>> {
+        let Block { ratifications, solutions, transactions, aborted_transaction_ids, authority, .. } = self;
+
+        // Check if Authority is a Quorum
+        let Authority::Quorum(subdag) = authority else {
+            bail!("Cannot convert block with Quorum Authority to compact subdag");
```
raychu86:
```suggestion
            bail!("Cannot convert block with Quorum Authority to full subdag");
```

ledger/block/src/verify.rs
```diff
@@ -508,72 +510,97 @@ impl<N: Network> Block<N> {
...
+                let solutions_len = solutions.as_ref().map(|c| c.len()).unwrap_or(0);
+                // get length of transactions.
+                let transactions_len = transactions.len() + aborted_transaction_ids.len();
+                // Get bitset of unique solution indices.
+                let solution_indices = subdag.solution_indices(solutions_len)?;
+                // Get bitset of unique transaction indices.
+                let tx_indices = subdag.transaction_indices(transactions_len)?;
+                // Ensure the Subdag contains all solution indices.
+                ensure!(solution_indices.len() == solutions_len, "Subdag does not contain all tx indices.");
+                for i in 0..2usize.pow(solutions_len as u32) {
```
raychu86: If we have 256 solutions, we're looping through 2^256 times and calling `contains` that many times

ledger/block/src/verify.rs
```diff
@@ -508,72 +510,97 @@ impl<N: Network> Block<N> {
...
+                    ensure!(solution_indices.contains(i), "Subdag does not contain all tx indices.");
+                }
+                // Ensure the Subdag contains all transaction indices.
+                ensure!(tx_indices.len() == transactions_len, "Subdag does not contain all tx indices.");
+                for i in 0..2usize.pow(transactions_len as u32) {
```
raychu86: Ditto on the above,. We can theoretically have `2^TRANSACTIONS_DEPTH` transactions in a block, so would in theory have to iterate and call `contains` `2^(2^TRANSACTIONS_DEPTH)` times

ledger/narwhal/compact-header/src/lib.rs
```diff
@@ -0,0 +1,340 @@
...
+    pub fn new<'a>(
+        batch_header: &BatchHeader<N>,
+        ratifications: impl ExactSizeIterator<Item = &'a N::RatificationID>,
+        solutions: Option<impl ExactSizeIterator<Item = &'a PuzzleCommitment<N>>>,
+        transactions: impl ExactSizeIterator<Item = &'a N::TransactionID>,
+        rejected_transactions: impl ExactSizeIterator<Item = &'a N::TransactionID>,
+    ) -> Result<Self> {
+        let transmission_ids = batch_header.transmission_ids();
+
+        ensure!(ratifications.len() == 0, "Invalid batch, contains ratifications");
```
ljedrz:
```suggestion
        ensure!(ratifications.is_empty(), "Invalid batch, contains ratifications");
```

ledger/narwhal/compact-header/src/lib.rs
```diff
@@ -0,0 +1,340 @@
...
+
+    /// Convert compact header to batch header
+    pub fn into_batch_header<'a>(
+        self,
+        ratifications: impl ExactSizeIterator<Item = &'a N::RatificationID>,
+        solutions: Option<impl Iterator<Item = &'a PuzzleCommitment<N>>>,
+        transactions: impl Iterator<Item = &'a N::TransactionID>,
+        rejected_transactions: impl Iterator<Item = &'a N::TransactionID>,
+    ) -> Result<BatchHeader<N>> {
+        ensure!(ratifications.len() == 0, "Invalid batch, contains ratifications");
```
ljedrz:
```suggestion
        ensure!(ratifications.is_empty(), "Invalid batch, contains ratifications");
```

ledger/narwhal/compact-header/src/bytes.rs
```diff
@@ -0,0 +1,189 @@
...
+        // Ensure the number of transmission IDs is within bounds.
+        if num_transmission_ids as usize > Self::MAX_TRANSMISSIONS {
+            return Err(error(format!(
+                "Number of transmission IDs ({num_transmission_ids}) exceeds the maximum ({})",
+                Self::MAX_TRANSMISSIONS,
+            )));
+        }
+        // Read the number of transaction indices.
+        let num_transaction_indices = u32::read_le(&mut reader)?;
+        let mut transaction_indices = Vec::with_capacity(num_transaction_indices as usize);
```
ljedrz, line 40: (applies to all deserialization-related pre-allocations) is there some reasonable limit lower than `u32::MAX` we can apply here? 4GiB is not a large number on its own, but a few malicious messages received at once could cause an OOM

ledger/narwhal/subdag/src/lib.rs
```diff
@@ -79,28 +88,62 @@ fn sanity_check_subdag_with_dfs<N: Network>(subdag: &BTreeMap<u64, IndexSet<Batc
...
 
 impl<N: Network> PartialEq for Subdag<N> {
     fn eq(&self, other: &Self) -> bool {
-        // Note: We do not check equality on `election_certificate_ids` as it would cause `Block::eq` to trigger false-positives.
-        self.subdag == other.subdag
+        match (self, other) {
+            // Note: We do not check equality on `election_certificate_ids` as it would cause `Block::eq` to trigger false-positives.
+            (Self::Full { subdag, .. }, Self::Full { subdag: other_subdag, .. }) => subdag == other_subdag,
+            (Self::Compact { subdag, .. }, Self::Compact { subdag: other_subdag, .. }) => subdag == other_subdag,
+            _ => false,
```
ljedrz: we may want to issue a `warn!` if an incompatible comparison (full vs. compact and vice versa) is made, as it could indicate a logic bug

ljedrz:
Have you considered the following alternative to the `Subdag`?
```
trait Certificate {
    /* possibly some methods that apply to both BatchCertificate<N> and CompactCertificate<N>, 
    with some default impls for those that don't differ */ 
}

impl<N: Network> Certificate for BatchCertificate<N> { ... }

impl<N: Network> Certificate for CompactCertificate<N> { ... }

struct Subdag<N: Network, C: Certificate> {
    subdag: BTreeMap<u64, IndexSet<C>>,
    election_certificate_ids: IndexSet<Field<N>>,
}
```
If viable, this change has the potential to deduplicate some of the logic and it would allow us to omit the enum variant checks.

ledger/narwhal/compact-header/src/bytes.rs
vicsn: Great catch.

If you have more general worries, want to open a new issue for this, maybe tracking existing recent PRs tackling our serialization? Perhaps we should make a design checklist which we can keep in mind to not have to worry about it anymore?

For this specific one, as Ray mentioned above a block has at most `2^TRANSACTIONS_DEPTH`=`2^16` txs. I'll go over the whole PR and use u16 where appropriate.

ljedrz: > If you have more general worries, want to open a new issue for this, maybe tracking existing recent PRs tackling our serialization?

I recently did a round of fuzzing and didn't detect any issues other than the ones I'd known (tracked [here](https://github.com/AleoHQ/snarkVM/issues/2097)), but before going live we should probably manually scan our deserialization code for any use of `T::with_capacity` that may be troublesome, as not everything will get caught via simple fuzzing.

ledger/block/src/lib.rs
vicsn: Fixed a7ed56b2f

ledger/block/src/lib.rs
vicsn: Fixed a7ed56b2f

ledger/block/src/verify.rs
vicsn: Fixed a7ed56b2f

ledger/block/src/verify.rs
vicsn: Fixed https://github.com/AleoHQ/snarkVM/commit/a7ed56b2f32df6d0ba20515535ce31133bf24d69

ledger/narwhal/compact-header/Cargo.toml
vicsn: 70f647f9e

ledger/narwhal/compact-header/src/lib.rs
vicsn: Removed the check, it was incorrect

ledger/narwhal/compact-header/src/lib.rs
vicsn: Removed the check, it was incorrect.

ledger/narwhal/subdag/src/lib.rs
vicsn: Good idea b1c833d92

vicsn:
> If viable, this change has the potential to deduplicate some of the logic and it would allow us to omit the enum variant checks.

As discussed offline, a challenge is that we may introduce lots of duplicate logic for (de)serialization of Subdag, as well as structs which use Subdag.

ledger/src/lib.rs
```diff
@@ -92,6 +92,29 @@ pub enum RecordsFilter<N: Network> {
     SlowUnspent(PrivateKey<N>),
 }
 
+/// Helper struct to wrap transmissions in the Subdag.
+pub struct SubdagTransmissions<N: Network> {
+    pub transmissions: IndexMap<TransmissionID<N>, Transmission<N>>,
```
ljedrz, line 98: are `prior_included_transmissions` and `aborted_transmissions` a part of the `transmissions`?

vicsn: Nope, hope these new comments clarify: 4fc3258f5
Adding an intermediary check that transmissions does not contain aborted feels not worth the compute overhead, and checks down the line should independently verify the validity of the received contents.

howardwu:
Closing, was not in scope.

PR #2213, opened 2023-11-28
[ZKS-04] Track committee size.

d0cd:
This PR:
- tracks the committee size in a `metadata` mapping.
- enforces that the committee size does not exceed `200`.
- enforces that the number of **delegators** does not exceed `100_000`.

The [benches/rewards](https://github.com/AleoHQ/snarkVM/tree/benches/rewards) branch measures the performance of `atomic_post_ratify` for various numbers of delegators distributed uniformly over 200 validators.
On a 2021 M1 MAX:
```
DB: `atomic_post_ratify` with 0 delegators and 200 validators
                        time:   [18.821 ms 19.708 ms 20.456 ms]

DB: `atomic_post_ratify` with 1000 delegators and 200 validators
                        time:   [88.696 ms 104.41 ms 131.43 ms]

DB: `atomic_post_ratify` with 10000 delegators and 200 validators
                        time:   [750.47 ms 793.26 ms 838.26 ms]

DB: `atomic_post_ratify` with 100000 delegators and 200 validators
                        time:   [4.0215 s 4.3443 s 4.7269 s]

DB: `atomic_post_ratify` with 1000000 delegators and 200 validators
                        time:   [37.494 s 37.753 s 38.058 s]
```
Using memory based storage, we found:
```
Memory: `atomic_post_ratify` with 0 delegators and 200 validators
                        time:   [53.368 ms 53.403 ms 53.433 ms]

Memory: `atomic_post_ratify` with 1000 delegators and 200 validators
                        time:   [181.87 ms 182.05 ms 182.29 ms]

Memory: `atomic_post_ratify` with 10000 delegators and 200 validators
                        time:   [1.3540 s 1.3547 s 1.3553 s]

Memory: `atomic_post_ratify` with 100000 delegators and 200 validators
                        time:   [13.088 s 13.224 s 13.480 s]

Memory: `atomic_post_ratify` with 1000000 delegators and 200 validators
                        time:   [131.51 s 131.59 s 131.68 s]
```

Helps with addressing Audit Finding: **[zksecurity 04] Garbage Collection Can Block Commits From Happening**

PR #2308, opened 2024-01-18
[Coinbase Go/No-Go] Bond balances in genesis block

vicsn:
## Motivation

We want to be able to bond balances in the genesis block.

This PR modifies `Genesis::Ratify`, such that it contains the desired state of the `committee`, `account`, and `bonded` mappings in `credits.aleo`. This initial state is checked for consistency before it is directly injected into the `FinalizeStore`.

## Related PRs

https://github.com/AleoHQ/snarkOS/pull/3016


ledger/block/src/ratify/mod.rs
```diff
@@ -24,11 +24,14 @@ use indexmap::IndexMap;
 type Variant = u8;
 /// A helper type to represent the public balances.
 type PublicBalances<N> = IndexMap<Address<N>, u64>;
+/// A helper type to represent the bonded balances.
+type BondedBalances<N> = IndexMap<Address<N>, u64>;
```
d0cd:
```suggestion
type BondedBalances<N> = IndexMap<Address<N>, (Address<N>, u64);
```
Gah! I realized that we should instead expect users to provide a validator address along with a bonded amount. This should now line up with the logic in `credits.aleo`.

ledger/block/src/ratify/mod.rs
```diff
@@ -24,11 +24,14 @@ use indexmap::IndexMap;
...
 
+#[allow(clippy::large_enum_variant)]
```
d0cd: I wonder if this has a non-trivial impact on object/storage size?

d0cd (dismissed):

synthesizer/src/vm/finalize.rs
```diff
@@ -716,6 +716,25 @@ impl<N: Network, C: ConsensusStorage<N>> VM<N, C> {
...
+                            Some(Value::Plaintext(Plaintext::Literal(Literal::U64(value), _))) => {
+                                (*value).saturating_add(*amount)
+                            }
+                            None => *amount,
+                            v => bail!("Critical bug in pre-ratify - Invalid bonded balance type ({v:?})"),
+                        })));
+                        // Update the bonded balance in finalize storage.
+                        let operation = store.update_key_value(program_id, bonded_mapping, key, next_value)?;
+                        finalize_operations.push(operation);
+                    }
```
raychu86: I haven't looked into this in depth, but outside of the `bonded` mapping, we need to update the `committee_state` in the `committee` mapping and also the number of delegators if/when this PR gets merged in - https://github.com/AleoHQ/snarkVM/pull/2213.

In addition, we need checks that the delegated bonding amount meets all the requirements (for example, delegation must be more than X credits)

ledger/block/src/ratify/mod.rs
vicsn: It might have some impact, but the jump from `(Committee<N>, PublicBalances<N>)` to `(Committee<N>, PublicBalances<N>, BondedBalances<N>)` won't be that big.

ledger/block/src/ratify/mod.rs
vicsn: Makes sense - 74d29bf2c
Can you explain why we don't have to use/set the validator_address in the value: https://github.com/AleoHQ/snarkVM/blob/bond_balances_in_genesis/synthesizer/src/vm/finalize.rs#L775
Even though it is clearly a part of `bond_state`: https://github.com/AleoHQ/snarkVM/blob/testnet3/synthesizer/program/src/resources/credits.aleo#L45

d0cd (commented):
Updated to address feedback.

raychu86:
Would be nice to have some tests that ensures the bond balances is equivalent to actual `bond_public` transactions.

synthesizer/src/vm/finalize.rs
```diff
@@ -775,17 +775,60 @@ impl<N: Network, C: ConsensusStorage<N>> VM<N, C> {
...
-                    for (validator, (microcredits, _)) in committee.members() {
-                        // Insert the validator into the stakers.
-                        stakers.insert(*validator, (*validator, *microcredits));
+                    // Calculate the stake per validator using `bonded_balances`.
+                    let mut stake_per_validator = IndexMap::with_capacity(committee.members().len());
+                    for (address, (validator_address, amount)) in bonded_balances {
+                        // Check that the amount meets the minimum requirement, depending on whether the address is a validator.
+                        if *address == *validator_address {
+                            ensure!(
+                                *amount >= 1_000_000_000_000_u64,
```
raychu86: Is there a way to ensure that these values match the `credits.aleo` minimum requirements? Might be good to write some comments and maybe add some tests for enforcement to catch any changes down the line.

ledger/block/src/ratify/mod.rs
howardwu: Considering every block contains this enum, and we will always pay the worst-case memory footprint, I don't understand why one chose to ignore this clippy warning.

You should follow the advise of `clippy` to resolve it (I assume its to use `Box`). Otherwise, every enum variant will always pay the worst case bound.

ledger/block/src/ratify/serialize.rs
```diff
@@ -19,11 +19,12 @@ impl<N: Network> Serialize for Ratify<N> {
     fn serialize<S: Serializer>(&self, serializer: S) -> Result<S::Ok, S::Error> {
         match serializer.is_human_readable() {
             true => match self {
-                Self::Genesis(committee, public_balances) => {
+                Self::Genesis(committee, public_balances, bonded_balances) => {
                     let mut input = serializer.serialize_struct("Ratify", 3)?;
```
howardwu: Why is this 3?
```suggestion
                    let mut input = serializer.serialize_struct("Ratify", 4)?;
```

synthesizer/src/vm/finalize.rs
```diff
@@ -1641,11 +1696,6 @@ finalize compute:
         // Check that the storage was not updated.
         let program_id = ProgramID::from_str("testing.aleo").unwrap();
         let mapping_name = Identifier::from_str("entries").unwrap();
-        let value = vm
-            .finalize_store()
-            .get_value_speculative(program_id, mapping_name, &Plaintext::from(Literal::Address(address)))
-            .unwrap();
-        println!("{:?}", value);
```
howardwu, line 1673: Why was this removed?

synthesizer/src/vm/finalize.rs
```diff
@@ -791,17 +792,66 @@ impl<N: Network, C: ConsensusStorage<N>> VM<N, C> {
...
+                        ensure!(
+                            expected_amount == amount,
+                            "Ratify::Genesis(..) inconsistent staked amount for validator {validator_address}",
+                        );
                     }
 
                     // Construct the next committee map and next bonded map.
                     let (next_committee_map, next_bonded_map) =
-                        to_next_commitee_map_and_bonded_map(committee, &stakers);
+                        to_next_commitee_map_and_bonded_map(committee, bonded_balances);
```
howardwu, line 860: Why is this `bonded_balances`? If this exists, what is the purpose of `stake_per_validator`?

Please clarify with a test case what the intended behavior and edge cases are, and share the commit with the test here.

synthesizer/src/vm/finalize.rs
```diff
@@ -791,17 +792,66 @@ impl<N: Network, C: ConsensusStorage<N>> VM<N, C> {
...
-                        // Insert the validator into the stakers.
-                        stakers.insert(*validator, (*validator, *microcredits));
+                    // Ensure that all members of the committee are open.
+                    ensure!(
+                        committee.members().iter().all(|(_, (_, is_open))| *is_open),
+                        "Ratify::Genesis(..) all members of the committee must be open"
+                    );
+
+                    // Calculate the stake per validator using `bonded_balances`.
+                    let mut stake_per_validator = IndexMap::with_capacity(committee.members().len());
```
howardwu, line 809: Rename this to `stakers`, why was this changed? Add back:
```
// Initialize the stakers.
let mut stakers = IndexMap::with_capacity(committee.members().len());
```

synthesizer/src/vm/finalize.rs
howardwu: How is there not a test catching this? Please share a commit that demonstrates catching this erroneous case.

synthesizer/src/vm/finalize.rs
```diff
@@ -791,17 +792,66 @@ impl<N: Network, C: ConsensusStorage<N>> VM<N, C> {
...
-                    let mut stakers = IndexMap::with_capacity(committee.members().len());
-                    // Iterate over the committee members.
-                    for (validator, (microcredits, _)) in committee.members() {
-                        // Insert the validator into the stakers.
-                        stakers.insert(*validator, (*validator, *microcredits));
+                    // Ensure that all members of the committee are open.
+                    ensure!(
+                        committee.members().iter().all(|(_, (_, is_open))| *is_open),
+                        "Ratify::Genesis(..) all members of the committee must be open"
+                    );
```
howardwu: Remove this. I'm not understanding why this design consideration was introduced.

synthesizer/src/vm/finalize.rs
```diff
@@ -791,17 +792,66 @@ impl<N: Network, C: ConsensusStorage<N>> VM<N, C> {
...
+                            ensure!(
+                                committee.is_committee_member_open(*validator_address),
+                                "Ratify::Genesis(..) the delegator {address} is delegating to a closed validator {validator_address}",
+                            );
+                        }
+                        // Ensure that each staker has an entry in `public_balances`.
+                        ensure!(
+                            public_balances.contains_key(address),
+                            "Ratify::Genesis(..) the staker {address} is missing from the public balances",
+                        );
```
howardwu: Why is this check necessary? Unless you decrement from `public_balances`, remove this check, it is extraneous.

synthesizer/src/vm/finalize.rs
```diff
@@ -791,17 +792,66 @@ impl<N: Network, C: ConsensusStorage<N>> VM<N, C> {
...
+                        // Accumulate the staked amount per validator.
+                        let total = stake_per_validator.entry(validator_address).or_insert(0u64);
+                        *total = total.saturating_add(*amount);
```
howardwu, line 830: Why does this logic exist? It does not do anything!

synthesizer/src/vm/finalize.rs
```diff
@@ -791,17 +792,66 @@ impl<N: Network, C: ConsensusStorage<N>> VM<N, C> {
...
+
+                    // Check that `committee` is consistent with `stake_per_validator`.
+                    for (validator_address, amount) in stake_per_validator {
+                        // Retrieve the expected validator stake from the committee.
+                        let expected_amount = match committee.members().get(validator_address) {
+                            Some((amount, _)) => *amount,
+                            None => bail!(
+                                "Ratify::Genesis(..) found a validator in the bonded balances that is not in the committee"
+                            ),
+                        };
```
howardwu: nit: use `let else` syntax for better readability.

synthesizer/src/vm/finalize.rs
d0cd: This check was introduced to be consistent with `credits.aleo`.
When a validator initially bonds through `bond_public`, they are default set to be open. See [here](https://github.com/AleoHQ/snarkVM/blob/2c3b775f9c869e8ba03d10e91b94d8bf8297d388/synthesizer/program/src/resources/credits.aleo#L129).

synthesizer/src/vm/finalize.rs
d0cd: This check was introduced to be consistent with `credits.aleo`.
In the finalize block of `bond_public`, we [check](https://github.com/AleoHQ/snarkVM/blob/2c3b775f9c869e8ba03d10e91b94d8bf8297d388/synthesizer/program/src/resources/credits.aleo#L163) that a corresponding entry in the `account` mapping exists for both validators and delegators.

synthesizer/src/vm/finalize.rs
d0cd: It appears to be a dangling `println` statement. 
`value` does not appear to be used elsewhere.

synthesizer/src/vm/finalize.rs
d0cd: `total` is a mutable reference to the entry in `stake_per_validator`.
The deref and assignment, on L828, updates the entry.

ledger/block/src/ratify/mod.rs
vicsn: > why one chose to ignore this clippy warning
By being too much in a get shit done mood, my bad.

@d0cd since you've been finishing off the PR, want to wrap the contents in Box? I also just recently noticed we're already using this pattern in snarkVM.

synthesizer/src/vm/finalize.rs
d0cd: `bonded_balances` contains the expected state of the `bonded` mapping in `credits.aleo`.
The purpose of `stake_per_validator` is to check that the stakes in `committee` in fact equals the sum of the stakes in `bonded_balances`.
The commit can be found [here](https://github.com/AleoHQ/snarkVM/pull/2308/commits/33c4af3488e7ed4736ed8a4d2432ac6e4decaaf4)

synthesizer/src/vm/finalize.rs
d0cd: This stakers (which can be both validators and delegators) is provided in `bonded_balances`, which is checked for consistency against the `committee` using `stake_per_validator`.
We could consider renaming `bonded_balances` to `stakers` for more clarity.
We could also consider removing some of the redundant information between `committee` and `bonded_balances`.

ledger/block/src/ratify/mod.rs
d0cd: Done! By boxing the elements of `Genesis::Ratify` the size reduces from 240 to 32 bytes.

synthesizer/src/vm/finalize.rs
```diff
@@ -1750,4 +1799,589 @@ finalize compute:
...
+        assert_eq!(committee_1, committee_2);
+
+        let bonded_1 = vm_1.finalize_store().get_mapping_confirmed(program_id, bonded_mapping_name).unwrap();
+        let bonded_2 = vm_2.finalize_store().get_mapping_confirmed(program_id, bonded_mapping_name).unwrap();
+        assert_eq!(bonded_1, bonded_2);
+
+        let unbonding_1 = vm_1.finalize_store().get_mapping_confirmed(program_id, unbonding_mapping_name).unwrap();
+        let unbonding_2 = vm_2.finalize_store().get_mapping_confirmed(program_id, unbonding_mapping_name).unwrap();
+        assert_eq!(unbonding_1, unbonding_2);
+
```
raychu86, line 2357: With #2213 merged, we should check the number of delegators and validators in the `metadata` mapping are equivalent.

synthesizer/src/vm/finalize.rs
```diff
@@ -791,20 +792,68 @@ impl<N: Network, C: ConsensusStorage<N>> VM<N, C> {
...
                     // Construct the next committee map and next bonded map.
                     let (next_committee_map, next_bonded_map) =
-                        to_next_commitee_map_and_bonded_map(committee, &stakers);
+                        to_next_commitee_map_and_bonded_map(committee, bonded_balances);
 
                     // Insert the next committee into storage.
-                    store.committee_store().insert(state.block_height(), committee.clone())?;
+                    store.committee_store().insert(state.block_height(), *(committee.clone()))?;
                     // Store the finalize operations for updating the committee and bonded mapping.
                     finalize_operations.extend(&[
```
raychu86, line 865: With #2213 merged, we need to update the `metadata` mapping here in the Genesis ratification.

synthesizer/src/vm/finalize.rs
```diff
@@ -1875,115 +1939,1009 @@ finalize compute:
...
+        // Get the account mapping.
+        let actual_account = vm.finalize_store().get_mapping_confirmed(program_id, account_mapping_name).unwrap();
+        let expected_account = public_balances
+            .iter()
+            .map(|(address, amount)| {
+                (Plaintext::from_str(&address.to_string()).unwrap(), Value::from_str(&format!("{amount}u64")).unwrap())
+            })
+            .collect_vec();
+        assert_eq!(actual_account.len(), expected_account.len());
+        // Check that all entries except for the one for the first validator are the same.
```
raychu86: Fix comment. Also why is the first validator entry different? 

```suggestion
        // Check that all entries except for the first validator are the same.
```

synthesizer/src/vm/finalize.rs
```diff
@@ -1875,115 +1939,1009 @@ finalize compute:
...
+            .map(|(address, (validator, amount))| {
+                (
+                    Plaintext::from_str(&address.to_string()).unwrap(),
+                    Value::from_str(&format!("{{ validator: {validator}, microcredits: {amount}u64 }}")).unwrap(),
+                )
+            })
+            .collect_vec();
+        assert_eq!(actual_bonded.len(), expected_bonded.len());
+        for entry in actual_bonded.iter() {
+            assert!(expected_bonded.contains(entry));
```
raychu86, line 2161: Why don't we check the amounts in the bonded and account mappings?

synthesizer/src/vm/finalize.rs
d0cd: This is because the first validator also executes a number of additional transactions in the `genesis_quorum` method. 
Consequently, its balance after adding the genesis block is not the same as the public balance provided in `Genesis::Ratify`.

synthesizer/src/vm/finalize.rs
raychu86: NVM, i see now that the `actual_bonded` and `expected_bonded` variables are Vectors, so the contains call will check that the balance is correct.

synthesizer/src/vm/finalize.rs
raychu86: Could you maybe clarify that in the comment? Otherwise, we won't know why down the line.

howardwu (approved):

Issue #2299, opened 2024-01-11
[Bug] Constraint with`Transactions::<N>::MAX_TRANSACTIONS` will make the network stuck in some circumstances


feezybabee:
# https://hackerone.com/reports/2310792

## Summary:

In snarkVM, the number of transactions and aborted_transactions is limited by `Transactions::<N>::MAX_TRANSACTIONS` (65536):

```rust
    fn verify_transactions(&self) -> Result<()> {
        let height = self.height();

        // Ensure there are transactions.
        ensure!(!self.transactions.is_empty(), "Block {height} must contain at least 1 transaction");

        // Ensure the number of transactions is within the allowed range.
        if self.transactions.len() > Transactions::<N>::MAX_TRANSACTIONS {
            bail!(
                "Cannot validate a block with more than {} confirmed transactions",
                Transactions::<N>::MAX_TRANSACTIONS
            );
        }

        // Ensure the number of aborted transaction IDs is within the allowed range.
        if self.aborted_transaction_ids.len() > Transactions::<N>::MAX_TRANSACTIONS {
            bail!(
                "Cannot validate a block with more than {} aborted transaction IDs",
                Transactions::<N>::MAX_TRANSACTIONS
            );
        }
```

However, this can be violated with BullShark in some circumstances and will block the whole network.

## Steps To Reproduce:

No

## Proof-of-Concept (PoC)

In Bullshark, the `subdag` may commit many transactions at one time. For example, the max number of transaction in one batch is:

```pub const MAX_TRANSMISSIONS_PER_BATCH: usize = 250; // transmissions```

Suppose we have 100 validators,  at each round's certificate, we have at least 34 certificates to commit. If in some circumstances, we commit a subdag with 20 rounds, then it will contain 34 \* 250 \* 20 = 165000 transactions. As 165000 > 2 \* MAX_TRANSACTIONS, the network will halt because snarkVM won't commit:

```
        atomic_finalize!(self.finalize_store(), FinalizeMode::DryRun, {
            // Ensure the number of transactions does not exceed the maximum.
            if num_transactions > 2 * Transactions::<N>::MAX_TRANSACTIONS {
                // Note: This will abort the entire atomic batch.
                return Err(format!(
                    "Too many transactions in the block - {num_transactions} (max: {})",
                    2 * Transactions::<N>::MAX_TRANSACTIONS
                ));
            }
```

[...]

PR #2309, opened 2024-01-18
Increase aborted transaction limit

evan-schott:
# Motivation
- The existing limit on the maximum number of aborted transactions is too small. 
- This PR bumps it from `65,536` to `2,500,000`, as is calculated by `MAX_GC_ROUNDS * MAX_COMMITTEE_SIZE * MAX_TRANSMISSIONS_PER_BATCH` 
- Although there is a negligible probability of encountering this theoretical worst case, the ramifications of being maximally safe only increase the block size by 18%. 
	- Worse case storage increase:
		- 80MB to store all aborted transactions
	- But we would already be paying: 
		- 254MB to store the subDAG
		- 185MB to store all executions at the very minimum
		- Having to store the subdag is the main storage cost in extreme cases like this as we are storing all the batch headers and signatures that any validator has seen from any other validator.

### DIFF
--- a/synthesizer/src/vm/finalize.rs
+++ b/synthesizer/src/vm/finalize.rs
@@ -172,11 +172,11 @@ impl<N: Network, C: ConsensusStorage<N>> VM<N, C> {
         // Perform the finalize operation on the preset finalize mode.
         atomic_finalize!(self.finalize_store(), FinalizeMode::DryRun, {
             // Ensure the number of transactions does not exceed the maximum.
-            if num_transactions > 2 * Transactions::<N>::MAX_TRANSACTIONS {
+            if num_transactions > Transactions::<N>::MAX_ABORTED_TRANSACTIONS {
                 // Note: This will abort the entire atomic batch.
                 return Err(format!(
                     "Too many transactions in the block - {num_transactions} (max: {})",
-                    2 * Transactions::<N>::MAX_TRANSACTIONS
+                    Transactions::<N>::MAX_ABORTED_TRANSACTIONS
                 ));
             }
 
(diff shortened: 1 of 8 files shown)


raychu86:
LGTM so far, except i think there is a missing check in `Block::from_bytes`. We need to update that check for the number of aborted transactions. I believe it is still using the old `MAX_TRANSACTIONS` variable.

This should also fix: #2299

evan-schott:
Should this be changed to `250`? 
https://github.com/AleoHQ/snarkVM/blob/c22922378001dcc6e561d8db193c8e43e7b18a90/ledger/narwhal/batch-header/src/lib.rs#L65

synthesizer/src/vm/finalize.rs
```diff
@@ -172,11 +172,11 @@ impl<N: Network, C: ConsensusStorage<N>> VM<N, C> {
         // Perform the finalize operation on the preset finalize mode.
         atomic_finalize!(self.finalize_store(), FinalizeMode::DryRun, {
             // Ensure the number of transactions does not exceed the maximum.
-            if num_transactions > 2 * Transactions::<N>::MAX_TRANSACTIONS {
+            if num_transactions > Transactions::<N>::MAX_ABORTED_TRANSACTIONS {
```
raychu86, line 175: This should be something like this:

```suggestion
            if num_transactions > Transactions::<N>::MAX_TRANSACTIONS + Transactions::<N>::MAX_ABORTED_TRANSACTIONS {
```

raychu86:
> Should this be changed to `250`?
> 
> https://github.com/AleoHQ/snarkVM/blob/c22922378001dcc6e561d8db193c8e43e7b18a90/ledger/narwhal/batch-header/src/lib.rs#L65

There is a difference between the `mainnet` and `testnet3` branch currently. I'm not sure what the state is here, but I believe it should be something like `MAX_TRANMISSIONS_PER_BATCH = 250`. Need to double check

howardwu:
I have dropped the `MAX_TRANSMISSIONS_PER_BATCH` to `50`.

So instead of `2,500,000` maximum aborted transmissions, this is now set to `500,000`.

howardwu (approved):

PR #2318, opened 2024-01-22
Fix compilation warnings

ljedrz:
These include:
- unused imports
- ineffective reexports of impls and macros
- one redundant call to `deref`

### DIFF
--- a/circuit/environment/src/lib.rs
+++ b/circuit/environment/src/lib.rs
@@ -29,6 +29,7 @@ pub mod helpers;
 pub use helpers::*;
 
 pub mod macros;
+#[allow(unused_imports)]
 pub use macros::*;
 
 pub mod traits;
--- a/circuit/environment/src/macros/mod.rs
+++ b/circuit/environment/src/macros/mod.rs
@@ -13,10 +13,13 @@
 // limitations under the License.
 
 mod metrics;
+#[allow(unused_imports)]
 pub use metrics::*;
 
 mod scope;
+#[allow(unused_imports)]
 pub use scope::*;
 
 mod witness;
+#[allow(unused_imports)]
 pub use witness::*;
--- a/fields/src/lib.rs
+++ b/fields/src/lib.rs
@@ -42,6 +42,7 @@ mod legendre;
 pub use legendre::*;
 
 mod to_field_vec;
+#[allow(unused_imports)]
 pub use to_field_vec::*;
 
 pub mod traits;
--- a/utilities/src/serialize/mod.rs
+++ b/utilities/src/serialize/mod.rs
@@ -19,6 +19,7 @@ mod helpers;
 pub use helpers::*;
 
 mod impls;
+#[allow(unused_imports)]
 pub use impls::*;
 
 mod flags;
--- a/synthesizer/process/src/stack/call/mod.rs
+++ b/synthesizer/process/src/stack/call/mod.rs
@@ -66,8 +66,7 @@ impl<N: Network> CallTrait<N> for Call<N> {
         let timer = timer!("Call::evaluate");
 
         // Load the operands values.
-        let inputs: Vec<_> =
-            self.operands().iter().map(|operand| registers.load(stack.deref(), operand)).try_collect()?;
+        let inputs: Vec<_> = self.operands().iter().map(|operand| registers.load(stack, operand)).try_collect()?;
 
         // Retrieve the substack and resource.
         let (substack, resource) = match self.operator() {
--- a/utilities/src/serialize/impls.rs
+++ b/utilities/src/serialize/impls.rs
@@ -13,7 +13,7 @@
 // limitations under the License.
 
 pub use crate::{
-    io::{self, Read, Write},
+    io::{Read, Write},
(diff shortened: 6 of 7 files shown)


joske (approved):

circuit/environment/src/lib.rs
```diff
@@ -29,7 +29,6 @@ pub mod helpers;
 pub use helpers::*;
 
 pub mod macros;
-pub use macros::*;
```
howardwu, line 32: Can you work around this instead of removing `pub use`?

circuit/environment/src/macros/mod.rs
```diff
@@ -13,10 +13,7 @@
 // limitations under the License.
 
 mod metrics;
-pub use metrics::*;
 
 mod scope;
-pub use scope::*;
 
 mod witness;
-pub use witness::*;
```
howardwu: Can you work around this instead of removing `pub use`?

fields/src/lib.rs
```diff
@@ -42,7 +42,6 @@ mod legendre;
 pub use legendre::*;
 
 mod to_field_vec;
-pub use to_field_vec::*;
```
howardwu, line 45: Can you work around this instead of removing `pub use`?

utilities/src/serialize/mod.rs
```diff
@@ -19,7 +19,6 @@ mod helpers;
 pub use helpers::*;
 
 mod impls;
-pub use impls::*;
```
howardwu, line 22: Can you work around this instead of removing `pub use`?

howardwu:
@ljedrz can you clarify the command that triggers these? I tried reproducing locally and could not.

ljedrz:
These pop up when you use `snarkVM` as a dependency (at least locally); I can see them when e.g. compiling my small binary to check block deserialization speed.

The reason you're having difficulty reproducing it from the level of `snarkOS` or `snarkVM` is probably due to the recent inclusion of the `rust-toolchain` file - the `1.72.1` version might not be able to detect these issues just yet.

ljedrz:
Rebased and applied the review suggestions.

howardwu:
I dont think clippy ignores is the solution either. We need to fix it properly. Whats the way to replicate the errors?

ljedrz:
> Whats the way to replicate the errors?

You can create an empty new project with `cargo new --lib <name>` with a path-based dependency on snarkvm:
```
snarkvm = { path = "/path/to/snarkVM" }
```

And then it's just a matter of `cargo check` in that project's main folder.

howardwu:
Followed those steps and I'm still not seeing it. What Rust are you on?

ljedrz:
I'm using `1.75` (current stable).

howardwu:
I'm also on 1.75, I only see the `.deref()` one.

howardwu:
Never mind, `cargo test` does it for me now.

fields/src/lib.rs
```diff
@@ -42,6 +42,7 @@ mod legendre;
 pub use legendre::*;
 
 mod to_field_vec;
+#[allow(unused_imports)]
```
howardwu, line 45:
```suggestion
```
I'm not seeing this case appear. Can you confirm or revert?

howardwu (changes requested):
I also didn't see any of the `snarkvm_utilities` cases appear, can you confirm or revert?

fields/src/lib.rs
ljedrz: I confirm:
```
    Checking snarkvm-fields v0.16.18 (/path/to/snarkVM/fields)
warning: unused import: `to_field_vec::*`
  --> /path/to/snarkVM/fields/src/lib.rs:45:9
   |
45 | pub use to_field_vec::*;
   |         ^^^^^^^^^^^^^^^
   |
   = note: `#[warn(unused_imports)]` on by default
```

ljedrz:
I confirm the `snarkvm-utilities` ones too:
```
warning: unused import: `self`
  --> /path/to/snarkVM/utilities/src/serialize/impls.rs:16:10
   |
16 |     io::{self, Read, Write},
   |          ^^^^
   |
   = note: `#[warn(unused_imports)]` on by default

warning: unused import: `impls::*`
  --> /path/to/snarkVM/utilities/src/serialize/mod.rs:22:9
   |
22 | pub use impls::*;
   |         ^^^^^^^^

warning: unused imports: `FromBytes`, `ToBytes`, `Vec`, `self`
  --> /path/to/snarkVM/utilities/src/serialize/traits.rs:17:10
   |
17 |     io::{self, Read, Write},
   |          ^^^^
18 |     FromBytes,
   |     ^^^^^^^^^
19 |     ToBytes,
   |     ^^^^^^^
20 |     Vec,
   |     ^^^

warning: `snarkvm-utilities` (lib) generated 3 warnings (run `cargo fix --lib -p snarkvm-utilities` to apply 3 suggestions)
```
The reason why you're not seeing them might be your IDE or some other tool doing partial `debug` compilation before you have a chance to compile manually.

howardwu (approved):

PR #2320, opened 2024-01-23
Assorted deserialization-related perf improvements

ljedrz:
This PR contains a few small perf improvements related to deserialization (of blocks in particular).

### DIFF
--- a/fields/src/macros.rs
+++ b/fields/src/macros.rs
@@ -104,12 +104,12 @@ macro_rules! sqrt_impl {
                 let l_minus_one_times_k = n - 1 - k_2;
                 let l_minus_one = l_minus_one_times_k / k;
                 let l = l_minus_one + 1;
-                let mut l_s: Vec<u64> = Vec::with_capacity(k as usize);
-                l_s.resize(l_s.len() + k_1 as usize, l_minus_one);
-                l_s.resize(l_s.len() + k_2 as usize, l);
+
+                let l_s =
+                    || std::iter::repeat(l_minus_one).take(k_1 as usize).chain(std::iter::repeat(l).take(k_2 as usize));
 
                 let mut l_sum = 0;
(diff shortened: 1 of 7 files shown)


fields/src/macros.rs
```diff
@@ -104,12 +104,12 @@ macro_rules! sqrt_impl {
                 let l_minus_one_times_k = n - 1 - k_2;
                 let l_minus_one = l_minus_one_times_k / k;
                 let l = l_minus_one + 1;
-                let mut l_s: Vec<u64> = Vec::with_capacity(k as usize);
-                l_s.resize(l_s.len() + k_1 as usize, l_minus_one);
-                l_s.resize(l_s.len() + k_2 as usize, l);
+
+                let l_s =
+                    || std::iter::repeat(l_minus_one).take(k_1 as usize).chain(std::iter::repeat(l).take(k_2 as usize));
```
howardwu, line 109: Are you sure these are equivalent vectors/iterators?

Notice before we have `l_minus_one` up to `k1` then between `k1` and `k2` lengths, we have `l`.

In the new version, we have `k1` elements of `l_minus_one` and `k2` elements of `l`.

Shouldn't the new version finish with `k2-k1` (checked/safe) elements of `l`?

ljedrz: I just double-checked it in the playground with some randomized values, the results are the same between both versions: https://play.rust-lang.org/?version=stable&mode=debug&edition=2021&gist=891bc02268eee1344a80fd81d8c147bf.

ljedrz: I think the current version is a bit misleading, due to its use of `Vec::len`; the length of `l_s` initially equals `0`, then `k_1` (after 1st resize), and finally `k_1 + k_2` (the 2nd resize)

howardwu (approved):

PR #2321, opened 2024-01-23
[Fix] Deployment verification for programs that create external records.

d0cd:
Addresses https://github.com/AleoHQ/snarkOS/issues/3020.


vicsn (approved):
LGTM, did you test that the new test `test_deployment_with_external_records` indeed fails before this PR?

howardwu (approved):

PR #2327, opened 2024-01-30
fix: is_bond and is_unbond match credits.aleo

randomsleep:
Match `is_bond` and `is_unbond` with `credits.aleo`.


howardwu (approved):

howardwu:
Great catch!

PR #2329, opened 2024-02-02
Fix bug in Poseidon internal squeezing function.

acoglio:
The permutation must be done unconditionally at this point of the code, which is reached if there are more field elements to squeeze than are left in the current state vector (i.e. if `squeeze_index + remaining.len() > RATE` holds). Whether the remaining elements to squeeze happen to be the same number as the rate or not, is irrelevant here.

[This code](https://github.com/AleoHQ/snarkVM/compare/testnet3...poseidon-squeeze-internal-bug) demonstrates the issue: it prints `Permutation.` once instead of twice.

This is an illustration, where the small arrows indicate the current squeeze index:
poseidon

This should not be exploitable by executing Aleo instructions because those should make a single call to the sponge function, while triggering the bug requires two calls. Nonetheless, this should be fixed because:
1. The `squeeze_internal` function violates its own specification. And we may get in trouble if we use this function for other purposes in the future.
2. The `if` condition takes extra time and generates extra R1CS constraints.

The other two implementations of Poseidon in snarkVM, one under `algorithms` and one under `console`, look fine.

### DIFF
--- a/circuit/algorithms/src/poseidon/hash_many.rs
+++ b/circuit/algorithms/src/poseidon/hash_many.rs
@@ -130,10 +130,9 @@ impl<E: Environment, const RATE: usize> Poseidon<E, RATE> {
             let num_squeezed = RATE - squeeze_index;
             remaining[..num_squeezed].clone_from_slice(&state[start..(start + num_squeezed)]);
 
-            // Unless we are done with squeezing in this call, permute.
-            if remaining.len() != RATE {
-                self.permute(state);
-            }
+            // Permute.
+            self.permute(state);
+
             // Repeat with the updated output slice and squeeze index.
             remaining = &mut remaining[num_squeezed..];
             squeeze_index = 0;


howardwu:
Was this formally checked?

From pencil and paper analysis, it seems we still need the if conditional, but gate keep it further on the rate squeeze index.

acoglio:
When we get to that point of the code (line 133), `squeeze_index + remaining.len() > RATE` holds (otherwise we would have `return`ed at line 126), we have just squeezed `remaining[..RATE - squeeze_index]` (line 131), and we have more elements to squeeze because of the inequality above. So we must always permute. If we don't, we'll cycle through the same outputs, as the test function `unchanging_outputs` that I've just added [here](https://github.com/AleoHQ/snarkVM/compare/testnet3...poseidon-squeeze-internal-bug) shows:
```
/Users/ac/.cargo/bin/cargo test --color=always --lib poseidon::hash_many::tests::unchanging_outputs --no-fail-fast --manifest-path /Users/ac/Work/aleo/snarkVM/circuit/algorithms/Cargo.toml -- --format=json --exact -Z unstable-options --show-output
Testing started at 11:21 PM ...
    Finished test [optimized + debuginfo] target(s) in 0.39s
     Running unittests src/lib.rs (target/debug/deps/snarkvm_circuit_algorithms-81d681ef621baf9c)
RATE = 4 and CAPACITY = 1
Permutation.
Squeeze 1:
Output = 8014138239963369935047609395555728362890000218898320263975885605530462638507field.constant
Squeeze 4:
Output = 3754602288276171440606829666192192209811444671877104196154271669470735360815field.constant
Output = 3520756939327963497196120795172468044619780785356975233775072768115467754699field.constant
Output = 7291943308998054908115465624790668716166155272010779330149867604185725261780field.constant
Output = 8014138239963369935047609395555728362890000218898320263975885605530462638507field.constant
Squeeze 4:
Output = 3754602288276171440606829666192192209811444671877104196154271669470735360815field.constant
Output = 3520756939327963497196120795172468044619780785356975233775072768115467754699field.constant
Output = 7291943308998054908115465624790668716166155272010779330149867604185725261780field.constant
Output = 8014138239963369935047609395555728362890000218898320263975885605530462638507field.constant
Squeeze 4:
Output = 3754602288276171440606829666192192209811444671877104196154271669470735360815field.constant
Output = 3520756939327963497196120795172468044619780785356975233775072768115467754699field.constant
Output = 7291943308998054908115465624790668716166155272010779330149867604185725261780field.constant
Output = 8014138239963369935047609395555728362890000218898320263975885605530462638507field.constant

Process finished with exit code 0
```

We have a formal specification of Poseidon in ACL2 [here](https://github.com/AleoHQ/aleo-acl2/blob/master/poseidon/poseidon.lisp), which uses a simpler logic to absorb and squeeze just one element at a time in a loop, and it has no condition guarding the permutation: when the index reaches the RATE, permute. To show the hash outputs of this formal specification I need to first find out the exact Poseidon parameters used in the tests.

acoglio:
After @bendyarm found and formalized the Poseidon parameters used in the tests, I ran the ACL2 specification of Poseidon and got the following, corresponding to the `unchanging_outputs` test I mentioned above:
```
POSEIDON !>(hash nil (rate-4-alpha-17-parameters) 13)
(8014138239963369935047609395555728362890000218898320263975885605530462638507
 3754602288276171440606829666192192209811444671877104196154271669470735360815
 3520756939327963497196120795172468044619780785356975233775072768115467754699
 7291943308998054908115465624790668716166155272010779330149867604185725261780
 4861252863280559719727425584389798213287935889129026076725228966690642923365
 7217939601206443418783545003015102564469781528773791656192329836691284752787
 3847069254229444540813803809449810546677347076231484631105669574751680204577
 6360940757099987115873607022528148012488874065495155845457272458813281528750
 2189011774880053270414424101496536814373490723215979638017381557547713684138
 4766479750890547465908202855408334758672072428811537677244762545286031523535
 8246536114497490327664290382136560398507374279846399949824149814607850183338
 2682923563373717802192849836083465272502749583724589006688086528014311083813
 3975305146268908912478510511280257329509292235984784671196366429535597245196)
POSEIDON !>
```
The first four are the same as in the Rust code, and the remaining nine change as they should.

howardwu (approved):

PR #2330, opened 2024-02-02
WIP: Introduce committee round lag

vicsn:
## Motivation

Add support for returning committee with lag.

## Test Plan

...

### DIFF
--- a/ledger/block/src/verify.rs
+++ b/ledger/block/src/verify.rs
@@ -193,13 +193,13 @@ impl<N: Network> Block<N> {
             }
             Authority::Quorum(subdag) => {
                 // Compute the expected leader.
-                let expected_leader = current_committee.get_leader(expected_round)?;
+                // let expected_leader = current_committee.get_leader(expected_round)?;
                 // Ensure the block is authored by the expected leader.
-                ensure!(
-                    subdag.leader_address() == expected_leader,
-                    "Quorum block {expected_height} is authored by an unexpected leader (found: {}, expected: {expected_leader})",
-                    subdag.leader_address()
-                );
+                // ensure!(
+                //     subdag.leader_address() == expected_leader,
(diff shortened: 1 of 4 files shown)


ledger/block/src/verify.rs
```diff
@@ -193,13 +193,13 @@ impl<N: Network> Block<N> {
             }
             Authority::Quorum(subdag) => {
                 // Compute the expected leader.
-                let expected_leader = current_committee.get_leader(expected_round)?;
+                // let expected_leader = current_committee.get_leader(expected_round)?;
```
raychu86, line 196: Why are we skipping this check?

ledger/src/lib.rs
```diff
@@ -215,6 +215,14 @@ impl<N: Network, C: ConsensusStorage<N>> Ledger<N, C> {
         }
     }
 
+    /// Returns the latest committee with the committee round lag. 
+    pub fn latest_committee_with_lag(&self) -> Result<Committee<N>> {
+        match self.current_committee.read().as_ref() {
```
raychu86, line 220: Is this correct? Why are we using `current_committee.read()` here?

ledger/store/src/program/committee.rs
```diff
@@ -287,6 +288,25 @@ pub trait CommitteeStorage<N: Network>: 'static + Clone + Send + Sync {
             None => Ok(None),
         }
     }
+
+    /// Returns the committee for the given `round` with the committee round lag. 
+    fn get_committee_for_round_with_lag(&self, round: u64) -> Result<Option<Committee<N>>> {
+        // If the round is less than the committee lag, return the genesis committee. 
+        if round < COMMITTEE_ROUND_LAG{
```
raychu86, line 295: This check is unnecessary, as the `saturating_sub` already covers this case.

raychu86:
Closing for now in favor of #2336

PR #2336, opened 2024-02-07
[ZKS-02] Introduce committee lookback

raychu86:
## Motivation

This PR introduces a committee round lag of 100 rounds to match the `MAX_GC_ROUNDS`. This committee round lag is used in consensus to ensure that the committee being used to validate the subdags is the committee at `round.saturating_sub(N::COMMITTEE_ROUND_LAG)`.

Audit Finding: **[zksecurity 02] Dynamic Committee Feature is Not Safe**

### DIFF
--- a/ledger/src/check_next_block.rs
+++ b/ledger/src/check_next_block.rs
@@ -83,10 +83,14 @@ impl<N: Network, C: ConsensusStorage<N>> Ledger<N, C> {
             self.vm.check_speculate(state, block.ratifications(), block.solutions(), block.transactions())?;
 
         // Ensure the block is correct.
+        let round_with_lag = self.latest_round().saturating_sub(N::COMMITTEE_ROUND_LAG);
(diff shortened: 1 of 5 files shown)


ledger/src/check_next_block.rs
```diff
@@ -82,11 +82,24 @@ impl<N: Network, C: ConsensusStorage<N>> Ledger<N, C> {
         let ratified_finalize_operations =
             self.vm.check_speculate(state, block.ratifications(), block.solutions(), block.transactions())?;
 
+        // Get the round number for the previous committee. Note, we subtract 2 from odd rounds,
+        // because committees are updated in even rounds.
+        let previous_round = match block.round() % 2 == 0 {
```
raychu86, line 87: Technically, we can just subtract 1, because block rounds are always even.

howardwu (approved):

PR #2338, opened 2024-02-08
Introduce field-related benches

ljedrz:
Since these are some of the most fundamental operations popping up in snarkOS profiles, I decided to take a closer look at their implementations and introduce some benchmarks for them. In addition, I aligned a few of the implementations (via minor refactoring and reordering), so that it's easier to compare the `fp_256` and `fp_384` files.

Some notes:
- the `Fp384::*_assign` operations seem to be faster than what one might expect, compared to the `Fp256` counterparts
- on the other hand, the other benchmarked operations are slower (by a factor of >2x instead of 1.3x) than I expected
- at least on `x86_64`, some of the unrolled loops can be made into loops without any performance penalty (I checked it, as these days cache friendliness may be more beneficial); I'm happy to file such changes as a follow-up if it's desired (it does make the code significantly shorter and more readable)

### DIFF
--- a/fields/Cargo.toml
+++ b/fields/Cargo.toml
@@ -23,6 +23,11 @@ include = [ "Cargo.toml", "src", "README.md", "LICENSE.md" ]
 license = "Apache-2.0"
 edition = "2021"
 
+[[bench]]
+name = "fields"
+path = "benches/fields.rs"
+harness = false
+
 [dependencies.snarkvm-utilities]
 path = "../utilities"
 version = "=0.16.18"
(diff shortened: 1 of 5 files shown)


fields/src/fp_384.rs
```diff
@@ -433,7 +433,6 @@ impl<P: Fp384Parameters> PrimeField for Fp384<P> {
         let mut tmp = self.0;
         let mut r = tmp.0;
         // Montgomery Reduction
-        // Iteration 0
```
howardwu, line 436: Revert

howardwu:
@ljedrz for now, let's try to avoid PRs like this where it contains non-essential changes. No performance was improved here, and it is taking our time as distraction.

howardwu (approved):

fields/Cargo.toml
```diff
@@ -60,6 +65,14 @@ version = "1.0"
...
 features = [ "derive" ]
 
+[dev-dependencies.criterion]
+version = "0.5"
+
+[dev-dependencies.snarkvm-curves]
+path = "../curves"
+version = "=0.16.18"
+default-features = false
+
```
howardwu, line 75: Remove

howardwu:
^Please change this, and please for the love of god, do not rebase your work.

PR #2352, opened 2024-02-12
[HackerOne-2354265] Restrict number of imports, program depth, and call depth.

d0cd:
This PR,
- introduces a limit for the number of imports in a program.
- checks that each function in a program makes at most `MAX_TRANSITIONS - 1` calls.
- introduces an upper bound in import depth
- caches `get_number_of_function_calls`.


synthesizer/program/src/traits/stack_and_registers.rs
```diff
@@ -62,6 +62,9 @@ pub trait StackProgram<N: Network> {
     /// Returns the program.
     fn program(&self) -> &Program<N>;
 
+    /// Returns the program depth.
+    fn program_depth(&self) -> usize;
```
howardwu: nit: can you move this below program ID?

synthesizer/process/src/stack/mod.rs
```diff
@@ -220,6 +224,12 @@ impl<N: Network> StackProgram<N> for Stack<N> {
         &self.program
     }
 
+    /// Returns the program depth.
+    #[inline]
+    fn program_depth(&self) -> usize {
```
howardwu: nit: can you move this below `program_id`?

console/network/src/lib.rs
```diff
@@ -160,6 +160,11 @@ pub trait Network:
     /// The maximum number of outputs per transition.
     const MAX_OUTPUTS: usize = 16;
 
+    /// The maximum program depth.
+    const MAX_PROGRAM_DEPTH: usize = 1024;
+    /// The maximum number of imports.
+    const MAX_IMPORTS: usize = 2048;
```
howardwu: Can you provide some context on why these depths and imports are where they're at?

Based on our design, I believe we're constraining other parameters already at the 32 mark. It would make sense to constraint both of these parameters alongside the 32 mark.

synthesizer/process/src/stack/helpers/initialize.rs
```diff
@@ -39,17 +41,49 @@ impl<N: Network> Stack<N> {
...
+                }
+            }
+            // Check that the number of calls does not exceed the maximum.
+            // Note that one transition is reserved for the fee.
+            ensure!(
+                num_calls < ledger_block::Transaction::<N>::MAX_TRANSITIONS,
+                "Number of calls exceeds the maximum allowed number of transitions"
+            );
+            // Add the number of calls to the stack.
+            stack.number_of_calls.insert(*function.name(), num_calls);
```
howardwu, line 84: What is the performance implication of this additional logic?

console/network/src/lib.rs
d0cd: The `MAX_PROGRAM_DEPTH` was decided based off of Ethereum's MAX_CALL_DEPTH.
I went with this as a healthy upper bound. I see no issue with restricting `MAX_PROGRAM_DEPTH` to 32.

`MAX_IMPORTS` should be larger because a user may want to create a program that reads the mappings of a large number of other programs.

synthesizer/process/src/stack/helpers/initialize.rs
d0cd: Analytically, it reduces O(n) work for `get_number_of_calls` to O(1) in exchange for space.
I will follow up with benchmarks.

console/network/src/lib.rs
d0cd: Ah actually, suppose that `program1` reads `program2` reads ... reads `program32`
If we `program0` to reads `program1`, then we need a greater `MAX_PROGRAM_DEPTH`.

synthesizer/process/src/stack/helpers/initialize.rs
d0cd: On `fix/import-depth`:
- `Stack::new` was `~27 us` for all depths greater than zero.
- `Stack::get_number_of_calls` was `~900 ns` for all depths.

On `benches/stack-operations` (a branch with the same benchmark file on top of `mainnet`):
- `Stack::new` was `~27 us` for all depths greater than zero.
- `Stack::get_number_of_calls` increased linearly for each depth. See below.

```
Depth 0 | Stack::get_number_of_calls
                        time:   [3.9844 µs 3.9941 µs 4.0081 µs]

Depth 1 | Stack::get_number_of_calls
                        time:   [1.4909 µs 1.4946 µs 1.5009 µs]

Depth 2 | Stack::get_number_of_calls
                        time:   [1.9188 µs 1.9222 µs 1.9276 µs]

Depth 4 | Stack::get_number_of_calls
                        time:   [2.7937 µs 2.7991 µs 2.8100 µs]

Depth 8 | Stack::get_number_of_calls
                        time:   [4.4999 µs 4.5912 µs 4.7143 µs]

Depth 16 | Stack::get_number_of_calls
                        time:   [8.0029 µs 8.0240 µs 8.0565 µs]

Depth 30 | Stack::get_number_of_calls
                        time:   [14.118 µs 14.143 µs 14.169 µs]
```

console/network/src/lib.rs
howardwu: Discussed offline, waiting for a change to 64 and 64

randomsleep:
One question less related this issue: Would it be better to set a larger 'TRANSACTION_DEPTH', (e.g. 8)? This will make it easier to increase 'MAX_TRANSITION' in the future. We can still keep 'MAX_TRANSITION' as 32 for now.

howardwu:
> One question less related this issue: Would it be better to set a larger 'TRANSACTION_DEPTH', (e.g. 8)? This will make it easier to increase 'MAX_TRANSITION' in the future. We can still keep 'MAX_TRANSITION' as 32 for now.

This has problems because it will change the Merklization to be much slower, and require a larger inclusion circuit for proving. Ultimately, these impact runtime for transaction generation.

howardwu (approved):

PR #2357, opened 2024-02-15
[Feature] Adds a histogram label

joske:
## Motivation

This adds a `histogram_label` method to allow passing a label to the histogram. A limitation is that you can only attach one label.

### DIFF
--- a/metrics/src/lib.rs
+++ b/metrics/src/lib.rs
@@ -98,3 +98,7 @@ pub fn histogram<V: Into<f64>>(name: &'static str, value: V) {
     let histogram = ::metrics::histogram!(name);
     histogram.record(value.into());
 }
+
+pub fn histogram_label<V: Into<f64>>(name: &'static str, label_key: &'static str, label_value: String, value: V) {
+    ::metrics::histogram!(name, label_key => label_value).record(value.into());
+}


vicsn:
@collinc97 can you adjust the settings so we can add @miazn as reviewer?

miazn:
We can actually remove`register_histogram` function, as the histogram! macro (just like counter or gauge) registers the metric the first time it sees a new metric name (you can see this in lines 93 and 98)-- since we are using the base metrics definitions of the counter,gauge and hist.

metrics/src/lib.rs
```diff
@@ -98,3 +98,17 @@ pub fn histogram<V: Into<f64>>(name: &'static str, value: V) {
     let histogram = ::metrics::histogram!(name);
     histogram.record(value.into());
 }
+
+pub fn histogram_label<V: Into<f64>>(name: &'static str, label_key: &'static str, label_value: String, value: V) {
+    let labels = vec![::metrics::Label::new(label_key, label_value)];
```
miazn: if we are only using a string label key and value, it might be easier to just use the macro, as i did here:
https://github.com/AleoHQ/snarkOS/pull/3064/files#diff-8663d09365b76641c32c45870cbd1915ab198258eb6b823b0aa779128a5b4ff1R93

this makes it cleaner to add the label function to the other two types (counter and gauge), which we should probably do?

joske: Indeed, now I switched to owned string for the label value, it's possible to use the macro. This was giving trouble with `'static` when it was still a `&str`.

joske:
I kept the register functions for backwards compatibility. They were already useless when we bumped to `metrics 0.22`.

metrics/src/lib.rs
joske: I could add label variants to the gauge and counter too, but there's no need at the moment?

howardwu (approved):

PR #2362, opened 2024-02-19
Add impl Ternary for Box<T: Ternary>

kpp:
## Motivation

Reduce the amount of boilerplate code


ljedrz (approved):
LGTM :+1:; there is likely more potential for improvements like this one.

PR #2365, opened 2024-02-20
Modify scope of the hash module in the coinbase module to pub use

cx19981114:
In the current code, the hash module is imported privately within the coinbase module. This prevents external modules from directly accessing and utilizing the functionalities of the hash module. This could be inconvenient for developers. I propose modifying the scope of the hash module to be public (pub use).


howardwu:
Closing as this PR is no longer relevant in light of #2401

PR #2373, opened 2024-02-25
feat: check transactions in atomic_speculate

randomsleep:
## Motivation

This PR adds checks for transactions in `atomic_speculate`. This will prevent invalid transactions from halting the network.

Previously we perform `check_transactions` for every transaction in `check_next_block`, after `construct_block_template`. This is too late and will halt the network in some cases:

1. If a transaction is an invalid format, `check_next_block` will reject: https://github.com/AleoHQ/snarkOS/issues/2991
2. Transactions with dependencies may pass `atomic_speculate` but get rejected in `check_next_block`(this not shown in previous reports). For example, consider two transactions `T1` sent by address `A1` and `T2` sent by address `A2`. The initial credit of `A2` is 0 so `T2` is an invalid transaction (not enough fee). In `T1`, address `A1` sends some credit to `A2`. Therefore, `T1` and `T2` together will pass in `atomic_speculate`. However, `T2` will be rejected in `check_next_block` because the check is independent. This will halt the network.

The reason why we should not add `check_transactions` for `BatchPropose` is that it will cause liveness issues: https://github.com/AleoHQ/snarkOS/pull/3098#issuecomment-1952321580

## Test Plan

(Write your test plan here)

## Related PRs

This should fix the issues about invalid transactions halt the network (including https://github.com/AleoHQ/snarkOS/issues/2991).

Related PR: https://github.com/AleoHQ/snarkOS/pull/3083 https://github.com/AleoHQ/snarkOS/pull/3098

### DIFF
--- a/ledger/src/check_next_block.rs
+++ b/ledger/src/check_next_block.rs
@@ -14,11 +14,9 @@
 
 use super::*;
 
-use rand::{rngs::StdRng, SeedableRng};
-
 impl<N: Network, C: ConsensusStorage<N>> Ledger<N, C> {
     /// Checks the given block is valid next block.
-    pub fn check_next_block<R: CryptoRng + Rng>(&self, block: &Block<N>, rng: &mut R) -> Result<()> {
+    pub fn check_next_block<R: CryptoRng + Rng>(&self, block: &Block<N>, _rng: &mut R) -> Result<()> {
         let height = block.height();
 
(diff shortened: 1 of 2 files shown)


raychu86:
Thank you for filing this PR, I am closing in favor of a more comprehensive fix of the same approach.

https://github.com/AleoHQ/snarkVM/pull/2376

PR #2374, opened 2024-02-26
[ZKS-02] Introduce Committee IDs 

raychu86:
## Motivation

This PR introduces the concepts of committee IDs, which is a hash of the committee state: `BHP(starting_round || members || total_stake)`. 

This committee ID is used in batch headers to refer to a specific committee that was used to create the batch header. Other validators will then only sign batch headers/proposals if they see that the committee ID used refers to the committee that was expected. This will be an added safety guarantee when building the sudag/blocks, because validators will only certify batches if they are on the same committee state.

Audit Finding: **[zksecurity 02] Dynamic Committee Feature is Not Safe**


acoglio (approved):

howardwu (approved):

ledger/block/src/verify.rs
```diff
@@ -233,6 +233,29 @@ impl<N: Network> Block<N> {
...
+
+            // Check that all all certificates on each round have the same committee ID.
+            cfg_iter!(subdag).try_for_each(|(round, certificates)| {
+                // Check that every certificate for a given round shares the same committee ID.
+                let expected_committee_id = certificates
+                    .first()
+                    .map(|certificate| certificate.committee_id())
+                    .ok_or(anyhow!("No certificates found for subdag round {round}"))?;
+                ensure!(
+                    certificates.iter().skip(1).all(|certificate| certificate.committee_id() == expected_committee_id),
```
mimoo, line 252: OK so I think that makes sense to me, most of the DAG will likely be the same committee id, but any vertex that's on the same round (or previous round) as the last committed anchor will have a different committee id

mimoo (commented):
LGTM

ledger/block/src/verify.rs
howardwu: I think the intent is to reject those out of date cases during the proposal phase.

Do you see any concern that an out of date vertex could get 2f+1 signatures, be included in block production here, and cause a halting concern?

PR #2376, opened 2024-03-01
[HackerOne-2311934] Verify transactions prior to speculation

### LINKED ISSUE #2991: [Bug] BatchPropose with invalid transaction can halt the network
# https://hackerone.com/reports/2311934

## Summary:

 BatchPropose with invalid transaction halt the network

## Steps To Reproduce:

    1. Use this branch: https://github.com/ghostant-1017/snarkOS/tree/demo/invalid_tx
    2. Run the devnet script: `.devnet` 
    3. Check the logs: ` Unable to advance to the next block - Invalid transaction found in the transactions list: Execution verification failed: Failed to verify a transition input`

## Proof-of-Concept (PoC)

The `BatchPropose` with invalid transactions will be certified and the `check_next_block` will fail as the transaction is invalid

## Supporting Material/References:

Logs: https://github.com/ghostant-1017/logs/blob/master/logs-20240111094535.tar.gz

## Summary:
BatchPropose with invalid transaction halt the network

raychu86:
## Motivation

This PR performs transaction verification in `VM::speculate` and aborts the transactions that are invalid. This is necessary because in snarkOS when we construct subdags, we do not validate transmissions found in proposals, which could lead to block production halting if there are malicious validators. With the new change, we can eliminate the invalid transactions prior to block production and ensure that `Block::check_next_block` doesn't error.

Note that now we perform transaction verification prior to creating the block itself, rather than just before the block is about to be added to storage. Because of the transaction partial verification cache, we do not perform duplicate verification on the execution/deployment, but we do have to perform the (cheaper) uniqueness checks twice.

## Related PRs

This PR is related to https://github.com/AleoHQ/snarkOS/pull/3104.

And should replace: https://github.com/AleoHQ/snarkOS/pull/3083 and https://github.com/AleoHQ/snarkOS/pull/3098.

This should fix: https://github.com/AleoHQ/snarkOS/issues/2991 and https://github.com/AleoHQ/snarkOS/issues/2990.


vicsn:
FYI when trying to compile this into snarkOS for testing I get the error:
```
error[E0038]: the trait `snarkos_node_bft_ledger_service::LedgerService` cannot be made into an object
   --> node/sync/src/block_sync.rs:118:49
    |
118 |     pub fn new(mode: BlockSyncMode, ledger: Arc<dyn LedgerService<N>>) -> Self {
    |                                                 ^^^^^^^^^^^^^^^^^^^^ `snarkos_node_bft_ledger_service::LedgerService` cannot be made into an object
    |
note: for a trait to be "object safe" it needs to allow building a vtable to allow the call to be resolvable dynamically; for more information visit <https://doc.rust-lang.org/reference/items/traits.html#object-safety>
   --> /Users/victorsintnicolaas/programs/snarkOS/node/bft/ledger-service/src/traits.rs:107:8
    |
107 |     fn prepare_advance_to_next_quorum_block<R: Rng + CryptoRng>(
    |        ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^ the trait cannot be made into an object because method `prepare_advance_to_next_quorum_block` has generic type parameters
```

synthesizer/src/vm/finalize.rs
```diff
@@ -28,31 +30,75 @@ impl<N: Network, C: ConsensusStorage<N>> VM<N, C> {
...
+                checked_transactions.into_iter().fold(
+                    (Vec::new(), Vec::new()),
+                    |(mut verified, mut aborted), (transaction, error_message)| {
+                        match error_message {
+                            None => verified.push(transaction),
+                            Some(e) => aborted.push((transaction, e)),
+                        };
+                        (verified, aborted)
+                    },
+                )
```
ljedrz: seems like https://doc.rust-lang.org/std/iter/trait.Iterator.html#method.partition would be a good fit here

ljedrz:
> FYI when trying to compile this into snarkOS for testing I get the error:

As for the object safety issue, it's not ideal, but we could work around it by having 2 versions of `prepare_advance_to_next_quorum_block`, with `TestRng` hardcoded for `#[cfg(test)]` scenarios :thinking:.

synthesizer/src/vm/finalize.rs
```diff
@@ -28,31 +30,75 @@ impl<N: Network, C: ConsensusStorage<N>> VM<N, C> {
...
+        let (verified_transactions, verification_aborted_transactions) = match block_store
+            .find_block_height_from_state_root(block_store.current_state_root())?
+        {
+            // If the current state root does not exist in the block store, then the genesis block has not been introduced yet.
+            None => (candidate_transactions, vec![]),
+            // Verify transactions for all non-genesis cases.
+            _ => {
+                let rngs = (0..candidate_transactions.len()).map(|_| StdRng::from_seed(rng.gen())).collect::<Vec<_>>();
+                // Verify the transactions and collect the error message if there is one.
+                let checked_transactions: Vec<_> = cfg_into_iter!(candidate_transactions)
```
vicsn: This comment still applies: https://github.com/AleoHQ/snarkOS/pull/3083#issuecomment-1936937113
I was able to kill all the nodes with some big batch proposals in a test suite.

With the system allocator (which is heavier than jemalloc):
- The maximum RAM to verify a 1M constraint program is 2GB.
- The maximum RAM for a 1 or 31-depth 1M constraint call execution is *negligible*,  but to be safe we could assume 10MB

A thorough but brittle approach would be to estimate memory usage based on the number of constraints. But an easier setup would be a separate deploy and an execute verification maximum:

```
const MAX_PARALLEL_DEPLOY_VERIFICATIONS = 5;
const MAX_PARALLEL_EXECUTE_VERIFICATIONS = 1000;

let (deploys, executions) = candidate_transactions.partition(|tx|tx.is_deploy());
let deploy_iter = deploys.chunks(MAX_PARALLEL_DEPLOY_VERIFICATIONS);
let execution_iter = executions.chunks(MAX_PARALLEL_EXECUTE_VERIFICATIONS);
let tx_iter = deploy_iter.chain(execution_iter);

while let Some(txs) = tx_iter.next() {
    cfg_into_iter!(txs.zip(rngs).map(|| ...verify...)
}
```

raychu86: How is this different from the current approach of transaction verification on the current `mainnet` branch in `check_next_block`?

Are you saying that fundamentally the verification logic should be updated?

synthesizer/src/vm/finalize.rs
raychu86: Good point, much cleaner.

raychu86:
> FYI when trying to compile this into snarkOS for testing I get the error:

@vicsn Similar to how we use a `thread_rng` for `check_next_block` and `check_transaction_basic`, this should allow testing to compile:

https://github.com/AleoHQ/snarkOS/compare/mainnet...abort-malicious-txes

synthesizer/src/vm/finalize.rs
vicsn: Yes, that piece of parallel verification logic was prone to the same vulnerability.

vicsn:
> @vicsn Similar to how we use a thread_rng for check_next_block and check_transaction_basic, this should allow testing to compile:

Thank you, will test it and report back.

ghostant-1017:
Hi guys,
I was notified that [HackerOne-2310746/2311934] were duplicated. As I understand it and Report-2310746 focuses only on the Fee and is easier to fix, while Report-2311934 is more comprehensive. Will you reconsider the results of these two report assessments? Thank you! @vicsn @raychu86

raychu86:
> Hi guys, I was notified that [HackerOne-2310746/2311934] were duplicated. As I understand it and Report-2310746 focuses only on the Fee and is easier to fix, while Report-2311934 is more comprehensive. Will you reconsider the results of these two report assessments? Thank you! @vicsn @raychu86

Although both of these issues can be resolved with a single fix, in my opinion, they could be considered separate reports. 

The snarkOS PR https://github.com/AleoHQ/snarkOS/pull/3104, should address the Fee finding, and this PR should address both of the findings (for added safety).

ghostant-1017:
Thank you! @raychu86

synthesizer/src/vm/finalize.rs
raychu86: I've implemented the OOM fix here - https://github.com/AleoHQ/snarkVM/pull/2387. If it is approved, we can apply the change to this PR as well.

howardwu (approved):

PR #2378, opened 2024-03-03
[ZKS-08][HackerOne-2289066] Make leader election tamper resistant

raychu86:
## Motivation

This PR updates the leader election algorithm to be tamper resistant by removing mutable fields (`starting_round` and `total_stake`) from the seed and changing the committee sorting to be based on the address x-coordinate.

Leader election will now be more predictable (if the committee sets do not change), but should no longer be able to be manipulated by changing stake via `bond/unbond`s. 

Discussion can be found here: https://github.com/AleoHQ/snarkOS/issues/2986#issuecomment-1955791181

## Test Plan

The current tests indicate that this approach is still stake weighted properly. 

Audit Finding: **[zksecurity 08] Potentially Biased Leader Election**

### DIFF
--- a/ledger/committee/src/lib.rs
+++ b/ledger/committee/src/lib.rs
@@ -181,7 +181,7 @@ impl<N: Network> Committee<N> {
         // Retrieve the total stake of the committee.
         let total_stake = self.total_stake();
         // Construct the round seed.
-        let seed = [self.starting_round, current_round, total_stake].map(Field::from_u64);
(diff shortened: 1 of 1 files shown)


raychu86:
@randomsleep This should address the leader manipulation concerns you mentioned. Let us know if you do not believe this is the case.

randomsleep (approved):
LGTM

howardwu (approved):

PR #2379, opened 2024-03-05
[Test] Execution cost.

d0cd:
This PR follows on #2369 with two tests checking that the finalize cost is calculated correctly. 
The finalize cost of an execution is the finalize cost of its root transition.
This should be equal to the sum of the cost of **all** commands in **each** finalize block in **each** transition in the execution.
As a by product, this PR also refactors the cost utilities to support `command_cost` which returns the cost of a command.

### DIFF
--- a/synthesizer/process/src/cost.rs
+++ b/synthesizer/process/src/cost.rs
@@ -21,22 +21,141 @@ use console::{
 use ledger_block::{Deployment, Execution};
 use synthesizer_program::{CastType, Command, Finalize, Instruction, Operand, StackProgram};
 
-/// The base and per byte costs for expensive operations.
+/// Returns the *minimum* cost in microcredits to publish the given deployment (total cost, (storage cost, synthesis cost, namespace cost)).
+pub fn deployment_cost<N: Network>(deployment: &Deployment<N>) -> Result<(u64, (u64, u64, u64))> {
+    // Determine the number of bytes in the deployment.
+    let size_in_bytes = deployment.size_in_bytes()?;
+    // Retrieve the program ID.
+    let program_id = deployment.program_id();
+    // Determine the number of characters in the program ID.
+    let num_characters = u32::try_from(program_id.name().to_string().len())?;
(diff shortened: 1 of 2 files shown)


synthesizer/process/src/cost.rs
```diff
@@ -21,150 +21,23 @@ use console::{
...
+const HASH_BASE_COST: u64 = 10_000;
+const HASH_PER_BYTE_COST: u64 = 30;
+const HASH_BHP_BASE_COST: u64 = 50_000;
+const HASH_BHP_PER_BYTE_COST: u64 = 300;
+const HASH_PSD_BASE_COST: u64 = 40_000;
+const HASH_PSD_PER_BYTE_COST: u64 = 75;
+const MAPPING_BASE_COST: u64 = 10_000;
+const MAPPING_PER_BYTE_COST: u64 = 10;
+const SET_BASE_COST: u64 = 10_000;
+const SET_PER_BYTE_COST: u64 = 100;
```
howardwu: Why did this get moved?

howardwu:
Why does this PR touch non-test code?

synthesizer/process/src/cost.rs
```diff
@@ -74,97 +74,89 @@ pub fn execution_cost<N: Network>(process: &Process<N>, execution: &Execution<N>
...
-                "Overflowed while computing the size of the operand '{operand}' in '{}/{}' (finalize)",
-                stack.program_id(),
-                finalize.name()
-            ))
-        })?;
-        // Return the cost.
-        Ok(base_cost.saturating_add(byte_multiplier.saturating_mul(size_of_operands)))
-    }
+/// Finalize costs for compute heavy operations, derived as:
+/// `BASE_COST + (PER_BYTE_COST * SIZE_IN_BYTES)`.
```
howardwu, line 78:
```suggestion
/// `BASE_COST + (PER_BYTE_COST * SIZE_IN_BYTES)`.

```

howardwu (approved):

PR #2384, opened 2024-03-05
[Fix] Limit program size to 100KBs

### LINKED ISSUE #3141: Add limits on program struct definitions
There was a dummy program written which sized up to 50MBs by leveraging the struct construction.

Following this discussion, the decision was made to reduce the number of structs that can be defined in a program, and to bound any other constructs of a program so that the size of a program is inherently restricted.

@vicsn and @d0cd will be responsible with implementing a PR to contain this.

_Originally posted by @howardwu in https://github.com/AleoHQ/snarkOS/issues/3124#issuecomment-1974921008_

vicsn:
## Motivation

Closes https://github.com/AleoHQ/snarkOS/issues/3141

Introduces:
- `MAX_PROGRAM_SIZE`
- `MAX_STRUCTS`
- `MAX_RECORDS`
- `MAX_CLOSURES`

Some notes:
- `MAX_STRUCTS`, `MAX_CLOSURES` and `MAX_RECORDS` don't add security but don't hurt either and were requested in the issue above. Even existing parameters like `MAX_INSTRUCTIONS` can't be fully utilized as it will now quickly lead to programs which surpass the size limit.
- I did not add custom error types to test on to keep it simple and because we don't use those anywhere yet.

Can't add you as reviewer do doing it manually: @miazn

## Test Plan

Added extensive unit test

### DIFF
--- a/console/network/src/lib.rs
+++ b/console/network/src/lib.rs
@@ -144,10 +144,21 @@ pub trait Network:
     /// The maximum number of entries in a record.
     const MAX_RECORD_ENTRIES: usize = Self::MIN_RECORD_ENTRIES.saturating_add(Self::MAX_DATA_ENTRIES);
 
+    /// The maximum program size in bytes.
+    const MAX_PROGRAM_BYTE_SIZE: usize = 3_500_000; // 3.5 MB
+    /// The maximum program size in characters.
+    const MAX_PROGRAM_CHAR_SIZE: usize = 7_000_000; // 7 MB
+
     /// The maximum number of mappings in a program.
     const MAX_MAPPINGS: usize = 31;
     /// The maximum number of functions in a program.
     const MAX_FUNCTIONS: usize = 31;
+    /// The maximum number of structs in a program.
+    const MAX_STRUCTS: usize = 10 * Self::MAX_FUNCTIONS;
+    /// The maximum number of records in a program.
+    const MAX_RECORDS: usize = 10 * Self::MAX_FUNCTIONS;
+    /// The maximum number of closures in a program.
+    const MAX_CLOSURES: usize = 2 * Self::MAX_FUNCTIONS;
     /// The maximum number of operands in an instruction.
     const MAX_OPERANDS: usize = Self::MAX_INPUTS;
     /// The maximum number of instructions in a closure or function.
--- a/console/network/src/lib.rs
+++ b/console/network/src/lib.rs
@@ -145,7 +145,7 @@ pub trait Network:
     const MAX_RECORD_ENTRIES: usize = Self::MIN_RECORD_ENTRIES.saturating_add(Self::MAX_DATA_ENTRIES);
 
     /// The maximum program size by number of characters.
-    const MAX_PROGRAM_SIZE: usize = 7_000_000; // 7 MB
+    const MAX_PROGRAM_SIZE: usize = 100_000; // 100 KB
 
     /// The maximum number of mappings in a program.
     const MAX_MAPPINGS: usize = 31;
--- a/console/network/src/lib.rs
+++ b/console/network/src/lib.rs
@@ -144,10 +144,8 @@ pub trait Network:
     /// The maximum number of entries in a record.
     const MAX_RECORD_ENTRIES: usize = Self::MIN_RECORD_ENTRIES.saturating_add(Self::MAX_DATA_ENTRIES);
 
-    /// The maximum program size in bytes.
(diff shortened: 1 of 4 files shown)


synthesizer/program/src/parse.rs
```diff
@@ -106,10 +106,16 @@ impl<N: Network, Instruction: InstructionTrait<N>, Command: CommandTrait<N>> Fro
 
     /// Returns a program from a string literal.
     fn from_str(string: &str) -> Result<Self> {
+        // Ensure the raw program string is less than MAX_PROGRAM_CHAR_SIZE.
+        ensure!(string.len() <= N::MAX_PROGRAM_CHAR_SIZE, "Program exceeds N::MAX_PROGRAM_CHAR_SIZE.");
```
howardwu: `string.len()` is equivalent to `string.to_bytes_le().len()`

Why are we checking two different values at 3.5MB and 7MB when it's the same value?

howardwu (changes requested):
We have `N::MAX_IMPORTS`. Can we also enforce that size here?

console/network/src/lib.rs
```diff
@@ -144,10 +144,21 @@ pub trait Network:
     /// The maximum number of entries in a record.
     const MAX_RECORD_ENTRIES: usize = Self::MIN_RECORD_ENTRIES.saturating_add(Self::MAX_DATA_ENTRIES);
 
+    /// The maximum program size in bytes.
+    const MAX_PROGRAM_BYTE_SIZE: usize = 3_500_000; // 3.5 MB
+    /// The maximum program size in characters.
+    const MAX_PROGRAM_CHAR_SIZE: usize = 7_000_000; // 7 MB
```
howardwu: Why is this 3.5MB and 7MB?

From initial eyeballing, 1MB may seem excessive. Consider the number of deployments in a block and the implication on block size. It would be excessive for a block to be 7 * 100 == 700MB if an attacker tried to DOS storage. It would also prevent the block from being transmitted in the network protocol.

We may want to cap the expressivity at between 64KB-512KB from an initial glance. It is a worthwhile effort to formalize what this number should be empirically.

synthesizer/program/src/parse.rs
vicsn: They are not equivalent, the original input string is usually longer, the byte vector is a more succinct representation. If you want a quick example print the values in this test: https://github.com/AleoHQ/snarkVM/blob/limit_program_size/synthesizer/program/src/bytes.rs#L164

However, I'm happy to re-use MAX_PROGRAM_BYTE_SIZE here. Let me know what you prefer.

console/network/src/lib.rs
vicsn: @miazn is in the process of getting empirical data on what was actually deployed on testnet3. Afterwards I'll consider deployment costs and report back here.

synthesizer/program/src/parse.rs
howardwu: I see, you're comparing `program.to_bytes_le().len()` instead of `string.to_bytes_le().len()`.

The `program.to_bytes_le()` does not grow linearly at a factor of 2x consistently for all program types. It should be sublinear (by a wide margin) from the string representation.

howardwu: I may recommend checking the string length for better intuition. It's easy for developers to enforce a program size elsewhere if they have a uniform metric to measure against.

vicsn: I like: 9e0f25160

console/network/src/lib.rs
vicsn: > It would be excessive for a block to be 7 * 100 == 700MB if an attacker tried to DOS storage.

Just to be on the same page, I think this is unlikely. 7MB deployments cost at least 7000 credits due to  `DEPLOYMENT_FEE_MULTIPLIER`, so we should not be seeing that many large deployments in a block already.

However, as previously discussed, lets bias towards safety for now. The string sizes of the top 20 deployed programs on testnet are at most 52KB. So I suggest to set the program limit to 100KB: 84a9a4518

howardwu:
> We have `N::MAX_IMPORTS`. Can we also enforce that size here?

This issue is still unaddressed @vicsn

vicsn:
> `MAX_IMPORTS`

The limit was already enforced. Test was added here: [81e51a8](https://github.com/AleoHQ/snarkVM/pull/2384/commits/81e51a8b0c3720173824474505adcf4f05180e6d)

Let me know if you have something else in mind.

howardwu:
I missed that commit, thanks @vicsn

howardwu (approved):

PR #2385, opened 2024-03-08
[Go/No-Go] ARC-0037 - Introduce withdrawal address for stakers

raychu86:
## Motivation

This PR implements withdrawal addresses that staker can use claim their unbonded credits in accordance to [ARC-0037](https://github.com/AleoHQ/ARCs/discussions/51).

Validators and delegators now need to specify an additional `withdrawal_address` when they call `bond_public`. While they are bonded this withdrawal address can not be changed, which is important because a core reason for this separation of addresses is so that the validator funds are secure even if their "hot" keys are compromised. This withdraw address will receive the credits that are unbonded by the staker in `claim_unbond_public`. 

**This is done with the following changes:**

- Added a new `withdraw` mapping
   - Key: staker address
   - Value: withdrawal address
   
- Updated `bond_public`:
  - Add additional withdrawal address as input
  - Check if `withdrawal_address` exists in the `withdraw` mapping:
	 - true -> Check that the withdrawal address is consistent with the one in the mapping (if it already exists)
	 - false -> Update the withdraw mapping.

- Updated `claim_unbond_public`: 
  - This is still called by the staker.
  - Withdraw the balance to the `withdrawal_address`
  - Remove the withdraw address if the bonded entry no longer exists for the staker

- Updated`BondedBalances`: 
  - Add a withdrawal address field
  - Update `atomic_pre_ratify` to process the withdrawal addresses
  - Allows specifying withdrawal addresses in the genesis block

## Test Plan

Tests have been added to ensure that:

- The withdrawal address receives the claimed unbonded credits.
- The withdrawal address can't be changed unless the staker is completely unbonded.
- Stakers can't call `bond_public` with a different withdrawal address that was was set in the `withdraw` mapping.
- The processing of `Ratify::Genesis` is correct with the new withdrawal addresses.

## Additional Notes

- A staker must be completely unbonded before they are able to change their withdrawal address
- Withdrawal addresses can't be changed once initial stake is bonded

## Related PRs

The sister PR can be found here - https://github.com/AleoHQ/snarkOS/pull/3165


synthesizer/src/vm/finalize.rs
```diff
@@ -2161,6 +2183,25 @@ finalize compute:
...
+                    Value::from_str(&withdrawal_address.to_string()).unwrap(),
+                )
+            })
+            .collect_vec();
+        // Note that `actual_withdraw` and `expected_withdraw` are vectors and not necessarily in the same order.
+        // By checking that the lengths of the vector are equal and that all entries in `actual_withdraw` are in `expected_withdraw`,
+        // we can ensure that the two vectors contain the same data.
+        assert_eq!(actual_withdraw.len(), expected_withdraw.len());
+        for entry in actual_withdraw.iter() {
+            assert!(expected_withdraw.contains(entry));
```
ljedrz, line 2268: a linear search could become slow if the `expected_withdraw` vec is large :thinking:; it doesn't need to be a hash map, but it could be sorted, so that a binary search could be utilized here instead

ljedrz: ah, but it's just a test; nvm then

synthesizer/program/src/resources/credits.aleo
```diff
@@ -216,82 +236,85 @@ finalize bond_public:
...
-    lte r25 100_000u32 into r26;
+    lte r27 100_000u32 into r28;
     // Enforce that the number of delegators is less than or equal to 100_000.
-    assert.eq r26 true;
+    assert.eq r28 true;
     // Set the new number of delegators.
-    set r25 into metadata[aleo1qgqqqqqqqqqqqqqqqqqqqqqqqqqqqqqqqqqqqqqqqqqqqqqqqqqqanmpl0];
+    set r27 into metadata[aleo1qgqqqqqqqqqqqqqqqqqqqqqqqqqqqqqqqqqqqqqqqqqqqqqqqqqqanmpl0];
+    // Set the withdrawal address.
+    set r2 into withdraw[r0];
```
vicsn, line 270: Just want to confirm this is correct, we're setting committee size and number of delegators just once, but `Set the withdrawal address` is noted twice.

ljedrz (commented):
LGTM so far :+1:.

synthesizer/program/src/resources/credits.aleo
```diff
@@ -627,17 +653,31 @@ finalize claim_unbond_public:
...
-    get.or_use account[r0] 0u64 into r3;
-    add r1.microcredits r3 into r4;
-    set r4 into account[r0];
+    // Get the withdrawal address for the caller.
+    get withdraw[r0] into r3;
+
+    // Add the unbonded amount to the withdrawal address public balance.
+    // Increments `account[r3]` by `r1`.
+    // If `account[r3]` does not exist, 0u64 is used.
+    // If `account[r3] + r2` overflows, `claim_unbond_public` is reverted.
```
vicsn: Should this note `account[r3] + r1` ?

vicsn (approved):
Just some nits

synthesizer/program/src/resources/credits.aleo
raychu86: There is 2 instances of `set _ into withdraw[_]` because it denotes the validator case and the delegator case.

This should only be called once during the initial bond of the validator or delegator. This is the same as the metadata mappings only being set during the initial bond.

synthesizer/program/src/resources/credits.aleo
raychu86: Yes, you are correct. Looks like the comment was previously erroneous as well.

d0cd (approved):

Tuneezi:
So what should we do here?

KEJANOW:
lest goo

jony0406:
api full fill work!

PR #2386, opened 2024-03-09
Update Staker requirements and `BatchHeader::MAX_CERTIFICATES`

raychu86:
## Motivation

This PR updates the minimum validator/delegator stakes and also the `BatchHeader::MAX_CERTIFICATES`:

`MIN_VALIDATOR_STAKE`: 1_000_000 credits -> 10_000_000 credits
`MIN_DELEGATOR_STAKE`: 10 credits -> 10_000 credits
`MAX_CERTIFICATES`: 200 -> 10

Note that `Committee::MAX_COMMITTEE_SIZE` and `BatchCertificate::MAX_SIGNATURES` will also be changed because they are equivalent to `BatchHeader::MAX_CERTIFICATES`.


raychu86:
@ljedrz @niklaslong @joske There seems to be an issue with how `Committee::MAX_COMMITTEE_SIZE` and `BatchCertificate::MAX_SIGNATURES` are being set in testing. They do not properly use the cfg(test) value of `BatchHeader::MAX_CERTIFICATES` when running the tests (they use 10 instead of 100). 

Would you happen to have a cleaner way to do this rather than adding the cfg flags to both `MAX_COMMITTEE_SIZE` and `MAX_SIGNATURES`?

ljedrz:
@raychu86 I usually prefer to introduce a `test` feature and then apply a `#[cfg(feature = "test")]` attribute where applicable; that way, as long as the crate is consistently imported as a `dev-dependency` with that feature enabled, it doesn't matter how the test is executed - the attribute applies.

edit: since I see that this is also in place, I would check if the crate in question is consistently imported as a `dev-depencency` with the `test` feature in the crates where the tests reside.

niklaslong (approved):

ljedrz (approved):

synthesizer/src/vm/helpers/committee.rs
```diff
@@ -249,7 +249,7 @@ pub(crate) mod test_helpers {
                 // Keep a tally of the remaining microcredits.
                 let remaining_microcredits = microcredits.saturating_sub(MIN_VALIDATOR_STAKE);
                 // Set the staker amount to 10 credit.
```
d0cd:
```suggestion
                // Set the staker amount to `MIN_DELEGATOR_STAKE` microcredits.
```

d0cd (approved):
LGTM! Left a minor nit.

PR #2387, opened 2024-03-12
Bound parallel verification of transactions

raychu86:
## Motivation

This PR bounds the memory usage during transactions verification by introducing `MAX_PARALLEL_DEPLOY_VERIFICATIONS` and `MAX_PARALLEL_EXECUTE_VERIFICATIONS`, where we verify a limited number of transactions at once instead of all at once. Previously the unbounded verification could lead to OOM errors when a node is forced to verify many large/expensive transactions at once.

Profiling memory usage of verification gives us some insights -

https://github.com/AleoHQ/snarkVM/pull/2376#discussion_r1509093306
> Deployment: The maximum RAM to verify a 1M constraint program is 2GB.
> Execution: The maximum RAM for a 1 or 31-depth 1M constraint call execution is negligible, but to be safe we could assume 10MB.

For an upper bound of 10GB of memory dedicated to transaction verification, we set the following values to:

`MAX_PARALLEL_DEPLOY_VERIFICATIONS`: 5
`MAX_PARALLEL_EXECUTE_VERIFICATIONS`: 1000

This means we verify deployments first (in chunks of 5), then executions after (in chunks of 1000).


ljedrz (approved):
LGTM :+1:

vicsn (approved):
Works like a charm

raychu86:
@vicsn Since the `pre-speculate` verification PR - https://github.com/AleoHQ/snarkVM/pull/2376 was merged, this PR had to be updated with the same change on the `VM::speculate` side.

ledger/src/check_transaction_basic.rs
```diff
@@ -24,4 +24,13 @@ impl<N: Network, C: ConsensusStorage<N>> Ledger<N, C> {
     ) -> Result<()> {
         self.vm().check_transaction(transaction, rejected_id, rng)
     }
+
+    /// Checks that the given list of transactions are well-formed and unique.
+    pub fn check_transactions_basic<R: CryptoRng + Rng>(
```
vicsn, line 29: Do you have a concrete usecase in mind for this function (in snarkOS)?

synthesizer/src/vm/finalize.rs
```diff
@@ -127,15 +109,16 @@ impl<N: Network, C: ConsensusStorage<N>> VM<N, C> {
...
         // Ensure each transaction is well-formed and unique.
         // NOTE: We perform the transaction checks here prior to `atomic_speculate` because we must
         // ensure that the `Fee` transactions are valid. We can't unify the transaction checks in `atomic_speculate`
         // because we run speculation on the unconfirmed variant of the transactions.
-        let rngs = (0..transactions.len()).map(|_| StdRng::from_seed(rng.gen())).collect::<Vec<_>>();
-        cfg_iter!(transactions).zip(rngs).try_for_each(|(transaction, mut rng)| {
-            self.check_transaction(transaction, transaction.to_rejected_id()?, &mut rng)
-                .map_err(|e| anyhow!("Invalid transaction found in the transactions list: {e}"))
-        })?;
+        self.check_transactions(&transactions_and_rejected_ids, rng)?;
```
vicsn, line 120: > We can't unify the transaction checks in `atomic_speculate` because we run speculation on the unconfirmed variant of the transactions.

Just for my understanding, can you give an example of why the post-speculation check_transactions is needed?

vicsn (approved):
LGTM, just two questions

synthesizer/src/vm/finalize.rs
raychu86: If any node receives a block while syncing, they need to perform the transaction validation through `VM::check_speculate` because they do not go through the path of constructing the block via `VM::speculate`

ledger/src/check_transaction_basic.rs
raychu86: Not particularly and we may not need this in snarkOS, but could be nice to have an interface to verify a bunch of transactions in parallel.

synthesizer/src/vm/verify.rs
```diff
@@ -31,6 +31,87 @@ macro_rules! ensure_is_unique {
...
+                cfg_into_iter!(transactions).zip(rngs).partition_map(|(transaction, mut rng)| {
+                    // Abort the transaction if it is a fee transaction.
+                    if transaction.is_fee() {
+                        return Either::Right((
+                            *transaction,
+                            "Fee transactions are not allowed in speculate".to_string(),
+                        ));
+                    }
+                    // Verify the transaction.
+                    match self.check_transaction(transaction, None, &mut rng) {
```
vicsn: I should have noticed this earlier, but be aware: if a malicious validator inserts a deployment into their Batch Proposal with a malformed verifying key, this thread may panic, causing all transmissions to be sent back into the mempool queue for verification.

The only way I see us preventing this is to perform validation on transactions of validator Batch Proposals before signing them.

I will make sure we have a consensus test case for this to reproduce.

raychu86: Good point, we should wrap the call with a `handle_halting` macro, like we do for other functions that have the capability of panicking.

raychu86: I added panic catchers in `check_transaction` and updated the test - [fbe604d](https://github.com/AleoHQ/snarkVM/pull/2387/commits/fbe604de0b9b7035ee07ab903a14c6da5202e8e6)

vicsn (approved):
LGTM for real now

howardwu (approved):

PR #2388, opened 2024-03-12
Move transaction verification into `atomic_speculate`

raychu86:
## Motivation

This PR is an extension of https://github.com/AleoHQ/snarkVM/pull/2376, where we move the pre-speculate transaction verification into `VM::atomic_speculate`. This will allow us to use the same TX verification logic in both `VM::speculate` and `VM::check_speculate`, which will reduce unnecessary duplication of transaction verification code.

### DIFF
--- a/ledger/src/check_next_block.rs
+++ b/ledger/src/check_next_block.rs
@@ -14,8 +14,6 @@
 
 use super::*;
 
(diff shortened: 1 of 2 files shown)


raychu86:
Closing due to undesired consequence of this change. We need to verify the Fee transaction types prior to calling `atomic_speculate`.

PR #2399, opened 2024-03-21
[Fix] Don't re-seed the TestRng when calling check_merkle_tree

ljedrz:
This PR fixes a small issue I've found when looking into the performance of `MerkleTree` - instead of reusing a single `TestRng` for the entirety of a single test, we are currently re-seeding it whenever we run `check_merkle_tree`, which makes the affected tests slower and more difficult to reproduce.

This is a test-only change and its introduction can be delayed for as long as necessary.


howardwu (approved):
LGTM.

For readers - this PR only touches test files. It does not touch non-test code.

PR #2402, opened 2024-03-21
[Feature] Add `transfer_public_as_signer` to `credits.aleo`

### LINKED ISSUE #2139: [Feature] Support transfer public credits from self.signer
## 🚀 Feature

I propose adding two new methods to `credits.aleo`, tentatively named `transfer_by_signer_public` and `transfer_by_signer_public_to_private`. These new methods would be similar to the existing `transfer_public` and `transfer_public_to_private`, but the payer in these new methods should be changed from `self.caller` to `self.signer`.

## Motivation

Currently, when `transfer_public` and `transfer_public_to_private` are invoked via a program, the cost is deducted from the program's account, not the actual initiator's (i.e., the signer's) account. This may not be the desired behavior in some scenarios.

For example, if a smart contract needs to make transfers on behalf of a user, the existing design deducts credits from the smart contract itself, rather than from the user (the signer). This could not only potentially deplete the smart contract's credits unexpectedly but may also introduce security and permission-related concerns.

By introducing these two new methods, we can ensure that the cost is correctly deducted from the signer's account, which is more logical and safer.

## Implementation

1. Define the two new methods, `transfer_by_signer_public` and `transfer_by_signer_public_to_private`, in the `credits.aleo` file.
2. The logic for these methods should be similar to that of the existing `transfer_public` and `transfer_public_to_private`, but `self.caller` should be replaced with `self.signer`.
3. Update the documentation and comments appropriately to reflect the existence and purpose of these new methods.
4. Add unit tests to validate the correctness and expected behavior of the new methods.

Affected components could include:
- The `credits.aleo` file
- Relevant unit tests
- Documentation

**Are you willing to open a pull request?**

Yes, I am willing to submit a pull request and engage in subsequent code reviews and testing.

d0cd:
## Motivation

This PR adds the `transfer_public_as_signer` function to `credits.aleo`. This function allows developers to write programs that interface with `credits.aleo` and transfer public credits through the written programs. 

## Test Plan

Additional tests were added checking that:
- `transfer_public_from_user`
- `transfer_public_from_program`
- `transfer_public_as_signer_from_user`
- `transfer_public_as_signer_from_program`

## Related PRs

Closes #2139


synthesizer/src/vm/mod.rs
```diff
@@ -1425,4 +1425,513 @@ finalize do:
...
+        // Verify the transaction.
+        vm.check_transaction(&transaction, None, rng).unwrap();
+
+        // Add the transaction to a block and update the VM.
+        let block = sample_next_block(&vm, &caller_private_key, &[transaction], rng).unwrap();
+
+        // Update the VM.
+        vm.add_next_block(&block).unwrap();
+
+        // Check the balance of the caller, which should be the same as before.
```
vicsn: Seems like the balance of the caller did decrease, you assert it to be `182_499_997_483_583` above.

d0cd: Ah good catch, it should decrease because the caller pays the fee.

howardwu (approved):

PR #2404, opened 2024-03-22
feat: add futures + argument serializer

miazn:
## Motivation

The `Future` and `Argument` structs' serializer/deserializer used the default `collect_str` method, which causes `future` types of Outputs to not convert to proper json, for example when you get a transaction from the REST endpoint it looks like

```  
"outputs": [
    {
      "type": "future",
      "id": "6056862602270631138152110549113103883216888689162817898151691740816207450024field",
      "value": "{\n  program_id: credits.aleo,\n  function_name: transfer_public,\n  arguments: [\n    aleo12z722rd0cyydzazcl5zu22e98yhqj4zs3s6pwqdsl6q04chu8czqtgwyje,\n    aleo1j4jk9243sqcydp0tek7pfnur6a2a7dd7xrfcgqpr2pekgpc9ls9sflvpyy,\n    27000202u64\n  ]\n}"
    }
  ],
```

This PR implements a serialize/deserialize for both the Future and Argument types to enable proper JSON outputs, like:
image

## Test Plan

Added serde/bincode tests for the `Future` struct


raychu86:
@d0cd Looks like CI is failing due to `A transition output value is missing` errors

miazn:
Fixed the transition output value, realized that it was still using the `From_str` which was causing a serialize/deserialize issue

howardwu:
@d0cd, @evan-schott, @ljedrz @vicsn can you carefully review this PR?

If we add JSON support, we run the risk that the order may matter and need to ensure the order of arrays and types remain consistent in this new design during deserialization.

vicsn:
Thank you for flagging Howard. This would not have occured to me immediately, but this PR does look safe to me.

JSON doesn't guarantee ordering of objects keys, but it does guarantee ordering of array elements. ([source](https://stackoverflow.com/questions/7214293/is-the-order-of-elements-in-a-json-list-preserved)) ([source's source](https://www.rfc-editor.org/rfc/rfc7159.html)).

If we're making changes to serialization, I think we should apply them uniformly across the API, so I would also suggest we also make changes to some other objects which have the same almost-but-not-quite JSON problem. A quick scan for `write!(f, "\n` gives me at least: Plaintext Records, Plaintext, Entry, RecordType.

I see the value in using an existing format for communicating our data. Otherwise @unzvfu and other eager users will keep writing custom JSON parsers. I think checking that all elements with order are serialized into arrays is sufficient.

ljedrz:
I agree with @vicsn; also, we shouldn't rely on the order of the key-value maps, even if we're using the `preserve_order` feature, for the sake of other implementations.

console/program/src/data/future/serialize.rs
```diff
@@ -15,21 +15,85 @@
 use super::*;
 
 impl<N: Network> Serialize for Future<N> {
-    /// Serializes the future into a string or as bytes.
```
d0cd, line 18: Nit: include the comment.

console/program/src/data/future/serialize.rs
```diff
@@ -15,21 +15,85 @@
...
+                state.serialize_field("arguments", &self.arguments)?;
+                state.end()
+            }
             false => ToBytesSerializer::serialize_with_size_encoding(self, serializer),
         }
     }
 }
 
 impl<'de, N: Network> Deserialize<'de> for Future<N> {
-    /// Deserializes the future from a string or bytes.
```
d0cd, line 28: Nit: include the comment.

d0cd (changes requested):
Left some nits. Overall, the implementation looks good! 

There remains the broader question of using JSON serialization for the other (program-level data types). I see the benefits in using a JSON format, but it does require additional work in standardizing the interface and ensuring that the change does not  negatively impact "programming layers" (Leo and Aleo Instructions). 

Was there a reason why we didn't go with JSON before?

howardwu:
I am still concerned that this breaks the transaction output format. Can anyone confirm? 

Does this change how we pass transaction inputs from CLI at all?

miazn:
closing for now bc merging this would require a lot of rework

PR #2407, opened 2024-03-23
Reduce zero-padding and cache zero-hashes in k-ary MerkleTree

ljedrz:
When creating k-ary Merkle trees, we currently pad all the "missing" leaves, which can result in a large number of empty hashes that all end up being allocated, and then repeatedly included in the calculation of further hashes. In order to improve performance for all such scenarios, we can omit some of the padding, and cache the hashes that derive from the zero hash, which is what this PR proposes.

While there are still ways in which the performance can be improved further, these changes aren't exactly trivial already, and it seems like a good moment to open them to discussion. Below are some benchmarks that I'll be referring to when outlining the potential for future improvement.

```
creation of a KaryMerkleTree with ARITY = 8, DEPTH = 8, NUM_LEAVES = ARITY^(DEPTH-1)+1, and SHA3 hashers

before:

alloc count: 48,061,051
heap use:    5.12 GiB
max heap:    10.17 GiB

after:

alloc count: 21,084,574 (-56%)
heap use:    1.62 GiB (-68%)
max heap:    3.17 GiB (-69%)

```
These are all wins, with a significantly lower both maximum and current (once the tree is created) heap use.
```
console/collections/benches/kary_merkle_tree:

KaryMerkleTree/new/0 (0% full)
                        time:   [21.247 µs 21.308 µs 21.383 µs]
                        change: [+93.568% +95.333% +96.860%] (p = 0.00 < 0.05)
                        Performance has regressed.
Found 2 outliers among 10 measurements (20.00%)
  1 (10.00%) high mild
  1 (10.00%) high severe

Benchmarking KaryMerkleTree/new/4194304 (25% full): Warming up for 3.0000 s
Warning: Unable to complete 10 samples in 5.0s. You may wish to increase target time to 20.4s.
KaryMerkleTree/new/4194304 (25% full)
                        time:   [2.0185 s 2.0246 s 2.0308 s]
                        change: [-51.205% -51.016% -50.847%] (p = 0.00 < 0.05)
                        Performance has improved.

Benchmarking KaryMerkleTree/new/8388608 (50% full): Warming up for 3.0000 s
Warning: Unable to complete 10 samples in 5.0s. You may wish to increase target time to 38.2s.
KaryMerkleTree/new/8388608 (50% full)
                        time:   [3.8503 s 3.8668 s 3.8835 s]
                        change: [-25.467% -25.025% -24.593%] (p = 0.00 < 0.05)
                        Performance has improved.

Benchmarking KaryMerkleTree/new/12582912 (75% full): Warming up for 3.0000 s
Warning: Unable to complete 10 samples in 5.0s. You may wish to increase target time to 57.7s.
KaryMerkleTree/new/12582912 (75% full)
                        time:   [5.7279 s 5.7525 s 5.7726 s]
                        change: [-8.8066% -8.3742% -7.9833%] (p = 0.00 < 0.05)
                        Performance has improved.
Found 2 outliers among 10 measurements (20.00%)
  2 (20.00%) low mild

Benchmarking KaryMerkleTree/new/16777216 (100% full): Warming up for 3.0000 s
Warning: Unable to complete 10 samples in 5.0s. You may wish to increase target time to 75.1s.
KaryMerkleTree/new/16777216 (100% full)
                        time:   [7.5454 s 7.5707 s 7.5963 s]
                        change: [+3.6674% +4.2182% +4.7320%] (p = 0.00 < 0.05)
                        Performance has regressed.
```
The creation of an empty and a full tree see regressions, as those scenarios don't benefit from the caching of empty hashes. I have an idea on how to improve them, which will be posted as an additional commit if the diff isn't too large.

### DIFF
--- a/console/collections/src/kary_merkle_tree/mod.rs
+++ b/console/collections/src/kary_merkle_tree/mod.rs
@@ -120,11 +120,6 @@ impl<LH: LeafHash<Hash = PH::Hash>, PH: PathHash, const DEPTH: u8, const ARITY: 
                     // Prepare an iterator over then children, being mindful of possible missing leaves.
                     let child_iter = || child_indexes::<ARITY>(i).map(|child_index| tree.get(child_index).copied());
 
-                    // At the leaf level just return the leaves, or `None` in case the node has none.
-                    if current_empty_node_hash == empty_hash {
-                        return child_iter().collect::<Option<Vec<_>>>();
-                    }
-
                     // Check if the children aren't all derived from empty hashes.
                     if child_iter().all(|hash| hash == Some(current_empty_node_hash)) {
                         return None;
--- a/console/collections/src/kary_merkle_tree/mod.rs
+++ b/console/collections/src/kary_merkle_tree/mod.rs
@@ -24,6 +24,7 @@ mod tests;
 use snarkvm_console_types::prelude::*;
 
 use aleo_std::prelude::*;
+use std::ops::Range;
 
 #[derive(Clone)]
 pub struct KaryMerkleTree<LH: LeafHash<Hash = PH::Hash>, PH: PathHash, const DEPTH: u8, const ARITY: u8> {
@@ -116,12 +117,13 @@ impl<LH: LeafHash<Hash = PH::Hash>, PH: PathHash, const DEPTH: u8, const ARITY: 
             // Construct the children for each node in the current level.
             let child_nodes = (start..end)
                 .take_while(|&i| child_indexes::<ARITY>(i).next().and_then(|idx| tree.get(idx)).is_some())
-                .map(|i| child_indexes::<ARITY>(i).map(|child_index| tree[child_index]).collect::<Vec<_>>())
+                .map(|i| &tree[child_indexes::<ARITY>(i)])
                 .collect::<Vec<_>>();
 
             // Compute and store the hashes for each node in the current level.
             let num_full_nodes = child_nodes.len();
-            tree[start..][..num_full_nodes].clone_from_slice(&path_hasher.hash_all_children(&child_nodes)?);
+            let hashes = path_hasher.hash_all_children(&child_nodes)?;
+            tree[start..][..num_full_nodes].clone_from_slice(&hashes);
             // Use the precomputed empty node hash for every empty node, if there are any.
             if start + num_full_nodes < end {
                 let empty_node_hash = path_hasher.hash_children(&vec![empty_hash; arity])?;
@@ -261,7 +263,7 @@ fn tree_depth<const DEPTH: u8, const ARITY: u8>(tree_size: usize) -> Result<u8> 
 }
 
 /// Returns the indexes of the children, given an index.
-fn child_indexes<const ARITY: u8>(index: usize) -> impl Iterator<Item = usize> {
+fn child_indexes<const ARITY: u8>(index: usize) -> Range<usize> {
     let start = index * ARITY as usize + 1;
     start..start + ARITY as usize
 }
--- a/console/collections/src/kary_merkle_tree/mod.rs
+++ b/console/collections/src/kary_merkle_tree/mod.rs
@@ -96,13 +96,6 @@ impl<LH: LeafHash<Hash = PH::Hash>, PH: PathHash, const DEPTH: u8, const ARITY: 
             num_nodes + leaves.len() + if leaves.len() % arity == 0 { 0 } else { arity - leaves.len() % arity },
         );
 
-        // Prepare a cache of hashes of empty nodes and their derivatives.
-        let mut empty_hash_cache = Vec::with_capacity(DEPTH as usize);
-        empty_hash_cache.push(path_hasher.hash_children(&vec![empty_hash; arity])?);
-        for i in 0..(DEPTH as usize - 1) {
-            empty_hash_cache.push(path_hasher.hash_children(&vec![empty_hash_cache[i]; arity])?);
-        }
-
         // Initialize the Merkle tree.
         let mut tree = vec![empty_hash; minimum_tree_size];
 
@@ -112,7 +105,7 @@ impl<LH: LeafHash<Hash = PH::Hash>, PH: PathHash, const DEPTH: u8, const ARITY: 
 
         // Compute and store the hashes for each level, iterating from the penultimate level to the root level.
         let mut start_index = num_nodes;
-        let mut current_level = 0;
+        let mut current_empty_node_hash = path_hasher.hash_children(&vec![empty_hash; arity])?;
         // Compute the start index of the current level.
         while let Some(start) = parent::<ARITY>(start_index) {
             // Compute the end index of the current level.
@@ -125,13 +118,12 @@ impl<LH: LeafHash<Hash = PH::Hash>, PH: PathHash, const DEPTH: u8, const ARITY: 
                     let child_iter = || child_indexes::<ARITY>(i).map(|child_index| tree.get(child_index).copied());
 
                     // At the leaf level just return the leaves, or `None` in case the node has none.
-                    if current_level == 0 {
+                    if current_empty_node_hash == empty_hash {
                         return child_iter().collect::<Option<Vec<_>>>();
                     }
 
                     // Check if the children aren't all derived from empty hashes.
-                    let current_composite_empty_hash = empty_hash_cache[current_level - 1];
-                    if child_iter().all(|hash| hash == Some(current_composite_empty_hash)) {
+                    if child_iter().all(|hash| hash == Some(current_empty_node_hash)) {
                         return None;
                     }
 
@@ -141,12 +133,11 @@ impl<LH: LeafHash<Hash = PH::Hash>, PH: PathHash, const DEPTH: u8, const ARITY: 
                 .collect::<Vec<_>>();
 
             // Compute and store the hashes for each node in the current level.
-            tree[start..end].clone_from_slice(
-                &path_hasher.hash_all_children::<ARITY>(&child_nodes, empty_hash_cache[current_level])?,
-            );
+            tree[start..end].clone_from_slice(&path_hasher.hash_all_children(&child_nodes, current_empty_node_hash)?);
             // Update the start index for the next level.
             start_index = start;
-            current_level += 1;
+            // Update the empty node hash for the next level.
+            current_empty_node_hash = path_hasher.hash_children(&vec![current_empty_node_hash; arity])?;
         }
         lap!(timer, "Hashed {} levels", tree_depth);
 
--- a/console/collections/src/kary_merkle_tree/mod.rs
+++ b/console/collections/src/kary_merkle_tree/mod.rs
@@ -129,8 +129,8 @@ impl<LH: LeafHash<Hash = PH::Hash>, PH: PathHash, const DEPTH: u8, const ARITY: 
             // Use the precomputed empty node hash for every empty node, if there are any.
             if start + num_full_nodes < end {
                 let empty_node_hash = path_hasher.hash_children(&vec![empty_hash; arity])?;
-                for idx in start + num_full_nodes..end {
-                    tree[idx] = empty_node_hash;
+                for node in tree.iter_mut().take(end).skip(start + num_full_nodes) {
+                    *node = empty_node_hash;
                 }
             }
             // Update the start index for the next level.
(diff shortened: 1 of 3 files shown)


ljedrz:
The regressions turned out to be simpler to iron out than expected:
```
KaryMerkleTree/new/0 (0% full)
                        time:   [12.495 µs 12.524 µs 12.564 µs]
                        change: [+12.772% +13.612% +14.468%] (p = 0.00 < 0.05)
                        Performance has regressed.
Found 2 outliers among 10 measurements (20.00%)
  1 (10.00%) high mild
  1 (10.00%) high severe

KaryMerkleTree/new/16777216 (100% full)
                        time:   [7.4351 s 7.4754 s 7.5136 s]
                        change: [-4.6663% -3.8930% -3.0814%] (p = 0.00 < 0.05)
                        Performance has improved.
```
The empty tree case is still a bit worse than before, but it's a tiny difference measured in µs and not the typical scenario. Also, that one seems likely to fluctuate - I just re-ran it (without any changes), and saw a -6% difference.

console/collections/src/kary_merkle_tree/tests/mod.rs
```diff
@@ -170,7 +170,7 @@ fn check_merkle_tree_depth_3_arity_3_padded<LH: LeafHash<Hash = PH::Hash>, PH: P
 
     // Construct the Merkle tree for the given leaves.
     let merkle_tree = KaryMerkleTree::<LH, PH, 3, ARITY>::new(leaf_hasher, path_hasher, &leaves)?;
-    assert_eq!(40, merkle_tree.tree.len());
+    assert_eq!(25, merkle_tree.tree.len());
```
howardwu, line 173: Why did this change?

ljedrz: With this PR, the nodes that would only contain empty hashes (padding) no longer need to allocate them at the leaf level, which means that the tree ends up being smaller; you can see a calculation of the difference in the current number of leaves - padded for the entire leaf level - and the one this PR proposes (only padded up to the arity) [here](https://play.rust-lang.org/?version=stable&mode=debug&edition=2021&gist=96f2533f23996449cc35364b53b75c79). In case of this particular test case (arity = 3, depth = 3, leaves = 10), the number of tree members (at the leaf level) that are just padding is currently 17, which this PR takes down to 2, which is a reduction of 15, leading to the corresponding reduction in tree size.

howardwu: Does this change apply to the `merkle-tree` too? (non-k-ary one) If so we shoudl optimize and profile that one too.

ljedrz: I've already done that; the wins aren't as great, but I'm happy to file a PR soon if this direction is desirable.

ljedrz: https://github.com/AleoHQ/snarkVM/pull/2415

howardwu:
@ljedrz is this a breaking PR change? Does the Merkle tree's expected outputs and roots change in any way?

ljedrz:
I don't think so - hashes for the nodes containing empty hashes are calculated as if the empty hashes were there, they are only not physically present in the tree; the number of leaves field is not affected either; are there any test vectors I could double-check against?

ljedrz:
@howardwu I've been running equivalence tests (both `KaryMerkleTree` and `MerkleTree`) by @d0cd for a large number of iterations and haven't encountered any failures.

d0cd:
For those interested, the equivalence tests can be found in [this](https://github.com/AleoHQ/snarkvm-equivalence-tests) repo. (currently private but the AleoHQ/snarkVM team has access)

A sample equivalence test for the `Puzzle` is below.
Note that `snarkvm_old` points to `mainnet-staging`, and `snarkvm_new` points to `ljedrz:perf/kary_merkle_tree_padding`
```rust
use snarkvm_old::ledger::puzzle::Puzzle as OldPuzzle;
use snarkvm_old::ledger::store::helpers::memory::ConsensusMemory as OldMemory;
use snarkvm_old::prelude::{
    Address as OldAddress, MainnetV0 as OldNetwork, Network as OldNetworkTrait, Uniform,
};
use snarkvm_old::synthesizer::vm::VM as OldVM;

use snarkvm_new::ledger::puzzle::Puzzle as NewPuzzle;
use snarkvm_new::ledger::store::helpers::memory::ConsensusMemory as NewMemory;
use snarkvm_new::prelude::{
    Address as NewAddress, MainnetV0 as NewNetwork, Network as NewNetworkTrait,
};
use snarkvm_new::synthesizer::vm::VM as NewVM;

use anyhow::Result;
use std::str::FromStr;

const ITERATIONS: u128 = 100;

fn check_equivalency(
    old_puzzle: &OldPuzzle<OldNetwork>,
    new_puzzle: &NewPuzzle<NewNetwork>,
) -> Result<()> {
    // Initialize an RNG.
    let rng = &mut rand::thread_rng();
    // Check that the puzzles produce equivalent solutions.
    for _ in 0..ITERATIONS {
        // Generate a random epoch hash.
        let old_epoch_hash = <OldNetwork as OldNetworkTrait>::BlockHash::rand(rng);
        let new_epoch_hash =
            <NewNetwork as NewNetworkTrait>::BlockHash::from_str(&format!("{old_epoch_hash}"))?;

        // Generate a random address and counter.
        let old_address = OldAddress::<OldNetwork>::rand(rng);
        let new_address = NewAddress::<NewNetwork>::from_str(&format!("{old_address}"))?;

        let old_counter = u64::rand(rng);
        let new_counter = u64::from_str(&format!("{old_counter}"))?;

        let old_solution = old_puzzle.prove(old_epoch_hash, old_address, old_counter, None)?;
        let new_solution = new_puzzle.prove(new_epoch_hash, new_address, new_counter, None)?;

        let old_proof_target = old_puzzle.get_proof_target(&old_solution)?;
        let new_proof_target = new_puzzle.get_proof_target(&new_solution)?;

        assert_eq!(old_proof_target, new_proof_target);
    }
    Ok(())
}

#[test]
fn test_puzzles_are_equivalent() -> Result<()> {
    // Construct the old puzzle.
    let old_puzzle = OldVM::<OldNetwork, OldMemory<OldNetwork>>::new_puzzle()?;

    // Construct the new puzzle.
    let new_puzzle = NewVM::<NewNetwork, NewMemory<NewNetwork>>::new_puzzle()?;

    // Check that the old and new puzzles produce equivalent solutions.
    check_equivalency(&old_puzzle, &new_puzzle)?;

    Ok(())
}

```

howardwu:
@vicsn @mdelle1 can each of you independently review this design change for correctness/completeness and soundness?

console/collections/src/kary_merkle_tree/mod.rs
```diff
@@ -89,28 +89,55 @@ impl<LH: LeafHash<Hash = PH::Hash>, PH: PathHash, const DEPTH: u8, const ARITY:
         // Compute the empty hash.
         let empty_hash = path_hasher.hash_empty::<ARITY>()?;
 
+        // Calculate the size of the tree which excludes leafless nodes.
+        let arity = ARITY as usize;
+        let minimum_tree_size = std::cmp::max(
+            1,
+            num_nodes + leaves.len() + if leaves.len() % arity == 0 { 0 } else { arity - leaves.len() % arity },
```
raychu86: nit: Could you make this more readable and add some documentation?

console/collections/src/kary_merkle_tree/mod.rs
```diff
@@ -89,28 +89,55 @@ impl<LH: LeafHash<Hash = PH::Hash>, PH: PathHash, const DEPTH: u8, const ARITY:
...
 
             // Construct the children for each node in the current level.
             let child_nodes = (start..end)
-                .map(|i| child_indexes::<ARITY>(i).map(|child_index| tree[child_index]).collect::<Vec<_>>())
+                .map(|i| {
+                    // Prepare an iterator over then children, being mindful of possible missing leaves.
+                    let child_iter = || child_indexes::<ARITY>(i).map(|child_index| tree.get(child_index).copied());
+
+                    // At the leaf level just return the leaves, or `None` in case the node has none.
+                    if current_empty_node_hash == empty_hash {
```
d0cd: Will this condition ever trigger? 
At leaf level, `current_empty_node_hash = path_hasher.hash_children(&vec![empty_hash; arity])?` which would never equal `empty_hash` no?

console/collections/src/kary_merkle_tree/mod.rs
```diff
@@ -89,28 +89,55 @@ impl<LH: LeafHash<Hash = PH::Hash>, PH: PathHash, const DEPTH: u8, const ARITY:
...
+                        return child_iter().collect::<Option<Vec<_>>>();
+                    }
+
+                    // Check if the children aren't all derived from empty hashes.
+                    if child_iter().all(|hash| hash == Some(current_empty_node_hash)) {
```
d0cd: Does the optimized pathway ever get triggered for any level past the first?

console/collections/src/kary_merkle_tree/mod.rs
ljedrz: Ah yes, you're right; I also tested it, and it never triggered; I'll remove it :+1:.

console/collections/src/kary_merkle_tree/mod.rs
ljedrz: It doesn't; just like with your previous comment, I think this was done before I realized that the empty hashes don't actually need to be there.

ljedrz: Although... I can somewhat see how this could be triggered in some cases? But I'm not hitting it in tests.

ljedrz:
I'm going to check if any of these review comments apply to https://github.com/AleoHQ/snarkVM/pull/2415 as well.

Edit: it seems that way, but let's finish this PR first so as to make sure that everything is in good shape.

console/collections/src/kary_merkle_tree/mod.rs
vicsn: Removing it is a win for simplicity anyway

vicsn (approved):
As far as I can tell: this PR does not impact completeness/soundness of the merkle tree opening proof, because it produces merkle tree nodes which are identical to before.

Note that due to the particular imlementation, we can't prove membership anymore of empty leaves outside of range of the originally supplied leaves, which seems like a desirable bonus.

ljedrz:
I'll rerun the benchmarks for this and the other PR shortly.

ljedrz:
The updated benchmark results (skipping the `0% full` case, as it's inconsistent) are:
```
Benchmarking KaryMerkleTree/new/4194304 (25% full): Warming up for 3.0000 s
KaryMerkleTree/new/4194304 (25% full)
                        time:   [1.0458 s 1.0503 s 1.0551 s]
                        change: [-73.821% -73.661% -73.488%] (p = 0.00 < 0.05)
                        Performance has improved.

Benchmarking KaryMerkleTree/new/8388608 (50% full): Warming up for 3.0000 s
KaryMerkleTree/new/8388608 (50% full)
                        time:   [1.9405 s 1.9510 s 1.9611 s]
                        change: [-58.942% -58.687% -58.449%] (p = 0.00 < 0.05)
                        Performance has improved.

Benchmarking KaryMerkleTree/new/12582912 (75% full): Warming up for 3.0000 s
KaryMerkleTree/new/12582912 (75% full)
                        time:   [2.8457 s 2.8569 s 2.8664 s]
                        change: [-47.101% -46.725% -46.350%] (p = 0.00 < 0.05)
                        Performance has improved.
Found 2 outliers among 10 measurements (20.00%)
  2 (20.00%) low mild

Benchmarking KaryMerkleTree/new/16777216 (100% full): Warming up for 3.0000 s
KaryMerkleTree/new/16777216 (100% full)
                        time:   [3.7669 s 3.8062 s 3.8520 s]
                        change: [-39.205% -38.257% -37.313%] (p = 0.00 < 0.05)
                        Performance has improved.

KaryMerkleTree/verify   time:   [14.508 µs 14.608 µs 14.737 µs]
                        change: [-7.9402% -6.7762% -5.4263%] (p = 0.00 < 0.05)
                        Performance has improved.
Found 10 outliers among 100 measurements (10.00%)
  10 (10.00%) high severe
```
Which is a further improvement across the board.

vicsn (approved):
ACK on the new commits

d0cd (approved):

PR #2415, opened 2024-04-01
Reduce zero-padding and cache zero-hashes in MerkleTree

ljedrz:
This PR is the "plain" Merkle tree counterpart to https://github.com/AleoHQ/snarkVM/pull/2407.

While some of the `console/collections/merkle_tree` benchmark results are mixed, any regressions are small; however, I also ran a custom benchmark locally, and that one shows a clear improvement in extreme scenarios:
```
creation of a BHP MerkleTree with DEPTH = 23 and 2^22 + 1 random leaves

before:

alloc count: 38,230,906
heap use:    1.59 GiB
max heap:    2.10 GiB
time:        51.88s

after:

alloc count: 31,932,664 (-16.5%)
heap use:    1.47 GiB (-7.5%)
max heap:    2.00 GiB (-4.8%)
time:        42.61s (-17.9%)
```

### DIFF
--- a/console/collections/src/merkle_tree/mod.rs
+++ b/console/collections/src/merkle_tree/mod.rs
@@ -96,14 +96,11 @@ impl<E: Environment, LH: LeafHash<Hash = PH::Hash>, PH: PathHash<Hash = Field<E>
         while let Some(start) = parent(start_index) {
             // Compute the end index of the current level.
             let end = left_child(start);
-            // Construct the children for each node in the current level.
+            // Construct the children for each node in the current level; the leaves are padded, which means
+            // that there either are 2 children, or there are none, at which point we may stop iterating.
             let tuples = (start..end)
-                .filter_map(|i| {
-                    // Procure the children of the node.
-                    let tuple = (tree.get(left_child(i)).copied(), tree.get(right_child(i)).copied());
-                    // If the left child is missing (in which case the right one also is), return `None`.
-                    if tuple.0.is_none() { None } else { Some((tuple.0.unwrap(), tuple.1.unwrap())) }
-                })
+                .take_while(|&i| tree.get(left_child(i)).is_some())
+                .map(|i| (tree[left_child(i)], tree[right_child(i)]))
                 .collect::<Vec<_>>();
             // Compute and store the hashes for each node in the current level.
             let num_full_nodes = tuples.len();
--- a/console/collections/src/merkle_tree/mod.rs
+++ b/console/collections/src/merkle_tree/mod.rs
@@ -77,8 +77,12 @@ impl<E: Environment, LH: LeafHash<Hash = PH::Hash>, PH: PathHash<Hash = Field<E>
         // Compute the empty hash.
         let empty_hash = path_hasher.hash_empty()?;
 
+        // Calculate the size of the tree which excludes leafless nodes.
+        let minimum_tree_size =
+            std::cmp::max(1, num_nodes + leaves.len() + if leaves.len() > 1 { leaves.len() % 2 } else { 0 });
+
         // Initialize the Merkle tree.
-        let mut tree = vec![empty_hash; tree_size];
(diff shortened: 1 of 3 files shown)


howardwu:
@vicsn @mdelle1 can each of you independently review this design change for correctness/completeness and soundness?

ljedrz:
Note: I recommend that we finish https://github.com/AleoHQ/snarkVM/pull/2407, as some of the review comments there are applicable to this PR too.

console/collections/src/merkle_tree/mod.rs
```diff
@@ -77,25 +77,46 @@ impl<E: Environment, LH: LeafHash<Hash = PH::Hash>, PH: PathHash<Hash = Field<E>
...
+            let tuples = (start..end)
+                .map(|i| {
+                    // Procure the children of the node.
+                    let tuple = (tree.get(left_child(i)).copied(), tree.get(right_child(i)).copied());
+                    // If both children are empty hashes, return `None`.
+                    if tuple.0 == Some(current_empty_node_hash) && tuple.1 == Some(current_empty_node_hash) {
+                        return None;
+                    }
+                    // If any of the children are missing, return `None`.
+                    if tuple.0.is_none() || tuple.1.is_none() {
```
vicsn: My understanding is that it would be a protocol violation if only one of them is None, because you `pad up to the tree's arity (making every node full)`. Maybe good to `ensure` it for clarity? Or otherwise leave a comment.

ljedrz:
Converting to a draft until https://github.com/AleoHQ/snarkVM/pull/2407 is reviewed, as they have many common features, and there are already review comments in that one that I've locally applied to this PR.

console/collections/src/merkle_tree/mod.rs
ljedrz: Since the k-ary PR has been approved, I updated this PR and this previously valid comment no longer applies.

console/collections/src/merkle_tree/mod.rs
```diff
@@ -91,9 +97,28 @@ impl<E: Environment, LH: LeafHash<Hash = PH::Hash>, PH: PathHash<Hash = Field<E>
...
-            let tuples = (start..end).map(|i| (tree[left_child(i)], tree[right_child(i)])).collect::<Vec<_>>();
+            let tuples = (start..end)
+                .filter_map(|i| {
+                    // Procure the children of the node.
+                    let tuple = (tree.get(left_child(i)).copied(), tree.get(right_child(i)).copied());
+                    // If both children are missing, return `None`.
+                    if tuple.0.is_none() && tuple.1.is_none() {
+                        None
+                    } else {
+                        Some((tuple.0.unwrap(), tuple.1.unwrap()))
```
vicsn: Given that we return a `Result`, I think we might as well `ensure` that the left and right tuple are both `Some` in order to prevent a panic.

vicsn (approved):
Apart from one nit (adding the ensure), LGTM. [This comment](https://github.com/AleoHQ/snarkVM/pull/2407#pullrequestreview-1982825648) applies here too.

I was also told equivalence tests have been run on the binary merkle tree just like they were for the k-ary.

console/collections/src/merkle_tree/mod.rs
ljedrz: Actually, it's impossible for only one of them to be `None`, because we pad single-child nodes with a zero hash; however, this means that we might as well only check either member of the tuple, which will be faster :+1:

ljedrz: Also, this comment made me realize that we could actually use `take_while` here for shorter iteration :eyes:.

ljedrz:
@vicsn PR updated; let me know if you agree with the second change, as it would also apply to the k-ary Merkle tree PR :grin:. Also, the equivalence tests still pass.

vicsn (approved):

d0cd (approved):

PR #2419, opened 2024-04-02
[Feature] new `network.id` opcode

evan-schott:
## Motivation
This PR introduces the opcode `network.id` to access the ID of the current network (ex: `0` for mainnet, or `3` for testnet3). This opcode can be used as a safeguard against replay attacks, as signatures or hashes of data can now be bound to a specific network. Users can now test program authorizations on testnet without worrying that the same authorizations be maliciously reused on mainnet. 

## Open questions
1. Should `network.id` be allowed in transition functions? 
    - We could not think of any potential use cases of such a feature that could not just be modified to attesting to a network ID value (as a transition function input), and then passing that value to finalize to verify it is correct. 
    - Additionally, @bendyarm has pointed out that including this may result in potential difficulties with resolving the compile time value of `network.id` as it could be obfuscated if it were involved in computation.
2. Should all block metadata be readily accessible? [Relevant ARC](https://github.com/AleoHQ/ARCs/discussions/69#discussion-6452698)

### DIFF
--- a/ledger/src/tests.rs
+++ b/ledger/src/tests.rs
@@ -23,6 +23,7 @@ use console::{
     account::{Address, PrivateKey},
     network::prelude::*,
     program::{Entry, Identifier, Literal, Plaintext, ProgramID, Value},
+    types::U16,
 };
 use ledger_block::{ConfirmedTransaction, Rejected, Transaction};
 use ledger_committee::{Committee, MIN_VALIDATOR_STAKE};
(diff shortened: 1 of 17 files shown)


evan-schott:
Should we add the network ID to the block-specific random seed?
https://github.com/AleoHQ/snarkVM/blob/69b5b17a9d7f32e4f3abbc562ef080adfb654877/synthesizer/program/src/logic/finalize_global_state/mod.rs#L57-L63

ledger/src/tests.rs
```diff
@@ -29,6 +29,7 @@ use ledger_committee::{Committee, MIN_VALIDATOR_STAKE};
 use ledger_store::{helpers::memory::ConsensusMemory, ConsensusStore};
 use synthesizer::{program::Program, vm::VM, Stack};
 
+use console::types::U16;
```
d0cd: Nit: move up to L22.

synthesizer/process/src/stack/evaluate.rs
```diff
@@ -82,6 +82,8 @@ impl<N: Network> StackEvaluate<N> for Stack<N> {
                     Operand::Caller => Ok(Value::Plaintext(Plaintext::from(Literal::Address(registers.caller()?)))),
                     // If the operand is the block height, throw an error.
                     Operand::BlockHeight => bail!("Cannot retrieve the block height from a closure scope."),
+                    // If the operand is the network id, throw an error.
+                    Operand::NetworkID => bail!("Cannot retrieve the network ID from a closure scope."),
```
d0cd, line 86: What is the reason why this shouldn't be available in evaluation?
Same for below.

synthesizer/process/src/stack/execute.rs
```diff
@@ -115,6 +115,10 @@ impl<N: Network> StackExecute<N> for Stack<N> {
                     Operand::BlockHeight => {
                         bail!("Illegal operation: cannot retrieve the block height in a closure scope")
                     }
+                    // If the operand is the network id, throw an error.
+                    Operand::NetworkID => {
+                        bail!("Illegal operation: cannot retrieve the network id in a closure scope")
```
d0cd, line 120: What is the reason why this shouldn't be available in execution?
Same for below.

synthesizer/process/src/stack/register_types/mod.rs
```diff
@@ -98,6 +98,7 @@ impl<N: Network> RegisterTypes<N> {
                 RegisterType::Plaintext(PlaintextType::Literal(LiteralType::Address))
             }
             Operand::BlockHeight => bail!("'block.height' is not a valid operand in a non-finalize context."),
+            Operand::NetworkID => bail!("'network.id' is not a valid operand in a non-finalize context."),
```
d0cd, line 101: Should it be valid?

d0cd (changes requested):
Left some feedback and nits.
The main question we should answer is whether or not to allow `network.id` in a non-finalize context.

synthesizer/process/src/stack/evaluate.rs
d0cd: Discussed offline. Excluding in execution and evaluation does not prevent the desired use case as you can directly pass in a `u32` and check it against the network in a finalize block.

synthesizer/process/src/stack/execute.rs
d0cd: Discussed offline. Excluding in execution and evaluation does not prevent the desired use case as you can directly pass in a u32 and check it against the network in a finalize block.

synthesizer/process/src/stack/register_types/mod.rs
d0cd: Discussed offline. Excluding in execution and evaluation does not prevent the desired use case as you can directly pass in a u32 and check it against the network in a finalize block.

d0cd:
> Should we add the network ID to the block-specific random seed?
> 
> https://github.com/AleoHQ/snarkVM/blob/69b5b17a9d7f32e4f3abbc562ef080adfb654877/synthesizer/program/src/logic/finalize_global_state/mod.rs#L57-L63

It's not necessary for this PR, but could be a useful property. Networks with different IDs have different RNG state.

d0cd (approved):

bendyarm (approved):

PR #2423, opened 2024-04-05
[Test] Proof target tests

iamalwaysuncomfortable:
## Motivation

This PR adds test coverage to the cumulative target adjustment logic in both block formation & and block verification over multiple blocks. This bases off #2422.


d0cd (approved):
LGTM!

PR #2424, opened 2024-04-05
Use a parallel sort to order txs while speculating

ljedrz:
A small follow-up to https://github.com/AleoHQ/snarkVM/pull/2421. The changes are:
- a `cfg_sort_unstable_by` macro for parallel sorting using a closure is introduced (note: the `unstable` part only means that **equal** elements may be reordered, and transaction IDs are never equal)
- the aforementioned macro is used to sort transactions when speculating
- the parallel iteration for the collection is removed, as it is slower than single-threaded iteration (since `Transaction::id` is basically free)
- instead of enumerating and collecting indices, [IndexSet::get_index_of](https://docs.rs/indexmap/latest/indexmap/set/struct.IndexSet.html#method.get_index_of) is used

### DIFF
--- a/synthesizer/src/vm/finalize.rs
+++ b/synthesizer/src/vm/finalize.rs
@@ -15,6 +15,7 @@
 use super::*;
 
 use ledger_committee::{MAX_DELEGATORS, MIN_DELEGATOR_STAKE, MIN_VALIDATOR_STAKE};
+use utilities::cfg_sort_unstable_by;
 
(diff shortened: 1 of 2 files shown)


ljedrz:
The CI jobs are failing due to a GitHub incident; will rerun a bit later.

ljedrz:
It seems that for whatever reason the CI can't be triggered successfully after the GitHub failure; I'm going to open a new PR with a new branch name.

PR #2425, opened 2024-04-05
Use a parallel sort to order txs while speculating

ljedrz:
A small follow-up to https://github.com/AleoHQ/snarkVM/pull/2421 and the replacement of https://github.com/AleoHQ/snarkVM/pull/2424. The changes are:
- a `cfg_sort_unstable_by` macro for parallel sorting using a closure is introduced (note: the `unstable` part only means that **equal** elements may be reordered, and transaction IDs are never equal)
- the aforementioned macro is used to sort transactions when speculating
- the parallel iteration for the collection is removed, as it is slower than single-threaded iteration (since `Transaction::id` is basically free)
- instead of enumerating and collecting indices, [IndexSet::get_index_of](https://docs.rs/indexmap/latest/indexmap/set/struct.IndexSet.html#method.get_index_of) is used

### DIFF
--- a/synthesizer/src/vm/finalize.rs
+++ b/synthesizer/src/vm/finalize.rs
@@ -15,6 +15,7 @@
 use super::*;
 
 use ledger_committee::{MAX_DELEGATORS, MIN_DELEGATOR_STAKE, MIN_VALIDATOR_STAKE};
+use utilities::cfg_sort_unstable_by;
 
 impl<N: Network, C: ConsensusStorage<N>> VM<N, C> {
     /// Speculates on the given list of transactions in the VM.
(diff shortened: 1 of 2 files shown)


howardwu:
Is the assumption that sort_unstable is safe because the IDs are supposed to be unique? If so, this should be documented in the code itself for future readers to know.

ljedrz:
The unstable sort is safe, it only may reorder equal elements.

However, I realized that due to lookups, caching the key would be more meaningful here; I added a macro for sorting by cached key and adjusted this PR to use it.
I kept the macro for unstable sorting, as it may be useful in the future, for applications where they key extraction function is cheap.

vicsn (approved):

niklaslong (approved):

PR #2428, opened 2024-04-08
[Feature] Mitigate payment-free verification

raychu86:
## Motivation

This PR adds a few rules to mitigate the case where malicious parties can force validators to perform transaction verification without having to pay transaction fees (due to transaction aborts).

### Example cases: 
In both of these cases, the validators will perform verification on all transactions, but only the first transaction will be accepted. All others will be aborted.

1. Send multiple transactions with the same private fee record.
2. Send multiple transactions with the same public fee payer. However only have enough balance to pay for the first transaction.

### The mitigating factors in this PR are:

1. To perform duplicate/spent checks (input ID, output ID, tpks, etc) and abort colliding transactions prior to transaction verification in `VM::prepare_for_speculate`.
    - This check is done in `VM::atomic_speculate`, but now we also perform it prior to tx verification.
2. Add only process 1 deployment per block per public payer. 
    - Abort all subsequent deployments.

_Note that these checks are added prior to verification, so it may be aggressive in aborting transactions. For example given transactions A and B who have the same fee record. If A is aborted for a separate reason (for example, invalid finalize operation):_ 
_1. This PR will abort both A and B._
_2. The previous logic will only abort A and accept B._

### Possible future mitigations:

1. Possibly process deployment finalize before executions.
   - Deployments are more expensive to verify than executions, so processing them first will reduce wasted verification work.
3. Add slashing for malicious validators who inject aborted transactions to proposals.

## Test Plan

Tests have been added to ensure that there can only be 1 deployment per block per public fee payer. In addition, more checks have been added to existing tests to ensure that the aborted transactions do not incur verification costs.

## Note

This could break existing testnets, and will require a chain reset.


synthesizer/src/vm/finalize.rs
```diff
@@ -469,6 +435,10 @@ impl<N: Network, C: ConsensusStorage<N>> VM<N, C> {
                         output_ids.extend(confirmed_transaction.transaction().output_ids());
                         // Add the transition public keys to the set of produced transition public keys.
                         tpks.extend(confirmed_transaction.transaction().transition_public_keys());
+                        // Add the deployment payer to the set of deployment payers.
```
vicsn:
```suggestion
                        // Add any public deployment payer to the set of deployment payers.
```

synthesizer/src/vm/finalize.rs
```diff
@@ -788,6 +758,72 @@ impl<N: Network, C: ConsensusStorage<N>> VM<N, C> {
...
+        for tpk in transaction.transition_public_keys() {
+            // If the transition public key is already produced in this block or previous blocks, abort the transaction.
+            if tpks.contains(tpk) || self.transition_store().contains_tpk(tpk).unwrap_or(true) {
+                return Some(format!("Duplicate transition public key {tpk}"));
+            }
+        }
+
+        // If the transaction is a deployment, ensure that it is not another deployment in the block from the same public fee payer.
+        if let Transaction::Deploy(_, _, _, fee) = transaction {
+            // If the deployment spender has already deployed in this block, abort the transaction.
```
vicsn:
```suggestion
            // If any public deployment payer has already deployed in this block, abort the transaction.
```

synthesizer/src/vm/finalize.rs
```diff
@@ -798,13 +834,61 @@ impl<N: Network, C: ConsensusStorage<N>> VM<N, C> {
...
+                None => {
+                    // Add the transition IDs to the set of produced transition IDs.
+                    transition_ids.extend(transaction.transition_ids());
+                    // Add the input IDs to the set of spent input IDs.
+                    input_ids.extend(transaction.input_ids());
+                    // Add the output IDs to the set of produced output IDs.
+                    output_ids.extend(transaction.output_ids());
+                    // Add the transition public keys to the set of produced transition public keys.
+                    tpks.extend(transaction.transition_public_keys());
+                    // Add the deployment payer to the set of deployment payers.
```
vicsn:
```suggestion
                    // Add any public deployment payer to the set of deployment payers.
```

vicsn (approved):
Just some comment nits.

vicsn:
I'll also rerun consensus test 8.- Load-Batch-Proposal-Attack.

synthesizer/src/vm/finalize.rs
```diff
@@ -788,6 +758,72 @@ impl<N: Network, C: ConsensusStorage<N>> VM<N, C> {
...
 
+    /// Returns `Some(reason)` if the transaction is aborted. Otherwise, returns `None`.
+    ///
+    /// The transaction will be aborted if any of the following conditions are met:
+    /// - The transaction is producing a duplicate transition
+    /// - The transaction is double-spending an input
+    /// - The transaction is producing a duplicate output
+    /// - The transaction is producing a duplicate transition public key
+    /// - The transaction is another deployment in the block from the same public fee payer.
+    fn should_abort_transaction(
```
vicsn, line 770: As there are other reasons and locations where a transaction may be aborted, perhaps a more descriptive name would be: `check_transaction_uniqueness_in_block`

d0cd (approved):

howardwu:
@vicsn @onetrickwolf can you guys run this branch with CI validation and consensus burn-in / test criteria checks?

vicsn:
Long-running e2e test passed, consensus test 8 passed as well and clearly logs many `Speculation safely aborted a transaction - Another deployment in the block from the same public fee`.

howardwu (approved):

PR #2431, opened 2024-04-10
Introduce deployment variable limit and cost

vicsn:
## Motivation

Even though we limit program size and number of constraints, @d0cd identified that it is possible for someone to create huge *constants* in programs. During deployment verification, an attacker can make a validator take forever or use a lot of memory, without paying for it, because we don't limit or price constants.

The solution is to limit constants similarly to how we limit the number of constraints.

Some design considerations:
- I'm limiting total number of variables, not just constants, as defense in depth. Our lack of variable limitation was nagging me anyway. Speed to synthesize is the same for constants v.s. variables (for their respective worst case opcode keccak v.s. psd)
- I'm pricing in variables at the same rate as constraints, to avoid DoS from small programs with lots of constants.
- credits.aleo uses ~150k variables, the biggest popular program 300k variables, so (1 << 20) should be a sufficient limit
- verification of a deployment with max constants of (1 << 20) takes ~2 seconds in debug mode, so likely 0.02-0.2 second in release).
- we cannot add num_constants to CircuitInfo, constants are not a known quantity at the Varuna level. That's why I added the value to the `Deployment` object. Given that we price deployments by their size, the cost for deployments is increased by 0.01 aleo credit per function. Technically, we could only track the constants and not the total number of variables in the deployment, but I fear its less understandable for outside developers.

## Test Plan

Adding two tests, one testing the variable limit, and one testing manipulation of reported variables.

## Related PRs

Similar to: https://github.com/AleoHQ/snarkVM/pull/2271

### DIFF
--- a/console/network/src/lib.rs
+++ b/console/network/src/lib.rs
@@ -116,7 +116,7 @@ pub trait Network:
     /// The cost in microcredits per byte for the deployment transaction.
     const DEPLOYMENT_FEE_MULTIPLIER: u64 = 1_000; // 1 millicredit per byte
     /// The cost in microcredits per constraint for the deployment transaction.
-    const SYNTHESIS_FEE_MULTIPLIER: u64 = 25; // 25 microcredits per constraint
+    const SYNTHESIS_FEE_MULTIPLIER: u64 = 12; // 12 microcredits per constraint
     /// The maximum number of constraints in a deployment.
     const MAX_DEPLOYMENT_CONSTRAINTS: u64 = 1 << 20; // 1,048,576 constraints
     /// The maximum number of variables in a deployment.
--- a/console/network/src/lib.rs
+++ b/console/network/src/lib.rs
@@ -120,7 +120,7 @@ pub trait Network:
     /// The maximum number of constraints in a deployment.
     const MAX_DEPLOYMENT_CONSTRAINTS: u64 = 1 << 20; // 1,048,576 constraints
     /// The maximum number of variables in a deployment.
-    const MAX_DEPLOYMENT_VARIABLES: u64 = 1 << 20; // 1,048,576 variables
+    const MAX_DEPLOYMENT_VARIABLES: u64 = 1 << 22; // 4,194,304 variables
     /// The maximum number of microcredits that can be spent as a fee.
     const MAX_FEE: u64 = 1_000_000_000_000_000;
     /// The maximum number of microcredits that can be spent on a finalize block.
--- a/console/network/src/lib.rs
+++ b/console/network/src/lib.rs
@@ -118,7 +118,9 @@ pub trait Network:
     /// The cost in microcredits per constraint for the deployment transaction.
     const SYNTHESIS_FEE_MULTIPLIER: u64 = 25; // 25 microcredits per constraint
     /// The maximum number of constraints in a deployment.
-    const MAX_DEPLOYMENT_LIMIT: u64 = 1 << 20; // 1,048,576 constraints
+    const MAX_DEPLOYMENT_CONSTRAINTS: u64 = 1 << 20; // 1,048,576 constraints
+    /// The maximum number of variables in a deployment.
+    const MAX_DEPLOYMENT_VARIABLES: u64 = 1 << 20; // 1,048,576 variables
     /// The maximum number of microcredits that can be spent as a fee.
     const MAX_FEE: u64 = 1_000_000_000_000_000;
     /// The maximum number of microcredits that can be spent on a finalize block.
--- a/console/network/src/lib.rs
+++ b/console/network/src/lib.rs
@@ -120,7 +120,7 @@ pub trait Network:
     /// The maximum number of constraints in a deployment.
     const MAX_DEPLOYMENT_CONSTRAINTS: u64 = 1 << 20; // 1,048,576 constraints
     /// The maximum number of variables in a deployment.
-    const MAX_DEPLOYMENT_VARIABLES: u64 = 1 << 22; // 4,194,304 variables
+    const MAX_DEPLOYMENT_VARIABLES: u64 = 1 << 20; // 1,048,576 variables
     /// The maximum number of microcredits that can be spent as a fee.
     const MAX_FEE: u64 = 1_000_000_000_000_000;
     /// The maximum number of microcredits that can be spent on a finalize block.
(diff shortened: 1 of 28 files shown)


howardwu:
You'll need to reflect `TestnetCircuit` changes and `TestnetV0` changes now as well.

synthesizer/process/src/stack/helpers/synthesize.rs
```diff
@@ -100,6 +103,8 @@ impl<N: Network> Stack<N> {
         // Insert the proving key.
         self.insert_proving_key(function_name, proving_key)?;
         // Insert the verifying key.
-        self.insert_verifying_key(function_name, verifying_key)
+        self.insert_verifying_key(function_name, verifying_key)?;
```
ljedrz, line 106: should we remove the proving key if the insertion of the verifying key fails (`insert_verifying_key` is fallible now)?

console/network/src/lib.rs
```diff
@@ -118,7 +118,9 @@ pub trait Network:
     /// The cost in microcredits per constraint for the deployment transaction.
     const SYNTHESIS_FEE_MULTIPLIER: u64 = 25; // 25 microcredits per constraint
     /// The maximum number of constraints in a deployment.
-    const MAX_DEPLOYMENT_LIMIT: u64 = 1 << 20; // 1,048,576 constraints
+    const MAX_DEPLOYMENT_CONSTRAINTS: u64 = 1 << 20; // 1,048,576 constraints
```
ljedrz, line 121: if the limit is `1 << 20`, is it reasonable to assume that we would never expect `1 << 64` to be supported? `u32` might be a better pick in that case, and IIRC that could have a performance impact on some of the R1CS operations (if the `Variable` was also capped at `1 << 32`)

ledger/store/src/transaction/deployment.rs
```diff
@@ -49,6 +49,8 @@ pub trait DeploymentStorage<N: Network>: Clone + Send + Sync {
     type VerifyingKeyMap: for<'a> Map<'a, (ProgramID<N>, Identifier<N>, u16), VerifyingKey<N>>;
     /// The mapping of `(program ID, function name, edition)` to `certificate`.
     type CertificateMap: for<'a> Map<'a, (ProgramID<N>, Identifier<N>, u16), Certificate<N>>;
+    /// The mapping of `(program ID, function name, edition)` to `variable count`.
+    type VariableCountMap: for<'a> Map<'a, (ProgramID<N>, Identifier<N>, u16), u64>;
```
ljedrz, line 53: this is not strictly related to this PR, but I wonder if, in case of several maps containing composite/large keys, it would make sense to have composite values too (e.g. a single map of `(ProgramID<N>, Identifier<N>, u16)` to `(VerifyingKey<N>, Certificate<N>, u64)`), reducing the number of queries and storage space dedicated to keys

synthesizer/process/src/stack/helpers/synthesize.rs
vicsn: `insert_verifying_key` was always fallible, we just don't return the Result yet because we have another function to check.

Indeed its not super clean that this doesn't happen atomically, but the only reason for failure at the moment is that the pk/vk/count already exists.

So I think we can leave it as is.

console/network/src/lib.rs
vicsn: Constraints won't exceed u32 anytime soon, especially as our SRS only supports `1 << 28`. Perhaps constant variables might end up being more. The `circuit` crate stores most data as u32, and its only a small amount of counters anyway, so at this point its not worth the rewrite I think.

ledger/store/src/transaction/deployment.rs
vicsn: Indeed that sounds like an area worthy of optimization

iamalwaysuncomfortable (commented):
I understand the defense-in-depth argument, however is putting a variable limit maybe going too far.

It's said that `300k` _should be sufficient_ but what's the basis of that assumption of sufficiency? 

If there are programs which need more variables to more innocent arithmetic operations, this prevents them from operating - this problem will be especially acute for people writing programs in Leo as it tends to lead to many more variables than programs written natively in Aleo instructions. 

My main concern with this is that how the circuit creates variables under the hood is effectively opaque to most users (minus those who are familiar with the circuits) and will create a bad UX as users pull out their hair trying to find out why their deployments are failing.

vicsn:
Good questions.

> It's said that 300k should be sufficient but what's the basis of that assumption of sufficiency?

The limit is 1M, same as the constraint limit, because synthesizing variables and constraints for our existing gadgets go hand in hand. If you grep in snarkVM for `Count::` you'll see all of our existing gadgets allocate roughly the same amount of variables as constraints (actually a bit less variables, because multiple constraints typically link the same variable). Compared to before, the only usage which this design limits is someone trying to generate large constants.

> My main concern with this is that how the circuit creates variables under the hood is effectively opaque to most users (minus those who are familiar with the circuits) and will create a bad UX as users pull out their hair trying to find out why their deployments are failing.

This was also the case for the constraint limit. If you have alternative design ideas let me know.

I agree we should report the reason of failure: https://github.com/AleoHQ/snarkOS/issues/2962

howardwu:
The design should have variable count in the verifying key. It is error prone to have it defined as an accessory, like in this PR.

AFAIK, the `insert_variable_count` is NOT called everywhere `insert_verifying_key` is called.

It would be helpful to know why this design was chosen, and whether it was validated on a snarkOS devnet.

synthesizer/process/src/stack/execute.rs
```diff
@@ -144,13 +144,16 @@ impl<N: Network> StackExecute<N> for Stack<N> {
     ) -> Result<Response<N>> {
         let timer = timer!("Stack::execute_function");
 
+        // Ensure constant circuit gadgets are initialized.
+        A::init_constants();
```
howardwu, line 148: Why do we need to `init_constants` on everything?

vicsn:
> It would be helpful to know why this design was chosen, and whether it was validated on a snarkOS devnet.

I did consider it. From this PR's README: "we cannot add num_constants to CircuitInfo, constants are not a known quantity at the Varuna level. That's why I added the value to the Deployment object." Adding more context:
- The `CircuitVerifyingKey` is from the algorithms crate, so it is also semantically not clean to add constants in there which are not relevant in that crate. I.e. we'd have to set them to 0 or None for any tests local to the Varuna crate.
- Another option is the `VerifyingKey` in the synthesizer crate. I didn't want to mess with that abstraction either, as it is used as a thin wrapper.

But go wild if you want to do it anyway.

> whether it was validated on a snarkOS devnet.

Not yet, otherwise I would/should have mentioned in the PR README or one of the comments.

> AFAIK, the insert_variable_count is NOT called everywhere insert_verifying_key is called.

I see, that looks like a potential issue. A simpler way around could be to unify just those 2 functions

howardwu:
Thanks for the insight. We're currently discussing using https://github.com/AleoHQ/snarkVM/pull/2444 as the replacement PR.

howardwu:
Closing in favor of #2444

PR #2434, opened 2024-04-16
Add support to fetch delegators for given validator

vicsn:
## Motivation

Adds support to fetch all delegators for a particular validator.

## Test Plan

- [x] Tested manually when bonding hundreds of delegators.
- [x] Moreover, @iamalwaysuncomfortable is working on [related benchmarks](https://github.com/AleoHQ/snarkVM/pull/2436).

## Related PRs

https://github.com/AleoHQ/snarkOS/pull/3219


iamalwaysuncomfortable (approved):
According the benchmarks, accessing the bonded mapping using `get_value_confirmed` is roughly the same between 10 entries and 100k entries thus I think this is acceptable to include at this level of the protocol.

The more operative piece to benchmark would be the performance of the actual REST endpoint itself as it serializes 100k delegators into a `json` response.

howardwu (approved):

PR #2435, opened 2024-04-17
Nit: more consistent tx verification 

vicsn:
## Motivation

I noticed a small discrepancy after Ray remarked tpk checks cover tcm/scm checks.

## Test Plan

No functionality is altered.


howardwu:
Because we do not know for certain that Poseidon is collision-resistant, Pratyush and I made the decision to include the `tcm` uniqueness check out of an abundance of caution.

Closing this PR.

PR #2437, opened 2024-04-17
[Feature] Introduces a maximum transaction size.

d0cd:
This PR introduces `MAX_TRANSACTION_SIZE` which upper bounds the number of bytes in a transaction. 
This limit is imposed in `check_transaction`, however constructing larger transactions is still permitted.

As part of the design, this PR introduces a `LimitedWriter` which is a wrapper aroound `Write`. The motivation is to ensure that additional bytes (beyond the maximum) are not written. Without this, the implementation in `check_transaction` have to write potentially larger amounts of data, before being able to check that the limit is not exceeded. 

Note that this implementation requires that `MAX_TRANSACTION_SIZE` is never decreased.

An alternate implementation was considered where `MAX_TRANSACTION_SIZE` was enforced in the (de)serializers. That design is prone to backwards compatibility issues. Note that this design is not free of backwards compatibility issues either; if `MAX_TRANSACTION_SIZE` is decreased, then older transactions can still be invalidated.


ledger/block/src/transaction/bytes.rs
```diff
@@ -17,7 +17,9 @@ use super::*;
 impl<N: Network> FromBytes for Transaction<N> {
     /// Reads the transaction from the buffer.
     #[inline]
-    fn read_le<R: Read>(mut reader: R) -> IoResult<Self> {
+    fn read_le<R: Read>(reader: R) -> IoResult<Self> {
+        // Restrict the reader to `MAX_TRANSACTION_SIZE` bytes.
+        let mut reader = reader.take(N::MAX_TRANSACTION_SIZE as u64);
```
iamalwaysuncomfortable: Two comments here:
- Is it entirely safe to simply truncate and proceed as normal? It feels like we shouldn't limit reading - because what if the transaction size limit is lowered in the future, how would we read transactions that are above the size limit that are already in the ledger?

utilities/src/bytes.rs
```diff
@@ -478,6 +478,40 @@ pub fn bytes_from_bits_le(bits: &[bool]) -> Vec<u8> {
...
+            }
+            Err(e) => Err(e),
+        }
+    }
+
+    fn flush(&mut self) -> IoResult<()> {
+        self.writer.flush()
+    }
+}
+
```
iamalwaysuncomfortable: This limits the amount of bytes serialized, but that maybe we shouldn't be using it in the fundamental serialization code?

One one hand it stops people from generating transactions above the `MAX_TRANSACTION_SIZE` limit, but the devil in the details might be that if someone a node database gets an otherwise valid transaction above the size limit in the database, nodes stop being able to read their transaction stores. The potential unintended consequences here might be larger than we should risk.

However, this `Writer` DOES prevent the DoS vector of people sending large transactions that nodes constantly have to call `write_le` on. Maybe the better path is to use this serializer within a transaction `size` or `check` method on the `Transaction` struct that checks the size up to 128kb and if it's above, throws a `bail!` - and this method can be used in the checks within `synthesizer/src/vm/verify.rs`

That way we get both the benefit of limiting bytes without risking future dragons in core byte serialization.

synthesizer/src/vm/verify.rs
```diff
@@ -76,6 +76,11 @@ impl<N: Network, C: ConsensusStorage<N>> VM<N, C> {
 
         /* Transaction */
 
+        // Ensure that the transaction is well formed and does not exceed the maximum size.
+        if let Err(error) = transaction.to_bytes_le() {
+            bail!("Transaction '{}' is not well-formed: {error}", transaction.id())
+        }
+
```
iamalwaysuncomfortable: This check should maybe also be in `atomic_speculate` as well to be extra sure?

iamalwaysuncomfortable (commented):
Unsure about the approach of limiting Transaction serialization itself to 128kb. But I think the `Writer` you developed can be used to create a check that only checks up to 128kb instead of serializing the entire transaction to bytes and then checking.

ledger/block/src/transaction/bytes.rs
howardwu: I echo @iamalwaysuncomfortable. I don't know if using `take` and truncating the bytestream is the best approach. However, I'm also not an expert on the implications.

Could @joske or @niklaslong chime in here? The intention is to limit transactions to 128kb in consensus, in order to ensure transactions exceeding this size don't land in a block. At the same time, we also want to minimize any risk of breakages to any other functionality in our codebase.

The priority is on safety over fast or simple results here.

niklaslong: I agree with the above, we _could_ limit reading and keep the same constant even if we further restrict tx size in future but I'd be in favour of avoiding this if possible. 

One consideration is DoS vectors in snarkOS, however we already limit inbound message sizes in the codecs so we should be covered there. That means we're left with the problem of detecting overly large transactions that are _smaller_ than the network's max message size as early as possible and I think that would be safe to check after deserialisation.

Is there another path in transaction processing or snarkVM use that wouldn't be covered by this?

niklaslong: Just for completeness and in order to avoid decoding overly large transactions in snarkOS, we could adjust the codecs to:
1. register the length of the raw bytes from the network
2. read the first two bytes containing the message ID
3. if it is a tx and longer than the max, we bail, if not we decode the rest

synthesizer/src/vm/verify.rs
d0cd: `check_transaction` is called before either instance of `atomic_speculate`. Is the redundancy necessary?

ledger/block/src/transaction/bytes.rs
d0cd: > Just for completeness and in order to avoid decoding overly large transactions in snarkOS, we could adjust the codecs to:
> 
>     1. register the length of the raw bytes from the network
> 
>     2. read the first two bytes containing the message ID
> 
>     3. if it is a tx and longer than the max, we bail, if not we decode the rest

I'd be in favor of this approach as the `MAX_EVENT_SIZE` is 256 MiB, while the `MAX_TRANSACTION_SIZE` is 128 kB.

utilities/src/bytes.rs
d0cd: Calling out another issue. If the transaction limit is ever lowered (through a network update), we need to ensure that the appropriate size limit is used for each transaction in `check_transaction`. However, `Transaction` objects do not contain a version.

ledger/block/src/transaction/bytes.rs
d0cd: By limiting it in the codec, does it give us the property: "honest validators will never sign/propose a proposal that contains transactions exceeding the transaction byte limit"?

niklaslong: We could restrict (outbound) serialisation in the same way but I think this should be a property enforced in snarkVM? i.e. an honest validator can't produce a tx that is too large in the first place? Or is it impossible to check for before serialisation?

niklaslong: Ah, I misunderstood the question. When it comes to _signing_ other proposals that might have large transactions in them, the assumption is that we've seen and stored the txs so restricting inbound size should already account for that. However, we might need to add a little nuance to the logic and error handling 🤔

niklaslong: To add to the above, we check that transactions are well formed in the proposal so we could maybe check for length there (we also deserialise txs at that stage).

niklaslong (approved):

utilities/src/bytes.rs
```diff
@@ -478,6 +478,40 @@ pub fn bytes_from_bits_le(bits: &[bool]) -> Vec<u8> {
...
+impl<W: Write> LimitedWriter<W> {
+    pub fn new(writer: W, limit: usize) -> Self {
+        Self { writer, limit, remaining: limit }
+    }
+}
+
+impl<W: Write> Write for LimitedWriter<W> {
+    fn write(&mut self, buf: &[u8]) -> IoResult<usize> {
+        if self.remaining == 0 {
+            return Err(std::io::Error::new(std::io::ErrorKind::Other, format!("Byte limit exceeded: {}", self.limit)));
```
niklaslong, line 497: Just mentioning that https://doc.rust-lang.org/std/io/enum.ErrorKind.html#variant.WriteZero exists, it might be more appropriate here?

d0cd: I wanted the error message to display the byte limit. If we don't need that information then I agree that `WriteZero` is the way to go.

PR #2439, opened 2024-04-22
[Perf] Don't recreate ReadOptions when querying the database

ljedrz:
Heap profiling indicates that plenty of allocations currently come from the use of `rocksdb_readoptions_create`, which is an FFI function called internally from `get_pinned`. This can easily be avoided by creating an instance of `ReadOptions` just once and reusing it via `get_pinned_opt`, which is the equivalent of `get_pinned` with a `&ReadOptions` parameter.

This change reduces (in a 15-minute run consisting of 4 `--dev` validators) the average number of allocations per second in a `dev` node that doesn't produce txs by roughly 30%, which is beneficial for the performance of any heap allocator.

### DIFF
--- a/ledger/store/src/helpers/rocksdb/internal/mod.rs
+++ b/ledger/store/src/helpers/rocksdb/internal/mod.rs
@@ -75,7 +75,6 @@ pub trait Database {
 }
 
 /// An instance of a RocksDB database.
-#[derive(Clone)]
 pub struct RocksDB {
     /// The RocksDB instance.
     rocksdb: Arc<rocksdb::DB>,
@@ -91,6 +90,22 @@ pub struct RocksDB {
     pub(super) atomic_depth: Arc<AtomicUsize>,
     /// A flag indicating whether the atomic writes are currently paused.
     pub(super) atomic_writes_paused: Arc<AtomicBool>,
+    /// This is an optimization that avoids some allocations when querying the database.
+    pub(super) default_readopts: rocksdb::ReadOptions,
+}
+
+impl Clone for RocksDB {
+    fn clone(&self) -> Self {
+        Self {
+            rocksdb: self.rocksdb.clone(),
+            network_id: self.network_id,
+            storage_mode: self.storage_mode.clone(),
+            atomic_batch: self.atomic_batch.clone(),
+            atomic_depth: self.atomic_depth.clone(),
+            atomic_writes_paused: self.atomic_writes_paused.clone(),
+            default_readopts: Default::default(),
+        }
+    }
 }
 
 impl Deref for RocksDB {
@@ -136,6 +151,7 @@ impl Database for RocksDB {
                     atomic_batch: Default::default(),
                     atomic_depth: Default::default(),
                     atomic_writes_paused: Default::default(),
+                    default_readopts: Default::default(),
                 })
             })?
             .clone();
@@ -309,6 +325,7 @@ impl RocksDB {
                 atomic_batch: Default::default(),
                 atomic_depth: Default::default(),
                 atomic_writes_paused: Default::default(),
+                default_readopts: Default::default(),
             })
         }?;
 
--- a/ledger/store/src/helpers/rocksdb/internal/map.rs
+++ b/ledger/store/src/helpers/rocksdb/internal/map.rs
@@ -535,7 +535,7 @@ impl<K: Serialize + DeserializeOwned, V: Serialize + DeserializeOwned> DataMap<K
         Q: Serialize + ?Sized,
     {
         let raw_key = self.create_prefixed_key(key)?;
-        match self.database.get_pinned(&raw_key)? {
+        match self.database.get_pinned_opt(&raw_key, &self.database.default_readopts)? {
             Some(data) => Ok(Some(data)),
             None => Ok(None),
         }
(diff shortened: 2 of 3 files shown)


ledger/store/src/helpers/rocksdb/internal/mod.rs
```diff
@@ -91,6 +90,22 @@ pub struct RocksDB {
     pub(super) atomic_depth: Arc<AtomicUsize>,
     /// A flag indicating whether the atomic writes are currently paused.
     pub(super) atomic_writes_paused: Arc<AtomicBool>,
+    /// This is an optimization that avoids some allocations when querying the database.
+    pub(super) default_readopts: rocksdb::ReadOptions,
+}
+
+impl Clone for RocksDB {
```
ljedrz, line 97: note: this is needed because `ReadOptions` doesn't implement `Clone`

ledger/store/src/helpers/rocksdb/internal/mod.rs
```diff
@@ -91,6 +90,22 @@ pub struct RocksDB {
     pub(super) atomic_depth: Arc<AtomicUsize>,
     /// A flag indicating whether the atomic writes are currently paused.
     pub(super) atomic_writes_paused: Arc<AtomicBool>,
+    /// This is an optimization that avoids some allocations when querying the database.
+    pub(super) default_readopts: rocksdb::ReadOptions,
```
iamalwaysuncomfortable, line 94: Under the hood this holds a foreign pointer which gets passed `get_pinned_opt`. Even though it doesn't get manipulated at all on the Rust side, is there certainty it would remain in the expected state for the lifetime of the Rocks object via the other side of the ffi?

My read of the source of [get_pinned_opt](https://docs.rs/rocksdb/latest/src/rocksdb/db.rs.html#991) is that if the pointer ends up null this would return an `Err` or potentially `panic` within the `unsafe` block of this function if the foreign pointer in `ReadOptions` ends up in an invalid state. However I'm not an expert on RocksDB! So the library may provide the proper safety guarantees for this.

iamalwaysuncomfortable (commented):
> the average number of allocations per second in a dev node that doesn't produce txs by roughly 30%

The number posted in this PR are attractive. However to clarify, under what conditions do you see the reduction? Is it just normal operation of a node minus any load?

Would it be possible to profile this under heavy read load?

ljedrz:
> under what conditions do you see the reduction? Is it just normal operation of a node minus any load?

The difference was measured for a `--dev 2` node which doesn't produce txs, but participates in consensus.

> Would it be possible to profile this under heavy read load?

I requested a devnet run to see the impact with more transactions.

ledger/store/src/helpers/rocksdb/internal/mod.rs
ljedrz: While it would be quite weird if that were to happen, it's a valid concern; I've traced [the C code](https://github.com/facebook/rocksdb/blob/3f3045a40511d0e15a63b134e0c92ace49983fc1/db/c.cc#L6628-L6642) that uses the `ReadOptions`, though, and it doesn't seem to me that the object gets modified in any way, and the underlying pointer won't go away until a call to `rocksdb_readoptions_destroy` is made (which happens on `Drop`).

iamalwaysuncomfortable: Thanks for tracing that down.

iamalwaysuncomfortable (commented):
> I requested a devnet run to see the impact with more transactions.

This can be done, but prior to that, to quantify the impact what are the measures we need + what tools do you use to get them?

kpandl:
Performed some memory profiling with ~2.5 hours of 25 TPS [here](https://github.com/AleoHQ/stress-observability/issues/115#issuecomment-2072450510).
After the test, snarkOS maximum memory usage was reduced by 8% and average memory usage was reduced by 4.4% through this PR (maximum/average of all 10 nodes).

iamalwaysuncomfortable:
A devnet test of this to profile has been kicked off

ljedrz:
Observations based on devnet metrics:
- memory use is a bit lower, but there is still growth
- log frequency is up, suggesting higher average throughput
- average block lag is down, and its spikes are much lower
- number of threads appears to be a bit down
- minor page faults are reduced slightly

Overall, this change results in some small performance wins, and its effects are more significant with workloads that involve many database reads.

howardwu:
I will await for approvals from reviewers prior to merging this one.

ledger/store/src/helpers/rocksdb/internal/mod.rs
ljedrz: @iamalwaysuncomfortable I've found a resource that states that `ReadOptions` may be reused freely: https://github.com/facebook/rocksdb/wiki/RocksDB-FAQ#basic-readwrite (the last Q in the linked section):
> Q: Can I reuse ReadOptions, WriteOptions, etc, across multiple threads?
> A: As long as they are const, you are free to reuse them.

ljedrz:
Update: the same can be done with `WriteOptions`; the win is smaller, as we write less frequently.

howardwu:
@joske can you also review this PR?

niklaslong (approved):
LGTM 👍

iamalwaysuncomfortable (approved):

PR #2440, opened 2024-04-22
[TOB] Fix length in `Serialize` impl for `Transition`

d0cd:
## Motivation

This PR fixes the `Serialize` implementation for `Transition`s by correcting the number of struct fields.

Finding: TOB-PUZZLE-3


raychu86 (approved):
LGTM. 

In the future, we may want to integrate `semgrep` checks to CI to prevent future instances of this case.

howardwu (approved):

PR #2444, opened 2024-04-29
Adds a variable limit for deployments

howardwu:
## Motivation

Note: This PR contains breaking changes. First, all verifying keys are serialized with 8 extra bytes to account for the `num_variables`, which represents the number of *constant*, public, and private variables in the circuit. In addition, the `DataID` was updated to address an open `TODO` which involves reordering the map IDs (and also a breaking change).

Original motivation from @vicsn :

> Even though we limit program size and number of constraints, @d0cd identified that it is possible for someone to create huge constants in programs. During deployment verification, an attacker can make a validator take forever or use a lot of memory, without paying for it, because we don't limit or price constants.
>
> The solution is to limit constants similarly to how we limit the number of constraints.
>
> Some design considerations:
>
> - I'm limiting total number of variables, not just constants, as defense in depth. Our lack of variable limitation was nagging me anyway. Speed to synthesize is the same for constants v.s. variables (for their respective worst case opcode keccak v.s. psd)
> - I'm pricing in variables at the same rate as constraints, to avoid DoS from small programs with lots of constants.
> - credits.aleo uses ~150k variables, the biggest popular program 300k variables, so (1 << 20) should be a sufficient limit
> - verification of a deployment with max constants of (1 << 20) takes ~2 seconds in debug mode, so likely 0.02-0.2 second in release).
> - we cannot add num_constants to CircuitInfo, constants are not a known quantity at the Varuna level. That's why I added the value to the Deployment object. Given that we price deployments by their size, the cost for deployments is increased by 0.01 aleo credit per function. Technically, we could only track the constants and not the total number of variables in the deployment, but I fear its less understandable for outside developers.

## Test Plan

Adding two tests, one testing the variable limit, and one testing manipulation of reported variables.

## Related PRs

Supersedes #2431.

Similar to: https://github.com/AleoHQ/snarkVM/pull/2271.

### DIFF
--- a/algorithms/src/snark/varuna/ahp/ahp.rs
+++ b/algorithms/src/snark/varuna/ahp/ahp.rs
@@ -107,7 +107,7 @@ impl<F: PrimeField, SM: SNARKMode> AHPForR1CS<F, SM> {
 
     /// Get all the strict degree bounds enforced in the AHP.
     pub fn get_degree_bounds(info: &CircuitInfo) -> Result<[usize; 4]> {
-        let num_variables = info.num_variables;
+        let num_variables = info.num_public_and_private_variables;
         let num_non_zero_a = info.num_non_zero_a;
         let num_non_zero_b = info.num_non_zero_b;
         let num_non_zero_c = info.num_non_zero_c;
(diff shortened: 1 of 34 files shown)


raychu86 (approved):
LGTM

PR #2450, opened 2024-05-09
Introduce cfg_sorted_by

ljedrz:
As requested in https://github.com/AleoHQ/snarkOS/issues/3251, and it works in the related scenario ([preview branch](https://github.com/ljedrz/snarkOS/tree/perf/faster_get_pending_certs)).

Note: there is also [par_sorted_unstable_by](https://docs.rs/indexmap/latest/indexmap/set/struct.IndexSet.html#method.par_sorted_unstable_by), which would likely be faster; happy to add/substitute if desirable.

### DIFF
--- a/utilities/src/parallel.rs
+++ b/utilities/src/parallel.rs
@@ -292,3 +292,19 @@ macro_rules! cfg_sort_by_cached_key {
         $self.par_sort_by_cached_key($closure);
     }};
 }
+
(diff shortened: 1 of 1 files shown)


howardwu:
Can you include any perf improvements for code changes from https://github.com/AleoHQ/snarkOS/pull/3247/files?

howardwu (approved):

PR #2452, opened 2024-05-13
[HackerOne-2498849] Abort fee earlier in prepare_for_speculate

vicsn:
## Motivation

Does not close but partially mitigates: https://github.com/AleoHQ/snarkVM/issues/2451

A malicious validator can generate an execution with mostly the same transitions, which will be aborted causing the victim's transaction to be filtered out and dropped.

## Test Plan

- [x] e2e test succeeded which starts a devnet, syncs, 10 TPS, bonding (no deployments yet).
- [x] Unit tests corrected.

## Related PRs

This was introduced by: https://github.com/AleoHQ/snarkVM/pull/2428

### DIFF
--- a/synthesizer/src/vm/finalize.rs
+++ b/synthesizer/src/vm/finalize.rs
@@ -855,6 +855,12 @@ impl<N: Network, C: ConsensusStorage<N>> VM<N, C> {
         // Abort the transactions that are have duplicates or are invalid. This will prevent the VM from performing
         // verification on transactions that would have been aborted in `VM::atomic_speculate`.
         for transaction in transactions.iter() {
+            // Abort the transaction early if it is a fee transaction.
+            if transaction.is_fee() {
+                aborted_transactions.push((*transaction, "Fee transactions are not allowed in speculate".to_string()));
+                continue;
+            }
+
             // Determine if the transaction should be aborted.
             match self.should_abort_transaction(
                 transaction,
@@ -900,14 +906,6 @@ impl<N: Network, C: ConsensusStorage<N>> VM<N, C> {
             // Verify the transactions and collect the error message if there is one.
             let (valid, invalid): (Vec<_>, Vec<_>) =
                 cfg_into_iter!(transactions).zip(rngs).partition_map(|(transaction, mut rng)| {
-                    // Abort the transaction if it is a fee transaction.
-                    if transaction.is_fee() {
-                        return Either::Right((
-                            *transaction,
-                            "Fee transactions are not allowed in speculate".to_string(),
-                        ));
-                    }
-
                     // Verify the transaction.
                     match self.check_transaction(transaction, None, &mut rng) {
                         // If the transaction is valid, add it to the list of valid transactions.
--- a/synthesizer/src/vm/finalize.rs
+++ b/synthesizer/src/vm/finalize.rs
@@ -835,9 +835,11 @@ impl<N: Network, C: ConsensusStorage<N>> VM<N, C> {
         transactions: &[&'a Transaction<N>],
         rng: &mut R,
     ) -> Result<(Vec<&'a Transaction<N>>, Vec<(&'a Transaction<N>, String)>)> {
-        // Construct the list of transactions that need to verified.
-        let mut transactions_to_verify = Vec::with_capacity(transactions.len());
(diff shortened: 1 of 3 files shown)


synthesizer/src/vm/finalize.rs
```diff
@@ -852,10 +854,24 @@ impl<N: Network, C: ConsensusStorage<N>> VM<N, C> {
...
+        // VM from performing expensive verification on invalid fees or
+        // deployments that would have been aborted in `VM::atomic_speculate`.
         for transaction in transactions.iter() {
-            // Determine if the transaction should be aborted.
+            // If the transaction is an Execution, we will fully verify it.
+            // Executions require full verification to avoid malleability attacks.
+            if transaction.is_execute() {
+                executions.push(*transaction);
+                continue;
+            }
```
raychu86: @vicsn I don't believe we want this. This opens up the attack where malicious parties and induce load on validators by verifying many transactions for free.

To have duplicate `input_id`, `output_ids`, and `tcm` you need to use the same private key to create the transaction, so it would only affect their own transactions.

The malleability attack from validators can only be done via converting the Execution to a Fee transaction. Which your subsequent check already catches.

synthesizer/src/vm/finalize.rs
```diff
@@ -852,10 +854,24 @@ impl<N: Network, C: ConsensusStorage<N>> VM<N, C> {
...
+
+            // Abort the transaction if it is a fee transaction.
+            if transaction.is_fee() {
+                aborted_transactions.push((*transaction, "Fee transactions are not allowed in speculate".to_string()));
+                continue;
+            }
```
raychu86: I believe adding this check should be enough. The ~~only~~ malleability attack validators can do is to convert the Execution to a Fee transaction.

The `FakeTransaction` attack can ~~only~~ be done this way. Having actual `input_id`, `output_id`, and `tcm` duplicates can only be crafted with the same private key, so malicious validators can't simply inject a transition from another transaction into their own. 

@ghostant-1017 Let me know if this makes sense and if I am missing anything.

raychu86:
Note that this is a potential breaking change, that may affect the ongoing networks. 

The validators being run most likely won't be throwing around these `Fee` transactions in the honest case, however we need to be sure that nobody ran that as a test case during the operation of the network.

synthesizer/src/vm/finalize.rs
vicsn: My understanding is that the attack entails a validator creating a completely invalid bogus transaction by adding a random transition. 

I pushed a proof of concept to the tests: 8949c47d0 . At least at this level in the code, I confirmed that without this PR, the bogus transaction "pushes out" the honest transaction.

Sidenote: This also made me realise that malicious validators can still cause payment-free verification of their malformed executions, though these should be cheap. I'll think about a resolution.

raychu86: I stand corrected. It looks like the validators can also mutate the transactions by swapping out the fee transaction. If the fee is too low, the transaction will be aborted (if the fee is high enough, then the transaction will be rejected and the fee is consumed.


The validator might also be able to mutate the transaction fields with garbage s.t. verification will fail and the TX will be aborted.

vicsn: > The validator might also be able to mutate the transaction fields with garbage s.t. verification will fail and the TX will be aborted.

Good point, I think I also see an angle where the attacker can force a deployment to be aborted. The attacker mutates the deployment's only transition, which is the fee, to change its outputs but keep its inputs. Or vice versa. Therefore the transaction_id will be different but it still partially collides in the `should_abort_transaction` check. 

Despite this, I have a weak opinion to keep calling `should_abort_transaction` on deployments, as this PR does, as it is still an effective DoS protection measure.

In general, we can consider proper malleability protection measures like a cheap hmac or signature over the full transaction contents which *must* be the first thing verified. But I think that's a post-mainnet discussion.

raychu86: This "attack" is a bit hard to detect, but slashing is something we need in the future to mitigate misbehavior by validators.

vicsn: Reverted skipping the execution check, now only aborting Fee early and added a test for it

howardwu (approved):

PR #2354, opened 2024-02-14
[Coinbase Go/No-Go] Introduce validator commission rates 

raychu86:
## Motivation

This PR implements commissioned stake payouts for stakers in accordance to [ARC-038](https://github.com/AleoHQ/ARCs/discussions/52).

Validators can now set a commission `rate` that will provide commission payouts to the validator based on the amount of stake delegated to them. The delegators will now have a commission (based on the `rate`) paid to the validator from each block's staking reward.

This rate is a percentage value between 0 and 100. Where 0 means no commissioned payouts to the validators and 100 means the entire staking reward is paid as commission to the validator.

## Test Plan

Tests have been added to ensure that:
- A commission rate of 0 equates to a full stake reward payout for the delegator
- A commission rate of 100 equates to no stake reward for the delegator
- A higher commission rate (compared to a lower commission rate) equates to less stake reward for the delegator

PR #2453, opened 2024-05-15
ARC-0041: Update credits program to support delegation before validator bonding and native commission.

evanmarshall:
## Summary

Currently, validators will need to have 10M credits in order to self.bond as a validator. The problem is that most or all of these validators won't actually have 10M credits. To bring Aleo in line with other ecosystems, we want to relax 2 existing constraints:

1. Delegators can delegate to any address (with the same 10,000 credit minimum & maximum of 100,000 delegators) even if that address is not in the committee
2. Validators only require a self bond of 100 credits to join the committee, given the total delegation for them is >= 10M credits.

Although the changes are significant, the staking abstraction should closely follow what existed before.

## Technical summary

The invariants of the `credits.aleo` program are now as follows:
```
// 1. contains committee[r0] =>
//   a. contains delegated[r0]
//      - The set of delegated addresses is superset of active validators.
// 2. delegated[*] >= 10,000
//   a. contain bonded[*].microcredits >= 10,000
//      - Delegators can bond to any address with the 10,000 credit minimum
// 3. delegated[r0] < 10,000,000 =>
//   a. !contain committee[r0]
//      - If the delegated total (which includes the self bond) is less than 10,000,000 credits, it won't be in the committee
// 4. bond[r0].validator == r0 =>
//   a. contains committee[r0]
//     - Self bonded addresses are always active validators. Delegators **can** force validators to unbond by removing sufficient delegation
// 5. committee[r0] =>
//   a. bonded[r0].microcredits >= 100 credits
//      - Self bonded addresses are always active validators with at least 100 credits self bonded
//   b. delegated[r0] >= 10,000,000 credits
//      - The total delegated amount for an active validator is at least 10,000,000 credits
// 6. delegated[r0] == sum bond[*].microcredits by bond[].validator == r0
// 7. bond[r0].microcredits < 10,000 =>
//   a. bond[r0].validator == r0
//   b. contains committee[r0]
//   - A bond value < 10,000 implies self-bonded active validator.
// 8. bond[r0].validator != r0 =>
//   a. bond[r0].microcredits >= 10,000
// 9. count bonded[] - count committee[] == metadata[aleo1qgqqqqqqqqqqqqqqqqqqqqqqqqqqqqqqqqqqqqqqqqqqqqqqqqqqanmpl0]
// 10. count committee[] == metadata[aleo1qqqqqqqqqqqqqqqqqqqqqqqqqqqqqqqqqqqqqqqqqqqqqqqqqqqq3ljyzc]
// 11. 100,000 + count committee[] >= count bonded[] >= count delegated[] >= count committee[]
// 12. bond[*].microcredits > 100
// 13. delegated[*] > 10,000
```

### API Changes

1. There is now a `bond_validator` function that should be used to add a validator to the committee instead of the `bond_public` that was used before. This separation function was added due to the change in interface to support specifying a commission.
2. The `withdrawal address` is now the address that needs to be used for `unbond_public` which applies to all stakers. There are 2 primary reasons for this change:
    * Security. Previously the validator hot key was used to `unbond_public` and `claim_unbond_public` while funds ended up at the address specified in the `withdraw` mapping. This meant that it was safe to lose exclusive access to the hot key but stakers still need to retain access to it to unlock their funds. 
    * Programmability. We anticipate the need for compliance features, commission structures and validator pools. To broadly enable new use cases, we need a key that can sign certificates in BFT ie the validator key and an address that can have programmatically control staked funds. The validator key will always need to be a hot private key but the withdraw address can be a program so we're giving the withdraw address sufficient authority to manage staked funds.
3. `claim_unbond_public` now requires an address for claiming. `claim_unbond_public` previously used `self.caller` for the address. The result of this change is that anyone can pay the fee and run the transaction for someone to claim their unbonded balance. This removes an unnecessary restriction to further enhance security for operations (by reducing the reliance on a hot key and enabling separation of cold keys) and enable programmability.

## Progress

- [x] Initial `credits.aleo` implementation
- [x] snarkVM commission on rewards & staker checks
- [x] snarkVM unit tests

## Testing

1. Unit testing in snarkVM will be updated as part of this PR.
2. Before merging, we will run an isolated network with these changes.
5. Before merging, we will have comprehensive audits of the credits.aleo program and associated snarkVM changes.
6. After merging, we will first deploy on canary network before eventually running on testnet beta.


iamalwaysuncomfortable:
Would it be possible to post a rendered markdown link to Arc-41 here?

synthesizer/program/src/resources/credits.aleo
```diff
@@ -106,556 +149,462 @@ record credits:
...
+function bond_validator:
+    input r0 as address.public;  // Withdrawal address.
+    input r1 as u64.public;      // Amount of microcredits to bond.
+    input r2 as u8.public;       // Commission rate.
 
-    // Determine if the amount is at least one credit.
-    gte r2 1_000_000u64 into r3;
-    // Enforce the amount is at least one credit.
+    // Determine if the amount is at least 100 credits.
+    gte r1 100_000_000u64 into r3;
```
iamalwaysuncomfortable: What is the rationale behind the choice of this number?

Additionally, can you explain the implications of the choice of this number on the safety of the consensus process?

howardwu:
This PR is not ready for review. Converting to a draft.

synthesizer/program/src/resources/credits.aleo
chirswomack: Do we want this constraint to be for every self-bond or only for the initial bond? As written, validators would only be able to increase their self-bond by >=100 credits after joining the committee.

synthesizer/program/src/resources/credits.aleo
```diff
@@ -106,556 +149,462 @@ record credits:
...
+    is.eq r0 r6 into r8;
+    is.eq r0 r7 into r9;
+    or r8 r9 into r10;
+    // Assert that the caller is the withdrawal address.
+    assert.eq r10 true;
+    /* End Check Caller Permission */
+
+    // Check if bonded to self (validator) or to a different address (delegator).
+    is.eq r5.validator r1 into r11;
+    branch.eq r11 false to unbond_validator;
```
chirswomack: `r5.validator = r1` when unbonding a validator
```suggestion
    branch.eq r11 true to unbond_validator;
```

synthesizer/program/src/resources/credits.aleo
```diff
@@ -106,556 +149,462 @@ record credits:
...
+
+        /* Update Delegated State */
+        // Calculate new delegated total for the validator.
+        sub r13 r17 into r20;
+        // Store the new delegated total.
+        set r20 into delegated[r5.validator];
+        /* End Update Delegated state */
+
+        // Check if the new bonded state is below 10,000 credits or validator is forcing unbond.
+        branch.eq r16 false to remove_delegator;
```
chirswomack:
```suggestion
        branch.eq r16 true to remove_delegator;
```

raychu86:
Just flagging that we had an alternative approach to ARC-38 commissions via a commission mapping (and not in the `Committee` struct) - https://github.com/AleoNet/snarkVM/pull/2354.

I realize the `commissions` have not been fully implemented in this PR, but I am unsure how you intend on updating the commission rates in the `Committee`.

synthesizer/program/src/resources/credits.aleo
```diff
@@ -106,556 +149,462 @@ record credits:
...
-    // Ends the `decrement_validator` logic.
-    branch.eq true true to end;
-
-    /*** Remove Validator ***/
-
-    // Starts the `remove_validator` logic.
+    // {
+        /* Setup Unbond Validator */
+        // Retrieve the committee state or fail if it doesn't exist.
+        get committee[r5.validator] into r25;
```
chirswomack: We will reach this line if a delegator unbonds before a validator reaches the 10M minimum to join the committee (we won't branch on L480). In this case we should jump to the end.
```suggestion
        contains committee[r5.validator] into r25;
        nor r11 r25 into r26;
        branch.eq r26 true to end;
        get committee[r5.validator] into r27;
```

evanmarshall:
> Just flagging that we had an alternative approach to ARC-38 commissions via a commission mapping (and not in the `Committee` struct) - #2354.
> 
> I realize the `commissions` have not been fully implemented in this PR, but I am unsure how you intend on updating the commission rates in the `Committee`.

The design for commissions here is that you can only set the commission as you enter the committee. Also, when trying to enter the committee, you cannot have any unbonding state. This effectively forces a validator to wait 360 blocks to leave and reenter the committee, to change their commission. The goal is to create a negative incentive for validators who may otherwise rapidly change their commission making it difficult for delegators to decide who to bond to.

evanmarshall:
> Would it be possible to post a rendered markdown link to Arc-41 here?

This doesn't really follow any ARC discussion thus far. I'm hoping with this Draft PR we can agree on a final design as there are many small decisions made here that may not be universally supported. Given the scope of the changes, I think it's more appropriate to file an ARC after we can agree upon a design to present to the community as the diff is quite large.

synthesizer/program/src/resources/credits.aleo
evanmarshall: I'll move it into the checks for a new validator entering the committee. It should only be 1 credit here as @chirswomack  points out that, you should be able to increment by 1 credit with an initial self bond of 100 credits.

In terms of the safety of consensus, you should be able to set the minimum here to 0 theoretically. There should be no game theoretic difference between being delegating 10M credits and self bonding 10M credits, especially after slashing or rewards scaling is introduced. This is also extremely standard across dozens of PoS ecosystems according to @kyle-u410.

The reason 100 credits here is used is mostly practical. We could have approached this in multiple ways but instead of introducing more mappings or a new `remove_validator` function, I choose a minimum amount that a validator could practically unbond. It's a somewhat arbitrary but small amount that enables to a validator to reasonably estimate when they're trying to full unbond while not knowing the exact value onchain when the transaction is included in a block.

evanmarshall: I'll fix this. It should only be on the initial self bond.

iamalwaysuncomfortable: Most of my concerns come from the perspective of safety within the consensus protocol. 

The byzantine threshold `f` is based on *stake* and committees are dynamic. A 100 credit minimum threshold allows almost anyone to enter or leave the committee which could have the following effects:

1. The amount of stake present in the committee could have a very high variance, and potentially exceed `f` in a given round. This would be non-ideal. 

2. The committee size could grow to the maximum quickly, and the network performance with 100+ validators is not yet known. This should be *rigorously* tested before a change like this is made in my opinion. 

3. Slashing and scaling is not yet implemented or voted on by bodies governing the network - and on many networks the implementation of such things takes a long time because it needs to be rigorously tested first. Because there is no mechanism such as slashing to prevent modified validators, 100 credit minimum makes it easy for an actor to try to do so. I don't believe this is good from the perspective of safety.

kyle-u410: The threshold to activate a validator remains at 10,000,000 credits. This change allows delegated credits to contribute to that minimum balance instead of requiring a validator to self bond.

synthesizer/program/src/resources/credits.aleo
```diff
@@ -106,573 +149,496 @@ record credits:
...
-    // Set the withdrawal address.
-    set r2 into withdraw[r0];
-
-    // Continues the rest of the `bond_validator` logic.
-    position continue_bond_validator;
+finalize bond_validator:
+    input r0 as address.public;  // Validator's address.
+    input r1 as address.public;  // Withdrawal address.
+    input r2 as u64.public;      // Amount of microcredits to bond.
+    input r3 as u8.public;       // Commission rate.
```
kyle-u410, line 198: This check from `bond_public` should be applied here in same order:
```
    // Retrieve the withdrawal address for the validator.
    get.or_use withdraw[r0] r1 into r4;
    // Ensure that the withdrawal address is consistent.
    assert.eq r1 r4;
    // Set withdrawal address.
    set r4 into withdrawal[r0];

File: synthesizer/program/src/resources/credits.aleo [hunk 6]
```diff
@@ -106,573 +149,496 @@ record credits:
...
+        assert.eq r17 true;
+
+        // Get the committee size.
+        get.or_use metadata[aleo1qqqqqqqqqqqqqqqqqqqqqqqqqqqqqqqqqqqqqqqqqqqqqqqqqqqq3ljyzc] 0u32 into r18;
+        // Increment the committee size by one.
+        add r18 1u32 into r19;
+        // Set the new committee size.
+        set r19 into metadata[aleo1qqqqqqqqqqqqqqqqqqqqqqqqqqqqqqqqqqqqqqqqqqqqqqqqqqqq3ljyzc];
+        // Set the withdrawal address.
+        set r1 into withdraw[r0];
```
kyle-u410: Set should be done at top right after validating withdrawal address is not changing.

synthesizer/program/src/resources/credits.aleo
```diff
@@ -106,573 +149,496 @@ record credits:
...
-    // Determine if the caller is attempting to bond as a validator.
-    is.eq self.caller r0 into r4;
-    // Determine if the caller is a user account.
-    is.eq self.caller self.signer into r5;
-    // If the caller is attempting to bond as a validator, enforce the caller is a user account.
-    ternary r4 r5 true into r6;
-    assert.eq r6 true;
+    // Commission must be a percentage in the range 0-100 inclusive.
+    gt r2 100u8 into r4;
+    assert.neq r4 true;
```
kyle-u410, line 178: Should we force validator to have a separate withdrawal address?
```
// Prevent using hot-on-box validator address from withdrawing self-bond and rewards.
assert.neq self.signer r0;
```

In other chains having a separate value is strongly recommended, but not strictly enforced. I just can not think of any reason to allow this.

lucas-u410: This seems like a good idea, will likely prevent future hacks.

synthesizer/program/src/resources/credits.aleo
```diff
@@ -106,573 +149,496 @@ record credits:
...
+        // Increment the number of bonded delegators by one.
+        add r8 1u32 into r9;
+        // Determine if the number of delegators is less than or equal to 100_000.
+        lte r9 100_000u32 into r10;
+        // Enforce that the number of delegators is less than or equal to 100_000.
+        assert.eq r10 true;
+        // Set the new number of delegators.
+        set r9 into metadata[aleo1qgqqqqqqqqqqqqqqqqqqqqqqqqqqqqqqqqqqqqqqqqqqqqqqqqqqanmpl0];
+        // Set the withdrawal address.
+        set r2 into withdraw[r0];
```
kyle-u410: Move this up.

synthesizer/program/src/resources/credits.aleo
```diff
@@ -106,573 +149,496 @@ record credits:
...
+    input r3 as u64.public;      // Amount of microcredits to bond.
 
-    // Increment the stake for the specified validator.
-    add r21.microcredits r3 into r22;
-    // Construct the updated committee state.
-    cast r22 r21.is_open into r23 as committee_state;
+    // Retrieve the withdrawal address for the staker.
+    get.or_use withdraw[r0] r2 into r4;
+    // Ensure that the withdrawal address is consistent.
+    assert.eq r2 r4;
```
kyle-u410, line 325: Add set here to prevent action at a distance:
```
set r4 into withdraw[r0];
```

synthesizer/program/src/resources/credits.aleo
```diff
@@ -106,573 +149,496 @@ record credits:
...
-    // Set the withdrawal address.
-    set r2 into withdraw[r0];
-
-    // Continues the rest of the `bond_validator` logic.
-    position continue_bond_validator;
+finalize bond_validator:
+    input r0 as address.public;  // Validator's address.
+    input r1 as address.public;  // Withdrawal address.
+    input r2 as u64.public;      // Amount of microcredits to bond.
+    input r3 as u8.public;       // Commission rate.
```
kyle-u410: s/Commission rate/Commission percent/

synthesizer/program/src/resources/credits.aleo
```diff
@@ -106,573 +149,496 @@ record credits:
...
     assert.eq r3 true;
 
-    // Determine if the caller is attempting to bond as a validator.
-    is.eq self.caller r0 into r4;
-    // Determine if the caller is a user account.
-    is.eq self.caller self.signer into r5;
-    // If the caller is attempting to bond as a validator, enforce the caller is a user account.
-    ternary r4 r5 true into r6;
-    assert.eq r6 true;
+    // Commission must be a percentage in the range 0-100 inclusive.
```
kyle-u410: s/percentage/percent/

synthesizer/program/src/resources/credits.aleo
```diff
@@ -106,573 +149,496 @@ record credits:
...
-    // Input the withdrawal address.
-    input r1 as address.public;
-    // Input the amount of microcredits to bond.
-    input r2 as u64.public;
+// bond_validator is used to initiall bond a validator into the active committee.
+// bond_validator can also be used to increase the self-bond of the validator.
+function bond_validator:
+    input r0 as address.public;  // Withdrawal address.
+    input r1 as u64.public;      // Amount of microcredits to bond.
+    input r2 as u8.public;       // Commission rate.
```
kyle-u410: s/Commission rate/Commission percent/

synthesizer/program/src/resources/credits.aleo
```diff
@@ -106,573 +149,496 @@ record credits:
...
-function bond_public:
-    // Input the validator's address.
-    input r0 as address.public;
-    // Input the withdrawal address.
-    input r1 as address.public;
-    // Input the amount of microcredits to bond.
-    input r2 as u64.public;
+// bond_validator is used to initiall bond a validator into the active committee.
+// bond_validator can also be used to increase the self-bond of the validator.
+function bond_validator:
```
kyle-u410, line 159: Add a comment describing `commission percent`:
```
// Commission percent is the portion of delegator rewards that the validator claims for itself.
// A commission of 100 would result in all delegator rewards being earned on the validator bond.
```

synthesizer/program/src/resources/credits.aleo
```diff
@@ -12,6 +12,38 @@
 // See the License for the specific language governing permissions and
```
kyle-u410, line 12: s/2023/2024/ in Copyright above?

synthesizer/program/src/resources/credits.aleo
```diff
@@ -12,6 +12,38 @@
...
+//      - The total delegated amount for an active validator is at least 10,000,000 credits
+// 6. delegated[r0] == sum bond[*].microcredits by bond[].validator == r0
+// 7. bond[r0].microcredits < 10,000 =>
+//   a. bond[r0].validator == r0
+//   b. contains committee[r0]
+//   - A bond value < 10,000 implies self-bonded active validator.
+// 8. bond[r0].validator != r0 =>
+//   a. bond[r0].microcredits >= 10,000
+// 9. count bonded[] - count committee[] == metadata[aleo1qgqqqqqqqqqqqqqqqqqqqqqqqqqqqqqqqqqqqqqqqqqqqqqqqqqqanmpl0]
+// 10. count committee[] == metadata[aleo1qqqqqqqqqqqqqqqqqqqqqqqqqqqqqqqqqqqqqqqqqqqqqqqqqqqq3ljyzc]
```
kyle-u410, line 43: Is there a good reason for using an address here instead of simple index?
The more I look at this `metadata[0u32]` seems like a better alias for `count committee[]`.

synthesizer/program/src/resources/credits.aleo
```diff
@@ -12,6 +12,38 @@
...
+//   b. delegated[r0] >= 10,000,000 credits
+//      - The total delegated amount for an active validator is at least 10,000,000 credits
+// 6. delegated[r0] == sum bond[*].microcredits by bond[].validator == r0
+// 7. bond[r0].microcredits < 10,000 =>
+//   a. bond[r0].validator == r0
+//   b. contains committee[r0]
+//   - A bond value < 10,000 implies self-bonded active validator.
+// 8. bond[r0].validator != r0 =>
+//   a. bond[r0].microcredits >= 10,000
+// 9. count bonded[] - count committee[] == metadata[aleo1qgqqqqqqqqqqqqqqqqqqqqqqqqqqqqqqqqqqqqqqqqqqqqqqqqqqanmpl0]
```
kyle-u410, line 41: A comment here may help:
`   - Delegator count consistent with metadata.`

synthesizer/program/src/resources/credits.aleo
```diff
@@ -27,10 +59,21 @@ mapping committee:
...
+    commission as u8;
+
+/**********************************************************************************************************************/
+
+// This mapping tracks total amount bonded to an address by a validator and its delegators.
+// It includes all validators whether or not they are yet part of the committee
+mapping delegated:
+    // The key represents the address stake is delegated to. It may not yet be a validator.
+    key as address.public;
+    // The amount of microcredits bonded to the validator by delegators and not the validator.
```
kyle-u410: This value included delegator self-bond. The comment should be updated to:
```
// The amount of microcredits bonded to the validator by delegators including validator self-bond.

File: synthesizer/program/src/resources/credits.aleo (hunk 7 above)
Comment (james-u410) [reply synthesizer/program/src/resources/credits.aleo:178]: It's definitely best practice to make them separate. This does require that another key be kept secure and accessible, but validators should have strong processes around this already.

File: synthesizer/program/src/resources/credits.aleo (hunk 1 above)
Comment (evanmarshall) [reply synthesizer/program/src/resources/credits.aleo]: To add further, the self bond of 100 credits is only possible when the total delegation to a validator is 10M credits. If any delegator removes sufficient delegation to put the validator total delegation below 10M credits, the validator will be removed from the committee. In theory, the self bond could be 0 but for practical purposes is set to 100 credits. 

The real difference is whether this 10M bond comes from a self bond or a delegators (with the same 10K credits minimum per delegator). Enabling delegators to provide the minimum 10M credits is how it works in most other ecosystems.

File: synthesizer/program/src/resources/credits.aleo [hunk 18]
```diff
@@ -106,573 +149,496 @@ record credits:
...
+
+            /* Update Metadata */
+            // Retrieve the current number of delegators.
+            get.or_use metadata[aleo1qgqqqqqqqqqqqqqqqqqqqqqqqqqqqqqqqqqqqqqqqqqqqqqqqqqqanmpl0] 0u32 into r22;
+            // Decrement the number of delegators by one.
+            sub r22 1u32 into r23;
+            // Store the new number of delegators.
+            set r23 into metadata[aleo1qgqqqqqqqqqqqqqqqqqqqqqqqqqqqqqqqqqqqqqqqqqqqqqqqqqqanmpl0];
+            /* End Update Metadata */
+        // }
```
kyle-u410: Maybe add comment to indicate intentional fallthrough.

synthesizer/program/src/resources/credits.aleo
```diff
@@ -708,7 +674,7 @@ finalize set_validator_state:
     get committee[r0] into r2;
 
     // Construct the updated committee state.
-    cast r2.microcredits r1 into r3 as committee_state;
+    cast r1 r2.commission into r3 as committee_state;
     // Update the committee state for the specified validator.
     set r3 into committee[r0];
```
kyle-u410, line 760: Optional addition

[Idea from Solana allow withdrawal address to change](https://docs.solanalabs.com/cli/usage#solana-vote-authorize-withdrawer):
```
// Must be called by withdrawal address.
function set_withdrawal_address:
    input r0 as address.public; // Bonded address that will have a new withdrawal address set.
    input r1 as address.public; // New withdrawal address.

    // Do not allow updating withdrawal address to same as bonded address.
    assert.neq r0 r1;

    async set_withdrawal_address self.caller r0 r1 into r1;
    output r1 as credits.aleo/set_withdrawal_address.future;

finalize set_withdrawal_address:
    input r0 as address.public; // Current withdrawal address.
    input r1 as address.public; // Bonded address that will have a new withdrawal address set.
    input r2 as address.public; // New withdrawal address.

    // Validate current withdrawal address against caller. Fail if withdraw address does not exist.
    get withdraw[r1] into r3;
    assert.eq r0 r3;

    // Update to new withdrawal addrss.
    set r2 into withdraw[r1];
```

synthesizer/program/src/resources/credits.aleo
```diff
@@ -106,573 +149,507 @@ record credits:
...
+        position remove_delegator;
+        // {
+            /* Update Bonded state */
+            // Remove the bonded state.
+            remove bonded[r1];
+            /* End Update Bonded state */
+
+            /* Update Metadata */
+            // Retrieve the current number of delegators.
+            get.or_use metadata[aleo1qgqqqqqqqqqqqqqqqqqqqqqqqqqqqqqqqqqqqqqqqqqqqqqqqqqqanmpl0] 0u32 into r22;
```
chirswomack: just use `get` here

synthesizer/program/src/resources/credits.aleo
```diff
@@ -106,573 +149,507 @@ record credits:
...
     // Construct the updated committee state.
-    cast r11 r10.is_open into r12 as committee_state;
+    cast r6.is_open r6.commission into r7 as committee_state;
 
     /* Bonded */
 
     // Construct the initial bond state.
-    cast r0 0u64 into r13 as bond_state;
+    cast r0 0u64 into r8 as bond_state;
     // Get the bond state for the caller, or default to the initial bond state.
```
chirswomack:
```suggestion
    // Get the bond state for the signer, or default to the initial bond state.
```

synthesizer/program/src/resources/credits.aleo
```diff
@@ -106,573 +149,507 @@ record credits:
...
-    assert.eq r16 true;
+    assert.eq r13 true;
 
     // Construct the updated bond state.
-    cast r0 r15 into r17 as bond_state;
+    cast r0 r10 into r14 as bond_state;
 
     /* Account */
 
     // Get the balance of the caller.
```
chirswomack:
```suggestion
    // Get the balance of the signer.
```

synthesizer/program/src/resources/credits.aleo
```diff
@@ -106,573 +149,507 @@ record credits:
...
     // If the account does not exist, this finalize scope will fail.
-    get account[r0] into r18;
+    get account[r0] into r15;
     // Decrement the balance of the caller.
```
chirswomack:
```suggestion
    // Decrement the balance of the signer.
```

synthesizer/program/src/resources/credits.aleo
```diff
@@ -106,573 +149,507 @@ record credits:
...
+        // Fail if current unbonding state
+        assert.eq r21 false;
+    // }
+
+    position validator_in_committee;
+
     // Update the committee state of the specified validator.
-    set r12 into committee[r0];
+    set r7 into committee[r0];
     // Update the bond state for the caller.
```
chirswomack:
```suggestion
    // Update the bond state for the signer.
```

synthesizer/program/src/resources/credits.aleo
```diff
@@ -106,573 +149,507 @@ record credits:
...
-    set r17 into bonded[r0];
+    set r14 into bonded[r0];
     // Update the balance of the caller.
```
chirswomack:
```suggestion
    // Update the balance of the signer.
```

synthesizer/program/src/resources/credits.aleo
```diff
@@ -106,573 +149,507 @@ record credits:
...
+
+        // Retrieve the bond state of the validator.
+        get bonded[r5.validator] into r28;
+
+        // Retrieve the total delegated balance for the validator.
+        get delegated[r5.validator] into r29;
+        /* End Setup Unbond Validator */
+
+        // Check if the current delegated amount is below 10M credits.
+        lte r29 10_000_000_000_000u64 into r30;
```
chirswomack:
```suggestion
        lt r29 10_000_000_000_000u64 into r30;
```

synthesizer/program/src/resources/credits.aleo
```diff
@@ -106,573 +149,507 @@ record credits:
...
+        branch.eq r30 true to remove_validator;
+
+        // Calculate the new delegated total after unbonding.
+        sub r29 r2 into r31;
+
+        // Calculate the new bond state after unbonding.
+        sub r28.microcredits r2 into r32;
+
+        // Check if the new bond state is below the required minimums.
+        gte r32 100u64 into r33; // Minimum self bond requirement: 100 credits.
```
chirswomack:
```suggestion
        gte r32 100_000_000u64 into r33; // Minimum self bond requirement: 100 credits.
```

synthesizer/program/src/resources/credits.aleo
```diff
@@ -106,573 +149,507 @@ record credits:
...
+        get committee[r5.validator] into r27;
+
+        // Retrieve the bond state of the validator.
+        get bonded[r5.validator] into r28;
+
+        // Retrieve the total delegated balance for the validator.
+        get delegated[r5.validator] into r29;
+        /* End Setup Unbond Validator */
+
+        // Check if the current delegated amount is below 10M credits.
```
chirswomack, line 577: Update comment to note this is only if the delegator caused this to drop below 10M above

synthesizer/program/src/resources/credits.aleo
```diff
@@ -106,573 +149,507 @@ record credits:
...
+
+        /* Update Delegated Total */
+        // Store the new delegated total.
+        set r31 into delegated[r1];
+        /* End Update Delegated Total */
+
+        /* Update Validator Bonded State */
+        // Construct and store the new bond state.
+        cast r5.validator r32 into r39 as bond_state;
+        set r39 into bonded[r1];
```
chirswomack:
```suggestion
        set r39 into bonded[r5.validator];
```

synthesizer/program/src/resources/credits.aleo
```diff
@@ -106,573 +149,507 @@ record credits:
...
+
+        /* Update Unbonding State */
+        // Retrieve or initialize the unbonding state.
+        get.or_use unbonding[r5.validator] r4 into r43;
+        // Increment the unbond amount by the full bonded amount.
+        add r28.microcredits r43.microcredits into r44;
+        // Set the updated unbond state.
+        cast r44 r3 into r45 as unbond_state;
+        // Store the new unbonding state.
+        set r45 into unbonding[r1];
```
chirswomack:
```suggestion
        set r45 into unbonding[r5.validator];
```

synthesizer/src/vm/finalize.rs
```diff
@@ -1048,7 +1050,7 @@ impl<N: Network, C: ConsensusStorage<N>> VM<N, C> {
                     //     }
                     // }
 
-                    // Calculate the stake per validator using `bonded_balances`.
+                    // Calculate the stake per validator using `bonded_balances`. No need to check delegated mapping as this is the genesis block.
```
fulltimemike: For safety, we could assert that the total bonded balance matches the total delegated balance, but as this is the genesis block, we shouldn't have arbitrary bonded balances to addresses not in the committee

synthesizer/src/vm/helpers/committee.rs
```diff
@@ -155,52 +191,67 @@ pub fn ensure_stakers_matches<N: Network>(
...
-pub fn to_next_commitee_map_and_bonded_map<N: Network>(
+pub fn to_next_credits_maps<N: Network>(
     next_committee: &Committee<N>,
     next_stakers: &IndexMap<Address<N>, (Address<N>, u64)>,
-) -> (Vec<(Plaintext<N>, Value<N>)>, Vec<(Plaintext<N>, Value<N>)>) {
+    next_delegated: &IndexMap<Address<N>, u64>,
+) -> (Vec<(Plaintext<N>, Value<N>)>, Vec<(Plaintext<N>, Value<N>)>, Vec<(Plaintext<N>, Value<N>)>) {
     // Prepare the identifiers.
     let validator_identifier = Identifier::from_str("validator").expect("Failed to parse 'validator'");
     let microcredits_identifier = Identifier::from_str("microcredits").expect("Failed to parse 'microcredits'");
```
fulltimemike, line 239: Can delete this

synthesizer/src/vm/helpers/rewards.rs
```diff
@@ -43,39 +43,87 @@ pub fn staking_rewards<N: Network>(
...
+            updated_stakers.insert(*staker, (*validator, *stake));
+            continue;
+        }
+        // Cast the staking reward as a u64.
+        // Note: This '.expect' is guaranteed to be safe, as we ensure the quotient is within a safe bound.
+        let staking_reward = u64::try_from(quotient).expect("Staking reward is too large");
+
+        // Compute the commission for the validator.
+        let commission = (staking_reward as u128)
+            .saturating_mul(commission_rate as u128)
```
fulltimemike: Might be worth adding a safety check that commission_rate is <= 100u8

synthesizer/program/src/resources/credits.aleo
```diff
@@ -12,6 +12,38 @@
...
+// 6. delegated[r0] == sum bond[*].microcredits by bond[].validator == r0
+// 7. bond[r0].microcredits < 10,000 =>
+//   a. bond[r0].validator == r0
+//   b. contains committee[r0]
+//   - A bond value < 10,000 implies self-bonded active validator.
+// 8. bond[r0].validator != r0 =>
+//   a. bond[r0].microcredits >= 10,000
+// 9. count bonded[] - count committee[] == metadata[aleo1qgqqqqqqqqqqqqqqqqqqqqqqqqqqqqqqqqqqqqqqqqqqqqqqqqqqanmpl0]
+// 10. count committee[] == metadata[aleo1qqqqqqqqqqqqqqqqqqqqqqqqqqqqqqqqqqqqqqqqqqqqqqqqqqqq3ljyzc]
+// 11. 100,000 >= count bonded[] >= count delegated[] >= count committee[]
```
raychu86: Is this correct?

The maximum number of delegators is 100,000. However, the `count bonded` includes the delegators and validators.

So `count bonded` could be `100,000 + number of validators` which is greater than `100,000`.

synthesizer/program/src/resources/credits.aleo
```diff
@@ -106,573 +149,515 @@ record credits:
...
-    input r2 as u64.public;
-
-    // Determine if the amount is at least one credit.
-    gte r2 1_000_000u64 into r3;
-    // Enforce the amount is at least one credit.
+// This function allows a validator to self-bond to join the active committee, and specify a commission percent and withdrawal address.
+// Commission percent is the portion of delegator rewards that the validator claims for itself.
+// A commission of 100 would result in all delegator rewards being retained by the validator.
+// bond_validator can also be used to increase the self-bond of the validator.
+function bond_validator:
```
raychu86, line 159: Add documentation explaining that the withdrawal address and the commission should be the same as before if the validator has already joined the validator set.

synthesizer/program/src/resources/credits.aleo
```diff
@@ -106,573 +149,515 @@ record credits:
...
+    input r0 as address.public;  // Withdrawal address.
+    input r1 as u64.public;      // Amount of microcredits to bond.
+    input r2 as u8.public;       // Commission percent.
+
+    // Prevent using hot-on-box validator address from withdrawing self-bond and rewards.
+    assert.neq self.signer r0;
+
```
raychu86, line 163: We need to prevent the validator from being a program address, like we did in - https://github.com/AleoNet/snarkVM/pull/2443.

Fortunately, I believe since this function is validator specific, a simple `assert.neq self.signer self.caller` would suffice.

synthesizer/program/src/resources/credits.aleo
```diff
@@ -106,573 +149,515 @@ record credits:
...
-    set r19 into account[r0];
+    // if (new_validator)
+    contains committee[r0] into r17;
+    branch.eq r17 true to validator_in_committee;
+    // {
+        // Check if initial bond amount is at least 100 credits
+        gte r2 100_000_000u64 into r18;
+        // Assert that initial self bond is at least 100 credits
+        assert.eq r18 true;
+
```
raychu86: Why not enforce this above by checking that r10 is greater than 100 credits?

That way we ensure that in the case of existing validators, we guarantee the self-bonded balance is at least 100 credits.

raychu86: Note: We already do this for delegators.

synthesizer/program/src/resources/credits.aleo
raychu86: @d0cd

synthesizer/program/src/resources/credits.aleo
```diff
@@ -106,573 +149,515 @@ record credits:
...
 
-// This function allows any staker to unbond their microcredits from a validator.
-// The corresponding function for 'unbond_public' is 'claim_unbond_public'.
+// Unbond a staker using their associated withdrawal address
+// **Validators**: It will remove the validator if the self bonded balance falls below 100 credits
+// or the total delegated balance falls below 10M credits.
+// **Delegators**: It will remove the validator if the total delegated balance falls below 10M credits.
+// It will remove the entire bond_state of the delegator if it falls below 10,000 credits.
+//
+// Validators can also forcibly unbond a delegator
```
raychu86: This line should be more descriptive. Does this mean forcibly unbonding individual delegators or all them once once the self bonded balance falls below 100?

parameters/src/testnet/resources/transfer_private.metadata
```diff
@@ -1,6 +1,6 @@
 {
   "prover_checksum": "6733de36be3291a1a6f95ed7b986ccb29fc69707788582fa672395abe3446e80",
   "prover_size": 75949172,
-  "verifier_checksum": "1291734485f70e932337080b7724a3ad49a965c5e2120ed3e47521c75cbf8c83",
-  "verifier_size": 665
+  "verifier_checksum": "86e2667e415b0e6266e0e9f3edb99746837debca70e9e82fd7d55f732ff7d995",
```
howardwu: Why do all of these irrelevant transition functions change parameters?

howardwu: Your intuition should give off a major red flag if you see behavior like this, and it should be documented if this change is well-founded.

synthesizer/program/src/resources/credits.aleo
```diff
@@ -106,573 +149,515 @@ record credits:
...
+    set r12 into delegated[r0];
 
-    /******* Bond Delegator *******/
+/**********************************************************************************************************************/
 
-    // Starts the `bond_delegator` logic.
-    position bond_delegator;
+// This function allows any staker to bond their microcredits to a validator and specify a withdrawal address.
+// The corresponding functions for 'bond_public' are 'unbond_public' and 'claim_unbond_public'.
+function bond_public:
```
raychu86, line 291: Why not rename this to `bond_delegator` to be clear it is for delegators only?

synthesizer/program/src/resources/credits.aleo
```diff
@@ -1,4 +1,4 @@
-// Copyright (C) 2019-2023 Aleo Systems Inc.
+// Copyright (C) 2019-2024 Aleo Systems Inc.
```
howardwu, line 1: revert: Aleo Systems Inc. does NOT retain IP control in 2024.

synthesizer/program/src/resources/credits.aleo
howardwu: revert: Aleo Systems Inc. does NOT retain IP control in 2024.

synthesizer/program/src/resources/credits.aleo
```diff
@@ -106,573 +149,515 @@ record credits:
...
-// The corresponding function for 'unbond_public' is 'claim_unbond_public'.
+// Unbond a staker using their associated withdrawal address
+// **Validators**: It will remove the validator if the self bonded balance falls below 100 credits
+// or the total delegated balance falls below 10M credits.
+// **Delegators**: It will remove the validator if the total delegated balance falls below 10M credits.
+// It will remove the entire bond_state of the delegator if it falls below 10,000 credits.
+//
+// Validators can also forcibly unbond a delegator
+
 function unbond_public:
```
raychu86, line 420: If we are separating `bond_public` into `bond_validator` and `bond_public`.

Why not do the same for `unbond_public`? `unbond_validator` and `unbond_delegator` may be easier to implement and read, since we are already going down the route of separation.

synthesizer/program/src/resources/credits.aleo
```diff
@@ -106,573 +149,515 @@ record credits:
...
-// The `claim_unbond_public` function allows any staker to claim their microcredits
+// The `claim_unbond_public` function allows any address to run a crank to enable a staker to claim their microcredits
 // to their withdrawal address after the unbonding period.
 //
 // This function also removes the staker's withdrawal address if the staker no longer has any bonded balance.
 function claim_unbond_public:
+    // The address of the staker
+    input r0 as address.public;
+    // The number of microcredits to claim
+    input r1 as u64.public;
```
raychu86: What is the point of adding "The number of microcredits to claim"? You already enforce in finalize that this value must be equivalent to the entire unbond amount. Why not just implicitly unbond it all like the previous logic?

synthesizer/program/src/resources/credits.aleo
raychu86: The proposed `bond_validator` function does not act as the "prebond" mechanism. It's for **validators** to add to its existing state as committee member, or join the committee if there is sufficient provided stake or enough "prebonded" delegator stake.

The proposed `bond_public` logic acts as a mechanism for **delegators** to bond to an existing validator in the committee, or "prebond" to a validator address that is not yet in the committee.

synthesizer/program/src/resources/credits.aleo
```diff
@@ -106,573 +149,515 @@ record credits:
...
+    get.or_use committee[r0] r5 into r6;
+
+    // Fail if attempting to change commission. Commission can only be set when validator bonds the first time.
+    assert.eq r3 r6.commission;
 
-    // Increment the stake for the specified validator.
-    add r10.microcredits r3 into r11;
     // Construct the updated committee state.
-    cast r11 r10.is_open into r12 as committee_state;
+    cast r6.is_open r6.commission into r7 as committee_state;
```
howardwu: This operation is technically unnecessary. You can use `r6` (which is equivalent to `r7`).

howardwu: Resolved in [1a13a11](https://github.com/AleoNet/snarkVM/pull/2453/commits/1a13a11ebd17b828ca7b8fbcbe70b3e38b8505e0)

synthesizer/program/src/resources/credits.aleo
evanmarshall: Any reason not to use a `u8` here or even a boolean?

synthesizer/program/src/resources/credits.aleo
evanmarshall: I think we should probably not add this now to minimize the complexity of changes.

synthesizer/program/src/resources/credits.aleo
evanmarshall: You're right, I'll update it.

synthesizer/program/src/resources/credits.aleo
evanmarshall: I added it b/c to me it made sense to check it in the branch that adds the validator to the committee as it's an initial bond_validator condition. Given the unbond conditions of removing the validator with a self bond of less than 100 credits, it seems like these should both be equivalent.

synthesizer/program/src/resources/credits.aleo
evanmarshall: It means they can choose to unbond any of their delegators, one by one. New comment will be:
```
// Validators have the permission to completely unbond any of their delegators
// When a validator unbonds a delegator, its entire bonded balance is unbonded regardless of the amount of microcredits input
// and may end up removing the validator from the committee if the total delegated balance falls below 10M credits
```

parameters/src/testnet/resources/transfer_private.metadata
evanmarshall: Good catch. As for a follow up from @raychu86:

```
 It’s not a big deal thankfully, especially since the proving keys are untouched. TLDR - it’s because of this PR -  https://github.com/AleoNet/snarkVM/pull/2444 that adds the 8 bytes of num_variables to the end the wrapper VKs. But this doesn’t seem to fail because we use the underlying Varuna VK when loading credits.aleo VKs from bytes.
 ```

File: synthesizer/program/src/resources/credits.aleo (hunk 40 above)
Comment (evanmarshall) [reply synthesizer/program/src/resources/credits.aleo:291]: As for naming conventions, I think both `bond_public` and `bond_delegator` make sense. I don't really have a preference here. I just kept it the same as to try to minimize changes. 

It'd probably make sense to keep it `bond_public` if at some point we plan on introducing a `bond_private`
Do you have a strong preference here?

File: synthesizer/program/src/resources/credits.aleo (hunk 42 above)
Comment (evanmarshall) [reply synthesizer/program/src/resources/credits.aleo:420]: I separated `bond_public` into `bond_validator` & `bond_public` as they now have different inputs. Primarily, the addition of a commission made it seem strange to combine them.

I left `unbond_public` as one function as the functionality overlaps a lot. Because a validator can unbond a delegator (letting us delete the `unbond_delegator_as_validator` function and because a delegator can unbond themselves and potentially unbond a validator as a result, I left it as a single function to avoid lots of duplication.

File: synthesizer/program/src/resources/credits.aleo (hunk 43 above)
Comment (evanmarshall) [reply synthesizer/program/src/resources/credits.aleo]: @kyle-u410 My understanding is that this is a coinbase requirement for know exactly how many credits were claimed at a later date.

Comment (kyle-u410) [reply synthesizer/program/src/resources/credits.aleo]: It is important to know what the final claim amount is, but that should be present onchain when `claim_unbond_public` is finalized.

Computing the value to specify here would be annoying.

Additionally, even if the amount is specified here, there would still need to be sufficient block indexing to read the finalized result to know if the claim was successfull.

So I suspect the only practical result of requiring microcredits here would be an increased number of failed `claim_unbond_public` transactions.

Comment (evanmarshall) [reply synthesizer/program/src/resources/credits.aleo]: I'll remove this argument.

**Comment by evanmarshall** [comment]
@howardwu @raychu86 Can you help recreate the genesis block? I'm not sure if we want to take advantage of using the minimum self bond for validators & create new delegators.

File: synthesizer/program/src/resources/credits.aleo (hunk 40 above)
Comment (raychu86) [reply synthesizer/program/src/resources/credits.aleo:291]: Not a strong preference, `bond_public` should be fine for now to minimize drift.

File: synthesizer/program/src/resources/credits.aleo (hunk 15 above)
Comment (raychu86) [reply synthesizer/program/src/resources/credits.aleo:43]: Initially `metadata` was supposed to be used for user accounts. We can change this to use a u8 if needed.

File: ledger/src/tests.rs [hunk 45]
```diff
@@ -489,6 +489,7 @@ finalize foo:
     assert_eq!(program, ledger.get_program(program_id).unwrap())
 }
 
+// TODO update this test for the new bonding/commission
```
howardwu: nit: this `TODO` needs to be addressed

synthesizer/src/vm/helpers/committee.rs
```diff
@@ -31,41 +31,69 @@ use std::str::FromStr;
...
                     // Return the committee state.
-                    Ok((*address, (microcredits, is_open)))
+                    (is_open, commission)
                 }
                 _ => bail!("Invalid committee value (missing struct) - {value}"),
-            }
+            };
+
+            // Extract the microcredits for the address from the delegated map.
+            let microcredits = delegated_map.iter().find_map(|(delegated_key, delegated_value)| {
```
howardwu: @ljedrz can you assess this for performance implications? (use `test_committee_and_delegated_maps_into_committee` for profiling)

We need a full committee with full delegated maps.

raychu86:
@evanmarshall Genesis blocks generated in - https://github.com/AleoNet/snarkVM/pull/2453/commits/8ba7dc5b7e6c0a35af131d3dd89f86e06104abb7

synthesizer/src/vm/helpers/committee.rs
ljedrz: That test takes roughly the same amount of time as before; however, due to the scope of these changes it would be best to run a profile in the IsoNet and compare it against one from a run with the current code if we'd like to make sure that there are no regressions.

synthesizer/program/src/resources/credits.aleo
d0cd: Switching to a `u8` would be fine.

vicsn:
Some tests are still failing. You can find fixes for some of them [in this branch](https://github.com/AleoHQ/snarkVM/tree/arc-41). I'll need input from @evanmarshall or @fulltimemike  on:
- `test_staking_rewards_when_staker_is_under_min_yields_no_reward` - I assume this is because of the commission rate. Not sure if we should just remove `assert_eq!(candidate_stake, stake);` or explicitly recalculate the commission in the test.
- `test_insufficient_public_fees`
- `test_max_committee_limit_with_bonds`

evanmarshall:
> Some tests are still failing. You can find fixes for some of them [in this branch](https://github.com/AleoHQ/snarkVM/tree/arc-41). I'll need input from @evanmarshall or @fulltimemike on:
> 
> * `test_staking_rewards_when_staker_is_under_min_yields_no_reward` - I assume this is because of the commission rate. Not sure if we should just remove `assert_eq!(candidate_stake, stake);` or explicitly recalculate the commission in the test.
> * `test_insufficient_public_fees`
> * `test_max_committee_limit_with_bonds`

These tests are fixed.

ledger/src/tests.rs
```diff
@@ -206,6 +206,8 @@ fn test_insufficient_public_fees() {
     // Sample recipient.
     let recipient_private_key = PrivateKey::new(rng).unwrap();
     let recipient_address = Address::try_from(&recipient_private_key).unwrap();
+    let withdrawal_private_key = PrivateKey::<MainnetV0>::new(rng).unwrap();
```
vicsn, line 209: Most other tests use an intermediary `CurrentNetwork`, please adjust accordingly for ease of review

vicsn (approved):
CI passes, review comments look incorporated, approving this so it can be tested in canary. If existing reviewers or auditors encounter issues, they can open new PRs.

apruden2008 (approved):
Looks good to me. We should work to add more test cases for this in the future, but in the meantime this looks good to merge in so we can start doing live testing on the canary net.

Great work @evanmarshall on pulling this together.

PR #2456, opened 2024-05-16
Fairly price storage costs for transactions

iamalwaysuncomfortable:
## Motivation

Transactions are currently limited to being `128kb` or less in size. 

Most execution transactions are small (for instance, `transfer_public` is `2880` bytes in size). However, transaction size can be greatly increased if the function being executed has very large inputs or very large outputs in the form of nested `structs` or `arrays`.

SnarkOS nodes must allocate enough memory to verify all transactions and carry out  `speculate` operations on them. If many executions near the `128kb` limit are sent to a network in a short timeframe, it can lead to a large amount of memory growth, long block times, long sync times and fast ledger growth. 

Below is an example of memory growth on a validator observed from sending `3 TPS` of `124kb` transactions over 2 hours.
Screen Shot 2024-05-10 at 3 38 13 PM

Currently a `128kb` transaction would only cost `.128` credits to execute, which makes it relatively cheap to flood the network with them. Given the deleterious effects of large transactions to network health, it makes sense to impose higher storage costs on larger transactions to ensure users of them are paying the appropriate costs for the effort it takes to process them.

This PR proposes that the storage cost of execution transactions above `5kb` in size be penalized with a quadratic function.

```math
\text{fee\_microcredits} =
\begin{cases}
\text{num\_bytes} & \text{if } 0 < \text{num\_bytes} \leq 5 \\
\frac{{num\_bytes}^2}{5000} & \text{otherwise} & \text{if } \text{num\_bytes} > 5 \\
\end{cases}
```

An exact function to price storage remains open for debate, and alternatives can certainly be proposed, but it is *strongly* encouraged storage costs be increased to properly pay for the extra effort the network must perform verify and store them.

NOTE: This would be a breaking change for Testnet, so this should be a pre-mainnet change and integrated into `mainnet-staging`.


synthesizer/process/src/cost.rs
```diff
@@ -60,7 +60,12 @@ pub fn deployment_cost<N: Network>(deployment: &Deployment<N>) -> Result<(u64, (
 /// Returns the *minimum* cost in microcredits to publish the given execution (total cost, (storage cost, finalize cost)).
 pub fn execution_cost<N: Network>(process: &Process<N>, execution: &Execution<N>) -> Result<(u64, (u64, u64))> {
     // Compute the storage cost in microcredits.
-    let storage_cost = execution.size_in_bytes()?;
+    let mut storage_cost = execution.size_in_bytes()?;
+
+    // Compute a storage cost penalty if above the size penalty threshold.
+    if storage_cost > EXECUTION_SIZE_PENALTY_THRESHOLD {
+        storage_cost = storage_cost.saturating_mul(100)
```
vicsn: For consistency, I would name the magic 100 number similar to DEPLOYMENT_FEE_MULTIPLIER and put both these new constants in the network trait

iamalwaysuncomfortable: Yeah the network trait is certainly a better place for these.

raychu86:
When a decision is made, can you post the before and after fees amounts for the `credits.aleo` functions?

iamalwaysuncomfortable:
> When a decision is made, can you post the before and after fees amounts for the `credits.aleo` functions?

The decision on the function has been made and posted above on the PR description:

The costs of major credits.aleo functions are as follows:
```
bond_public: 306229 (storage: 1519, finalize: 304710)
join: 2087 (storage: 2087, finalize: 0)
split: 2131 (storage: 2131, finalize: 0)
transfer_private: 2242 (storage: 2242, finalize: 0)
transfer_private_to_public: 26961 (storage: 2141, finalize: 24820)
transfer_public: 51060 (storage: 1420, finalize: 49640)
transfer_public_to_private: 26439 (storage: 1619, finalize: 24820)
unbond_public: 325149 (storage: 1309, finalize: 323840)
```

There haven't been any changes in storage costs in credits.aleo functions under this pricing model.

console/network/src/lib.rs
```diff
@@ -115,6 +115,12 @@ pub trait Network:
     const STARTING_SUPPLY: u64 = 1_500_000_000_000_000; // 1.5B credits
     /// The cost in microcredits per byte for the deployment transaction.
     const DEPLOYMENT_FEE_MULTIPLIER: u64 = 1_000; // 1 millicredit per byte
+    /// The degree of the polynomial used to calculate the storage fee.
+    const EXECUTION_STORAGE_FEE_DEGREE: u32 = 2;
+    /// The constant that divides the storage polynomial.
+    const EXECUTION_STORAGE_FEE_SCALING_FACTOR: u64 = 1000;
+    /// The maximum size execution transactions can be before a quadratic storage penalty applies.
```
vicsn, line 123:
```suggestion
    /// The maximum size (in bytes) which execution transactions can be before a quadratic storage penalty applies.
```

vicsn (dismissed):
Await merge until Howard approved and CI passes

synthesizer/process/src/cost.rs
```diff
@@ -60,7 +60,14 @@ pub fn deployment_cost<N: Network>(deployment: &Deployment<N>) -> Result<(u64, (
 /// Returns the *minimum* cost in microcredits to publish the given execution (total cost, (storage cost, finalize cost)).
 pub fn execution_cost<N: Network>(process: &Process<N>, execution: &Execution<N>) -> Result<(u64, (u64, u64))> {
     // Compute the storage cost in microcredits.
-    let storage_cost = execution.size_in_bytes()?;
+    let mut storage_cost = execution.size_in_bytes()?;
+
+    // Compute a quadratic storage cost penalty if above the size penalty threshold.
+    if storage_cost > N::EXECUTION_STORAGE_PENALTY_THRESHOLD {
+        storage_cost = storage_cost
+            .saturating_pow(N::EXECUTION_STORAGE_FEE_DEGREE)
```
howardwu: I strongly advise we use a quadratic function otherwise use an approximation algorithm for `pow`, similar to how to do `retarget` in the puzzle.

The concern here is determinism across architectures and across programming languages.

synthesizer/process/src/cost.rs
```diff
@@ -60,7 +60,14 @@ pub fn deployment_cost<N: Network>(deployment: &Deployment<N>) -> Result<(u64, (
 /// Returns the *minimum* cost in microcredits to publish the given execution (total cost, (storage cost, finalize cost)).
 pub fn execution_cost<N: Network>(process: &Process<N>, execution: &Execution<N>) -> Result<(u64, (u64, u64))> {
     // Compute the storage cost in microcredits.
-    let storage_cost = execution.size_in_bytes()?;
+    let mut storage_cost = execution.size_in_bytes()?;
+
+    // Compute a quadratic storage cost penalty if above the size penalty threshold.
+    if storage_cost > N::EXECUTION_STORAGE_PENALTY_THRESHOLD {
```
howardwu: I would strongly advise introducing unit tests that explicitly **clamp** the expected upper bound and lower bound numeric value behaviors prior to the approval and merge of this PR.

synthesizer/process/src/cost.rs
Meshiest: Is there a reason why the penalty is applied to the entire `storage_cost` rather than just overage?

A jump in fee by 20,010 for 1 additional byte seems a little concerning

|$(storageCost ^{degree})/scalingFactor$|new storage_cost|
|-|-|
|$(5001 ^2)/1000$|25,010|
|$(5100 ^2)/1000$|26,010|
|$(6000 ^2)/1000$|36,000|
|$(32,000 ^2)/1000$|1,024,000|
|$(64,000 ^2)/1000$|4,096,000|
|$(128,000 ^2)/1000$|16,384,000|

gluax: May I ask how this would be non-determinism here? Both `storage_cost` and the fee degree are integers which power in rust is deterministic for.

And also, what is meant by other languages? WASM compatibility?

iamalwaysuncomfortable: > Is there a reason why the penalty is applied to the entire storage_cost rather than just overage?

Namely because as transaction size grows, block size does as well. Large blocks with many large transactions require each node to fetch these transmissions to complete their view of the `dag` which causes stress on the BFT and syncing processes. 

While perhaps not the most mathematically beautiful, it prioritizes network health by ensuring penalties are applied to larger transactions starting early. It's likely that most functions on the network will have transaction sizes under 5kb, which is the reasoning for the initially linear pricing. 

Pricing information like the table you crafted above can be added to network documentation ensure this is explained.

synthesizer/process/src/cost.rs
iamalwaysuncomfortable: These should be able to added before the end of this week.

synthesizer/process/src/cost.rs
iamalwaysuncomfortable: > And also, what is meant by other languages? WASM compatibility?

`WASM` is certainly a good example. Wasm does have 64-bit types but those are represented in 32-bit memory addresses so "saturation" operations might not translate here.

synthesizer/process/src/cost.rs
apruden2008: @iamalwaysuncomfortable what's the status of these unit tests? Per @howardwu recommendation, we don't want to merge until we get those tests included.

iamalwaysuncomfortable: @apruden2008 The tests have been added.

vicsn (dismissed):
LGTM, great tests

evanmarshall (dismissed):
:shipit:

synthesizer/process/src/cost.rs
iamalwaysuncomfortable: The modifier is now changed to 5000 - so this discontinuity no longer exists.

evanmarshall (approved):
:shipit:

synthesizer/process/src/cost.rs
Meshiest: |$(storageCost ^{degree})/scalingFactor$|new storage_cost|
|-|-|
|$(5001 ^2)/5000$|5002|
|$(5100 ^2)/5000$|5202|
|$(6000 ^2)/5000$|7200|
|$(32,000 ^2)/5000$|204,800|
|$(64,000 ^2)/5000$|819,200|
|$(128,000 ^2)/5000$|3,276,800|

PR #2458, opened 2024-05-20
[Perf] Reduce allocations in ToBits for Plaintext

ljedrz:
Calls to `Plaintext::write_bits_le` are responsible for a large number of allocations when performing deployments and executions; using a few simple tweaks we can reduce that number by as much as ~18% in some scenarios (my run was performed on a large deployment and execution).

Also, since we are dealing with `OnceCell` here, the additional calls to `Vec::shrink_to_fit` should be perfectly acceptable and slightly reduce heap use.

### DIFF
--- a/circuit/program/src/data/plaintext/to_bits.rs
+++ b/circuit/program/src/data/plaintext/to_bits.rs
@@ -23,10 +23,12 @@ impl<A: Aleo> ToBits for Plaintext<A> {
             Self::Literal(literal, bits_le) => {
                 // Compute the bits of the literal.
                 let bits = bits_le.get_or_init(|| {
-                    let mut bits_le = vec![Boolean::constant(false), Boolean::constant(false)]; // Variant bit.
+                    let mut bits_le = Vec::new();
+                    bits_le.extend([Boolean::constant(false), Boolean::constant(false)]); // Variant bit.
                     literal.variant().write_bits_le(&mut bits_le);
                     literal.size_in_bits().write_bits_le(&mut bits_le);
                     literal.write_bits_le(&mut bits_le);
+                    bits_le.shrink_to_fit();
                     bits_le
                 });
                 // Extend the vector with the bits of the literal.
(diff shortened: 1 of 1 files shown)


circuit/program/src/data/plaintext/to_bits.rs
```diff
@@ -35,15 +37,17 @@ impl<A: Aleo> ToBits for Plaintext<A> {
...
+                    let mut bits_le = Vec::new();
+                    bits_le.extend([Boolean::constant(false), Boolean::constant(true)]); // Variant bit.
                     U8::constant(console::U8::new(members.len() as u8)).write_bits_le(&mut bits_le);
                     for (identifier, value) in members {
                         let value_bits = value.to_bits_le();
                         identifier.size_in_bits().write_bits_le(&mut bits_le);
                         identifier.write_bits_le(&mut bits_le);
                         U16::constant(console::U16::new(value_bits.len() as u16)).write_bits_le(&mut bits_le);
-                        bits_le.extend_from_slice(&value_bits);
+                        bits_le.extend(value_bits);
```
ljedrz, line 48: note: `Vec::extend` is preferable here, as it avoids cloning the members of of `value_bits`.

circuit/program/src/data/plaintext/to_bits.rs
```diff
@@ -23,10 +23,12 @@ impl<A: Aleo> ToBits for Plaintext<A> {
             Self::Literal(literal, bits_le) => {
                 // Compute the bits of the literal.
                 let bits = bits_le.get_or_init(|| {
-                    let mut bits_le = vec![Boolean::constant(false), Boolean::constant(false)]; // Variant bit.
+                    let mut bits_le = Vec::new();
```
ljedrz, line 26: note: this is a conservative tweak which ensures that the vector starts out with a capacity larger than 2 elements

howardwu (approved):

howardwu:
We need CI here

iamalwaysuncomfortable (approved):
These look like safe wins, but I'd agree that it needs CI before merge.

PR #2459, opened 2024-05-21
[Refactor] Use type_name instead of literals in LiteralType::parse

ljedrz:
As discussed here: https://github.com/AleoNet/snarkVM/pull/2454#issuecomment-2120937539. It's a simple refactoring, and due to the use of `const fn` there will be no difference in performance.


vicsn (approved):
When CI is happy I'm happy

howardwu (approved):

PR #2460, opened 2024-05-22
[Perf] Faster Poseidon permutation and Field multiplication

ljedrz:
This was something that I found while heap-profiling a large deployment and execution; `Poseidon::hash_many` is prominent, and responsible for many allocations. Most of these are caused by the `pow` operations involved, and by adjusting one of their internals (switching from `Mul` to `MulAssign` specifically), we can avoid some of them. The preexisting `Mul` operation is also adjusted for a small improvement.

In addition, the `new_state` allocation used in `apply_mds` can be reused, and its contents can be swapped with - instead of cloned to - the `state`.

These changes result in a decrease in allocs/s by ~**6%** (likely more if adjusted for node startup) in a `--dev` node during a large deployment+execution. In addition, the new `mul_assign` operation is ~**36%** faster than the `mul` one, which will positively impact node performance.

Note: we currently don't seem to have a benchmark for `Field` multiplication (I tested it by introducing an ad-hoc benchmark for `MulAssign<&Field<E>> for Field<E>`); I can add one or more of those if we'd like.

### DIFF
--- a/circuit/environment/src/helpers/linear_combination.rs
+++ b/circuit/environment/src/helpers/linear_combination.rs
@@ -17,7 +17,7 @@ use snarkvm_fields::PrimeField;
 
 use core::{
     fmt,
-    ops::{Add, AddAssign, Mul, Neg, Sub},
+    ops::{Add, AddAssign, Mul, MulAssign, Neg, Sub},
 };
 
 // Before high level program operations are converted into constraints, they are first tracked as linear combinations.
@@ -433,14 +433,10 @@ impl<F: PrimeField> Mul<&F> for LinearCombination<F> {
     fn mul(self, coefficient: &F) -> Self::Output {
         let mut output = self;
         output.constant *= coefficient;
-        output.terms = output
-            .terms
-            .into_iter()
-            .filter_map(|(v, current_coefficient)| {
-                let res = current_coefficient * coefficient;
-                (!res.is_zero()).then_some((v, res))
-            })
-            .collect();
+        output.terms.iter_mut().for_each(|(_v, current_coefficient)| {
+            *current_coefficient *= coefficient;
+        });
+        output.terms.retain(|(_v, res)| !res.is_zero());
         output.value *= coefficient;
         output
     }
@@ -460,6 +456,17 @@ impl<F: PrimeField> Mul<&F> for &LinearCombination<F> {
 
     fn mul(self, coefficient: &F) -> Self::Output {
         self.clone() * coefficient
+    }
+}
+
+impl<F: PrimeField> MulAssign<&F> for LinearCombination<F> {
+    fn mul_assign(&mut self, coefficient: &F) {
+        self.constant *= coefficient;
+        self.terms.iter_mut().for_each(|(_v, current_coefficient)| {
+            *current_coefficient *= coefficient;
+        });
+        self.terms.retain(|(_v, res)| !res.is_zero());
+        self.value *= coefficient;
     }
 }
 
--- a/circuit/environment/src/helpers/linear_combination.rs
+++ b/circuit/environment/src/helpers/linear_combination.rs
@@ -433,10 +433,14 @@ impl<F: PrimeField> Mul<&F> for LinearCombination<F> {
     fn mul(self, coefficient: &F) -> Self::Output {
         let mut output = self;
         output.constant *= coefficient;
-        output.terms.iter_mut().for_each(|(_v, current_coefficient)| {
-            *current_coefficient *= coefficient;
-        });
-        output.terms.retain(|(_v, res)| !res.is_zero());
+        output.terms = output
+            .terms
+            .into_iter()
+            .filter_map(|(v, current_coefficient)| {
+                let res = current_coefficient * coefficient;
+                (!res.is_zero()).then_some((v, res))
+            })
+            .collect();
         output.value *= coefficient;
         output
     }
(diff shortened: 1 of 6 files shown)


circuit/environment/src/helpers/linear_combination.rs
```diff
@@ -463,6 +459,17 @@ impl<F: PrimeField> Mul<&F> for &LinearCombination<F> {
...
+    fn mul_assign(&mut self, coefficient: &F) {
+        self.constant *= coefficient;
+        self.terms.iter_mut().for_each(|(_v, current_coefficient)| {
+            *current_coefficient *= coefficient;
+        });
+        self.terms.retain(|(_v, res)| !res.is_zero());
+        self.value *= coefficient;
+    }
+}
+
```
iamalwaysuncomfortable: This struct in the `circuit` crate is a central object in constructing the R1CS so there's zero room for error here.  

At the very least, there needs to be several unit tests written around this operation to ensure the correct amount of variables and constraints come out of this. Can you write unit tests surrounding it both here and in the mul operation that's employing this change? 

It should also be ensured any tests that employ test vectors which explicitly or implicitly employ the `mul` operation are re-run (if they're not done so by default in CI).

iamalwaysuncomfortable (changes requested):
> Note: we currently don't seem to have a benchmark for Field multiplication (I tested it by introducing an ad-hoc benchmark for MulAssign<&Field<E>> for Field<E>); I can add one or more of those if we'd like.

I think this would be really helpful if you can add these benchmarks. The changes here are to a very core set of operations so it would be good to see if the performance improvements are worth it.

Also since these operations are so core to the circuit construction, we need to ensure they're correct. Some tests around the `MulAssign` implementation and then the `mul` operation that uses them should be added before this is merged.

All that said, I personally appreciate thought being put into where performance gains can be had!

circuit/environment/src/helpers/linear_combination.rs
ljedrz: While I believe that this change (`filter_map` changed to `for_each`+`retain`) doesn't result in any changes to crypto logic (i.e. the order and number of operations involved), I agree that it should be reviewed carefully :+1:.

Since I'm not a crypto expert, I will let the other team members decide if the existing tests are sufficient or if we need more of them (and how they would need to look like).

In the meantime, I'll be happy to add a relevant benchmark.

ljedrz:
I returned the original `Mul` code for now to reduce the scope of changes; the most important change is the fact that we avoid cloning the `LinearCombination`.

circuit/types/field/src/mul.rs
```diff
@@ -59,7 +59,7 @@ impl<E: Environment> MulAssign<Field<E>> for Field<E> {
 impl<E: Environment> MulAssign<&Field<E>> for Field<E> {
     fn mul_assign(&mut self, other: &Field<E>) {
         match (self.is_constant(), other.is_constant()) {
-            (true, true) | (false, true) => *self = (&self.linear_combination * *other.eject_value()).into(),
+            (true, true) | (false, true) => self.linear_combination *= &*other.eject_value(),
```
ljedrz, line 62: This change allows us to go from `MulAssign<&Field<E>> for Field<E>` to the new `MulAssign<&F> for LinearCombination<F>` instead of going through `Mul<&F> for &LinearCombination<F>` (and ultimately `Mul<&F> for LinearCombination<F>`), which clones the left-hand side of this operation.

ljedrz:
As requested, I added a benchmark for `Field::mul_assign`, which is the operation affected by this PR; these are the results, and they are quite consistent:
```
Field::mul_assign       time:   [43.078 ns 43.184 ns 43.301 ns]
                        change: [-33.164% -32.969% -32.787%] (p = 0.00 < 0.05)
                        Performance has improved.
Found 1 outliers among 100 measurements (1.00%)
  1 (1.00%) high mild
```

circuit/environment/src/helpers/linear_combination.rs
ljedrz: @iamalwaysuncomfortable I've [updated](https://github.com/AleoNet/snarkVM/pull/2460/commits/79f2a1309ed0a9d5e1122b2c4c0be4d912459ae6) the PR so that it doesn't affect the existing `Mul` impl, and uses a virtually identical implementation for the newly introduced `MulAssign` :+1:.

vicsn: I agree we should make sure this is covered by tests. We currently test for a wide range of gadgets, incl. Poseidon, whether the number of constraints and variables remain unchanged, as well as that the final `value()` encoded in the circuit equals the native result value. For examples you can search for `assert_scope!`. Assuming that the hash gadgets are sound, any change which `MulAssign` introduces (e.g. multiplying by different values) should also reflect in the final value. So I think we're covered and no additional tests are needed.

@acoglio I can't add you as a reviewer but do you want to have a look as well?

howardwu:
@ljedrz does this change any circuit constructions?

Can you please open this on `AleoHQ/snarkVM` first for thorough review prior to introducing it on `AleoNet`?

vicsn:
> @ljedrz does this change any circuit constructions?

I believe this does not change circuit constructions, and [also think our tests will catch it if it does](https://github.com/AleoNet/snarkVM/pull/2460#discussion_r1611609392). Mike and Alessandro should independently review, we should burn-in.

@ljedrz given how understandably concerned this makes everyone for "just" a ~6% improvement, I'd argue let's stall this PR until post-mainnet.

ljedrz:
Closing this PR in order to have it reviewed in `AleoHQ` first :+1:.

PR #2461, opened 2024-05-22
[Perf] Use SmallVec in the LinearCombination

ljedrz:
Despite the tiny diff and the somewhat conservative approach, this change has a large impact on deployment+execution runs. In a local test run using the tx cannon, `--dev` nodes, and a large transaction, it decreased the allocs/s in the target node by ~**32%**; it also made the 33-64B "allocation bucket" much less of an outlier in terms of the alloc count, which should improve the performance of specialized allocators like `jemalloc`.

I'm marking this PR as a draft, as I'd like to double-check it with a devnet run first.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -2676,6 +2676,7 @@ dependencies = [
  "num-traits",
  "once_cell",
  "serial_test",
+ "smallvec",
  "snarkvm-algorithms",
  "snarkvm-circuit",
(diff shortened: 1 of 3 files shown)


ljedrz:
Closing this one for now, as the results are inconclusive under high load; in addition, these objects are a bit more long-lived than I'd like, making `SmallVec` less attractive.

PR #2462, opened 2024-05-22
[Fix] Add `num_variables` to verifying key serialization

raychu86:
## Motivation

This PR updates the verifying key bytes in `parameters` to include the new `num_variables` in serialization. Proving keys are unchanged and untouched.

This is mainly for consistency and correctness; snarkVM ran without errors even without this change because it uses the underlying `VarunaVerifyingKey` instead of the `VerifyingKey` wrapper (that includes the `num_variables`) for the credits and inclusion keys. This is because these specific keys are first-class citizens of the protocol.


vicsn (approved):
Was this simply the result of running the existing generation script?

iamalwaysuncomfortable (approved):
Thanks for this. All tools that build transactions (any sdks, tx-cannon, clis, etc.) should likely be based on this rev until we're able to land future PRs into mainnet-staging

iamalwaysuncomfortable:
@zosorock would you be able to run CI on this? This is an important one to merge since it has some crucial fixes within it.

howardwu (approved):

zosorock:
> @zosorock would you be able to run CI on this? This is an important one to merge since it has some crucial fixes within it.

Yup, just fixed the CI on you guy's fork, and it is running now:
https://app.circleci.com/pipelines/github/AleoHQ/snarkVM/6

PR #2472, opened 2024-05-30
[Chore] Update documentation in `atomic_speculate`

byod8:
## Motivation

It's possible that a `split` transaction have no `fee`. So this is not a foundational bug we need to abort the transaction here.


synthesizer/src/vm/finalize.rs
```diff
@@ -413,9 +413,16 @@ impl<N: Network, C: ConsensusStorage<N>> VM<N, C> {
...
-                                None => Err("Rejected execute transaction has no fee".to_string()),
+                                // This can only happen when the transaction is `credits.aleo/split`
+                                None => {
+                                    // Note: On failure, skip this transaction, and continue speculation.
+                                    #[cfg(debug_assertions)]
+                                    eprintln!("Execution failed without a fee");
+                                    // Store the aborted transaction.
+                                    aborted.push((transaction.clone(), "Execution failed without a fee".to_string()));
+                                    // Continue to the next transaction.
+                                    continue 'outer;
```
iamalwaysuncomfortable: It is potentially conceivable that an invalid `split` transaction might end up triggering this code if the transaction gets this far (previous `check_transaction` checks should take care of this), but this case should be handled by explicitly checking the transaction is `credits.aleo/split`

The way this code is written it avoids the above safety mechanism entirely, if someone manages to get a fee-less transaction into a finalize scope, the entire finalize scope will be executed without a fee which is a DoS vector and part of the reasoning why the entire batch gets aborted. 

`// This can only happen when the transaction is credits.aleo/split` - basically there are no guaruntees that this will be true.

I'd advise NOT to remove this safety mechanism, to put explicit checks for `credits.aleo/split` and only abort the transaction is that function, otherwise to the original abortion of the entire batch.

iamalwaysuncomfortable (changes requested):
Appreciate the community contribution! However, this PR avoids the original safety mechanism entirely. I've left comments about how to modify the original approach to ensure the original safety mechanism is left in place.

synthesizer/src/vm/finalize.rs
vicsn: Small correction: the foundational bug is regarding when `finalize_execution` returns an `Err`, the transaction is expected to be rejected and it is expected to have a Fee. 

So under no condition should this code be refactored to allow abortion for specific program calls.

However, a refactor is likely possible which would make this correct-by-construction and handled by the compiler, so its impossible to hit this codepath.

I suggest for now ANF can close the PR

byod8: @vicsn The `Fee` field of `Transaction` is optional because `credits.aleo/split` transaction can be valid without a fee. So it's possible a `credits.aleo/split` transaction in `finalize_execution` returns an `Err` but it doesn't have a fee. Should this be considered as a foundational bug?

vicsn: I see, apologies to you both for not reading more closely. I'd lke to follow up here after I do some extra research and understand better whether `credits.aleo/split` should remain free.

vicsn: Just to confirm: `credits.aleo/split` does not have a Fee, but it does consume public account balance within its aleo instructions logic.

Can you please add (or modify) a unit test to showcase that the abortion now happens as expected? Apologies for the extra nuisance, but its proven useful to catch unanticipated behaviour. You can see some recent examples of adjusted tests in `ledger/src/tests.rs` in: https://github.com/ProvableHQ/snarkVM/pull/13/files

byod8: After trying to add a test, I find it rarely get there. This is because `credits.aleo/split` is a simple transition without finalize operation. If it is invalid, it will be aborted by `prepare_for_speculate` function. It seems impossible that a `credits.aleo/split` transaction fails in `finalize_execution` in `atomic_speculate`.

vicsn: Perhaps it can fail in finalize if the spender's public account balance is spent by a previous transaction (in the same block)

byod8: It seems that `credits.aleo/split` only consume private fee in the input record?

vicsn: You are right, I should have spotted that, makes sense.

Thank god for unit testing, I guess that means the logic was correct all along and this PR can be closed.

byod8: Yes, I will revert the code changes and add comment on that.

vicsn (approved):

byod8:
Hi @iamalwaysuncomfortable, please review.

PR #2478, opened 2024-06-06
Authorization deserialization fix

lukenewman:
## Motivation

When deserializing an `Authorization` from string a call stack that included multiple transitions, the structural assertion logic would erroneously fail. This PR fixes the assertion logic but does not change the underlying structure/order of the `Authorization` fields.

## Test Plan

Our (Puzzle) delegated proving stack takes in user-generated `Authorization`s in string form. We then parse these and execute the proof request using snarkVM crates. I've tested that these `Authorization` deserializations and subsequent proofs do indeed land onchain properly.

I can provide screenshots/videos on request. For now, you can see the following transaction that was generated through our stack and has landed onchain: https://testnet.aleoscan.io/transaction?id=at14d4edujs6ztghev2zppp55v4xlffu7q3p379dlvdsu3yuu2cluysd84ktk

I am also willing to create a sample project showing this fix in action.

A more formal test plan would involve embedding a multi-transition `Authorization` unit test into snarkVM itself and not just relying on a `fee_public` test case.

## Related PRs

none


lukenewman:
based this off `mainnet` instead of `mainnet-staging` -- closing this and will open a subsequent

PR #2479, opened 2024-06-06
Fix deserialization of authorizations with multiple transitions

lukenewman:
## Motivation

When deserializing an `Authorization` from string a call stack that included multiple transitions, the structural assertion logic would erroneously fail. This PR fixes the assertion logic but does not change the underlying structure/order of the `Authorization` fields.

## Test Plan

Added two unit tests for authorization deserializations:
* single transition is in `synthesizer/process/src/stack/authorization/mod.rs`
* multiple transitions is in `synthesizer/src/vm/mod.rs`

I'm happy to relocate the latter to `authorization/mod.rs` but stuck it in `vm/mod.rs` since it needs to deploy several programs to the test VM.

## Related PRs

none


vicsn:
Great find! 

> embedding a multi-transition Authorization unit test

Yes please. You can see some recent examples of adjusted tests in `ledger/src/tests.rs` in: https://github.com/ProvableHQ/snarkVM/pull/13/files . There are also a lot in `synthesizer/src/vm/mod.rs`

lukenewman:
@vicsn this is good for another pass! added unit tests

synthesizer/process/src/stack/authorization/mod.rs
```diff
@@ -62,8 +62,13 @@ impl<N: Network> TryFrom<(Vec<Request<N>>, Vec<Transition<N>>)> for Authorizatio
...
             transitions.len()
         );
+        // Move the first request to the back in order to match the transitions.
+        let mut requests_deque = VecDeque::from(requests.clone());
+        if let Some(first_request) = requests_deque.pop_front() {
+            requests_deque.push_back(first_request);
+        }
         // Ensure the requests and transitions are in order.
-        for (index, (request, transition)) in requests.iter().zip_eq(&transitions).enumerate() {
+        for (index, (request, transition)) in requests_deque.iter().zip_eq(&transitions).enumerate() {
```
niklaslong: I'm not yet familiar with the full context behind this change (it's rather odd these are not ordered the same way by default) but just looking at the rust, you could probably do something like this to avoid building a second `VecDeque` and calling `clone`, `pop` and `insert`:
```suggestion
        for (index, (request, transition)) in requests.iter().skip(1).chain(requests.iter().take(1)).zip_eq(&transitions).enumerate() {
```

lukenewman:
@niklaslong thanks for the feedback! it's been addressed.

niklaslong:
I've had a dig around and we seem to:
1. order requests from parent to child: they are passed in when constructing the `Authorization`
2. order transitions from child to parent, as inner calls are processed first and added one by one to the `Authorization`

This means the mismatch isn't due to deserialisation but rather the expectation that requests and transitions be ordered in the same way within an `Authorization`.  In other words, the reordering proposed in this PR might not be sufficient, should there be more than two nested calls.

Perhaps we could either establish the correspondence between requests and transitions using the program id in the check or (more simply) rely on the collections being in reverse order relative to each other for the time being?

lukenewman:
alright @niklaslong @vicsn we're now matching based on the transition commitment

synthesizer/process/src/stack/authorization/mod.rs
```diff
@@ -65,10 +68,21 @@ impl<N: Network> TryFrom<(Vec<Request<N>>, Vec<Transition<N>>)> for Authorizatio
             requests.len(),
             transitions.len()
         );
```
d0cd, line 74: A comment at the top of this method documenting why this technique is needed would be helpful. Something along the lines of:
```
Requests are added to an authorization in the pre-order traversal sequence (parents before children), while transitions are added to an authorization in the post-order traversal sequence (children before parents). 
This method takes in as input requests and transitions, in pre-order and post-order traversal sequence respectively, which is why the requests and transitions need to be checked independent of the ordering
```

d0cd: Assuming my understanding is correct :)

lukenewman:
@d0cd thanks for the review! addressed your feedback and also simplified the `tcm_indices` creation

niklaslong (approved):

PR #2484, opened 2024-06-07
Make CI 3x faster

vicsn:
## Motivation

Already approved and tested here: https://github.com/ProvableHQ/snarkVM/pull/8

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -709,7 +709,7 @@ jobs:
     resource_class: << pipeline.parameters.twoxlarge >>
     steps:
       - run_serial:
-          flags: --test '*' -- --skip keccak --skip psd --skip sha
+          flags: --test '*' -- --skip keccak --skip psd --skip sha --skip instruction::is --skip instruction::equal --skip instruction::commit
           workspace_member: synthesizer/program
           cache_key: snarkvm-synthesizer-program-cache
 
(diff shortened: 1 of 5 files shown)


.circleci/config.yml
```diff
@@ -743,6 +743,36 @@ jobs:
           workspace_member: synthesizer/program
           cache_key: snarkvm-synthesizer-program-cache
 
+  synthesizer-program-integration-instruction-is:
+    docker:
+      - image: cimg/rust:1.76.0 # Attention - Change the MSRV in Cargo.toml and rust-toolchain as well
+    resource_class: << pipeline.parameters.2xlarge >>
```
zosorock: we need to change these from `2xlarge` to `twoxlarge` - here and in other locations (lines 759 and 769)

.circleci/config.yml
```diff
@@ -743,6 +743,36 @@ jobs:
...
+    steps:
+      - run_serial:
+          flags: instruction::is --test '*'
+          workspace_member: synthesizer/program
+          cache_key: snarkvm-synthesizer-program-cache
+
+  synthesizer-program-integration-instruction-equal:
+    docker:
+      - image: cimg/rust:1.76.0 # Attention - Change the MSRV in Cargo.toml and rust-toolchain as well
+    resource_class: << pipeline.parameters.2xlarge >>
```
zosorock: we need to change these from `2xlarge` to `twoxlarge`

.circleci/config.yml
```diff
@@ -743,6 +743,36 @@ jobs:
...
+    steps:
+      - run_serial:
+          flags: instruction::equal --test '*'
+          workspace_member: synthesizer/program
+          cache_key: snarkvm-synthesizer-program-cache
+
+  synthesizer-program-integration-instruction-commit:
+    docker:
+      - image: cimg/rust:1.76.0 # Attention - Change the MSRV in Cargo.toml and rust-toolchain as well
+    resource_class: << pipeline.parameters.2xlarge >>
```
zosorock: we need to change these from `2xlarge` to `twoxlarge`

zosorock (changes requested):

synthesizer/program/tests/instruction/assert.rs
```diff
@@ -28,7 +28,7 @@ use synthesizer_process::{Process, Stack};
 type CurrentNetwork = MainnetV0;
 type CurrentAleo = AleoV0;
 
-const ITERATIONS: usize = 100;
+const ITERATIONS: usize = 25;
```
gluax, line 31: These types of tests should be avoided in CI, where we iterate and do random values. These randoms breaks unit and integration test philosophy, and also cause the CI time to increase. Unit tests should try to cover all edge cases, while random/fuzz tests are best done on a dedicated machine off CI. If a random/fuzz test does find something you would then convert it to a unit test.

gluax (approved):
Still approving for urgency, but how testing is being done should be reconsidered.

evanmarshall (commented):
It's not clear to me why any of the values initially existed and why this new value is lower but still sufficiently acceptable. 

Any reason we shouldn't just set them to `1` or `5` or `10`?

Meshiest (approved):

voximity (approved):

zosorock (approved):

vicsn:
> Any reason we shouldn't just set them to 1 or 5 or 10?

No valid reason, agree with @gluax ' [comment](https://github.com/AleoNet/snarkVM/pull/2484#discussion_r1633980016) and its worth investigating a refactor in the future. Somewhat related: there are some heavy `#[ignore]`'d tests which could potentially also fit in a separate fuzzing / CI / integration testing flow.

PR #2486, opened 2024-06-07
Increase mainnet MAX_CERTIFICATES, and thereby max number of validators

vicsn:
## Motivation

CI passed and already approved here: https://github.com/ProvableHQ/snarkVM/pull/14


Meshiest (approved):

gluax (approved):

evanmarshall (approved):
This is good but I feel like we need a comment that's like: "Here's the think you change if you want change the maximum number of validators"

voximity (approved):

PR #2487, opened 2024-06-08
Initializes the restrictions list for each network

vicsn:
## Motivation

Mirror of: https://github.com/ProvableHQ/snarkVM/pull/15
Passed review and CI.


Meshiest (approved):
These changes look good assuming they are used with caution. I suspect if a restriction is added retroactively (on block ranges that have already been added to the ledger), it will corrupt the mappings for newly syncing nodes and result in a desync between those and currently synced nodes.

vicsn:
> I suspect if a restriction is added retroactively (on block ranges that have already been added to the ledger), it will corrupt the mappings for newly syncing nodes and result in a desync between those and currently synced nodes.

Caution indeed advised! But the above scenario is not possible because the node with updated list will not connect and sync blocks from nodes with other lists: https://github.com/AleoNet/snarkOS/pull/3306

console/network/src/canary_v0.rs
```diff
@@ -152,6 +152,11 @@ impl Network for CanaryV0 {
         snarkvm_parameters::canary::GenesisBytes::load_bytes()
     }
 
+    /// Returns the restrictions list as a JSON-compatible string.
+    fn restrictions_list_as_str() -> &'static str {
```
evanmarshall, line 156: I'm not sure about the naming here. I think I'd prefer `restrictions_list()` or `restrictions_list_json()` to be more clear about the intent.

console/network/src/lib.rs
```diff
@@ -215,6 +215,9 @@ pub trait Network:
     /// Returns the genesis block bytes.
     fn genesis_bytes() -> &'static [u8];
 
+    /// Returns the restrictions list as a JSON-compatible string.
+    fn restrictions_list_as_str() -> &'static str;
```
evanmarshall, line 219: Same comment here about the naming.

console/network/src/mainnet_v0.rs
```diff
@@ -149,6 +149,11 @@ impl Network for MainnetV0 {
         snarkvm_parameters::mainnet::GenesisBytes::load_bytes()
     }
 
+    /// Returns the restrictions list as a JSON-compatible string.
+    fn restrictions_list_as_str() -> &'static str {
```
evanmarshall, line 153: Same here about the naming

console/network/src/testnet_v0.rs
```diff
@@ -152,6 +152,11 @@ impl Network for TestnetV0 {
         snarkvm_parameters::testnet::GenesisBytes::load_bytes()
     }
 
+    /// Returns the restrictions list as a JSON-compatible string.
+    fn restrictions_list_as_str() -> &'static str {
```
evanmarshall, line 156: Same here about the naming

ledger/benches/bonded_mapping.rs
```diff
@@ -22,12 +22,13 @@ use console::{
     program::{Identifier, Literal, Plaintext, ProgramID, Value},
 };
 use ledger_committee::{MAX_DELEGATORS, MIN_DELEGATOR_STAKE};
+use ledger_store::ConsensusStore;
+use synthesizer::VM;
```
evanmarshall, line 26: Should we add dependency ordering to the pre-commit hooks? Not here but perhaps as a follow up issue.

synthesizer/src/restrictions/mod.rs
```diff
@@ -0,0 +1,386 @@
...
+
+impl<N: Network> Restrictions<N> {
+    /// Initializes the `Restrictions` instance for the current network.
+    pub fn load() -> Result<Self> {
+        // Load the restrictions list from the network.
+        let restrictions = Self::from_str(N::restrictions_list_as_str())?;
+        // Ensure the restrictions ID matches the computed value.
+        let expected_restrictions_id =
+            Self::compute_restrictions_id(&restrictions.programs, &restrictions.functions, &restrictions.arguments)?;
+        if restrictions.restrictions_id != expected_restrictions_id {
```
evanmarshall, line 69: It seems like we could leave the `restrictions_id` out of the json serialized file. It seems like we need either: a script to generate the restrictions_id from `programs`, `functions`, and `arguments` or we could just omit the `restrictions_id` from the json and calculate it.

synthesizer/src/restrictions/mod.rs
```diff
@@ -0,0 +1,386 @@
...
+    /// e.g. `restricted.aleo/bar _ aleo1zkpxxxxx _ _` => `..10` (up to block 10)
+    /// e.g. `restricted.aleo/bar _ aleo1zkpxxxxx _ _` => `10..20` (from block 10 to block 20)
+    ///
+    /// Note: This design intentionally minimizes the number of total lookups required to check
+    /// for restrictions when a transition matches the described profile. In summary:
+    /// - When a transition does not match the program ID or function name, the total lookup cost is `O(1)`.
+    /// - When a transition matches the program ID & function name, the initial lookup cost is `O(num_inputs + num_outputs)`.
+    ///    - If an input or output index does not match, the additional lookup cost is `0`.
+    ///    - If an input or output index matches, the additional lookup cost is `O(n)` for `n` restricted arguments with the same index.
+    arguments: IndexMap<Locator<N>, IndexMap<ArgumentLocator, IndexMap<Literal<N>, BlockRange>>>,
```
evanmarshall, line 58: How would we handle the case where we want to ban an address from doing anything? It seems structurally more difficult than these checks. I imagine our best option right now would be to restriction `credits.aleo/fee` for some address but this is somewhat easily bypassed as the fee payer can be a different signer.

evanmarshall (dismissed):
Overall looks good to me. A couple small changes & a couple issues to file from these changes.

Meshiest (dismissed):

Meshiest:
> But the above scenario is not possible because the node with updated list will not connect and sync blocks from nodes with other lists: [AleoNet/snarkOS#3306](https://github.com/AleoNet/snarkOS/pull/3306)

1. validators sync to block 30k
2. restriction added on block range 29k... for transfers for a certain address
3. all nodes restart with the new restrictions
4. incoming clients (or 1k behind validators) sync up to 29k and start hitting the newly restricted transfers
5. validators that were at 30k have old mappings that do not reflect the restrictions, new clients/lagging validators have different mappings that reflect the restricted transactions

vicsn:
Thank you for explaining @Meshiest . Indeed that sounds like a serious UX issue. 

To further my own understanding and documentation: given that validators frequently lag multiple blocks behind one another even in a healthy network, with the current design "social consensus" would have to enforce restrictions **at least** 10 blocks into the future when any network restart happens which updates the restrictions list. Enforcing such a delay in code is not possible because a validator can't know independently if its behind tip or not, especially because other validators might have already been switched off for the restart.

console/network/src/canary_v0.rs
vicsn: Appreciate the suggestion, but let's leave it as is for now.

ledger/benches/bonded_mapping.rs
vicsn: I think the more style-enforcing pre-commit hooks the better!

synthesizer/src/restrictions/mod.rs
vicsn: I agree its superfluous, and the UX benefit for certain scenario's might be achievable in many other ways (a restrictions_id script, a checksum utility, ...). But I assume Howard put thought into it and discussed with ANF, so let's leave it unless there is a specific vulnerability.

synthesizer/src/restrictions/mod.rs
vicsn: Indeed that would be a nice extension. A first step within the current design could be to extend the check against plaintext literal inputs to also plaintext container inputs. But there's a large design space ahead.

voximity (approved):

damons (approved):
Approved under duress.  I don't like allowing features which have not been tested.  To me, If it doesn't have a test, it doesn't work.

apruden2008:
> Approved under duress. I don't like allowing features which have not been tested. To me, If it doesn't have a test, it doesn't work.

Working on testing in canarynet now, per discussion with @damons

PR #2493, opened 2024-06-12
[Feature] Support historical logging for `delegated`, `bonded, and `unbonding` mappings.

d0cd:
This PR supports historical logging for the `delegated`, `bonded`, and `unbonding` mappings in `credits.aleo`.

If the `history` feature is enabled, the VM will store the above mappings as JSON to a set local directory (the same directory as the `.ledger`).

This data is stored and updated as follows:
- A top-level `.history*` directory is created at the storage path.
- When a block `i` is added to the ledger, a subdirectory `block-<i>` is created containing: `block-<i>-delegated.json`, `block-<i>-bonded.json`, and `block-<i>-unbonding.json`.
- If the history directory is deleted, it is created once again.
- Data is stored as a JSON string for easier consumption. We considered storing it at bytes, but opted for usability over compression.

Pruning and management of the `.history` directory is left to the node runner. 

Note. The entirety of this PR is gated under the `history` feature.

[CI passed mostly here](https://app.circleci.com/pipelines/github/ProvableHQ/snarkVM/124/workflows/a2a9c330-035f-4016-a094-3873d39df80f). The failing tests are succeeding locally and unrelated to this PR.


synthesizer/src/vm/helpers/history.rs
```diff
@@ -0,0 +1,108 @@
...
+    {
+        // Get the path to the block directory.
+        let block_path = self.path.join(format!("block-{height}"));
+        // Create the block directory if it does not exist.
+        if !block_path.exists() {
+            std::fs::create_dir_all(&block_path)?;
+        }
+
+        // Write the entry to the block directory.
+        let entry_path = block_path.join(format!("block-{height}-{mapping}.json"));
```
vicsn: https://stackoverflow.com/questions/466521/how-many-files-can-i-put-in-a-directory

synthesizer/src/vm/finalize.rs
```diff
@@ -1258,6 +1258,22 @@ impl<N: Network, C: ConsensusStorage<N>> VM<N, C> {
...
                     // Insert the next committee into storage.
                     store.committee_store().insert(state.block_height(), next_committee)?;
+
+                    #[cfg(feature = "history")]
+                    {
+                        // Load a `History` object.
+                        let history = History::new(N::ID, store.storage_mode().clone());
+
+                        // Write the delegated mapping as JSON.
+                        history.store_mapping(state.block_height(), MappingName::Delegated, &next_delegated_map)?;
```
vicsn: Can we make a back-of-the-envelope calculation how much disk space the 3 mappings require? Imagine the worst case 100_000 delegators, wouldn't this already be *at least* 3MiB? And at 5 seconds per block, that would be 18 TiB per year? This number might be too optimistic (because JSON is bigger than 3MiB) or too pessimistic (because there won't be 100_000 delegators that quickly.

I think the only way to overcome this is to only store incremental data...

synthesizer/src/vm/finalize.rs
```diff
@@ -1258,6 +1258,22 @@ impl<N: Network, C: ConsensusStorage<N>> VM<N, C> {
 
                     // Insert the next committee into storage.
                     store.committee_store().insert(state.block_height(), next_committee)?;
+
+                    #[cfg(feature = "history")]
```
vicsn: During "normal" operation of the validators, I see `Failed to post-ratify` printed quite regularly. Try running a simple 4 validator network - potentially it only shows up if you restart 1 node so its syncing (instead of participating in consensus). This was also discussed here: https://github.com/AleoNet/snarkOS/pull/3268#issuecomment-2152708302

So then either:
- the stored data should be deleted/cleaned up when post ratify fails
- OR we only store the data in a later stage, e.g. when check_next_block succeeds, and perform an extra pass over the block's succeeded transactions to store the mapping diffs that it has introduced. Not sure if this is possible.

synthesizer/src/vm/finalize.rs
d0cd: We considered this in our design process.
An incremental data solution would have a smaller footprint, but require much more complicated indexing logic. 
In favor of simplicity, we leave management of the `history` directory to the node operators.

synthesizer/src/vm/finalize.rs
d0cd: In your observations, what happened next? How did the node recover and move forward?
The reason I ask is because if it recovers, then a correct block will be written and the history will be updated accordingly.

Performance-wise, this is the best place to store the data since you already have the `bonded` and `delegated` mappings in memory. Moving this later would require pulling them from DB.

synthesizer/src/vm/helpers/history.rs
d0cd: Great observation! We'll opt for a 2 tier storage system, each with up to 2^16 files. 
That is sufficient for all blocks and under the limits of most modern server OSes.

synthesizer/src/vm/finalize.rs
vicsn: > if it recovers, then a correct block will be written and the history will be updated accordingly.

Ah that's great, yes that solves the issue

synthesizer/src/vm/helpers/history.rs
```diff
@@ -0,0 +1,123 @@
...
+        // Get the path the directory group.
+        let group = Self::group(height);
+        let path = self.path.join(format!("group-{group}"));
+        // Get the path to the block directory.
+        path.join(format!("block-{height}"))
+    }
+
+    // A helper function to calculate the group number for a given block height.
+    fn group(height: u32) -> u32 {
+        height.saturating_div(2u32 << 16)
```
vicsn:
```suggestion
        height.saturating_div(1u32 << 16)
```
Or perhaps even clearer, use: `u16::MAX`

vicsn (dismissed):
LGTM!

As for CI ([thx for pushing this already!](https://app.circleci.com/pipelines/github/ProvableHQ/snarkVM/123/workflows/e0f0949a-3d1f-4fec-9bfb-0c53973b181e)):
Can you also push a clippy fix? I assume maybe a new Rust version came out.
Can you also run the synthesizer tests locally to double check they pass?

Meshiest (dismissed):
LGTM. We use a tool that uses a similar mechanism that captures the raw binary values of all mappings and they are quite compact (under 4-16KB per block)

d0cd:
> LGTM!
> 
> As for CI ([thx for pushing this already!](https://app.circleci.com/pipelines/github/ProvableHQ/snarkVM/123/workflows/e0f0949a-3d1f-4fec-9bfb-0c53973b181e)): Can you also push a clippy fix? I assume maybe a new Rust version came out. Can you also run the synthesizer tests locally to double check they pass?

Done and the synthesizer tests passed locally.

d0cd:
> LGTM. We use a tool that uses a similar mechanism that captures the raw binary values of all mappings and they are quite compact (under 4-16KB per block)

Do you have opinions on whether the data should be stored as raw JSON strings or as bytes?

apruden2008:
We ran our own internal CI against this and is passing. Given the fact this is gated behind a CLI flag, I think this is a safe feature to add in here at this point in the development timeline. Thank you @Meshiest and @damons for your review

apruden2008 (commented):
See comment above. LGTM based on internal CI passing and prior reviews both inside and outside of Provable

HarukaMa:
I just saw this and I'd highly recommend anyone who use this to store the jsons in a fs that supports compression, since with a busy network those data could get huge really fast. Had to manually implement a similar function to debug my explorer before. Also one of the reasons I don't keep a whole history of bonded and deletaged mappings - they are _really_ huge on a live network.

PR #2498, opened 2024-06-19
Fix check for bond_validator increasing number of validators beyond M…

evanmarshall:
…AX_COMMITTEE_SIZE

## Summary

This is a bug fix related to ARC-41. It fixes an issue where we were not properly extracting the validator address from the `bond_validator` call. The result is that the check for exceeding the maximum number of validators didn't work.

## Testing

Running the new unit test with the pre-existing confirms that the change ensures the check now works again.

## Release

This shouldn't require a reset or fork. It should be deployable as a patch and validators can update asynchronously.

### DIFF
--- a/synthesizer/src/vm/finalize.rs
+++ b/synthesizer/src/vm/finalize.rs
@@ -1783,7 +1783,6 @@ finalize transfer_public:
         // Initialize a new address.
         let new_validator_private_key = PrivateKey::<CurrentNetwork>::new(rng).unwrap();
         let new_validator_address = Address::try_from(&new_validator_private_key).unwrap();
-        let new_validator_withdrawal_address = Address::<CurrentNetwork>::new(rng.gen());
 
         // Construct the committee.
         // Track the allocated amount.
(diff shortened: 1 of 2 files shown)


vicsn:
The change looks fine to me, but given my lack of context, I'll let Ray review.

> The result is that the check for exceeding the maximum number of validators didn't work.

I did run your new unit test excluding the finalize changes, and it also passed before this PR's changes.

evanmarshall:
> The change looks fine to me, but given my lack of context, I'll let Ray review.
> 
> > The result is that the check for exceeding the maximum number of validators didn't work.
> 
> I did run your new unit test excluding the finalize changes, and it also passed before this PR's changes.

That's my mistake. I tried `true => None` to check if the unit test is working and that does cause the unit test to fail. The reason the unit test passes without any modification is due to the original check actually extracting a valid address, the withdrawal address.

So you could've only bypassed this check if you set your withdrawal address to another pre-existing validator. I'll update the test to catch this specific edge case

raychu86 (approved):
LGTM.

One thing I would note is that this will not require a ledger reset ONLY IF the following conditions do not exist in the last 100 rounds before GC:
1.  there were transactions that bonded multiple validators to the same withdrawal address.
2. If we aborted any `bond_validator` calls due to reaching the `MAX_COMMITTEE_SIZE`. If we are at `MAX_COMMITTEE_SIZE`, then all `bond_validator` calls will be aborted, even if they are valid. Since this logic is considering the withdraw address as a new_validator address.

@evanmarshall An easy way to check the first condition is to check the `withdraw` mapping for each of the validator addresses and sure uniqueness. The second condition is already solved if there are < `MAX_COMMITTEE_SIZE` validators in the committee.

PR #2504, opened 2024-06-26
`RwLock.read()` instead of `RwLock.write()`

ghostant-1017:


ljedrz:
I think I recall these, but counter-intuitively these need to be `write`, since the collection in question is an `LruCache`, no?

Edit: indeed, this is the signature of [`get`](https://docs.rs/lru/latest/lru/struct.LruCache.html#method.get).

ghostant-1017:
> I think I recall these, but counter-intuitively these need to be `write`, since the collection in question is an `LruCache`, no?
> 
> Edit: indeed, this is the signature of [`get`](https://docs.rs/lru/latest/lru/struct.LruCache.html#method.get).

Thank you! You are right.

ljedrz:
That being said, I don't know why that lock was changed to an `RwLock`, as only write guards are used, meaning `Mutex` would be a better fit.

PR #2510, opened 2024-07-01
[Perf] Don't re-synthesize credits.aleo

### LINKED ISSUE #2416: [Optimization] Inefficient fetching of `credits.aleo` PKs
It looks like our VM does not fetch the `credits.aleo` proving keys from the `resources` folder and simply resynthesizes them from scratch at time of use. This is completely safe, however will incur a one-time synthesis cost.

When we fetch the proving keys functions via `Stack::get_proving_key`, we check if the function is a `credits.aleo` function and read the proving key directly from the bytes in `.resources` if it is. This is done in `Stack::execute` - https://github.com/AleoHQ/snarkVM/blob/2cbf34a1010bf781277cdc6ff1ae966230cf97c1/synthesizer/process/src/stack/execute.rs#L462

However a few lines above, we are always re-synthesizing the key at first time of use:  https://github.com/AleoHQ/snarkVM/blob/2cbf34a1010bf781277cdc6ff1ae966230cf97c1/synthesizer/process/src/stack/execute.rs#L425-L429 This ignores the optimization we have for the PKs where we can simply read from bytes instead of resynthesizing.

_Note: The current behavior does **NOT** impact security, but can be optimized._

ljedrz:
Closes https://github.com/AleoNet/snarkVM/issues/2416.

### DIFF
--- a/synthesizer/process/src/stack/execute.rs
+++ b/synthesizer/process/src/stack/execute.rs
@@ -432,8 +432,9 @@ impl<N: Network> StackExecute<N> for Stack<N> {
         if matches!(registers.call_stack(), CallStack::Synthesize(..))
             || matches!(registers.call_stack(), CallStack::Execute(..))
         {
-            // If the proving key does not exist, then synthesize it.
-            if !self.contains_proving_key(function.name()) {
+            // If the proving key does not exist, then synthesize it. This is not needed for `credits.aleo`.
+            if self.program_id() != &ProgramID::from_str("credits.aleo")? && !self.contains_proving_key(function.name())
+            {
                 // Add the circuit key to the mapping.
                 self.synthesize_from_assignment(function.name(), &assignment)?;
                 lap!(timer, "Synthesize the {} circuit key", function.name());


raychu86 (approved):
LGTM, pending CI runs successfully

raychu86 (changes requested):
This change actually causes the `test_vm_execute_and_finalize` on my machine to hang and never finish. Without the change, it completes in around 8 minutes.

 This should be investigated before we proceed with the changes.

ljedrz:
I can reproduce this; that being said, I'm not familiar with the associated plumbing so I'm not sure what the nature of the issue is.

vicsn:
Hmm I can't think immediately of why this would fail. However, given that this optimization is about resynthesis of **proving keys** in `CallStack::Execute` mode, its not something validators run into (who only use CallStack::CheckDeployment), so this seems very low priority

vicsn:
Closing for now because not a priority to fix. Keeping the issue open.

PR #2514, opened 2024-07-09
[Perf] Reserve the exact capacity in ToBits for BigIntegers

ljedrz:
While heap-profiling the loading of the Canary ledger, I was surprised by the number of temporary allocations pointing to the `ToBits` impl for `BigInteger256`; it appears that the `Extend` impl that's used there is not able to take advantage of the `ExactSizeIterator` impl for `BitIteratorLE`, and we need to be more explicit in order to avoid them.

The results at height 330598 are as follows:
- total allocs **-14%**
- temp allocs: **-28%**
- ledger loading time: **-9%**


joske (approved):

ljedrz:
[CI run link](https://app.circleci.com/pipelines/github/ProvableHQ/snarkVM?branch=perf%2Fexact_extend_bigint_to_bits-ci)

vicsn (approved):
Color me surprised as well. Cool find.

PR #2515, opened 2024-07-09
[Perf] Speed up the fetching of the latest block height

ljedrz:
Another heap-profiling find applicable to the loading of the ledger; by counting the number of heights instead of deserializing them, we can avoid a large number of allocations.

The results at height 330598 are as follows:
- total allocs **-12%**
- temp allocs: **-15%**
- ledger loading time: **-7%**

### DIFF
--- a/ledger/store/src/block/mod.rs
+++ b/ledger/store/src/block/mod.rs
@@ -1014,16 +1014,19 @@ impl<N: Network, B: BlockStorage<N>> BlockStore<N, B> {
 
         // Compute the block tree.
         let tree = {
-            // Prepare an iterator over the block heights.
-            let heights = storage.id_map().keys_confirmed();
+            // Find the maximum block height.
+            let max_height = storage.id_map().len_confirmed().checked_sub(1).map(u32::try_from);
             // Prepare the leaves of the block tree.
-            let hashes = match heights.max() {
-                Some(height) => cfg_into_iter!(0..=cow_to_copied!(height))
-                    .map(|height| match storage.get_block_hash(height)? {
-                        Some(hash) => Ok(hash.to_bits_le()),
-                        None => bail!("Missing block hash for block {height}"),
-                    })
-                    .collect::<Result<Vec<Vec<bool>>>>()?,
+            let hashes = match max_height {
+                Some(height) => {
+                    let height = height?;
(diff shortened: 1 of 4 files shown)


ljedrz:
[CI run link](https://app.circleci.com/pipelines/github/ProvableHQ/snarkVM?branch=perf%2Ffaster_max_height-ci)

vicsn (dismissed):

vicsn:
Can you point readers to the logic/evidence that the change does not introduce an off-by-one error?

ledger/store/src/block/mod.rs
```diff
@@ -1360,6 +1360,11 @@ impl<N: Network, B: BlockStorage<N>> BlockStore<N, B> {
         self.storage.id_map().keys_confirmed()
     }
 
+    /// Returns the height of the latest block in the storage.
+    pub fn max_height(&self) -> Option<u32> {
+        u32::try_from(self.storage.id_map().len_confirmed()).ok()?.checked_sub(1)
```
ljedrz: in order to make sure this matches the current logic, there are 2 scenarios to consider:
1. empty ledger: in that case `len_confirmed == 0` (there are no records mapping block height to hash), and `checked_sub(1)` results in `None` being returned (which is equivalent to the current `heights().max()` for an empty `id_map`)
2. `N` blocks (including `1` for just the genesis): `len_confirmed == N`, and `checked_sub(1)` returns `Some(N - 1)`, which is the same as max block height (since they are all unique and consecutive)

in addition, `u32::try_from` will never fail here, as the block height can not have a value greater than `u32::MAX`

ljedrz:
@vicsn done; also, I noticed that originally there actually _would_ have been an issue with a height equal to `u32::MAX` (at which point there are arguably more pressing issues :smiley:), which is now fixed.

vicsn (dismissed):

ljedrz:
I just found 2 more instances of max block height lookup (without using `heights`, which is how I missed them before); I updated the CI branch too.

One of them will further improve ledger loading time.

vicsn (approved):

gluax (approved):

zkxuerb (approved):
LGTM

vicsn (approved):

zklimaleo (approved):

PR #2516, opened 2024-07-10
[Perf] Reduce allocs when parsing Program

ljedrz:
(transferred from https://github.com/ProvableHQ/snarkVM/pull/5, and asking @d0cd for a review as suggested there)

`Program` parsing is something that has always been prominent in node heap profiles, and this PR targets its highest-level bits.

While parsing its components, we currently use a combination of the functions `many1` and `alt`, which cause us to clone the parsed input at least once (more if it's not the expected component) for every component; this can be avoided by removing the use of `alt` in favor of a new intermediate function which is able to determine the type of the parsed component in advance and select the right parsing function without "making guesses". This is what the 1st commit does.

The 2nd commit avoids all the clones of program components that we currently perform presumably as a workaround introduced in order to be able to manage the notoriously tricky `nom` errors. With some refactoring (i.e. applying `map_res` only to potential errors), we are able to remove many allocations.

These changes reduce the number of allocations caused by the parsing of a `Program` in a 15-minute run of a `--dev` node by ~**8%** (an absolute decrease of over 54k allocs), making the process faster. The impact on the performance of the node as a whole depends on the number and size of the `Program`s it has to process.


ljedrz:
[CI run link](https://app.circleci.com/pipelines/github/ProvableHQ/snarkVM?branch=perf%2Fprogram_parsing2-ci)

ljedrz:
Converting to draft for now, as the CI complains :mag:.

Edit: fixed.

gluax (approved):

vicsn:
Superseded by https://github.com/AleoNet/snarkVM/pull/2562 to resolve `cargo fmt` conflicts

PR #2518, opened 2024-07-12
[Perf] Only use raw iterators with RocksDB and speed up ledger load

ljedrz:
The signatures of iteration-related methods (with the `rocksdb` feature) have been haunting my heap profiles for a long time now, and while I initially believed this was something we could fine-tune with some configuration, all such attempts were unsatisfactory. It turns out that the "default" (higher-level) RocksDB iterators are just inherently inefficient, and only the `DBRawIterator` is able to avoid a massive number of allocations, some of which may remain lingering in the RSS and contribute to its inflation over time. The raw iterator is somewhat trickier to use correctly, but since it's what the regular iterator is built upon, there is nothing particularly novel about it, and we're already using it to count records.

With these changes I ran profiling with both an instrumented OS allocator and `jemalloc` and - contrary to the usual results - it was the latter that experienced larger relative differences when loading the ledger:
- total allocs and temp allocs were both reduced by **98%**
- ledger load time was down **14%**
- max RSS was reduced by **13%**

In addition, `heaptrack` ran **73%** faster and produced a **99%** smaller profile, which is very practical for future profiling needs.

Alongside these changes I realized that the current `len_confirmed` method - while working correctly and passing all tests - can be made more solid by including iterator validity checks; the only situation where this would truly be needed is if it was called on the very last map in the database and if the map was empty, but [this commit](https://github.com/AleoNet/snarkVM/commit/f8fc68ca22ce9369f3e82107c48f3738bf00af50) guards against that edge case.

The [final commit](https://github.com/AleoNet/snarkVM/commit/1dbf87418379118a74c2f53fd9f45b444c6c2d6b) is a further improvement on one of the changes from https://github.com/AleoNet/snarkVM/pull/2515 (and made possible with the switch to raw iterators).

[CI run link](https://app.circleci.com/pipelines/github/ProvableHQ/snarkVM?branch=perf%2Fraw_rocks_iter-ci)

### DIFF
--- a/ledger/store/src/block/mod.rs
+++ b/ledger/store/src/block/mod.rs
@@ -1015,17 +1015,10 @@ impl<N: Network, B: BlockStorage<N>> BlockStore<N, B> {
         // Compute the block tree.
         let tree = {
             // Prepare an iterator over the block heights.
-            let heights = storage.id_map().keys_confirmed();
+            let mut heights_hashes = storage.id_map().iter_confirmed().collect::<Vec<_>>();
+            heights_hashes.sort_unstable_by(|(h1, _), (h2, _)| h1.cmp(h2));
             // Prepare the leaves of the block tree.
-            let hashes = match heights.max() {
-                Some(height) => cfg_into_iter!(0..=cow_to_copied!(height))
-                    .map(|height| match storage.get_block_hash(height)? {
-                        Some(hash) => Ok(hash.to_bits_le()),
-                        None => bail!("Missing block hash for block {height}"),
-                    })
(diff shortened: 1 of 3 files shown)


vicsn (approved):
98%?! Amazing

Meshiest (approved):
LGTM. Impressive gains

gluax (approved):

zklimaleo (approved):

vicsn:
Superseded by https://github.com/AleoNet/snarkVM/pull/2561 to merge `staging` and fix `cargo fmt`

PR #2519, opened 2024-07-12
Feat/lazy load deployment

joske:
## Motivation

We've seen that RSS size is correlated with deployments loaded (and staying) in memory. This PR keeps `MAX_STACKS` programs in memory (chosen arbitrarily at 1000). 

## Test Plan

Tested by running a local devnet, deploying programs and executing them. Then restarting the network and trying to execute these programs. The `Stack`s were lazily loaded from storage, execution worked. Also tested executing programs that are not deployed.

### DIFF
--- a/synthesizer/process/Cargo.toml
+++ b/synthesizer/process/Cargo.toml
@@ -78,6 +78,9 @@ package = "snarkvm-ledger-store"
 path = "../../ledger/store"
 version = "=0.16.19"
 
+[dependencies.lru]
+version = "0.12"
+
 [dependencies.synthesizer-program]
 package = "snarkvm-synthesizer-program"
 path = "../../synthesizer/program"
@@ -93,9 +96,12 @@ package = "snarkvm-utilities"
 path = "../../utilities"
 version = "=0.16.19"
 
+[dependencies.tracing]
+version = "0.1"
+
 [dependencies.aleo-std]
 version = "0.1.24"
-default-features = false
+features = ["storage"]
 
 [dependencies.colored]
 version = "2"
--- a/synthesizer/process/src/lib.rs
+++ b/synthesizer/process/src/lib.rs
@@ -88,7 +88,7 @@ use ledger_store::helpers::memory::ConsensusMemory;
 use ledger_store::helpers::rocksdb::ConsensusDB;
 
 #[cfg(not(any(test, feature = "test")))]
-const MAX_STACKS: usize = 50;
+const MAX_STACKS: usize = 1000;
 
 #[cfg(any(test, feature = "test"))]
 const MAX_STACKS: usize = 2;
--- a/synthesizer/process/src/lib.rs
+++ b/synthesizer/process/src/lib.rs
@@ -91,8 +91,10 @@ use ledger_store::helpers::rocksdb::ConsensusDB;
 const MAX_STACKS: usize = 1000;
 
 #[cfg(any(test, feature = "test"))]
-const MAX_STACKS: usize = 2;
+// This must be at least 64 (same as MAX_IMPORTS) to avoid breaking test_max_imports.
+const MAX_STACKS: usize = 64;
 
+#[cfg(feature = "rocks")]
 #[derive(Clone)]
 pub struct Process<N: Network> {
     /// The universal SRS.
@@ -101,8 +103,21 @@ pub struct Process<N: Network> {
     credits: Option<Arc<Stack<N>>>,
     /// The mapping of program IDs to stacks.
     stacks: Arc<Mutex<LruCache<ProgramID<N>, Arc<Stack<N>>>>>,
-    /// The storage mode.
-    storage_mode: Option<StorageMode>,
+    /// The storage.
+    store: Option<ConsensusStore<N, ConsensusDB<N>>>,
+}
+
+#[cfg(not(feature = "rocks"))]
+#[derive(Clone)]
+pub struct Process<N: Network> {
+    /// The universal SRS.
+    universal_srs: Arc<UniversalSRS<N>>,
+    /// The Stack for credits.aleo
+    credits: Option<Arc<Stack<N>>>,
+    /// The mapping of program IDs to stacks.
+    stacks: Arc<Mutex<LruCache<ProgramID<N>, Arc<Stack<N>>>>>,
+    /// The storage.
+    store: Option<ConsensusStore<N, ConsensusMemory<N>>>,
 }
 
 impl<N: Network> Process<N> {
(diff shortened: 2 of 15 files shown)


raychu86:
Would love some unit tests for this as well! An easy way to do this would be to use a `#[cfg(any(test, feature = "test"))]` and set `MAX_STACKS` to be a small value for testing purposes only.

synthesizer/process/src/lib.rs
```diff
@@ -205,26 +250,74 @@ impl<N: Network> Process<N> {
...
+
+    fn load_stack(&self, program_id: impl TryInto<ProgramID<N>>) -> Result<()> {
+        let program_id = program_id.try_into().map_err(|_| anyhow!("Invalid program ID"))?;
+        info!("Lazy loading stack for {program_id}");
+        let storage_mode = self.storage_mode.as_ref().ok_or_else(|| anyhow!("Failed to get storage mode"))?.clone();
+        // try to lazy load the stack
+        #[cfg(feature = "rocks")]
+        let store = ConsensusStore::<N, ConsensusDB<N>>::open(storage_mode);
+        #[cfg(not(feature = "rocks"))]
+        let store = ConsensusStore::<N, ConsensusMemory<N>>::open(storage_mode);
```
raychu86: It may be easier to either have `C: ConsensusStorage` as a generic or to pass in `storage` somehow. 

It might break some APIs, but it feels cleaner than reloading the storage it at each instance.

joske: yes, this was the original idea. but adding the generic cascaded everywhere... Will look into that.

joske: This cascades too much. The perf hit to open storage is about 2s (on a 15GB ledger). That's not negligible, but hopefully we don't need to keep so many programs in memory.

joske: found a way to avoid opening storage every time :tada:

kpandl:
When running the snarkOS `./devnet.sh` script (with the commits of this PR attached to snarkVM) with e.g. 10 validators (and no clients), the nodes fail to load the ledger and the network doesn't start.

Typically, the logs look similar to this:
9 validators log:

> DEBUG snarkvm_synthesizer_process: Opening storage
> thread 'main' panicked at cli/src/commands/start.rs:176:84:
> Failed to parse the node: Failed to load ledger (run 'snarkos clean' and try again)
> 
> IO error: While lock file: /Users/user/dev/snarkOS/.ledger-0-0/LOCK: Resource temporarily unavailable

1 validator logs:
> INFO snarkvm_ledger: Loading the ledger from storage...
> thread 'main' panicked at cli/src/commands/start.rs:176:84:
> Failed to parse the node: Failed to load ledger (run 'snarkos clean' and try again)
> Mismatching network ID or storage mode in the database

joske:
`lock file: /Users/user/dev/snarkOS/.ledger-0-0/LOCK: Resource temporarily unavailable` This is a bit strange. Do all nodes have this?

kpandl:
Yes, all nodes had it. On aws, validator 0 worked, all other validators got stuck on this:

> 2024-09-03T14:50:51.110986Z DEBUG Opening storage
> 2024-09-03T14:50:51.172513Z DEBUG Opened storage
> 2024-09-03T14:51:08.816863Z  INFO Loading the ledger from storage...

(the `--no-dev-txs` flag was set)

joske:
did you put `--dev 0` on all by accident?

synthesizer/process/src/lib.rs
```diff
@@ -130,20 +173,50 @@ impl<N: Network> Process<N> {
...
+        };
     }
 }
 
 impl<N: Network> Process<N> {
     /// Initializes a new process.
     #[inline]
     pub fn load() -> Result<Self> {
+        // Assumption: this is only called in test code.
+        Process::load_from_storage(Some(aleo_std::StorageMode::Development(0)))
```
vicsn, line 191: Noting this may be the source of our troubles, and we may want to enclose the function in `#[cfg(any(test, feature = "test"))]` if possible.

joske: Yes this code needs to be updated.

joske: marking as draft again.

vicsn: [4aa125b](https://github.com/AleoNet/snarkVM/pull/2519/commits/4aa125bd3b9abef974d26c7f644307ffb221cd07) looks good.

~If enclosing `fn load()` in something like `#[cfg(any(test, feature = "test"))]` is too difficult (perhaps because its called both by tests and benchmarks?), another safer approach could be to get rid of `fn load()` alltogether. Should take you just one search and replace to call `Process::load_from_storage(Some(aleo_std::StorageMode::Development(0)))` directly, and that keeps the `aleo_std::StorageMode::Development(0)` logic contained to within tests and benchmarks. That makes a future developer less likely to make mistakes.~ UPDATE: See new comment below.

vicsn: I realise also that the current approach can lead to deadlocks on the storage when running tests, because all the tests will try to access the same database... Looks like we also didn't run CI on this PR yet (you can ask @ljedrz for how he has been running CI with a separate -ci branch).

New proposal: what about we let `fn load` use `store: None`. Just like `load_from_storage `, it should load the `credits.aleo` program because the tests need it. To avoid adding too many lines of code, you can still abstract similar logic to a shared function.

joske: We can not use `None` as this breaks many tests.

synthesizer/process/src/tests/test_process.rs
```diff
@@ -0,0 +1,92 @@
+// Copyright (C) 2019-2023 Aleo Systems Inc.
```
vicsn, line 1: This license needs to be updated, see e.g. https://github.com/AleoNet/snarkVM/blob/mainnet-staging/fields/src/fp12_2over3over2.rs#L1

synthesizer/process/Cargo.toml
```diff
@@ -93,9 +96,12 @@ package = "snarkvm-utilities"
 path = "../../utilities"
 version = "=0.16.19"
 
+[dependencies.tracing]
```
vicsn, line 99: Hate to nit, but these are ordered alphabetically, should be moved after `[dependencies.serde_json]`

synthesizer/process/Cargo.toml
```diff
@@ -78,6 +78,9 @@ package = "snarkvm-ledger-store"
 path = "../../ledger/store"
 version = "=0.16.19"
 
+[dependencies.lru]
```
vicsn, line 81: Same nit, all the non-Aleo dependencies are below

synthesizer/process/src/lib.rs
vicsn: Clear. Then we should rename `load` to `load_testing_only`.

vicsn:
Closed by https://github.com/AleoNet/snarkVM/pull/2553

PR #2520, opened 2024-07-19
[Fix] Add robustness to solution verification

raychu86:
## Motivation

This PR updates block production to correctly address malicious solutions. This fix overwrites the solution target with value calculated during solution verification. So even if a malicious solution is injected by a malicious validator, honest validators will always process them and safely convert them to honest solutions.

CI run is [here](https://app.circleci.com/pipelines/github/ProvableHQ/snarkVM?branch=fix%2Fsolution-verification-ci).

## Test Plan

Tests have been added to ensure that malicious solutions are handled correctly.

### DIFF
--- a/ledger/puzzle/src/lib.rs
+++ b/ledger/puzzle/src/lib.rs
@@ -266,6 +266,34 @@ impl<N: Network> Puzzle<N> {
         Ok(())
     }
 
+    /// ATTENTION: This function will update the target if the solution target is different from the calculated one.
+    /// Returns `Ok(())` if the solution is valid.
+    pub fn check_solution_mut(
+        &self,
+        solution: &mut Solution<N>,
+        expected_epoch_hash: N::BlockHash,
+        expected_proof_target: u64,
+    ) -> Result<()> {
+        // Ensure the epoch hash matches.
+        if solution.epoch_hash() != expected_epoch_hash {
+            bail!(
+                "Solution does not match the expected epoch hash (found '{}', expected '{expected_epoch_hash}')",
+                solution.epoch_hash()
+            )
+        }
+        // Calculate the proof target of the solution.
+        let proof_target = self.get_proof_target_unchecked(solution)?;
+
+        // Set the target with the newly calculated proof target value.
+        solution.target = proof_target;
+
+        // Ensure the solution is greater than or equal to the expected proof target.
+        if proof_target < expected_proof_target {
+            bail!("Solution does not meet the proof target requirement ({proof_target} < {expected_proof_target})")
+        }
+        Ok(())
+    }
+
     /// Returns `Ok(())` if the solutions are valid.
     pub fn check_solutions(
         &self,
--- a/ledger/src/advance.rs
+++ b/ledger/src/advance.rs
@@ -154,14 +154,10 @@ where
         };
 
         // Verify the solutions in the chunk.
-        let verification_results: Vec<_> = candidates_chunk
-            .iter_mut()
-            .rev()
-            .map(|solution| {
-                let verified = verification_fn(solution);
-                (solution, verified)
-            })
-            .collect();
+        let verification_results = candidates_chunk.iter_mut().rev().map(|solution| {
+            let verified = verification_fn(solution);
+            (solution, verified)
+        });
 
         // Process the results of the verification.
         for (solution, is_valid) in verification_results.into_iter() {
--- a/ledger/src/advance.rs
+++ b/ledger/src/advance.rs
@@ -127,8 +127,8 @@ pub fn split_candidate_solutions<T, F>(
     verification_fn: F,
 ) -> (Vec<T>, Vec<T>)
 where
-    T: Sized,
-    F: Fn(&T) -> bool,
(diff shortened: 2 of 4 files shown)


ledger/src/tests.rs
```diff
@@ -2668,4 +2669,141 @@ mod valid_solutions {
...
+            assert_eq!(block.solutions().len(), 1);
+            assert_eq!(block.aborted_solution_ids().len(), 0);
+
+            // Fetch the solution from the block.
+            let (solution_id, solution) = block.solutions().as_ref().unwrap().first().unwrap();
+            assert_eq!(*solution_id, expected_solution_id, "Check that the block has the correct solution ID");
+            assert_eq!(*solution, expected_solution, "Check that the block has the correct solution");
+        };
+
+        // Case 1: The valid solution is included in the block.
```
vicsn, line 2743: Looking deeper in the code, wouldn't `self.proof_target_cache` save the first target it sees? What happens if Case 1 is the malicious solution?

vicsn: For completeness, could you add two more cases with a second solution, just switching the the order (first inserting the malicious solution, then the valid one).

ledger/src/tests.rs
```diff
@@ -2668,4 +2669,141 @@ mod valid_solutions {
...
+            if puzzle.get_proof_target(&solution).unwrap() >= minimum_proof_target {
+                valid_solution = Some(solution);
+            }
+        }
+        // Unwrap the valid solution.
+        let valid_solution = valid_solution.unwrap();
+
+        // Construct a malicious solution with a different target.
+        let malicious_solution =
+            Solution::new(*valid_solution.partial_solution(), valid_solution.target().saturating_add(1));
```
vicsn: Could we use `minimum_proof_target - 1` instead? This more closely reflects what an attacker would do

ledger/src/advance.rs
```diff
@@ -147,28 +147,28 @@ where
...
         let verification_results: Vec<_> = candidates_chunk
-            .into_iter()
+            .iter_mut()
             .rev()
             .map(|solution| {
-                let verified = verification_fn(&solution);
+                let verified = verification_fn(solution);
                 (solution, verified)
             })
             .collect();
```
ljedrz: this isn't introduced in this PR, but it doesn't seem that we need this intermediate `collect` into a vector?

ledger/src/tests.rs
raychu86: You shouldn't be able to have both in the block, since snarkOS filters out duplicates based on the `solutionID`. 

We are just trying to prevent a deviation case where different nodes processes the same solution ID with different solutions.

ledger/src/advance.rs
raychu86: Added optimization with [fa3a03d](https://github.com/AleoNet/snarkVM/pull/2520/commits/fa3a03dd151929a8af4eb879f9fc51499d13e33b). 

However, I believe this entire function can be overhauled/optimized in subsequent PRs. We currently aren't parallelizing verification.

ledger/src/tests.rs
raychu86: Resolved with - [3adb1e2](https://github.com/AleoNet/snarkVM/pull/2520/commits/3adb1e26db563692d811dcdf40994f2e55975a76)

ledger/src/tests.rs
vicsn: The `ledger` object is re-used across tests, potentially doing caching of the solution target. So even though there's no vulnerability, I think a slightly more proper test would be to let Case 1 be malicious, and Case 2 succesful.

raychu86: Resolved with [8f2c014](https://github.com/AleoNet/snarkVM/pull/2520/commits/8f2c0144c77a164e066f253153001dd072ec6103)

ledger/puzzle/src/lib.rs
```diff
@@ -266,6 +266,34 @@ impl<N: Network> Puzzle<N> {
...
+                solution.epoch_hash()
+            )
+        }
+        // Calculate the proof target of the solution.
+        let proof_target = self.get_proof_target_unchecked(solution)?;
+
+        // Set the target with the newly calculated proof target value.
+        solution.target = proof_target;
+
+        // Ensure the solution is greater than or equal to the expected proof target.
```
ljedrz, line 290: doesn't this suggest that the comparison below should be reversed :thinking:?

vicsn: I think both the comment and the code are consistent and correct. But using `ensure` or rewriting the comment would make that clearer

vicsn (approved):

ledger/puzzle/src/lib.rs
ljedrz: ah, you're right, I think my morning brain just expected an `ensure!` there instead of `bail!`

ljedrz (approved):
LGTM

raychu86:
Closing this PR for now due to unifying the fix in https://github.com/AleoNet/snarkVM/pull/2521.

PR #2521, opened 2024-07-20
[Fix] Add checksum to `TransmissionID`

raychu86:
## Motivation

This PR introduces the concept of a `TransmissionChecksum`. This checksum is a `Sha3_256` hash on the underlying `Transmission` object that is truncated to fit a `u128`. The intention is to prevent malicious validators from tampering with solutions or transactions and broadcasting these malformed objects to other validators. The addition of these checksums, alongside additional snarkOS level checks will allow honest validators to detect and disregard the attack from the malicious validators.

This PR also unifies the changes from https://github.com/AleoNet/snarkVM/pull/2520, as it aims to solve the same attack. The change there updates block production to correctly address malicious solutions. This additional fix overwrites the solution target with value calculated during solution verification. So even if a malicious solution is injected by a malicious validator, honest validators will always process them and safely convert them to honest solutions.

The CI is run [here](https://app.circleci.com/pipelines/github/ProvableHQ/snarkVM?branch=fix%2Ftransmission-checksum-ci).

**NOTE**: This updates the block serialization, so it is not backwards compatible with the networks (Testnet Beta and Canary) being run. These will require a reset.

#### Benchmarks
The benchmarks indicate that the added runtime overhead is quite miniscule. `Data::to_checksum` benchmarks (on a 64 GB M1 MAX):
- Solution: 10μs
- `transfer_public` transaction: 82μs
- Largest possible transaction (128 kb): 5.5ms 

## Related PRs

The sister PR is https://github.com/AleoNet/snarkOS/pull/3367.

### DIFF
--- a/ledger/block/src/verify.rs
+++ b/ledger/block/src/verify.rs
@@ -545,8 +545,9 @@ impl<N: Network> Block<N> {
             .collect::<Result<Vec<_>>>()?;
         let mut unconfirmed_transactions = unconfirmed_transactions.iter().peekable();
 
-        // Initialize a set of already seen transmission IDs.
-        let mut seen_transmission_ids = HashSet::new();
+        // Initialize a set of already seen transaction and solution IDs.
+        let mut seen_transaction_ids = HashSet::new();
+        let mut seen_solution_ids = HashSet::new();
 
         // Initialize a set of aborted or already-existing solution IDs.
         let mut aborted_or_existing_solution_ids = HashSet::new();
@@ -555,9 +556,22 @@ impl<N: Network> Block<N> {
 
         // Iterate over the transmission IDs.
         for transmission_id in subdag.transmission_ids() {
-            // If the transmission ID has already been seen, then continue.
-            if !seen_transmission_ids.insert(transmission_id) {
-                continue;
+            // If the transaction or solution ID has already been seen, then continue.
+            // Note: This is done instead of checking `TransmissionID` directly, because we need to
+            // ensure that each transaction or solution ID is unique. The `TransmissionID` is guaranteed
+            // to be unique, however the transaction/solution ID may not be due to malleability concerns.
+            match transmission_id {
+                TransmissionID::Ratification => {}
+                TransmissionID::Solution(solution_id, _) => {
+                    if !seen_solution_ids.insert(solution_id) {
+                        continue;
+                    }
+                }
+                TransmissionID::Transaction(transaction_id, _) => {
+                    if !seen_transaction_ids.insert(transaction_id) {
+                        continue;
+                    }
+                }
             }
 
             // Process the transmission ID.
--- a/ledger/block/src/verify.rs
+++ b/ledger/block/src/verify.rs
@@ -539,11 +539,11 @@ impl<N: Network> Block<N> {
     ) -> Result<(Vec<SolutionID<N>>, Vec<N::TransactionID>)> {
         // Prepare an iterator over the solution IDs.
         let mut solutions = solutions.as_ref().map(|s| s.deref()).into_iter().flatten().peekable();
-        // Prepare an iterator over the unconfirmed transaction IDs.
-        let unconfirmed_transaction_ids = cfg_iter!(transactions)
-            .map(|confirmed| confirmed.to_unconfirmed_transaction_id())
+        // Prepare an iterator over the unconfirmed transactions.
+        let unconfirmed_transactions = cfg_iter!(transactions)
+            .map(|confirmed| confirmed.to_unconfirmed_transaction())
             .collect::<Result<Vec<_>>>()?;
-        let mut unconfirmed_transaction_ids = unconfirmed_transaction_ids.iter().peekable();
+        let mut unconfirmed_transactions = unconfirmed_transactions.iter().peekable();
 
         // Initialize a set of already seen transmission IDs.
         let mut seen_transmission_ids = HashSet::new();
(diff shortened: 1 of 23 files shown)


ledger/block/src/verify.rs
```diff
@@ -563,10 +563,15 @@ impl<N: Network> Block<N> {
...
             match transmission_id {
                 TransmissionID::Ratification => {}
-                TransmissionID::Solution(solution_id) => {
+                TransmissionID::Solution(solution_id, checksum) => {
                     match solutions.peek() {
                         // Check the next solution matches the expected solution ID.
-                        Some((_, solution)) if solution.id() == *solution_id => {
+                        Some((_, solution))
+                            if solution.id() == *solution_id
+                                && Data::<Solution<N>>::Buffer(solution.to_bytes_le()?.into())
```
ljedrz, line 585: we might want to introduce (probably in `snarkvm-utilities`) a free function with a bound on `Network` that converts a `&[u8]` to a checksum in order to avoid having to use a `Data` wrapper here and in a few other spots; that function would then be reused internally in `Data::to_checksum`.

ljedrz (dismissed):
LGTM, left one minor suggestion.

vicsn:
~Even though adding the checksum to the `proof_target_cache` is technically not needed if https://github.com/AleoNet/snarkVM/pull/2520 is implemented, want to add it anyway as defense in depth?~ On second thought, due to the large overhead of verification, I'd much prefer if we just merge https://github.com/AleoNet/snarkVM/pull/2520 and leave the `proof_target_cache` just dependent on the solution id

vicsn (dismissed):

ledger/narwhal/transmission-id/src/lib.rs
```diff
@@ -27,41 +27,50 @@ pub enum TransmissionID<N: Network> {
...
     /// Converts the solution ID into a transmission ID.
-    fn from(solution_id: SolutionID<N>) -> Self {
-        Self::Solution(solution_id)
+    fn from((solution_id, checksum): (SolutionID<N>, N::TransmissionChecksum)) -> Self {
+        Self::Solution(solution_id, checksum)
     }
 }
 
-impl<N: Network> From<&N::TransactionID> for TransmissionID<N> {
+impl<N: Network> From<(&N::TransactionID, &N::TransmissionChecksum)> for TransmissionID<N> {
```
vicsn, line 42: Nit: this impl uses references, while the From impl for solution_id uses values

raychu86: I was unable to do this because the rust type system does not allow multiple impl of the same trait with the same type ownership:

```
error[E0119]: conflicting implementations of trait `From<(snarkvm_ledger_puzzle::SolutionID<_>, &_)>` for type `TransmissionID<_>`
  --> ledger/narwhal/transmission-id/src/lib.rs:42:1
   |
35 | impl<N: Network> From<(SolutionID<N>, N::TransmissionChecksum)> for TransmissionID<N> {
   | ------------------------------------------------------------------------------------- first implementation here
...
42 | impl<N: Network> From<(N::TransactionID, N::TransmissionChecksum)> for TransmissionID<N> {
   | ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^ conflicting implementation for `TransmissionID<_>
   ```

File: ledger/narwhal/transmission-id/src/string.rs [hunk 3]
```diff
@@ -43,8 +52,8 @@ impl<N: Network> Display for TransmissionID<N> {
     fn fmt(&self, f: &mut Formatter<'_>) -> fmt::Result {
         match self {
             Self::Ratification => write!(f, "ratification"),
-            Self::Solution(id) => write!(f, "{}", id),
-            Self::Transaction(id) => write!(f, "{}", id),
+            Self::Solution(id, checksum) => write!(f, "{}.{}", id, checksum),
```
Meshiest, line 55: it may be worth considering encoding the checksum as padded base16 or bech32 without a human-readable-part rather than base10

Meshiest (dismissed):

ledger/block/src/verify.rs
gluax: We could also consider implementing `Eq` between the solution and the `Data` wrapper type directly so her it could be simplified to `solution == *checksum`.

ledger/src/advance.rs
```diff
@@ -147,28 +147,24 @@ where
...
         for (solution, is_valid) in verification_results.into_iter() {
             if is_valid && valid_candidate_solutions.len() < max_solutions {
-                valid_candidate_solutions.push(solution);
+                valid_candidate_solutions.push(*solution);
             } else {
-                aborted_candidate_solutions.push(solution);
+                aborted_candidate_solutions.push(*solution);
             }
         }
     }
```
gluax, line 170: Could this not be done in the same iterator instead of iterating twice? Maybe the compiler is smart enough to simplify this but seems wasteful to me.

gluax (dismissed):

ledger/src/advance.rs
raychu86: It's possible, but we would likely have to refactor `verification_fn`. It also isn't in the scope of this PR, so ideally any optimizations should be done in a future PR.

ljedrz (dismissed):

raychu86:
Note: [a6d175a](https://github.com/AleoNet/snarkVM/pull/2521/commits/a6d175aa0b4a4f738cae4e44aa9405a028041216) was added to further enforce the uniqueness of solution/transaction IDs that make it into blocks. This is also done on the snarkOS level with - [64e58a0](https://github.com/AleoNet/snarkOS/pull/3367/commits/64e58a053b92a1a6ce91452cb88316cc31a8c781)

ljedrz (approved):

snarkworld (approved):
LGTM

zosorock:
The failed `ledger` CI passed on Provable's side so we will not re-run:
https://app.circleci.com/pipelines/github/ProvableHQ/snarkVM/194/workflows/70ed53e9-b4d8-4d22-8b32-a88bbc223c79/jobs/14663

PR #2522, opened 2024-07-30
Updating Canary genesis block with TransmissionID checksum - v0.4.0

zosorock:
## Motivation
Updating to Canary genesis block v0.4.0

`sha256 == e6419d92fa653a9560310a168ac2f3c65cb3edc9ffb966e1fdb1d208bc0929d2`

## Related PRs
https://github.com/AleoNet/snarkVM/pull/2521
https://github.com/AleoNet/snarkOS/pull/3367


raychu86 (dismissed):
LGTM, though I'm not sure I like having the JSON objects because all of the information already exists in `block.genesis`. 

It is another surface that will need to be updated if `block.genesis` needs to be resampled, and this feels easy to forget down the line.

In the context of this PR, it's obvious that these files were used to create the genesis block, however it isn't clear from an outside perspective. At the least, I think a name change is warranted; for example, `validators.json` -> `genesis_validators.json`

snarkworld (dismissed):
LGTM, and agree with @raychu86 point.

Meshiest (dismissed):

snarkworld (approved):
LGTM given CI has passed.

PR #2523, opened 2024-07-31
[Fix] Cache parameter downloads and limit test parallelization

### LINKED ISSUE #2494: Flaky Tests on CI
## 🐛 Bug Report ##
Issue: Flaky Tests on CI
Severity: Medium

### Description

We have observed that some tests in our CI pipeline exhibit flaky behavior, requiring multiple runs to pass. This inconsistency is affecting the reliability and efficiency of our development process.

### Affected Tests

`algorithms` - This test often fails unpredictably, and the root cause is currently unknown.

While not explicitly mentioned, other tests may also exhibit similar behavior, requiring multiple attempts to pass.

### Steps to Reproduce

- Run the CI pipeline.
- Observe the failure of the algorithms test (and potentially others) intermittently.
- Re-run the failed tests.
- Notice that the tests may pass on subsequent attempts.

### Expected Behavior

All tests should pass consistently on the first run, provided that the code is correct.

### Actual Behavior

The algorithms test (and potentially others) fail intermittently without any changes to the code.
These tests often require multiple attempts to pass, leading to wasted time and resources.

### Impact

Decreases confidence in the CI results.
Slows down the development process due to the need for re-running tests.
Makes it difficult to identify genuine issues in the codebase.

### Possible Causes

Race conditions or timing issues within the tests or the code being tested.
Environmental issues related to the CI infrastructure.
Dependencies on external services or resources that may not be consistently available.

### Suggested Actions

Investigation and Diagnosis
    - Conduct a thorough investigation to identify the root cause of the flakiness in the algorithms test.
    - Review the test code and the associated application code for potential issues.

Test Stabilization
    - Implement fixes to address any identified issues causing the flakiness.
    - Ensure that tests do not have hidden dependencies on external resources or timing conditions.

Enhancement of CI Infrastructure
    - Ensure that the CI environment is consistent and reliable.
    - Consider introducing additional logging or diagnostics to capture more information about the failures.

Documentation and Communication
    - Document the findings and the steps taken to address the flaky tests.
    - Communicate any changes to the team to ensure that everyone is aware of the improvements and any new best practices.

### Additional Information

Please provide any logs or additional context that might help in diagnosing the issue.
If you have observed flaky behavior in other tests, please list them here as well.

vicsn:
## Motivation

Closes https://github.com/AleoNet/snarkVM/issues/2494 by:
- caching parameter downloads so we don't get rate limited by AWS.
- limiting parallelization of tests using `-- --test-threads`. The other option would be to use `cargo nextest`, but that would add a new dependency.

Also fixes caching, which previously wasn't actually used. By recommendation from [CircleCI](https://circleci.com/docs/configuration-reference/): "Given the immutability of caches, it might be helpful to start all your cache keys with a version prefix v1-.... That way you will be able to regenerate all your caches just by incrementing the version in this prefix."

Total runtime is now 30 minutes.

Future optimization ideas:
- Do not try to download the largest parameters in a test, which is several GiB and costs 16 min.
- Get more fine-grained data on how much memory each test uses. `nextext` allows configuring of granular test profiles depending on their load. My current tests showed most tests just use 60 MiB, but that must have been wrong since we see OOM with as little as `--test-threads=4` sometimes.

## Test Plan

- [Succesful CI run](https://app.circleci.com/pipelines/github/ProvableHQ/snarkVM/240/workflows/48ffb50d-097d-4171-a209-0ba3a9d94605)
- Additionally ran CI multiple times in parallel to see if the issues are resolved.

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -66,16 +66,16 @@ commands:
       - run:
           name: Prepare environment and install dependencies
           command: |
-            export SCCACHE_CACHE_SIZE=200M
+            echo 'export "RUSTC_WRAPPER"="sccache"' >> $BASH_ENV
+            echo 'export "SCCACHE_CACHE_SIZE"="200M"' >> $BASH_ENV
             export WORK_DIR="$CIRCLE_WORKING_DIRECTORY/.cache/sccache"
             export SCCACHE_DIR="$CIRCLE_WORKING_DIRECTORY/.cache/sccache"
             mkdir -p "$CIRCLE_WORKING_DIRECTORY/.bin"
             wget https://github.com/mozilla/sccache/releases/download/v0.3.0/sccache-v0.3.0-x86_64-unknown-linux-musl.tar.gz
             tar -C "$CIRCLE_WORKING_DIRECTORY/.bin" -xvf sccache-v0.3.0-x86_64-unknown-linux-musl.tar.gz
-            mv $CIRCLE_WORKING_DIRECTORY/.bin/sccache-v0.3.0-x86_64-unknown-linux-musl/sccache $CIRCLE_WORKING_DIRECTORY/.bin/sccache
-            export PATH="$PATH:$CIRCLE_WORKING_DIRECTORY/.bin"
-            export RUSTC_WRAPPER="sccache"
-            rm -rf "$CIRCLE_WORKING_DIRECTORY/.cargo/registry"
+            chmod +x $CIRCLE_WORKING_DIRECTORY/.bin/sccache-v0.3.0-x86_64-unknown-linux-musl/sccache
+            mv $CIRCLE_WORKING_DIRECTORY/.bin/sccache-v0.3.0-x86_64-unknown-linux-musl/sccache /home/circleci/bin/sccache
+            rm -rf "/home/circleci/.cargo/registry"
             DEBIAN_FRONTEND=noninteractive sudo apt-get update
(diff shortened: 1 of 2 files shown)


raychu86:
Overall, looks great to me. I only have 1 question:

Is it possible to still allow downloads of the smaller files to ensure that the parameter fetching pipeline is still tested and working in CI? (only cache the large objects)

vicsn:
> Is it possible to still allow downloads of the smaller files to ensure that the parameter fetching pipeline is still tested and working in CI? (only cache the large objects)

Probably many ways to do this. We'd have to disable circleci cache for the `parameters` crate, AND either one of:
1. `#[ignore]` higher degree tests (just like we do for some other tests). A special CI runner could run all the `#[ignore]`'d tests separately. Unfortunately this requires Rust macro work in `impl_remote` so I fear its too much of a timesink right now, at least for me.
2. We're not even using the highest degree parameters, and we won't be for a long time, and if we do need it we'll notice prover side, so we can even just comment out high degree `impl_remote`. The main downside it may confuse developers in 2-3 years time.

raychu86 (dismissed):
LGTM!

Would be nice to have some tests to ensure that `impl_remote` is working properly. Maybe some new dedicated tests to download a select few parameters (even if it exists in the CI cache).

vicsn:
> Would be nice to have some tests to ensure that impl_remote is working properly. Maybe some new dedicated tests to download a select few parameters (even if it exists in the CI cache).

Added, new test succeeds.

snarkworld (approved):
LGTM - successful CI in PR

raychu86 (approved):
LGTM

PR #2524, opened 2024-08-01
Don't assume solution checksum consistency in check_subdag_transmissions

vicsn:
## Motivation

This is a small improvement to https://github.com/AleoNet/snarkVM/pull/2521 . Due to the fixes introduced, honest validators may mutate a seemingly invalid solution to a valid one. As a result, there's a deviation between the checksum in the subdag and the block. If such an attack were to happen, it can be easily recovered from by implementing this PR.

The fix is to ignore the checksum. We're already guaranteed only unique solution id's enter the block.

## Test Plan

Tested by simulating malicious validator.
[CI run here](https://app.circleci.com/pipelines/github/ProvableHQ/snarkVM/208/workflows/7b445769-6e16-4261-b737-a004e292d50a)

### DIFF
--- a/ledger/block/src/verify.rs
+++ b/ledger/block/src/verify.rs
@@ -577,15 +577,11 @@ impl<N: Network> Block<N> {
             // Process the transmission ID.
             match transmission_id {
                 TransmissionID::Ratification => {}
-                TransmissionID::Solution(solution_id, checksum) => {
+                TransmissionID::Solution(solution_id, _checksum) => {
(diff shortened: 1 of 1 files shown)


raychu86 (approved):
LGTM.

This one-line change we determined to be needed to fully ensure that we guarantee solution manipulation by malicious validators does not prevent block production in any way. I've also tested locally to ensure it maintains the existing desired properties, while also fixing 1 more edge-case.

PR #2525, opened 2024-08-02
[Fix] Set console caller and signer when executing closures.

d0cd:
This PR ensures that the console `caller` and `signer` registers are appropriately set when executing a `closure`.

The CI run is [here](https://github.com/ProvableHQ/snarkVM/tree/fix/caller-in-closure-ci).


synthesizer/process/src/stack/execute.rs
```diff
@@ -45,8 +45,15 @@ impl<N: Network> StackExecute<N> for Stack<N> {
 
         // Initialize the registers.
         let mut registers = Registers::new(call_stack, self.get_register_types(closure.name())?.clone());
+
+        use circuit::Eject;
+
+        // Set the transaction signer.
+        registers.set_signer(signer.eject_value());
```
vicsn, line 52: Can you wrap the two new sets with `if let CallStack::Execute(..) = registers.call_stack()` just like we do below? Then we can keep the `use Eject` to just within the Execute blocks. I hope the tests still pass then O:)

d0cd: Do we actually want to do that? In the corresponding logic in `execute_function` we set in all modes, not just `CallStack::Execute`.

vicsn: Hmm good call out. I guess we can leave the PR as is then.

I appended a diff to test if wrapping all of the console sets in `Execute` mode would work, and indeed the tests still pass (modulo requiring new expectations because I changed a test). Maybe this can inform a hypothetical future refactor of `execute_{function, closure}`.

[diff.txt](https://github.com/user-attachments/files/16515242/diff.txt)

vicsn:
FYI it's on my list to re-review and re-open this one.

PR #2526, opened 2024-08-05
avoid comments about (3f + 1) and (2f + 1)

bendyarm:
## Motivation

In `snarkvm_ledger_committee`, `quorum_threshold` is correctly calculated as `N - f`.
Since `N` is not always `3f + 1` (it could be `3f + 2` or `3f + 3`),
the quorum threshold is not always `2f + 1`.

This PR changes some comments to avoid the imprecise `3f + 1` and `2f + 1`.

NOTE: no need for CI to pass since only comments were changed.

### DIFF
--- a/ledger/committee/src/lib.rs
+++ b/ledger/committee/src/lib.rs
@@ -173,6 +173,7 @@ impl<N: Network> Committee<N> {
     pub fn quorum_threshold(&self) -> u64 {
         // Assuming `N = 3f + 1 + k`, where `0 <= k < 3`,
         // then `2N/3 + 1 = 2f + 1 + (2k + 2)/3 = 2f + 1 + k = N - f`.
+        // In the line above, `/` means integer division.
         self.total_stake().saturating_mul(2).saturating_div(3).saturating_add(1)
     }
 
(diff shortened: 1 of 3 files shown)


ledger/committee/src/lib.rs
```diff
@@ -169,14 +169,14 @@ impl<N: Network> Committee<N> {
         self.total_stake().saturating_add(2).saturating_div(3)
     }
 
-    /// Returns the amount of stake required to reach a quorum threshold `(2f + 1)`.
+    /// Returns the amount of stake required to reach a quorum threshold `(N - f)`.
     pub fn quorum_threshold(&self) -> u64 {
         // Assuming `N = 3f + 1 + k`, where `0 <= k < 3`,
-        // then `(2N + 3) / 3 = 2f + 1 + (2k + 2)/3 = 2f + 1 + k = N - f`.
+        // then `2N/3 + 1 = 2f + 1 + (2k + 2)/3 = 2f + 1 + k = N - f`.
```
vicsn, line 175: This was already there, but can you explain the jump `2f + 1 + (2k + 2)/3 = 2f + 1 + k`? I guess it should be `<=`

acoglio: The `/` is integer division, corresponding to `saturating_div` in the code. Maybe we should add a note clarifying that. (Since k may be 0 or 1 or 2, then (2k+2) div 3 is respectively 0 or 1 or 2, i.e. the same as k.)

vicsn (dismissed):

acoglio (approved):

raychu86 (approved):
LGTM!

PR #2532, opened 2024-08-22
[refactor] Improve the Display impl for ProgramCore

ljedrz:
This is technically a refactoring, but with some performance implications; this impl also shouldn't cause a panic anymore in case of a formatting error.

Found via fuzzing, with `Stack::new` causing multiple crashes during the parse check.

### DIFF
--- a/synthesizer/program/src/parse.rs
+++ b/synthesizer/program/src/parse.rs
@@ -143,56 +143,55 @@ impl<N: Network, Instruction: InstructionTrait<N>, Command: CommandTrait<N>> Deb
     }
 }
 
-#[allow(clippy::format_push_string)]
 impl<N: Network, Instruction: InstructionTrait<N>, Command: CommandTrait<N>> Display
     for ProgramCore<N, Instruction, Command>
 {
     /// Prints the program as a string.
     fn fmt(&self, f: &mut Formatter<'_>) -> fmt::Result {
-        // Initialize a string for the program.
-        let mut program = String::new();
-
         if !self.imports.is_empty() {
             // Print the imports.
             for import in self.imports.values() {
-                program.push_str(&format!("{import}\n"));
+                writeln!(f, "{import}")?;
             }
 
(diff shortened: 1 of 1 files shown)


synthesizer/program/src/parse.rs
```diff
@@ -130,56 +130,55 @@ impl<N: Network, Instruction: InstructionTrait<N>, Command: CommandTrait<N>> Deb
...
                 },
                 ProgramDefinition::Function => match self.functions.get(identifier) {
-                    Some(function) => program.push_str(&format!("{function}\n\n")),
+                    Some(function) => writeln!(f, "{function}")?,
                     None => return Err(fmt::Error),
                 },
             }
+            // Omit the last newline.
+            if identifier_iter.peek().is_some() {
+                writeln!(f)?;
```
vicsn, line 190: Curious black magic, just to confirm, this doesn't *add* a newline but removes it (as the comment makes me think)?

vicsn (approved):
Seems harmless if tests pass

synthesizer/program/src/parse.rs
ljedrz: This adds a newline only if this wasn't the very last element (which is equivalent to adding it to all and removing the final one, which is what the former version does).

zosorock:
The same CI jobs have failed 3 times. @vicsn can you let us know if you feel these are relevant or not, please?

vicsn:
Merging upstream should do the trick for `fmt`, @ljedrz let me know if you can tackle it otherwise I'll clone this PR to fix it!

The parameters test is unrelated, not sure why it fails, will look into it if it becomes a recurring problem.

ljedrz:
@vicsn sure, I'll rebase shortly :ok_hand:

gluax (approved):

zkxuerb (approved):
LGTM

gluax (approved):

PR #2533, opened 2024-08-22
[refactor] Use explicit wrapping addition in Fp*::mul_assign

ljedrz:
While overflowing addition will not panic in `--release` mode ([it will instead wrap around](https://doc.rust-lang.org/stable/book/ch03-02-data-types.html#integer-overflow)), it will cause a crash under `--debug`. This PR switches to explicit wrapping addition for the `mul_assign` method for `Fp256` and `Fp384`.

Found via fuzzing.

### DIFF
--- a/fields/src/fp_256.rs
+++ b/fields/src/fp_256.rs
@@ -767,7 +767,7 @@ impl<'a, P: Fp256Parameters> MulAssign<&'a Self> for Fp256<P> {
 
         r[3] = fa::mac_with_carry(r[3], (self.0).0[3], (other.0).0[0], &mut carry1);
         r[2] = fa::mac_with_carry(r[3], k, P::MODULUS.0[3], &mut carry2);
-        r[3] = carry1 + carry2;
+        r[3] = carry1.wrapping_add(carry2);
 
         // Iteration 1.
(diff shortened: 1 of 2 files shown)


vicsn:
Can you share the error you saw, was it "Failed to eject scalar value: The scalar is greater than or equal to the modulus." by any chance?

ljedrz:
@vicsn sadly, this has only occurred in one of my initial runs, where I built with `--debug` by accident (since it is not the intended way of building snarkVM), so I no longer have those artifacts. That being said, I consider this a cleanup (as opposed to a fix), as this is what we are already de facto doing except for some of the tests (which is what the linked PR noted).

vicsn:
Currently I'd recommend to close this PR. Perhaps for now you can draft it.

Reason is that IF the wrapping behaviour is not something we actually want, it would be nice if it will at least show up in a panicking test.

I didn't evaluate whether the wrapping is acceptable or reachable under normal conditions. One day that part of the code can be re-tested and re-reviewed.

ljedrz:
Closing, as overflows are unreachable with the current crypto primitives (only via brute-force fuzzing), and we never run in debug mode regardless. Indeed, we'd want this to surface in some test case in case we ever add new primitives.

PR #2534, opened 2024-08-23
[Fix] Avoid ejecting the scalar value when loading and storing a circuit.

d0cd:
This PR avoids ejecting the scalar value when loading and storing a circuit. 

## Failing Case

This issue was found in a deployed program.

Suppose that we have a program `cast` operation from a `field` to a `scalar`.
Suppose that `r0` is a field element that is larger than the scalar modulus.
```
function foo:
    input r0 as field.private;
    cast r0 into r1 as scalar;
```
When the `cast` instruction is executed, constraints are added to check that the casted scalar value is correct.
During deployment, these constraints are not checked for satisfaction (by design), and an invalid scalar element is stored into the registers.
[`store_circuit`](https://github.com/AleoNet/snarkVM/blob/be171ce0720544c2bc51e27e1674a62fff585adc/synthesizer/process/src/stack/registers/load.rs#L162) ejects the stored value and checks for type equality.
However, the implementation of `Eject` for `Scalar` requires that the element is a valid a scalar and panics otherwise.

## Solution

The [eject](https://github.com/AleoNet/snarkVM/blob/be171ce0720544c2bc51e27e1674a62fff585adc/synthesizer/process/src/stack/registers/load.rs#L162) is done to check that the type of the stored value is correct. 
To avoid this failing case, we directly check for type equality against the circuit value.

## Testing

This PR includes an integration test that enumerates all `LiteralType` combinations of `cast` and `cast.lossy` and tests them in a program context.

CI is running in [this](https://github.com/ProvableHQ/snarkVM/tree/fix/ejection-in-load-store-ci) branch.

## Considerations

We considered alternate solutions including:
- restricting the `cast` operation on `scalar`s
- ensuring that `Scalar::eject` does not fail
however, these solutions would either require a migration or impact a large portion of the codebase.

### DIFF
--- a/synthesizer/process/src/stack/registers/load.rs
+++ b/synthesizer/process/src/stack/registers/load.rs
@@ -159,7 +159,23 @@ impl<N: Network, A: circuit::Aleo<Network = N>> RegistersLoadCircuit<N, A> for R
         match self.register_types.get_type(stack, register) {
             // Ensure the stack value matches the register type.
             Ok(register_type) => {
-                stack.matches_register_type(&circuit::Eject::eject_value(&circuit_value), &register_type)?
+                // Check if the register type and circuit value are both scalar.
+                let register_type_is_scalar =
+                    matches!(register_type, RegisterType::Plaintext(PlaintextType::Literal(LiteralType::Scalar)));
+                let circuit_value_is_scalar = matches!(
+                    circuit_value,
+                    circuit::Value::Plaintext(circuit::Plaintext::Literal(circuit::Literal::Scalar(_), _))
+                );
+
+                // Check if the register type matches the type in the circuit value.
+                // We do a special check for scalar values, as there is a possibility of an overflow via
+                // field to scalar conversion in deployment verification.
+                match register_type_is_scalar && circuit_value_is_scalar {
+                    true => {}
+                    false => {
+                        stack.matches_register_type(&circuit::Eject::eject_value(&circuit_value), &register_type)?
+                    }
+                }
             }
             // Ensure the register is defined.
             Err(error) => bail!("Register '{register}' is not a member of the function: {error}"),
--- a/synthesizer/process/src/stack/registers/store.rs
+++ b/synthesizer/process/src/stack/registers/store.rs
@@ -90,7 +90,24 @@ impl<N: Network, A: circuit::Aleo<Network = N>> RegistersStoreCircuit<N, A> for 
                 match self.register_types.get_type(stack, register) {
                     // Ensure the stack value matches the register type.
                     Ok(register_type) => {
-                        stack.matches_register_type(&circuit::Eject::eject_value(&circuit_value), &register_type)?
+                        // Check if the register type and circuit value are both scalar.
+                        let register_type_is_scalar = matches!(
+                            register_type,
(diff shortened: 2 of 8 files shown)


synthesizer/program/src/logic/instruction/operation/cast.rs
```diff
@@ -169,6 +169,25 @@ pub struct CastOperation<N: Network, const VARIANT: u8> {
 }
 
 impl<N: Network, const VARIANT: u8> CastOperation<N, VARIANT> {
+    /// Initializes a new `cast` instruction.
+    #[inline]
+    pub fn new(operands: Vec<Operand<N>>, destination: Register<N>, cast_type: CastType<N>) -> Result<Self> {
```
howardwu, line 174: Is this new method only use for testing purposes?

If so, I would advise as an independent reviewer that this method be denoted as `CastOperation::new_testing_only` with a `#[cfg(test)]` guard on it.

synthesizer/process/src/stack/registers/load.rs
```diff
@@ -159,7 +159,23 @@ impl<N: Network, A: circuit::Aleo<Network = N>> RegistersLoadCircuit<N, A> for R
...
+                let circuit_value_is_scalar = matches!(
+                    circuit_value,
+                    circuit::Value::Plaintext(circuit::Plaintext::Literal(circuit::Literal::Scalar(_), _))
+                );
+
+                // Check if the register type matches the type in the circuit value.
+                // We do a special check for scalar values, as there is a possibility of an overflow via
+                // field to scalar conversion in deployment verification.
+                match register_type_is_scalar && circuit_value_is_scalar {
+                    true => {}
```
howardwu, line 174: The comment above states that `We do a special check for scalar values` however there is no check here on `L174` for the scalar case. Can you clarify what exactly should be checked here?

synthesizer/process/src/stack/registers/store.rs
```diff
@@ -90,7 +90,24 @@ impl<N: Network, A: circuit::Aleo<Network = N>> RegistersStoreCircuit<N, A> for
...
+                        let circuit_value_is_scalar = matches!(
+                            circuit_value,
+                            circuit::Value::Plaintext(circuit::Plaintext::Literal(circuit::Literal::Scalar(_), _))
+                        );
+
+                        // Check if the register type matches the type in the circuit value.
+                        // We do a special check for scalar values, as there is a possibility of an overflow via
+                        // field to scalar conversion in deployment verification.
+                        match register_type_is_scalar && circuit_value_is_scalar {
+                            true => {}
```
howardwu, line 107: The comment above states that `We do a special check for scalar values` however there is no check here on `L107` for the scalar case. Can you clarify what exactly should be checked here?

synthesizer/process/src/stack/registers/load.rs
```diff
@@ -159,7 +159,23 @@ impl<N: Network, A: circuit::Aleo<Network = N>> RegistersLoadCircuit<N, A> for R
...
+                    false => {
+                        stack.matches_register_type(&circuit::Eject::eject_value(&circuit_value), &register_type)?
```
raychu86, line 176: What about other types that might have scalars internally? 

e.g. Structs that have `scalar` elements or arrays of `scalar`s?

d0cd:
Closed in favor of #2535

PR #2535, opened 2024-08-23
[Fix] Make deployment verification consistent by deterministically seeding the RNG.

d0cd:
This PR:
- fixes deployment verification by deterministically seeding the RNG.

## Failing Case

This issue was found in a deployed program with a similar structure to:
```
program test.aleo;
function foo:
    input r0 as field.private;
    cast r0 into r1 as scalar;
```

At a high-level, deployments are verified by [randomly generating inputs](https://github.com/AleoNet/snarkVM/blob/be171ce0720544c2bc51e27e1674a62fff585adc/synthesizer/process/src/stack/deploy.rs#L107), [executing the program](https://github.com/AleoNet/snarkVM/blob/be171ce0720544c2bc51e27e1674a62fff585adc/synthesizer/process/src/stack/deploy.rs#L161), [extracting the assignment](https://github.com/AleoNet/snarkVM/blob/be171ce0720544c2bc51e27e1674a62fff585adc/synthesizer/process/src/stack/deploy.rs#L165), and [checking that the associated certificates match the structure of the assignment](https://github.com/AleoNet/snarkVM/blob/be171ce0720544c2bc51e27e1674a62fff585adc/algorithms/src/snark/varuna/varuna.rs#L280). **Note that the actual values do not matter. [Source](https://github.com/AleoNet/snarkVM/blob/be171ce0720544c2bc51e27e1674a62fff585adc/algorithms/src/snark/varuna/ahp/indexer/indexer.rs#L135)**

The core issue is that RNG used in generating random inputs can differ across different runs of [verify_deployment](https://github.com/AleoNet/snarkVM/blob/be171ce0720544c2bc51e27e1674a62fff585adc/synthesizer/process/src/stack/deploy.rs#L58).

The failing case manifests in the following way:
- The execution of `cast r0 into r1 as scalar;` creates an invalid scalar field element when `r0` is larger than the scalar field modulus.
- The resulting value is [ejected](https://github.com/AleoNet/snarkVM/blob/be171ce0720544c2bc51e27e1674a62fff585adc/synthesizer/process/src/stack/registers/store.rs#L93) before storing the result into the register `r1`.
- [Scalar::eject_value](https://github.com/AleoNet/snarkVM/blob/be171ce0720544c2bc51e27e1674a62fff585adc/circuit/types/scalar/src/lib.rs#L75) halts, if the scalar field element is invalid.
Consequently, a deployment of the above program can fail to verify depending on the random inputs that were used. 

## Solution

To fix this issue, this PR deterministically seeds the RNG used in deployment verification with the lower 64 bits of the deployment ID. This ensures that validators will use the same seed for a particular deployment, which ensures that they will always agree on whether the deployment is valid or not.

## Testing

This PR includes a test that attempts to deploy a program that loads and stores data types with invalid scalar field elements. The test verifies that the status of the deployment is consistent: always valid or always invalid, independent of the RNG.

CI for this branch is running [here](https://app.circleci.com/pipelines/github/ProvableHQ/snarkVM?branch=fix%2Fdeployment-verification-ci).

## Considerations

In addition to this PR, additional mitigations can be put in place to ensure that:
- `Scalar::eject_value` does not halt
- Loading and storing circuits to registers does not invoke `eject_value`. This is dependent on whether or not the invocations of `eject_value` are intended to serve as a value check or a less strict type check.

## Related PRs

Use this PR over #2534

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -3091,6 +3091,7 @@ dependencies = [
  "rand",
  "rayon",
  "serde_json",
+ "snarkvm-circuit",
  "snarkvm-console",
  "snarkvm-ledger-authority",
  "snarkvm-ledger-block",
@@ -3101,6 +3102,7 @@ dependencies = [
  "snarkvm-ledger-store",
  "snarkvm-ledger-test-helpers",
  "snarkvm-synthesizer",
+ "snarkvm-utilities",
  "time",
  "tracing",
 ]
@@ -3444,6 +3446,7 @@ dependencies = [
  "once_cell",
  "parking_lot",
  "rand",
+ "rand_chacha",
  "rayon",
  "serde_json",
  "snarkvm-circuit",
--- a/fields/src/fp_256.rs
+++ b/fields/src/fp_256.rs
@@ -767,7 +767,7 @@ impl<'a, P: Fp256Parameters> MulAssign<&'a Self> for Fp256<P> {
 
         r[3] = fa::mac_with_carry(r[3], (self.0).0[3], (other.0).0[0], &mut carry1);
         r[2] = fa::mac_with_carry(r[3], k, P::MODULUS.0[3], &mut carry2);
-        r[3] = carry1.wrapping_add(carry2);
+        r[3] = carry1 + carry2;
 
         // Iteration 1.
         r[0] = fa::mac(r[0], (self.0).0[0], (other.0).0[1], &mut carry1);
@@ -781,7 +781,7 @@ impl<'a, P: Fp256Parameters> MulAssign<&'a Self> for Fp256<P> {
 
         r[3] = fa::mac_with_carry(r[3], (self.0).0[3], (other.0).0[1], &mut carry1);
         r[2] = fa::mac_with_carry(r[3], k, P::MODULUS.0[3], &mut carry2);
-        r[3] = carry1.wrapping_add(carry2);
+        r[3] = carry1 + carry2;
 
         // Iteration 2.
         r[0] = fa::mac(r[0], (self.0).0[0], (other.0).0[2], &mut carry1);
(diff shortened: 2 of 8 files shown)


MaximizeYourGPU (approved):
Test rev

vicsn (changes requested):
I would recommend to **not use** the rng argument at all within `verify_deployment`. The existing assumptions which this PR makes also applies to the `burner_private_key`: its value is not important and if god forbid there's a bug in the key generation code then we'd rather want all validators to abort equally.

That being said, this recommendation is not absolutely essential, so if we're in a time crunch, we can also leave it for later. In a future PR, we can also consider removing the rng argument in `check_next_block` entirely to prevent future developers from introducing non-determinism.

vicsn (dismissed):
I would recommend to not use the rng argument at all within verify_deployment. The existing assumptions which this PR makes also applies to the burner_private_key: its value is not important and if god forbid there's a bug in the key generation code then we'd rather want all validators to abort equally.

That being said, this recommendation is not essential, so if we're in a time crunch, we can also leave it for later. In a future PR, we can also consider removing the rng argument in check_next_block entirely to prevent future developers from introducing non-determinism.

zkxuerb (dismissed):
LGTM

evanmarshall (dismissed):
:shipit:

lukenewman (approved):
🫡

raychu86:
Added a tweak to minimize change-set and clean up the testing; the logic should be equivalent.

Unfortunately this dismissed the previous reviews.

vicsn (approved):

zkxuerb (approved):

lukenewman (approved):

snarkworld (approved):
LGTM

evanmarshall (approved):

PR #2539, opened 2024-09-02
add test for `NUM_SEQUENCE_INSTRUCTIONS`

zhiqiangxu:
According to the comment:

```
/// Instruction count of the longest sequence.
```

The `NUM_SEQUENCE_INSTRUCTIONS` is assumed to be the instruction count of the longest sequence.

This PR adds a test case which ensures it's correct.

### DIFF
--- a/ledger/puzzle/epoch/src/synthesis/helpers/mod.rs
+++ b/ledger/puzzle/epoch/src/synthesis/helpers/mod.rs
@@ -346,6 +346,23 @@ pub(crate) mod tests {
     const ITERATIONS: u64 = 25;
 
     #[test]
+    fn test_longest_sequence() {
+        let instruction_set_weights = instruction_set::<CurrentNetwork>();
+        let mut longest_sequence: usize = 0;
+        for isw in instruction_set_weights {
+            if isw.0.len() > longest_sequence {
+                longest_sequence = isw.0.len();
+            }
+        }
+        assert!(
+            longest_sequence == NUM_SEQUENCE_INSTRUCTIONS,
+            "longest sequence length({}) not equal to NUM_SEQUENCE_INSTRUCTIONS({})",
+            longest_sequence,
+            NUM_SEQUENCE_INSTRUCTIONS
+        )
+    }
+
+    #[test]
     fn test_sample_instructions() {
         let mut rng = TestRng::default();
 


raychu86:
@d0cd How relevant is this test? It seems quite innocent, but unsure if it tests any desirable properties.

ledger/puzzle/epoch/src/synthesis/helpers/mod.rs
```diff
@@ -345,6 +345,23 @@ pub(crate) mod tests {
...
+        let instruction_set_weights = instruction_set::<CurrentNetwork>();
+        let mut longest_sequence: usize = 0;
+        for isw in instruction_set_weights {
+            if isw.0.len() > longest_sequence {
+                longest_sequence = isw.0.len();
+            }
+        }
+        assert!(
+            longest_sequence == NUM_SEQUENCE_INSTRUCTIONS,
+            "longest sequence length({}) not equal to NUM_SEQUENCE_INSTRUCTIONS({})",
```
d0cd, line 359:
```suggestion
            "The longest sequence length({}) is not equal to `NUM_SEQUENCE_INSTRUCTIONS`({})",
```

ledger/puzzle/epoch/src/synthesis/helpers/mod.rs
```diff
@@ -345,6 +345,23 @@ pub(crate) mod tests {
 
     const ITERATIONS: u64 = 25;
 
+    #[test]
+    fn test_longest_sequence() {
+        let instruction_set_weights = instruction_set::<CurrentNetwork>();
+        let mut longest_sequence: usize = 0;
+        for isw in instruction_set_weights {
```
d0cd, line 352: Could do this via an iterator for more clarity.
Something like: `instruction_set_weights.iter().map(|(_, sequence) sequence.len()).max()`

d0cd (commented):
Thanks @zhiqiangxu!
Overall a good test to have! Left some nits.

d0cd:
Thanks for the contribution @zhiqiangxu!
We love to see people add more safety checks to snarkVM.
In this case, I believe this is already checked [here](https://github.com/ProvableHQ/snarkVM/blob/4a22b0ada6e293bfcaa2711e4d61dc309a3649d3/ledger/puzzle/epoch/src/synthesis/helpers/instruction_set.rs#L2830)

vicsn:
Thank you for the contribution! Closing due to duplication with existing test.

PR #2543, opened 2024-09-04
fix `minimum_tree_size` for `KaryMerkleTree`

zhiqiangxu:
This issue is not detected because the `leaves` parameter is now ensured to be a power of 8: [source](https://github.com/AleoNet/snarkVM/blob/5bb50a845c774f10247657947b9f1c9ec7f21d81/ledger/puzzle/epoch/src/synthesis/program/to_leaves.rs#L45) .

### DIFF
--- a/console/collections/src/kary_merkle_tree/mod.rs
+++ b/console/collections/src/kary_merkle_tree/mod.rs
@@ -94,11 +94,9 @@ impl<LH: LeafHash<Hash = PH::Hash>, PH: PathHash, const DEPTH: u8, const ARITY: 
         // The minimum tree size is either a single root node or the calculated number of nodes plus
         // the supplied leaves, and empty hashes that pad up to the tree's arity (making every node full).
         let arity = ARITY as usize;
-        let all_nodes_are_full = leaves.len() % arity == 0;
-        let minimum_tree_size = std::cmp::max(
-            1,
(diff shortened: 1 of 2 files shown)


ljedrz (approved):
I think this is correct, but I'd recommend double-checking with someone else as well, as I'm a bit rusty with this object.

Also, it would be great if we could have a test proving that the current implementation fails under different assumptions, and that it works correctly with the change in place.

Also, we need a CI run.

zhiqiangxu:
It seems the CI is out , is there a plan to make it work soon?

vicsn:
> This issue is not detected because the leaves parameter is now ensured to be a power of 8

An example calculation for why the issue is currently not triggered would also be appreciated, will help future reviewers

ljedrz (commented):
Overriding the approval to comment mode in order to reflect the request for a test; as for CI runs, those are currently triggered manually by the Foundation.

ljedrz (changes requested):
Hmm, apparently only "request changes" will override my approval.

zhiqiangxu:
After further checking, I found that both are correct, but the original code is more memory efficient because of [this logic](https://github.com/AleoNet/snarkVM/blob/02994a1e34e7db8da6c27587537b165aa488868c/console/collections/src/kary_merkle_tree/mod.rs#L128-L133).

ljedrz:
Thanks for double-checking :pray:!

PR #2544, opened 2024-09-04
fix : mail typo

Budalebah:
fix : mail typo

## Motivation

(Write your motivation here)

## Test Plan

(Write your test plan here)

## Related PRs

(Link your related PRs here)


zosorock:
Thank you for the contribution @Budalebah but we prefer it the way it is.

PR #2545, opened 2024-09-05
make `variable_limit` tight

zhiqiangxu:
if `Self::num_variables() == limit`, should not allow new variables.


vicsn:
Appreciate the contribution, but I recommend to close the PR because:
1. Semantically, limit can be interpreted as exclusive
2. The previous setup was extensively tested so should not be adjusted without significant motivation

PR #2552, opened 2024-09-28
ci: update actions/checkout and actions/cache to v4

Killpit:
## Motivation

-Updated `actions/checkout` and `actions/cache` to v4 to upgrade the pipelines and provide modern functionality for the pipelines with the latest abilities possible according to GitHub documentation as these are the latest versions.


raychu86 (approved):
LGTM

PR #2554, opened 2024-10-07
Feat/test targets

vicsn:
## Motivation

Add feature flag to be able to lower coinbase and proof targets for testing purposes.

Keeping the original `test` feature in `MainnetV0` to avoid the need for a larger test refactor.

## Test Plan

Ran networks and landed solutions succesfully.

Need to run CI still.

## Related PRs

Replaces https://github.com/AleoNet/snarkVM/pull/2529


zkxuerb (approved):
LGTM

PR #2558, opened 2024-10-15
[Update] MSRV and clippy.

d0cd:
This PR updates the minimum Rust version to v1.81.0 and addresses clippy issues associated with the upgrade.
As a result, this PR:
- removes unused features
- refactors code
- updates CI configurations accordingly
These changes should be carefully reviewed to ensure that behavior is not changed.

This PR also introduces an update to the naming convention for `cache_key`s. 
The proposed convention is `v<SNARKVM_VERSION>-rust-<RUST_VERSION>-<NAME>`.
@reviewers if we are considering changes to the key naming convention, consider:
  - including a checksum of the `Cargo.lock` file which would capture any dependency updates
  - using a variable for Rust version and snarkOS version

CI runs are [here](https://app.circleci.com/pipelines/github/ProvableHQ/snarkVM?branch=update%2Frust-version%2Bclippy-ci)


vicsn:
Regarding on the failing parameters test: good chance it will pass if we just retrigger that single one. Because individually it doesn't surpass AWS' download rate limits.

vicsn (approved):

gluax (approved):

zosorock:
@vicsn 
> Regarding on the failing parameters test: good chance it will pass if we just retrigger that single one. Because individually it doesn't surpass AWS' download rate limits.

Yeah, I have seen this on and off, for some reason it fails from CI but always works when we do manual test. I pinged CircleCI about it and they said they are aware and that I should just "try again until it works". I have a suspicion it has to do with how snarkVM handles TLS. May be it needs to be update to run like Ray did here for the CDN?
https://github.com/AleoNet/snarkOS/pull/3408/files

zosorock (approved):

PR #2561, opened 2024-10-22
[Perf] Only use raw iterators with RocksDB and speed up ledger load

vicsn:
## Motivation

Replaces https://github.com/AleoNet/snarkVM/pull/2518

Original message:

The signatures of iteration-related methods (with the rocksdb feature) have been haunting my heap profiles for a long time now, and while I initially believed this was something we could fine-tune with some configuration, all such attempts were unsatisfactory. It turns out that the "default" (higher-level) RocksDB iterators are just inherently inefficient, and only the DBRawIterator is able to avoid a massive number of allocations, some of which may remain lingering in the RSS and contribute to its inflation over time. The raw iterator is somewhat trickier to use correctly, but since it's what the regular iterator is built upon, there is nothing particularly novel about it, and we're already using it to count records.

With these changes I ran profiling with both an instrumented OS allocator and jemalloc and - contrary to the usual results - it was the latter that experienced larger relative differences when loading the ledger:

    total allocs and temp allocs were both reduced by 98%
    ledger load time was down 14%
    max RSS was reduced by 13%

In addition, heaptrack ran 73% faster and produced a 99% smaller profile, which is very practical for future profiling needs.

Alongside these changes I realized that the current len_confirmed method - while working correctly and passing all tests - can be made more solid by including iterator validity checks; the only situation where this would truly be needed is if it was called on the very last map in the database and if the map was empty, but [this commit](https://github.com/AleoNet/snarkVM/commit/f8fc68ca22ce9369f3e82107c48f3738bf00af50) guards against that edge case.

The [final commit](https://github.com/AleoNet/snarkVM/commit/1dbf87418379118a74c2f53fd9f45b444c6c2d6b) is a further improvement on one of the changes from https://github.com/AleoNet/snarkVM/pull/2515 (and made possible with the switch to raw iterators).

## Test Plan

[CI run link](https://app.circleci.com/pipelines/github/ProvableHQ/snarkVM?branch=perf%2Fraw_rocks_iter-ci)


zkxuerb (approved):
LGTM 👏

ledger/store/src/helpers/rocksdb/internal/map.rs
```diff
@@ -277,17 +277,21 @@ impl<
 
         // Count the number of keys belonging to the map.
         let mut len = 0usize;
-        while let Some(key) = iter.key() {
-            // Only compare the map ID - the network ID is guaranteed to
-            // remain the same as long as there is more than a single map.
-            if key[2..][..2] != self.context[2..][..2] {
-                // If the map ID is different, it's the end of iteration.
+        while iter.valid() {
```
evanmarshall, line 280: What's the purpose of this change to the loop iteration?

As far as I can tell, the original loop: `while let Some(key) = iter.key()` should be equivalent but should avoid the redundant check? 

Is there additional validation somewhere that I'm missing?

ledger/store/src/helpers/rocksdb/internal/nested_map.rs
```diff
@@ -372,18 +372,22 @@ impl<
 
         // Count the number of keys belonging to the nested map.
         let mut len = 0usize;
-        while let Some(key) = iter.key() {
-            // Only compare the nested map - the network ID and the outer map
-            // ID are guaranteed to remain the same as long as there is more
-            // than a single map in the database.
-            if !key[PREFIX_LEN + 4..].starts_with(serialized_map) {
-                // If the nested map ID is different, it's the end of iteration.
+        while iter.valid() {
```
evanmarshall, line 375: Also applies here: https://github.com/AleoNet/snarkVM/pull/2561/files#r1827079669

evanmarshall (approved):
Looks great!

One potential concern about data consistency and unexpected mutations while the database is running. Should be we iterating over snapshots to ensure that concurrent writes don't potentially cause rare & non-deterministic errors that are difficult to reproduce?

Snapshots should be pretty lightweight overall and as long as we ensure it falls out of scope after it's used, it may be a good way to avoid potential errors from this performance optimization.

PR #2562, opened 2024-10-22
[Perf] Reduce allocs when parsing Program

vicsn:
## Motivation

Replaces https://github.com/AleoNet/snarkVM/pull/2516

Original PR message:

(transferred from https://github.com/ProvableHQ/snarkVM/pull/5, and asking @d0cd for a review as suggested there)

Program parsing is something that has always been prominent in node heap profiles, and this PR targets its highest-level bits.

While parsing its components, we currently use a combination of the functions many1 and alt, which cause us to clone the parsed input at least once (more if it's not the expected component) for every component; this can be avoided by removing the use of alt in favor of a new intermediate function which is able to determine the type of the parsed component in advance and select the right parsing function without "making guesses". This is what the 1st commit does.

The 2nd commit avoids all the clones of program components that we currently perform presumably as a workaround introduced in order to be able to manage the notoriously tricky nom errors. With some refactoring (i.e. applying map_res only to potential errors), we are able to remove many allocations.

These changes reduce the number of allocations caused by the parsing of a Program in a 15-minute run of a --dev node by ~8% (an absolute decrease of over 54k allocs), making the process faster. The impact on the performance of the node as a whole depends on the number and size of the Programs it has to process.

## Test Plan

[CI run link](https://app.circleci.com/pipelines/github/ProvableHQ/snarkVM?branch=perf%2Fprogram_parsing2-ci)


synthesizer/program/src/parse.rs
```diff
@@ -45,58 +45,70 @@ impl<N: Network, Instruction: InstructionTrait<N>, Command: CommandTrait<N>> Par
         // Parse the semicolon ';' keyword from the string.
         let (string, _) = tag(";")(string)?;
 
+        fn intermediate<N: Network, Instruction: InstructionTrait<N>, Command: CommandTrait<N>>(
```
d0cd, line 48: Nit. Add comment describing this function.

d0cd: It would be good to explain in comments why we use this method.

d0cd:
@ljedrz nice find! I'd like to confirm my understanding as I'm not an expert in `nom` internals.
Is the first optimization avoiding [this](https://github.com/rust-bakery/nom/blob/54557471141b73ca3b2d07be88d6709a43495b10/src/multi/mod.rs#L190) allocation?
Is the second optimization avoiding [this](https://github.com/rust-bakery/nom/blob/54557471141b73ca3b2d07be88d6709a43495b10/src/internal.rs#L619) allocation?

d0cd (commented):
In general, I am in favor of this PR.
I think it would be good to establish a more rigorous evaluation process for these sorts of parser changes.
For example, a regression test suite that checks against all programs deployed on-chain.

ljedrz:
@d0cd 
> Is the first optimization avoiding [this](https://github.com/rust-bakery/nom/blob/54557471141b73ca3b2d07be88d6709a43495b10/src/multi/mod.rs#L190) allocation?

It removes [this](https://github.com/AleoNet/snarkVM/pull/2562/commits/f70404a0791e46ffa022bfb0b0cd2eda0fd0d4c4#diff-eaacff66a7ddd6f00711cf72f1428bc66633782cca0e1ac37f4d5dd03530f4efL48) use of `alt`, so the avoided allocation is (ultimately) [this](https://docs.rs/nom/latest/src/nom/branch/mod.rs.html#135).

> Is the second optimization avoiding [this](https://github.com/rust-bakery/nom/blob/54557471141b73ca3b2d07be88d6709a43495b10/src/internal.rs#L619) allocation?

This one is actually simpler; it avoids [these clones](https://github.com/AleoNet/snarkVM/pull/2516/commits/39c48dd1cf13a0c0c791f7f9cea305ba3363a26d#diff-eaacff66a7ddd6f00711cf72f1428bc66633782cca0e1ac37f4d5dd03530f4efL86-L90) in snarkVM code.

synthesizer/program/src/parse.rs
ljedrz: good idea; it seems that there there is some heavy-lifting done around branches right now, though, so my suggestion is to add it separately in order to avoid any merge-related issues; I've added it to my TODO, so it won't be forgotten

zosorock:
@gluax had originally reviewed and approved #2516.

PR #2563, opened 2024-10-22
[Perf] Use heuristics to avoid allocations in Sanitizer::str_till_eol

vicsn:
## Motivation

Replaces https://github.com/AleoNet/snarkVM/pull/2517

Original PR message:

(transferred from https://github.com/ProvableHQ/snarkVM/pull/6, and asking @d0cd for a review as suggested there)

The Sanitizer is used very prominently in our parsing functions, and it is also a source of many allocations, most of which are temporary and avoidable.

The potential perf improvements are quite large, and I've measured them both with a 15-minute run of a --dev node and using hyperfine on a small binary that parsed all the valid .aleo programs currently present in the snarkVM codebase.

dev node:

    all allocs are down ~15%, of which almost all are temporary
    in Program::from_str specifically, allocs are reduced by ~64%, of which temp allocs ~88%

parsing all .aleo programs using Program::from_str:

    allocs are reduced by ~41%
    growth reallocs are down ~70%
    runtime is reduced by ~31%

## Test Plan

[CI run link](https://app.circleci.com/pipelines/github/ProvableHQ/snarkVM?branch=perf%2Fparsing_sanitizer2-ci)


console/network/environment/src/helpers/sanitizer.rs
```diff
@@ -117,12 +117,33 @@ impl Sanitizer {
...
+        if let Some((before, after)) = string.split_once('\n') {
+            let is_multiline = before.ends_with('\\');
+
+            if !is_multiline {
+                let contains_unsafe_chars = !before.chars().all(is_char_supported);
+
+                if !contains_unsafe_chars {
+                    Ok((after, before))
+                } else {
+                    recognize(Self::till(value((), Sanitizer::parse_safe_char), Self::eol))(before)
```
d0cd, line 131: Does this need to use a `till` operator until `Self::eol` since this is knows to be a single line string?

console/network/environment/src/helpers/sanitizer.rs
```diff
@@ -117,12 +117,33 @@ impl Sanitizer {
...
     /// Discard any leading newline.
     fn str_till_eol(string: &str) -> ParserResult<&str> {
-        map(
-            recognize(Self::till(alt((value((), tag("\\\n")), value((), Sanitizer::parse_safe_char))), Self::eol)),
-            |i| {
-                if i.as_bytes().last() == Some(&b'\n') { &i[0..i.len() - 1] } else { i }
-            },
-        )(string)
+        // A heuristic approach is applied here in order to avoid
+        // costly parsing operations in the most common scenarios.
```
d0cd, line 121: Nit. Can you explain the heuristic in the comment?

console/network/environment/src/helpers/sanitizer.rs
```diff
@@ -117,12 +117,33 @@ impl Sanitizer {
...
+        if let Some((before, after)) = string.split_once('\n') {
+            let is_multiline = before.ends_with('\\');
+
+            if !is_multiline {
+                let contains_unsafe_chars = !before.chars().all(is_char_supported);
+
+                if !contains_unsafe_chars {
+                    Ok((after, before))
```
d0cd, line 129: Is the goal to avoid allocations in [this](https://github.com/rust-bakery/nom/blob/54557471141b73ca3b2d07be88d6709a43495b10/src/multi/mod.rs#L287) method?

d0cd (commented):
Overall, I'm in favor of this PR. It would be helpful to add test cases for the this specific parser, especially for when the heuristic is and isn't taken.

console/network/environment/src/helpers/sanitizer.rs
ljedrz: good point; I'm not sure, but since I'm not super savvy with `nom`, I preferred to leave it intact

console/network/environment/src/helpers/sanitizer.rs
ljedrz: added to my TODO (as with the `Program`-related comment, I'd suggest to add it separately to avoid merge-related issues)

console/network/environment/src/helpers/sanitizer.rs
ljedrz: IIRC the allocations that are avoided are primarily:
- the ones from [`verify`](https://docs.rs/nom/latest/src/nom/combinator/mod.rs.html#428)
- the ones from [`recognize`](https://docs.rs/nom/latest/src/nom/combinator/mod.rs.html#513)
- the ones from [`value`](https://docs.rs/nom/latest/src/nom/combinator/mod.rs.html#460)
- but most importantly, the `alt`-related ones from [here](https://github.com/AleoNet/snarkVM/pull/2563/files#diff-218cced7e98aba78cca56fdda8f839eb6a20a2414541d2805ec337d4e7e51d01R136) and [here](https://github.com/ProvableHQ/snarkVM/blob/3d42aa04a058cd5f46a1880b421313e1c04a63dc/console/network/environment/src/helpers/sanitizer.rs#L92)

gluax (approved):

Meshiest (approved):

zkxuerb (approved):
LGTM

ljedrz:
@vicsn I updated the original branch with 2 commits addressing the notes on [extra docs](https://github.com/AleoNet/snarkVM/pull/2517/commits/5e928d8828e53bf3cd766073969e1ee79079628d) and [the redundant use of `eol`](https://github.com/AleoNet/snarkVM/pull/2517/commits/42d999b02eab19f2661bd6c57e4479761a254f0b). Feel free to cherry-pick those, they should apply cleanly.

@d0cd as for extra tests, note that `str_till_eol` is only ever used in `parse_comment`, and there are already several applicable [test cases](https://github.com/AleoNet/snarkVM/blob/c9c83f99a9634ed802c42cf42a601d9df9e58dc9/console/network/environment/src/helpers/sanitizer.rs#L271-L286) for it, including ones that would utilize the faster approach. Do you feel like those are missing any variants?

d0cd:
@ljedrz on initial scan, I don't see test case for a multi-line comment, multiple single line comments, multiple multi-line comments, and interleavings of the two. I'll leave it to your discretion, but I would at least recommend a multi-line comment and multiple single line comments.

I also saw that you use `eoi` in the update. Looks good to me, but I would also recommend explaining in a comment that this is valid because `before` is a single line string.

The reason I am being so insistent on comments is that these parsers are quite sensitive and not many people have a deep expertise in `nom`.

ljedrz:
@vics I addressed the comments in 2 new commits:
- https://github.com/AleoNet/snarkVM/pull/2517/commits/3dd61c2ea8d6818ff83f680bce945721dee1a813
- https://github.com/AleoNet/snarkVM/pull/2517/commits/39ab91f4f0016a27b885c9b78df78476219c0d02

ljedrz:
While not a big deal (extra test cases, docs, and one optimization), this PR was still missing the 4 extra commits mentioned at the end. I can include them in a follow-up shortly.

vicsn:
@ljedrz apologies I overlooked that I had to update my branch

PR #2565, opened 2024-10-25
[Feature] Introduce `BLOCK_SPEND_LIMIT`

vicsn:
## Motivation

This PR introduces `BLOCK_SPEND_LIMIT`, the maximum amount in microcredits that can be spent on finalizing executions and synthesizing deployments in a block. This gives much greater certainty of maximum blocktimes. The advantages of this limit, analogous to a gas limit, are:
- more consistent blocktimes, so improved chain UX and consensus scalability
- fees are no longer required to be high to prevent against DoS attacks
- ~more accurate (validator/mining) rewards if they're assuming a particular blocktime~ No longer relevant

The new limit is only enabled from `CONSENSUS_V2_HEIGHT` onwards.

A limit of 950 credits was chosen because:
- `BLOCKTIME` is supposed to be 10 seconds.
- Given the assumption 100 credits == 1 second of runtime, 950 credits translates to 9.5 seconds
- Which leaves space for `Solution` verification, this incurs significant free compute. 4 solutions take [180ms to verify on macbook M1 max](https://github.com/AleoNet/snarkVM/blob/staging/ledger/puzzle/epoch/docs/spec.md#benchmarks).
- To account for execution processing, `EXECUTION_FIXED_COST` of 2 credits is introduced. This is not reflected in fees, as it is subsidized, but it does count towards the `BLOCK_SPEND_LIMIT`. Benchmarks on M2 max imply between 15ms and 25ms verification time for executions.
- The biggest block seen on mainnet so far had `137` executions, which would still remain below this `BLOCK_SPEND_LIMIT`. 

### Open questions
- **Flexible limits**: could we have a `ROUND_SPEND_LIMIT*num_rounds` or `PROPOSAL_SPEND_LIMIT*num_proposals`? Or better: could we impose a spend limit per individual proposal?
- Only snarkOS-level change or only snarkVM-level change?
- Should we have to verify transactions/fees before enforcing the spend limit?
- Note for migration that existing programs will not retroactively be rejected.

### Out of scope / future work
- **Ensuring taking fees**: We should review and ensure we always take fees from aborted/invalid transactions. Because any block limit effectively lowers the available blockspace by a well targeted attacker who knows how to make their transactions be aborted. 950 credits can be filled through 9 executions or 19 deployments.
- **Reinserting transactions**: if transactions ever surpass the spend limit, they will be aborted. Ideally, they are reinserted into batch proposals so that users don't have to retransmit. However, this would require a significant reworking all over snarkOS to ensure duplicate transmission ids are accepted under this circumstance.

## Test Plan

- [ ] Local network succeeded, surviving the transition into the new consensus rules.
- [ ] Deployed canary network succeeded, surviving the transition, surviving the transition into the new consensus rules.

## Related PRs

Replaces https://github.com/AleoNet/snarkVM/pull/2371

### DIFF
--- a/synthesizer/process/src/cost.rs
+++ b/synthesizer/process/src/cost.rs
@@ -91,7 +91,7 @@ fn execution_storage_cost<N: Network>(size_in_bytes: u64) -> u64 {
 
 /// Returns the fixed cost for an execution.
 /// NOTE: this constant reflects the compute cost of an execution, but is not required to be paid by the user.
-pub fn execution_fixed_cost<N: Network>() -> u64 {
+pub const fn execution_fixed_cost<N: Network>() -> u64 {
     N::EXECUTION_FIXED_COST
 }
 
--- a/synthesizer/process/src/cost.rs
+++ b/synthesizer/process/src/cost.rs
@@ -13,13 +13,15 @@
 // See the License for the specific language governing permissions and
 // limitations under the License.
 
+use std::sync::Arc;
+
 use crate::{Process, Stack, StackProgramTypes};
 
 use console::{
     prelude::*,
     program::{FinalizeType, Identifier, LiteralType, PlaintextType},
 };
-use ledger_block::{Deployment, Execution};
+use ledger_block::{Deployment, Execution, Transaction};
 use synthesizer_program::{CastType, Command, Finalize, Instruction, Operand, StackProgram};
 
 /// Returns the *minimum* cost in microcredits to publish the given deployment (total cost, (storage cost, synthesis cost, namespace cost)).
@@ -37,7 +39,7 @@ pub fn deployment_cost<N: Network>(deployment: &Deployment<N>) -> Result<(u64, (
         .ok_or(anyhow!("The storage cost computation overflowed for a deployment"))?;
 
     // Compute the synthesis cost in microcredits.
-    let synthesis_cost = synthesis_cost(deployment)?;
+    let synthesis_cost = deployment_synthesis_cost(deployment)?;
 
     // Compute the namespace cost in credits: 10^(10 - num_characters).
     let namespace_cost = 10u64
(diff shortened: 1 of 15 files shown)


synthesizer/src/vm/finalize.rs
```diff
@@ -813,13 +844,45 @@ impl<N: Network, C: ConsensusStorage<N>> VM<N, C> {
...
+                if synthesis_cost.saturating_add(microcredits_spent_on_compute) > N::BLOCK_SPEND_LIMIT {
+                    return Some("Exceeds block spend limit".to_string());
+                }
+            }
+        }
+
+        // Activate extra checks from CONSENSUS_V2_HEIGHT onwards.
+        if current_block_height >= N::CONSENSUS_V2_HEIGHT {
+            // If the transaction is an execution, ensure that the total finalize cost does not exceed the block spend limit.
+            if let Transaction::Execute(_, execution, _) = transaction {
```
ljedrz: nit: it might be a good idea to swap these conditions in order to not suggest that the code block below contains the only `CONSENSUS_V2` logic (since there is also extra logic for deploy txs); alternatively, that extra logic applicable to deploy txs could be moved under this code block (without altering the order), keeping all the `V2` logic in one place

synthesizer/process/src/cost.rs
```diff
@@ -86,6 +89,12 @@ fn execution_storage_cost<N: Network>(size_in_bytes: u64) -> u64 {
     }
 }
 
+/// Returns the fixed cost for an execution.
+/// NOTE: this constant reflects the compute cost of an execution, but is not required to be paid by the user.
+pub fn execution_fixed_cost<N: Network>() -> u64 {
```
ljedrz: are there plans to extend this logic (in which case an alternative name might be more fitting)? I'm wondering about the rationale for the existence of a (non-const) function wrapper for a `const`

vicsn: I swear this was in my git staging area already: 3443c3ef7

> are there plans to extend this logic (in which case an alternative name might be more fitting)?

Unclear at the moment. Only other name I can imagine is `execution_compute_cost`?

synthesizer/src/vm/finalize.rs
vicsn: Agree I like the clarity: fcbbb688b

vicsn:
@zkxuerb @zosorock do you have permission to draft this PR? Otherwise I'll close it. Still needs significant thinking.

0xMattieu:
.

vicsn:
Closing in favour of https://github.com/ProvableHQ/snarkOS/pull/3471

PR #2568, opened 2024-10-28
[Perf] Downgrade a write guard on Process to read

ljedrz:
Moved from https://github.com/ProvableHQ/snarkVM/pull/19.

The reasoning is that there is no benefit to holding a write guard here, as in addition to keeping the object frozen, it prohibits any potential readers from accessing it during finalization; a read guard in its place still won't allow the `process` to be modified for the entire duration of the atomic batch, but it _will_ allow other potential concurrent readers in the meantime. Unless the state of the `process` (or any of its contents) during finalization should not be visible to anyone else or should be considered outright invalid until it concludes, a read guard is the more performant choice.

I haven't been able to measure the impact of this change locally, but I'm confident it would make a difference in networks involving more traffic.

### DIFF
--- a/synthesizer/src/vm/finalize.rs
+++ b/synthesizer/src/vm/finalize.rs
@@ -263,10 +263,10 @@ impl<N: Network, C: ConsensusStorage<N>> VM<N, C> {
 
             /* Perform the atomic finalize over the transactions. */
 
-            // Acquire the write lock on the process.
+            // Acquire the read lock on the process.
             // Note: Due to the highly-sensitive nature of processing all `finalize` calls,
-            // we choose to acquire the write lock for the entire duration of this atomic batch.
-            let process = self.process.write();
+            // we choose to acquire the read lock for the entire duration of this atomic batch.
+            let process = self.process.read();
 
             // Initialize a list of the confirmed transactions.
             let mut confirmed = Vec::with_capacity(num_transactions);


vicsn (approved):

synthesizer/src/vm/finalize.rs
```diff
@@ -263,10 +263,10 @@ impl<N: Network, C: ConsensusStorage<N>> VM<N, C> {
 
             /* Perform the atomic finalize over the transactions. */
 
-            // Acquire the write lock on the process.
+            // Acquire the read lock on the process.
             // Note: Due to the highly-sensitive nature of processing all `finalize` calls,
-            // we choose to acquire the write lock for the entire duration of this atomic batch.
-            let process = self.process.write();
+            // we choose to acquire the read lock for the entire duration of this atomic batch.
+            let process = self.process.read();
```
raychu86, line 269: The point of the write lock here is so that there is NO concurrency on finalize scope state. Because transaction verification is highly dependent on state that can be changed between transactions we would much rather lean on safety guarantees.

Imagine transaction A and B where they both spend 10 credits from the same mapping. If they are read concurrently, both are considered valid, which should not be the case.

 In addition the `atomic_finalize` function below also uses this same `let process = self.process.write();` pattern; this breaks the consistency.

At the very least there should be some testing guarantee the safety of this change, but **I do not believe this is safe**.

raychu86 (changes requested):

synthesizer/src/vm/finalize.rs
ljedrz: Since these strict requirements go against the assumption from the description, my recommendation is to close this PR and I'll soon propose one extending the related comments/docs so that it is clear in the future :+1:.

vicsn: +1 on closing this PR, while I think this PR is safe, it is not putting any guardrails in place to ensure future developments remain safe. Some comments to grow our (mostly my own) understanding:

> In addition the atomic_finalize function below also uses this same let process = self.process.write(); pattern; this breaks the consistency.

Indeed, but in contrast to finalize (which needs write access to *finalize*), speculate conceptually only needs read access.

> Imagine transaction A and B where they both spend 10 credits from the same mapping. If they are read concurrently, both are considered valid, which should not be the case.

`self.atomic_lock` prevents this. Moreover, what is critical: `Process` doesn't contain data related to transaction verification, which is why the read-lock currently suffices. So maybe the comment could be adjusted to something like: "There are no guardrails in place to ensure Process will never contain data relevant for transmission verification".

Maybe a targeted unit test or some smart types could be sufficient for that, but is likely too much effort for now.

raychu86: Agreed, `self.atomic_lock` does indeed prevent the double-spend issue.

I still think that any changes related to the finalize scope and state management requires more scrutiny and additional testing.

ljedrz:
Closing as per the comments; while this change would currently be safe due to the use of `atomic_lock`, future changes might render it unsafe. Without additional compile-time safeguards or at least dedicated tests, this modification needs to be shelved.

PR #2569, opened 2024-11-01
[ARC-0042] Adjust reward algorithms to utilize timestamps

raychu86:
## Motivation

This PR is intended to implement and address the concerns highlighted in [ARC-0042](https://github.com/AleoNet/ARCs/discussions/76).

The gist is that the current block times are not aligned with our current algorithms and causing the following:
1. `block_reward` is emitting at a higher inflation rate than the expected 5%
2. `coinbase_reward` is reducing too quickly and not over the expected 10 year timeline.

This PR addresses the problem by using `timestamps` as the peg rather than `block_height`. Both the inflation rate and the `coinbase_reward` reduction-over-time are on human timescales; it makes sense for us to align the formulas to the human timescale with `timestamps` rather than relying on `block_heights`, which are on a network timescale.

### `block_reward`:
- Rename original to `block_reward_v1`
- Implement `block_reward_v2`
- Use `time_since_last_block` to determine how much reward to emit for that particular block in relation to how much inflation that `time_since_last_block` seconds should cause.
- This `time_since_last_block` is bound by a block interval of 60 seconds to prevent the block reward from becoming too large in the event of a long block interval such as a network outage or network upgrade.

### `coinbase_reward`:
- Rename original to `coinbase_reward_v1`
- Implement `coinbase_reward_v2`
- Use `anchor_block_reward_at_timestamp` instead of `anchor_block_reward_at_height`
- Uses the number of seconds the block is to the 10 year mark to determine the reward. The closer we are, the less reward.

Some emission analysis can be seen in the comments below - https://github.com/AleoNet/snarkVM/pull/2569#issuecomment-2456020249

## **Migration:**
- Introduce `N::CONSENSUS_V2_HEIGHT` to Network trait 
    - This is the height for each network that snarkVM will swap from the `v1` reward impls to the `v2` reward impls.
- SnarkOS will require nodes to swap to the new version before height `N::CONSENSUS_V2_HEIGHT`, otherwise there may be some issues.

### **Mainnet** - Block 2,800,000 (~Dec 10, 2024 at the current 3.0s block times)

| Network | Release Date | Buffer | Consensus V2 |
| Canary  | Nov 12, 2024 | 3 days | Nov 15, 2024 |
| Testnet | Nov 19, 2024 | 3 days | Nov 22, 2024 |
| Mainnet | Dec 3, 2024  | 7 days | Dec 10, 2024 |

## Test Plan

Tests have been added to ensure that the V2 `block_reward` and `coinbase_reward` implementations match the expected behavior. 

CI can be found [here](https://app.circleci.com/pipelines/github/ProvableHQ/snarkVM?branch=feat%2Fblock_reward_v2-ci).

### DIFF
--- a/ledger/block/src/helpers/target.rs
+++ b/ledger/block/src/helpers/target.rs
@@ -18,9 +18,28 @@ use console::prelude::{Network, Result, ensure};
 /// A safety bound (sanity-check) for the coinbase reward.
 pub const MAX_COINBASE_REWARD: u64 = 190_258_739; // Coinbase reward at block 1.
 
-/// A the maximum block interval in seconds. This is used to cap the block interval in the V2 block reward calculation to
+/// A the maximum block interval in seconds. This is used to upper bound the block interval in the V2 block reward calculation to
 /// prevent the block reward from becoming too large in the event of a long block interval.
-const BLOCK_REWARD_V2_MAX_BLOCK_INTERVAL: i64 = 60; // 1 minute.
+const V2_MAX_BLOCK_INTERVAL: i64 = 60; // 1 minute.
+/// A the minimum block interval in seconds. This is used to lower bound the block interval in the V2 block reward calculation to
+/// prevent the block reward from becoming too small in the event of an extremely short block interval.
+const V2_MIN_BLOCK_INTERVAL: i64 = 1; // 1 second.
+
+/// Calculate the block reward based on the network’s consensus version, determined by the given block height.
+pub fn block_reward<N: Network>(
+    block_height: u32,
+    total_supply: u64,
+    block_time: u16,
+    time_since_last_block: i64,
+    coinbase_reward: u64,
+    transaction_fees: u64,
+) -> u64 {
+    // Determine which block reward version to use.
+    match block_height < N::CONSENSUS_V2_HEIGHT {
+        true => block_reward_v1(total_supply, block_time, coinbase_reward, transaction_fees),
+        false => block_reward_v2(total_supply, time_since_last_block, coinbase_reward, transaction_fees),
+    }
+}
 
 /// Calculate the V1 block reward, given the total supply, block time, coinbase reward, and transaction fees.
 ///     R_staking = floor((0.05 * S) / H_Y1) + CR / 3 + TX_F.
@@ -56,10 +75,10 @@ pub fn block_reward_v2(
     const SECONDS_IN_A_YEAR: u64 = 60 * 60 * 24 * 365;
     // Compute the annual reward: (0.05 * S).
     let annual_reward = total_supply / 20;
-    // Compute the seconds since last block with a maximum of `MAX_BLOCK_INTERVAL` seconds.
-    // Compute the block reward: (0.05 * S) * min(I, MAX_BLOCK_INTERVAL) / S_Y.
-    let block_reward =
-        annual_reward * time_since_last_block.min(BLOCK_REWARD_V2_MAX_BLOCK_INTERVAL) as u64 / SECONDS_IN_A_YEAR;
+    // Compute the seconds since last block with a maximum of `V2_MAX_BLOCK_INTERVAL` seconds and minimum of `V2_MIN_BLOCK_INTERVAL` seconds;
+    let time_since_last_block = time_since_last_block.max(V2_MIN_BLOCK_INTERVAL).min(V2_MAX_BLOCK_INTERVAL);
+    // Compute the block reward: (0.05 * S) * min(max(I, MIN_BLOCK_INTERVAL), MAX_BLOCK_INTERVAL) / S_Y.
+    let block_reward = annual_reward * time_since_last_block as u64 / SECONDS_IN_A_YEAR;
     // Return the sum of the block reward, coinbase reward, and transaction fees.
     block_reward + (coinbase_reward / 3) + transaction_fees
 }
@@ -331,7 +350,7 @@ pub fn to_next_targets<N: Network>(
 #[cfg(test)]
 mod tests {
     use super::*;
-    use console::network::{MainnetV0, prelude::*};
+    use console::network::{MainnetV0, TestnetV0, prelude::*};
 
     type CurrentNetwork = MainnetV0;
 
@@ -469,6 +488,53 @@ mod tests {
     }
 
     #[test]
+    fn test_block_reward() {
+        let mut rng = TestRng::default();
+
+        // Ensure that a block height of `TestnetV0::CONSENSUS_V2_HEIGHT` uses block reward V2.
+        let time_since_last_block = rng.gen_range(1..=V2_MAX_BLOCK_INTERVAL);
+        let reward = block_reward::<TestnetV0>(
+            TestnetV0::CONSENSUS_V2_HEIGHT,
+            TestnetV0::STARTING_SUPPLY,
+            TestnetV0::BLOCK_TIME,
+            time_since_last_block,
+            0,
+            0,
+        );
+        let expected_reward = block_reward_v2(TestnetV0::STARTING_SUPPLY, time_since_last_block, 0, 0);
+        assert_eq!(reward, expected_reward);
+
+        for _ in 0..100 {
+            // Check that the block reward is correct for the first consensus version.
+            let consensus_v1_height = rng.gen_range(0..TestnetV0::CONSENSUS_V2_HEIGHT);
+            let consensus_v1_reward = block_reward::<TestnetV0>(
+                consensus_v1_height,
+                TestnetV0::STARTING_SUPPLY,
+                TestnetV0::BLOCK_TIME,
+                0,
+                0,
+                0,
+            );
+            let expected_reward = block_reward_v1(TestnetV0::STARTING_SUPPLY, TestnetV0::BLOCK_TIME, 0, 0);
+            assert_eq!(consensus_v1_reward, expected_reward);
+
+            // Check that the block reward is correct for the second consensus version.
+            let consensus_v2_height = rng.gen_range(TestnetV0::CONSENSUS_V2_HEIGHT..u32::MAX);
+            let time_since_last_block = rng.gen_range(1..=V2_MAX_BLOCK_INTERVAL);
+            let consensus_v2_reward = block_reward::<TestnetV0>(
+                consensus_v2_height,
+                TestnetV0::STARTING_SUPPLY,
+                TestnetV0::BLOCK_TIME,
+                time_since_last_block,
+                0,
+                0,
+            );
+            let expected_reward = block_reward_v2(TestnetV0::STARTING_SUPPLY, time_since_last_block, 0, 0);
+            assert_eq!(consensus_v2_reward, expected_reward);
+        }
+    }
+
+    #[test]
     fn test_block_reward_v1() {
         let reward = block_reward_v1(CurrentNetwork::STARTING_SUPPLY, CurrentNetwork::BLOCK_TIME, 0, 0);
         assert_eq!(reward, EXPECTED_STAKING_REWARD);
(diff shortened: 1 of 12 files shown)


ledger/block/src/helpers/target.rs
```diff
@@ -108,6 +230,69 @@ fn anchor_block_reward_at_height(block_height: u32, starting_supply: u64, anchor
...
+    // Compute the unadjusted reward at the given block timestamp.
+    let reward_at_block_timestamp =
+        block_reward_at_timestamp(block_timestamp, genesis_timestamp, starting_supply, anchor_time);
+    // Compute the anchor block reward.
+    reward_at_block_timestamp.max(reward_at_year_9)
+}
+
+/// Returns the timestamp for a given year, relative to the genesis timestamp.
+pub const fn timestamp_at_year(genesis_timestamp: i64, num_years: u32) -> i64 {
+    // Calculate the number of seconds in a year.
```
vicsn: Blasphemy. To silence future reviewers from wasting time you can explicitly mention leap years are ignored because its not significant enough.

ledger/block/src/helpers/target.rs
```diff
@@ -108,6 +230,69 @@ fn anchor_block_reward_at_height(block_height: u32, starting_supply: u64, anchor
...
+    const fn block_reward_at_timestamp(
+        block_timestamp: i64,
+        genesis_timestamp: i64,
+        starting_supply: u64,
+        anchor_time: u16,
+    ) -> u128 {
+        // Calculate the timestamp at year 10.
+        let timestamp_at_year_10 = timestamp_at_year(genesis_timestamp, 10) as u128;
+        // Calculate the number of seconds elapsed in 10 years.
+        let number_of_seconds_in_10_years = timestamp_at_year(0, 10) as u128;
```
vicsn: I think this use of the function is slightly confusing, better to just do the calculation directly `(seconds_per_year * num_years)`  or make a new function `seconds_per_year`

ledger/block/src/helpers/target.rs
```diff
@@ -35,20 +58,83 @@ pub const fn block_reward(total_supply: u64, block_time: u16, coinbase_reward: u
...
+///     MIN_BI = Minimum block interval.
+///     MAX_BI = Maximum block interval.
+pub fn block_reward_v2(
+    total_supply: u64,
+    time_since_last_block: i64,
+    coinbase_reward: u64,
+    transaction_fees: u64,
+) -> u64 {
+    // Calculate the number of seconds in a year.
+    const SECONDS_IN_A_YEAR: u64 = 60 * 60 * 24 * 365;
```
vicsn: Defined twice, can be defined at the top of the file.

synthesizer/src/vm/finalize.rs
```diff
@@ -32,9 +32,11 @@ impl<N: Network, C: ConsensusStorage<N>> VM<N, C> {
     ///     `Ratify::BlockReward(block_reward)` and `Ratify::PuzzleReward(puzzle_reward)`
     ///     to the front of the `ratifications` list.
     #[inline]
+    #[allow(clippy::too_many_arguments)]
     pub fn speculate<'a, R: Rng + CryptoRng>(
         &self,
         state: FinalizeGlobalState,
+        time_since_last_block: i64, // TODO (raychu86): Consider moving this value into `FinalizeGlobalState`.
```
vicsn, line 39: Just a note: while it increases randomness of the seed, it also adds nontrivial computation because `hash_bhp768` is used in there, so I would then not add it to the preimage since we don't really need the pseudorandomness at this stage.

vicsn (commented):
Just nits today, will review the tests and ideas for testcases tomorrow

Since the code is the defacto documentation on the reward mechanisms at the moment, do you mind adding slightly more detail:
- In the docblock for `fn puzzle_reward` what the conceptual meaning is compared to e.g. the `coinbase_reward`
- In the docblock for `fn anchor_block_reward_at_timestamp` what the conceptual meaning is compared to e.g. the `coinbase_reward` (namely, it is an upper bound, ultimate payout depends on the target of the solutions)

synthesizer/src/vm/finalize.rs
raychu86: It's possible to add to `FinalizeGlobalState` and not include it as part of the seed, but I refrained from doing something so confusing and inconsistent.

d0cd:
We simulated the block rewards without coinbase rewards and observed the following results.
The y-axis is the block rewards and the x-axis is the UNIX timestamp in seconds.
Note that the end of the x-axis is 2030-09-04 23:59:59.
Each of the plots represent an assumed a "fixed" block time.
| Plot Name | Block Time (s)  |
|---|---|
| Rewards No Solutions-0-0  |  3  |
| Rewards No Solutions-1-0  | 5  |
| Rewards No Solutions-2-0  | 10  |
| Rewards No Solutions-3-0  | 15  |
| Rewards No Solutions-4-0  | 30  |
| Rewards No Solutions-5-0  | 45  |
| Rewards No Solutions-6-0  | 60  |
| Rewards No Solutions-7-0  | 90  |
| Rewards No Solutions-8-0  | 120 |

These results are consistent with desired goals of this PR. 
While each of the plots start with a different block reward emission rate, they all have the same rate after block 2M.
This indicates that after the `CONSENSUS_V2_HEIGHT` is reached, the base block reward emission rate will be steady regardless of block time.

plot-Rewards No Solutions

Below is a plot of anchor block rewards over time for the same set of block times.
As expected, we see that after block 2M, the anchor block reward rate converges to the desired decay function.

plot-Anchor Block Rewards

ledger/block/src/helpers/target.rs
raychu86: Resolved with [9098995](https://github.com/AleoNet/snarkVM/pull/2569/commits/9098995c53f6d01f0dc2171a6c51d72c8d1b5de7)

ledger/block/src/helpers/target.rs
raychu86: Resolved with [9098995](https://github.com/AleoNet/snarkVM/pull/2569/commits/9098995c53f6d01f0dc2171a6c51d72c8d1b5de7)

ledger/block/src/helpers/target.rs
raychu86: Resolved with [9098995](https://github.com/AleoNet/snarkVM/pull/2569/commits/9098995c53f6d01f0dc2171a6c51d72c8d1b5de7)

ledger/block/src/helpers/target.rs
```diff
@@ -18,13 +18,39 @@ use console::prelude::{Network, Result, ensure};
 /// A safety bound (sanity-check) for the coinbase reward.
 pub const MAX_COINBASE_REWARD: u64 = 190_258_739; // Coinbase reward at block 1.
 
-/// Calculate the block reward, given the total supply, block time, coinbase reward, and transaction fees.
+/// A the maximum block interval in seconds. This is used to upper bound the block interval in the V2 block reward calculation to
+/// prevent the block reward from becoming too large in the event of a long block interval.
+const V2_MAX_BLOCK_INTERVAL: i64 = 60; // 1 minute.
```
iamalwaysuncomfortable, line 23: Perhaps in the discussion of `ARC-0042` there should be some further discussion about how high this number *should* be. I think we're seeing in `monte-carlo` simulations run by @d0cd that emissions are lowered for shorter block times, but actually **increased** for higher block times (this makes sense with the math proposed). IMHO `60` is too high and increases emissions in the case that block times go over 10 seconds. 

This is likely to be the case as more validators and more traffic come onto the network, and perhaps antithetical to the original intent of the `ARC`. What do you think @raychu86 (and anyone else involved in this discussion) about proposing lowering this number to a maximum of perhaps `20` to incentivize the network to keep lower block times and not risk inflation in the case that block times go over 10 seconds?

iamalwaysuncomfortable:
> These results are consistent with desired goals of this PR

@d0cd I think they're consistent with the goals of this PR and are a faithful prototype of `ARC-0042`, but perhaps the numbers actually suggested in the ARC itself are inconsistent with the stated goals of the ARC to reduce inflation if block times go beyond 10 seconds.

The monte-carlo simulation shows the emission rate **increasing** (sometimes with quite a significant slope increase) for block times beyond 10 seconds. This doesn't reduce inflation if block times increase with the addition of further validators or more traffic. 

The discussion should be brought to `ARC-0042` about reducing the `V2_MAX_BLOCK_INTERVAL` from `60` to perhaps a lower number like `20`. @HarukaMa do you have any thoughts on this?

HarukaMa:
@iamalwaysuncomfortable Are you calculating the emission rate against timestamp for block time >10s? For per-block reward it's expected to increase because the block did cover more time, so it should have emitted more credits. If you plot it against the actual time though, the emission rate _should_ remain the same. (EDIT: remember that we are trying to address the issue that credits emission is too high relative to time, as the design is 5% block reward _per year_ regardless of how many blocks we have generated over the year. If the block interval is longer than the max block interval, the emission rate will actually decrease as the exceeded time won't be rewarded. This in turn could actually become an incentive to make validators push out blocks faster.)

EDIT: also thanks for mentioning me, wasn't aware there is already a PR for this.

raychu86:
@iamalwaysuncomfortable @HarukaMa 

The algorithm will have a `block_reward` emission rate of 5% as long as the block times are consistently within the `V2_MAX_BLOCK_INTERVAL`. Even if block times are consistently 60 seconds exactly, the emission rate should remain 5%. If the blocktime is consistently larger than `V2_MAX_BLOCK_INTERVAL` (because of network load or too many validators), then the emission rate will be lower than expected for the year and be less than 5% because we are capping the amount of rewards.

However, I do see your point that reducing this would incentivize block times that are always lower than `V2_MAX_BLOCK_INTERVAL`, because validators want to ensure maximum payout. Definitely worth considering, but it is a balance we need to make to help guarantee steady inflation and keep block times low enough.

I would suggest keeping at 60 seconds to be conservative or 30 seconds if we want to be a bit more aggressive.

ledger/block/src/helpers/target.rs
raychu86: See comment https://github.com/AleoNet/snarkVM/pull/2569#issuecomment-2457458722 for my thoughts.

ledger/block/src/verify.rs
```diff
@@ -388,9 +388,12 @@ impl<N: Network> Block<N> {
         )?;
 
         // Calculate the expected coinbase reward.
-        let expected_coinbase_reward = coinbase_reward(
+        let expected_coinbase_reward = coinbase_reward::<N>(
```
ljedrz, line 391: nit: once `coinbase_reward` is bounded by `N: Network`, all the parameters that are derived from `N` can be omitted to reduce boilerplate (they can be obtained ad-hoc in the function body)

ledger/block/src/verify.rs
```diff
@@ -402,9 +405,17 @@ impl<N: Network> Block<N> {
         let expected_transaction_fees =
             self.transactions.iter().map(|tx| Ok(*tx.priority_fee_amount()?)).sum::<Result<u64>>()?;
 
+        // Calculate the time since last block.
+        let time_since_last_block = timestamp.saturating_sub(previous_block.timestamp());
         // Compute the expected block reward.
-        let expected_block_reward =
-            block_reward(N::STARTING_SUPPLY, N::BLOCK_TIME, expected_coinbase_reward, expected_transaction_fees);
+        let expected_block_reward = block_reward::<N>(
```
ljedrz, line 411: ditto

ledger/block/src/verify.rs
raychu86: I had initially thought about this as well, but refrained to keep the formula open s.t. `N: Network` being passed into `coinbase_reward` and `block_reward` is only used for `N::CONSENSUS_V2_HEIGHT`. 

This enables better testing for variables that go into the actual calculation of the rewards rather than being opinionated on fixed trait values.

vicsn (approved):
L G T M

d0cd (approved):

ljedrz (approved):
LGTM :+1:

iamalwaysuncomfortable (commented):
> Even if block times are consistently 60 seconds exactly, the emission rate should remain 5%. If the blocktime is consistently larger than V2_MAX_BLOCK_INTERVAL (because of network load or too many validators), then the emission rate will be lower than expected for the year and be less than 5% because we are capping the amount of rewards.

The tradeoff of block time vs. rewards emission rate is something all the members of the Aleo community should do some deeper analysis of ultimately. However, if the community is currently staunchly committed to the `5%` inflation rate `60` seconds likely provides reasonable statistical guarantees that this rate will be preserved.

zkxuerb (approved):
LGTM

PR #2576, opened 2024-11-19
Expose all parallelization utility functions

vicsn:
## Motivation

Came across this when trying to parallelize stuff in snarkOS, and realized I couldn't access all known functions.

## Test Plan

No tests needed, only changes exposed functions.


ljedrz (approved):

console/network/environment/src/lib.rs
```diff
@@ -72,14 +72,21 @@ pub mod prelude {
...
         cfg_find_map,
         cfg_into_iter,
         cfg_iter,
         cfg_iter_mut,
         cfg_keys,
+        cfg_par_bridge,
         cfg_reduce,
+        cfg_reduce_with,
+        cfg_sort_by_cached_key,
+        cfg_sort_unstable_by,
```
iamalwaysuncomfortable, line 86: This one shouldn't be exposed as the documentation states it may not preserve order of equal elements. Exposing that makes me uncofmortable.

```suggestion
```

iamalwaysuncomfortable (commented):
This seems fine except for the exposure of `cfg_sort_unstable_by`

console/network/environment/src/lib.rs
ljedrz: That quality is perfectly fine in almost all scenarios, and it's faster.

raychu86 (approved):
LGTM as there are not logic changes and only expose more functions.

console/network/environment/src/lib.rs
raychu86: For the most part I think `cfg_sort_unstable_by` is safe if the comparator is defined clearly. I would hesitate to use it on more complex types, but we do use `sort_unstable_by` on some integers. 

I could be swayed either way, but I think for the most part as long as `cfg_sort_unstable_by` is being used responsibly, it's fine.

PR #2580, opened 2024-11-28
[Perf] Cache fee verification

vicsn:
## Motivation

When processing hundreds of transmissions during block generation, transaction verification is a large bottleneck, taking up 90% of compute. **Half** of that is spent verifying the SNARK proofs associated to fees. We should cache that. 

Some useful details:
- Despite changing caching logic, this does not require a migration because the cache is in memory and will be cleared on a necessary upgrade restart.
- There are 3 cases for the fee: its from a Deployment, Execution, or Fee transaction. Given that Fee transactions are only created during block generation and are only verified once, it is technically superfluous to add those to the cache. But its cleaner at negligible cost.
- Since the PR which introduced the cache, I realize we can cache potentially way more checks, but the SNARK verification is by far the largest cost anyway, so we can stick with the existing setup.

## Test Plan

- [x] Added cache state expectations to unit tests. `Fee` tests are not in there because the respective test-helper doesn't expose the rejected id.
- [x] Ran succesfully on deployed network

## Related PRs

- Cache was introduced in: https://github.com/AleoNet/snarkVM/pull/2270
- Importance of caching highlighted here: https://github.com/AleoNet/snarkOS/pull/3451


raychu86 (approved):
Logic overall LGTM! I don't recall why we didn't do the fee cacheing before when we had initially introduced it. 

One thing I would ask you to think about as the author of the PR is the following:

Fee's can be swapped between transactions and Executions/Deployments can be broadcasted with different Fee transitions. Does this PR cover those cases? (I think it does, but want to be doubly sure)

vicsn:
> Fee's can be swapped between transactions and Executions/Deployments can be broadcasted with different Fee transitions. Does this PR cover those cases? (I think it does, but want to be doubly sure)

Good question, we should be safe. We only cache verification on a `Transaction` basis. So caching for a transaction won't cause *any* verification to be avoided for transactions afterwards which have the same Fee.

vicsn:
> Logic overall LGTM! I don't recall why we didn't do the fee cacheing before when we had initially introduced it.

Also, I think that was at least partially because of my lack of understanding that the fee path also incurs a full proof verification.

PR #2581, opened 2024-11-29
Nit: Do not hardcode the hiding mode

vicsn:
## Motivation

Found this while browsing through the code. We should not be hardcoding the hiding mode, but set it during a test/library invocation. This particular instance was harmless, but should be fixed to safeguard future development nonetheless.

## Test Plan

No actual logic is changed worth testing.

A unit test failed, likely for an unrelated reason (parameter fetching), please retrigger 🙏


raychu86:
Huh, why did the Varuna tests that used `VarunaNonHidingMode` not fail since we were hardcoding `VarunaHidingMode`?

vicsn:
> Huh, why did the Varuna tests that used VarunaNonHidingMode not fail since we were hardcoding VarunaHidingMode?

Because the function we're calling, `index_polynomial_labels_single`, doesn't actually *use* anything related to the type. This PR does not change any runtime values.

It might make sense to move all of the pure functions in `AHPForR1CS` elsewhere, but as it is this PR is a pareto improvement for clarity.

raychu86 (approved):
LGTM

zkxuerb (approved):

PR #2582, opened 2024-12-02
Add additional `Input/Output` checks 

raychu86:
## Motivation

This PR adds additional input/output checks to transaction verification in order to cover all cases. Previously we added length checks in https://github.com/AleoNet/snarkVM/pull/2511; for completeness, we now add variant type checks as well.

CI is run [here](https://app.circleci.com/pipelines/github/ProvableHQ/snarkVM/413/workflows/c0c98765-3e0d-4b56-8be6-b255bdd6d8ec).

## Test Plan

Tested on local devnets and isonets. Sync tests have also been run to ensure that existing networks are fully compatible.

### DIFF
--- a/console/program/src/data_types/value_type/mod.rs
+++ b/console/program/src/data_types/value_type/mod.rs
@@ -22,6 +22,8 @@ use snarkvm_console_network::prelude::*;
 
 use enum_index::EnumIndex;
 
+pub type Variant = u8;
+
 #[derive(Clone, PartialEq, Eq, Hash, EnumIndex)]
 pub enum ValueType<N: Network> {
     /// A constant type.
@@ -36,6 +38,20 @@ pub enum ValueType<N: Network> {
     ExternalRecord(Locator<N>),
     /// A publicly-visible future.
     Future(Locator<N>),
+}
+
+impl<N: Network> ValueType<N> {
+    /// Returns the variant of the value type.
+    pub const fn variant(&self) -> Variant {
+        match self {
+            ValueType::Constant(..) => 0,
+            ValueType::Public(..) => 1,
+            ValueType::Private(..) => 2,
+            ValueType::Record(..) => 3,
+            ValueType::ExternalRecord(..) => 4,
+            ValueType::Future(..) => 5,
+        }
+    }
 }
 
 impl<N: Network> From<EntryType<N>> for ValueType<N> {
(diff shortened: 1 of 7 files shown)


vicsn (approved):

d0cd (approved):

console/program/src/data_types/value_type/mod.rs
```diff
@@ -38,6 +40,20 @@ pub enum ValueType<N: Network> {
     Future(Locator<N>),
 }
 
+impl<N: Network> ValueType<N> {
+    /// Returns the variant of the value type.
+    pub const fn variant(&self) -> Variant {
```
ljedrz, line 45: nit: the same could be achieved with `EnumIndex`, which is already derived for `ValueType`

synthesizer/process/src/verify_execution.rs
```diff
@@ -112,6 +112,14 @@ impl<N: Network> Process<N> {
             ensure!(function.inputs().len() == num_inputs, "The number of transition inputs is incorrect");
             ensure!(function.outputs().len() == num_outputs, "The number of transition outputs is incorrect");
 
+            // Ensure the input and output types are equivalent to the ones defined in the function.
+            // We only need to check that the variant type matches because we already check the hashes in
+            // the `Input::verify` and `Output::verify` functions.
+            let transition_input_variants = transition.inputs().iter().map(Input::variant).collect::<Vec<_>>();
```
ljedrz, line 118: nit: these allocations could be avoided via something like
```
ensure!(function.input_variants().iter().zip(transition.inputs().iter().map(Input::variant)).all(|(v1, v2)| v1 == v2));
```
(length equality has already been checked at this point)

synthesizer/program/src/function/mod.rs
```diff
@@ -69,6 +69,11 @@ impl<N: Network, Instruction: InstructionTrait<N>, Command: CommandTrait<N>> Fun
         self.inputs.iter().map(|input| input.value_type()).cloned().collect()
     }
 
+    /// Returns the function input type variants.
+    pub fn input_variants(&self) -> Vec<Variant> {
```
ljedrz, line 73: nit: these could return an `impl Iterator<Item = Variant>`

ljedrz (approved):

zkxuerb (approved):
LGTM

synthesizer/program/src/function/mod.rs
raychu86: I left it consistent with the existing `output_types` and `input_types` functions. We can revisit in the future for possible optimizations.

console/program/src/data_types/value_type/mod.rs
raychu86: I added a `Variant` for consistency and to be more explicit in matching the `Variant` between types. I felt it was a bit vague to use EnumIndex.

synthesizer/process/src/verify_execution.rs
raychu86: ditto below: can leave for future optimization

PR #2584, opened 2024-12-05
[Feature] Introduce a console literal for raw bytes

ljedrz:
This PR is a draft/RFC proposal for a new console literal containing raw bytes, which would allow us to e.g. handle hashes ([example use case](https://github.com/AleoNet/ARCs/discussions/81)) and proofs in Aleo programs.

ARC link: https://github.com/ProvableHQ/ARCs/discussions/83

Open points:
- the encoding to be used in programs
  - it is currently a hex encoding for simplicity, but that incurs a 100% penalty in size when encoded
  - e.g. `base64` would reduce that penalty to ~33%
- the [length limit(s)](https://github.com/AleoNet/snarkVM/compare/staging...ljedrz:feat/console_literal_bytes?expand=1#diff-c71ebd9fc8799045828435c4a5636c18ad9555ea752491c9f555faedc73f0b6cR60-R63)
  - is `u16::MAX` enough/too much?
  - should the length prefix be an order of magnitude greater for potential future compatibility? this is the reason for the additional implementation in `snarkvm-utilities`
- a call to `unimplemented!()` in [`circuit::Literal::new`](https://github.com/AleoNet/snarkVM/compare/staging...ljedrz:feat/console_literal_bytes?expand=1#diff-2e86ff26ab0d9dfe8c50f1a57f846af5dd70e193c776030b2512941b66b447bbR99)
  - it is safe to make it `unimplemented!` as long as `Inject::new` is not intended to be used externally
  - this new feature could be extended with a `circuit` counterpart, removing the need for it

### DIFF
--- a/circuit/program/src/data/literal/cast/mod.rs
+++ b/circuit/program/src/data/literal/cast/mod.rs
@@ -115,6 +115,9 @@ macro_rules! impl_cast_body {
             LiteralType::String => {
                 bail!(concat!("Cannot cast a ", stringify!($type_name), " literal to a string type."))
             }
+            LiteralType::Bytes => {
+                bail!(concat!("Cannot cast a ", stringify!($type_name), " literal to a bytes type."))
+            }
         }
     };
 }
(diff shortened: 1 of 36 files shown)


ljedrz:
Closing due to inactivity - the codebase has diverged significantly, and it would need to be rewritten.

PR #2585, opened 2024-12-05
[Fix] Update permissions on Continuous Benchmarks CI

raychu86:
## Motivation

After the Repo ownership change, the Continuous Benchmarks github action has not been completed due to permissions differences. This fixes the permissions and allows benchmarks to be run whenever PRs are merged into `staging`, `testnet`, or `mainnet`. 

This allows us to accurately track the performance history.

Benchmarks can be found here - https://provablehq.github.io/snarkVM/dev/bench/

Note: I believe the previous benchmarks were also lost with the repo migration

### DIFF
--- a/.github/workflows/benchmarks.yml
+++ b/.github/workflows/benchmarks.yml
@@ -4,8 +4,6 @@ on:
   push:
     branches:
       - 'staging'
-      - 'testnet'
-      - 'mainnet'
 
 jobs:
   # Run benchmarks and stores the output to a file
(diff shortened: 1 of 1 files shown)


vicsn:
Consider waiting with this until we confirmed whether to keep the staging/testnet/mainnet branches.

vicsn:
Shall we do just the `staging` branch? That'll simplify reporting, and I don't think we'll miss too much from the occasional hotfix - which will find its way into the staging branch anyway.

raychu86:
> Shall we do just the `staging` branch? That'll simplify reporting, and I don't think we'll miss too much from the occasional hotfix - which will find its way into the staging branch anyway.

Agreed, I set it to `staging` only

vicsn (approved):

PR #2586, opened 2024-12-20
[Feat] Increase `MAX_CERTIFICATES` on `MainnetV0`

raychu86:
## Motivation

This PR plans to increase the number `N::MAX_CERTIFICATES` value for `MainnetV0`. The main result of this change is that the maximum number of validators in a committee is also increased. Since Mainnet was launched, the codebase has undergone some extensive upgrades and optimizations and based on internal/external testing, we believe the number of validators can be expanded safely. The original maximum is 16 and this change increases that value to 25 (subject to change). 

Increasing the number of validators creates a more robust and more decentralized network. This increase also has a number of other effects (increasing TPS, increasing network overhead, and increasing block size to name a few), so we must be vigilant when deciding to expand this number. As we further improve the network, we can iteratively increase the maximum number of validators in order to ensure network stability, even if it feels conservative.

## **Migration:**
- Introduce `N::CONSENSUS_V3_HEIGHT` to Network trait 
    - This is the height for each network that snarkVM will swap from the `v2` consensus to `v3` consensus.
- SnarkOS will require nodes to swap to the new version before height `N::CONSENSUS_V3_HEIGHT`, otherwise there may be some issues.

### **Mainnet** - Block 4,900,000 (~Feb 18, 2025 at the current 3.0s block times)

| Network | Release Date | Buffer | Consensus V3 |
| Canary  | Jan 21, 2025 | 3 days | Jan 24, 2025 |
| Testnet | Jan 28, 2025 | 3 days | Jan 31, 2025 |
| Mainnet | Feb 11, 2025  | 7 days | Feb 18, 2025 |

## Test Plan

Tests have been added to ensure that the migration properly transitions the old `MAX_CERTIFICATES_BEFORE_V3` value to the new `MAX_CERTIFICATES` value.

CI can be found [here](https://app.circleci.com/pipelines/github/ProvableHQ/snarkVM?branch=feat%2Fincrease_validators-ci).

## Related PRs

Migration will coincide with https://github.com/AleoNet/snarkVM/pull/2575.

## TODO

- [ ] Determine final numbers for `MAX_CERTIFICATES` and migration block heights
- [ ] Run extensive devnet testing

### DIFF
--- a/console/network/src/canary_v0.rs
+++ b/console/network/src/canary_v0.rs
@@ -136,10 +136,16 @@ impl Network for CanaryV0 {
     /// The block height from which consensus V2 rules apply.
     #[cfg(not(any(test, feature = "test")))]
     const CONSENSUS_V2_HEIGHT: u32 = 2_900_000;
-    // TODO (raychu86): Update this value based on the desired canary height.
     /// The block height from which consensus V2 rules apply.
     #[cfg(any(test, feature = "test"))]
-    const CONSENSUS_V2_HEIGHT: u32 = 0;
+    const CONSENSUS_V2_HEIGHT: u32 = 10;
+    // TODO: (raychu86): Update this value based on the desired mainnet height.
+    // The block height from which consensus V3 rules apply.
+    #[cfg(not(any(test, feature = "test")))]
+    const CONSENSUS_V3_HEIGHT: u32 = 4_560_000;
+    /// The block height from which consensus V3 rules apply.
+    #[cfg(any(test, feature = "test"))]
+    const CONSENSUS_V3_HEIGHT: u32 = 20;
     /// The network edition.
     const EDITION: u16 = 0;
     /// The genesis block coinbase target.
--- a/console/network/src/canary_v0.rs
+++ b/console/network/src/canary_v0.rs
@@ -145,7 +145,7 @@ impl Network for CanaryV0 {
     const CONSENSUS_V3_HEIGHT: u32 = 4_560_000;
     /// The block height from which consensus V3 rules apply.
     #[cfg(any(test, feature = "test"))]
-    const CONSENSUS_V3_HEIGHT: u32 = 20;
+    const CONSENSUS_V3_HEIGHT: u32 = 11;
     /// The network edition.
     const EDITION: u16 = 0;
     /// The genesis block coinbase target.
--- a/console/network/src/canary_v0.rs
+++ b/console/network/src/canary_v0.rs
@@ -166,6 +166,8 @@ impl Network for CanaryV0 {
     const INCLUSION_FUNCTION_NAME: &'static str = MainnetV0::INCLUSION_FUNCTION_NAME;
     /// The maximum number of certificates in a batch.
     const MAX_CERTIFICATES: u16 = 100;
+    /// The maximum number of certificates in a batch before consensus V3 rules apply.
+    const MAX_CERTIFICATES_BEFORE_V3: u16 = 100;
     /// The network name.
     const NAME: &'static str = "Aleo Canary (v0)";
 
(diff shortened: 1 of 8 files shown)


console/network/src/canary_v0.rs
```diff
@@ -160,6 +166,8 @@ impl Network for CanaryV0 {
     const INCLUSION_FUNCTION_NAME: &'static str = MainnetV0::INCLUSION_FUNCTION_NAME;
     /// The maximum number of certificates in a batch.
     const MAX_CERTIFICATES: u16 = 100;
+    /// The maximum number of certificates in a batch before consensus V3 rules apply.
+    const MAX_CERTIFICATES_BEFORE_V3: u16 = 100;
```
vicsn, line 170: Hmm I guess this was very intentional but I have a weak preference for using sth like:
-  `max_certificates_from_genesis` or `max_certificates_from_v0`
-  `max_certificates_from_v3`

Then all these constants are named similarly and reflect similar logic.

Curious to hear your thoughts.

synthesizer/src/vm/finalize.rs
```diff
@@ -995,8 +999,13 @@ impl<N: Network, C: ConsensusStorage<N>> VM<N, C> {
                     bond_validator_addresses.into_iter().filter(|address| !committee_members.contains(address)).count();
                 // Compute the next committee size.
                 let next_committee_size = committee_members.len().saturating_add(num_new_validators);
+                // Determine the maximum committee size to use.
+                let max_committee_size = match state.block_height() < N::CONSENSUS_V3_HEIGHT {
```
vicsn, line 1003: Given that we know there will be future additions to this: we can match just on the block height, and compare in the branches `if height < v3_height`. Then in the future we only have to adjust two branches.

Ideally that same format is used for all consensus height matches, but I usually defer formatting refactors until necessary.

console/network/src/canary_v0.rs
vicsn: An alternative approach worth mentioning is explicitly specifying all dynamic values for every consensus update, e.g. in a config file. I think Aptos does this. I'll tinker on it and get back, but that'll be after this PR is done.

kpandl:
I've tested the PR, specifically bonding validators at various block heights (old and new consensus heights) and for various sizes (within the max number of validators, and above it). Most of the tests work fine.

There's just one edge case we should fix: it is now possible to run a new network with the latest `N::MAX_CERTIFICATES` from height 0 on. For example, you can run a new mainnet network using the devnet script with 25 validators. This shouldn't be possible, at height 0 only 16 validators should be allowed. To fix it, we need to make the following section dependent on the block height: https://github.com/AleoNet/snarkVM/blob/be48f30bfeb78f327a57d08f23f7a0aff4e65a95/ledger/committee/src/lib.rs#L73-L82

console/network/src/canary_v0.rs
raychu86: I see your suggestions but have a weak preference for the current version for the following reasons:

1. It keeps `MAX_CERTIFICATES` as the canonical value and only introduced a singular new variable.
2. This reduces the change set because all downstream usages of `MAX_CERTIFICATES` won't need to be updated.


Happy to reconsider if there is a stronger push for an alternative

synthesizer/src/vm/finalize.rs
raychu86: Agreed. I also think that our current approach of `CONSENSUS_V*_HEIGHT` is a bit naive and will lead to inevitable tech-debt/confusion in as the number of migrations increase.

Ideally, we have a cleaner approach overall, but for now I'll keep things simple until we have a more concrete effort.

Separately, we can do a formatting refactor in a subsequent PR. I would opt to keep this format the same to match the existing usages of  `CONSENSUS_V*_HEIGHT`

raychu86:
ATTENTION: THIS PR WILL BE SUPERCEDED BY https://github.com/ProvableHQ/snarkVM/pull/2595 DUE TO BRANCHING ISSUES. 

ALL DISCUSSION WILL BE MOVED THERE AND THIS BRANCH WILL BE CLOSED

PR #2588, opened 2025-01-09
Hashes - ARC-81

anstylian:
## Motivation

This pull request introduces changes to the Hash instructions in the `synthesizer/program/src/logic/instruction/operation/hash.rs` file. The main focus is on improving the handling of destination types and adding support for additional array types in hashing operations (sha3 and keccak).

It is a step forward for the implementation of [ARC-81](https://github.com/ProvableHQ/ARCs/discussions/81).

## Test Plan
Let me know how you prefer to test this

### DIFF
--- a/synthesizer/program/src/logic/instruction/operation/hash.rs
+++ b/synthesizer/program/src/logic/instruction/operation/hash.rs
@@ -105,14 +105,19 @@ fn check_number_of_operands(variant: u8, opcode: Opcode, num_operands: usize) ->
 }
 
 /// Returns 'true' if the destination type is valid.
-fn is_valid_destination_type<N: Network>(destination_type: &PlaintextType<N>) -> bool {
-    !matches!(
-        destination_type,
+fn is_valid_destination_type<N: Network>(variant: u8, destination_type: &PlaintextType<N>) -> bool {
+    match destination_type {
         PlaintextType::Literal(LiteralType::Boolean)
-            | PlaintextType::Literal(LiteralType::String)
-            | PlaintextType::Struct(..)
-            | PlaintextType::Array(..)
-    )
+        | PlaintextType::Literal(LiteralType::String)
+        | PlaintextType::Struct(..) => false,
+        PlaintextType::Array(array) => match variant {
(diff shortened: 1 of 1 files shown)


synthesizer/program/src/logic/instruction/operation/hash.rs
```diff
@@ -196,6 +203,120 @@ impl<N: Network, const VARIANT: u8> HashInstruction<N, VARIANT> {
...
+    let hash: Vec<_> = hash
+        .chunks(8)
+        .map(|element| {
+            let value = element
+                .iter()
+                .enumerate()
+                .fold(0, |acc, (i, bit)| if bit.eject_value() { acc | (1 << i) } else { acc });
+
+            let value = console::types::U8::new(value);
+            circuit::Plaintext::Literal(
```
d0cd: I don't believe this circuit is constrained sufficiently. You'll need to use the `from_bits` and `to_bits` utilities.

synthesizer/program/src/logic/instruction/operation/hash.rs
```diff
@@ -256,14 +462,9 @@ impl<N: Network, const VARIANT: u8> HashInstruction<N, VARIANT> {
             (_, PlaintextType::Struct(..)) => bail!("Cannot hash into a struct"),
             (_, PlaintextType::Array(..)) => bail!("Cannot hash into an array (yet)"),
         };
-        // Cast the output to the destination type.
-        let output = match self.destination_type {
```
d0cd, line 260: Generally, it would nice to have fewer deltas on the variants that are not being changed.

d0cd:
Thanks @anstylian for taking an attempt at ARC-81! I think this approach will need careful consideration before we move forward. Specifically we'd like to:
- Ensure the approach is consistent in design. In particular, I'd like to be certain that the bespoke variants don't create confusion for users and tech debt in the future.
- Ensure the circuits are sufficiently constrained.
- Thoroughly test the implementation. One property that needs testing is that the output arrays are correctly constructed. I imagine there are more.

anstylian:
@d0cd, do you have any suggestions for testing?
1. Should we consider adding the result of the hash in arrays of u128?
1. Should we consider a specific type for hash results?

PR #2590, opened 2025-01-10
[Refactor] simplify and unify the storage setup for tests

ljedrz:
The hybrid nature of our storage makes testing it tricky, and there are several spots where we are currently using a mix of in-memory and persistent storage, which complicates things further. The changes proposed in this PR are something that I've hoped to see in snarkVM for a while now, and since some other, concurrent snarkVM work requires a stricter separation between the storage variants, it was a good opportunity to finally propose them.

The following is a rough summary of these changes:
- all the tests now implicitly either use the persistent storage or the in-memory one; there is no mixing, unless it's done explicitly
- as far as storage is concerned, all the tests can now run concurrently
- the `open_(.*)testing` methods are no longer needed, there is no test special-casing for rocksdb either
- the `open` methods no longer require `S` to be `Clone`
- instead of a `static DB`, there is now a `static DATABASES`, which has more than a single member only in tests
- the `ledger-with-rocksdb` CI job should be runnable again

What this fixes:
- there are currently several storage-related tests that use the `serial` attribute, which **doesn't** affect the use of `static` objects (specifically `static DB`), meaning the tests involving RocksDB storage can get cross-contaminated
- some tests are currently (re)using the same folder that's used in `--dev` mode, which can lead to cross-contamination and/or can overwrite storage related to development tests
- one of the tests related to [stacks](https://github.com/ProvableHQ/snarkVM/pull/2591) failed due to the wrong storage type being used

Future improvements:
- if the `open` methods for the stores didn't rely so much on `StorageMode` (i.e. if we could use an instance of actual storage to open the stores), we wouldn't need `static DATABASES` (or `DB`, as is the case currently); that being said, I'd prefer it to be done separately
- we currently only store the production genesis block, but doing the same for the testnet one would speed up some of the tests (it would allow us to run some of the slower tests in `--release` mode)

Requires: https://github.com/ProvableHQ/aleo-std/pull/12


ljedrz:
While that one program (`rand_chacha_check`) in `test_vm_execute_and_finalize` is returning a different outcome than the expectation regardless of features (`rocksdb`, `test`), I'm not sure why it happens just yet, but the PR may already be reviewed, and hopefully the cause will be revealed in the process.

For reference, this is the one mismatch:
```
============================================================
TEST
------------------------------------------------------------
program: test_rand.aleo
function: rand_chacha_check
inputs:
- 1field
- true

============================================================
EXPECTED
------------------------------------------------------------
verified: true
execute:
  test_rand.aleo/rand_chacha_check:
    outputs:
    - '{"type":"future","id":"818878742790741579153893179075772445872751227433677932822653185952935999557field","value":"{\n  program_id: test_rand.aleo,\n  function_name: rand_chacha_check,\n  arguments: [\n    1field,\n    true\n  ]\n}"}'
speculate: the execution was rejected
add_next_block: succeeded.

============================================================
ACTUAL
------------------------------------------------------------
verified: true
execute:
  test_rand.aleo/rand_chacha_check:
    outputs:
    - '{"type":"future","id":"818878742790741579153893179075772445872751227433677932822653185952935999557field","value":"{\n  program_id: test_rand.aleo,\n  function_name: rand_chacha_check,\n  arguments: [\n    1field,\n    true\n  ]\n}"}'
speculate: the execution was accepted
add_next_block: succeeded.

============================================================
```

ljedrz:
I have a [branch](https://github.com/ProvableHQ/aleo-std/compare/main...ljedrz:feat/test_storage_mode?expand=1) ready with a `StorageMode::Test` proposal, which would reduce the reliance of this PR on the `test` feature. I'd be happy to update this PR based on it if we can release an update to `aleo-std` soon.

ljedrz:
Update: I no longer have that one test mismatch with updated `aleo-std`, so I'll double-check the other tests and update this PR shortly.

ljedrz:
I updated the PR to use a new `aleo-std-storage` feature which makes things simpler and more explicit; it hasn't been merged yet, so there are temporarily some dependencies on a specific branch, but other than that this PR is ready for review now.

ledger/store/src/helpers/rocksdb/internal/mod.rs
```diff
@@ -122,12 +124,16 @@ impl Database for RocksDB {
     ///
     /// In production mode, the database opens directory `~/.aleo/storage/ledger-{network}`.
     /// In development mode, the database opens directory `/path/to/repo/.ledger-{network}-{id}`.
-    fn open<S: Clone + Into<StorageMode>>(network_id: u16, storage: S) -> Result<Self> {
-        static DB: OnceCell<RocksDB> = OnceCell::new();
+    /// In tests, the database opens an ephemeral directory in the OS temporary folder.
```
vicsn, line 134: Nit: can you add a comment for Custom for completeness

vicsn (changes requested):
LGTM, once this and aleo-std are approved you can update the cargo.toml and we may need Ray for publishing aleo-std.

ledger/store/src/helpers/rocksdb/internal/mod.rs
ljedrz: Done; I'll include it with the next rebase or any other review-related changes.

vicsn:
`aleo-std` has been updated. Please also merge (not rebase) `staging`

ljedrz:
@vicsn Done. If you'd like, I can always rebase after the reviews are complete.

vicsn (approved):
LGTM.

Please run the following test and report the results back: run a local devnet using the `devnet.sh` script before this PR, then restart the nodes without removing any database with this PR.

The same will be done by Provable's canary validator before the next canary release as part of the standard testing routine.

ljedrz:
@vicsn Test passed.

raychu86:
There are some merge conflicts that came from the "lock-tick" changes

ljedrz:
I updated the branch; one manual adjustment was needed.

niklaslong (approved):

PR #2578, opened 2024-11-21
Revert "Feat/lazy load deployment"

raychu86:
This PR reverts the optimization changes from AleoNet/snarkVM#2553 as it was a prematurely merged without enough testing.

PR #2553, opened 2024-10-07
Feat/lazy load deployment

vicsn:
## Motivation (from original PR)

We've seen that RSS size is correlated with deployments loaded (and staying) in memory. This PR keeps MAX_STACKS programs in memory (chosen at `Self::MAX_TRANSMISSIONS_PER_BATCH as usize * Self::MAX_CERTIFICATES as usize * 4`).

## Test Plan

Tested by running a local and deployed devnet, deploying programs and executing them. Then restarting the network and trying to execute these programs. The Stacks were lazily loaded from storage, execution worked. Also tested executing programs that are not deployed.

## Related PRs

Replaces https://github.com/AleoNet/snarkVM/pull/2519

PR #2591, opened 2025-01-10
Only keep track of recently used stacks in memory.

vicsn:
## Motivation

RSS growth is correlated with deployments. By lazy loading deployments, hopefully unbounded memory growth is eliminated.

The cache will take up at most MAX_PROGRAM_DEPTH x MAX_IMPORTS x 100kb x 10 ~= 4GB of data.

Note that programs, represented as `Stack`s, have imports. This means the cache is essentially a DAG with multiple roots. To avoid memory leaks, we only evict root stacks from the cache. In the following diagram, you can see in the call graph below where the cache is (temporarily) locked and updated. The "long" loop has some nasty edge cases, it would be better to pass all imports directly into `load_deployment` and `Stack::new`, but that would be a much bigger refactor.

Screenshot 2025-01-10 at 16 30 57

## Test Plan

- [ ] Unit tests pass. Requires https://github.com/ProvableHQ/snarkVM/pull/2590 to fix the `test_real_example_cache_evict` test.
- [ ] Local network passes, using tx-cannon to deploy and fetch multiple sets of maximally nested deployments.
- [ ] Deployed network passes, using tx-cannon to deploy and fetch multiple sets of maximally nested deployments.
- [ ] Consider fixing how test storage works, which currently leaks states across tests.

## Related PRs

https://github.com/ProvableHQ/snarkVM/pull/2519
https://github.com/ProvableHQ/snarkVM/pull/2553
https://github.com/ProvableHQ/snarkVM/pull/2578


console/network/src/lib.rs
```diff
@@ -195,6 +195,14 @@ pub trait Network:
     /// The maximum number of certificates in a batch.
     const MAX_CERTIFICATES: u16;
 
+    /// The maximum number of stacks in the process.
+    /// Must be at least Self::MAX_PROGRAM_DEPTH * Self::MAX_IMPORTS to be safe.
+    #[cfg(not(any(test, feature = "test")))]
+    const MAX_STACKS: usize = Self::MAX_PROGRAM_DEPTH * Self::MAX_IMPORTS * 10;
+    /// The maximum number of stacks in the process.
+    #[cfg(any(test, feature = "test"))]
+    const MAX_STACKS: usize = 65;
```
ljedrz: is this just a randomly chosen lower number that worked fine, or a minimum below which the tests don't pass?

ledger/narwhal/batch-header/src/lib.rs
```diff
@@ -66,6 +66,7 @@ impl<N: Network> BatchHeader<N> {
     /// The maximum number of rounds to store before garbage collecting.
     pub const MAX_GC_ROUNDS: usize = 100;
     /// The maximum number of transmissions in a batch.
```
ljedrz:
```suggestion
```

console/network/src/lib.rs
vicsn: A minimum below which the tests don't pass. But ideally, I should just get rid of this testing constant because future tests might require more (up to `Self::MAX_PROGRAM_DEPTH * Self::MAX_IMPORTS`).

synthesizer/process/src/lib.rs
```diff
@@ -66,30 +66,61 @@ use synthesizer_program::{
...
+#[derive(Clone)]
+pub struct Process<N: Network> {
+    /// The universal SRS.
+    universal_srs: Arc<UniversalSRS<N>>,
+    /// The Stack for credits.aleo
+    credits: Option<Arc<Stack<N>>>,
+    /// The mapping of program IDs to stacks.
+    stacks: Arc<Mutex<LruCache<ProgramID<N>, Arc<Stack<N>>>>>,
+    /// The storage.
+    store: Option<ConsensusStore<N, ConsensusDB<N>>>,
```
ljedrz: perhaps feature-gating would work on just the `store` member? i.e.
```
    /// The storage.
    #[cfg(feature = "rocks")]
    store: Option<ConsensusStore<N, ConsensusDB<N>>>,
    #[cfg(not(feature = "rocks"))]
    store: Option<ConsensusStore<N, ConsensusMemory<N>>>,
```

synthesizer/process/src/lib.rs
```diff
@@ -66,30 +66,61 @@ use synthesizer_program::{
...
+}
+
+#[cfg(not(feature = "rocks"))]
 #[derive(Clone)]
 pub struct Process<N: Network> {
     /// The universal SRS.
     universal_srs: Arc<UniversalSRS<N>>,
```
ljedrz, line 91: while we're at it, the `UniversalSRS` doesn't need to be in an `Arc`, as it's already arced internally and derives `Clone`
```suggestion
    universal_srs: UniversalSRS<N>,
```

synthesizer/process/src/lib.rs
```diff
@@ -108,7 +139,7 @@ impl<N: Network> Process<N> {
         lap!(timer, "Synthesize credits program keys");
 
         // Add the 'credits.aleo' stack to the process.
-        process.add_stack(stack);
+        process.credits = Some(Arc::new(stack));
```
ljedrz, line 128: would it make sense for `Stack::new` to already return an `Arc<Stack>`?

synthesizer/process/src/lib.rs
```diff
@@ -66,30 +66,61 @@ use synthesizer_program::{
...
     pub fn setup<A: circuit::Aleo<Network = N>, R: Rng + CryptoRng>(rng: &mut R) -> Result<Self> {
         let timer = timer!("Process:setup");
-
         // Initialize the process.
-        let mut process = Self { universal_srs: Arc::new(UniversalSRS::load()?), stacks: IndexMap::new() };
+        let mut process = Self {
+            universal_srs: Arc::new(UniversalSRS::load()?),
+            credits: None,
+            stacks: Arc::new(Mutex::new(LruCache::new(NonZeroUsize::new(N::MAX_STACKS).unwrap()))),
+            store: None,
```
ljedrz: since the store is now a possible member, perhaps it would be a good idea to reflect it in this function's name that it doesn't produce one? e.g. `setup_wo_store` or `new_without_store`

synthesizer/process/src/lib.rs
```diff
@@ -123,28 +154,103 @@ impl<N: Network> Process<N> {
...
+        let programs_to_add = stack.all_external_stacks();
+        // Obtain the lock on the stacks.
+        let mut stacks = self.stacks.lock();
+        // Determine which stacks still need to be added into the process.
+        let programs_to_add = programs_to_add
+            .into_iter()
+            .chain(std::iter::once((*stack.program_id(), stack))) // add the root stack.
+            .unique_by(|(id, _)|*id) // don't add duplicates.
+            .filter(|(program_id, _)| program_id != &ProgramID::<N>::from_str("credits.aleo").unwrap()) // don't add the credits.aleo stack.
+            .filter(|(program_id, _)| !stacks.contains(program_id)) // don't add stacks present in the cache.
```
ljedrz: the filter should ideally be singular, or we may end up dealing with 2 instances of the `Filter` iterator and slow things down
```suggestion
            .filter(|(program_id, _)| {
                program_id != &ProgramID::<N>::from_str("credits.aleo").unwrap() // don't add the credits.aleo stack.
                && !stacks.contains(program_id) // don't add stacks present in the cache.
            })
```

synthesizer/process/src/lib.rs
```diff
@@ -123,28 +154,103 @@ impl<N: Network> Process<N> {
...
+            .collect::<Vec<_>>();
+        // Determine the required capacity.
+        let current_capacity = stacks.cap().get().saturating_sub(stacks.len());
+        let mut required_capacity = programs_to_add.len().saturating_sub(current_capacity);
+        if required_capacity > 0 {
```
ljedrz:
```suggestion
        if required_capacity != 0 {
```
nit: `clippy` might complain about it someday

synthesizer/process/src/lib.rs
```diff
@@ -123,28 +154,103 @@ impl<N: Network> Process<N> {
...
+            debug!("Evicting {required_capacity} stacks from the cache.");
+        }
+        // If the new stacks require more capacity, attempt to remove the least recently used stacks.
+        while required_capacity > 0 {
```
ljedrz:
```suggestion
        while required_capacity != 0 {
```
ditto

synthesizer/process/src/lib.rs
```diff
@@ -123,28 +154,103 @@ impl<N: Network> Process<N> {
...
+        // Determine which stacks still need to be added into the process.
+        let programs_to_add = programs_to_add
+            .into_iter()
+            .chain(std::iter::once((*stack.program_id(), stack))) // add the root stack.
+            .unique_by(|(id, _)|*id) // don't add duplicates.
+            .filter(|(program_id, _)| program_id != &ProgramID::<N>::from_str("credits.aleo").unwrap()) // don't add the credits.aleo stack.
+            .filter(|(program_id, _)| !stacks.contains(program_id)) // don't add stacks present in the cache.
+            .collect::<Vec<_>>();
+        // Determine the required capacity.
+        let current_capacity = stacks.cap().get().saturating_sub(stacks.len());
```
ljedrz:
```suggestion
        let remaining_capacity = stacks.cap().get().saturating_sub(stacks.len());
```
more clear IMO

synthesizer/process/src/lib.rs
```diff
@@ -123,28 +154,103 @@ impl<N: Network> Process<N> {
...
+        // If the new stacks require more capacity, attempt to remove the least recently used stacks.
+        while required_capacity > 0 {
+            // To avoid dangling stacks, we only remove stacks that do not have more than 1 reference (e.g. as external stack).
+            // For this not to loop endlessly, we assume to always quickly find stacks with just one reference, so:
+            // 1. queried stacks must be short-lived, which is essential anyway to avoid memory leaks.
+            // 2. queried stacks must expire before a querying thread queries the next stack.
+            let root_program_ids = stacks
+                .iter()
+                .rev()
+                .filter_map(|(program_id, stack)| (std::sync::Arc::<_>::strong_count(stack) == 1).then_some(program_id))
```
ljedrz:
```suggestion
                .filter_map(|(program_id, stack)| (Arc::strong_count(stack) == 1).then_some(program_id))
```
doesn't that work? we've already imported `Arc` into this module

synthesizer/process/src/lib.rs
```diff
@@ -123,28 +154,103 @@ impl<N: Network> Process<N> {
...
+                .take(required_capacity)
+                .cloned()
+                .collect::<Vec<_>>();
+            // Lower the required capacity.
+            required_capacity = required_capacity.saturating_sub(root_program_ids.len());
```
ljedrz: just for excruciating clarity I'd move this right below the popping (which would require `root_program_ids` to be iterated by reference, which should be fine) or subtract one-by-one with every pop

synthesizer/process/src/lib.rs
```diff
@@ -123,28 +154,103 @@ impl<N: Network> Process<N> {
...
+        // Determine the required capacity.
+        let current_capacity = stacks.cap().get().saturating_sub(stacks.len());
+        let mut required_capacity = programs_to_add.len().saturating_sub(current_capacity);
+        if required_capacity > 0 {
+            debug!("Evicting {required_capacity} stacks from the cache.");
+        }
+        // If the new stacks require more capacity, attempt to remove the least recently used stacks.
+        while required_capacity > 0 {
+            // To avoid dangling stacks, we only remove stacks that do not have more than 1 reference (e.g. as external stack).
+            // For this not to loop endlessly, we assume to always quickly find stacks with just one reference, so:
```
ljedrz: if we're worried that there is a slight chance we wouldn't find stacks to pop, perhaps we could break and return an error once a specified number of iterations is reached?

synthesizer/process/src/lib.rs
```diff
@@ -123,28 +154,103 @@ impl<N: Network> Process<N> {
...
+        for (program_id, stack) in programs_to_add {
+            stacks.put(program_id, stack);
+        }
+        Ok(())
+    }
+
+    /// Returns the size of the cache.
+    #[inline]
+    pub fn num_stacks(&self) -> usize {
+        self.stacks.lock().len()
```
ljedrz, line 226: depending on the frequency of calls to `num_stacks()`, we might actually want to use an `RwLock<LruCache>` this time and `read` here instead

synthesizer/process/src/lib.rs
```diff
@@ -123,28 +154,103 @@ impl<N: Network> Process<N> {
...
+    }
+
     /// Initializes a new process.
     #[inline]
-    pub fn load() -> Result<Self> {
-        let timer = timer!("Process::load");
+    pub fn load_from_storage(storage_mode: Option<StorageMode>) -> Result<Self> {
+        let timer = timer!("Process::load_from_storage");
+
+        debug!("Opening storage");
```
ljedrz: this is a bit too vague to me if I were to see it in logs at random; perhaps we can specify the context a bit more or remove this and the following `debug` log if they're no longer needed for dev purposes

synthesizer/process/src/lib.rs
```diff
@@ -123,28 +154,103 @@ impl<N: Network> Process<N> {
...
+        let storage_mode = storage_mode.clone().ok_or_else(|| anyhow!("Failed to get storage mode"))?;
```
ljedrz:
```suggestion
        let storage_mode = storage_mode.ok_or_else(|| anyhow!("Failed to get storage mode"))?;
```

synthesizer/process/src/lib.rs
```diff
@@ -123,28 +154,103 @@ impl<N: Network> Process<N> {
...
+        // try to lazy load the stack
```
ljedrz:
```suggestion
        // Try to lazily load the stack.
```

ljedrz: that being said, I don't think this comment applies here specifically

synthesizer/process/src/lib.rs
```diff
@@ -170,9 +276,24 @@ impl<N: Network> Process<N> {
...
+        process.credits = Some(Arc::new(stack));
+
+        finish!(timer, "Process::load_from_storage");
+        // Return the process.
+        Ok(process)
+    }
+
+    /// Initializes a new process without creating the 'credits.aleo' program.
+    #[inline]
+    pub fn load_no_storage() -> Result<Self> {
```
ljedrz, line 296: is the only difference between the current `setup` function and this one the addition of `credits.aleo`? in that case one should ideally be redefined using the other

synthesizer/process/src/lib.rs
```diff
@@ -203,29 +329,68 @@ impl<N: Network> Process<N> {
...
+                Err(err) => debug!("Could not retrieve transaction ID for program ID: {err}"),
+            }
+        }
+
+        false
+    }
+
+    /// Returns `true` if the process contains the program with the given ID.
+    #[inline]
+    pub fn contains_program_in_memory(&self, program_id: &ProgramID<N>) -> bool {
```
ljedrz:
```suggestion
    pub fn contains_program_in_cache(&self, program_id: &ProgramID<N>) -> bool {
```
just to avoid suggesting that it's related to the in-memory storage

synthesizer/process/src/lib.rs
```diff
@@ -203,29 +329,68 @@ impl<N: Network> Process<N> {
...
+            return true;
+        }
+        // Retrieve the stores.
+        if let Some(store) = self.store.as_ref() {
+            let transaction_store = store.transaction_store();
+            let deployment_store = transaction_store.deployment_store();
+            // Check if the program ID exists in the storage.
+            match deployment_store.find_transaction_id_from_program_id(program_id) {
+                Ok(Some(_)) => return true,
+                Ok(None) => debug!("Program ID not found in storage"),
```
ljedrz: is this actionable (if it's somewhat unexpected, we might want to `WARN` instead)? we could also include the program ID in the log

synthesizer/process/src/lib.rs
```diff
@@ -203,29 +329,68 @@ impl<N: Network> Process<N> {
...
+
+        false
+    }
+
+    /// Returns `true` if the process contains the program with the given ID.
+    #[inline]
+    pub fn contains_program_in_memory(&self, program_id: &ProgramID<N>) -> bool {
+        // Check if the program ID is 'credits.aleo'.
+        if self.credits.as_ref().map_or(false, |stack| stack.program_id() == program_id) {
+            return true;
```
ljedrz, line 367: what if the `Process` was created using `load_no_storage`?

synthesizer/process/src/lib.rs
```diff
@@ -203,29 +329,68 @@ impl<N: Network> Process<N> {
...
-        // Retrieve the stack.
-        let stack = self.stacks.get(&program_id).ok_or_else(|| anyhow!("Program '{program_id}' does not exist"))?;
-        // Ensure the program ID matches.
-        ensure!(stack.program_id() == &program_id, "Expected program '{}', found '{program_id}'", stack.program_id());
-        // Return the stack.
-        Ok(stack)
+        // Check if the program is 'credits.aleo'.
+        if program_id == ProgramID::<N>::from_str("credits.aleo")? {
+            return self.credits.clone().ok_or_else(|| anyhow!("Failed to get stack for 'credits.aleo'"));
+        }
```
ljedrz: I think [this](https://github.com/ProvableHQ/snarkVM/pull/2591/files#diff-8462dc9ab98f319ac224ee6e645a5cc14132ad1aaa1ca7a15344e5abeb6e534bR358) way of checking is faster

synthesizer/process/src/lib.rs
```diff
@@ -203,29 +329,68 @@ impl<N: Network> Process<N> {
...
+        // Try to retrieve the stack from the LRU cache.
+        if let Some(stack) = self.stacks.lock().get(&program_id) {
+            // Ensure the program ID matches.
+            ensure!(
+                stack.program_id() == &program_id,
+                "Expected program '{}', found '{program_id}'",
+                stack.program_id()
+            );
```
ljedrz: is a mismatch possible? I'd expect this to be a development-only safeguard

ljedrz (commented):
First review pass complete :+1:. I'll probably do another one after the comments are addressed.

synthesizer/process/src/lib.rs
vicsn: Hmm I think I prefer the flexibility and more idiomatic return of Stack

synthesizer/process/src/lib.rs
vicsn: Only using it in tests, will try to conditionally compile

synthesizer/process/src/lib.rs
```diff
@@ -173,48 +177,65 @@ impl<N: Network> Process<N> {
...
-            // 2. queried stacks must expire before a querying thread queries the next stack.
+            // To avoid dangling stacks, we only remove stacks which are not imported by any other stack.
             let root_program_ids = stacks
                 .iter()
                 .rev()
-                .filter_map(|(program_id, stack)| (std::sync::Arc::<_>::strong_count(stack) == 1).then_some(program_id))
+                .filter_map(|(program_id, (_, num_parents))| (*num_parents == 0).then_some(program_id))
                 .take(required_capacity)
                 .cloned()
                 .collect::<Vec<_>>();
```
ljedrz, line 186: this shouldn't need to be collected anymore

vicsn: The iterator is holding (immutable) references to a part of `stacks`. As long as part of the iterator lives, I cannot call any `stacks.pop(&program_id)`.

ljedrz: ah, that's fair; I thought we only needed to collect it to have the count 👍

synthesizer/process/src/lib.rs
vicsn: This breaks at least one other place in the code which assumes this is an Arc, probably trivial to fix but can be done outside of this PR.

synthesizer/process/src/lib.rs
vicsn: No longer using strong_count here.

synthesizer/process/src/lib.rs
vicsn: Done

synthesizer/process/src/lib.rs
vicsn: Changed to a new design fc0ec5f37

synthesizer/process/src/lib.rs
vicsn: Removing them

synthesizer/process/src/lib.rs
vicsn: Then I think this logic is still accurate: it'll only return `true` if credits.aleo is actually stored.

synthesizer/process/src/lib.rs
vicsn: Agreed, removed it.

synthesizer/process/src/lib.rs
vicsn: 8b1a28acb

synthesizer/process/src/lib.rs
vicsn: d3ab60211

synthesizer/process/src/lib.rs
vicsn: 9b61b7805

synthesizer/process/src/lib.rs
vicsn: Good one f812a4b5f

synthesizer/process/src/lib.rs
vicsn: c630a369a

synthesizer/process/src/lib.rs
vicsn: f3fb877b3

synthesizer/process/src/lib.rs
vicsn: f1022e3e8

synthesizer/process/src/lib.rs
vicsn: Indeed with your other suggestion there's only `load_no_storage` now (and a `load_web`)

synthesizer/process/src/lib.rs
vicsn: b1e2a0a8c

console/network/src/lib.rs
vicsn: 12345aa05

synthesizer/process/src/lib.rs
```diff
@@ -123,28 +140,123 @@ impl<N: Network> Process<N> {
...
-    /// Adds a new stack to the process.
+    /// Adds a new stack to the LRU cache in the process.
     /// If you intend to `execute` the program, use `deploy` and `finalize_deployment` instead.
     #[inline]
-    pub fn add_stack(&mut self, stack: Stack<N>) {
-        // Add the stack to the process.
-        self.stacks.insert(*stack.program_id(), Arc::new(stack));
+    pub fn add_stack(&self, stack: Arc<Stack<N>>) -> Result<()> {
+        // Construct the 'credits.aleo' program ID.
+        let credits_id = ProgramID::<N>::from_str("credits.aleo")?;
```
vicsn: TODO: referencing `self.credits` wil be faster

vicsn: 815eecd95

synthesizer/process/src/lib.rs
```diff
@@ -150,7 +150,10 @@ impl<N: Network> Process<N> {
     #[inline]
     pub fn add_stack(&self, stack: Arc<Stack<N>>) -> Result<()> {
         // Construct the 'credits.aleo' program ID.
-        let credits_id = ProgramID::<N>::from_str("credits.aleo")?;
+        let credits_id = self
+            .credits
+            .as_ref()
+            .map_or(ProgramID::<N>::from_str("credits.aleo").unwrap(), |stack| *stack.program_id());
```
ljedrz, line 156:
```suggestion
            .map_or_else(|| ProgramID::<N>::from_str("credits.aleo").unwrap(), |stack| *stack.program_id());
```
(to avoid eager evaluation)

synthesizer/process/src/lib.rs
```diff
@@ -123,28 +140,125 @@ impl<N: Network> Process<N> {
...
+        // Collect all direct and indirect external stacks.
+        let programs_to_add = stack.all_external_stacks();
+        // Obtain the lock on the stacks.
+        let mut stacks = self.stacks.lock();
+        // Determine which stacks still need to be added into the process.
+        let programs_to_add = programs_to_add
+            .into_iter()
+            .chain(std::iter::once((*stack.program_id(), stack))) // add the root stack.
+            .unique_by(|(id, _)|*id) // don't add duplicates.
```
ljedrz, line 165:
```suggestion
            .unique_by(|(id, _)| *id) // don't add duplicates.
```
tiny nit, `rustfmt` might have missed it

synthesizer/process/src/lib.rs
```diff
@@ -182,7 +311,12 @@ impl<N: Network> Process<N> {
     #[cfg(feature = "wasm")]
     pub fn load_web() -> Result<Self> {
         // Initialize the process.
-        let mut process = Self { universal_srs: Arc::new(UniversalSRS::load()?), stacks: IndexMap::new() };
+        let mut process = Self {
+            universal_srs: Arc::new(UniversalSRS::load()?),
+            credits: None,
+            stacks: Arc::new(Mutex::new(LruCache::new(NonZeroUsize::new(N::MAX_STACKS).unwrap()))),
+            store: None,
+        };
```
ljedrz, line 319:
```suggestion
        let mut process = Self::load_no_storage()?;
```

synthesizer/process/src/lib.rs
```diff
@@ -123,28 +140,125 @@ impl<N: Network> Process<N> {
...
 
         // Initialize the process.
-        let mut process = Self { universal_srs: Arc::new(UniversalSRS::load()?), stacks: IndexMap::new() };
+        let mut process = Self {
+            universal_srs: Arc::new(UniversalSRS::load()?),
+            credits: None,
+            stacks: Arc::new(Mutex::new(LruCache::new(NonZeroUsize::new(N::MAX_STACKS).unwrap()))),
+            store: Some(store),
+        };
```
ljedrz, line 261:
```suggestion
        let mut process = Self::load_no_storage()?;
        process.store = Some(store);
```

ljedrz (commented):
2nd review pass complete, left a few comments; also, it seems like one of the new tests is currently failing.

kpandl:
Tested it on a 5 validator network, with 4 instance doing nested deployments, 4 instances doing nested executions, and a program probe instance running every 500 ms (total ~100 program deployments).
Validators reliably halted after ~70 deployments.

For testing, added the locktick features to snarkVM (commit `c12cbf66d334159108d75f85d910d073420869b8` [here](https://github.com/kpandl/snarkVM/commits/stack_cache_reloaded_reduced_max_stacks/)) and snarkOS (commit `246996bb2cdbc0dc75d4029b21aad1d32db6af2d` [here](https://github.com/ProvableHQ/snarkOS-staging/tree/kp_test_cache_pr_vm2591)).
This revealed that locks are held for a long time (>2s) for deep nested programs.

Example logs:
```
2025-02-13T09:54:21.240609Z TRACE snarkos: [locktick] checking for active lock guards
2025-02-13T09:54:21.240971Z TRACE snarkos: /Users/kp/.cargo/git/checkouts/snarkvm-438da7bfff6ff07c/c12cbf6/ledger/src/advance.rs@93:33 (Write): 304; 1 active; avg d: 2.09742095s; avg w: 58ns
2025-02-13T09:54:21.240999Z TRACE snarkos: /Users/kp/.cargo/git/checkouts/snarkvm-438da7bfff6ff07c/c12cbf6/synthesizer/src/vm/finalize.rs@591:27 (Write): 305; 1 active; avg d: 2.065852041s; avg w: 79ns
2025-02-13T09:54:21.241006Z TRACE snarkos: /Users/kp/.cargo/git/checkouts/snarkvm-438da7bfff6ff07c/c12cbf6/synthesizer/src/vm/finalize.rs@554:28 (Lock): 305; 1 active; avg d: 2.06651786s; avg w: 99ns
2025-02-13T09:54:21.241015Z TRACE snarkos: /Users/kp/.cargo/git/checkouts/snarkvm-438da7bfff6ff07c/c12cbf6/synthesizer/src/vm/mod.rs@321:27 (Lock): 305; 1 active; avg d: 2.097369753s; avg w: 97ns
2025-02-13T09:54:21.241165Z TRACE snarkos: /Users/kp/dev/stress-observability/test_suites/single-region-tests/playbooks/snarkos-shallow/246996bb2cdbc0dc75d4029b21aad1d32db6af2d/node/bft/src/bft.rs@459:38 (Lock): 2590; 1 active; avg d: 6.866325ms; avg w: 201ns
```

Some fix ideas:
* use `try_lock`
* find a minimal reproduction case
* print when we call `contains_program_in_cache` and `get_stack` so we understand where the "pressure" is

raychu86:
How relevant/applicable is this with the new Process + Stack changes?

I assume the optimizations here using caches may still be useful, but there are quite a number of merge conflicts. In addition, there is now the assumption of upgradable stacks due to program upgradability.

vicsn:
Another note for whoever continues this: the introduction of dynamic dispatch may significantly alter the required design

PR #2594, opened 2025-01-14
[Fix] CI for memory-intensive synthesizer tests

niklaslong:
Builds on #2593 and achieves sub 20 minute CI times for synthesiser tests on `2xlarge` instances. Memory-intensive tests are moved to a dedicated multi-threaded job (`synthesizer-mem-heavy`) and the other synthesizer tests can now be run in parallel without OOM'ing. 

I have also commented out the highest power test cases as the parameter download fails for them at the moment. We don't use these in production but it might be possible to restore their testing on CI by:
1. ssh'ing into the machine while the main job sleeps (don't `pkill sleep` or the job will fail)
2. running the tests manually to download the parameters (retries should eventually succeed)
4. making sure the `resources` directory gets cached

### DIFF
--- a/parameters/src/mainnet/mod.rs
+++ b/parameters/src/mainnet/mod.rs
@@ -42,6 +42,10 @@ impl_remote!(Degree25, REMOTE_URL, "resources/", "powers-of-beta-25", "usrs");
 impl_remote!(Degree26, REMOTE_URL, "resources/", "powers-of-beta-26", "usrs");
 impl_remote!(Degree27, REMOTE_URL, "resources/", "powers-of-beta-27", "usrs");
 // TODO (nkls): restore on CI.
+// The SRS is only used for proving and we don't currently support provers of
+// this size. When a users wants to create a proof, they load the appropriate
+// powers for the circuit in `batch_circuit_setup` which calls `max_degree`
+// based on the domain size.
 // impl_remote!(Degree28, REMOTE_URL, "resources/", "powers-of-beta-28", "usrs");
 
 // Shifted Degrees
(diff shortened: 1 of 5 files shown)


vicsn (approved):

parameters/src/mainnet/mod.rs
```diff
@@ -41,7 +41,8 @@ impl_remote!(Degree24, REMOTE_URL, "resources/", "powers-of-beta-24", "usrs");
 impl_remote!(Degree25, REMOTE_URL, "resources/", "powers-of-beta-25", "usrs");
 impl_remote!(Degree26, REMOTE_URL, "resources/", "powers-of-beta-26", "usrs");
 impl_remote!(Degree27, REMOTE_URL, "resources/", "powers-of-beta-27", "usrs");
-impl_remote!(Degree28, REMOTE_URL, "resources/", "powers-of-beta-28", "usrs");
+// TODO (nkls): restore on CI.
+// impl_remote!(Degree28, REMOTE_URL, "resources/", "powers-of-beta-28", "usrs");
```
raychu86, line 49: Can you explain the PR why disabling these powers are safe? Can you show that they are not being used?

niklaslong: I am given to understand (with help from @vicsn—this isn't context I'm fully familiar with, so bear with me) that the SRS is only used for proving and we don't currently support provers of the size that we're disabling. This is all set down in the `algorithms` crate. When a users wants to create a proof, they load the appropriate powers for the circuit in [`batch_circuit_setup`](https://github.com/ProvableHQ/snarkVM/blob/staging/algorithms/src/snark/varuna/varuna.rs#L85) which calls [`max_degree`](https://github.com/ProvableHQ/snarkVM/blob/staging/algorithms/src/snark/varuna/ahp/ahp.rs#L96) based on the domain size.

d0cd: Is this due to the limits on the size of the constraint system? 
If so, it would be nice to have a TODO wherever we specify the size limit to restore this when it is increased.
It's worth noting that we may eventually have to support these in CI.

vicsn: For reference @niklaslong the size limits are `MAX_DEPLOYMENT_CONSTRAINTS` and `MAX_VARIABLE_CONSTRAINTS`.

PR #2595, opened 2025-01-15
[Feat] Increase `MAX_CERTIFICATES` on `MainnetV0`

raychu86:
## Motivation

This PR plans to increase the number `N::MAX_CERTIFICATES` value for `MainnetV0`. The main result of this change is that the maximum number of validators in a committee is also increased. Since Mainnet was launched, the codebase has undergone some extensive upgrades and optimizations and based on internal/external testing, we believe the number of validators can be expanded safely. The original maximum is 16 and this change increases that value to 25 (subject to change). 

Increasing the number of validators creates a more robust and more decentralized network. This increase also has a number of other effects (increasing TPS, increasing network overhead, and increasing block size to name a few), so we must be vigilant when deciding to expand this number. As we further improve the network, we can iteratively increase the maximum number of validators in order to ensure network stability, even if it feels conservative.

## **Migration:**
- Introduce `N::CONSENSUS_V3_HEIGHT` to Network trait 
    - This is the height for each network that snarkVM will swap from the `v2` consensus to `v3` consensus.
- SnarkOS will require nodes to swap to the new version before height `N::CONSENSUS_V3_HEIGHT`, otherwise there may be some issues.

### **Mainnet** - Block 4,900,000 (~Feb 18, 2025 at the current 3.0s block times)

| Network | Release Date | Buffer | Consensus V3 |
| Canary  | Jan 21, 2025 | 3 days | Jan 24, 2025 |
| Testnet | Jan 28, 2025 | 3 days | Jan 31, 2025 |
| Mainnet | Feb 11, 2025  | 7 days | Feb 18, 2025 |

## Test Plan

Tests have been added to ensure that the migration properly transitions the old `MAX_CERTIFICATES_BEFORE_V3` value to the new `MAX_CERTIFICATES` value.

CI can be found [here](https://app.circleci.com/pipelines/github/ProvableHQ/snarkVM?branch=feat%2Fincrease_validators-ci).

## Related PRs

This PR replaces https://github.com/ProvableHQ/snarkVM/pull/2586 due to branching issues.

Migration will coincide with https://github.com/AleoNet/snarkVM/pull/2575.

### DIFF
--- a/console/network/src/canary_v0.rs
+++ b/console/network/src/canary_v0.rs
@@ -136,10 +136,16 @@ impl Network for CanaryV0 {
     /// The block height from which consensus V2 rules apply.
     #[cfg(not(any(test, feature = "test")))]
     const CONSENSUS_V2_HEIGHT: u32 = 2_900_000;
-    // TODO (raychu86): Update this value based on the desired canary height.
     /// The block height from which consensus V2 rules apply.
     #[cfg(any(test, feature = "test"))]
-    const CONSENSUS_V2_HEIGHT: u32 = 0;
+    const CONSENSUS_V2_HEIGHT: u32 = 10;
+    // TODO: (raychu86): Update this value based on the desired mainnet height.
+    // The block height from which consensus V3 rules apply.
+    #[cfg(not(any(test, feature = "test")))]
+    const CONSENSUS_V3_HEIGHT: u32 = 4_560_000;
+    /// The block height from which consensus V3 rules apply.
+    #[cfg(any(test, feature = "test"))]
+    const CONSENSUS_V3_HEIGHT: u32 = 20;
     /// The network edition.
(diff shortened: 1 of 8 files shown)


raychu86:
@kpandl Your comment https://github.com/ProvableHQ/snarkVM/pull/2586#issuecomment-2575025844 should be addressed with the new commit - [36971cf](https://github.com/ProvableHQ/snarkVM/pull/2595/commits/36971cf0543d52a7ace5cc2ac89d73b578313ff4).

Would appreciate a confirmation on your end as well.

vicsn (approved):
LGTM, might want to remove the two todo's

kpandl:
LGTM! Tested the edge case from https://github.com/ProvableHQ/snarkVM/pull/2586#issuecomment-2575025844 again and everything works as it should.

PR #2596, opened 2025-01-16
[Fix] Selection of transition ID in finalize.

d0cd:
## Motivation

The motivating issue for this PR is [here](https://github.com/AleoNet/snarkVM/issues/2570).

The original PR is [here](https://github.com/ProvableHQ/snarkVM/pull/2575).

From the original discussion:
>Fundamentally, while the call graph is constructed correctly, it is used in an unsound way to find the transition ID that corresponds to `Future` that is passed in as an operand to an `await` command ([source](https://github.com/AleoNet/snarkVM/blob/4eb83d7f7276514baa9c44b920155750cfe855e7/synthesizer/process/src/finalize.rs#L269)). The `child_transition_id` linked above is used to initialize the `FinalizeRegisters`. The `transition_id` in `FinalizeRegisters` is only used to initialize an RNG in the `rand` instructions. The purpose is the ensure that on-chain logic (in finalize scopes) have a unique seed each time they are invoked. 

>The current implementation has the following issues:
>1. In the case where an async `function` calls a mix of async and non-async functions, a [non-existent](https://github.com/AleoNet/snarkVM/blob/4eb83d7f7276514baa9c44b920155750cfe855e7/synthesizer/process/src/finalize.rs#L272) transition is used in the call graph. Specifically it must an async function call that itself makes an async function call that precedes a non-async call. This results in the transaction getting rejected. **In this case, the impacted programs are ones that use a specific mix of async and non-async calls.**
>2. In the case where only `async` calls are made, but are awaited in a different order than the call graph, an incorrect transition ID is used for the `FinalizeRegisters`. The only place where the transition ID is used is in the `rand.chacha` instruction. **In this case, the impacted programs are ones that make async calls to programs with `rand.chacha` but await the futures in a different order.**

>There are a number of possible ways this could be addressed:
>1. Update the implementation to get the correct transition ID. This would require more information to be passed in during finalization.
>2. Add a transition ID to the `Future`. This would be a breaking change in the data format, but conceptually sound.
>3. Fix the use of the transition ID to not require the precise sub-transition, while maintaining the invariant that RNG seeds are unique.

This PR proposes a solution to this issue by introducing a nonce to `FinalizeRegisters`. This nonce is used to seed the `rand` commands. Furthermore, instead of attempting to determine the `child_transition_id` that corresponds to an awaited `Future` from the call graph, this approach uses the main transition ID to initialize all `FinalizeRegisters` for a given transaction. The main transition ID along with the nonce ensures that each finalize context uses a unique seed for the RNG, while removing the need to correctly determine the transition ID for a given `Future` (a complicated process).

## Migration

This proposal has been written to migrate at `N::CONSENSUS_V2_HEIGHT`.
Given timelines, it's more likely that a new `N::CONSENSUS_V3_HEIGHT` would need to be introduced. The migration would follow the process introduced in [ARC-0042](https://github.com/AleoNet/snarkVM/pull/2569).

## Test Plan

This PR includes a [test](https://github.com/ProvableHQ/snarkVM/blob/fix/rng-finalize/synthesizer/tests/tests/vm/execute_and_finalize/complex_finalization_2.aleo), whose [expected output](https://github.com/ProvableHQ/snarkVM/blob/fix/rng-finalize/synthesizer/tests/expectations/vm/execute_and_finalize/complex_finalization_2.out) demonstrates the the failure and the fix after `CONSENSUS_V2_HEIGHT` is reached.

Included is the [CI branch](https://github.com/ProvableHQ/snarkVM/tree/fix/rng-finalize-ci) and the [CI diff](https://github.com/ProvableHQ/snarkVM/compare/fix/rng-finalize...fix/rng-finalize-ci).

Note that the `test_rand` integration test expectation changed because the program integration tests now use the `test` feature which has different block metadata, thus changing the blockhash, thus changing the on-chain RNG seed.

## Impact

As stated above, there are two classes of programs that are impacted by this issue:
1. those that use a mix of async and non-async calls in a specific way.
2. that make async calls to programs with `rand.chacha` but await the futures in a different order that the async functions were called.

In scanning all Aleo programs deployed on Mainnet as of **2/11/25 8AM PT**:
1. There are 13 functions among all programs deployed to mainnet, which contain a non-async call, followed by an async call. In the above functions, none of the async calls take a `Future` as input, thus do not exhibit this issue.

2. There are 28 programs that use the `rand.chacha` command and 0 programs that import, and consequently, call them.

**This was confirmed by a static analyzer on all programs on mainnet at the above date and time** and double-checked by manual audit.

## Related PRs
- https://github.com/ProvableHQ/leo/pull/28443

## To Reviewers

An important function of this PR is to provide an understanding of how async execution mechanism works and why this PR is needed. If reviewers need context or clarification, please feel to post your questions in this thread. I am also happy to do a call explaining the original and proposed design/code.


ljedrz (commented):
LGTM engineering-wise

vicsn (approved):

raychu86 (approved):

PR #2230, opened 2023-12-04
Use evaluate_function for CallStack::PackageRun; Sample response for CallStack::CheckDeployment

vicsn:
## Motivation

If we `evaluate_function` during deployment, we risk hitting false positive failed constraints on our random assignment.

So instead, this PR proposes to sample a response. After sampling, construction of Response is handled the same way evaluate_function.

## Test Plan

In the future we will test a wider range of programs.

## Related PRs

https://github.com/AleoHQ/snarkVM/pull/2176
https://github.com/AleoHQ/snarkVM/pull/2224
https://github.com/AleoHQ/snarkVM/pull/2226

PR #2598, opened 2025-01-17
Refactor CallStack::Synthesize to produce consistent dummy outputs for faster nested import deployments

kpandl:
## Summary

This PR refactors the way `CallStack::Synthesize` is run, decoupling the `Synthesize` mode from the old `execute_function` approach. It produces consistent dummy outputs (inspired by `CheckDeployment`) more rapidly and skips sub-circuit construction for nested import calls. The result is significantly faster deployments for deeply nested imports, while still generating cryptographically coherent record nonces.

## Motivation

Previously, `Authorize` and `Synthesize` modes were lumped together. As such, calls in CallStack::Synthesize were forced to run a real sub-circuit via `execute_function`. This is slow for nested imports and has led to an exponential growth in deployment time w.r.t. the import depth.

Thus, this PR refactors the `Synthesize` branch to skip full circuit execution and produce “dummy” outputs with consistent record nonces.

## Security

This PR is not meant to impact *any* validator operations, because the `CallStack::Synthesize` is only used by provers, not by validators/verifiers.

## Note

This PR changes expected future IDs in 3 of the `execute_and_finalize` tests. The reason is that previously, Synthesize always ran real sub-circuits, consuming RNG draws. With this PR, certain sub-circuits are skipped by producing dummy outputs. As a result, the RNG is executed less times, leading to different transaction commitments and future IDs.

## Test Plan

Tested the deployment and execution of programs with deep imports in a local devnet.

## Related PRs

* https://github.com/ProvableHQ/snarkVM/pull/2176
* https://github.com/ProvableHQ/snarkVM/pull/2230


vicsn:
NOTE: this PR will also significantly speed up tests using nested deployments, e.g. `test_deep_nested_execution_cost`.

synthesizer/process/src/stack/call/mod.rs
```diff
@@ -45,7 +45,7 @@ pub trait CallTrait<N: Network> {
     /// Executes the instruction.
     fn execute<A: circuit::Aleo<Network = N>, R: CryptoRng + Rng>(
         &self,
-        stack: &(impl StackEvaluate<N> + StackExecute<N> + StackMatches<N> + StackProgram<N>),
+        stack: &Stack<N>,
```
vicsn: Why change the interface? Can we keep it the way it was? Same for the other function.

synthesizer/process/src/stack/mod.rs
```diff
@@ -316,6 +316,12 @@ impl<N: Network> StackProgram<N> for Stack<N> {
             .ok_or_else(|| anyhow!("Function '{function_name}' does not exist"))
     }
 
+    /// Returns true if the proving key for the given function name already exists in this stack.
+    #[inline]
+    fn has_proving_key(&self, function_name: &Identifier<N>) -> bool {
```
vicsn: Why add this extra wrapper?

synthesizer/process/src/stack/call/mod.rs
```diff
@@ -256,6 +257,98 @@ impl<N: Network> CallTrait<N> for Call<N> {
...
+                        // Push the request onto the call stack.
+                        call_stack.push(request.clone())?;
+
+                        // Execute the request.
+                        let response = substack.execute_function::<A, R>(call_stack, console_caller, root_tvk, rng)?;
+
+                        // Return the request and response.
+                        (request, response)
+                    }
+                    CallStack::Synthesize(_, private_key, ..) => {
```
vicsn: Missing comment like you added to the other synthesize and authorize cases

synthesizer/process/src/stack/call/mod.rs
```diff
@@ -256,6 +257,98 @@ impl<N: Network> CallTrait<N> for Call<N> {
...
+                                ValueType::Record(record_name) => {
+                                    let index = match output.operand() {
+                                        Operand::Register(Register::Locator(index)) => Field::from_u64(*index),
+                                        _ => bail!("Expected a `Register::Locator` operand for a record output."),
+                                    };
+                                    // Compute the randomizer.
+                                    let randomizer = N::hash_to_scalar_psd2(&[*request.tvk(), index])?;
+                                    // Construct the record nonce from that randomizer.
+                                    let record_nonce = N::g_scalar_multiply(&randomizer);
+                                    // Sample the record with that nonce.
```
vicsn: Could this be made a part of `sample_record`, so that callers of `sample_record` don't have to think about it anymore and all are guaranteed correct usage?

synthesizer/process/src/stack/call/mod.rs
vicsn: Is there any difference to the contents of `CallStack::Synthesize` and `CallStack::CheckDeployment`? Perhaps they can be merged into the same case.

vicsn (changes requested):

synthesizer/process/src/stack/call/mod.rs
kpandl: Addressed here: https://github.com/ProvableHQ/snarkVM/pull/2598/commits/774efae04d38e1688477ecefa5d389e02eb5a69e

synthesizer/process/src/stack/call/mod.rs
kpandl: Addressed both here: https://github.com/ProvableHQ/snarkVM/pull/2598/commits/815a4dfeec0f8aae47b3b04ee7b2f61f85fa10d2
`Synthesize` and `CheckDeployment` are indeed the same!

synthesizer/process/src/stack/mod.rs
kpandl: I think if we don't want to have this wrapper, I need to revert https://github.com/ProvableHQ/snarkVM/commit/774efae04d38e1688477ecefa5d389e02eb5a69e.
I guess the cleanest solution is if I go ahead and to that, so we can directly call `contains_proving_key`?

synthesizer/process/src/stack/call/mod.rs
kpandl: Created a new function `sample_record_using_tvk` here: https://github.com/ProvableHQ/snarkVM/pull/2598/commits/0503e5cd906617c121aec2cac97b981442a312ed

synthesizer/process/src/stack/mod.rs
vicsn: Ah I see. Hmm I don't know the reasoning behind the choice of `StackProgram<N>` trait existing. But let's try to keep that unchanged.

One approach which could be somewhat consistent, is to create a new trait `StackKeys`. And these particular functions (contains/get/insert proving/verifying keys) could `impl<N> StackKeys<N>`.

synthesizer/process/src/stack/call/mod.rs
```diff
@@ -265,22 +267,24 @@ impl<N: Network> CallTrait<N> for Call<N> {
                             inputs.iter(),
                             &function.input_types(),
                             root_tvk,
-                            is_root,
+                            false,
```
vicsn: Why use `false`?

synthesizer/process/src/stack/call/mod.rs
```diff
@@ -289,38 +293,38 @@ impl<N: Network> CallTrait<N> for Call<N> {
...
```
vicsn: Why use `false`?

kpandl: addressed here: https://github.com/ProvableHQ/snarkVM/pull/2598/commits/d7070d9af8d06f488d4841d52c8e835d104f6d48

synthesizer/process/src/stack/call/mod.rs
kpandl: addressed here: https://github.com/ProvableHQ/snarkVM/pull/2598/commits/d7070d9af8d06f488d4841d52c8e835d104f6d48

synthesizer/process/src/stack/mod.rs
kpandl: addressed here: https://github.com/ProvableHQ/snarkVM/pull/2598/commits/6ebe64296cddc7bfdd88b4e2f8b267587ceeab3a

kpandl:
note: the failing `snarkvm` CI test is a CI/CDN issue

raychu86:
@d0cd This may be relevant to some of your `Stack` related work

synthesizer/process/src/stack/mod.rs
```diff
@@ -223,6 +223,82 @@ impl<N: Network> Stack<N> {
     }
 }
 
+impl<N: Network> StackKeys<N> for Stack<N> {
+    /// Returns `true` if the proving key for the given function name exists.
+    #[inline]
+    fn contains_proving_key(&self, function_name: &Identifier<N>) -> bool {
```
vicsn, line 229: Now that these functions are defined in `impl<N: Network> StackKeys<N> for Stack<N>`, can they be removed in `impl<N: Network> Stack<N>`?

kpandl: Good catch, addressed here: https://github.com/ProvableHQ/snarkVM/pull/2598/commits/46c8498d673fc1a90527b4a405822b750e8daeca

vicsn (approved):
LGTM, CI flakiness is unrelated

d0cd (commented):
Sorry for the delay here folks! Great find and thank you for doing the tedious exercise of stepping into the RNG internals. PR looks good! 

If you'd like to dedup some code, you could also combine the match arms for `PackageRun` and `Synthesize if pk_missing` by adding `pk_missing` to the match expression. It's really only a style thing and can be addressed at any time.

PR #2600, opened 2025-01-21
perf: Change link-time optimization in dev and test

Raphexion:
## Motivation

This change improves the development cycle in both dev and test, when running the same command multiple times in a row.

On my machine I can see improvements from minutes to seconds. For example when running `cargo run` and `cargo test`.

## Test Plan

Compare running a commands 3 times in a row with or without this patch.

## Related PRs

None

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -292,7 +292,7 @@ debug-assertions = false
 
 [profile.dev]
 opt-level = 3
-lto = "thin"
+lto = "off"
 incremental = true
 
 [profile.test]


vicsn:
"This branch is out-of-date with the base branch"

Raphexion:
@vicsn Sorry about that, I had forgotten to set the upstream repo. Now it is onto of 7c4a910f7 `staging`

vicsn:
Regarding [Lukasz' comment](https://github.com/ProvableHQ/snarkOS/pull/3441#pullrequestreview-2446031585).

Seems like the snarkVM CircleCI tests do run slower, 30 minutes on this PR instead of 20 or 22 minutes on recent PRs. This difference will matter a lot when we need to push a hotfix through.

Can you remove the link-time optimization for the test profile? We can keep it as the default on the dev profile. :)

Also from this and a couple other PRs it seems that the `ledger-puzzle-epoch` job is flaky. Can you upgrade its resource class in `.circleci/config.yml` to `medium`?

vicsn (approved):

PR #2601, opened 2025-01-22
[Feature] Batch proposal spend limits

niklaslong:
This PR introduces a `BATCH_SPEND_LIMIT` which is currently set to `100 credits * tx limit / 20`  and continues the exploratory work done in https://github.com/ProvableHQ/snarkVM/pull/2565. Under the assumption that 100 credits equals 1s of single-threaded runtime and given that a block contains `NUM_MAX_CERTIFICATES`,  the upper bound should be around 1 min for blocks.  The chosen factor of 20 will likely need to be modified when priority fees are introduced and when the validator set grows but for now it leads to a reasonable upper bound on block time.

### DIFF
--- a/console/network/src/lib.rs
+++ b/console/network/src/lib.rs
@@ -131,7 +131,7 @@ pub trait Network:
     const MAX_DEPLOYMENT_CONSTRAINTS: u64 = 1 << 20; // 1,048,576 constraints
     /// The maximum number of microcredits that can be spent as a fee.
     const MAX_FEE: u64 = 1_000_000_000_000_000;
-    /// The maximum number of microcredits that can be spent on a finalize block.
+    /// The maximum number of microcredits that can be spent on a transaction's finalize scope.
     const TRANSACTION_SPEND_LIMIT: u64 = 100_000_000;
     /// The maximum number of microcredits that can be spent by the transactions in a batch.
     /// This implies the block spend limit is bounded at `TRANSACTION_SPEND_LIMIT * N::NUM_MAX_CERTIFICATES`.
--- a/console/network/src/lib.rs
+++ b/console/network/src/lib.rs
@@ -133,7 +133,7 @@ pub trait Network:
     const MAX_FEE: u64 = 1_000_000_000_000_000;
     /// The maximum number of microcredits that can be spent on a transaction's finalize scope.
     const TRANSACTION_SPEND_LIMIT: u64 = 100_000_000;
-    /// The maximum number of microcredits that can be spent by the transactions in a batch.
+    /// The maximum number of microcredits that can be spent on compute by the transactions in a batch.
     /// This implies the block spend limit is bounded at `TRANSACTION_SPEND_LIMIT * N::NUM_MAX_CERTIFICATES`.
     const BATCH_SPEND_LIMIT: u64 = Self::TRANSACTION_SPEND_LIMIT;
 
--- a/console/network/src/lib.rs
+++ b/console/network/src/lib.rs
@@ -76,6 +76,7 @@ pub enum ConsensusVersion {
     V1 = 1,
     V2 = 2,
     V3 = 3,
+    V4 = 4,
 }
 
 pub trait Network:
@@ -223,7 +224,7 @@ pub trait Network:
 
     /// A list of (consensus_version, block_height) pairs indicating when each consensus version takes effect.
     /// Documentation for what is changed at each version can be found in `N::CONSENSUS_VERSION`
-    const CONSENSUS_VERSION_HEIGHTS: [(ConsensusVersion, u32); 3];
+    const CONSENSUS_VERSION_HEIGHTS: [(ConsensusVersion, u32); 4];
     ///  A list of (consensus_version, size) pairs indicating the maximum number of validators in a committee.
     //  Note: This value must **not** decrease without considering the impact on serialization.
     //  Decreasing this value will break backwards compatibility of serialization without explicit
--- a/console/network/src/lib.rs
+++ b/console/network/src/lib.rs
@@ -132,6 +132,9 @@ pub trait Network:
     const MAX_FEE: u64 = 1_000_000_000_000_000;
     /// The maximum number of microcredits that can be spent on a finalize block.
     const TRANSACTION_SPEND_LIMIT: u64 = 100_000_000;
+    /// The base cost in microcredits to verify an execution.
+    /// NOTE: this constant reflects the compute cost of an execution, but is not required to be paid by the user.
+    const EXECUTION_BASE_COST: u64 = 2_000_000; // 2 million microcredits.
 
     /// The anchor height, defined as the expected number of blocks to reach the coinbase target.
     const ANCHOR_HEIGHT: u32 = Self::ANCHOR_TIME as u32 / Self::BLOCK_TIME as u32;
--- a/console/network/src/lib.rs
+++ b/console/network/src/lib.rs
@@ -132,6 +132,9 @@ pub trait Network:
     const MAX_FEE: u64 = 1_000_000_000_000_000;
     /// The maximum number of microcredits that can be spent on a finalize block.
     const TRANSACTION_SPEND_LIMIT: u64 = 100_000_000;
+    /// The maximum number of microcredits that can be spent by the transactions in a batch.
+    /// This implies the block spend limit is bounded at `TRANSACTION_SPEND_LIMIT * N::NUM_MAX_CERTIFICATES`.
+    const BATCH_SPEND_LIMIT: u64 = Self::TRANSACTION_SPEND_LIMIT;
     /// The base cost in microcredits to verify an execution.
     /// NOTE: this constant reflects the compute cost of an execution, but is not required to be paid by the user.
     const EXECUTION_BASE_COST: u64 = 2_000_000; // 2 million microcredits.
--- a/console/network/src/lib.rs
+++ b/console/network/src/lib.rs
@@ -136,9 +136,6 @@ pub trait Network:
     /// The maximum number of microcredits that can be spent by the transactions in a batch.
     /// This implies the block spend limit is bounded at `TRANSACTION_SPEND_LIMIT * N::NUM_MAX_CERTIFICATES`.
     const BATCH_SPEND_LIMIT: u64 = Self::TRANSACTION_SPEND_LIMIT;
-    /// The base cost in microcredits to verify an execution.
-    /// NOTE: this constant reflects the compute cost of an execution, but is not required to be paid by the user.
-    const EXECUTION_BASE_COST: u64 = 2_000_000; // 2 million microcredits.
 
     /// The anchor height, defined as the expected number of blocks to reach the coinbase target.
     const ANCHOR_HEIGHT: u32 = Self::ANCHOR_TIME as u32 / Self::BLOCK_TIME as u32;
--- a/console/network/src/lib.rs
+++ b/console/network/src/lib.rs
@@ -133,9 +133,6 @@ pub trait Network:
     const MAX_FEE: u64 = 1_000_000_000_000_000;
     /// The maximum number of microcredits that can be spent on a transaction's finalize scope.
     const TRANSACTION_SPEND_LIMIT: u64 = 100_000_000;
-    /// The maximum number of microcredits that can be spent on compute by the transactions in a batch.
-    /// This implies the block spend limit is bounded at `TRANSACTION_SPEND_LIMIT * N::NUM_MAX_CERTIFICATES`.
-    const BATCH_SPEND_LIMIT: u64 = Self::TRANSACTION_SPEND_LIMIT;
 
     /// The anchor height, defined as the expected number of blocks to reach the coinbase target.
     const ANCHOR_HEIGHT: u32 = Self::ANCHOR_TIME as u32 / Self::BLOCK_TIME as u32;
--- a/console/network/src/testnet_v0.rs
+++ b/console/network/src/testnet_v0.rs
@@ -140,7 +140,7 @@ impl Network for TestnetV0 {
         (ConsensusVersion::V1, 0),
         (ConsensusVersion::V2, 2_950_000),
         (ConsensusVersion::V3, 4_800_000),
-        (ConsensusVersion::V4, 6_625_000),
+        (ConsensusVersion::V4, 6_635_000),
     ];
     /// A list of (consensus_version, block_height) pairs indicating when each consensus version takes effect.
     /// Documentation for what is changed at each version can be found in `N::CONSENSUS_VERSION`
--- a/console/network/src/testnet_v0.rs
+++ b/console/network/src/testnet_v0.rs
@@ -140,7 +140,7 @@ impl Network for TestnetV0 {
         (ConsensusVersion::V1, 0),
         (ConsensusVersion::V2, 2_950_000),
         (ConsensusVersion::V3, 4_800_000),
-        (ConsensusVersion::V4, 5_000_000),
+        (ConsensusVersion::V4, 6_625_000),
     ];
     /// A list of (consensus_version, block_height) pairs indicating when each consensus version takes effect.
     /// Documentation for what is changed at each version can be found in `N::CONSENSUS_VERSION`
--- a/console/network/src/testnet_v0.rs
+++ b/console/network/src/testnet_v0.rs
@@ -136,13 +136,17 @@ impl Network for TestnetV0 {
     /// A list of (consensus_version, block_height) pairs indicating when each consensus version takes effect.
     /// Documentation for what is changed at each version can be found in `N::CONSENSUS_VERSION`
     #[cfg(not(any(test, feature = "test")))]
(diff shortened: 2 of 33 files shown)


console/network/src/lib.rs
```diff
@@ -132,6 +133,12 @@ pub trait Network:
     const MAX_FEE: u64 = 1_000_000_000_000_000;
     /// The maximum number of microcredits that can be spent on a finalize block.
     const TRANSACTION_SPEND_LIMIT: u64 = 100_000_000;
+    /// The maximum number of microcredits that can be spent by the transactions in a batch.
+    /// This implies the block spend limit is bounded at `TRANSACTION_SPEND_LIMIT * N::NUM_MAX_CERTIFICATES`.
+    const BATCH_SPEND_LIMIT: u64 = Self::TRANSACTION_SPEND_LIMIT;
+    /// The base cost in microcredits to verify an execution.
+    /// NOTE: this constant reflects the compute cost of an execution, but is not required to be paid by the user.
+    const EXECUTION_BASE_COST: u64 = 2_000_000; // 2 million microcredits.
```
raychu86: "is not required to be paid by the user" What does this mean? 

A `transfer_public` is currently around 51k microcredits, why is there a need for the additional 2 million microcredits?

raychu86: Is this to arbitrarily say that a proposal is upper bounded by `BATCH_SPEND_LIMIT` / `EXECUTION_BASE_COST` number of executions?

vicsn: Executions do incur compute, we just don't price it in at the moment. So it makes sense to take it into account when determining the spend limit - or better wording: 'compute limit'.

Right now it doesn't matter too much because we have a batch proposal size limit, we could leave it it.

However, in the future we may want to consider getting rid of the batch proposal size limit, and only keep the compute limit, at which point we better not forget that proof verification takes up a significant amount of compute.

I agree 2 credits seems like a lot - but note that single-threaded proof verification costs multiple milliseconds. I think in any case this can likely be lowered to 1 or 0.5 for now.

raychu86: It seems like `cost` is an overloaded term here, as it implies it's used for the execution fees paid by the user.

In theory, the execution fee should already price in the cost of compute. An execution fee is the cost of finalize + storage cost; the storage cost can be considered the base level of payment for storing and performing the verification on the transaction. 

It feels confusing that a constant is tacked on here to limit the proposal spend by misrepresenting and inflating the "cost" of each transaction.

console/network/src/lib.rs
```diff
@@ -132,6 +133,12 @@ pub trait Network:
     const MAX_FEE: u64 = 1_000_000_000_000_000;
     /// The maximum number of microcredits that can be spent on a finalize block.
```
raychu86: nit
```suggestion
    /// The maximum number of microcredits that can be spent on a transaction's finalize scope.
```

console/network/src/lib.rs
vicsn: I agree it is confusing. Just for the record and for anyone reading this: execution (snark proof) verification and its compute is currently not priced in.

However, I realized yesterday that actually execution verification is no longer on the hot path, due to this PR merged this week: https://github.com/ProvableHQ/snarkOS/pull/3451

So thank you for flagging, I think we can remove this "virtual proof verification cost". The spend limit's only aim is to limit the amount of compute in the block generation hot path.

niklaslong: Removed in [ed5b491](https://github.com/ProvableHQ/snarkVM/pull/2601/commits/ed5b491c2018b2b4ee4e2f55ded088e72df4f088).

console/network/src/lib.rs
niklaslong: [e645d1d](https://github.com/ProvableHQ/snarkVM/pull/2601/commits/e645d1da2a7cb5827acb5b99d23bcf28c1fd9e05)

niklaslong:
Just as a reminder, I've opened the PR for review but we'll need to make sure the heights for the consensus version changes are still appropriate before we merge.

synthesizer/process/src/cost.rs
```diff
@@ -460,6 +465,29 @@ pub fn cost_in_microcredits_v1<N: Network>(stack: &Stack<N>, function_name: &Ide
         })
 }
 
+/// Returns the compute cost for a transaction in microcredits.
+/// This is used to limit the amount of compute in the block generation hot
+/// path. This does NOT represent the full costs which a user has to pay.
+pub fn compute_cost<N: Network>(process: &Process<N>, transaction: &Transaction<N>) -> Result<u64> {
```
vicsn: Could we call `compute_cost` from `deployment_cost` and `execution_cost_v*`?

vicsn (changes requested):
Currently planned release dates for this consensus version are:

Canary: 4/1/2025
Testnet: 4/8/2025
Mainnet: 4/22/2025

Can you @niklaslong populate the version heights based on extrapolation of the existing blocktimes and block heights for the three networks, adding a similar 3-/5-/14- day leeway as for the previous upgrade? And double check around 3/30/2025 whether the heights are still accurate or need a minor update.

IMO we can then merge this PR, so that other PRs can also target this new consensus version. Happy to hear reviewer's thoughts.

console/network/src/lib.rs
```diff
@@ -130,8 +131,11 @@ pub trait Network:
     const MAX_DEPLOYMENT_CONSTRAINTS: u64 = 1 << 20; // 1,048,576 constraints
     /// The maximum number of microcredits that can be spent as a fee.
     const MAX_FEE: u64 = 1_000_000_000_000_000;
-    /// The maximum number of microcredits that can be spent on a finalize block.
+    /// The maximum number of microcredits that can be spent on a transaction's finalize scope.
     const TRANSACTION_SPEND_LIMIT: u64 = 100_000_000;
+    /// The maximum number of microcredits that can be spent by the transactions in a batch.
```
vicsn:
```suggestion
    /// The maximum number of microcredits that can be spent on compute by the transactions in a batch.
```

console/network/src/lib.rs
```diff
@@ -130,8 +131,11 @@ pub trait Network:
...
+    /// This implies the block spend limit is bounded at `TRANSACTION_SPEND_LIMIT * N::NUM_MAX_CERTIFICATES`.
+    const BATCH_SPEND_LIMIT: u64 = Self::TRANSACTION_SPEND_LIMIT;
```
vicsn: What do you think about adding a check in `fn compute_cost` that we don't surpass the `TRANSACTION_SPEND_LIMIT`? I see two advantages:
- With one check we will cover both the finalize ánd synthesis cost not getting too big - which will be very useful since we're planning to increase the constraint limit soon.
- Any "consumers" of `deployment_cost`/`execution_cost`/`compute_cost` such as `snarkos developer execute` or transaction cannons will have to explicitly handle or show an error when surpassing the limit.

console/network/src/testnet_v0.rs
```diff
@@ -136,13 +136,17 @@ impl Network for TestnetV0 {
     /// A list of (consensus_version, block_height) pairs indicating when each consensus version takes effect.
     /// Documentation for what is changed at each version can be found in `N::CONSENSUS_VERSION`
     #[cfg(not(any(test, feature = "test")))]
-    const CONSENSUS_VERSION_HEIGHTS: [(ConsensusVersion, u32); 3] =
-        [(ConsensusVersion::V1, 0), (ConsensusVersion::V2, 2_950_000), (ConsensusVersion::V3, 4_800_000)];
+    const CONSENSUS_VERSION_HEIGHTS: [(ConsensusVersion, u32); 4] = [
+        (ConsensusVersion::V1, 0),
+        (ConsensusVersion::V2, 2_950_000),
+        (ConsensusVersion::V3, 4_800_000),
+        (ConsensusVersion::V4, 5_000_000),
```
raychu86: Note: These values will need to be updated once the code logic is complete

synthesizer/process/src/cost.rs
niklaslong: Afaict, neither of those functions have access to all the inputs required for `compute_cost`? While I'm all for code reuse, I'm not sure the extra indirection and type-wrangling would be worth it here. But maybe I'm missing something, happy to take a closer look 🕵️‍♂️

vicsn:
```
pub fn deployment_cost<N: Network>(deployment: &Deployment<N>)
pub fn execution_cost_v2<N: Network>(process: &Process<N>, execution: &Execution<N>)
```

You can create a Transaction from the Deployment or Execution. If that's too ugly we could have a separate `deployment_compute_cost` and `execute_compute_cost`.

console/network/src/lib.rs
vicsn: @niklaslong on second thought - we should take into account that the constant may be dependent on blockheight in the future... Maybe its easier to keep the check outside of the function anyway... Up to you.

niklaslong: Imo, decoupling the check from the call is a good practice as we might enforce limits differently in future but we could make sure such a check exists higher up for e.g. `snarkos developer execute` instead? This would avoid burying the check and would make it more explicit. After all, these are low level APIs, is there any calling code you think might bypass these checks if they're not implemented here?

vicsn: > but we could make sure such a check exists higher up for e.g. snarkos developer execute instead

Yes that would be great to add to the existing snarkOS PR.

@iamalwaysuncomfortable do you know what functions are likely to be called by e.g. a wallet using the SDK? Should we leave it up to the wallet to be aware of enforce the spend limit (because we expect those developers to monitor the release changelog)?

synthesizer/process/src/cost.rs
niklaslong: Yeah, it's the extra fields (various ids, the fee) that make this cumbersome but I think the second option would be fine 👍

console/network/src/lib.rs
niklaslong: [e39f3fa](https://github.com/ProvableHQ/snarkVM/pull/2601/commits/e39f3fa79807aa085a166aa6b9cd31579ef43d5b)

console/network/src/testnet_v0.rs
niklaslong: Updated in [c931eae](https://github.com/ProvableHQ/snarkVM/pull/2601/commits/c931eae9df1920606a1fcfd7c7d1a9cc0b7aed03) and will double-check on closer to the activation dates 👍

synthesizer/process/src/cost.rs
niklaslong: I added `execution_compute_cost` in [f3393c0](https://github.com/ProvableHQ/snarkVM/pull/2601/commits/f3393c088f9e2506935b3aed6e22a0417253011b), hopefully this is what you had in mind? There was already a `deployment_synthesis_cost` so I left that one as is for now but happy to align naming if that would be clearer.

vicsn (approved):
LGTM!

niklaslong:
CI seems to be having a moment, will retry tomorrow morning 🕵️‍♂️

raychu86 (dismissed):
Logic here looks innocent enough. The bulk/scrutiny will be in the snarkOS side, where enforcement is done.

Also, once block heights are decided, please update the PR description to match the previous migration PRs with information regarding block heights and approximate dates of migration.

synthesizer/process/src/cost.rs
```diff
@@ -19,9 +19,23 @@ use console::{
...
+use ledger_block::{Deployment, Execution, Transaction};
 use synthesizer_program::{CastType, Command, Finalize, Instruction, Operand, StackProgram};
 
+/// Returns the compute cost for a transaction in microcredits.
+/// This is used to limit the amount of compute in the block generation hot
+/// path. This does NOT represent the full costs which a user has to pay.
+pub fn compute_cost<N: Network>(process: &Process<N>, transaction: &Transaction<N>) -> Result<u64> {
+    match transaction {
+        // Synthesis cost accounts for the majority of deployment transaction compute.
+        Transaction::Deploy(_, _, _, deployment, _) => deployment_synthesis_cost(deployment),
```
raychu86: I realized now that `TRANSACTION_SPEND_LIMIT` was based on a maximum spent on finalize, but now we are also considering for `deployment_synthesis_cost`.

Can you estimate how this affects the bounds? How much would a maximum contraint size deployment fill up? How many deployments can we expect in a proposal, etc.

raychu86: In that vein, it would be good to add metrics like how many "transfer_publics" could fill a proposal. (Obviously, this is prior to fee reductions).

vicsn: Might be a good idea to add a test ensuring that the synthesis_cost < batch_spend_limit - as a precaution when we increase the synthesis limit in the future.

It wouldnt be a safety issue necessarily if in a future update synthesis_cost > batch_spend_limit, but it would be good to be aware of.

For executions the test is technically not necessary because the batch_spend_limit literally equals the max_finalize_cost

niklaslong: I've added a test case, hopefully this is what you had in mind 🤞

synthesizer/process/src/cost.rs
```diff
@@ -575,4 +575,15 @@ function over_five_thousand:
...
     }
+
+    fn max_synthesis_cost<N: Network>() -> u64 {
+        N::MAX_DEPLOYMENT_VARIABLES.saturating_add(N::MAX_DEPLOYMENT_CONSTRAINTS) * N::SYNTHESIS_FEE_MULTIPLIER
+    }
+
+    #[test]
+    fn test_max_synthesis_cost_below_batch_spend_limit() {
+        assert!(max_synthesis_cost::<MainnetV0>() < CanaryV0::BATCH_SPEND_LIMIT);
+        assert!(max_synthesis_cost::<MainnetV0>() < TestnetV0::BATCH_SPEND_LIMIT);
```
vicsn: Consider using `TestnetV0` and `CanaryV0` above :)

niklaslong: oop, good catch!

niklaslong: fixed

synthesizer/process/src/cost.rs
vicsn: > it would be good to add metrics like how many "transfer_publics" could fill a proposal.

3000 transfer publics fit within the spend limit. So potentially the proposal limit of 50 can be greatly increased in the future, will need to be tested as at that scale bandwidth and memory usage might explode.

raychu86: How much of the proposal spend limit does a max size deployment take up?

A few programs took almost 50 credits to deploy, which makes me believe a single deployment could potentially take up a majority of a proposal. Not a bad thing, but good think to know

niklaslong: `BATCH_SPEND_LIMIT / max_synthesis_cost ≈ 1.9` in the test I added, this is without the storage cost so it would currently take up most of the proposal (as we don't pack the remaining batch/block space).

console/network/src/lib.rs
iamalwaysuncomfortable: One might assume it'd be `transfer_public` but [actual usage stats](https://explorer.provable.com/programs) show that increasingly functions with complicated call chains are being called with outputs that call multiple futures. I however highly doubt you'll see a 100 credit execution. I'd imagine you're more likely to run into this with a deployment.

I'd echo Niklas's comment that the check is better separated from the call - so the wallets should enforce it. Estimation functions should give true information on what any given transaction would cost so in the rare case anyone does hit a limit, they can make changes and re-estimate to get it below the limit.

niklaslong:
Rebased.

niklaslong:
(Re)-drafted pending more discussion on the fee structure.

ledger/narwhal/batch-header/src/lib.rs
```diff
@@ -56,6 +56,9 @@ pub struct BatchHeader<N: Network> {
 }
 
 impl<N: Network> BatchHeader<N> {
+    /// The maximum number of microcredits that can be spent on compute by the transactions in a batch.
+    /// This implies the block spend limit is bounded at `TRANSACTION_SPEND_LIMIT * N::NUM_MAX_CERTIFICATES`.
+    pub const BATCH_SPEND_LIMIT: u64 = N::TRANSACTION_SPEND_LIMIT * Self::MAX_TRANSMISSIONS_PER_BATCH as u64;
```
niklaslong: Note: this would be identical to the current theoretical limit (and not an effective bound on block time) so we'll likely need this to be lower.

raychu86: If this is the limit we want to establish for now, there is no need for the migration logic above because there is fundamentally no change in possible bounds.

Optionally, for a more holistic approach, future change to this limit and migration can be introduced together with the future Fee reduction efforts. This would simplify the logic in the PR and the sister snarkOS PR to establish the logic behind enforcing the limits.

niklaslong:
Note: I have updated the branch and the PR description. Now that #2657 has been merged, the diff is much smaller.

PR #2603, opened 2025-01-27
Further reduce test parameter downloads to reduce CI flakiness

vicsn:
## Motivation

A parameter test still failed recently: https://app.circleci.com/pipelines/github/ProvableHQ/snarkVM/13487/workflows/4c90c4b7-c758-4774-a323-6dd68c421a7f/jobs/597293?utm_campaign=vcs-integration-link&utm_medium=referral&utm_source=github-checks-link&utm_content=summary

Further reducing the downloads, as already proposed by @raychu86 in the past

## Test Plan

Existing tests should suffice.

## Related PRs

The previous reduction in the tests was implemented in https://github.com/ProvableHQ/snarkVM/pull/2594

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -669,7 +669,7 @@ jobs:
   parameters-large:
     docker:
       - image: cimg/rust:1.81.0 # Attention - Change the MSRV in Cargo.toml and rust-toolchain as well
-    resource_class: << pipeline.parameters.small >>
+    resource_class: << pipeline.parameters.medium >>
     steps:
       - run_serial:
           flags: --no-run --features large_params
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -666,6 +666,16 @@ jobs:
           workspace_member: parameters
           cache_key: v1.0.0-rust-1.81.0-snarkvm-parameters-cache
 
+  parameters_large:
+    docker:
+      - image: cimg/rust:1.81.0 # Attention - Change the MSRV in Cargo.toml and rust-toolchain as well
+    resource_class: << pipeline.parameters.small >>
+    steps:
+      - run_serial:
+          flags: --no-run --features large_params
+          workspace_member: parameters
+          cache_key: v1.0.0-rust-1.81.0-snarkvm-parameters-cache
+
   parameters-uncached:
     docker:
       - image: cimg/rust:1.81.0 # Attention - Change the MSRV in Cargo.toml and rust-toolchain as well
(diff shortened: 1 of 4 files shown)


raychu86:
Is this really the correct approach? I think at least we need a unit test to prove that these SRS can't be used based on the protocol limits.

Alternatively, we can disable the CI, but still allow the downloads of these SRS. (This might require more LoC changed)

vicsn:
> Is this really the correct approach? I think at least we need a unit test to prove that these SRS can't be used based on the protocol limits.

We already approved this approach in https://github.com/ProvableHQ/snarkVM/pull/2594 and the reasoning for why its safe is in there and in the code comments.

> Alternatively, we can disable the CI, but still allow the downloads of these SRS. (This might require more LoC changed)

Yes that would be ideal and I think we should do so, or think of a better CI setup, when it becomes prioritiy.

vicsn:
Happy to host a review session on the topic if you want.

d0cd:
Is there a way to reduce the parameter downloads only during CI via some feature flags?

niklaslong:
> Is there a way to reduce the parameter downloads only during CI via some feature flags?

Yep! Feature flags in place of the commented out code should work fine—we could introduce a `ci` feature.

vicsn:
> Is there a way to reduce the parameter downloads only during CI via some feature flags?

@raychu86 does that sound acceptable to you?

raychu86:
> > Is there a way to reduce the parameter downloads only during CI via some feature flags?
> 
> Yep! Feature flags in place of the commented out code should work fine—we could introduce a `ci` feature.

Yes let's do that!

niklaslong (approved):
LGTM 👍

.circleci/config.yml
```diff
@@ -666,6 +666,16 @@ jobs:
           workspace_member: parameters
           cache_key: v1.0.0-rust-1.81.0-snarkvm-parameters-cache
 
+  parameters-large:
+    docker:
+      - image: cimg/rust:1.81.0 # Attention - Change the MSRV in Cargo.toml and rust-toolchain as well
+    resource_class: << pipeline.parameters.large >>
+    steps:
+      - run_serial:
+          flags: --no-run --features large_params
```
niklaslong, line 675: I'd maybe add a comment here to indicate this should be restored at some point.

vicsn: Thx for re-reviewing! Added checking large parameter downloads to the Aleo <> Eiger testing roadmap, same priority as getting periodic fuzzing and code coverage to run.

PR #2604, opened 2025-01-27
[Deps] Run cargo update

ljedrz:
A pure lockfile bump; if no issues are detected, a counterpart PR will follow for snarkOS.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -4,24 +4,24 @@ version = 3
 
 [[package]]
 name = "addr2line"
-version = "0.21.0"
+version = "0.24.2"
 source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "8a30b2e23b9e17a9f90641c7ab1549cd9b44f296d3ccbf309d2863cfe398a0cb"
+checksum = "dfbe277e56a376000877090da837660b4427aad530e3028d44e0bffe4f89a1c1"
 dependencies = [
  "gimli",
 ]
 
 [[package]]
-name = "adler"
-version = "1.0.2"
+name = "adler2"
+version = "2.0.0"
 source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "f26201604c87b1e01bd3d98f8d5d9a8fcbb815e8cedb41ffccbeb4bf593a35fe"
+checksum = "512761e0bb2578dd7380c6baaa0f4ce03e84f95e960231d1dec8bf4d7d6e2627"
 
 [[package]]
 name = "ahash"
-version = "0.8.7"
+version = "0.8.11"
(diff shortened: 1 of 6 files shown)


ljedrz:
The `wasm` feature is producing [new warnings](https://github.com/rustwasm/wasm-bindgen/issues/4161) now; I special-cased them, but an alternative approach might be preferable.

ljedrz:
Hmm, I can't reproduce the `wasm` job failure locally; I'll try re-running it.

niklaslong:
It might be the nodeJS version? I see `12.22.9~dfsg-1ubuntu3.6` in the output which is pretty old?

ljedrz:
@niklaslong it's entirely possible - I attempted to reproduce the issue on 2 other systems with cutting-edge packages and was unable to do so.

Cargo.lock
```diff
@@ -478,66 +487,56 @@ dependencies = [
...
 dependencies = [
  "encode_unicode",
- "lazy_static",
  "libc",
- "unicode-width",
- "windows-sys 0.52.0",
-]
-
-[[package]]
-name = "console_error_panic_hook"
```
iamalwaysuncomfortable, line 527: Why was this removed? This lowers visibility into wasm panics.

iamalwaysuncomfortable (commented):
A couple of things

1. The `console_error_panic_hook` seems to have been removed - why?
2. The wasm bindings aren't being found. It could have to do with the version updates you applied via Cargo update not being compatible with `Cargo`'s version of `wasm-pack`. Change the [install of wasm-pack](https://rustwasm.github.io/wasm-pack/installer/) to `curl https://rustwasm.github.io/wasm-pack/installer/init.sh -sSf | sh` and see if that helps.

It could also be related to circleci failing to carry the bindings from cache, the error is:
`code: 'MODULE_NOT_FOUND',
  requireStack: [ '/tmp/.tmp4821I1/run.cjs' ]`

This indicates a wasm binding is missing from the `/tmp` folder, indicating it might just be getting deleted in some step. If changing the wasm-pack installation target doesn't work, re-run the CI with ssh and check if the wasm bindings are in the place they're expected to be. If they're not, it points to them getting deleted somehow in CI.

Cargo.lock
ljedrz: This is a lockfile change, so it must have been caused indirectly by one of the updates; it seems that it's no longer used by `wasm-bindgen-test`.

niklaslong (approved):
LGTM 🚀

vicsn (approved):

vicsn:
Manually reran `cargo update` to confirm no packages were manually added.

PR #2605, opened 2025-01-29
Improve consensus versioning clarity and guarantees

vicsn:
## Motivation

At the time of writing, consensus-relevant constants for different networks at different block heights are defined in the Environment trait: https://github.com/ProvableHQ/snarkVM/blob/mainnet/console/network/src/

Upgrades to the values of consensus-relevant constants are specified as e.g.:
- CONSENSUS_V2_HEIGHT and CONSENSUS_V3_HEIGHT
- MAX_CERTIFICATES and MAX_CERTIFICATES_BEFORE_V3

However, ideally we have a system which satisfies the following design criteria:
- A developer should be able to see what are the consensus values at a given height, and which changes were introduced at which heights, in order to audit and validate the logic changes.
- When introducing a change to a particular value at a given height, code changes should be minimal. Meaning a developer should only have to specify a {height, value} pair, should be able to retrieve it using something like `CONSTANT.get(height)`, or should only have to adjust a `match` arm.
- When introducing a new consensus version, a developer should only have to add values to constants which are actually changing.
- We don’t have to necessarily express *all* config values in a new format, just the ones which depend on height.
- Constants may only start to be defined from a certain height onwards.

## Why we can't have nice things
- Rust does not allow matching directly on runtime variables or generics, which is why this PR uses `height if height < CONSENSUS_HEIGHT(X)? => { }`.
- Many parts of the code are currently checking that certain values don't surpass `MAX_CERTIFICATES` or `MAX_COMMITTEE_SIZE`, which is not semantically correct, because those values change over time. To limit the scope of this PR, those constants have been left **unchanged**, and a new `N::MAX_COMMITTEE_SIZE` array is introduced along with a test that its largest value matches `MAX_CERTIFICATES`.

## Test Plan

- Added a bunch of unit tests with documentation of what they achieve for the new constants.

### DIFF
--- a/console/network/src/canary_v0.rs
+++ b/console/network/src/canary_v0.rs
@@ -122,6 +122,10 @@ impl Environment for CanaryV0 {
 impl Network for CanaryV0 {
     /// The block hash type.
     type BlockHash = AleoID<Field<Self>, { hrp2!("ab") }>;
+    /// The consensus block height.
+    type ConsensusHeight = u32;
+    /// The consensus block height.
+    type ConsensusVersion = u16;
     /// The ratification ID type.
     type RatificationID = AleoID<Field<Self>, { hrp2!("ar") }>;
     /// The state root type.
@@ -146,6 +150,9 @@ impl Network for CanaryV0 {
     /// The block height from which consensus V3 rules apply.
     #[cfg(any(test, feature = "test"))]
     const CONSENSUS_V3_HEIGHT: u32 = 11;
+    /// The block heights at which consensus versions are updated.
+    /// For a description of each upgrade, see the declaration of `CONSENSUS_VERSIONS` in the trait.
+    const CONSENSUS_VERSIONS: [(Self::ConsensusHeight, Self::ConsensusVersion); 2] = [(0, 0), (100, 1)];
     /// The network edition.
     const EDITION: u16 = 0;
     /// The genesis block coinbase target.
@@ -168,6 +175,9 @@ impl Network for CanaryV0 {
     const MAX_CERTIFICATES: u16 = 100;
     /// The maximum number of certificates in a batch before consensus V3 rules apply.
     const MAX_CERTIFICATES_BEFORE_V3: u16 = 100;
+    // The maximum number of certificates per round for each consensus version.
+    // This effectively limits the number of validators per round as well.
+    const MAX_CERTIFICATES_PER_ROUND: [(Self::ConsensusHeight, u16); 2] = [(0, 16), (1, 25)];
     /// The network name.
     const NAME: &'static str = "Aleo Canary (v0)";
 
--- a/console/network/src/canary_v0.rs
+++ b/console/network/src/canary_v0.rs
@@ -136,13 +136,13 @@ impl Network for CanaryV0 {
     /// The block heights at which consensus versions are updated.
     /// Documentation for what is changed at each version can be found in `Network::CONSENSUS_HEIGHT`.
     #[cfg(not(any(test, feature = "test")))]
-    const CONSENSUS_VERSIONS: [(u32, ConsensusVersion); 3] =
-        [(0, ConsensusVersion::V1), (2_900_000, ConsensusVersion::V2), (4_560_000, ConsensusVersion::V3)];
+    const CONSENSUS_VERSION_HEIGHTS: [(ConsensusVersion, u32); 3] =
+        [(ConsensusVersion::V1, 0), (ConsensusVersion::V2, 2_900_000), (ConsensusVersion::V3, 4_560_000)];
     /// The block heights at which consensus versions are updated.
     /// Documentation for what is changed at each version can be found in `Network::CONSENSUS_HEIGHT`.
     #[cfg(any(test, feature = "test"))]
-    const CONSENSUS_VERSIONS: [(u32, ConsensusVersion); 3] =
-        [(0, ConsensusVersion::V1), (10, ConsensusVersion::V2), (11, ConsensusVersion::V3)];
+    const CONSENSUS_VERSION_HEIGHTS: [(ConsensusVersion, u32); 3] =
+        [(ConsensusVersion::V1, 0), (ConsensusVersion::V2, 10), (ConsensusVersion::V3, 11)];
     /// The network edition.
     const EDITION: u16 = 0;
     /// The genesis block coinbase target.
@@ -169,16 +169,16 @@ impl Network for CanaryV0 {
     const MAX_CERTIFICATES: u16 = 100;
     /// The maximum number of validators in a committee.
     #[cfg(any(test, feature = "test"))] // TODO: or should we use the 'test-helpers' feature?
-    const MAX_COMMITTEE_SIZE: [(u32, u16); 2] = [(0, 16), (4_560_000, 25)];
+    const MAX_COMMITTEE_SIZE: [(ConsensusVersion, u16); 2] = [(ConsensusVersion::V1, 16), (ConsensusVersion::V3, 25)];
     /// The maximum number of validators in a committee.
     #[cfg(not(any(test, feature = "test")))] // TODO: or should we use the 'test-helpers' feature?
-    const MAX_COMMITTEE_SIZE: [(u32, u16); 2] = [(0, 100), (4_560_000, 100)];
+    const MAX_COMMITTEE_SIZE: [(ConsensusVersion, u16); 2] = [(ConsensusVersion::V1, 100), (ConsensusVersion::V3, 100)];
     /// The network name.
     const NAME: &'static str = "Aleo Canary (v0)";
 
     /// Returns the height at which a specified consensus version becomes active.
     fn CONSENSUS_HEIGHT(version: ConsensusVersion) -> Result<u32> {
-        Ok(Self::CONSENSUS_VERSIONS.get(version as usize).ok_or(anyhow!("Invalid consensus version"))?.0)
+        Ok(Self::CONSENSUS_VERSION_HEIGHTS.get(version as usize).ok_or(anyhow!("Invalid consensus version"))?.1)
     }
 
     /// Returns the genesis block bytes.
--- a/console/network/src/canary_v0.rs
+++ b/console/network/src/canary_v0.rs
@@ -161,44 +161,14 @@ impl Network for CanaryV0 {
     const ID: u16 = 2;
     /// The function name for the inclusion circuit.
     const INCLUSION_FUNCTION_NAME: &'static str = MainnetV0::INCLUSION_FUNCTION_NAME;
-    /// The maximum number of certificates in a batch.
+    /// A list of (consensus_version, size) pairs indicating the maximum number of certificates in a batch.
     #[cfg(any(test, feature = "test"))]
-    const MAX_CERTIFICATES: u16 = 25;
-    /// The maximum number of certificates in a batch.
+    const MAX_CERTIFICATES: [(ConsensusVersion, u16); 2] = [(ConsensusVersion::V1, 16), (ConsensusVersion::V3, 25)];
+    /// A list of (consensus_version, size) pairs indicating the maximum number of certificates in a batch.
     #[cfg(not(any(test, feature = "test")))]
-    const MAX_CERTIFICATES: u16 = 100;
-    /// A list of (consensus_version, size) pairs indicating the maximum number of validators in a committee.
-    #[cfg(any(test, feature = "test"))]
-    const MAX_COMMITTEE_SIZE: [(ConsensusVersion, u16); 2] =
-        [(ConsensusVersion::V1, 16), (ConsensusVersion::V3, Self::MAX_CERTIFICATES)];
-    /// A list of (consensus_version, size) pairs indicating the maximum number of validators in a committee.
-    #[cfg(not(any(test, feature = "test")))]
-    const MAX_COMMITTEE_SIZE: [(ConsensusVersion, u16); 2] =
-        [(ConsensusVersion::V1, 100), (ConsensusVersion::V3, Self::MAX_CERTIFICATES)];
+    const MAX_CERTIFICATES: [(ConsensusVersion, u16); 2] = [(ConsensusVersion::V1, 100), (ConsensusVersion::V3, 100)];
     /// The network name.
     const NAME: &'static str = "Aleo Canary (v0)";
-
-    /// Returns the consensus version which is active at the given height.
-    fn CONSENSUS_VERSION(seek_height: u32) -> anyhow::Result<ConsensusVersion> {
-        match Self::CONSENSUS_VERSION_HEIGHTS.binary_search_by(|(_, height)| height.cmp(&seek_height)) {
-            // If a consensus version was found at this height, return it.
-            Ok(index) => Ok(Self::CONSENSUS_VERSION_HEIGHTS[index].0),
-            // If the specified height was not found, determine whether to return an appropriate version.
-            Err(index) => {
-                if index == 0 {
-                    Err(anyhow!("Expected consensus version 1 to exist at height 0."))
-                } else {
-                    // Return the appropriate version belonging to the height *lower* than the sought height.
-                    Ok(Self::CONSENSUS_VERSION_HEIGHTS[index - 1].0)
-                }
-            }
-        }
-    }
-
-    /// Returns the height at which a specified consensus version becomes active.
-    fn CONSENSUS_HEIGHT(version: ConsensusVersion) -> Result<u32> {
-        Ok(Self::CONSENSUS_VERSION_HEIGHTS.get(version as usize - 1).ok_or(anyhow!("Invalid consensus version"))?.1)
-    }
 
     /// Returns the genesis block bytes.
     fn genesis_bytes() -> &'static [u8] {
--- a/console/network/src/canary_v0.rs
+++ b/console/network/src/canary_v0.rs
@@ -134,13 +134,15 @@ impl Network for CanaryV0 {
     type TransmissionChecksum = u128;
 
     /// The block heights at which consensus versions are updated.
-    /// Documentation for what is changed at each version can be found in `Network::HEIGHT_V`.
+    /// Documentation for what is changed at each version can be found in `Network::CONSENSUS_HEIGHT`.
     #[cfg(not(any(test, feature = "test")))]
-    const CONSENSUS_VERSIONS: [(u32, u16); 3] = [(0, 1), (2_900_000, 2), (4_560_000, 3)];
+    const CONSENSUS_VERSIONS: [(u32, ConsensusVersion); 3] =
+        [(0, ConsensusVersion::V1), (2_900_000, ConsensusVersion::V2), (4_560_000, ConsensusVersion::V3)];
     /// The block heights at which consensus versions are updated.
-    /// Documentation for what is changed at each version can be found in `Network::HEIGHT_V`.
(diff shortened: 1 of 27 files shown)


ledger/block/src/helpers/target.rs
```diff
@@ -36,12 +36,14 @@ pub fn block_reward<N: Network>(
...
     transaction_fees: u64,
-) -> u64 {
+) -> Result<u64> {
     // Determine which block reward version to use.
-    match block_height < N::CONSENSUS_V2_HEIGHT {
-        true => block_reward_v1(total_supply, block_time, coinbase_reward, transaction_fees),
-        false => block_reward_v2(total_supply, time_since_last_block, coinbase_reward, transaction_fees),
-    }
+    Ok(match block_height {
+        height if height < N::HEIGHT_V(2)? => {
```
raychu86: IMO this code snippet is more confusing than it was before; without context, `height < N::HEIGHT_V(2)` is less clear than `block_height < N::CONSENSUS_V2_HEIGHT`.

It could be the variable naming or the use of a non-constant value of `2`, but it feels a bit unnatural.

console/network/src/canary_v0.rs
```diff
@@ -165,9 +160,17 @@ impl Network for CanaryV0 {
...
+    #[cfg(not(any(test, feature = "test")))] // TODO: or should we use the 'test-helpers' feature?
     const MAX_CERTIFICATES: u16 = 100;
-    /// The maximum number of certificates in a batch before consensus V3 rules apply.
-    const MAX_CERTIFICATES_BEFORE_V3: u16 = 100;
+    /// The maximum number of validators in a committee.
+    #[cfg(any(test, feature = "test"))] // TODO: or should we use the 'test-helpers' feature?
+    const MAX_COMMITTEE_SIZE: [(u32, u16); 2] = [(0, 16), (4_560_000, 25)];
+    /// The maximum number of validators in a committee.
+    #[cfg(not(any(test, feature = "test")))] // TODO: or should we use the 'test-helpers' feature?
+    const MAX_COMMITTEE_SIZE: [(u32, u16); 2] = [(0, 100), (4_560_000, 100)];
```
raychu86: Is there a way to const-ify the height declarations? For example, `4_560_000` is declared 3 times instead of declared once in a unified way.

console/network/src/mainnet_v0.rs
```diff
@@ -134,19 +134,14 @@ impl Network for MainnetV0 {
     /// The transmission checksum type.
     type TransmissionChecksum = u128;
 
-    /// The block height from which consensus V2 rules apply.
+    /// The block heights at which consensus versions are updated.
```
raychu86:
```suggestion
    /// A list of (block_height, consensus_version) pairs indicating when each consensus version takes effect.
```

Documentation should explain what the tuple values are.

console/network/src/mainnet_v0.rs
```diff
@@ -170,12 +165,25 @@ impl Network for MainnetV0 {
...
+    #[cfg(any(test, feature = "test"))] // TODO: or should we use the 'test-helpers' feature?
     const MAX_CERTIFICATES: u16 = 25;
-    /// The maximum number of certificates in a batch before consensus V3 rules apply.
-    const MAX_CERTIFICATES_BEFORE_V3: u16 = 16;
+    /// The maximum number of certificates in a batch.
+    #[cfg(not(any(test, feature = "test")))] // TODO: or should we use the 'test-helpers' feature?
+    const MAX_CERTIFICATES: u16 = 100;
+    /// The maximum number of validators in a committee.
+    #[cfg(any(test, feature = "test"))] // TODO: or should we use the 'test-helpers' feature?
+    const MAX_COMMITTEE_SIZE: [(u32, u16); 2] = [(0, 16), (4_900_000, 25)];
```
raychu86: Is it better to use block height as the delineator here or consensus version?

ledger/block/src/helpers/target.rs
raychu86: Maybe a wrapper `enum` would be better than `2`.

For instance: `CONSENSUS::V2`, `CONSENSUS_RULES::V2`, `CONSENSUS_VERSION::V2`.

console/network/src/canary_v0.rs
ljedrz: indeed, it would be best to reference `CONSENSUS_VERSIONS` and be consistent about it

console/network/src/canary_v0.rs
```diff
@@ -165,9 +160,17 @@ impl Network for CanaryV0 {
     /// The function name for the inclusion circuit.
     const INCLUSION_FUNCTION_NAME: &'static str = MainnetV0::INCLUSION_FUNCTION_NAME;
     /// The maximum number of certificates in a batch.
+    #[cfg(any(test, feature = "test"))] // TODO: or should we use the 'test-helpers' feature?
```
ljedrz: perhaps we should consider joining both features?

console/network/src/canary_v0.rs
```diff
@@ -503,6 +506,11 @@ impl Network for CanaryV0 {
     ) -> bool {
         path.verify(&*CANARY_POSEIDON_4, &*CANARY_POSEIDON_2, root, leaf)
     }
+
+    /// Returns the height at which a specified consensus version becomes active.
+    fn HEIGHT_V(version: usize) -> Result<u32> {
+        Ok(Self::CONSENSUS_VERSIONS.get(version).ok_or(anyhow!("Invalid consensus version"))?.0)
```
ljedrz: since we're indexing the versions starting from `1`, shouldn't this instead be the following?
```suggestion
        Ok(Self::CONSENSUS_VERSIONS.get(version - 1).ok_or(anyhow!("Invalid consensus version"))?.0)
```
The input `version` could be a `NonZeroUsize` to avoid any confusion about the beginning.

edit: I'm pretty sure this is the cause of the current CI job failures

ljedrz: with such a setup the tuples would IMO be redundant (assuming that the versions are going to always be contiguous)

console/network/src/mainnet_v0.rs
vicsn: 4d3b948e0

console/network/src/canary_v0.rs
vicsn: Makes sense, updated the setup: 0b5e9f775

ledger/block/src/helpers/target.rs
vicsn: Great idea! c928f35f3

console/network/src/mainnet_v0.rs
vicsn: Now using the consensus version to further reduce the chance of mistakes. The only downside is that the lookup of a constant value by its height now requires a double lookup - first we find the `version` in `CONSENSUS_HEIGHTS` and then with that version in hand we find the value. See the updated `consensus_config_value`. It seems acceptable to me.

console/network/src/canary_v0.rs
vicsn: Thx, should be fixed by c928f35

vicsn: Thx, should be fixed by c928f35

console/network/src/canary_v0.rs
vicsn: One day, one day. Removed the comments, I just wanted to see if CI would pass as is: 70c4ec66a

synthesizer/src/vm/verify.rs
```diff
@@ -243,9 +243,11 @@ impl<N: Network, C: ConsensusStorage<N>> VM<N, C> {
                             .block_store()
                             .find_block_height_from_state_root(execution.global_state_root())?
                             .unwrap_or_default();
-                        let (cost, (_, _)) = match block_height < N::CONSENSUS_V2_HEIGHT {
-                            true => execution_cost_v1(&self.process().read(), execution)?,
-                            false => execution_cost_v2(&self.process().read(), execution)?,
+                        let (cost, (_, _)) = match block_height {
```
d0cd: If we had a utility like `N::CONSENSUS_VERSION(height: u32) -> ConsensusVersion`, this block could be rewritten as: 
```
match N::CONSENSUS_VERSION(block_height) {
    ConsensusVersion::V1 => execution_cost_v1(&self.process().read(), execution)?
    _ => execution_cost_v2(&self.process().read(), execution)?
}
```
Would that improve ergonomics? Ditto for similar patterns in this PR.

console/network/src/testnet_v0.rs
```diff
@@ -165,12 +162,25 @@ impl Network for TestnetV0 {
...
+    #[cfg(not(any(test, feature = "test")))]
     const MAX_CERTIFICATES: u16 = 100;
-    /// The maximum number of certificates in a batch before consensus V3 rules apply.
-    const MAX_CERTIFICATES_BEFORE_V3: u16 = 100;
+    /// A list of (consensus_version, size) pairs indicating the maximum number of validators in a committee.
+    #[cfg(any(test, feature = "test"))]
+    const MAX_COMMITTEE_SIZE: [(ConsensusVersion, u16); 2] = [(ConsensusVersion::V1, 16), (ConsensusVersion::V3, 25)];
+    /// A list of (consensus_version, size) pairs indicating the maximum number of validators in a committee.
+    #[cfg(not(any(test, feature = "test")))]
+    const MAX_COMMITTEE_SIZE: [(ConsensusVersion, u16); 2] = [(ConsensusVersion::V1, 100), (ConsensusVersion::V3, 100)];
```
raychu86: What is the intention for adding this to the network trait?

It should be coupled with `MAX_CERTIFICATES`. And then `MAX_COMMITTEE_SIZE` is set using `N::MAX_CERTIFICATES`

synthesizer/src/vm/verify.rs
vicsn: Great idea 🙏 9d26eadfa

console/network/src/testnet_v0.rs
vicsn: Added a very light coupling here: de2f3d36c

If we only want a single unique `MAX_CERTIFICATES` declaration, which I agree would be better, we end up with this is roughly the smallest diff I can come up with: a3124053e

ledger/block/src/helpers/target.rs
```diff
@@ -36,12 +36,12 @@ pub fn block_reward<N: Network>(
...
     coinbase_reward: u64,
     transaction_fees: u64,
-) -> u64 {
+) -> Result<u64> {
     // Determine which block reward version to use.
-    match block_height < N::CONSENSUS_V2_HEIGHT {
-        true => block_reward_v1(total_supply, block_time, coinbase_reward, transaction_fees),
-        false => block_reward_v2(total_supply, time_since_last_block, coinbase_reward, transaction_fees),
-    }
+    Ok(match N::CONSENSUS_VERSION(block_height)? as usize {
```
d0cd: While the number of versions is small, I'd vote for matching on the enums for more clarity.

console/network/src/lib.rs
```diff
@@ -220,6 +215,50 @@ pub trait Network:
...
+        }
+    }
+    /// Returns the height at which a specified consensus version becomes active.
+    #[allow(non_snake_case)]
+    fn CONSENSUS_HEIGHT(version: ConsensusVersion) -> Result<u32> {
+        Ok(Self::CONSENSUS_VERSION_HEIGHTS.get(version as usize - 1).ok_or(anyhow!("Invalid consensus version"))?.1)
+    }
+    /// Returns the last `MAX_CERTIFICATES` value.
+    #[allow(non_snake_case)]
+    fn LAST_MAX_CERTIFICATES() -> Result<u16> {
```
raychu86: nit
```suggestion
    fn LATEST_MAX_CERTIFICATES() -> Result<u16> {
```

ledger/block/src/helpers/target.rs
raychu86: Why cast to `usize` instead of directly comparing against `ConsensusVersion::V*`?

ledger/narwhal/batch-certificate/src/lib.rs
```diff
@@ -45,14 +45,16 @@ pub struct BatchCertificate<N: Network> {
...
+        N::LAST_MAX_CERTIFICATES()
+    }
 }
 
 impl<N: Network> BatchCertificate<N> {
     /// Initializes a new batch certificate.
     pub fn from(batch_header: BatchHeader<N>, signatures: IndexSet<Signature<N>>) -> Result<Self> {
         // Ensure that the number of signatures is within bounds.
-        ensure!(signatures.len() <= Self::MAX_SIGNATURES as usize, "Invalid number of signatures");
+        ensure!(signatures.len() <= N::LAST_MAX_CERTIFICATES()? as usize, "Invalid number of signatures");
```
raychu86: nit
```suggestion
        ensure!(signatures.len() <= Self::max_signatures()? as usize, "Invalid number of signatures");
```

raychu86 (approved):
Logic looks good. Left a few nits and have some formatting requests. 

Would appreciate additional reviews on this - especially on the macro logic

ledger/block/src/helpers/target.rs
vicsn: I agree it would be clearer for now, but it will also create an *inevitable* moment of confusion to a future developer when they have to potentially refactor hundreds of lines of code to use a range comparison again. I vote for matching on integers still.

vicsn: It is very likely we will have 100+ consensus versions. If branch A applies to the first 50, and branch B applies to the next 50, the cleanest way to implement this is to use `Range`s.

We can't use ranges on enums, only on ints. 

Maybe there is additional syntax sugar we can apply, but match arm logic needs to be const, so that would have to be additional macro magic.

ledger/narwhal/batch-certificate/src/lib.rs
vicsn: Thx! 3300451f7

console/network/src/lib.rs
vicsn: 47183c4a9

console/network/src/lib.rs
```diff
@@ -410,3 +449,132 @@ pub trait Network:
...
+    /// It is possible this invariant no longer holds in the future, e.g. due to pruning or novel types of constants.
+    fn consensus_constants_at_genesis<N: Network>() {
+        let height = N::CONSENSUS_VERSION_HEIGHTS.first().unwrap().1;
+        assert_eq!(height, 0);
+        let consensus_version = N::CONSENSUS_VERSION_HEIGHTS.first().unwrap().0;
+        assert_eq!(consensus_version, ConsensusVersion::V1);
+        assert_eq!(consensus_version as usize, 1);
+    }
+
+    /// Ensure that the consensus *versions* are unique, incrementing and start with 0.
```
ljedrz:
```suggestion
    /// Ensure that the consensus *versions* are unique, incrementing and start with 1.
```

console/network/src/lib.rs
```diff
@@ -410,3 +449,132 @@ pub trait Network:
...
+    fn consensus_versions<N: Network>() {
+        let mut previous_version = N::CONSENSUS_VERSION_HEIGHTS.first().unwrap().0;
+        // Ensure that the consensus versions start with 0.
```
ljedrz:
```suggestion
        // Ensure that the consensus versions start with 1.
```
unless what is meant here is the height

ledger/narwhal/batch-header/src/lib.rs
```diff
@@ -56,20 +56,6 @@ pub struct BatchHeader<N: Network> {
...
-    pub const MAX_CERTIFICATES: u16 = 100;
-    /// The maximum number of certificates in a batch before consensus V3 rules apply.
-    #[cfg(not(any(test, feature = "test-helpers")))]
-    pub const MAX_CERTIFICATES_BEFORE_V3: u16 = N::MAX_CERTIFICATES_BEFORE_V3;
-    /// The maximum number of certificates in a batch before consensus V3 rules apply.
-    /// This is deliberately set to a high value (100) for testing purposes only.
-    #[cfg(any(test, feature = "test-helpers"))]
-    pub const MAX_CERTIFICATES_BEFORE_V3: u16 = 100;
     /// The maximum number of rounds to store before garbage collecting.
     pub const MAX_GC_ROUNDS: usize = 100;
```
ljedrz, line 60: nit: in the future we might want to move these final 2 consts to the `Network` too

synthesizer/process/src/finalize.rs
```diff
@@ -104,10 +104,10 @@ impl<N: Network> Process<N> {
         lap!(timer, "Verify the number of transitions");
 
         // Construct the call graph.
-        // If the height is greater than or equal to `CONSENSUS_V3_HEIGHT`, then provide an empty call graph, as it is no longer used during finalization.
-        let call_graph = match state.block_height() < N::CONSENSUS_V3_HEIGHT {
-            true => self.construct_call_graph(execution)?,
-            false => HashMap::new(),
+        // If the height is greater than or equal to `ConsensusVersion::V3`, then provide an empty call graph, as it is no longer used during finalization.
+        let call_graph = match N::CONSENSUS_VERSION(state.block_height())? as usize {
+            1..3 => self.construct_call_graph(execution)?,
```
ljedrz:
```suggestion
            1..=2 => self.construct_call_graph(execution)?,
```
if possible - for ultimate clarity; applies to the other 2 instances of this range in this module and the next one

ljedrz (approved):
LGTM, left a handful of nits

niklaslong (approved):

synthesizer/process/src/finalize.rs
vicsn: Good idea, 5b32053b7

console/network/src/lib.rs
vicsn: Thx 4beaa7a82

raychu86:
@vicsn Looks like there is some failing CI, I assume it's due to feature flagging with "test".

ledger/block/src/helpers/target.rs
vicsn: New idea: instead or on top of defining consensus versions in an enum, we define a bunch of `const CONSENSUS_V1: usize = 1;`. That may be clearer than a raw `1`. But it doesn't help *that* much and its a line more boilerplate to write for each version. Let me know what you prefer.

d0cd (approved):

ledger/block/src/helpers/target.rs
raychu86: I think keeping a usize comparison is fine, but would prefer to compare it against `ConsensusVersion::V1` instead of just `1`.

Ranges can be done with: `0..ConsensusVersion::V99` and `ConsensusVersion::V99..`

vicsn: I wrote above: We can't use ranges on enums, only on ints.

raychu86: You should be able to cast it to an integer. 0..(ConsensusVersion::V99 as usize). 

If that doesn't work you can have a static function as well - `ConsensusVersion::V99.as_int()`

vicsn: Unfortunately non-const values or functions are not allowed. Even the `as_int()` is not allowed in a match arm.

However, I realise that we don't use the advantage of using match arms anyway. The compiler could help us and complain whether we're covering all versions, but this doesn't apply because we're using a catch all `_` case for the latest versions to avoid having to update all checks for every consensus version.

Implemented comparison against `ConsensusVersion` using if-else statements in a new commit here: 91018e522

PR #2609, opened 2025-02-12
Fix bug in parsing of comments.

acoglio:
When parsing a line comment `// ...` that ends without a newline (because it ends with the input), the parser was neglecting to ensure that there are no unsafe characters in the comment. This commit adds that check, and also adds a test that would fail without this fix.

Also improve and fix some documentation comments in the code, both in the file that was fixed and in a couple of other files.

### DIFF
--- a/console/network/environment/src/helpers/sanitizer.rs
+++ b/console/network/environment/src/helpers/sanitizer.rs
@@ -51,7 +51,7 @@ impl Sanitizer {
         )(string)
     }
 
-    /// Parse a safe character (in the sense explained in [string_parser::is_char_supported]).
+    /// Parse a safe character (in the sense explained in [is_char_supported]).
     /// Returns an error if no character is found or a non-safe character is found.
     /// The character is returned, along with the remaining input.
     ///
@@ -84,9 +84,8 @@ impl Sanitizer {
         }
     }
 
-    /// A newline parser that accepts:
-    ///
-    /// - A newline.
+    /// A parser that accepts:
+    /// - A newline, either `CR LF` or just `LF`.
     /// - The end of input.
     fn eol(string: &str) -> ParserResult<()> {
         alt((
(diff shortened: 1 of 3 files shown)


vicsn (approved):

vicsn:
CC @mikebenfield can't add you as reviewer, but please do :)

console/network/environment/src/helpers/sanitizer.rs
```diff
@@ -113,15 +112,19 @@ impl Sanitizer {
 
     /// Parse a string until the end of line.
     ///
-    /// This parser accepts the multiline annotation (\) to break the string on several lines.
+    /// This parser accepts the multiline annotation (`\ LF`) to break the string on several lines.
+    ///
+    /// The line may end with a newline, either `CR LF` or just `LF`; or it may end with the input.
```
bendyarm:
```suggestion
    /// The line may end with a newline (either `CR LF` or just `LF`), or it may end with the input.
```

bendyarm (approved):
Minor rewording to make it clear there are two kinds of newlines, not three.

console/network/environment/src/helpers/sanitizer.rs
```diff
@@ -113,15 +112,19 @@ impl Sanitizer {
...
     ///
-    /// Discard any leading newline.
+    /// Return the body of the comment, i.e. what is between `//` and the end of line.
+    /// If the line ends with `CR LF`, the `CR` is included in the returned body.
+    /// The `LF`, if present, is never included in the returned body.
     fn str_till_eol(string: &str) -> ParserResult<&str> {
         // A heuristic approach is applied here in order to avoid costly parsing operations in the
         // most common scenarios: non-parsing methods are used to verify if the string has multiple
         // lines and if there are any unsafe characters.
         if let Some((before, after)) = string.split_once('\n') {
```
mikebenfield, line 126: You can unify these cases to make this function much briefer like this:

```rust
let (before, after) = string.split_once('\n').unwrap_or((string, ""));
```

console/network/environment/src/helpers/sanitizer.rs
```diff
@@ -140,12 +144,19 @@ impl Sanitizer {
...
                     |i| {
+                        // Exclude the final `LF`, if any, from the comment body.
                         if i.as_bytes().last() == Some(&b'\n') { &i[0..i.len() - 1] } else { i }
                     },
                 )(string)
             }
-        } else {
+        } else if string.chars().all(is_char_supported) {
+            // There is no `LF`. We return all the characters up to the end of file.
             Ok((string, ""))
```
mikebenfield: Isn't this backwards? I see that it's unchanged from the previous code, but since above you return `(after, before)`, shouldn't you here return `("", string)`? Am I missing something from the logic here?

acoglio: I think you are right: this was a second, pre-existing, independent bug. It's an easy fix, good to do as part of this PR.

console/network/environment/src/helpers/sanitizer.rs
acoglio: Would this treat the case of no LF till end of file the same as the case of just one LF at the very end of the file? If so, would is_multiline be incorrect if there is no LF but the file ends with backslash (it would be true, even though there is no `\ LF`)? Maybe the rest of the code in this function all works out nonetheless, but I'm thinking that the current PR is pretty clear, and maybe we could try to abbreviate the function in a separate PR?

mikebenfield: Yeah it does get slightly hairy; all good if you want to leave it as is.

acoglio:
I looked at the CI failures but they don't seem related to my changes (?).

vicsn:
> I looked at the CI failures but they don't seem related to my changes (?).

Yes you can ignore.

mikebenfield (approved):
LGTM

PR #2612, opened 2025-02-17
[Deps] Run cargo update, update the toolchain

ljedrz:
A lockfile bump, plus an update to the toolchain (1.81 > 1.83) that will facilitate the use of the recently adjusted consensus-related constants and associated functions in snarkOS.


niklaslong (approved):
LGTM

vicsn (approved):

PR #2615, opened 2025-02-21
Add additional subdag verification

raychu86:
## Motivation

This PR adds check to enforce that all certificates in a block subdag have sufficient quorum requirements based on the signatures.

Additionally, a few constructions of `committee_lookback` have been refactored to properly use the `Ledger::get_committee_lookback_for_round` helper method.

### DIFF
--- a/ledger/src/check_next_block.rs
+++ b/ledger/src/check_next_block.rs
@@ -81,36 +81,17 @@ impl<N: Network, C: ConsensusStorage<N>> Ledger<N, C> {
         )?;
 
         // Retrieve the committee lookback.
-        let committee_lookback = {
-            // Determine the round number for the previous committee. Note, we subtract 2 from odd rounds,
-            // because committees are updated in even rounds.
-            let previous_round = match block.round() % 2 == 0 {
-                true => block.round().saturating_sub(1),
-                false => block.round().saturating_sub(2),
-            };
-            // Determine the committee lookback round.
-            let committee_lookback_round = previous_round.saturating_sub(Committee::<N>::COMMITTEE_LOOKBACK_RANGE);
-            // Output the committee lookback.
-            self.get_committee_for_round(committee_lookback_round)?
-                .ok_or(anyhow!("Failed to fetch committee for round {committee_lookback_round}"))?
-        };
+        let committee_lookback = self
+            .get_committee_lookback_for_round(block.round())?
+            .ok_or(anyhow!("Failed to fetch committee lookback for round {}", block.round()))?;
 
         // Retrieve the previous committee lookback.
         let previous_committee_lookback = {
             // Calculate the penultimate round, which is the round before the anchor round.
             let penultimate_round = block.round().saturating_sub(1);
-            // Determine the round number for the previous committee. Note, we subtract 2 from odd rounds,
-            // because committees are updated in even rounds.
-            let previous_penultimate_round = match penultimate_round % 2 == 0 {
-                true => penultimate_round.saturating_sub(1),
-                false => penultimate_round.saturating_sub(2),
-            };
-            // Determine the previous committee lookback round.
-            let penultimate_committee_lookback_round =
-                previous_penultimate_round.saturating_sub(Committee::<N>::COMMITTEE_LOOKBACK_RANGE);
-            // Output the previous committee lookback.
-            self.get_committee_for_round(penultimate_committee_lookback_round)?
-                .ok_or(anyhow!("Failed to fetch committee for round {penultimate_committee_lookback_round}"))?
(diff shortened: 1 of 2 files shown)


raychu86:
@acoglio

ljedrz (approved):
LGTM :+1:

acoglio:
To provide context for others, this results from a discussion that @raychu86 and I had at the office yesterday.

When a validator starts up, it only knows the genesis block and the genesis committees as "roots of trust". When receiving blocks from its peers, some of which may be faulty, how does the validator know whether those are indeed the right blocks that the network has agreed on so far, as opposed to some otherwise valid blocks (i.e. with valid transactions, hashes, etc.) but not quite the ones that were agreed upon? It's not an easy question to answer, but if the validator can (perhaps implicitly) reconstruct and validate the full DAG, from the sub-DAGs in the blocks, then it can trace things back to the aforementioned "roots of trust". A critical check for validating the DAG is that the certificates have a quorum of signatures in the active committee for the round. This PR adds that check. Note that all blocks are required to have a sub-DAG authority; the beacon authority is only used for testing (and in the genesis block, but just as a placeholder of sorts).

@raychu86 I thought a bit more about the possibility of only checking the anchor. That might work, but the formal proof for that has to involve a more explicit notion of "cryptographically injective" hashes. Doable, but not what we have now. So I still recommend checking all certificates for now, since it's fairly cheap — signatures were already verified during deserialization. If/when we have a formal model and proofs involving hashes, we might more confidently decide to just check the anchors.

ledger/src/check_next_block.rs
```diff
@@ -125,6 +106,9 @@ impl<N: Network, C: ConsensusStorage<N>> Ledger<N, C> {
             ratified_finalize_operations,
         )?;
 
+        // Ensure the certificates in the block subdag have met quorum requirements.
+        self.check_block_subdag_quorum(block)?;
```
vicsn, line 110: How would we feel about moving this to the syncing logic in snarkOS? Reasoning:
- This check is only needed when syncing, so it is better separation of concerns.
- This check might actually be expensive, because it deserializes a large amount of addresses from disk, which is a key reason why in snarkOS there is now a `committee_cache`.

raychu86: Ideally it is constricted to `snarkVM` so we do not require splitting consensus checks between repos. However I do see the concern of performance, which is why I inquired about potentially just checking the anchor certificate rather than all of the certificates and relying on Bullshark properties to guarantee safety.

Open to other ideas. Maybe snarkVM can have an interface for the caches? Or we can benchmark block verification on different validator sizes to measure impact.

vicsn: > Ideally it is constricted to snarkVM so we do not require splitting consensus checks between repos.

Yeah agreed. I'm curious whether some of the `sync_storage_with_block` logic in snarkOS might now be duplicate or should be moved into snarkVM as well then. @acoglio want to look into that - in due time?

> Open to other ideas. Maybe snarkVM can have an interface for the caches?

Sounds like this would be the fastest + safest route forward. @ljedrz since you made the original committee_cache and I think have some bandwidth, want to make a snarkVM+snarkOS PR next week which moves the cache into snarkVM?

acoglio:
Looking at sync_storage_with_block(), I see no duplicate functionality with the code added by this commit, which checks quorum signatures for sub-DAG certificates, while the checks in sync_storage_with_block() concern voting thresholds (> f) for leader certificates.

Whether that, or some of that, should be moved to snarkVM, I'm not sure. Perhaps easier to discuss live.

vicsn (approved):

ledger/src/check_next_block.rs
vicsn: It lives: https://github.com/ProvableHQ/snarkVM/pull/2626

ledger/src/check_next_block.rs
```diff
@@ -145,6 +129,43 @@ impl<N: Network, C: ConsensusStorage<N>> Ledger<N, C> {
...
+            // because that is done during construction in `BatchCertificate::new`.
+            cfg_iter!(certificates).try_for_each(|certificate| {
+                // Collect the signature authors.
+                let authors = certificate.signatures().map(|signature| signature.to_address()).collect();
+                // Ensure that the signers of the certificate reach the quorum threshold.
+                ensure!(
+                    committee_lookback.is_quorum_threshold_reached(&authors),
+                    "Certificate '{}' for round {round} does not meet quorum requirements",
+                    certificate.id()
+                );
```
raychu86, line 158: The error here is that we need to include the author of the certificate as a part of the signers.

Also I misnamed the variable `authors` when it should be `signers`.

PR #2621, opened 2025-02-27
[Perf] Simplify the R1CS > Assignment conversion

ljedrz:
The next (after the independent https://github.com/ProvableHQ/snarkVM/pull/2620) chapter in the synthesis optimization saga.

The conversion from `R1CS` to the `Assignment` needed for `ConstraintSynthesizer::generate_constraints` is very costly due to the allocations involved. However, most of it is actually unnecessary, and despite my worries that it had been done in order to reduce memory use, skipping it actually leads to lower utilization of RAM.

The performance impact, measured with the same synthesis as in the aforementioned other PR, is as follows:
- runtime **-21%**
- total allocs **and** deallocs **-18%** each
- peak RAM **-43%**

As for the benchmark from the aforementioned other PR:
```
CheckDeployment for do  time:   [847.87 ms 848.42 ms 848.94 ms]
                        change: [-35.625% -34.795% -34.243%] (p = 0.00 < 0.05)
                        Performance has improved.
```
However, this one also has an impact on another new benchmark from that PR:
```
CheckDeployment for transfer_private
                        time:   [139.91 ms 140.01 ms 140.20 ms]
                        change: [-10.901% -9.3525% -7.7967%] (p = 0.00 < 0.05)
                        Performance has improved.
```
(there is also a new `transfer_public` benchmark, but it's not as stable on my machine, returning results ranging from -3% to +3%)

### DIFF
--- a/circuit/environment/src/helpers/counter.rs
+++ b/circuit/environment/src/helpers/counter.rs
@@ -16,17 +16,17 @@
 use crate::*;
 use snarkvm_fields::PrimeField;
 
-use std::{mem, rc::Rc};
+use std::{mem, sync::Arc};
 
 #[derive(Debug, Default, Hash)]
 pub(crate) struct Counter<F: PrimeField> {
     scope: Scope,
-    constraints: Vec<Rc<Constraint<F>>>,
+    constraints: Vec<Arc<Constraint<F>>>,
     constants: u64,
     public: u64,
     private: u64,
     nonzeros: (u64, u64, u64),
-    parents: Vec<(Scope, Vec<Rc<Constraint<F>>>, u64, u64, u64, (u64, u64, u64))>,
+    parents: Vec<(Scope, Vec<Arc<Constraint<F>>>, u64, u64, u64, (u64, u64, u64))>,
 }
 
 impl<F: PrimeField> Counter<F> {
@@ -93,7 +93,7 @@ impl<F: PrimeField> Counter<F> {
     }
 
     /// Increments the number of constraints by 1.
-    pub(crate) fn add_constraint(&mut self, constraint: Rc<Constraint<F>>) {
+    pub(crate) fn add_constraint(&mut self, constraint: Arc<Constraint<F>>) {
         let (a_nonzeros, b_nonzeros, c_nonzeros) = constraint.num_nonzeros();
         self.nonzeros.0 += a_nonzeros;
         self.nonzeros.1 += b_nonzeros;
--- a/circuit/environment/src/helpers/r1cs.rs
+++ b/circuit/environment/src/helpers/r1cs.rs
@@ -21,7 +21,7 @@ use snarkvm_fields::PrimeField;
 
 #[cfg(feature = "save_r1cs_hashes")]
 use sha2::{Digest, Sha256};
-use std::rc::Rc;
+use std::sync::Arc;
 #[cfg(feature = "save_r1cs_hashes")]
 use std::{
     hash::{Hash, Hasher},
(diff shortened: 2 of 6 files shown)


circuit/environment/src/helpers/variable.rs
```diff
@@ -90,6 +90,19 @@ impl<F: PrimeField> Variable<F> {
             }
         }
     }
+
+    ///
+    /// Returns the relative index and value of the variable.
+    ///
```
vicsn, line 96: Might be nice to document as reminder to readers that `Public` variables occupy indices [0, m] for some m, and `Private` will then occupy [m + 1, n].

circuit/environment/src/helpers/r1cs.rs
```diff
@@ -19,18 +19,18 @@ use crate::{
...
 
 #[derive(Debug)]
 pub struct R1CS<F: PrimeField> {
     constants: Vec<Variable<F>>,
-    public: Vec<Variable<F>>,
-    private: Vec<Variable<F>>,
-    constraints: Vec<Rc<Constraint<F>>>,
+    pub(crate) public: Vec<Variable<F>>,
+    pub(crate) private: Vec<Variable<F>>,
+    pub(crate) constraints: Vec<Arc<Constraint<F>>>,
```
vicsn, line 67: Can you make a tracking issue for remaining future synthesis speedup ideas? 
- You can note that you didn't find a speedup with arenas.
- I suspect we might be able to hide some of this tracking in `R1CS` and `Counter` behind a feature flag. (applies to `CheckSynthesis` benchmark)
- Maybe we can get rid of the intermediary `Converter` and new `LinearCombination` in `generate_constraints` (applies to `verify_vk` benchmark)
- Maybe we can pass in `Assignment` by value. (applies to `verify_vk` benchmark)

circuit/environment/src/helpers/counter.rs
```diff
@@ -16,17 +16,17 @@
 use crate::*;
 use snarkvm_fields::PrimeField;
 
-use std::{mem, rc::Rc};
+use std::{mem, sync::Arc};
```
vicsn, line 19: Why the change to Arc? I recall Howard mention several times that this part of the code was explicitly designed to be only used in single-threaded runtimes, so Arc might give the wrong impression.

ljedrz: The switch is needed in order for those objects to work during the `Assignment` stage or afterwards. `Arc` is perfectly viable as a substitute for `Rc` even if no parallelism is involved, and we can always make an appropriate comment indicating why it is not an `Rc`.

kpandl (approved):
Nice optimization, LGTM!

circuit/environment/src/helpers/variable.rs
ljedrz: Don't they have their individual indices?
```
    pub(crate) fn new_public(&mut self, value: F) -> Variable<F> {
        let variable = Variable::Public(Arc::new((self.public.len() as u64, value)));
        self.public.push(variable.clone());
        self.counter.increment_public();
        self.num_variables += 1;
        variable
    }

    pub(crate) fn new_private(&mut self, value: F) -> Variable<F> {
        let variable = Variable::Private(Arc::new((self.private.len() as u64, value)));
        self.private.push(variable.clone());
        self.counter.increment_private();
        self.num_variables += 1;
        variable
    }

```

ljedrz:
Rebased.

niklaslong (approved):
The rust looks good to me, I'll defer to @vicsn on the protocol but that also seems good 👍

vicsn (approved):
LGTM, the intermediary object contained exactly the same indices, and the r1cs object is passed in by value, so I see no angle for any logic to be different.

PR #2622, opened 2025-02-27
[Perf] Use SmallVec in LinearCombination

ljedrz:
I think this PR, third in the series (https://github.com/ProvableHQ/snarkVM/pull/2620, https://github.com/ProvableHQ/snarkVM/pull/2621), concludes the heap-related optimizations that can be applied to synthesis, at least based on the new `large_program` benchmark. It is also the simplest - it only substitutes one `Vec` with a `SingleVec`, which is optimized for the scenario that is most common/impactful - `LinearCombination`s consisting of a single term. I've also run profiling using a `SmallVec<[T; N]>` with `N` ranging from `1` to `8`, and it performed favorably to all of them.

This change has the following impact:
- total allocs & deallocs **-47%**
- max RAM **-2%**
```
CheckDeployment for do  time:   [867.22 ms 869.05 ms 871.85 ms]
                        change: [-32.516% -31.569% -30.881%] (p = 0.00 < 0.05)
                        Performance has improved.
```

### DIFF
--- a/circuit/environment/Cargo.toml
+++ b/circuit/environment/Cargo.toml
@@ -64,6 +64,9 @@ version = "0.10"
 default-features = false
 optional = true
 
+[dependencies.singlevec]
+version = "1.7"
+
 [dev-dependencies.snarkvm-algorithms]
 path = "../../algorithms"
 features = [ "polycommit_full", "snark", "test" ]
--- a/circuit/environment/Cargo.toml
+++ b/circuit/environment/Cargo.toml
@@ -64,8 +64,8 @@ version = "0.10"
 default-features = false
 optional = true
 
-[dependencies.singlevec]
-version = "1.7"
+[dependencies.smallvec]
+version = "1.14"
 
 [dev-dependencies.snarkvm-algorithms]
 path = "../../algorithms"
--- a/circuit/environment/src/helpers/linear_combination.rs
+++ b/circuit/environment/src/helpers/linear_combination.rs
@@ -20,6 +20,7 @@ use core::{
     fmt,
     ops::{Add, AddAssign, Mul, Neg, Sub},
 };
+use singlevec::SingleVec;
 
 // Before high level program operations are converted into constraints, they are first tracked as linear combinations.
 // Each linear combination corresponds to a portion or all of a single row of an R1CS matrix, and consists of:
@@ -35,7 +36,7 @@ use core::{
 pub struct LinearCombination<F: PrimeField> {
     constant: F,
     /// The list of terms is kept sorted in order to speed up lookups.
-    terms: Vec<(Variable<F>, F)>,
+    terms: SingleVec<(Variable<F>, F)>,
     /// The value of this linear combination, defined as the sum of the `terms` and `constant`.
     value: F,
 }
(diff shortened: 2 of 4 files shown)


circuit/environment/Cargo.toml
```diff
@@ -59,6 +59,9 @@ version = "0.2"
 [dependencies.once_cell]
 version = "1.18.0"
 
+[dependencies.singlevec]
```
vicsn: 👁️ 👁️ 
Screenshot 2025-03-08 at 21 34 11

How big would the diff be if we just use tinyvec? Do snark{VM, OS} have other dependencies which are only maintained by 1-2 folks? I only know of `deadline`. In such cases, I'm thinking vendoring might be the better way forward so we take full ownership over the auditing...

ljedrz: I'm happy to try out `tinyvec` as well, but I suspect it would also have worse performance (as long as it doesn't special-case a single-element collection).
The `singlevec` crate is very simple (in fact, a lot simpler than `smallvec` or `tinyvec`), IIRC it's basically a single file; the reason why I'm listed as a contributor is that I needed `insert` and `remove` methods that weren't there at the time, and it took me only a few minutes to include them.

niklaslong (commented):
All in favour of this change. Though a dependency with a single maintainer calls for a little caution, tuffy's other crates have been actively tended to and `singlevec` is quite small, so easily auditable. I'd be curious to see a comparison with `tinyvec` as well.

kpandl:
Nice optimization!
I'd be interested to see a comparison with `tinyvec` as well ... then we we can decide if we go with that for higher security, or if we should consider vendoring `smallvec`.

ljedrz:
Actually, `tinyvec` can't be used due to it requiring that the elements implement `Default` (which may be ok for `PrimeField`, but I don't think it's viable for `Variable`); that being said, I don't think it would be much different from `smallvec`'s performance:
```
CheckDeployment for do  time:   [935.02 ms 946.89 ms 958.37 ms]
                        change: [-27.104% -25.752% -24.542%] (p = 0.00 < 0.05)
                        Performance has improved.
```
However, I can think of one more alternative optimization here; I'll advise soon.

ljedrz:
> I can think of one more alternative optimization here; I'll advise soon.

Nope; using an arena for the vectors, while potentially viable, would require a big refactoring.

@kpandl I'm assuming you meant vendoring `singlevec`, which is perfectly viable, as opposed to `smallvec` (which we are already using anyway).

vicsn:
Thx, let's use smallvec for now

ljedrz:
Switched to `smallvec` and rebased.

vicsn (approved):

vicsn (commented):
Could `AssignmentLC.terms` also be a smallvec?

ljedrz:
> Could AssignmentLC.terms also be a smallvec?

Potentially, but such a change would work better after https://github.com/ProvableHQ/snarkVM/pull/2621 is in (as there's good potential for synergy here). I'll add it to the tracing issue, and happy to look into it once this PR and the linked one are merged.

ljedrz:
> Could AssignmentLC.terms also be a smallvec?

@vicsn Actually, I forgot that that object was refactored away already in the other PR; that being said, I can take another look at the setup now that both related PRs have been merged.

PR #2626, opened 2025-03-04
[Refactor] Move committee cache from snarkOS

ljedrz:
As requested [here](https://github.com/ProvableHQ/snarkVM/pull/2615#discussion_r1975347420); it's a simple 1:1 move.

### DIFF
--- a/ledger/src/get.rs
+++ b/ledger/src/get.rs
@@ -23,7 +23,30 @@ impl<N: Network, C: ConsensusStorage<N>> Ledger<N, C> {
 
     /// Returns the committee for the given `round`.
     pub fn get_committee_for_round(&self, round: u64) -> Result<Option<Committee<N>>> {
-        self.vm.finalize_store().committee_store().get_committee_for_round(round)
+        // Check if the committee is already in the cache.
+        if let Some(committee) = self.committee_cache.lock().get(&round) {
+            return Ok(Some(committee.clone()));
+        }
+
+        match self.vm.finalize_store().committee_store().get_committee_for_round(round)? {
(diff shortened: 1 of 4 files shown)


vicsn (approved):

vicsn (changes requested):
Can you hook up existing calls to `self.vm.finalize_store().committee_store()` already? Then it'll be easier for https://github.com/ProvableHQ/snarkVM/pull/2615 - or perhaps nothing needs be even changed there if you tuck away the calls to the cache.

And perhaps then we don't have to expose `pub fn committee_cache(&self)`

ledger/src/get.rs
```diff
@@ -23,7 +23,30 @@ impl<N: Network, C: ConsensusStorage<N>> Ledger<N, C> {
...
+            Some(committee) => {
+                // Insert the committee into the cache.
+                self.committee_cache.lock().push(round, committee.clone());
+                // Return the committee.
+                Ok(Some(committee))
+            }
+            // Return the current committee if the round is equivalent.
+            None => {
+                // Retrieve the current committee.
+                let current_committee = self.latest_committee()?;
```
vicsn, line 42: For future reviewers: this additional logic came from the snarkOS side: https://github.com/ProvableHQ/snarkOS/pull/3526/

I think at least its good that all calls to `get_committee_for_round` now use the same logic...

vicsn (approved):
LGTM.

Note that we could hypothetically get more gains if we cache based on `height` down in the committee store, but that seems like future work - when needed.

raychu86 (approved):
LGTM

PR #2627, opened 2025-03-04
[Refactor] Update deployment storage to handle future nonzero editions.

d0cd:
This PR introduces `IDMapV2` which stores the edition along with the program ID. 
The current design assumes that editions are fixed, which may not be true in the future.
The intention is to be backwards compatible with the existing design.

### Migration
This PR will need a migration height, after which `IDMapV2` will be used instead of `IDMapV1`. 
Old deployments will still live in `IDMapV1`, all which will have a zero edition.
New deployments will live in `IDMapV2`, which start with the zero edition, incrementing on each upgrade.

**TODO. Having stared at this PR in isolation, I have some ideas on how to better test. Will follow up with more commits.**

### DIFF
--- a/ledger/store/src/helpers/rocksdb/internal/id.rs
+++ b/ledger/store/src/helpers/rocksdb/internal/id.rs
@@ -105,7 +105,8 @@ pub enum CommitteeMap {
 #[derive(Clone, Copy, Debug, PartialEq, Eq, Hash)]
 #[repr(u16)]
 pub enum DeploymentMap {
-    ID = DataID::DeploymentIDMap as u16,
+    IDV1 = DataID::DeploymentIDV1Map as u16,
+    IDV2 = DataID::DeploymentIDV2Map as u16,
     Edition = DataID::DeploymentEditionMap as u16,
     ReverseID = DataID::DeploymentReverseIDMap as u16,
     Owner = DataID::DeploymentOwnerMap as u16,
@@ -248,7 +249,7 @@ enum DataID {
     RoundToHeightMap,
     CommitteeMap,
     // Deployment
-    DeploymentIDMap,
+    DeploymentIDV1Map,
     DeploymentEditionMap,
     DeploymentReverseIDMap,
     DeploymentOwnerMap,
@@ -305,4 +306,7 @@ enum DataID {
     Test4,
     #[cfg(test)]
     Test5,
+
+    // Additional
+    DeploymentIDV2Map,
 }
--- a/ledger/store/src/transaction/deployment.rs
+++ b/ledger/store/src/transaction/deployment.rs
@@ -37,7 +37,9 @@ use std::borrow::Cow;
 /// A trait for deployment storage.
 pub trait DeploymentStorage<N: Network>: Clone + Send + Sync {
     /// The mapping of `transaction ID` to `program ID`.
-    type IDMap: for<'a> Map<'a, N::TransactionID, ProgramID<N>>;
+    type IDMapV1: for<'a> Map<'a, N::TransactionID, ProgramID<N>>;
+    /// The mapping of `transaction ID` to `(program ID, edition)`.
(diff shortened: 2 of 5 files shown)


ledger/store/src/helpers/rocksdb/internal/id.rs
```diff
@@ -305,4 +306,7 @@ enum DataID {
     Test4,
     #[cfg(test)]
     Test5,
+
+    // Additional
+    DeploymentIDV2Map,
```
ljedrz, line 311: nit: this new variant could also be above the `Test` ones, as those are only ever created in tests - never in the production database

ledger/store/src/transaction/deployment.rs
```diff
@@ -228,12 +241,27 @@ pub trait DeploymentStorage<N: Network>: Clone + Send + Sync {
...
         atomic_batch_scope!(self, {
             // Remove the program ID.
-            self.id_map().remove(transaction_id)?;
-            // Remove the edition.
-            self.edition_map().remove(&program_id)?;
+            if in_v1_id_map {
+                self.id_map_v1().remove(transaction_id)?;
+            } else {
+                self.id_map_v2().remove(transaction_id)?;
+            }
```
ljedrz, line 253: perhaps we could just call remove on both maps? we're already going to need an atomic write, and this would avoid a separate lookup

ljedrz (approved):
LGTM :+1:; I understand that the current number of entries in the legacy map (i.e. the performance impact) is low enough that we don't need to do a migration? Because it would be slightly trickier to do it later on.

ledger/store/src/transaction/deployment.rs
```diff
@@ -278,6 +305,25 @@ pub trait DeploymentStorage<N: Network>: Clone + Send + Sync {
...
 
+    /// Returns the transaction ID that contains the given `program ID` and `edition`.
+    fn find_transaction_id_from_program_id_and_edition(
+        &self,
+        program_id: &ProgramID<N>,
+        edition: u16,
+    ) -> Result<Option<N::TransactionID>> {
+        // Check if the pogram ID is for `credits.aleo`.
+        // This case is handled separately, as it is a default program of the VM.
+        // TODO: Note on removal in `find_transaction_id_from_program_id`.
```
vicsn, line 316: TODO left...

ledger/store/src/transaction/deployment.rs
```diff
@@ -211,15 +223,16 @@ pub trait DeploymentStorage<N: Network>: Clone + Send + Sync {
         })
     }
 
+    /// TODO (@d0cd) After the migration only the V2 map is used, so do we need to handle the old map?
```
vicsn, line 226: Can you explain any migration plan and steps in the PR description?

ledger/store/src/transaction/deployment.rs
```diff
@@ -184,9 +195,10 @@ pub trait DeploymentStorage<N: Network>: Clone + Send + Sync {
         let program_id = *program.id();
 
         atomic_batch_scope!(self, {
-            // Store the program ID.
-            self.id_map().insert(*transaction_id, program_id)?;
-            // Store the edition.
+            // Store the program ID and edition.
+            // Note that the `id_map_v1` is intentionally excluded.
```
vicsn, line 199: Despite there being follow up PRs, might be nice to add a similarly explicit note about id_map_v1 being deprecated in favour of id_map_v2.

ledger/store/src/transaction/deployment.rs
d0cd: This was intended to be reference to a prior TODO in the above method
I updated the language to make that more clear

d0cd:
I'm closing this PR because I think the order of operations is wrong. 
We're going to want to make the migration more explicit around V1 and V2 programs.
The PR chain will be updated accordingly.

PR #2628, opened 2025-03-04
[Refactor] Global stack map.

d0cd:

This PR removes `external_stacks` from `Stack` and references all stacks through the global stack map in `Process`.
This significantly reduces bookkeeping needed to keep `Process` consistent as stacks are modified or evicted from memory.
The changes made in this PR should be semantically equivalent to the prior design.
While the weak reference from `Stack` to `Process` may have some overhead, it should not introduce memory leaks.

## An example

Suppose we have the following programs:
- `one.aleo`
- `two.aleo`, which imports `one.aleo`
- `three.aleo`, which imports `one.aleo` and `two.aleo`
- `four.aleo`, which imports `one.aleo`, `two.aleo` and `three.aleo`

The existing `Process` and `Stack`s are organized in this way in memory:
image

The new design is organized in this way:
image

_The blue arrows are `Arc`s and the red arrows are `Weak`s._

## Profiling
-  A [memory profile](https://private-user-images.githubusercontent.com/3750347/419973700-2957decd-6064-4c3c-baad-0f8a5216bd42.png?jwt=eyJhbGciOiJIUzI1NiIsInR5cCI6IkpXVCJ9.eyJpc3MiOiJnaXRodWIuY29tIiwiYXVkIjoicmF3LmdpdGh1YnVzZXJjb250ZW50LmNvbSIsImtleSI6ImtleTUiLCJleHAiOjE3NDE3Mjc3MTQsIm5iZiI6MTc0MTcyNzQxNCwicGF0aCI6Ii8zNzUwMzQ3LzQxOTk3MzcwMC0yOTU3ZGVjZC02MDY0LTRjM2MtYmFhZC0wZjhhNTIxNmJkNDIucG5nP1gtQW16LUFsZ29yaXRobT1BV1M0LUhNQUMtU0hBMjU2JlgtQW16LUNyZWRlbnRpYWw9QUtJQVZDT0RZTFNBNTNQUUs0WkElMkYyMDI1MDMxMSUyRnVzLWVhc3QtMSUyRnMzJTJGYXdzNF9yZXF1ZXN0JlgtQW16LURhdGU9MjAyNTAzMTFUMjExMDE0WiZYLUFtei1FeHBpcmVzPTMwMCZYLUFtei1TaWduYXR1cmU9MDZiMzM1M2RmZjM3YjkwNDYzOWYxNjJjMTg1ZGY4MjBhOTg0YWE3ODQ5ODVkNzE2MjMyZDNjZTY5MDhiOTY3ZCZYLUFtei1TaWduZWRIZWFkZXJzPWhvc3QifQ.3Uia6YdTLAQO5EOK7aAzg7COHgu01g5-IJSlWDPrKMo) was done to ensure that a leak does not exist

### DIFF
--- a/synthesizer/process/src/finalize.rs
+++ b/synthesizer/process/src/finalize.rs
@@ -230,13 +230,12 @@ fn finalize_transition<N: Network, P: FinalizeStorage<N>>(
         states.pop()
     {
         // Get the finalize logic.
-        let finalize = match stack.get_function_ref(registers.function_name())?.finalize_logic() {
-            Some(finalize) => finalize,
-            None => bail!(
+        let Some(finalize) = stack.get_function_ref(registers.function_name())?.finalize_logic() else {
+            bail!(
                 "The function '{}/{}' does not have an associated finalize block",
                 stack.program_id(),
                 registers.function_name()
-            ),
+            )
         };
         // Evaluate the commands.
         while counter < finalize.commands().len() {
@@ -245,13 +244,8 @@ fn finalize_transition<N: Network, P: FinalizeStorage<N>>(
             // Finalize the command.
             match &command {
                 Command::BranchEq(branch_eq) => {
-                    let result = try_vm_runtime!(|| branch_to(
-                        counter,
-                        branch_eq,
-                        finalize.positions(),
-                        stack.deref(),
-                        &registers
-                    ));
+                    let result =
+                        try_vm_runtime!(|| branch_to(counter, branch_eq, finalize.positions(), &stack, &registers));
                     match result {
                         Ok(Ok(new_counter)) => {
                             counter = new_counter;
@@ -263,13 +257,8 @@ fn finalize_transition<N: Network, P: FinalizeStorage<N>>(
                     }
                 }
                 Command::BranchNeq(branch_neq) => {
-                    let result = try_vm_runtime!(|| branch_to(
-                        counter,
-                        branch_neq,
-                        finalize.positions(),
-                        stack.deref(),
-                        &registers
-                    ));
+                    let result =
+                        try_vm_runtime!(|| branch_to(counter, branch_neq, finalize.positions(), &stack, &registers));
                     match result {
                         Ok(Ok(new_counter)) => {
                             counter = new_counter;
--- a/synthesizer/process/src/finalize.rs
+++ b/synthesizer/process/src/finalize.rs
@@ -558,7 +558,7 @@ function compute:
         let (stack, _) =
             process.finalize_deployment(sample_finalize_state(1), &finalize_store, &deployment, &fee).unwrap();
         // Add the stack *manually* to the process.
-        process.add_stack(stack).unwrap();
+        process.add_stack(stack);
 
         // Ensure the program exists.
         assert!(process.contains_program(program.id()));
--- a/synthesizer/process/src/finalize.rs
+++ b/synthesizer/process/src/finalize.rs
@@ -404,9 +404,9 @@ fn initialize_finalize_state<N: Network>(
     nonce: u64,
 ) -> Result<FinalizeState<N>> {
     // Get the stack.
-    let stack = match stack.get_external_stack(future.program_id()) {
-        Ok(stack) => stack,
-        Err(_) => stack.clone(),
+    let stack = match stack.program_id() == future.program_id() {
+        true => stack.clone(),
+        false => stack.get_external_stack(future.program_id())?,
     };
     // Get the finalize logic and check that it exists.
     let finalize = match stack.get_function_ref(future.function_name())?.finalize_logic() {
--- a/synthesizer/process/src/finalize.rs
+++ b/synthesizer/process/src/finalize.rs
@@ -54,7 +54,7 @@ impl<N: Network> Process<N> {
             // Retrieve the fee stack.
             let fee_stack = self.get_stack(fee.program_id())?;
             // Finalize the fee transition.
-            finalize_operations.extend(finalize_fee_transition(state, store, fee_stack, fee)?);
+            finalize_operations.extend(finalize_fee_transition(state, store, &fee_stack, fee)?);
             lap!(timer, "Finalize transition for '{}/{}'", fee.program_id(), fee.function_name());
 
             /* Finalize the deployment. */
@@ -116,7 +116,7 @@ impl<N: Network> Process<N> {
             // Finalize the root transition.
             // Note that this will result in all the remaining transitions being finalized, since the number
             // of calls matches the number of transitions.
-            let mut finalize_operations = finalize_transition(state, store, stack, transition, call_graph)?;
+            let mut finalize_operations = finalize_transition(state, store, &stack, transition, call_graph)?;
 
             /* Finalize the fee. */
 
@@ -124,7 +124,7 @@ impl<N: Network> Process<N> {
                 // Retrieve the fee stack.
                 let fee_stack = self.get_stack(fee.program_id())?;
                 // Finalize the fee transition.
-                finalize_operations.extend(finalize_fee_transition(state, store, fee_stack, fee)?);
+                finalize_operations.extend(finalize_fee_transition(state, store, &fee_stack, fee)?);
                 lap!(timer, "Finalize transition for '{}/{}'", fee.program_id(), fee.function_name());
             }
 
@@ -150,7 +150,7 @@ impl<N: Network> Process<N> {
             // Retrieve the stack.
             let stack = self.get_stack(fee.program_id())?;
             // Finalize the fee transition.
-            let result = finalize_fee_transition(state, store, stack, fee);
+            let result = finalize_fee_transition(state, store, &stack, fee);
             finish!(timer, "Finalize transition for '{}/{}'", fee.program_id(), fee.function_name());
             // Return the result.
             result
@@ -162,7 +162,7 @@ impl<N: Network> Process<N> {
 fn finalize_fee_transition<N: Network, P: FinalizeStorage<N>>(
     state: FinalizeGlobalState,
     store: &FinalizeStore<N, P>,
-    stack: &Stack<N>,
+    stack: &Arc<Stack<N>>,
     fee: &Fee<N>,
 ) -> Result<Vec<FinalizeOperation<N>>> {
     // Construct the call graph.
@@ -187,7 +187,7 @@ fn finalize_fee_transition<N: Network, P: FinalizeStorage<N>>(
 fn finalize_transition<N: Network, P: FinalizeStorage<N>>(
     state: FinalizeGlobalState,
     store: &FinalizeStore<N, P>,
-    stack: &Stack<N>,
+    stack: &Arc<Stack<N>>,
     transition: &Transition<N>,
     call_graph: HashMap<N::TransitionID, Vec<N::TransitionID>>,
 ) -> Result<Vec<FinalizeOperation<N>>> {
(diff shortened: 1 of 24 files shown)


synthesizer/process/src/lib.rs
```diff
@@ -124,17 +124,22 @@ impl<N: Network> Process<N> {
...
         Ok(())
     }
 
     /// Adds a new stack to the process.
     /// If you intend to `execute` the program, use `deploy` and `finalize_deployment` instead.
     #[inline]
-    pub fn add_stack(&mut self, stack: Stack<N>) {
-        // Add the stack to the process.
-        self.stacks.insert(*stack.program_id(), Arc::new(stack));
+    pub fn add_stack(&mut self, stack: Stack<N>) -> Result<()> {
```
ljedrz:
```suggestion
    pub fn add_stack(&mut self, stack: Stack<N>) {
```
the body is infallible

synthesizer/process/src/lib.rs
```diff
@@ -90,7 +89,8 @@ impl<N: Network> Process<N> {
         let timer = timer!("Process:setup");
 
         // Initialize the process.
-        let mut process = Self { universal_srs: Arc::new(UniversalSRS::load()?), stacks: IndexMap::new() };
+        let mut process =
+            Self { universal_srs: Arc::new(UniversalSRS::load()?), stacks: Arc::new(RwLock::new(IndexMap::new())) };
```
ljedrz:
```suggestion
            Self { universal_srs: Arc::new(UniversalSRS::load()?), stacks: Default::default() };
```

synthesizer/process/src/lib.rs
```diff
@@ -145,7 +150,8 @@ impl<N: Network> Process<N> {
         let timer = timer!("Process::load");
 
         // Initialize the process.
-        let mut process = Self { universal_srs: Arc::new(UniversalSRS::load()?), stacks: IndexMap::new() };
+        let mut process =
+            Self { universal_srs: Arc::new(UniversalSRS::load()?), stacks: Arc::new(RwLock::new(IndexMap::new())) };
```
ljedrz:
```suggestion
            Self { universal_srs: Arc::new(UniversalSRS::load()?), stacks: Default::default() };
```

synthesizer/process/src/lib.rs
```diff
@@ -183,7 +189,8 @@ impl<N: Network> Process<N> {
     #[cfg(feature = "wasm")]
     pub fn load_web() -> Result<Self> {
         // Initialize the process.
-        let mut process = Self { universal_srs: Arc::new(UniversalSRS::load()?), stacks: IndexMap::new() };
+        let mut process =
+            Self { universal_srs: Arc::new(UniversalSRS::load()?), stacks: Arc::new(RwLock::new(IndexMap::new())) };
```
ljedrz:
```suggestion
            Self { universal_srs: Arc::new(UniversalSRS::load()?), stacks: Default::default() };
```

synthesizer/process/src/stack/mod.rs
```diff
@@ -324,42 +324,22 @@ impl<N: Network> StackProgram<N> for Stack<N> {
...
+        let result = self
+            .stacks
+            .upgrade()
+            .ok_or_else(|| anyhow!("Process-level stack map does not exist"))?
+            .read()
+            .get(program_id)
+            .cloned()
+            .ok_or_else(|| anyhow!("External stack for '{program_id}' does not exist"));
+        // Return the external stack.
+        result
```
ljedrz:
```suggestion
        self
            .stacks
            .upgrade()
            .ok_or_else(|| anyhow!("Process-level stack map does not exist"))?
            .read()
            .get(program_id)
            .cloned()
            .ok_or_else(|| anyhow!("External stack for '{program_id}' does not exist"))
```
nit: it's only a matter of time before `clippy` complains about an immediate return

ljedrz (commented):
I've left a few comments, but I'm in favor of these changes in general. Is there some specific  benchmark where the memory impact could be measured? I'd be happy to run the numbers.

synthesizer/process/src/stack/mod.rs
```diff
@@ -175,8 +175,8 @@ impl<N: Network> CallStack<N> {
 pub struct Stack<N: Network> {
     /// The program (record types, structs, functions).
     program: Program<N>,
-    /// The mapping of external stacks as `(program ID, stack)`.
-    external_stacks: IndexMap<ProgramID<N>, Arc<Stack<N>>>,
+    /// A reference to the global stack map.
+    stacks: Weak<RwLock<IndexMap<ProgramID<N>, Arc<Stack<N>>>>>,
```
vicsn, line 179: Is it possible for the Stack not to hold any reference to the global set of Stacks, and instead pass in relevant imports for the respective functions which need them (notably `initialize`)? Might be a cleaner design, and at least in the context of storing stacks to disk it would have made things easier - not sure if this still holds after program upgradability.

d0cd: I explored this idea initially, but it involves a larger refactor. 
Basically, you'd end up passing `Process` around wherever you call `get_external_stack`. 
Some places like `initialize` are straightforward, however you get into really hairy territory when you get to `call.rs`, where the stack is trait-ified.

d0cd:
@ljedrz I have a branch [here](https://github.com/ProvableHQ/aleo-program-regressions/pull/6) where I test out building a `Process` on all of the program on `mainnet` and `testnet`. It could be helpful in sniffing out performance issues and memory leaks. Note that it currently points to a rev for [this](https://github.com/ProvableHQ/snarkVM/pull/2629) PR

synthesizer/process/src/lib.rs
```diff
@@ -133,8 +132,12 @@ impl<N: Network> Process<N> {
...
     #[inline]
     pub fn add_stack(&mut self, stack: Stack<N>) {
-        // Add the stack to the process.
-        self.stacks.insert(*stack.program_id(), Arc::new(stack));
+        // Get the program ID.
+        let program_id = *stack.program_id();
+        // Acquire the write lock.
+        let mut stacks = self.stacks.write();
+        // Insert the stack into the process, replacing the existing stack if it exists.
+        stacks.insert(program_id, Arc::new(stack));
```
ljedrz:
```suggestion
        // Acquire an Arced instance of a stack in advance to limit the scope of the write lock.
        let stack = Arc::new(stack);
        // Insert the stack into the process, replacing the existing stack if it exists.
        self.stacks.write().insert(program_id, stack);
```

ljedrz: (since the creation of an `Arc` involves a heap allocation)

synthesizer/process/src/finalize.rs
```diff
@@ -361,39 +382,34 @@ fn finalize_transition<N: Network, P: FinalizeStorage<N>>(
...
-    // Get the finalize logic and the stack.
-    let (finalize, stack) = match stack.program_id() == future.program_id() {
-        true => (stack.get_function_ref(future.function_name())?.finalize_logic(), stack),
-        false => {
-            let stack = stack.get_external_stack(future.program_id())?.as_ref();
-            (stack.get_function_ref(future.function_name())?.finalize_logic(), stack)
-        }
+) -> Result<FinalizeState<N>> {
+    // Get the stack.
+    let stack = match stack.get_external_stack(future.program_id()) {
```
vicsn: Since we're not checking if `stack.program_id() == future.program_id()` anymore, should we keep it as a lazy defense-in-depth `ensure!`?

synthesizer/process/src/lib.rs
```diff
@@ -207,26 +210,22 @@ impl<N: Network> Process<N> {
...
     }
 
     /// Returns the stack for the given program ID.
     #[inline]
-    pub fn get_stack(&self, program_id: impl TryInto<ProgramID<N>>) -> Result<&Arc<Stack<N>>> {
+    pub fn get_stack(&self, program_id: impl TryInto<ProgramID<N>>) -> Result<Arc<Stack<N>>> {
         // Prepare the program ID.
         let program_id = program_id.try_into().map_err(|_| anyhow!("Invalid program ID"))?;
+        // Acquire the read lock.
+        let stacks = self.stacks.read();
```
vicsn: Is there any chance that the reads happen concurrently using e.g. rayon?

@kpandl found that my recent attempt at a stacks cache [resulted in a deadlock](https://github.com/ProvableHQ/snarkVM/pull/2591#issuecomment-2656749691) - the cause still needs to be investigated but one potential reason could be https://github.com/ProvableHQ/snarkOS/pull/3440/commits/970386ff51cecdd47005d76dc60c4d41e2732dc2

@kpandl would be great if you can run a stress test with the same setup against this PR, you can let it run for half a day, and if there is no halt, increase the request pressure and we try again for half a day.

synthesizer/process/src/finalize.rs
d0cd: Good call out, I think it is safer and clearer to go with:
```
let stack = match stack.program_id() == future.program_id() {
        true => stack.clone(),
        false => stack.get_external_stack(future.program_id())?,
    };
```

synthesizer/process/src/lib.rs
d0cd: There's a broader question on what is the expected concurrency model of `Process`.

I am assuming that there is a main thread which has R/W privilege and peripheral threads with R privilege.
One concern could be that the main thread runs into a read-write deadlock, but we've kept lock scopes minimal to avoid that. Perhaps the assumptions are wrong.

Curious to see the results of the stress test.

vicsn: Just to clarify what I think happened in https://github.com/ProvableHQ/snarkOS/commit/970386ff51cecdd47005d76dc60c4d41e2732dc2 in order: 1. rayon spawns a number of threads which want to acquire the read-lock 2. the main thread wants to acquire the write-lock 3. another aforementioned rayon thread wants to acquire a new read-lock. However, rayon will only end all threads and release all read-locks atomically, so this new read-lock can't be acquired nor released, nor can the previous write-lock.

d0cd: That's helpful context! I believe we are using [parking_lot::RwLock](https://docs.rs/parking_lot/0.12.3/parking_lot/type.RwLock.html) which should prevent writer starvation. Perhaps there's a subtle way this is getting triggered.

synthesizer/process/src/lib.rs
```diff
@@ -80,7 +79,7 @@ pub struct Process<N: Network> {
     /// The universal SRS.
     universal_srs: Arc<UniversalSRS<N>>,
     /// The mapping of program IDs to stacks.
-    stacks: IndexMap<ProgramID<N>, Arc<Stack<N>>>,
+    stacks: Arc<RwLock<IndexMap<ProgramID<N>, Arc<Stack<N>>>>>,
```
ljedrz, line 82: do we actually need an `RwLock` here? I only see a single call to `write`, and it's in the context of `&mut self`

ljedrz: ah, but the entire collection is now `Arc`ed; I guess that's fine if we expect it to be large

ljedrz (approved):
LGTM :ok_hand:

niklaslong (approved):

synthesizer/process/src/lib.rs
kpandl: Ran the tests, didn't find any issue with this branch.

For reference: I was running a light network to test it. First with the same settings where I found the past deadlock, then with more challenging settings (4x the number of deployed nested programs, faster program probe). [[2nd run on Grafana]](https://aleostresstest.grafana.net/d/single-region-tests-kp/single-region-tests-kp?from=2025-03-11T14:56:58.191Z&to=2025-03-12T09:46:46.871Z&refresh=5s&timezone=browser&orgId=1)

vicsn (approved):
LGTM! Gladded the stress test passed.

synthesizer/process/src/stack/mod.rs
raychu86: If cleanliness is the main factor here, then a helper function wrapping `RwLock<IndexMap<ProgramID<N>, Arc<Stack<N>>>` would be better.

I assume the `Weak` reference is to avoid cycles?

raychu86 (approved):
As long as tests been run to ensure nodes can still sync blocks and bootup correctly, this PR overall looks good to me.

PR #2629, opened 2025-03-05
[Redesign] Remove cached program metadata and add protections against large recursion.

d0cd:

The `Stack` caches information about a program including the number of calls, the finalize cost, and the program depth. If a program is updated, this information is invalid. 

This PR:
- removes the cached data and opts to compute the information on the fly. 
- adds protections to ensure that runtimes for the calculations are bounded.
- uses an iterative method to load a VM from storage. (See **Performance** section)
The resulting code should be **semantically** equivalent.

## Performance

See [this](https://github.com/ProvableHQ/aleo-program-regressions/tree/master/regressions/profile_process_load_from_store) regression test to evaluate the performance of loading a VM from DB on `canary`, `testnet`, and `mainnet`. In general, these improvements performed comparably or better.

## Notes

The integration expectations change for the following reasons:
- `user_callable.out`, `program_callable.out`. Fixed formatting of `assert.eq`
- `read_external_mapping.out`. By removing the cache, the finalize cost is updated after the test migration height is crossed.
- `branch_with_future.out`, `complex_finalization.out`, `future_out_of_order.out`. These expectations are also different on `staging` but were not committed earlier. They do not trigger a CI failure because `child_outputs` are not checked to be the same.


synthesizer/process/src/cost.rs
```diff
@@ -404,60 +407,60 @@ pub fn cost_per_command<N: Network>(
...
+        // Note that one transition is reserved for the fee.
+        ensure!(
+            num_finalizes < Transaction::<N>::MAX_TRANSITIONS,
+            "The number of finalize blocks must be less than '{}'",
+            Transaction::<N>::MAX_TRANSITIONS
+        );
+        // Get the finalize logic. If the function does not have a finalize scope then no cost is incurred.
+        if let Some(finalize) = match &stack_ref {
+            StackRef::Internal(stack_ref) => stack_ref.get_function_ref(&function_name)?.finalize_logic(),
+            StackRef::External(stack_ref) => stack_ref.get_function_ref(&function_name)?.finalize_logic(),
```
ljedrz: it would be best if we could avoid separate `match` branches here, as their logic is the same

synthesizer/process/src/stack/authorization/mod.rs
```diff
@@ -145,8 +145,16 @@ impl<N: Network> Authorization<N> {
     }
 
     /// Appends the given `Request` to the authorization.
-    pub fn push(&self, request: Request<N>) {
+    pub fn push(&self, request: Request<N>) -> Result<()> {
+        // Check that the number of requests is less than the maximum.
+        ensure!(
+            self.len() < Transaction::<N>::MAX_TRANSITIONS,
```
ljedrz, line 151: shouldn't it be `=<`? `MAX_` implies "no more than"

synthesizer/process/src/stack/mod.rs
```diff
@@ -135,9 +135,18 @@ impl<N: Network> CallStack<N> {
             CallStack::Authorize(requests, ..)
             | CallStack::Synthesize(requests, ..)
             | CallStack::CheckDeployment(requests, ..)
-            | CallStack::PackageRun(requests, ..) => requests.push(request),
-            CallStack::Evaluate(authorization) => authorization.push(request),
-            CallStack::Execute(authorization, ..) => authorization.push(request),
+            | CallStack::PackageRun(requests, ..) => {
+                // Check that the number of requests does not exceed the maximum.
+                ensure!(
+                    requests.len() < Transaction::<N>::MAX_TRANSITIONS,
```
ljedrz, line 141: ditto

synthesizer/process/src/stack/mod.rs
```diff
@@ -363,10 +351,47 @@ impl<N: Network> StackProgram<N> for Stack<N> {
...
+        // Initialize the base number of calls.
+        let mut num_calls = 1;
+        // Initialize a queue of functions to check.
+        let mut queue = vec![(StackRef::Internal(self), *function_name)];
+        // Iterate over the queue.
+        while let Some((stack_ref, function_name)) = queue.pop() {
+            // Ensure that the number of calls does not exceed the maximum.
+            // Note that one transition is reserved for the fee.
+            ensure!(
+                num_calls < Transaction::<N>::MAX_TRANSITIONS,
```
ljedrz, line 363: ditto

synthesizer/process/src/stack/mod.rs
```diff
@@ -363,10 +351,47 @@ impl<N: Network> StackProgram<N> for Stack<N> {
...
+                "Number of calls must be less than '{}'",
+                Transaction::<N>::MAX_TRANSITIONS
+            );
+            // Determine the number of calls for the function.
+            for instruction in match &stack_ref {
+                StackRef::Internal(stack) => stack.get_function_ref(&function_name)?.instructions(),
+                StackRef::External(stack) => stack.get_function_ref(&function_name)?.instructions(),
```
ljedrz: as before, `StackRef::Internal(stack) | StackRef::External(stack) => stack.get_function_ref(&function_name)?.instructions()`, or avoiding a `match` altogether would be preferable due to common logic

ljedrz: seems like the `Deref` impl would work here

ljedrz (commented):
I'm not too familiar with this particular logic; left a few comments.

synthesizer/tests/expectations/vm/execute_and_finalize/read_external_mapping.out
```diff
@@ -92,4 +92,4 @@ additional:
 - child_outputs:
     credits.aleo/fee_public:
       outputs:
-      - '{"type":"future","id":"8401929608727184541382889469312440652682460087199782119730570970943355959220field","value":"{\n  program_id: credits.aleo,\n  function_name: fee_public,\n  arguments: [\n    aleo1xe2fps8f9xpdas2q0fqy22uraenk84tvvzetrsyxgnwy6445h59s6wv78x,\n    12331u64\n  ]\n}"}'
+      - '{"type":"future","id":"4369745470310636653607709784601243833431868722463490242434780063887791579010field","value":"{\n  program_id: credits.aleo,\n  function_name: fee_public,\n  arguments: [\n    aleo1xe2fps8f9xpdas2q0fqy22uraenk84tvvzetrsyxgnwy6445h59s6wv78x,\n    3831u64\n  ]\n}"}'
```
vicsn, line 95: Why did `12331u64` change to `3831u64`?

synthesizer/src/vm/verify.rs
```diff
@@ -235,6 +235,12 @@ impl<N: Network, C: ConsensusStorage<N>> VM<N, C> {
                         } else {
                             execution_cost_v2(&self.process().read(), execution)?
                         };
+                        // Ensure the cost does not exceed the transaction spend limit.
+                        ensure!(
```
vicsn, line 239: For the interested reader: this check was originally done in `Stack::initialize` and is still there. Ergo, we don't have to regression test this.

synthesizer/process/src/tests/test_execute.rs
```diff
@@ -2528,7 +2528,8 @@ fn test_long_import_chain() {
     ))
     .unwrap();
     let result = process.add_program(&program);
-    assert!(result.is_err());
+    // Programs may create long import chains as long as number of calls does not exceed the maximum number of transitions.
+    assert!(result.is_ok());
```
vicsn, line 2531: For other reviewers, my understanding: because the check on `MAX_PROGRAM_DEPTH` in `Stack::initialize` is removed, we no longer fail at this level of abstraction. However, equivalent checks against maximum transitions and imports still exist when transactions and blocks are verified.

console/network/src/lib.rs
```diff
@@ -194,6 +194,7 @@ pub trait Network:
     const MAX_OUTPUTS: usize = 16;
 
     /// The maximum program depth.
+    /// This restriction is **not** in use.
```
vicsn: IMO you can remove.

synthesizer/process/src/stack/mod.rs
```diff
@@ -363,10 +351,44 @@ impl<N: Network> StackProgram<N> for Stack<N> {
...
-            .ok_or_else(|| anyhow!("Function '{function_name}' does not exist"))
+        // Initialize the base number of calls.
+        let mut num_calls = 1;
+        // Initialize a queue of functions to check.
+        let mut queue = vec![(StackRef::Internal(self), *function_name)];
+        // Iterate over the queue.
+        while let Some((stack_ref, function_name)) = queue.pop() {
+            // Ensure that the number of calls does not exceed the maximum.
+            // Note that one transition is reserved for the fee.
+            ensure!(
```
vicsn, line 370: I think it can be safely checked once after the while loop.

vicsn (commented):
Love the cleanup!

synthesizer/tests/expectations/vm/execute_and_finalize/read_external_mapping.out
d0cd: This transition is executed after the test migration height, so the fees are reduced.

synthesizer/process/src/stack/mod.rs
d0cd: I believe it needs to happen on each loop iteration to be safe.
Otherwise, you may execute for an unnecessarily long time (possibly indefinitely) when traversing large (or cyclic) call graphs.

synthesizer/process/src/stack/mod.rs
```diff
@@ -468,3 +490,23 @@ impl<N: Network> PartialEq for Stack<N> {
 }
 
 impl<N: Network> Eq for Stack<N> {}
+
+// A helper enum to avoid cloning stacks.
+#[derive(Clone)]
+pub(crate) enum StackRef<'a, N: Network> {
```
ljedrz, line 504: Hmm, I think we could just use `Arc<Stack<N>>` instead? Cloning an `Arc` is an atomic operation, and while it has some runtime penalty, it's a tiny one (not unlike using an `enum` as this one).

vicsn (approved):

console/network/src/lib.rs
d0cd: Should `MAX_PROGRAM_DEPTH` be removed entirely? Otherwise, it could be helpful for readers to know that it's not relevant.

synthesizer/process/src/stack/mod.rs
d0cd: In `cost.rs` and `get_number_of_calls` we wrap a `&Stack` in `StackRef::Internal` to avoid cloning.
We use the `StackRef` to persist an `&Stack` or `Arc<Stack>` across loop iterations.
Any ideas on how to avoid that pattern? If we have an alternative solution, then `StackRef` is no longer necessary.

ljedrz: it seems to me that we could pass `&Arc<Stack>` to the cost functions and call `clone` on it when necessary, and it might be possible for `get_number_of_calls` to have a signature of
```
fn get_number_of_calls(self: &Arc<Self>, function_name: &Identifier<N>)
```
so it can be shallowly cloned as well

d0cd: Hmm, do you forsee any issues with a change like that? 
I feel like the `Arc` is less general than the reference.

ljedrz: no, I'd only expect it to be simpler (i.e. no extra `StackRef` object); also, `Arc<T>` dereferences to `T`, so you can always obtain a simple reference from it

edit: this assumes that the cost calculation and `get_number_of_calls` is called in contexts where we already have an `Arc` handy - otherwise it would indeed not work

kpandl (changes requested):
EDIT: After applying @ljedrz recent PR that fixes `staging`, I'm now testing this branch for deadlocks ...

I ran a stresstest on the `feat/cycle-protections` snarkVM branch (commit `2250f82`, based of snarkOS branch staging commit `4d4805796036709c367add54e3ab31f5900fa196`), but it fails to produce blocks, also just on a simple local devnet run:

```
2025-03-12T14:49:55.112773Z ERROR Unable to advance to the next block - Certificate '3259985370026459528435539507197131813256664066034522549222049650037512599165field' for round 1 does not meet quorum requirements
2025-03-12T14:49:55.112839Z  WARN Unable to reinsert transmission at1eur0kqnp6c5kj...[2m2716236445986438..[0m into the memory pool - Transaction 'at1eur0kqnp6c5kj...[2m2716236445986438..[0m' already exists.
2025-03-12T14:49:55.112890Z  WARN Unable to reinsert transmission at1zlfafrjz70mu3...[2m1199453133453965..[0m into the memory pool - Transaction 'at1zlfafrjz70mu3...[2m1199453133453965..[0m' already exists.
2025-03-12T14:49:55.112950Z  WARN Unable to reinsert transmission at1uqs8kztr6gvk2...[2m1750320161395995..[0m into the memory pool - Transaction 'at1uqs8kztr6gvk2...[2m1750320161395995..[0m' already exists.
2025-03-12T14:49:55.112998Z  WARN Unable to reinsert transmission at1ugcmj42xwe4ql...[2m3764518595941297..[0m into the memory pool - Transaction 'at1ugcmj42xwe4ql...[2m3764518595941297..[0m' already exists.
2025-03-12T14:49:55.113003Z ERROR BFT failed to advance the subdag for round 2 - Certificate '3259985370026459528435539507197131813256664066034522549222049650037512599165field' for round 1 does not meet quorum requirements
```

Is there something I should change to the test?

ljedrz:
@kpandl i think the certification issue is already in `staging`, I'm already dissecting it.

synthesizer/src/vm/mod.rs
```diff
@@ -109,81 +112,71 @@ impl<N: Network, C: ConsensusStorage<N>> VM<N, C> {
...
-                    // Load the deployment and its imports.
-                    load_deployment_and_imports(&process, transaction_store, **transaction_id)
-                })
-                .collect::<Result<Vec<_>>>()?;
-
-            for (program_id, deployment) in deployments.iter().flatten() {
-                // Load the deployment if it does not exist in the process yet.
-                if !process.contains_program(program_id) {
-                    process.load_deployment(deployment)?;
+            // Load the deployments and their block heights.
```
raychu86: We updated program deployment loading here - https://github.com/ProvableHQ/snarkVM/pull/2644.

I am unsure how much overhead the additional block height lookups will add. But, should be minimal since it's only 2 db lookups + deserializations per deployment.

There might be a combination of approaches that may be better, but will require some benchmarking. The sorted-by-block-height method should alleviate issues, but won't necessarily solve it outright.

Outside of speed, is there a reason to do this sorted version? Is it required for cyclic program imports? Will need to consider based on how upgraded program stacks are handled.

raychu86: Also the final deployment loading in your case is not parallelized

raychu86: I am fine with any approach as long as it is performant and clean.

raychu86: Also edge case check: There can't be 2 deployments in a block where one deployment imports the other right? That could break your iterative approach here.

The naive case could be fine, but just checking in the case of malicious validators injecting both deployments into a proposal.

console/network/src/lib.rs
raychu86: For clarity, `MAX_PROGRAM_DEPTH` is no longer relevant because upgradability can introduce import cycles. However, we are still bounded on the function level with MAX_TRANSITIONS. Is that correct?

If it is, then we can remove `MAX_PROGRAM_DEPTH` as long as we can guarantee there is no deployment rejected because of this rule.

synthesizer/src/vm/mod.rs
d0cd: The recursive depth of `load_deployment_from_imports` is dependent on the import graph. If the depth is too large, then you have have a stack overflow. While we currently have a `MAX_PROGRAM_DEPTH`, enforcing this restriction in the presence of program updates will be challenging. The iterative method is an attempt at a safer approach.

There can't be dependent deployments in the same block. This is because deployment verification will initialize a `Stack` and check that all imports exist for all deployments prior to finalizing them. I added a test case in the latest commit to confirm.

kpandl (approved):
LGTM, no deadlock / halt found: [Grafana](https://aleostresstest.grafana.net/d/single-region-tests-kp/single-region-tests-kp?from=2025-03-12T16:43:21.621Z&to=2025-03-13T09:25:05.969Z&refresh=5s&timezone=browser&orgId=1)

console/network/src/lib.rs
d0cd: That's correct, and to clarify do you mean no **existing** deployment rejected?

raychu86: I meant to check if that any deployments were rejected because of the `MAX_PROGRAM_DEPTH`. Because if the rule is removed then block processing may be different.

raychu86: I just double-checked, deployments are rejected in 2 cases:
- Deployment already exists
- Failed `finalize_deployment` 

This should not affect that case.

d0cd: Ah gotcha. `Stack::initialize`, which is called during `verify_deployment` will check `MAX_PROGRAM_DEPTH`. This would be a "verification aborted" transaction which is not a "rejected" transaction so we do not need to create a migration. Is that correct?

raychu86: I believe that is correct. Would appreciate a double check here to make sure that removing this condition does not affect existing onchain state.

synthesizer/src/vm/mod.rs
raychu86: Ok I buy that, I think performance-wise this sufficient as well.

With upgradability, am I correct in assuming that you will be loading the old version of deployments first, then updating the process with their updated stacks as they come in? This would be for cyclic dependencies because you can't add stacks unless you've added their imports (Because the first iteration of program deployment can't have a cyclic dependency - cycles can only be created on upgrades)

d0cd: That's right, you're basically loading from genesis to tip. I think there are still some optimizations still left, but can be tackled later.

synthesizer/src/vm/mod.rs
```diff
@@ -109,81 +112,71 @@ impl<N: Network, C: ConsensusStorage<N>> VM<N, C> {
...
+                {
+                    Some(Ok(Some(height))) => height,
+                    _ => {
+                        bail!("Block height for deployment transaction '{transaction_id}' is not found in storage.")
+                    }
+                };
+                // Return the deployment and height.
+                let transaction_id = transaction_id.clone();
+                let local_map = HashMap::from([(*transaction_id, deployment)]);
+                let local_vec = vec![(*transaction_id, height)];
```
ljedrz: why the extra allocations for each deployment and height? can't we just use `push` and `insert` for individual items respectively in the `reduce` block?

synthesizer/src/vm/mod.rs
```diff
@@ -109,81 +112,71 @@ impl<N: Network, C: ConsensusStorage<N>> VM<N, C> {
...
                 }
-            }
+            )?;
+            // Extend the deployments and heights.
+            deployments.extend(local_deployments);
+            heights.extend(local_heights);
+        }
+
+        // Sort the deployments by their block heights.
+        heights.sort_by(|(_, a), (_, b)| a.cmp(b));
```
ljedrz:
```suggestion
        heights.sort_unstable_by(|(_, a), (_, b)| a.cmp(b));
```
I'm assuming equal heights (if we can even have duplicate heights here) may be reordered

synthesizer/src/vm/mod.rs
d0cd: Would I need to share the underlying `HashMap` and `Vec` need to be shared across threads? 
I was originally trying to avoid lock contention.

I did consider a solution where I collect the deployments and heights and then add it into the `HashMap` and `Vec` serially, but it turned out to slightly slower on [this](https://github.com/ProvableHQ/aleo-program-regressions/tree/master/regressions/profile_process_load_from_store)

ljedrz: no, you should be able to just return individual objects from this iterator by value; you would only need to change the later calls to `extend` to `push` and `insert`

synthesizer/src/vm/mod.rs
vicsn: @d0cd updated the base branch, you can resolve the conflict now!

raychu86 (approved):

synthesizer/process/src/stack/mod.rs
d0cd: I don't think this is always the case. `Stack::initialize` calls it without having an `Arc<Stack>`.
Unless you see any glaring issues, I think we could keep the interface as is.

console/network/src/lib.rs
d0cd: A genesis to tip sync test should be sufficient right?

synthesizer/src/vm/mod.rs
d0cd: Resolved!

synthesizer/process/src/stack/mod.rs
ljedrz: no, it's not a problem, it would only be a bit simpler (but also more restrictive)

ljedrz:
I did some profiling, and while these changes are barely noticeable in `mainnet`, the improvements on `canarynet` and `testnet` are **very** significant (improvements in the range of 80%-95%) in terms of runtime, the number of allocations, and max RAM use.

ljedrz (approved):
Profiling passed with flying colors.

console/network/src/lib.rs
vicsn: A genesis to tip sync will be run before the next release.

vicsn (approved):
LGTM

PR #2633, opened 2025-03-06
[Feat] Feature-gated R1CS hash collection

ljedrz:
The goal of this PR is to allow snarkVM users to verify that the R1CS objects aren't affected by potentially related changes, i.e. to facilitate regression testing.

I didn't feature-gate the additional `Hash` impls, as I don't think it would be impactful (especially not to runtime performance), and there have been times when I was interested in hashing some of the underlying objects during development.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -131,6 +131,7 @@ cli = [
 aleo-cli = [ "snarkvm-synthesizer/aleo-cli" ]
 async = [ "snarkvm-ledger/async", "snarkvm-synthesizer/async" ]
 cuda = [ "snarkvm-algorithms/cuda" ]
+save_r1cs_hashes = [ "snarkvm-circuit/save_r1cs_hashes" ]
 history = [ "snarkvm-synthesizer/history" ]
 parameters_no_std_out = [ "snarkvm-parameters/no_std_out" ]
 noconfig = [ ]
--- a/circuit/environment/Cargo.toml
+++ b/circuit/environment/Cargo.toml
@@ -59,6 +59,11 @@ version = "0.2"
 [dependencies.once_cell]
 version = "1.18.0"
 
+[dependencies.xxhash-rust]
+version = "0.8"
+features = ["xxh3"]
+optional = true
+
 [dev-dependencies.snarkvm-algorithms]
 path = "../../algorithms"
 features = [ "polycommit_full", "snark", "test" ]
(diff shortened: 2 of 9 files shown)


vicsn (approved):

ljedrz:
Note: the hash algorithm will shortly be changed to SHA-256 in order to reduce the likelihood of collisions; this shouldn't affect the overall shape of the PR.

kpandl (approved):

circuit/environment/src/helpers/r1cs.rs
```diff
@@ -20,10 +20,23 @@ use crate::{
...
 pub type Scope = String;
 
-#[derive(Debug)]
+/// A list of hashes of all the R1CS objects that have reached the
+/// coversion to Assignment stage. It's a vector in case there are
+/// any duplicates (which could indicate no change or redundant
+/// work) and since they need to eventually be sorted in order to
+/// have deterministic order, as they may be created in parallel.
+#[cfg(feature = "save_r1cs_hashes")]
+pub static R1CS_HASHES: Mutex<Vec<u64>> = Mutex::new(Vec::new());
```
niklaslong: It's a pity `Default` isn't const, oh well...

Cargo.toml
```diff
@@ -131,6 +131,7 @@ cli = [
 aleo-cli = [ "snarkvm-synthesizer/aleo-cli" ]
 async = [ "snarkvm-ledger/async", "snarkvm-synthesizer/async" ]
 cuda = [ "snarkvm-algorithms/cuda" ]
+save_r1cs_hashes = [ "snarkvm-circuit/save_r1cs_hashes" ]
```
niklaslong, line 134: We seem to have two conventions for feature flags (dashes or snake_case). Not something for this PR but we should consider aligning these in future. IIRC snarkOS mainly uses the former.

niklaslong (commented):
LGTM, pending the hash function change 👍

niklaslong (approved):

circuit/environment/Cargo.toml
```diff
@@ -59,6 +59,11 @@ version = "0.2"
 [dependencies.once_cell]
 version = "1.18.0"
 
+[dependencies.sha2]
```
vicsn, line 62: Can you explain why this crate is safe, or the safest option for SHA2

ljedrz: we're already using it for cryptographic purposes: https://github.com/ProvableHQ/snarkVM/blob/595d80778f24d15cedead12f4b001d94f09cc502/algorithms/Cargo.toml#L140

vicsn (approved):

PR #2634, opened 2025-03-06
[Feature] Introduces V2 programs.

d0cd:

This PR introduces the **syntax** for V2 programs.
At a high level, V2 programs:
- start with a `program$2` keyword
- allow `struct`, `record`, `mapping`, `function`, `closure`, and `import`.
- introduce an optional constructor via the `_init` keyword. Constructors must be specified after the above constructs.
- introduce user-defined `$metadata` to the program. Metadata must be specified after the constructor.

Here is an example of a V2 program.
```
import eth.aleo;
import usdc.aleo;

program$2 swap.aleo;

function swap:
    input r0 as eth.aleo/eth.record;
    input r1 as usdc.aleo/usdc.record;
    call eth.aleo/transfer r0 r1.owner r0.amount into r2 r3;
    call usdc.aleo/transfer r1 r0.owner r1.amount into r4 r5;
    output r2 as eth.aleo/eth.record;
    output r4 as usdc.aleo/usdc.record;

_init:
    assert.eq true true;
    
$metadata edition: 0u8;
$metadata owner: aleo1rhgdu77hgyqd3xjj8ucu3jj9r2krwz6mnzyd80gncr5fxcwlh5rsvzp9px;
$metadata upgradable: false;
```

**The implementation for V2 programs will be included in a following PR.**


synthesizer/program/src/v2/bytes.rs
```diff
@@ -0,0 +1,199 @@
...
+    use super::*;
+    use crate::Program;
+    use console::network::MainnetV0;
+
+    type CurrentNetwork = MainnetV0;
+
+    #[test]
+    fn test_bytes() -> Result<()> {
+        let program = r"
+program$2 token.aleo;
```
vicsn, line 165: From my humble POV, I'm not a fan of adding a new symbol `$` to the syntax - even if we don't typically expose aleo instructions to end users. Also because it doesn't work the same way like it does in the other syntax everyone knows (shell scripts). 
- For **program version**: what about a new keyword like `version` which we declare after `import` and before `program`? Related: I think `metadata.version` in the PR description should be changed to `metadata.edition`.
- For **metadata** we could simply drop the `$`.

d0cd: I don't have strong opinions here. I'll leave it to the reviewers to arbitrate on syntax.
FWIW, I feel like the `version` keyword doesn't create enough of a conceptual separation but I can be convinced if I stare at some programs long enough.

raychu86: The frequency of introducing new program versions should also be taken into account. For instance, in Solidity versions are updated quite frequently. (Though the counterpart for solidity is Leo, and Aleo instructions is bytecode).

IMO program versions won't be updated frequently, so something more natural would be preferred.

raychu86: How integral is the addition of `$2`? Is the constructor or metadata be sufficient of a delineator for program version?

Program "version" in byte serialization is already hidden from the aleo instructions anyways. Is it possible to just keep retain "program" as the same keyword?

d0cd: Constructors are optional. However, in a later PR, we do require that V2 programs contain `program_owner`, `edition`, and `updatable` metadata declarations. Perhaps that enough of a conceptual difference?

In general, `$2` is not integral. Since it sounds like the majority prefers to keep the syntax closer, I will attempt an alternate solution.

synthesizer/program/src/bytes.rs
```diff
@@ -21,46 +21,15 @@ impl<N: Network, Instruction: InstructionTrait<N>, Command: CommandTrait<N>> Fro
...
-                4 => program.add_function(FunctionCore::read_le(&mut reader)?).map_err(|e| error(e.to_string()))?,
-                // Invalid variant.
-                _ => return Err(error(format!("Failed to parse program. Invalid component variant '{variant}'"))),
-            }
-        }
+        // Ensure the version is valid and initialize the program.
+        let program = match version {
+            1 => Self::ProgramV1(ProgramCoreV1::read_le_internal(&mut reader).map_err(|e| error(e.to_string()))?),
+            2 => Self::ProgramV2(ProgramCoreV2::read_le_internal(&mut reader).map_err(|e| error(e.to_string()))?),
+            _ => return Err(error("Invalid program version")),
```
raychu86: nit
```suggestion
            invalid_version => return Err(error("Invalid program version ({invalid_version})")),
```

synthesizer/program/src/v2/bytes.rs
raychu86: I do like that this creates a clear indication to the program creator or users that the program is of a specific version instead of having to infer based on program syntax.

d0cd:
After review, the team has decided to go with a different design that supports constructors in the original programs. The latest updates can be found here: https://github.com/ProvableHQ/snarkVM/issues/2654

PR #2635, opened 2025-03-07
Add test-exports feature for accessing snarkVM/algorithms TestCircuit in external projects

kpandl:
## Motivation

Enable access to test utilities (specifically the Varuna `TestCircuit` in snarkVM/algorithms) from external crates without modifying the original visibility modifiers or core logic. This provides a clean way to use testing components in dependent projects while maintaining clear separation between test and production code.

Tested usage in an external project on execution fee estimation.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -149,7 +149,7 @@ parameters = [ "snarkvm-parameters" ]
 synthesizer = [ "snarkvm-synthesizer" ]
 utilities = [ "snarkvm-utilities" ]
 wasm = [ "snarkvm-wasm" ]
-test-exports = [ "snarkvm-algorithms/test-exports" ]
+test_exports = [ "snarkvm-algorithms/test_exports" ]
 test_targets = [ "snarkvm-console/test_targets" ]
 
 [dependencies.snarkvm-algorithms]
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -149,6 +149,7 @@ parameters = [ "snarkvm-parameters" ]
 synthesizer = [ "snarkvm-synthesizer" ]
 utilities = [ "snarkvm-utilities" ]
 wasm = [ "snarkvm-wasm" ]
+test-exports = [ "snarkvm-algorithms/test-exports" ]
 test_targets = [ "snarkvm-console/test_targets" ]
 
 [dependencies.snarkvm-algorithms]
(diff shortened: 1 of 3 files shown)


algorithms/Cargo.toml
```diff
@@ -212,3 +212,4 @@ serial = [
   "snarkvm-utilities/serial"
 ]
 snark = [ "crypto_hash", "fft", "msm", "polycommit", "r1cs" ]
+test-exports = ["test"]
```
ljedrz: I wouldn't reuse the `test` feature, as that one has the potential to alter some logic (introducing additional checks, using different mix/max values, etc.); instead, I'd change the attribute from
```
#[cfg(any(test, feature = "test"))]
```
to
```
#[cfg(any(test, feature = "test", feature = "test-exports"))]
```
on methods/modules/objects we'd like to use in external tools.

algorithms/src/snark/varuna/data_structures/mod.rs
```diff
@@ -35,5 +35,13 @@ pub(super) mod test_circuit;
 #[cfg(any(test, feature = "test"))]
 pub use test_circuit::*;
 
+/// Test exports module that provides access to test utilities for external crates.
+#[cfg(feature = "test-exports")]
+pub mod test_exports {
+    /// Test circuit for Varuna (for testing purposes only).
+    #[cfg(feature = "test")]
```
ljedrz:
```suggestion
```
I believe this won't be needed once the other comment is addressed.

ljedrz (changes requested):
While this may not be an issue right now, I have one suggestion regarding future-proofing.

algorithms/Cargo.toml
kpandl: thanks, addressed here: https://github.com/ProvableHQ/snarkVM/pull/2635/commits/0d92219c67e210cfea8b2fcc368f002b94dde93b

ljedrz (approved):
LGTM 👍

Cargo.toml
```diff
@@ -149,6 +149,7 @@ parameters = [ "snarkvm-parameters" ]
 synthesizer = [ "snarkvm-synthesizer" ]
 utilities = [ "snarkvm-utilities" ]
 wasm = [ "snarkvm-wasm" ]
+test-exports = [ "snarkvm-algorithms/test-exports" ]
```
vicsn: The feature flag right below it uses underscores

kpandl: addressed here: https://github.com/ProvableHQ/snarkVM/pull/2635/commits/caff7c6ce84bf97902e22957141c41b014a0a350

ljedrz (approved):

PR #2639, opened 2025-03-10
[Bug] Fix cargo doc compilation

kaimast:
## Motivation
`cargo doc` currently fails for snarkVM. This is because in `curves/src/edwards_bls12/fr.rs` a constant is set to `unimplemented!()`. This PR simply disables documentation for that one constant.

## Test Plan
This PR only adds an attribute to existing code and does not change any functionality. Existing tests should be sufficient to verify that it is safe to merge.

## Notes
`cargo doc` still prints some warnings, such as links being formatted incorrectly. I want to keep this PR minimal, but I can address those with a future PR.

### DIFF
--- a/curves/src/edwards_bls12/fr.rs
+++ b/curves/src/edwards_bls12/fr.rs
@@ -33,6 +33,8 @@ impl Fp256Parameters for FrParameters {}
 impl FftParameters for FrParameters {
     type BigInteger = BigInteger;
 
+    // `cargo doc` will fail without this attribute
+    #[doc(hidden)]
     const POWERS_OF_ROOTS_OF_UNITY: &'static [BigInteger] = unimplemented!();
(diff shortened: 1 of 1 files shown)


kaimast:
@niklaslong maybe you can take a quick look at this?

niklaslong:
Beautiful! Thank you for fixing this 🙌

niklaslong (approved):

vicsn:
Cool, can we add some kind of cargo doc check to the git commit hook to prevent regressions? Perhaps cargo test --doc, maybe with --norun

vicsn:
> Cool, can we add some kind of cargo doc check to the git commit hook to prevent regressions? Perhaps cargo test --doc, maybe with --norun

On second thought, this is unlikely to break, and its part of the release workflow to look into the doc interface diff, so not worth it.

PR #2640, opened 2025-03-10
Doc fixes and additions for block verification/construction.

acoglio:
Fix some doc comments.

Make two similar error messages consistent.

Clarify that a naturally expected check is done elsewhere.

Make some doc comments more precise (were confusing).

Make internal "unchecked" constructor non-public.

Document some redundant block checks. These are already performed during block construction, so there is no need to repeat them during block verification. I didn't see ways in which block verification is called without having called block construction first.

### DIFF
--- a/ledger/block/src/header/mod.rs
+++ b/ledger/block/src/header/mod.rs
@@ -44,7 +44,7 @@ pub struct Header<N: Network> {
     ratifications_root: Field<N>,
     /// The solutions root of the puzzle.
     solutions_root: Field<N>,
-    /// The subdag root of the authority.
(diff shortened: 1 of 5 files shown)


vicsn:
Thank you. Can you revert 315354c? The check is extremely cheap and IMO doesn't create confusion, and its more work for reviewers to ascertain this can be safely ommitted.

acoglio:
I've restored the checks but documented their redundancy.

I've also just added a commit that documents another redundant check (discussed on Slack some time ago, along with the others). This one, on sub-DAG transmissions, may not be as cheap as the others.

vicsn (approved):

PR #2641, opened 2025-03-10
[Feature] Introduce the `metadata.get` command.

d0cd:

This PR introduces the `metadata.get` command to both V1 and V2 programs, allowing them to request their own metadata or metadata of other programs. Note that V1 programs only have `_checksum` metadata, while V2 programs can also use metadata declarations.

### DIFF
--- a/synthesizer/process/src/stack/finalize_types/initialize.rs
+++ b/synthesizer/process/src/stack/finalize_types/initialize.rs
@@ -492,10 +492,8 @@ impl<N: Network> FinalizeTypes<N> {
         metadata_get: &MetadataGet<N>,
     ) -> Result<()> {
         // Ensure that the program is defined.
-        // Note that we do not check that the metadata name is valid, as that can change in future updates.
-        if let CallOperator::Locator(locator) = metadata_get.name() {
-            // Ensure the program ID is defined.
-            let program_id = locator.program_id();
+        // Note that we do not check that the metadata name is valid, as that can change by an update to the program.
+        if let Some(program_id) = metadata_get.program() {
             // Ensure the program ID is imported by the current program.
             if !stack.program().imports().keys().contains(program_id) {
                 bail!("External program '{program_id}' is not imported by '{}'.", stack.program_id());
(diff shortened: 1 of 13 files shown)


synthesizer/process/src/stack/finalize_types/initialize.rs
```diff
@@ -483,6 +484,35 @@ impl<N: Network> FinalizeTypes<N> {
...
+            stack.get_external_stack(program_id)?;
+        }
+        // Get the destination register.
+        let destination = metadata_get.destination().clone();
+        // Ensure the destination register is a locator (and does not reference an access).
+        ensure!(matches!(destination, Register::Locator(..)), "Destination '{destination}' must be a locator.");
+        // Insert the destination register.
+        self.add_destination(
+            destination,
+            FinalizeType::Plaintext(PlaintextType::Literal(*metadata_get.destination_type())),
```
vicsn, line 511: Are we or should we be constraining the destination type anywhere?

synthesizer/program/src/logic/command/metadata_get.rs
```diff
@@ -0,0 +1,303 @@
...
+    Checksum,
+    /// A declared metadata identifier.
+    Identifier(Identifier<N>),
+}
+
+/// A command to get metadata about a program, e.g. `metadata.get program_owner into r1 as address;`.
+/// Gets the value with the `name` from the program and stores it in the `destination` register.
+/// The value is checked to be of the `destination_type`.
+///
+/// `metadata.get _checksum into r1 as u128;` is a special case where the metadata is not retrieved from the program.
```
vicsn, line 38: Given that the checksum behaves quite differently, did you consider making a separate checksum command? What is the biggest downside, more lines of code?

synthesizer/process/src/stack/finalize_types/initialize.rs
d0cd: `destination_type()` returns a `LiteralType`
Is there a particular constraint you had in mind?

synthesizer/program/src/logic/command/metadata_get.rs
d0cd: Yeah exactly. But it could be split out fairly easily.

d0cd:
After review, the team has decided to go with a different design that supports constructors in the original programs. The same metadata can be accessed via some newly introduced operands. The latest updates can be found here: https://github.com/ProvableHQ/snarkVM/issues/2654

PR #2642, opened 2025-03-11
[Feature] Support program updates

d0cd:

This PR adds support for updates to V2 programs.


d0cd:
After review, the team has decided to go with a different design that supports constructors in the original programs. A bulk of this work will be ported over to a new PR. The latest updates can be found here: https://github.com/ProvableHQ/snarkVM/issues/2654

PR #2644, opened 2025-03-12
[Hotfix] Optimize deployment loading on initialization

raychu86:
## Motivation

This PR implements optimizations to `load_deployment_and_imports` when a VM is initialized. The previous logic was quite naive in that each individual parallel thread would potentially perform a deep recursion of imports. The fix was the introduce shared state between the threads to ensure that wasted work is limited.

### DIFF
--- a/synthesizer/src/vm/mod.rs
+++ b/synthesizer/src/vm/mod.rs
@@ -97,7 +97,7 @@ impl<N: Network, C: ConsensusStorage<N>> VM<N, C> {
     #[inline]
     pub fn from(store: ConsensusStore<N, C>) -> Result<Self> {
         // Initialize a new process.
-        let process = Arc::new(RwLock::new(Process::load()?));
+        let mut process = Process::load()?;
 
         // Initialize the store for 'credits.aleo'.
         let credits = Program::<N>::credits()?;
@@ -111,9 +111,10 @@ impl<N: Network, C: ConsensusStorage<N>> VM<N, C> {
 
         // A helper function to retrieve all the deployments.
         fn load_deployment_and_imports<N: Network, T: TransactionStorage<N>>(
-            process: Arc<RwLock<Process<N>>>,
+            process: &Process<N>,
             transaction_store: &TransactionStore<N, T>,
             transaction_id: N::TransactionID,
+            deployments: Arc<RwLock<IndexMap<ProgramID<N>, Deployment<N>>>>,
         ) -> Result<()> {
             // Retrieve the deployment from the transaction ID.
             let deployment = match transaction_store.get_deployment(&transaction_id)? {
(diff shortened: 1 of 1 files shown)


d0cd (approved):
LGTM! One observation is that the number of stack frames created is equivalent to the import depth. This should be addressed before upgradability is rolled in.

synthesizer/src/vm/mod.rs
```diff
@@ -126,17 +127,14 @@ impl<N: Network, C: ConsensusStorage<N>> VM<N, C> {
...
             }
 
-            // Prepare a vector for the deployments.
-            let mut deployments = vec![];
-
             // Iterate through the program imports.
             for import_program_id in program.imports().keys() {
                 // Add the imports to the process if does not exist yet.
-                if !process.contains_program(import_program_id) {
+                if !process.contains_program(import_program_id) && !deployments.read().contains_key(program_id) {
```
ljedrz, line 137:
```suggestion
                if !(process.contains_program(import_program_id) || deployments.read().contains_key(program_id)) {
```

synthesizer/src/vm/mod.rs
```diff
@@ -145,24 +143,25 @@ impl<N: Network, C: ConsensusStorage<N>> VM<N, C> {
...
                 }
             }
 
             // Once all the imports have been included, add the parent deployment.
-            deployments.push((*program_id, deployment));
+            if !deployments.read().contains_key(program_id) {
+                // Note: There may be re-insertions due to parallelism, but it is safe to
+                //  reinsert into the IndexMap because it should be the same object.
+                deployments.write().entry(*program_id).or_insert(deployment);
+            }
```
ljedrz, line 155: If we don't expect a lot of misses (i.e. duplicate entries), keeping only the inner expression would be faster.

synthesizer/src/vm/mod.rs
ljedrz: (personal preference for lumping together similar terms, feel free to ignore)

ljedrz (approved):
LGTM, left one nit and one suggestion.

synthesizer/src/vm/mod.rs
```diff
@@ -145,24 +143,25 @@ impl<N: Network, C: ConsensusStorage<N>> VM<N, C> {
...
 
-            Ok(deployments)
+            Ok(())
         }
 
         // Retrieve the transaction store.
         let transaction_store = store.transaction_store();
         // Retrieve the list of deployment transaction IDs.
         let deployment_ids = transaction_store.deployment_transaction_ids().collect::<Vec<_>>();
+        // TODO (raychu86): Load them in order of block height to limit recursion.
```
vicsn: Not sure ordering by block height and still using 256-sized chunks will necessarily limit recursion, will make sure its investigated some day when finishing https://github.com/ProvableHQ/snarkVM/pull/2591

vicsn (approved):
LGTM

d0cd:
There is a regression test in [this](https://github.com/ProvableHQ/aleo-program-regressions/pull/10) PR, under `profile_process_load_from_store` which measures the time to load a VM from DB for `mainnet` and `testnet` programs.
- **V1.3.0 Testnet:** 6.5445 seconds
- **V1.3.0 Mainnet:** 58.05 ms
- **Hotfix Testnet:** 5.8576 seconds
- **Hotfix Mainnet:** 49.99 ms

This code could also be run under a profiler to get a better idea of CPU/memory characteristics.

synthesizer/src/vm/mod.rs
vicsn:
```suggestion
        // TODO (raychu86): Investigate loading them in order to limit recursion.
```

PR #2648, opened 2025-03-12
Implement execute_authorization_with_response.

mikebenfield:
This way callers can access the `Response` and, in particular, the `Output`s.

### DIFF
--- a/synthesizer/src/vm/execute.rs
+++ b/synthesizer/src/vm/execute.rs
@@ -41,7 +41,7 @@ impl<N: Network, C: ConsensusStorage<N>> VM<N, C> {
         // Determine if a priority fee is declared.
         let is_priority_fee_declared = priority_fee_in_microcredits > 0;
         // Compute the execution.
-        let execution = self.execute_authorization_raw(authorization, query.clone(), rng)?;
+        let (execution, _) = self.execute_authorization_raw(authorization, query.clone(), rng)?;
         // Compute the fee.
         let fee = match is_fee_required || is_priority_fee_declared {
             true => {
@@ -83,6 +83,9 @@ impl<N: Network, C: ConsensusStorage<N>> VM<N, C> {
     }
 
     /// Returns a new execute transaction for the given authorization.
+    ///
+    /// This is identical to `execute_authorization_with_response` except that it
+    /// discards the `Response`.
     pub fn execute_authorization<R: Rng + CryptoRng>(
         &self,
         execute_authorization: Authorization<N>,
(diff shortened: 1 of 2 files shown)


vicsn (approved):
LGTM!

mikebenfield:
I don't think that failed check can be related to this PR can it? Either way I don't think I can rerun it; could you do that for me?

synthesizer/src/vm/execute.rs
```diff
@@ -15,6 +15,8 @@
 
 #![allow(clippy::too_many_arguments)]
 
+use console::program::Response;
```
niklaslong: I think the convention for VM is to have imports in the mod file of a subdirectory. Would it make sense to move this one up a level?

synthesizer/src/vm/execute.rs
```diff
@@ -83,22 +85,37 @@ impl<N: Network, C: ConsensusStorage<N>> VM<N, C> {
...
+    /// This is identical to `execute_authorization_with_response` except that it
+    /// discards the `Response`.
     pub fn execute_authorization<R: Rng + CryptoRng>(
         &self,
         execute_authorization: Authorization<N>,
         fee_authorization: Option<Authorization<N>>,
         query: Option<Query<N, C::BlockStorage>>,
         rng: &mut R,
     ) -> Result<Transaction<N>> {
+        Ok(self.execute_authorization_with_response(execute_authorization, fee_authorization, query, rng)?.0)
```
niklaslong: Style option: would explicitly writing `let (execution, _)` be clearer here?

synthesizer/src/vm/execute.rs
```diff
@@ -154,7 +171,7 @@ impl<N: Network, C: ConsensusStorage<N>> VM<N, C> {
                 lap!(timer, "Compute the proof");
 
                 // Return the execution.
-                Ok(cast_ref!(execution as Execution<N>).clone())
+                Ok((cast_ref!(execution as Execution<N>).clone(), cast_ref!(response as Response<N>).clone()))
```
niklaslong, line 174: I haven't gone through these macros yet but do you know why these clones are necessary? 🕵️‍♂️

niklaslong (approved):
Left a few thoughts and a question but the diff looks good to me!

synthesizer/src/vm/execute.rs
mikebenfield: the `cast_ref` macro returns a reference with a lifetime that ends in the local scope, and these types don't implement `Copy`, so we can't just dereference. 

I'd really like to understand the `cast_ref` macro; it seems very strange to me. I think we're doing all executions using the `Mainnet` parameter for `N` and then casting everything we get to whatever other network we need. Why this is necessary is a mystery to me, and no one I've asked knows.

niklaslong: Thanks for the context! Perhaps something for later, in that case 👍

vicsn:
@mikebenfield please no force pushes on snarkVM/OS after reviews have started, makes final review easier. We may add this to `CONTRIBUTING.md` but I'll make sure to remind folks too.

PR #2650, opened 2025-03-13
[Fix] Add deserialization check on max solutions.

acoglio:
During deserialization, it is important to check that the number of next items to deserialize is within bounds. For block deserialization, this is done for transactions, aborted transaction IDs, and and aborted solution IDs, but the check for solutions was missing. This commit adds it.

Also fix a comment in some deserialization code.


acoglio:
The CI failure seems unrelated to my changes.

ljedrz (approved):
LGTM

raychu86 (approved):

PR #2653, opened 2025-03-18
[Feature] Introduce `constructor`, `Operand::Checksum` and `Operand::Edition`

d0cd:

**DO NOT MERGE WITHOUT AUDIT**

## Motivation

This PR introduces:
- `constructor`s to programs. Constructors are blocks of code that are run on-chain when finalizing a deployment.
- `Operand::Checksum`. An operand that returns a field element containing the program's checksum. The checksum is defined as the `BHP1024` hash over the program's bits.
- `Operand::Edition`. The edition for the program. This is currently enforced to be `N::EDITION` (0) for all programs.

**Note: The keyword `constructor` is now restricted after `ConsensusVersion::V6` by [this](https://github.com/ProvableHQ/snarkVM/pull/2685) PR.** This is enforced at the consensus-level as opposed to the program-level to ensure backwards compatibility.

## Test Plan

Integration test cases were added to verify:
- that a program is not deployed if its constructor fails.
- a constructor can set a value in a mapping and that the new operands are functional.

Regression [tests](https://github.com/ProvableHQ/aleo-program-regressions/pull/15) were re-ran to verify that the changes to the execution cost calculations do not affect existing programs.


ledger/block/src/transactions/confirmed/mod.rs
```diff
@@ -79,10 +79,11 @@ impl<N: Network> ConfirmedTransaction<N> {
                     program.mappings().len(),
                 )
             }
-            // Ensure the number of finalize operations matches the number of 'UpdateKeyValue' finalize operations.
-            if num_update_key_values != fee.num_finalize_operations() {
+            // Ensure the number of fee finalize operations lower bounds the number of 'UpdateKeyValue' finalize operations.
+            // The lower bound is due to the fact that constructors can issue 'UpdateKeyValue' operations as part of the deployment.
+            if num_update_key_values < fee.num_finalize_operations() {
```
raychu86: Is there a way to fetch the number of finalize operations from the program constructor? 

I realize that `num_finalize_operations` is a function that only exists for fees currently and is hardcoded to 1 or 0.

synthesizer/process/src/cost.rs
```diff
@@ -405,6 +409,20 @@ pub fn cost_per_command<N: Network>(
...
 }
 
+/// Returns the minimum number of microcredits required to run the constructor in the given program.
+/// Each command in a constructor costs 100_000 microcredits.
+/// If a constructor does not exist, no cost is incurred.
+pub fn constructor_cost_in_microcredits<N: Network>(program: &Program<N>) -> Result<u64> {
+    match program.constructor() {
+        Some(constructor) => {
+            let num_commands = constructor.commands().len() as u64;
+            let cost = num_commands.checked_mul(100_000).ok_or(anyhow!("Constructor cost overflowed"))?;
```
raychu86: How was this constructor cost chosen? 

Was `TRANSACTION_SPEND_LIMIT` factored in? At current limit of 100,000,000, this restricts constructors to 100 commands.

raychu86: Additionally, is it intentional for all commands to be priced the same? An `add` is equivalent to the largest possible `hash` or `set`.

synthesizer/process/src/finalize.rs
```diff
@@ -183,6 +190,101 @@ fn finalize_fee_transition<N: Network, P: FinalizeStorage<N>>(
...
+    let constructor_types = stack.get_constructor_types()?.clone();
+
+    // Initialize the finalize registers.
+    let mut registers = FinalizeRegisters::new(state, transition_id, *program_id.name(), constructor_types, nonce);
+
+    // Initialize a counter for the commands.
+    let mut counter = 0;
+
+    // Evaluate the commands.
+    while counter < constructor.commands().len() {
```
raychu86, line 229: How much of this logic can be abstracted between `finalize_constructor` and `finalize_transition`?

Aside from the `Command::Await`s, most of the logic looks equivalent

ledger/block/src/transactions/confirmed/mod.rs
d0cd: You could get an upper bound based on the number of `set`, `remove`, (double-check on others) commands that exist in the constructor definition.
You can't always get an exact number since a constructor may contain branches.

synthesizer/process/src/finalize.rs
```diff
@@ -68,6 +68,13 @@ impl<N: Network> Process<N> {
             }
             finish!(timer, "Initialize the program mappings");
 
+            // If the program has a constructor, execute it and extend the finalize operations.
```
raychu86, line 71: Can you add the timer profiling to this section as well?

synthesizer/process/src/cost.rs
d0cd: No analysis was done to back into the costs.
It was an arbitrary restriction chosen to limit the number of commands in a constructor.
We could go with a similar pricing for executions and add a constructor multiplier.
In general, it's fairly tricky to re-use the same pricing logic for execution costs, but with some refactoring, we could probably get it to work.

synthesizer/process/src/finalize.rs
d0cd: I believe that the `Command::BranchEq`, `Command::BranchNeq`, and `_ =>` match arms can be shared across the two.

The loop structures are fairly different, because `finalize_transition` deals with `Future`s and `await`s.

synthesizer/process/src/stack/helpers/initialize.rs
```diff
@@ -23,12 +23,15 @@ impl<N: Network> Stack<N> {
...
             stacks: Arc::downgrade(&process.stacks),
+            constructor_types: Default::default(),
             register_types: Default::default(),
             finalize_types: Default::default(),
             universal_srs: process.universal_srs().clone(),
             proving_keys: Default::default(),
             verifying_keys: Default::default(),
             program_address: program.id().to_address()?,
+            program_checksum: program.checksum()?,
+            program_edition: U16::new(N::EDITION),
```
raychu86, line 34: Will we need migration logic for disallowing deployments with old editions once `Network::EDITION` is updated?

d0cd: With updatability, deployments will no longer use `N::EDITION`. Instead, they will start from 0 and increment.
Because `N::EDITION` is currently 0 for all current networks, I don't believe we will need migration logic.

synthesizer/program/src/lib.rs
```diff
@@ -641,6 +679,9 @@ impl<N: Network, Instruction: InstructionTrait<N>, Command: CommandTrait<N>> Pro
         "impl",
         "type",
         "future",
+        "constructor",
```
raychu86: Does adding keywords invalidate existing programs? Do we need special logic to allow existing programs to keep functions named "constructor", "edition", or "checksum"?

d0cd: I checked existing programs on `mainnet`, `testnet`, and `canary`.
None of these keywords were found so far.
[[source]](https://github.com/ProvableHQ/aleo-program-regressions/blob/test/more-regressions-2/regressions/count_keyword_instances/src/main.rs)

synthesizer/program/src/logic/command/contains.rs
```diff
@@ -84,9 +84,9 @@ impl<N: Network> Contains<N> {
             CallOperator::Resource(mapping_name) => (*stack.program_id(), mapping_name),
         };
 
-        // Ensure the mapping exists in storage.
-        if !store.contains_mapping_confirmed(&program_id, &mapping_name)? {
-            bail!("Mapping '{program_id}/{mapping_name}' does not exist in storage");
+        // Ensure the mapping exists.
+        if !store.contains_mapping_speculative(&program_id, &mapping_name)? {
```
raychu86, line 88: I assume the intention for this change is to allow constructors to be properly executed. How can we check that we hit all of these cases/opcodes?

raychu86: Also, does this now enable transactions to be executed in the same block that their program was deployed? (via a malicious validator injecting the execution after the deployment into his proposal)

d0cd: Added an integration test that uses all of the changed opcodes in a `constructor`.
Happy to take more suggestions

synthesizer/process/src/finalize.rs
raychu86: If it is messier to abstract, then don't attempt it. Just wanted to see if we can remove code duplication

synthesizer/program/src/logic/instruction/operand/bytes.rs
```diff
@@ -49,6 +51,14 @@ impl<N: Network> ToBytes for Operand<N> {
             Self::Caller => 4u8.write_le(&mut writer),
             Self::BlockHeight => 5u8.write_le(&mut writer),
             Self::NetworkID => 6u8.write_le(&mut writer),
+            Self::Checksum(program_id) => {
+                7u8.write_le(&mut writer)?;
+                program_id.write_le(&mut writer)
```
raychu86: Since `program_id` is an option, you may need to do something like the following:

```
                match fee {
                    None => 0u8.write_le(&mut writer),
                    Some(fee) => {
                        1u8.write_le(&mut writer)?;
                        fee.write_le(&mut writer)
                    }
```

synthesizer/process/src/cost.rs
d0cd: [Here](https://github.com/ProvableHQ/snarkVM/pull/2658) is a draft PR with an alternate design that is based off our existing execution pricing.
The issue is that the interface for `deployment_cost` needs to change because you need a `Stack` to compute the size of the data types.

synthesizer/program/src/logic/instruction/operand/bytes.rs
d0cd: Done in the byte utils: https://github.com/ProvableHQ/snarkVM/blob/d03247689d5017b886116ab6044e51234cb6c4ff/utilities/src/bytes.rs#L459

synthesizer/process/src/finalize.rs
```diff
@@ -183,6 +192,101 @@ fn finalize_fee_transition<N: Network, P: FinalizeStorage<N>>(
...
+    // Get the constructor logic. If the program does not have a constructor, return early.
+    let Some(constructor) = stack.program().constructor() else {
+        return Ok(finalize_operations);
+    };
+
+    // Get the constructor types.
+    let constructor_types = stack.get_constructor_types()?.clone();
+
+    // Initialize the finalize registers.
+    let mut registers = FinalizeRegisters::new(state, transition_id, *program_id.name(), constructor_types, nonce);
```
howardwu, line 223: Auditors: The parameters used to initialize the FinalizeRegisters here should be reviewed for correctness and soundness.

synthesizer/program/src/lib.rs
vicsn: If a program is deployed at the wrong moment this can fork the network. With so many features flying around, I'd rather play it safe, can this be made conditional on a network upgrade? It might even save you time because you don't have to run the regression test anymore. Moreover, we will likely keep adding keywords, so this migration logic may be a useful thing to have in the future.

Because many PRs will depend on the next network upgrade, do you want to make a separate PR for the next v5 upgrade already, similar to: https://github.com/ProvableHQ/snarkVM/pull/2657

synthesizer/process/src/cost.rs
```diff
@@ -53,10 +59,11 @@ pub fn deployment_cost<N: Network>(deployment: &Deployment<N>) -> Result<(u64, (
     // Compute the total cost in microcredits.
     let total_cost = storage_cost
         .checked_add(synthesis_cost)
+        .and_then(|x| x.checked_add(constructor_cost))
         .and_then(|x| x.checked_add(namespace_cost))
         .ok_or(anyhow!("The total cost computation overflowed for a deployment"))?;
 
-    Ok((total_cost, (storage_cost, synthesis_cost, namespace_cost)))
+    Ok((total_cost, (storage_cost, synthesis_cost, constructor_cost, namespace_cost)))
```
vicsn, line 66: After https://github.com/ProvableHQ/snarkOS/pull/3471 is merged, you'll have to make a snarkOS PR to support the interface update.

In case we wait with this PR being merged until audits are completed, maybe a single snarkOS draft PR for all the program upgradability stuff is easier.

synthesizer/program/src/lib.rs
vicsn: Moreover, does this PR already "enable" the use of constructors/checksum/edition? Migration logic will also technically be needed for all of those aspects, because it changes what is regarded as a valid deployment. However, I can imagine that placing appropriate consensus version guards might become extremely unwieldy and time consuming, so we can also coordinate with validators to upgrade very quickly or even synchronously.

vicsn:
Can you make an issue drafting documentation for the constructor and new opcodes, ignoring program upgradability for now: https://github.com/aleoNet/welcome/issues/new

synthesizer/program/src/logic/command/contains.rs
vicsn: > Also, does this now enable transactions to be executed in the same block that their program was deployed? (via a malicious validator injecting the execution after the deployment into his proposal).

I don't think so because all transactions are verified in `prepare_for_speculate` before any (speculative) deployment happens. And even if it did, I can't see any vulnerability arising from this, though we should understand it.

synthesizer/program/src/constructor/parse.rs
```diff
@@ -0,0 +1,128 @@
...
+#[cfg(test)]
+mod tests {
+    use super::*;
+    use crate::Constructor;
+    use console::network::MainnetV0;
+
+    type CurrentNetwork = MainnetV0;
+
+    #[test]
+    fn test_constructor_parse() {
```
vicsn, line 88: Can/should we add some negative tests too? E.g. that an input or output is not allowed, perhaps other similar conceptual differences

synthesizer/program/src/lib.rs
d0cd: Migration incoming in this PR: https://github.com/ProvableHQ/snarkVM/pull/2659
Basically, no constructors or additional operands before V5.

synthesizer/process/src/cost.rs
d0cd: The draft has been merged in.

synthesizer/process/src/finalize.rs
d0cd: During live review, we decided to keep them separate.

synthesizer/program/src/logic/instruction/operand/bytes.rs
d0cd: It's since been removed from the byte utils and inlined in `operand/bytes.rs`

synthesizer/process/src/stack/registers/load.rs
```diff
@@ -126,6 +142,26 @@ impl<N: Network, A: circuit::Aleo<Network = N>> RegistersLoadCircuit<N, A> for R
             Operand::BlockHeight => bail!("Cannot load the block height in a non-finalize context"),
             // If the operand is the network ID, throw an error.
             Operand::NetworkID => bail!("Cannot load the network ID in a non-finalize context"),
+            // If the operand is the checksum, load the value of the checksum.
+            Operand::Checksum(program_id) => {
```
vicsn, line 146: Are we using checksum and edition in a non-finalize context? Consider bailing like we do above.

d0cd: Yes, we use in both on-chain and off-chain contexts.

synthesizer/program/src/constructor/parse.rs
d0cd: Good suggestion! In review, we settled on making constructors non-empty. I'll be following up with a PR that adds this change and includes negative tests. That PR will likely be made against https://github.com/ProvableHQ/snarkVM/pull/2655, since constructors are used heavily there.

synthesizer/process/src/stack/finalize_types/initialize.rs
```diff
@@ -709,7 +725,7 @@ impl<N: Network> FinalizeTypes<N> {
                 _ => bail!("Instruction '{instruction}' is not for opcode '{opcode}'."),
             },
             Opcode::Command(opcode) => {
-                bail!("Fatal error: Cannot check command '{opcode}' as an instruction in 'finalize {finalize_name}'.")
+                bail!("Fatal error: Cannot check command '{opcode}' as an instruction.")
```
vicsn, line 728: What about we pass in `Option<Identifier>`, where `None` will imply it was the constructor? Not sure what these logs are useful for, but then we don't regress.

synthesizer/process/src/stack/finalize_types/initialize.rs
```diff
@@ -16,6 +16,29 @@
...
+        stack: &(impl StackMatches<N> + StackProgram<N>),
+        constructor: &Constructor<N>,
+    ) -> Result<Self> {
+        // Initialize a map of registers to their types.
+        let mut finalize_types = Self { inputs: IndexMap::new(), destinations: IndexMap::new() };
+
+        // Check the commands are well-formed.
+        for command in constructor.commands() {
+            // Verify that the command is not an await.
+            if command.is_await() {
```
vicsn, line 32: Are there any other finalize-specific commands or operands we should prevent from using? Are these fine? https://developer.aleo.org/guides/aleo/special_operands

synthesizer/process/src/stack/execute.rs
```diff
@@ -355,6 +375,26 @@ impl<N: Network> StackExecute<N> for Stack<N> {
                     Operand::NetworkID => {
                         bail!("Illegal operation: cannot retrieve the network id in a function scope")
                     }
+                    // If the operand is the checksum, retrieve the checksum from the stack.
+                    Operand::Checksum(program_id) => {
+                        let checksum = match program_id {
```
vicsn, line 380: Are we allowing these operands in the function scope?

ledger/block/src/transactions/confirmed/mod.rs
```diff
@@ -51,20 +51,22 @@ impl<N: Network> ConfirmedTransaction<N> {
...
-        let (num_initialize_mappings, num_update_key_values) =
-            finalize_operations.iter().try_fold((0, 0), |(init, update), operation| match operation {
-                FinalizeOperation::InitializeMapping(..) => Ok((init + 1, update)),
-                FinalizeOperation::UpdateKeyValue(..) => Ok((init, update + 1)),
+        // Count the number of `InitializeMapping` and `*KeyValue` finalize operations.
+        let (num_initialize_mappings, num_key_values) =
+            finalize_operations.iter().try_fold((0, 0), |(init, key_value), operation| match operation {
+                FinalizeOperation::InitializeMapping(..) => Ok((init + 1, key_value)),
+                FinalizeOperation::InsertKeyValue(..) // At the time of writing, `InsertKeyValue` is only used in tests. However, it is added for completeness, as it is a valid operation.
+                | FinalizeOperation::RemoveKeyValue(..)
```
vicsn, line 63: Just for the record, this may be another theoretical adjustment to transaction validity.

synthesizer/process/src/stack/execute.rs
vicsn: We can consider removing, just to reduce the scope of the PR. It doesn't directly benefit program upgradability, and it doesn't seem consistent to allow `Edition` but not `NetworkID`.

synthesizer/program/src/lib.rs
acoglio: Given the fairly restrictive syntactic structure of AVM code, technically many of these "grammatical identifiers" do not need to be reserved, for parsing to be still unambiguous. For instance, `constructor` is only allowed to appear at the top level within a `program`, and every other thing that can appear there starts with something known and different from `constructor`:
```
program = *import
          cws %s"program" ws program-id ws ";"
          1*( mapping / struct / record / closure / function )
          cws
```
Note that no "regular identifier" can appear there.

On the other hand, it's cleaner and less confusing to reserve anything that has a special meaning.

synthesizer/program/src/logic/instruction/operand/parse.rs
```diff
@@ -28,6 +28,13 @@ impl<N: Network> Parser for Operand<N> {
...
             map(tag("block.height"), |_| Self::BlockHeight),
             map(tag("network.id"), |_| Self::NetworkID),
+            // Note that `Operand::Checksum` and `Operand::Edition` must be parsed before `Operand::ProgramID`s, since an edition or checksum may be prefixed with a program ID.
+            map(pair(opt(terminated(ProgramID::parse, tag("/"))), tag("checksum")), |(program_id, _)| {
+                Self::Checksum(program_id)
+            }),
+            map(pair(opt(terminated(ProgramID::parse, tag("/"))), tag("edition")), |(program_id, _)| {
+                Self::Edition(program_id)
+            }),
             // Note that `Operand::ProgramID`s must be parsed before `Operand::Literal`s, since a program ID can be implicitly parsed as a literal address.
```
acoglio, line 38: Is this comment out of date? Literal::parse() doesn't handle program IDs, which are handled by the explicit ProgramID::parse() below.

acoglio:
The parsing looks good. These are matching changes to the ABNF grammar:
```
operand = ...
        / program-id ; if not followed by "/"
        / [ program-id "/" ] %s"checksum"
        / [ program-id "/" ] %s"edition"
        / ...

constructor = cws %s"constructor" ws ":" *command

program = ...
          1*( constructor / mapping / struct / record / closure / function )
          ...
```

According to some discussion in this PR, we expect a change of the new rule to
```
constructor = cws %s"constructor" ws ":" 1*command
```
i.e. at least one command in a constructor.

acoglio (approved):
I just reviewed the parsing code.

synthesizer/program/src/constructor/mod.rs
```diff
@@ -0,0 +1,205 @@
...
+            // Ensure the branch target does not reference an earlier position.
+            ensure!(!self.positions.contains_key(position), "Cannot branch to an earlier position '{position}'");
+        }
+
+        // Check if the command is a position command.
+        if let Some(position) = command.position() {
+            // Ensure the position is not yet defined.
+            ensure!(!self.positions.contains_key(position), "Cannot redefine position '{position}'");
+            // Ensure that there are less than `u8::MAX` positions.
+            ensure!(self.positions.len() < u8::MAX as usize, "Cannot add more than {} positions", u8::MAX);
```
ljedrz, line 91: I think this should either be `<=`, or the error message should be adjusted

ljedrz: ah no, it's for adding; nvm

synthesizer/program/src/constructor/parse.rs
```diff
@@ -0,0 +1,128 @@
...
+        // Parse the colon ':' keyword from the string.
+        let (string, _) = tag(":")(string)?;
+
+        // Parse the commands from the string.
+        let (string, commands) = many0(Command::parse)(string)?;
+
+        map_res(take(0usize), move |_| {
+            // Initialize a new constructor.
+            let mut constructor = Self::new();
+            if let Err(error) = commands.iter().cloned().try_for_each(|command| constructor.add_command(command)) {
```
ljedrz, line 37:
```suggestion
            if let Err(error) = commands.into_iter().try_for_each(|command| constructor.add_command(command)) {
```
if possible, but seems like it to me

ljedrz (approved):
Reviewed the engineering aspects, left a suggestion.

synthesizer/program/src/lib.rs
d0cd: For AVM code, we should prefer to be as restrictive as possible.

d0cd:
> Can you make an issue drafting documentation for the constructor and new opcodes, ignoring program upgradability for now: https://github.com/aleoNet/welcome/issues/new

@vicsn since we'd like to do updatability all in one go, does it make more sense to add docs at once?

synthesizer/process/src/stack/finalize_types/initialize.rs
d0cd: We could certainly do that. I preferred not to since it felt a little "unclean" to me. 
My own opinion, these error messages have rarely helped me in debugging.
I'll leave it to the reviewers to arbitrate.

synthesizer/process/src/stack/finalize_types/initialize.rs
d0cd: This is the only one that stands out to me.
Are there others we should consider?

synthesizer/process/src/stack/execute.rs
d0cd: I think we want to allow them.
The use case I had in mind was:
- program passes either `Edition` or `Checksum` to finalize
- finalize checks that it matches the expected on-chain version
A developer could use this pattern to invalidate "old" transactions from users, providing a cleaner UX.
Obviously a malicious developer can still muck about.

ledger/block/src/transactions/confirmed/mod.rs
d0cd: Are there any checks we should put in place here?

synthesizer/program/src/logic/instruction/operand/parse.rs
d0cd: `Literal::parse` does handle `ProgramID`s [[source]](https://github.com/ProvableHQ/snarkVM/blob/06964e8c81ee81ff8c0f2062234196151d6767ba/console/program/src/data/literal/parse.rs#L41).
This was needed to allow users to pass in program IDs as address value arguments but also allow program IDs to be explicitly declared as `Operand`s.

ledger/block/src/transactions/confirmed/mod.rs
vicsn: On additional thought, no need for checks, but can you document in the docblocks of `{accepted, rejected}_{deploy, execute}` whether `finalize_operations` contains e.g. operations from `Fee`, `constructor`, etc.

synthesizer/program/src/constructor/parse.rs
d0cd: In making the change, I run into an issue that the encapsulating closure is `FnMut`.

synthesizer/program/src/logic/instruction/operand/parse.rs
acoglio: Ah, I looked at [the Literal::parse() under circuits](https://github.com/ProvableHQ/snarkVM/blob/06964e8c81ee81ff8c0f2062234196151d6767ba/circuit/program/src/data/literal/mod.rs#L158), instead of the one under console. (I didn't know/remember there were two.)

synthesizer/program/src/logic/command/contains.rs
d0cd: @ljedrz during review, specific attention was called to this change. Do you see any issues or potential vulnerabilities?

ljedrz: This change means that it would now suffice for the mapping to be part of the ongoing atomic operation, i.e. it doesn't have to be _physically_ in the persistent storage yet. As long as there is no way of undoing/changing that mapping throughout the rest of this specific storage operation, this should technically be fine, since at the end of it, both the mapping and all the remaining pending storage entries will have been persisted.

d0cd: In the atomic operation (for constructors), the mapping can have entries added or removed. If the execution of the constructor fails, then then entire batch including the initialization of the mappings should be aborted.

ljedrz: this is how it works for all the atomic operations, and this change doesn't affect it

synthesizer/program/src/lib.rs
d0cd: **Update.** I previously verified that these keywords were not used at the top level. However, I later found that `edition` was used inside a record definition. @acoglio and I believe that it is not necessary for `checksum` and `edition` to be keywords and they have since been removed.

synthesizer/program/src/constructor/parse.rs
d0cd: Note. The negative tests were added directly to `feat/program-updatability`

synthesizer/process/src/cost.rs
vicsn: Can you draft a snarkOS PR? :)

d0cd: https://github.com/ProvableHQ/snarkOS/pull/3587

synthesizer/program/src/lib.rs
vicsn: I think we can close this thread in favour of resolving this on the snarkOS level: https://github.com/ProvableHQ/snarkOS/pull/3587#issuecomment-2788586686

d0cd: The more I think about this, the more I find it tricky to enforce `constructor` being in the reserved keywords. Unless a quorum during migration is synchronous, there is always a chance for someone to create some confusion by throwing in a program that uses "constructor". Per @acoglio's comment, it might be easiest to not have it as a keyword.

vicsn: @d0cd the way this is solved in https://github.com/ProvableHQ/snarkOS/pull/3587#issuecomment-2788586686 is to not sign transmissions containing programs containing the `constructor` keyword until the migration date.

Up to you if you want to make it a keyword or not.

d0cd: I think its a little more complicated.
As of now, the `constructor` keyword is not restricted. 
Between now and when the upgrade is rolled out, someone can deploy a program that uses constructor as  mapping name, function name, etc.
Once it becomes a keyword, then that program is no longer considered a valid one.

vicsn: Thank you for clarifying, I agree that would be very confusing. If we can avoid having it as keyword that would be best.

d0cd: Removed the reserved keyword.

howardwu:
@vicsn @d0cd -- pushing back on removing the reservation of keyword `constructor`.

It is easy to remove the restriction in the future, however it is difficult to introduce later.

I would advocate for adding it even sooner to "bookmark" it in an earlier release to this feature if possible.

d0cd:
> @vicsn @d0cd -- pushing back on removing the reservation of keyword `constructor`.
> 
> It is easy to remove the restriction in the future, however it is difficult to introduce later.
> 
> I would advocate for adding it even sooner to "bookmark" it in an earlier release to this feature if possible.

If it is to be added, then +1 on adding it sooner.
Currently, there are no programs on `testnet` or `mainnet` that contain the keyword constructor (scanned as of 12PM PST 4/15/25)
The process of reserving this keyword will be challenging. Some options included a (nearly) synchronous upgrade of quorum, or versioning the reserved keywords before and after an upgrade height. The fundamental issue is that there is nothing preventing the use of the keyword ATM.

vicsn:
> @vicsn @d0cd -- pushing back on removing the reservation of keyword constructor.

>  The process of reserving this keyword will be challenging. Some options included a (nearly) synchronous upgrade of quorum, or versioning the reserved keywords before and after an upgrade height. The fundamental issue is that there is nothing preventing the use of the keyword ATM.

The benefits of rushing in a synchronous upgrade are limited, let's add it to [Consensus V6](https://github.com/ProvableHQ/snarkVM/pull/2670) (see the PR description for release dates). @d0cd can you make:
- a snarkVM PR introducing the `constructor` keyword
~- a snarkOS PR disallowing **deployments** containing that keyword (similar as [proposed here](https://github.com/ProvableHQ/snarkOS/pull/3587#issuecomment-2788586686)). Note that even if a deployment with that keyword exists before V6, users can continue to create executions for it, so its not detrimental to UX.~

d0cd:
> The benefits of rushing in a synchronous upgrade are limited, let's add it to [Consensus V6](https://github.com/ProvableHQ/snarkVM/pull/2670) (see the PR description for release dates). @d0cd can you make:
> 
>     * a snarkVM PR introducing the `constructor` keyword
> 
>     * a snarkOS PR disallowing **deployments** containing that keyword (similar as [proposed here](https://github.com/ProvableHQ/snarkOS/pull/3587#issuecomment-2788586686)). Note that even if a deployment with that keyword exists before V6, users can continue to create executions for it, so its not detrimental to UX.

See [this](https://github.com/ProvableHQ/snarkVM/pulls/2685) PR.

synthesizer/program/src/lib.rs
d0cd: The `constructor` keyword has been added to the list of consensus restricted keywords in this PR
https://github.com/ProvableHQ/snarkVM/pull/2685

synthesizer/program/src/logic/command/contains.rs
d0cd: We also added the restriction that a program cannot be executed in the same block that it was upgraded in.
In this case, the execution will get aborted.

synthesizer/process/src/finalize.rs
vicsn: @d0cd make sure to go over all unresolved comments and copy them over in the new PR you're making.

d0cd: Gonna keep this PR, but I think we're all resolved in this one no?

d0cd:

PR #2655, opened 2025-03-19
[Feature] Support program upgradability.

d0cd:

## Motivation

This PR 
- supports program upgradability, by allowing programs with constructors to be updated. Constructors are immutable, allowing developers to implement custom update logic for their applications. 
- modifies the program checksum to use SHA256 and changes its type from `field` to a `[u8; 32u32]`
- strictly defines program equality to also include order
- adds an optional `program_checksum` field to `Deployment`, which implicitly versions the object. 
- updates the deployment ID calculation to use the program checksum. Benchmarks for the checksum calculation are included below.

## Test Plan

- Process-level tests were added to verify the syntactic restrictions on upgrades.
- VM/Ledger-level tests were added to verify upgrade semantics. 

## Benchmarks
**Run on a 2021 MBP w/ an M1 Max & 32GB RAM**
- SHA3_256 Hash - input size 1000 bytes - [3.4339 µs 3.4435 µs 3.4539 µs]
- SHA3_256 Hash - input size 10000 bytes - [28.346 µs 28.500 µs 28.747 µs]
- SHA3_256 Hash - input size 100000 bytes - [280.97 µs 282.03 µs 283.11 µs] **(max program size)**
- SHA3_256 Hash - input size 1000000 bytes -  [2.7798 ms 2.7846 ms 2.7907 ms]


ledger/store/src/helpers/memory/transaction.rs
```diff
@@ -92,6 +92,8 @@ impl<N: Network> TransactionStorage<N> for TransactionMemory<N> {
 pub struct DeploymentMemory<N: Network> {
     /// The ID map.
     id_map: MemoryMap<N::TransactionID, ProgramID<N>>,
+    /// The ID edition map.
+    id_edition_map: MemoryMap<N::TransactionID, u16>,
```
d0cd, line 96: Consider renaming these for more clarity.
@reviewers, any preferences?

ledger/store/src/transaction/deployment.rs
```diff
@@ -38,6 +38,9 @@ use std::borrow::Cow;
 pub trait DeploymentStorage<N: Network>: Clone + Send + Sync {
     /// The mapping of `transaction ID` to `program ID`.
     type IDMap: for<'a> Map<'a, N::TransactionID, ProgramID<N>>;
+    /// The mapping of `transaction ID` to `edition`.
+    /// Note: The `IDEditionMap` stores the non-zero editions for deployment transactions.
+    type IDEditionMap: for<'a> Map<'a, N::TransactionID, u16>;
     /// The mapping of `program ID` to `edition`.
     type EditionMap: for<'a> Map<'a, ProgramID<N>, u16>;
```
d0cd: Consider renaming `IDEditionMap` and `EditionMap` and `EditionMap` to `LatestEditionMap` for more clarity.

synthesizer/process/src/stack/deploy.rs
```diff
@@ -69,6 +70,18 @@ impl<N: Network> Stack<N> {
         deployment.check_is_ordered()?;
         // Ensure the program in the stack and deployment matches.
         ensure!(&self.program == deployment.program(), "The stack program does not match the deployment program");
+        // Ensure that edition in the stack and deployment matches.
```
d0cd, line 72: Do we want these checks here in addition to `VM::verify`?
If we can remove these, it's better since it keeps the deployment check "pure" w.r.t. the `Stack`

ledger/store/src/helpers/rocksdb/internal/id.rs
```diff
@@ -293,6 +294,8 @@ enum DataID {
     // Program
     ProgramIDMap,
     KeyValueMap,
+    // Nonzero editions
```
raychu86: Won't the current `insert` logic will insert editions of 0?

We just will be missing the 0 edition of the existing programs

ledger/store/src/transaction/deployment.rs
```diff
@@ -186,7 +197,9 @@ pub trait DeploymentStorage<N: Network>: Clone + Send + Sync {
         atomic_batch_scope!(self, {
             // Store the program ID.
             self.id_map().insert(*transaction_id, program_id)?;
-            // Store the edition.
+            // Store the edition for the transaction ID.
+            self.id_edition_map().insert(*transaction_id, edition)?;
```
raychu86: This will insert a 0 edition for new programs, but have a missing item for existing programs

ledger/store/src/helpers/rocksdb/internal/id.rs
d0cd: Ah yes, that's a bad comment. Addressing...

ledger/store/src/helpers/memory/transaction.rs
```diff
@@ -92,6 +92,8 @@ impl<N: Network> TransactionStorage<N> for TransactionMemory<N> {
...
     /// The edition map.
     edition_map: MemoryMap<ProgramID<N>, u16>,
```
raychu86, line 99: We may want to rename this to `latest_edition_map`. 

Should be fine on the rocksdb level, but will need to do some testing and ensure existing ledgers boot and operate properly.

ledger/store/src/transaction/deployment.rs
```diff
@@ -257,7 +283,7 @@ pub trait DeploymentStorage<N: Network>: Clone + Send + Sync {
         })
     }
 
-    /// Returns the transaction ID that contains the given `program ID`.
+    /// Returns the latest transaction ID that contains the given `program ID`.
     fn find_transaction_id_from_program_id(&self, program_id: &ProgramID<N>) -> Result<Option<N::TransactionID>> {
```
raychu86: It's already quite verbose, but we may want to consider renaming this to `find_latest_transaction_id_from_program_id`.

synthesizer/process/src/stack/helpers/check_update.rs
```diff
@@ -0,0 +1,109 @@
...
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+use super::*;
+
+impl<N: Network> Stack<N> {
+    /// Checks that the new program definition is a valid update.
```
raychu86: Could you list the invariants and conditions in the documentation?

synthesizer/src/vm/verify.rs
```diff
@@ -149,17 +149,61 @@ impl<N: Network, C: ConsensusStorage<N>> VM<N, C> {
...
                 ensure!(owner.verify(*deployment_id), "Invalid owner signature for deployment transaction '{id}'");
-                // Ensure the edition is correct.
-                if deployment.edition() != N::EDITION {
-                    bail!("Invalid deployment transaction '{id}' - expected edition {}", N::EDITION)
-                }
-                // Ensure the program ID does not already exist in the store.
-                if self.transaction_store().contains_program_id(deployment.program_id())? {
-                    bail!("Program ID '{}' is already deployed", deployment.program_id())
+                // Verify the program checksum, if it exists.
+                // TODO (@d0cd): Add requirement for the program checksum after migration height. Use similar logic to how the execution fee is checked.
```
raychu86: Migration height check for checksum might be sufficient since they are coupled, but we may want to check the absence of a constructor as well if the migration height is not reached.

d0cd: Good point, I believe it's needed

d0cd: I have an attempt migration PR [here](https://github.com/ProvableHQ/snarkVM/pull/2659). It needs work as setting V5 height invalidates many of tests that use constructors.

ledger/block/src/transaction/deployment/mod.rs
```diff
@@ -28,25 +28,39 @@ use console::{
...
+    /// This field creates a backwards-compatible implicit versioning mechanism for deployments.
+    /// Before the migration height where this feature is enabled, the checksum will **not** be allowed.
+    /// After the migration height where this feature is enabled, the checksum will be required.
+    program_checksum: Option<Field<N>>,
 }
 
+impl<N: Network> PartialEq for Deployment<N> {
+    fn eq(&self, other: &Self) -> bool {
+        self.edition == other.edition && self.verifying_keys == other.verifying_keys && self.program == other.program
+    }
```
d0cd, line 49: Note that deployment IDs will be different for the same deployment with and without a checksum.

ledger/block/src/transaction/merkle.rs
```diff
@@ -116,18 +116,23 @@ impl<N: Network> Transaction<N> {
     pub fn deployment_tree(deployment: &Deployment<N>) -> Result<DeploymentTree<N>> {
         // Ensure the number of leaves is within the Merkle tree size.
         Self::check_deployment_size(deployment)?;
-        // Retrieve the program.
-        let program = deployment.program();
+        // Prepare the header for the hash.
+        let header = match deployment.program_checksum() {
```
howardwu: Attention: The rules for a deployment ID's preimage are changed here.

Now, we use the `program_checksum` instead of the `program_id` in the transaction leaf.

ledger/block/src/transaction/deployment/mod.rs
```diff
@@ -57,13 +71,6 @@ impl<N: Network> Deployment<N> {
     pub fn check_is_ordered(&self) -> Result<()> {
         let program_id = self.program.id();
 
-        // Ensure the edition matches.
-        ensure!(
-            self.edition == N::EDITION,
```
vicsn, line 62: Should we keep this check if there is no checksum?

ledger/block/src/transaction/deployment/bytes.rs
```diff
@@ -45,16 +45,25 @@ impl<N: Network> FromBytes for Deployment<N> {
             verifying_keys.push((identifier, (verifying_key, certificate)));
         }
 
+        // If the deployment version is 2, read the program checksum.
+        let program_checksum = match version {
+            DeploymentVersion::V1 => None,
+            DeploymentVersion::V2 => {
+                let checksum: [u8; 32] = FromBytes::read_le(&mut reader)?;
```
vicsn: Should we validate the checksum here?

Example: validator receives a Deployment transaction from the network, and a MITM messes with the checksum - is it "blindly" stored in rocksdb or is the checksum validated somewhere? And if so, we might want to move the checksum validation here into the deserialization code for clarity and consistency with other id validation.

ledger/block/src/transaction/deployment/mod.rs
```diff
@@ -175,6 +187,29 @@ impl<N: Network> Deployment<N> {
     }
 }
 
+impl<N: Network> Deployment<N> {
+    /// Sets the program checksum.
+    /// Note: This method is intended to be used by the synthesizer **only**, and should not be called by the user.
+    #[doc(hidden)]
+    pub fn set_program_checksum_raw(&mut self, program_checksum: Option<[U8<N>; 32]>) {
+        self.program_checksum = program_checksum;
```
vicsn, line 205: Do we assume this is None before? If so we can check and fail if its not. If not we can just validate its correct.

If its only used by the test below you can make it testing only.

ledger/block/src/transaction/merkle.rs
vicsn: @d0cd I'd recommend making a deployment_tree_v1 and deployment_tree_v2.

Then we can make sure to only activate v2 from the new consensus height to avoid forking. Even if we decide to go for a synchronous update because not all validation logic can be feasibly guarded by the new consensus version, this change in particular will surely cause clients which didn't synchronously update not to sync anymore.

Less importantly, if someone wants to snarkify our consensus or validate some old data, they stil have access to v1 in snarkVM. We can consider deprecating all outdated logic one day, but currently we're keeping it around.

ledger/store/src/transaction/mod.rs
```diff
@@ -456,15 +503,29 @@ impl<N: Network, T: TransactionStorage<N>> TransactionStore<N, T> {
...
     }
 
+    /// Returns an iterator over the programs and editions, for all deployments.
+    pub fn programs_with_editions(
+        &self,
+    ) -> impl '_ + Iterator<Item = (Cow<'_, ProgramIDEdition<N>>, Cow<'_, Program<N>>)> {
+        self.storage.deployment_store().programs_with_editions()
+    }
+
     /// Returns an iterator over the `((program ID, function name, edition), verifying key)`, for all deployments.
```
vicsn, line 534: Are these all deployments or only the latest deployment for any given program? Are there any other method comments which require a similar update?

ledger/store/src/transaction/deployment.rs
```diff
@@ -213,15 +246,17 @@ pub trait DeploymentStorage<N: Network>: Clone + Send + Sync {
 
     /// Removes the deployment transaction for the given `transaction ID`.
     fn remove(&self, transaction_id: &N::TransactionID) -> Result<()> {
```
vicsn, line 248: Where is this called from? Do we have unit tests to clarify desired behaviour?

ledger/store/src/transaction/deployment.rs
```diff
@@ -310,17 +380,43 @@ pub trait DeploymentStorage<N: Network>: Clone + Send + Sync {
...
+                };
+                // Get the latest edition for the program ID.
+                let latest_edition = match self.get_latest_edition_for_program(&program_id)? {
+                    Some(edition) => edition,
+                    None => return Ok(None),
+                };
+                // Verify that the latest edition is zero.
+                // Prior to `ConsensusVersion::V5`, this must be the case because if a program is not in the `IDEditionMap` but exists,
+                // then it must have been deployed before program upgrades were introduced.
+                ensure!(latest_edition == 0, "Failed to get the edition for transaction '{transaction_id}'");
```
vicsn, line 425: Should this be guarded by consensus version? Or is this code never called after the upgrade happens?

synthesizer/process/src/finalize.rs
```diff
@@ -61,8 +81,8 @@ impl<N: Network> Process<N> {
 
             // Retrieve the program ID.
             let program_id = deployment.program_id();
-            // Iterate over the mappings.
-            for mapping in deployment.program().mappings().values() {
+            // Iterate over the mappings that must be initialized.
```
vicsn, line 84: Where are we running the constructor?

synthesizer/process/src/cost.rs
```diff
@@ -642,18 +642,18 @@ function dummy:",
 
             // Verify the deployment costs.
             let deployment_0 = process.deploy::<A, _>(&program_0, rng).unwrap();
-            assert_eq!(deployment_cost(&process, &deployment_0).unwrap(), (2442725, (815000, 577725, 50000, 1000000)));
+            assert_eq!(deployment_cost(&process, &deployment_0).unwrap(), (2474725, (847000, 577725, 50000, 1000000)));
```
vicsn, line 645: Constructors should only be possible after new consensus version to avoid forking

synthesizer/process/src/stack/deploy.rs
vicsn: Dont feel too strongly about it and would stick to what the current code does. ~Only thing I'm personally worried about is that we only compute the program_checksum once. Nevermind not an issue anymore with keccak.~

synthesizer/src/vm/finalize.rs
```diff
@@ -77,6 +77,9 @@ impl<N: Network, C: ConsensusStorage<N>> VM<N, C> {
         let unordered_aborted_transaction_ids: IndexMap<N::TransactionID, &String> =
             verification_aborted_transaction_ids.chain(speculation_aborted_transaction_ids).collect();
 
+        println!("Verification aborted transactions: {:?}", verification_aborted_transactions);
```
vicsn: Remove or guard by debug macro

synthesizer/src/vm/mod.rs
```diff
@@ -3117,9 +3139,412 @@ function adder:
...
+                Identifier::from_str("first").unwrap(),
+                &Plaintext::from_str("1field").unwrap(),
+            )
+            .unwrap()
+        else {
+            panic!("Expected a valid field.");
+        };
+        assert_eq!(field, Field::from_str("1field").unwrap());
+
+        // Fix what is wrong.
```
vicsn: Not sure I understand this comment, perhaps a leftover from development

synthesizer/src/vm/verify.rs
```diff
@@ -149,17 +149,79 @@ impl<N: Network, C: ConsensusStorage<N>> VM<N, C> {
...
+                        );
+                    }
                 }
-                // Ensure the program ID does not already exist in the store.
-                if self.transaction_store().contains_program_id(deployment.program_id())? {
-                    bail!("Program ID '{}' is already deployed", deployment.program_id())
+                // If the program checksum exists, then verify that it is correct.
+                if let Some(given_checksum) = deployment.program_checksum() {
+                    // Compute the expected checksum.
+                    let expected_checksum = deployment.program().to_checksum();
```
vicsn, line 179: Given how expensive this can get, I would not recompute - many of the checks in this file are redundant and also in this case, the checksum is already verified or computed at this point.

synthesizer/src/vm/deploy.rs
```diff
@@ -32,9 +32,15 @@ impl<N: Network, C: ConsensusStorage<N>> VM<N, C> {
         rng: &mut R,
     ) -> Result<Transaction<N>> {
         // Compute the deployment.
-        let deployment = self.deploy_raw(program, rng)?;
+        let mut deployment = self.deploy_raw(program, rng)?;
         // Ensure the transaction is not empty.
         ensure!(!deployment.program().functions().is_empty(), "Attempted to create an empty transaction deployment");
+        // Unset the checksum if the `CONSENSUS_VERSION` is less than `V5`.
```
vicsn: Hmm, if the program contains a constructor, it will still be rejected by the recipient node.

What about we "fail loudly" and return an error here, if the query is below V5 and the program contains a constructor?

synthesizer/src/vm/finalize.rs
```diff
@@ -786,6 +790,8 @@ impl<N: Network, C: ConsensusStorage<N>> VM<N, C> {
     /// - The transaction is producing a duplicate output
     /// - The transaction is producing a duplicate transition public key
     /// - The transaction is another deployment in the block from the same public fee payer.
+    /// - The transaction is an execution for a program following its deployment or redeployment in this block.
```
vicsn, line 790: I'm still in favour of aborting executions which reference a state root which is older than the latest program upgrade corresponding to any of the transitions in the execution.

vicsn (changes requested):

synthesizer/src/vm/verify.rs
vicsn: Nevermind, checksum is now computed using keccak

synthesizer/src/vm/deploy.rs
d0cd: The checksum must also be unset for programs without a constructor.
To clarify, the checksum is a partially-independent requirement for transactions after the migration.

synthesizer/process/src/finalize.rs
d0cd: See [`finalize_constructor`](https://github.com/ProvableHQ/snarkVM/blob/cacd25afd852c14ddba23464a8e2ecb9a525ebe9/synthesizer/process/src/finalize.rs#L2160)

ledger/block/src/transaction/deployment/mod.rs
d0cd: We assume this is `Some`.
The checksum is explicitly unset to `None` in `vm/deploy.rs` if the consensus height is below the migration height.
It is `#[doc(hidden)]` to discourage improper usage.

synthesizer/src/vm/mod.rs
d0cd: Leftover from development.

synthesizer/src/vm/finalize.rs
d0cd: I think this is a fair point. 
The lookup could be done by:
1. Look up the latest transaction ID that deployed or upgraded a program.
2. Look up the corresponding block height for the program.
3. Look up the corresponding block height for the state root.
4. Verify that the state root's block height is less than the program's last deployed/updated height.

One thing to consider is that this check would create a dependence on the state root for "non-record" transactions. Concretely, exchanges and other developer services tend to "fix" a state root to reduce the number of network calls needed for their transactions. It could change the UX for consumer software but is not an unreasonable thing to do.

I'll leave it to reviewers to arbitrate.

synthesizer/process/src/cost.rs
d0cd: The check is enforced in [`vm/verify.rs`](https://github.com/ProvableHQ/snarkVM/blob/cacd25afd852c14ddba23464a8e2ecb9a525ebe9/synthesizer/src/vm/verify.rs#L165).

ledger/store/src/transaction/deployment.rs
d0cd: I don't believe it needs to be. 
The lookup here is intended to be backwards compatible and the check is a sanity check.

Prior to V5, all deployments have the zero edition, are not upgradable (because they do not have a constructor), and do not have a checksum.
Only after V5 is the `id_edition_map` used.
This means that their corresponding entry in the `edition_map` must be zero.

Let me know if you buy this argument

ledger/store/src/transaction/deployment.rs
d0cd: This is only called when we roll back a block with `remove_last_n`.
There is a test for this specific function at the bottom of the file.

ledger/store/src/transaction/mod.rs
d0cd: This is for all deployments.
Does the comment need an update here?
We made updates to the comments but it could be the case that some were missed. Do you see any?

ledger/block/src/transaction/deployment/bytes.rs
d0cd: We could! The checksum is currently only validated in `vm/verify.rs`.
Since it's now been changed to a `SHA256` hash, it's fairly cheap to verify during deserialization.
Are there any other implications of adding it in?

ledger/block/src/transaction/deployment/mod.rs
d0cd: We should remove `N::EDITION` entirely.
However, since this is expected behavior, it wouldn't hurt to add a check that the edition is zero.
What do you think?

synthesizer/process/src/stack/deploy.rs
d0cd: By "stick to what the current code does", do you mean omit these checks?

ledger/block/src/transaction/deployment/mod.rs
vicsn: Agreed to remove `N::EDITION` but keep that check here!

ledger/block/src/transaction/merkle.rs
d0cd: The checksum is not allowed before the V5 height and required after.
Do you mean to explicitly refactor the code to have separate functions for `deployment_tree_v1` and `deployment_tree_v2` like how we do for execution costs?

ledger/block/src/transaction/deployment/bytes.rs
vicsn: Great, I think its a pure win. 

Only thing to ponder is whether to not serialize it to save on bandwidth, but e.g. `batch_id` is also serialized to let's leave it as is.

ledger/block/src/transaction/deployment/mod.rs
vicsn: Would e.g. `#[cfg(test)]` also work?

ledger/store/src/transaction/mod.rs
vicsn: No all good then, just double checking!

ledger/store/src/transaction/deployment.rs
vicsn: I see, thank you for explaining. I think a full migration to the new data maps would be better in the long run - but similar to how it took a while to clean up the `ConsensusVersion`'s, we can wait with this until we did this process a few times.

synthesizer/src/vm/finalize.rs
vicsn: Sounds great. I can't imagine this to be a major performance bottleneck.

However, we do need to check the # program for *each* transition. So maybe a cache is warranted.

CC @raychu86

synthesizer/process/src/stack/deploy.rs
vicsn: What I meant was to keep the same "style" of checks in the same places.

Let's just keep it in for now, don't think it hurts.

ledger/block/src/transaction/deployment/mod.rs
d0cd: No, it's needed in [`VM::deploy`](https://github.com/ProvableHQ/snarkVM/blob/cacd25afd852c14ddba23464a8e2ecb9a525ebe9/synthesizer/src/vm/deploy.rs#L41)

ledger/block/src/transaction/merkle.rs
vicsn: >  Do you mean to explicitly refactor the code to have separate functions for deployment_tree_v1 and deployment_tree_v2 like how we do for execution costs?

Indeed!

synthesizer/src/vm/finalize.rs
vicsn: One more reason why this is a good idea:

If a user's local state is behind the network, then they might use an outdated edition or checksum value as witness, and as a result, the proof will fail to verify with an extremely opaque error message.

Now, unfortunately, users are used to transactions failing for opaque reasons. But we can imagine a point where explorers surface this information to users, or developers see this in validator logs for debugging purposes. 

It would be a lot better if the user or developer would see e.g. "Execution failed due to the use of a state root older than the latest program upgrade."

ledger/block/src/transaction/merkle.rs
d0cd: See this commit: [0e274f1](https://github.com/ProvableHQ/snarkVM/pull/2655/commits/0e274f124a48fd47451d937c89fbe4c2aecc32ce)

synthesizer/src/vm/finalize.rs
vicsn: For the same reason as my last message above (verifier witness data may change after a program update), we should clear executions from the `partially_verified_transactions` cache when the underlying program for any of its transitions has been updated.

This should happen already before a block is finalized, so that if a program is updated, the next execution within the same block might now be rejected/aborted during finalization.

Maybe this ends up being superfluous if we add a separate explicit check for the state root, but it doesn't hurt to evict a cache early.

raychu86: This has been deliberated on a few times. I think it boils down to how we weight usability and user clarity. 

In Ethereum, (to my knowledge) there is no concept of this type of protection when working with upgradable programs via proxy contracts. The conclusion there was to shift the responsibility of safety onto the developer/user in favor of usability.

raychu86: Note that we are already aborting transactions that follow upgrades in the same block. An enforcement of state root "liveness" could cover that case as well.

IMO there is value in providing the safety for users because that is our preference as protocol developers, however it would be valuable to get community and developer feedback as well.

d0cd: I see both arguments for both sides.
As a user, I may want to "know what I am executing". 
On the other hand, a user might feel burdened by the need constantly get the latest state root, especially if the code does not deal with records.

Perhaps we can pull in ecosystem developers into this code review weigh in?

synthesizer/program/src/logic/command/get_or_use.rs
```diff
@@ -31,10 +31,8 @@ use console::{
 pub struct GetOrUse<N: Network> {
     /// The mapping.
     mapping: CallOperator<N>,
-    /// The key to access the mapping.
-    key: Operand<N>,
-    /// The default value.
-    default: Operand<N>,
+    /// The operands.
+    operands: [Operand<N>; 2],
```
raychu86, line 35: Why the change here? It felt more clear with the direct variables `key` and `default`.

At the very least, the documentation of `operands` should reflect the structure.

raychu86: Ditto with the other commands that have this unification

d0cd: This was needed to support the `operands()` on commands without allocating `Vec`s
The instructions are actually implemented in a similar way.
TBH the implementations of both instructions and commands needs some refactoring to standardize.

d0cd: For additional context, we need to iterate through all of the operands in a program and check that they don't match `Operand::Checksum` or `Operand::Edition` before a specific consensus version. It would be ideal if we can do this without cloning and allocating.

synthesizer/src/vm/finalize.rs
raychu86: My preference is to not add the state root check on transactions in favor of keeping the user experience cleaner and easier.

Particularly for now, unless we see sufficient pushback that favors the more conservative approach.

vicsn: > Perhaps we can pull in ecosystem developers into this code review weigh in?

I like it! I'll share a plan with you.

Note that in any case we will need to clear appropriate executions from `partially_verified_transactions` - though the location where it is cleared might depend on that user feedback.

d0cd: I made a branch with the check ([diff](https://github.com/ProvableHQ/snarkVM/compare/feat/program-updatability..feat/reject-old-execs-after-upgrade))
I'll make a note of it in the tracking issue as well.

d0cd: Refreshed the above branch to do the check in `should_abort_transaction`

howardwu: I genuinely would love to require a recent state root to ensure freshness.

However, I do feel for offline wallets, cold storage schemes, and HSMs (in the future). Namely, that they do not have the privilege or luxury to go and fetch live state like this on a continual basis. And a custodian servicing millions of addresses for their customers does not have the bandwidth to stay on top of live refreshes like this (with today's state of Aleo infrastructure and tooling).

As such, I strongly recommend keeping the design in its current form and re-evaluating down the line.

d0cd: I do agree that fetching live state is a pain for offline wallets, HSMs, etc. 
However, these solutions will already need to do this to spend records. 

The proposed design can use any state root after the the program has been upgraded.
Moreover, it's unlikely for programs to get upgraded frequently, even once a month would be a lot.
And so is unlikely to add meaningful overhead to existing requirement for state root fetching.

This restriction does improve security for users, limiting the ways a malicious developer could "rug" a user.

d0cd:

PR #2660, opened 2025-03-21
[Release] Mainnet 1.4.0

raychu86:
## Motivation

This PR merges the latest `testnet` branch in accordance with the release process and updates the crate version from `v1.3.0` to `v1.4.0`.

This corresponds with the latest `crates.io` release - https://crates.io/crates/snarkvm.


vicsn (approved):
NOTE: circleci caches are still at 1.3.0. but the benefit is negligible of updating that frequently.

PR #2665, opened 2025-03-25
Add more test-helpers for batch certificates

kaimast:
To write tests for [this PR for snarkOS](https://github.com/ProvableHQ/snarkOS/pull/3537) by @acoglio, I needed helper functions that can build certificates with valid signatures from a valid committee, not just signatures from random keys. This new functionality allows generating a committee and its private keys, as well as a helper that generates a new certificate using the committee keys.

Considering there are already test helpers in snarkVM, it makes sense to have these new helpers live there as well.  This way the functions can reuse existing functionality, so the overall changes are fairly minimal.
For example, `sample_batch_certificate_for_round_with_previous_certificate_ids` is now a two-line function that simply generates the keys. It then calls `sample_batch_certificate_for_round_with_committee`.

## Testing
The code only affects unit/integrations tests, and I verified that this does not break any tests in snarkVM or snarkOS.

## Notes
* If there is similar functionality already somewhere else in the code, please let me know. I couldn't find a way to do this with the existing helpers.

* These function names are getting a little long. A nicer way to implement this would be using the [Builder pattern](https://rust-unofficial.github.io/patterns/patterns/creational/builder.html), which I am happy to do, but that would involve more changes.

### DIFF
--- a/console/types/address/src/lib.rs
+++ b/console/types/address/src/lib.rs
@@ -36,6 +36,7 @@ pub use snarkvm_console_types_boolean::Boolean;
 pub use snarkvm_console_types_field::Field;
 pub use snarkvm_console_types_group::Group;
 
+/// The unique ID of an Aleo account, which is derived from its view (public) key.
 #[derive(Copy, Clone, PartialEq, Eq, Hash)]
 pub struct Address<E: Environment> {
     /// The underlying address.
--- a/console/types/address/src/lib.rs
+++ b/console/types/address/src/lib.rs
@@ -36,7 +36,7 @@ pub use snarkvm_console_types_boolean::Boolean;
 pub use snarkvm_console_types_field::Field;
 pub use snarkvm_console_types_group::Group;
 
-/// The unique ID of an Aleo account, which is derived from its view (public) key.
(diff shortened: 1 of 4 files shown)


console/types/address/src/lib.rs
```diff
@@ -36,6 +36,7 @@ pub use snarkvm_console_types_boolean::Boolean;
 pub use snarkvm_console_types_field::Field;
 pub use snarkvm_console_types_group::Group;
 
+/// The unique ID of an Aleo account, which is derived from its view (public) key.
```
raychu86: View keys are not intended to be publicly shared, I would be careful with the wording

raychu86:
Added a documentation nit, but the extra helper methods for sampling test instances are very welcome

console/types/address/src/lib.rs
kaimast: I added this when trying to understand how to derive the ID from a private key.

Do you have suggestions on how to rephrase it? I can also remove it, as it is not closely related to the PR.

kaimast: I will just remove the "(public)" part for now.

vicsn (approved):

PR #2668, opened 2025-03-28
Improves safety & readability on transaction root derivation logic

howardwu:
## Motivation

This PR improves safety and readability upon [PR 2577](https://github.com/ProvableHQ/snarkVM/pull/2577).

- [Rename `Deployment::len` to `Deployment::num_functions`](https://github.com/ProvableHQ/snarkVM/commit/02cfd2368f8b3254e3858c7562d94d754c6ab7f3)
    - Addresses [@ljedrz's comment](https://github.com/ProvableHQ/snarkVM/pull/2577/files#r1856200235) in the original PR
- [Removes `fee_index` from `Transaction::transaction_tree`](https://github.com/ProvableHQ/snarkVM/pull/2668/commits/33781045c7d6f5cfb01383d4050e12d87cadcf03)
    - This commit also adds a safety check to ensure `fee_index` does not exceed `MAX_TRANSITIONS`

## Related PRs

- https://github.com/ProvableHQ/snarkVM/pull/2577

### DIFF
--- a/ledger/block/src/transaction/deployment/mod.rs
+++ b/ledger/block/src/transaction/deployment/mod.rs
@@ -113,13 +113,8 @@ impl<N: Network> Deployment<N> {
     }
 
     /// Returns the number of program functions in the deployment.
-    pub fn len(&self) -> usize {
+    pub fn num_functions(&self) -> usize {
         self.program.functions().len()
-    }
-
-    /// Returns `true` if the deployment is empty.
-    pub fn is_empty(&self) -> bool {
-        self.program.functions().is_empty()
     }
 
     /// Returns the edition.
--- a/ledger/block/src/transaction/merkle.rs
+++ b/ledger/block/src/transaction/merkle.rs
@@ -111,22 +111,23 @@ impl<N: Network> Transaction<N> {
         mut deployment_or_execution_tree: TransactionTree<N>,
         fee: Option<&Fee<N>>,
     ) -> Result<TransactionTree<N>> {
+        // Retrieve the fee index, defined as the last index in the transaction tree.
+        let fee_index = deployment_or_execution_tree.number_of_leaves();
+        // Ensure the fee index is within the Merkle tree size.
+        ensure!(
+            fee_index <= N::MAX_FUNCTIONS,
+            "The fee index ('{fee_index}') in the transaction tree must be less than {}",
+            N::MAX_FUNCTIONS
+        );
+        // Ensure the fee index is within the Merkle tree size.
+        ensure!(
+            fee_index < Self::MAX_TRANSITIONS,
+            "The fee index ('{fee_index}') in the transaction tree must be less than {}",
+            Self::MAX_TRANSITIONS
+        );
+
         // If a fee is provided, append the fee leaf to the transaction tree.
         if let Some(fee) = fee {
-            // Retrieve the fee index, defined as the last index in the transaction tree.
-            let fee_index = deployment_or_execution_tree.number_of_leaves();
-            // Ensure the fee index is within the Merkle tree size.
-            ensure!(
-                fee_index < Self::MAX_TRANSITIONS,
-                "The fee index ('{fee_index}') in the transaction tree must be less than {}",
-                Self::MAX_TRANSITIONS
-            );
-            // Ensure the fee index is within the Merkle tree size.
-            ensure!(
-                fee_index <= N::MAX_FUNCTIONS,
-                "The fee index ('{fee_index}') in the transaction tree must be less than {}",
-                N::MAX_FUNCTIONS
-            );
             // Construct the transaction leaf.
             let leaf = TransactionLeaf::new_fee(u16::try_from(fee_index)?, **fee.transition_id()).to_bits_le();
             // Append the fee leaf to the transaction tree.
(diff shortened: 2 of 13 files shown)


ledger/block/src/transaction/merkle.rs
```diff
@@ -123,22 +141,16 @@ impl<N: Network> Transaction<N> {
...
-            .map(|(index, function)| {
-                // Construct the transaction leaf.
-                Ok(TransactionLeaf::new_deployment(
-                    u16::try_from(index)?,
-                    N::hash_bhp1024(&to_bits_le![header, function.to_bytes_le()?])?,
-                )
-                .to_bits_le())
-            })
-            .collect::<Result<Vec<_>>>()?;
+        let leaves = deployment.program().functions().values().enumerate().map(|(index, function)| {
```
raychu86: Unrelated to these changes, but I believe we can parallelize this and some of the other iterators

ledger/block/src/transaction/merkle.rs
```diff
@@ -112,6 +106,30 @@ impl<N: Network> Transaction<N> {
...
+        mut deployment_or_execution_tree: TransactionTree<N>,
+        fee: Option<&Fee<N>>,
+    ) -> Result<TransactionTree<N>> {
+        // If a fee is provided, append the fee leaf to the transaction tree.
+        if let Some(fee) = fee {
+            // Retrieve the fee index, defined as the last index in the transaction tree.
+            let fee_index = deployment_or_execution_tree.number_of_leaves();
+            // Ensure the fee index is within the Merkle tree size.
+            ensure!(
+                fee_index < Self::MAX_TRANSITIONS,
```
raychu86: If you want to bound it based on the transaction type, then `MAX_TRANSITIONS` is relevant only for executions. `MAX_FUNCTIONS` is the deployment related limit.

Right now we keep them all at 2^5, but if any are changed in the future, this check may not be correct.

ledger/block/src/transaction/merkle.rs
howardwu: Considering it will be at most `32`, I think the parallelization will actually slowdown the performance

ledger/block/src/transaction/merkle.rs
howardwu: Good point.

As this method could contain either a deployment tree OR an execution tree, should I check both `<= MAX_FUNCTIONS` and `< MAX_TRANSITIONS`?

raychu86: It's probably a good idea to do both checks. 

Alternatively, because we always know what type of tree we are passing into this function, we can do a conditional check (using a boolean or const generic)

ledger/block/src/transaction/deployment/mod.rs
```diff
@@ -113,15 +113,10 @@ impl<N: Network> Deployment<N> {
     }
 
     /// Returns the number of program functions in the deployment.
-    pub fn len(&self) -> usize {
+    pub fn num_functions(&self) -> usize {
```
vicsn, line 124: For consistency, can you also change the equivalent function for the execution? https://github.com/ProvableHQ/snarkVM/blob/staging/ledger/block/src/transaction/execution/mod.rs#L120

vicsn (approved):
LGTM, modulo Ray's and my open comments

ledger/block/src/transaction/merkle.rs
d0cd: Do we want the check `number_of_leaves < Self::MAX_{TRANSITIONS/FUNCTIONS}` do also be run when there is no fee for added safety?

Do `Self::deployment_tree` and `Self::execution_tree` need similar checks?

d0cd (approved):
Good with the overall changes! Left a question on additional safety checks.

ledger/block/src/transaction/deployment/mod.rs
howardwu: I would prefer to keep these two concepts unrelated. If Deployment requires is_empty, we can rediscuss bringing it back to len.

ledger/block/src/transaction/merkle.rs
howardwu: @raychu86 - added check here: 05f30a5

@d0cd - this is not necessary. The tree is correct by construction.

howardwu: After discussing offline with @d0cd, we've identified additional deployment size checks that should be added into the tree checks.

See [bfa667f](https://github.com/ProvableHQ/snarkVM/pull/2668/commits/bfa667f33a49f6e5349500ce3b63cb47e1a30842) for the updates.

howardwu:
This PR is ready for re-review and merging if LGTM'ed.

ledger/block/src/transaction/merkle.rs
```diff
@@ -178,20 +186,28 @@ impl<N: Network> Transaction<N> {
...
+            num_functions == verifying_keys.len(),
+            "Number of functions ('{num_functions}') and verifying keys ('{}') do not match",
             verifying_keys.len()
         );
+        // Ensure there are functions.
+        ensure!(num_functions > 0, "Deployment must contain at least one function");
+        // Ensure the number of functions is within the allowed range.
+        ensure!(
+            num_functions <= N::MAX_FUNCTIONS,
+            "Deployment must contain less than {} functions, found {num_functions}",
```
vicsn:
```suggestion
            "Deployment must contain at most {} functions, found {num_functions}",
```

vicsn (approved):

d0cd (approved):

PR #2669, opened 2025-03-28
Removes `PartialEq` and `Eq` on Map `V` traits

howardwu:
## Motivation

This PR removes `PartialEq` and `Eq` on Map `V` traits.

These properties are not needed, and reduces the dependencies required for an object to be supported by our maps.


raychu86 (approved):
LGTM. I did a check and there doesn't seem to be any case were we do comparisons directly on the maps (nor would we want to).

niklaslong (approved):
In principle, this change should be fine. One caveat is if we ever have a requirement to compare maps in tests, though there are workarounds like re-implementing the trait (only for tests) or relying on the implementations of the test values. LGTM 👍

ljedrz (approved):

vicsn:
The failing CI is related to the base branch - merging because its going into a feature branch.

PR #2672, opened 2025-03-31
Check that subDAGs are closed

kaimast:
## Motivation

This adds additional checks during synchronization to verify that subDAGs are well-formed.

## Tests and Related PR

The PR adds new unit tests, to ensure it works as expected, and a [PR in SnarkOS](https://github.com/ProvableHQ/snarkOS/pull/3552), to ensure it does not break anything in production.


ledger/src/check_next_block.rs
```diff
@@ -129,6 +132,40 @@ impl<N: Network, C: ConsensusStorage<N>> Ledger<N, C> {
...
+            .certificates()
+            .flat_map(|cert| cert.previous_certificate_ids().iter().map(|id| (cert.round() - 1, id)))
+            .filter(|(_, id)| !subdag_certs.contains(id))
+            .collect();
+
+        for (round, cert_id) in leaf_edges {
+            // Find the block for this certificate.
+            // We might check the same block multiple times, but, so far, the check
+            // simply ensures that the block is a valid ancestor.
+            match self.vm.block_store().get_block_for_certificate(cert_id, Some(round))? {
```
acoglio: I don't think we need to test the round, because elsewhere we check that each certificate in the sub-DAG has a quorum of signers, which means (under the usual fault tolerance assumptions) that at least one correct validator has ensured that the certificate's previous certificate IDs identify certificates in the round just before the certificate's round.

ledger/src/check_next_block.rs
```diff
@@ -129,6 +132,40 @@ impl<N: Network, C: ConsensusStorage<N>> Ledger<N, C> {
...
+                Some(prev_block) => {
+                    ensure!(
+                        prev_block.height() < block.height(),
```
acoglio: Should this condition be always satisfied? The next block should have height 1 + latest block height.

ledger/src/check_next_block.rs
```diff
@@ -129,6 +132,40 @@ impl<N: Network, C: ConsensusStorage<N>> Ledger<N, C> {
...
+                        "Leaf is pointing to a block that is not an ancestor"
+                    );
+                }
+                None => bail!("Leaf is pointing to certificate that is not associated with a previous block"),
+            }
+        }
+
```
acoglio, line 167: There is a rare but possible case in which we need to relax these checks: if the sub-DAG spans the full 100 Subdag::MAX_ROUNDS, it is possible that the certificates in the smallest round of the sub-DAG have dangling references to previous certificates, because when a sub-DAG is committed, at most 100 rounds starting from the anchor (and going backwards) are considered, as it has emerged from [recent discussion](https://github.com/ProvableHQ/snarkOS/pull/3564/files).

ledger/src/check_next_block.rs
kaimast: Yes, that makes sense. I will remove the check and add a note here.

ledger/src/check_next_block.rs
kaimast: Probably true. I don't think an extra sanity check here hurts, though?

acoglio: I suppose. Maybe we could add a comment saying that the expect the check to be redundant. I'll see if I can find if/where it gets checked outside this code.

acoglio: When we are at this point in the code, we have executed line 98 in this file, which calls `Block::verify`, which calls `Header::verify()`, which calls `Metadata::verify()`, which [checks the block height against the expected height](https://github.com/ProvableHQ/snarkVM/blob/40ace73fc3a1125518d73b9cf707f17217775d5b/ledger/block/src/header/metadata/verify.rs#L44), where the expected height is the previous block height + 1, calculated [here](https://github.com/ProvableHQ/snarkVM/blob/0a07c0bb66b4d477d80c89b58cfa14cec757445a/ledger/block/src/verify.rs#L160) in `Block::verify_authority()` (it seems odd that the expected height is calculated when verifying the authority, but that's a separate issue).

Any case, now it seems clear to me that the check `prev_block.height() < block.height()` would be redundant here, also given the invariant that the blocks in the blockchain have ordered strictly increasing heights. We could add a comment to that effect.

ledger/src/check_next_block.rs
kaimast: This should be addressed with the most recent change. I also added a test for it.

ledger/src/check_next_block.rs
kaimast: I am happy to remove it entirely, considering it is fully redundant.

kaimast:
This is ready to merge now.

The PR is quite big because I needed to add helper code to write more sophisticated tests.

ledger/src/check_next_block.rs
```diff
@@ -129,6 +134,46 @@ impl<N: Network, C: ConsensusStorage<N>> Ledger<N, C> {
...
+            return Ok(());
+        };
+
+        // Store the IDs of all certificates in this subDAG.
+        // This allows determining which edges point to other subDAGs/blocks.
+        let subdag_certs: HashSet<_> = subdag.certificate_ids().collect();
+        let leaf_edges: HashSet<_> = subdag
+            .certificates()
+            .flat_map(|cert| {
+                cert.previous_certificate_ids().iter().map(|prev_id| (cert.id(), cert.round() - 1, prev_id))
```
raychu86: It seems possible to store many instances of `prev_id` if multiple certificates point to it. 

This means we are doing multiple `get_block_for_certificate(prev_id)` DB lookups in the subsequent iterator.

kaimast: Good catch. I added that for debugging and forgot to remove it. Should be resolved now.

raychu86 (approved):
LGTM.

With each of these checks, we run the risk of invalidating existing blocks. Is there a pre-release sync of testnet/mainnet blocks to ensure the same behavior? @vicsn

vicsn:
> With each of these checks, we run the risk of invalidating existing blocks. Is there a pre-release sync of testnet/mainnet blocks to ensure the same behavior? @vicsn

Genesis to tip sync tests will be performed on all three networks before the next release.

We will soon also have a small automated sync test on a few hundred blocks for all three networks which we can run more often.

PR #2677, opened 2025-04-04
[Doc] Clarify starting round of committee.

acoglio:


ledger/committee/src/lib.rs
```diff
@@ -50,7 +50,7 @@ pub const MAX_DELEGATORS: u32 = 100_000u32;
 pub struct Committee<N: Network> {
     /// The committee ID, defined as the hash of the starting round, members, and total stake.
     id: Field<N>,
-    /// The starting round number for this committee.
+    /// The round number (of the block) at which this committee is created (always even).
```
kaimast, line 53: nit: shouldn't it be "was created"? If the committee struct is instantiated, I assume the committee exists?

kaimast: Is this the round when the committee becomes active? "created" is still a little confusing

kaimast: Should we also update the documentation for `Committee::starting_round()`? It's currently identical to the previous old doc comment (_"Returns the starting round number for this committee"_).

kaimast (changes requested):
Left two small questions/comments.

ledger/committee/src/lib.rs
acoglio: The committee becomes active 101 rounds after this one. For example, if block 8 contains bonding/unbonding transactions, those give rise to a committee, whose "starting round" is 8, whose first bonded round is 9 (i.e. the first round after the committee has been created), and whose first active round is 109 (i.e. where the committee runs consensus). They are three different rounds associated to the committee, and the term "starting" for round 8 is a bit confusing in my opinion. Actually, I don't believe we need the notion (and the data field) for the starting round at all, but that's a separate issue.

acoglio: Yes, pushed.

vicsn (approved):

PR #2679, opened 2025-04-04
fix: incorrect string interpolation in `TestRng::fixed`

0xsenty:


vicsn:
I ran a local test to verify, and it seems that the original code interpolates just fine. Looking forward to your future contributions, really appreciate the efforts here.

PR #2684, opened 2025-04-10
Store unconfirmed id in rejected transaction

### LINKED ISSUE #2683: [Feature] Rejected IDs should reference unconfirmed IDs
## 🚀 Feature

When a tx is rejected, its rejected id shows up in a block. The rejected id is made using the transition id of the fee.

It would be nice for users to be able to trace back if the original tx failed. As such, can we also serialize the original txid as part of the rejected transaction? Or do we have to expose a separate endpoint for this?

vicsn:
## Motivation

Currently, in order to know whether a transaction was rejected, one has to track and store its fee transition id, and match that to the rejected transaction's fee transition id. This requires exchanges, explorers and other users to track extra information. At least one exchange, one explorer and multiple developers have complained about this, and they continue to do so, showing its a real issue.

This PR suggests to store the `unconfirmed_id` in rejected transactions  from `ConsensusVersion::V6` onwards.

Closes https://github.com/ProvableHQ/snarkVM/issues/2683

## Test Plan

- [ ] [devnet] Run a validator and client with and without this PR in the same network before the migration height
- [ ] [devnet] Run a set of validators and clients with this PR as they surpass the migration height (without this PR they would halt)
- [ ] Gather early feedback from every major explorer that they can handle such a change.

### DIFF
--- a/ledger/block/src/transactions/rejected/serialize.rs
+++ b/ledger/block/src/transactions/rejected/serialize.rs
@@ -20,14 +20,29 @@ impl<N: Network> Serialize for Rejected<N> {
     fn serialize<S: Serializer>(&self, serializer: S) -> Result<S::Ok, S::Error> {
         match serializer.is_human_readable() {
             true => match self {
-                Self::Deployment(program_owner, deployment) => {
+                Self::Deployment(Some(unconfirmed_id), program_owner, deployment) => {
+                    let mut object = serializer.serialize_struct("Rejected", 4)?;
+                    object.serialize_field("type", "deployment")?;
+                    object.serialize_field("unconfirmed_id", &unconfirmed_id)?;
+                    object.serialize_field("program_owner", program_owner)?;
+                    object.serialize_field("deployment", deployment)?;
+                    object.end()
+                }
+                Self::Deployment(None, program_owner, deployment) => {
                     let mut object = serializer.serialize_struct("Rejected", 3)?;
                     object.serialize_field("type", "deployment")?;
                     object.serialize_field("program_owner", program_owner)?;
                     object.serialize_field("deployment", deployment)?;
                     object.end()
                 }
-                Self::Execution(execution) => {
+                Self::Execution(Some(unconfirmed_id), execution) => {
+                    let mut object = serializer.serialize_struct("Rejected", 3)?;
+                    object.serialize_field("type", "execution")?;
+                    object.serialize_field("unconfirmed_id", &unconfirmed_id)?;
+                    object.serialize_field("execution", execution)?;
+                    object.end()
+                }
+                Self::Execution(None, execution) => {
                     let mut object = serializer.serialize_struct("Rejected", 2)?;
                     object.serialize_field("type", "execution")?;
                     object.serialize_field("execution", execution)?;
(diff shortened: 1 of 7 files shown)


ledger/block/src/transactions/rejected/serialize.rs
```diff
@@ -20,14 +20,29 @@ impl<N: Network> Serialize for Rejected<N> {
     fn serialize<S: Serializer>(&self, serializer: S) -> Result<S::Ok, S::Error> {
         match serializer.is_human_readable() {
             true => match self {
-                Self::Deployment(program_owner, deployment) => {
+                Self::Deployment(Some(unconfirmed_id), program_owner, deployment) => {
```
miazn, line 23: Is there a way to take care of optional unconfirmed id in the original two match statements?
I think that surfacing this to the top level and creating different json fields based on an optional value adds a lot of complexity, and the new `unconfirmed_id` field should probably be present in all rejected transactions

vicsn: Just for context, we certainly can't do this for the bytes serialization, because then as soon as we'd upgrade we would not be able to communicate anymore with nodes which did not upgrade yet.

We could opt to introduce the `unconfirmed_id` JSON field immediately as soon as nodes upgrade to this version. But that also means that any incompatible client software **might** break at an unexpected time.

With this PR's design, at the designated V6 block height, there will be a migration, and from then onwards all validators and clients will use the `unconfirmed_id` field. And we could hypothetically migrate the historic rejected transaction objects in the database and always include `unconfirmed_id` once that migration was succesful.

raychu86:
I believe the `block_store.get_transaction(transaction_id)` should already account for this? If you pass in the original transaction ID, it will return a rejected transaction in response. 

If you pass in your transaction_id and it returns a `Fee` transaction, then you know your transaction was rejected.

The documentation can and should be improved, but here is the entry point of these values - https://github.com/ProvableHQ/snarkVM/blob/d1009ec23bf29660f6715e30dcdfb39f7108ae4c/ledger/store/src/block/mod.rs#L388-L393

raychu86:
For clarity: Is the problem that the exchanges need to pull the original transaction ID from the block states that have Rejected transactions (and they are unable to calculate the unconfirmed id themselves)?

ledger/block/src/transactions/rejected/serialize.rs
```diff
@@ -47,6 +62,12 @@ impl<'de, N: Network> Deserialize<'de> for Rejected<N> {
                 // Parse the rejected transaction from a string into a value.
                 let mut object = serde_json::Value::deserialize(deserializer)?;
 
+                // Try to parse the unconfirmed ID.
+                let contains_unconfirmed_id = object.get("unconfirmed_id").is_some();
+                let unconfirmed_id: Option<N::TransactionID> = contains_unconfirmed_id
+                    .then(|| DeserializeExt::take_from_value::<D>(&mut object, "unconfirmed_id"))
+                    .transpose()?;
```
niklaslong, line 69: Style option: this might make the check more explicit? 
```suggestion
                let unconfirmed_id: Option<N::TransactionID> = if object.get("unconfirmed_id").is_some() {
                    Some(DeserializeExt::take_from_value::<D>(&mut object, "unconfirmed_id")?)
                } else {
                    None
                };

```

niklaslong (commented):
Did a first pass, the rust looks good. Does this change have any impact on network messages sent by nodes (i.e. will we need a message version bump alongside the consensus version change)?

vicsn:
> For clarity: Is the problem that the exchanges need to pull the original transaction ID from the block states that have Rejected transactions (and they are unable to calculate the unconfirmed id themselves)?

That is correct. The usecase is as follows: "We have a service that parse blocks and record transactions. We would like to receive a rejected transaction, find the original txID and store the metadata. [...] We cannot scan through all the original txs."

> If you pass in your transaction_id and it returns a Fee transaction, then you know your transaction was rejected.

They were already aware but it doesn't work for their system.

ledger/block/src/transactions/rejected/serialize.rs
vicsn: I was so proud of this one Niklas.

howardwu:
Are we certain this is safe? This is easily malleable by any MITM party. It is part of the thinking for why we did not blindly include it originally.

vicsn:
Superseded by https://github.com/ProvableHQ/snarkOS/pull/3597

PR #2685, opened 2025-04-17
[Feat] Consensus-restricted keywords.

d0cd:
## Motivation

As programming standards evolve on Aleo, it would be useful to be able to enforce syntactic restrictions at the consensus layer. This PR:
- adds a mechanism for describing versioned syntax restrictions and enforces them in `check_transactions`.
- adds a CI runner with the `test` feature enabled for the `synthesizer` crate.

## Test Plan

A test was added verifying that syntax restrictions are not enforced before a given version height and that they are enforced after.

### DIFF
--- a/synthesizer/program/src/lib.rs
+++ b/synthesizer/program/src/lib.rs
@@ -574,7 +574,7 @@ impl<N: Network, Instruction: InstructionTrait<N>, Command: CommandTrait<N>> Pro
     // New keywords should be enforced through `RESTRICTED_KEYWORDS` instead, if possible.
     // Adding keywords to this list will require a backwards-compatible versioning for programs.
     #[rustfmt::skip]
-    const KEYWORDS: &'static [&'static str] = &[
+    pub const KEYWORDS: &'static [&'static str] = &[
         // Mode
         "const",
         "constant",
@@ -653,7 +653,7 @@ impl<N: Network, Instruction: InstructionTrait<N>, Command: CommandTrait<N>> Pro
     /// If the current consensus version is greater than or equal to the specified version,
     /// the keywords in the list should be restricted.
     #[rustfmt::skip]
-    const RESTRICTED_KEYWORDS: &'static [(ConsensusVersion, &'static [&'static str])] = &[
+    pub const RESTRICTED_KEYWORDS: &'static [(ConsensusVersion, &'static [&'static str])] = &[
         (ConsensusVersion::V6, &["constructor"])
     ];
 
--- a/synthesizer/program/src/lib.rs
+++ b/synthesizer/program/src/lib.rs
@@ -48,44 +48,47 @@ mod parse;
 mod serialize;
 
 use console::{
-    network::prelude::{
-        Debug,
-        Deserialize,
-        Deserializer,
-        Display,
-        Err,
-        Error,
-        ErrorKind,
-        Formatter,
-        FromBytes,
-        FromBytesDeserializer,
-        FromStr,
-        IoResult,
-        Network,
-        Parser,
-        ParserResult,
-        Read,
-        Result,
-        Sanitizer,
-        Serialize,
-        Serializer,
-        ToBytes,
-        ToBytesSerializer,
-        TypeName,
-        Write,
-        anyhow,
-        bail,
-        de,
-        ensure,
-        error,
-        fmt,
-        make_error,
-        many0,
-        many1,
-        map,
-        map_res,
-        tag,
-        take,
+    network::{
+        ConsensusVersion,
+        prelude::{
+            Debug,
+            Deserialize,
+            Deserializer,
+            Display,
+            Err,
+            Error,
+            ErrorKind,
+            Formatter,
+            FromBytes,
+            FromBytesDeserializer,
+            FromStr,
+            IoResult,
+            Network,
+            Parser,
+            ParserResult,
+            Read,
+            Result,
+            Sanitizer,
+            Serialize,
+            Serializer,
+            ToBytes,
+            ToBytesSerializer,
+            TypeName,
+            Write,
+            anyhow,
+            bail,
+            de,
+            ensure,
+            error,
+            fmt,
+            make_error,
+            many0,
+            many1,
+            map,
+            map_res,
+            tag,
+            take,
+        },
     },
     program::{Identifier, PlaintextType, ProgramID, RecordType, StructType},
 };
@@ -93,7 +96,7 @@ use console::{
 use indexmap::IndexMap;
 
 #[derive(Copy, Clone, PartialEq, Eq, Hash)]
-enum ProgramDefinition {
+pub enum ProgramDefinition {
     /// A program mapping.
     Mapping,
     /// A program struct.
@@ -154,6 +157,11 @@ impl<N: Network, Instruction: InstructionTrait<N>, Command: CommandTrait<N>> Pro
     /// Returns the ID of the program.
     pub const fn id(&self) -> &ProgramID<N> {
         &self.id
+    }
+
+    /// Returns the identifiers in the program.
+    pub fn identifiers(&self) -> &IndexMap<Identifier<N>, ProgramDefinition> {
+        &self.identifiers
     }
 
     /// Returns the imports in the program.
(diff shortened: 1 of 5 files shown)


d0cd:
@reviewers I'm not too happy with the name `RESTRICTED_KEYWORDS`, any ideas for a better one?

acoglio:
`RESERVED_KEYWORDS` is perhaps more common?

synthesizer/src/vm/verify.rs
```diff
@@ -161,6 +161,49 @@ impl<N: Network, C: ConsensusStorage<N>> VM<N, C> {
...
                     bail!("Program ID '{}' already exists", deployment.program_id());
                 }
+                // Enforce the syntax restrictions on the programs based on the current consensus version.
+                // Note: We intentionally do not enforce restrictions on the import names in case they were deployed before the restrictions were added.
+                let current_block_height = self.block_store().current_block_height();
+                let consensus_version = N::CONSENSUS_VERSION(current_block_height)?;
+                // Get all keywords that are restricted for the current version.
+                let keywords =
+                    Program::<N>::restricted_keywords_for_consensus_version(consensus_version).collect::<Vec<_>>();
+                // Check if the program name is a restricted keywords.
```
vicsn: Slight preference for moving this logic into its own function, perhaps `Program::<N>::contains_restricted_keywords(keywords)`

synthesizer/program/src/lib.rs
```diff
@@ -642,6 +650,15 @@ impl<N: Network, Instruction: InstructionTrait<N>, Command: CommandTrait<N>> Pro
         "type",
         "future",
     ];
+    /// A list of restricted keywords for Aleo programs.
+    /// These restrictions are used to enforce program hygiene.
+    /// Each entry is a tuple of the consensus version and a list of keywords.
+    /// If the current consensus version is greater than or equal to the specified version,
+    /// the keywords in the list should be restricted.
+    #[rustfmt::skip]
+    const RESTRICTED_KEYWORDS: &'static [(ConsensusVersion, &'static [&'static str])] = &[
```
vicsn: Can we document the relation to `KEYWORDS`?

And also add a docblock above `KEYWORDS` that this new const exists for keywords which were newly added.

vicsn (approved):
LGTM! Left two style suggestions.

synthesizer CI is failing when run with the `--test` feature because `test_migration_v3_maximum_validator_increase` assumes max validator set increases at consensus v3, whereas with the `test` feature enabled the max validator set is static at 100.

In other words, our usage of the `test` feature is a bit of an undocumented mess.


As a "quickfix", one option would be to use the same `sample_finalize_state` approach as in `test_migration_v3_maximum_validator_increase`, and then we can also get rid of introducing yet another CI job.

synthesizer/src/vm/verify.rs
raychu86: +1 on this. `check_transaction` is already way too large as it is

synthesizer/src/vm/verify.rs
```diff
@@ -161,6 +161,49 @@ impl<N: Network, C: ConsensusStorage<N>> VM<N, C> {
...
+                let program_name = deployment.program().id().name().to_string();
+                if keywords.iter().any(|keyword| program_name == *keyword) {
+                    bail!("Program name '{}' is a restricted keyword for the current consensus version", program_name)
+                }
+                // Check that all top-level program components are not restricted keywords.
+                for identifier in deployment.program().identifiers().keys() {
```
raychu86: Does this encompass branch position names?

Ditto for the original checks, were we checking branch position names with `is_reserved_keyword`?

d0cd: It doesn't and we weren't enforcing for `Position` or `Branch` instructions.

d0cd:
> `RESERVED_KEYWORDS` is perhaps more common?

I had mapped the concept of "reserved" keywords to the `KEYWORDS` array.
This feels more like syntactic recommendations to me.
Like linting...?

synthesizer/src/vm/verify.rs
d0cd: Good call out! Moving it back to `Program`, which should also simplify the code a bit more.

d0cd:
> As a "quickfix", one option would be to use the same `sample_finalize_state` approach as in `test_migration_v3_maximum_validator_increase`, and then we can also get rid of introducing yet another CI job.

If it's not too costly, it might be nice to have a CI run with the `test` feature enabled.
I noticed a few other tests ([1](https://github.com/ProvableHQ/snarkVM/blob/643bf03d9db60a95f1ec283ad57e1f17d24e3d43/synthesizer/src/vm/execute.rs#L374), [2](https://github.com/ProvableHQ/snarkVM/blob/643bf03d9db60a95f1ec283ad57e1f17d24e3d43/synthesizer/src/vm/execute.rs#L404), [3](https://github.com/ProvableHQ/snarkVM/blob/643bf03d9db60a95f1ec283ad57e1f17d24e3d43/synthesizer/src/vm/finalize.rs#L1913), [4](https://github.com/ProvableHQ/snarkVM/blob/643bf03d9db60a95f1ec283ad57e1f17d24e3d43/synthesizer/src/vm/verify.rs#L798), [5](https://github.com/ProvableHQ/snarkVM/blob/643bf03d9db60a95f1ec283ad57e1f17d24e3d43/synthesizer/src/vm/verify.rs#L892)) that also use it.

synthesizer/src/vm/verify.rs
d0cd: We do have to option start enforcing these in the same way at the VM layer.
I did a quick scan (using [this](https://github.com/ProvableHQ/aleo-program-regressions/pull/19)) and didn't find any programs on `testnet`, `canary`, or `mainnet` that have keywords in their positions. (Likely because Leo uses a very specific structure for the name)

synthesizer/program/src/lib.rs
```diff
@@ -642,6 +648,14 @@ impl<N: Network, Instruction: InstructionTrait<N>, Command: CommandTrait<N>> Pro
         "type",
         "future",
     ];
+    /// A list of restricted keywords for Aleo programs, enforced at the VM-level for program hygiene.
+    /// Each entry is a tuple of the consensus version and a list of keywords.
+    /// If the current consensus version is greater than or equal to the specified version,
+    /// the keywords in the list should be restricted.
+    #[rustfmt::skip]
+    pub const RESTRICTED_KEYWORDS: &'static [(ConsensusVersion, &'static [&'static str])] = &[
+        (ConsensusVersion::V6, &["constructor"])
```
vicsn, line 657: Given that our transaction validity rules change from the V6 consensus height onwards, we need to clear the `partially_verified_transactions` cache.

Given that this is a one-time action (as opposed to program upgrades which can happen more frequently), you can just copy this approach: https://github.com/ProvableHQ/snarkVM/blob/mainnet/synthesizer/src/vm/mod.rs#L440

@raychu86 what do you think about just automatically clearing the entire cache on every consensus version upgrade so we don't have to think about it anymore. This could be e.g. sth like `if N::CONSENSUS_VERSION_HEIGHTS.iter().map(|_,h|h).contains(block.height()) { clear the cache }`.

vicsn (changes requested):

synthesizer/program/src/lib.rs
```diff
@@ -660,6 +674,68 @@ impl<N: Network, Instruction: InstructionTrait<N>, Command: CommandTrait<N>> Pro
...
+        let keywords =
+            Program::<N>::restricted_keywords_for_consensus_version(consensus_version).collect::<IndexSet<_>>();
+        // Check if the program name is a restricted keywords.
+        let program_name = self.id().name().to_string();
+        if keywords.contains(&program_name.as_str()) {
+            bail!("Program name '{program_name}' is a restricted keyword for the current consensus version")
+        }
+        // Check that all top-level program components are not restricted keywords.
+        for identifier in self.identifiers.keys() {
+            if keywords.contains(identifier.to_string().as_str()) {
```
niklaslong, line 703: I wish we could avoid these allocations by storing the keywords as `Identifier`s but I don't think that's possible with the type's current implementations?

niklaslong (commented):
LGTM overall, pending the changes requested by other reviewers.

synthesizer/program/src/lib.rs
d0cd: Just to double check my understanding, it seems like `partially_verified_transactions` is only read once [here](https://github.com/ProvableHQ/snarkVM/blob/5d0d1745d05619e2d3aecdcf54cf65106e52f4b7/synthesizer/src/vm/verify.rs#L141).

If `is_partially_verified` is `true`, then an internal check in `check_fee` and `check_deployment_internal` are skipped. 

The check is added [here](https://github.com/ProvableHQ/snarkVM/blob/45c3ba4723d5fe0d119aec83dc6e3ac5abf5d324/synthesizer/src/vm/verify.rs#L167), so would that mean it's run every time, despite whether or not it is cached?

vicsn: You are right, I think I got this right the first time I reviewed but I forgot, thx for catching.

vicsn (approved):

synthesizer/program/src/lib.rs
d0cd: We could initialize the keywords as `Identifier`s and store them in an `IndexSet`.
You'll only have a fixed number of allocations.

raychu86 (approved):

synthesizer/program/src/lib.rs
d0cd: Added a benchmark, current impl is at ~4us on a 2021 MBP Pro M1 MAX with 32GB RAM.

d0cd: Actually, you're correct. The current impl wouldn't parse `address` as an identifier.

PR #2686, opened 2025-04-17
fix: avoid double signature check

joske:
## Motivation

Avoid double signature verification: https://github.com/ProvableHQ/snarkOS/issues/3558

## Test Plan
manually tested
## Related PRs

https://github.com/ProvableHQ/snarkOS/pull/3598

### DIFF
--- a/ledger/narwhal/batch-certificate/src/lib.rs
+++ b/ledger/narwhal/batch-certificate/src/lib.rs
@@ -53,17 +53,7 @@ impl<N: Network> BatchCertificate<N> {
 impl<N: Network> BatchCertificate<N> {
     /// Initializes a new batch certificate.
     pub fn from(batch_header: BatchHeader<N>, signatures: IndexSet<Signature<N>>) -> Result<Self> {
-        // Ensure that the number of signatures is within bounds.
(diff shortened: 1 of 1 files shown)


ledger/narwhal/batch-certificate/src/lib.rs
```diff
@@ -82,6 +72,20 @@ impl<N: Network> BatchCertificate<N> {
         // Return the batch certificate.
         Ok(Self { batch_header, signatures })
     }
+
+    pub fn check_signature_basic(batch_header: &BatchHeader<N>, signatures: &IndexSet<Signature<N>>) -> Result<()> {
```
kaimast, line 76: Could you document this function? It should be clear what it does compared to the checks in `BatchCertificate::from`. 

Maybe you could document `BatchCertificate::from` a little better as well.

kaimast (changes requested):
Looks mostly good to me. I added a note about additional documentation. Imho, every new public function should be documented.

Would it be hard to add unit tests, by the way? There might already be some tests for `BatchCertificate::from` that you could just adopt.

Final note: Have you thought about adding another `from_*` constructor instead? As of now, this new function is only called right before `BatchCertificate::from_unchecked`.

joske:
@kaimast I'm no longer assigned to Aleo, sorry

PR #2691, opened 2025-04-21
Filter records by ownership before decryption in `find_records`

christianwwwwwwww:
## Motivation

This change ensures that `find_records` filters out records not owned by the provided view key _before_ attempting decryption. Previously billions of records were fetched and decrypted, and then filtered for ownership. This adjustment avoids unnecessary computation and adds an extra layer of correctness in filtering.

## Test Plan

A new test `test_find_records_filters_by_ownership` was added to validate that records not owned by the view key are correctly excluded. Existing tests covering `find_records` behavior also continue to pass.


Antonio95:
This seems great! It's surprising ownership was not being checked in the `RecordsFilter::All` case: the enum's description makes it clear that should be restricted to to records owned by the account too.

I will be happy to greenlight this if @christianwwwwwwww or @vicsn can confirm the following:

Informally speaking, if `view_key` is not the owner of `record`, then
- `ledger.contains_tag(Record::tag(sk_tag, commitment))` is (with overwhelming probability) `false`
- `ledger.contains_serial_number(Record::serial_number(sk_tag, commitment))` is (with overwhelming probability) `false`

where, in both cases, `sk_tag` is derived from `view_key` in the usual way.

Antonio95:
I also see two failures inside the CI `main-workflow` which boil down to the following tests:

```
tests::test_bond_and_unbond_validator
transaction::fee::bytes::tests::test_bytes
transaction::serialize::tests::test_serde_json
transactions::confirmed::bytes::tests::test_bytes
transactions::serialize::tests::test_serde_json
transition::serialize::tests::test_bincode
```

@christianwwwwwwww would you mind looking into that? Perhaps syncing with the latest version of `staging` will help.

Antonio95:
@vicsn both commands of the failing windows workflow fail with
```
Chocolatey v2.4.1
Uninstalling the following packages:
rust

Too long with no output (exceeded 10m0s): context deadline exceeded
```

Is this a known issue unrelated to the PR or should it be addressed here?

Antonio95 (approved):
Contingent on the clarification above on the intended behaviour of `contains_tag` and `contains_serial_number`, great 👌

vicsn:
Great!

PR #2696, opened 2025-04-24
CI: migrate workflows to checkout v4

dizer-ti:
Bumps checkout to v4 for future-proofing against Node 20 runner updates. Workflows compile the same.


dizer-ti:
Hey ! [vicsn](https://github.com/vicsn)
Сould you review my PR when you have time?
Thanks

kaimast (approved):
LGTM!

I merged staging again to trigger another round of CI, but just realized that these GitHub actions are not run during CI. 

However, I do not see how this change could break anything.  We do not use any options for this action that could have been deprecated or removed.

vicsn (approved):

PR #2693, opened 2025-04-23
Feature to test consensus heights

kpandl:
## Motivation

This PR adds a `test_consensus_heights` feature to test consensus heights, as suggested [here](https://github.com/ProvableHQ/snarkOS/pull/3572#issuecomment-2809325130).

Should be merged together with https://github.com/ProvableHQ/snarkOS/pull/3572.

## Test Plan

Tested by running a local devnet and printing consensus heights in conjunction with this PR: https://github.com/ProvableHQ/snarkOS/pull/3572

## Related PRs

https://github.com/ProvableHQ/snarkOS/pull/3572

PR #2700, opened 2025-05-02
Don't change MAX_CERTIFICATES with the test_consensus_heights feature

kpandl:
## Motivation

The `test_consensus_heights` feature should only lower the consensus version heights, but not change the `MAX_CERTIFICATES`. This PR introduces such a fix.

## Related PRs

https://github.com/ProvableHQ/snarkVM/pull/2693
https://github.com/ProvableHQ/snarkOS/pull/3572


vicsn (approved):
LGTM, this reverts it to what it was before.

PR #2705, opened 2025-05-07
Avoid duplicate logs in fn staking_rewards

### LINKED ISSUE #2689: [Bug] `fn staking_rewards` prints massive amounts of duplicate logs
## 🐛 Bug Report

`fn staking_rewards` iteratess over all stakers, and then prints something for every staker every time if there's an issue, e.g. if the validator is not in the committee. When there's a ~100.000 stakers and ~25 validators, that design doesn't make sense.

More specifically, to display the following logs, we could just iterate over the validators once, e.g.:

```
// If the list of stakers is empty, there is no stake, or the block reward is 0, return the stakers.
if stakers.is_empty() || committee.total_stake() == 0 || block_reward == 0 {
    return stakers.clone();
}

for validator in committee.members() {
            // If the validator is not in the committee, skip the staker.
            let Some((validator_stake, _is_open, commission_rate)) = committee.members().get(validator) else {
                trace!("Validator {validator} is not in the committee - skipping {staker}");
            };
            // If the validator has more than 25% of the total stake, skip the staker.
            if *validator_stake > committee.total_stake().saturating_div(4) {
                trace!("Validator {validator} has more than 25% of the total stake - skipping {staker}");
            }
}

// Now continue as usual

// Compute the updated stakers.
cfg_iter!(stakers)
    .map(|(staker, (validator, stake))| {
```

kpandl:
## Motivation

There were cases of massive duplicated logs in `fn staking_rewards`, since we have ~100.000 stakers and ~25 validators, as described here: https://github.com/ProvableHQ/snarkVM/issues/2689

This PR addresses it by doing checks before processing all stakers:
* Retrieve and log invalid validators (invalid commission rates, > 25% of total stake)
* Retrieve and log validators that are not in the committee

As a result, these things will only be logged once per validator, not per staker. The level of the logs remains unchanged.

Closes https://github.com/ProvableHQ/snarkVM/issues/2689


synthesizer/src/vm/helpers/rewards.rs
```diff
@@ -51,27 +51,42 @@ pub fn staking_rewards<N: Network>(
         return stakers.clone();
     }
 
+    // First, validate all validators and create a map of valid validators.
+    let valid_validators: IndexMap<_, _> = committee
+        .members()
+        .iter()
+        .filter_map(|(validator, (validator_stake, _, commission_rate))| {
```
vicsn: Nit: please stick to original formatting, use `_is_open`.

synthesizer/src/vm/helpers/rewards.rs
```diff
@@ -51,27 +51,42 @@ pub fn staking_rewards<N: Network>(
...
+
     // Compute the updated stakers.
     cfg_iter!(stakers)
         .map(|(staker, (validator, stake))| {
-            // If the validator is not in the committee, skip the staker.
-            let Some((validator_stake, _is_open, commission_rate)) = committee.members().get(validator) else {
-                trace!("Validator {validator} is not in the committee - skipping {staker}");
+            // If the validator is not in the valid validators list, skip the staker.
+            let Some((validator_stake, commission_rate)) = valid_validators.get(validator) else {
+                // No logging here to avoid log spam.
```
vicsn: This is a bad comment, log spam refers to previous behaviour. Always think about a future developer reading this, not yourself.

synthesizer/src/vm/helpers/rewards.rs
```diff
@@ -51,27 +51,42 @@ pub fn staking_rewards<N: Network>(
...
+                trace!("Validator {validator} has more than 25% of the total stake - skipping all stakers");
+                return None;
+            }
+            Some((*validator, (*validator_stake, *commission_rate)))
+        })
+        .collect();
+
+    // Pre-check for validators not in committee.
+    let mut non_committee_validators = std::collections::HashSet::new();
+    for (_, (validator, _)) in stakers {
```
vicsn: Iterating a second time over stakers is a massive performance penalty. We should collect the appropriate information during the existing parallelized iteration below, and print relevant info after the iteration.

vicsn: Consider returning and collecting validators which were not found from the iteration...

Is it possible to use a dynamic `.filter()` on the iteration so we start automatically skipping respective stakers?

vicsn (changes requested):

synthesizer/src/vm/helpers/rewards.rs
raychu86: +1 to victor's point here. This check can be done in the following `cfg_iter!(stakers)` iterator rather than having it's own (which is quite expensive).

Adding this log + check with a `continue` in the existing iterator should be sufficient.

kpandl: I think the `.filter()` approach would require some caution, since it would cause tests like `test_staking_rewards_to_validator_not_in_committee` to fail, which assert that no staker was dropped.

Unless we want to change this behavior, I see 3 main options - using a fold/reduce pattern (large diff), using a Mutex+HashSet (small diff but potentially some performance penalty due to locks), or using a DashSet (new crate).

I implemented the Mutex+HashSet option for now, but happy to discuss if other options would be preferable: https://github.com/ProvableHQ/snarkVM/pull/2705/commits/3d8b55df5e195a3c4d6c6ad9507411f8aa8fdbbc

synthesizer/src/vm/helpers/rewards.rs
kpandl: addressed here: https://github.com/ProvableHQ/snarkVM/pull/2705/commits/bc4a392677c6c0029c81058ac63540a423871f6c

synthesizer/src/vm/helpers/rewards.rs
kpandl: addressed here: https://github.com/ProvableHQ/snarkVM/pull/2705/commits/bc4a392677c6c0029c81058ac63540a423871f6c

synthesizer/src/vm/helpers/rewards.rs
vicsn: Cc @ljedrz for design

ljedrz: I wouldn't expect this to be a bottleneck in the current implementation; one thing we could do to improve its performance with minimal changes would be to pre-allocate the expected/maximum capacity so as not to have to reallocate while locked.

kpandl: addressed here: https://github.com/ProvableHQ/snarkVM/pull/2705/commits/a669a35fbb6682db671f7c844b6cb1469e8f5b90

synthesizer/src/vm/helpers/rewards.rs
```diff
@@ -51,27 +51,46 @@ pub fn staking_rewards<N: Network>(
...
+            }
+            Some((*validator, (*validator_stake, *commission_rate)))
+        })
+        .collect();
+
+    // Track validators not in committee.
+    // Pre-allocating with an expected capacity prevents reallocation while the mutex is held.
+    let hashset_capacity = committee.members().len();
+    let missing_validators =
+        std::sync::Mutex::new(std::collections::HashSet::<Address<N>>::with_capacity(hashset_capacity));
```
vicsn: At 100k stakers, a BTreeSet will be meaningfully faster than a HashSet

ljedrz: is hashing this expensive here? otherwise a hashset should be faster for both insertions and lookups.

kpandl: I agree `BTreeSet` is an interesting solution when the set size is small and hashing cost is high. `BTreeSet` would need `Ord` on `Address`.

Generally, we already call a hash function for all stakers in `valid_validators.get(validator)`, and the second hash call happens only for delegators whose validator is not in `valid_validators`. Assuming there is only a small percentage of stakers with missing validators, the overhead should be small.

Let me know if you’d prefer I make that change.

vicsn: Just ran a benchmark, BTreeSet is ~twice as fast, the difference is indeed relatively small because Address is a small type so hashing is fast.

https://play.rust-lang.org/?version=stable&mode=debug&edition=2024&gist=32e94e12522f49e996d258625cfb1d13

Still I think this'll be worth it due to the large amount of stakers and data.

More importantly as Konstantin points out, valid_validators be turned into a BTreeSet if the ordering doesn't matter.

We should not impl Ord, but can we use to_byte_le to compare the Adress bytes? Or to_x_coordinate if there are no additional costly checks in that call.

vicsn: Made an issue: https://github.com/ProvableHQ/snarkVM/issues/2738

vicsn (approved):

synthesizer/src/vm/helpers/rewards.rs
```diff
@@ -51,27 +51,46 @@ pub fn staking_rewards<N: Network>(
...
     cfg_iter!(stakers)
         .map(|(staker, (validator, stake))| {
-            // If the validator is not in the committee, skip the staker.
-            let Some((validator_stake, _is_open, commission_rate)) = committee.members().get(validator) else {
-                trace!("Validator {validator} is not in the committee - skipping {staker}");
+            // If the validator is not in the valid validators list, skip the staker.
+            let Some((validator_stake, commission_rate)) = valid_validators.get(validator) else {
+                // Log validator not in committe.
+                if !committee.members().contains_key(validator) {
+                    let mut logged = missing_validators.lock().unwrap();
```
niklaslong: Why use a `std::sync::Mutex` here instead of `parking_lot`?

kaimast (changes requested):
See my comment on `log_enabled`.

synthesizer/src/vm/helpers/rewards.rs
```diff
@@ -51,27 +51,46 @@ pub fn staking_rewards<N: Network>(
...
+
     // Compute the updated stakers.
     cfg_iter!(stakers)
         .map(|(staker, (validator, stake))| {
-            // If the validator is not in the committee, skip the staker.
-            let Some((validator_stake, _is_open, commission_rate)) = committee.members().get(validator) else {
-                trace!("Validator {validator} is not in the committee - skipping {staker}");
+            // If the validator is not in the valid validators list, skip the staker.
+            let Some((validator_stake, commission_rate)) = valid_validators.get(validator) else {
+                // Log validator not in committe.
```
kaimast: Can you disable this block if we are at the default verbosity level (which is Debug IIRC)?

You can use [log::log_enabled!](https://docs.rs/log/latest/log/macro.log_enabled.html) for this.

kaimast (changes requested):

synthesizer/src/vm/helpers/rewards.rs
kaimast: (Sorry, I put the comment at the wrong location before. It seems like this is the only part we can skip.)

vicsn: I don't see `log_enabled` used anywhere else in snarkVM 🤔 If you think we need it maybe you can make a new issue for it?

synthesizer/src/vm/helpers/rewards.rs
kpandl: addressed here: https://github.com/ProvableHQ/snarkVM/pull/2705/commits/78fbf1392ff295a62a5b36d3733ed2fc6ab07afe

synthesizer/src/vm/helpers/rewards.rs
kaimast: I haven't seen any other parts in the code where we do additional computation just to log something.

If you use `log::trace!(..)` everything in the macro will already be skipped if the trace loglevel is not enabled, but if you perform computation beforehand like here, it will run even if the log message is not printed, assuming the compiler or CPU doesn't somehow optimize this.

Grabbing a lock multiple times might not be that expensive, but it still seems like something we should avoid if possible.

vicsn (approved):

kaimast (approved):
I pushed a small change with my suggested improvement. It uses `tracing::enabled!` to only lock if we are actually printing the log message.

 Looks good to merge now.

PR #2710, opened 2025-05-14
Speed up subDAG check

kaimast:
This PR removes an unnecessary block retrieval. 

`Ledger::check_block_subdag_leaves` only needs to know that there exists a block for a given certificate ID, but the function does not inspect the block's content.

### DIFF
--- a/ledger/store/src/block/mod.rs
+++ b/ledger/store/src/block/mod.rs
@@ -773,8 +773,8 @@ pub trait BlockStorage<N: Network>: 'static + Clone + Send + Sync {
     }
 
     /// Returns true if there is a block that contains the specified certificate.
-    fn has_block_for_certificate(&self, certificate_id: &Field<N>) -> Result<bool> {
-        Ok(self.certificate_map().get_confirmed(certificate_id)?.is_some())
+    fn contains_block_for_certificate(&self, certificate_id: &Field<N>) -> Result<bool> {
+        self.certificate_map().contains_key_confirmed(certificate_id)
     }
 
     /// Returns the batch certificate for the given `certificate ID`.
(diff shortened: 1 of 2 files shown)


ledger/store/src/block/mod.rs
```diff
@@ -772,20 +772,9 @@ pub trait BlockStorage<N: Network>: 'static + Clone + Send + Sync {
...
-        };
-
-        // Retrieve the block hash.
-        let Some(block_hash) = self.get_block_hash(block_height)? else {
-            bail!("The block hash for block '{block_height}' is missing in block storage")
-        };
-        // Retrieve the block.
-        self.get_block(&block_hash)
+    /// Returns true if there is a block that contains the specified certificate.
+    fn has_block_for_certificate(&self, certificate_id: &Field<N>) -> Result<bool> {
```
raychu86: We try to use `contains_*` for these types of functions
```suggestion
    fn contains_block_for_certificate(&self, certificate_id: &Field<N>) -> Result<bool> {
```

ledger/store/src/block/mod.rs
```diff
@@ -772,20 +772,9 @@ pub trait BlockStorage<N: Network>: 'static + Clone + Send + Sync {
...
+        Ok(self.certificate_map().get_confirmed(certificate_id)?.is_some())
```
raychu86: There is no need to deserialize the entire certificate. We have a contains check in the map.
```suggestion
        Ok(self.certificate_map().contains_key_confirmed(certificate_id)?)
```

kaimast:
@raychu86 thanks for the feedback! I addressed all comments.

niklaslong (approved):

PR #2716, opened 2025-05-18
Fix invalid URL in utilities/derives README

VolodymyrBg:
Corrected the Aleo authors badge URL by removing an erroneous '../' prefix from the URL path.


ljedrz:
The updated badge still doesn't seem to work; I think the leading `/` is the issue.

VolodymyrBg:
> The updated badge still doesn't seem to work; I think the leading `/` is the issue.

You're right, sorry for this mistake

niklaslong (approved):

niklaslong:
Thank you for the contribution @VolodymyrBg!

PR #2722, opened 2025-05-21
[Feature] Reject old executions after an upgrade.

d0cd:

**THE CHANGES MADE IN THIS PR ARE REMOVED BY #2758.**
This PR exists for the purpose of keeping audits consistent.

This PR rejects executions whose state roots are before the latest upgrade. 
This ensures that users do not create executions for program logic that may have changed.
This does require that executions use "fresh" state roots, but the overhead should not be meaningful w.r.t to fetching state roots for records.


synthesizer/src/vm/finalize.rs
```diff
@@ -847,16 +850,82 @@ impl<N: Network, C: ConsensusStorage<N>> VM<N, C> {
...
+            if current_version >= ConsensusVersion::V5 {
+                // Track the maximum block height and the associated program ID.
+                let mut max_block_height = 0;
+                let mut latest_program = None;
+                // For each transition in the execution, get the block height at which the program was deployed or upgraded and update the maximum block height.
+                for transition in execution.transitions() {
+                    // Get the program ID.
+                    let program_id = transition.program_id();
+                    // If the program is `credits.aleo`, set the appropriate state and continue.
+                    if program_id.to_string() == "credits.aleo" {
```
vicsn, line 879: Familiar nit: can we instead compute the ID once so we dont convert to string every time?

synthesizer/src/vm/tests/test_vm_upgrade.rs
```diff
@@ -2247,3 +2247,142 @@ constructor:
...
+    let rng = &mut TestRng::default();
+
+    // Initialize a new caller.
+    let caller_private_key = sample_genesis_private_key(rng);
+    let caller_address = Address::try_from(&caller_private_key).unwrap();
+
+    // Initialize the VM at height 0.
+    let vm = sample_vm_at_height(0, rng);
+
+    // Generate two executions of `transfer_public`.
```
vicsn, line 2347: Actually can you generate these at e.g. height 1? I think there's a bug in the above code which will cause them to be aborted, because max_block_height is never updated for the credits.aleo case

vicsn (changes requested):

synthesizer/src/vm/tests/test_vm_upgrade.rs
d0cd: Can you elaborate on the bug? The test verifies that the executions are in fact accepted.

synthesizer/src/vm/finalize.rs
d0cd: Made an adjustment in the latest commit.
Let me know what you think.
I'm not sure I see another way that's meaningfully more efficient.

synthesizer/src/vm/tests/test_vm_upgrade.rs
vicsn: My bad, on second read I see the logic is correct, sorry for the delay

vicsn (approved):

synthesizer/src/vm/finalize.rs
kaimast: Could you do `transition.name() == credits && transition.network() == aleo` (where credits and aleo are constants) to avoid creating a string on every iteration?

d0cd:

PR #2723, opened 2025-05-22
fix: Check for expected block height early

kaimast:
This PR changes `Ledger::check_next_block` to not only check that no block with the given height exists, but also that the given block is the direct successor of the latest block in the ledger.

Currently, height is not checked until later in `Metadata::verify`. If you give `check_next_block` a block that is higher than the expected height, it will fail with something like "invalid previous state root", which is not super helpful to debug the problem.
This change also avoids an extra lookup in the storage for whether a block with the given height already exists. Instead, it uses the height from the previous block already cached in the ledger.

Additionally, this PR adds another height check in `Ledger::advance_to_next_block` to make sure nothing has changed between calling `check_next_block` and `advance_to_next_block`.

### DIFF
--- a/ledger/src/advance.rs
+++ b/ledger/src/advance.rs
@@ -91,6 +91,20 @@ impl<N: Network, C: ConsensusStorage<N>> Ledger<N, C> {
     pub fn advance_to_next_block(&self, block: &Block<N>) -> Result<()> {
         // Acquire the write lock on the current block.
         let mut current_block = self.current_block.write();
+        // Check again for any possible race conditions.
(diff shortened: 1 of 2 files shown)


ljedrz (approved):
LGTM :+1:; it's an easy win perf-wise, and I've certainly seen "invalid previous state root", and worried that it's a lot more severe, despite not experiencing any apparent issues.

niklaslong (approved):

Meshiest:
This change enforces custom genesis blocks to have exactly 4 public_to_private transactions

PR #2724, opened 2025-05-22
Don't panic when a rocksdb database can't be opened

### LINKED ISSUE #3641: [Bug] Nodes should not panic if the ledger is locked.
## 🐛 Bug Report

Nodes panic when the local RocksDB storage is locked. 

## Steps to Reproduce

1. Start a snarkOS node.
2. Pause it (using Ctrl+Z) or open another terminal window/tab.
3. Start a second snarkOS node in the same folder.

You will get a panic looking like this:
```
2025-05-20T22:02:02.116981Z  INFO snarkvm_ledger: Loading the ledger from storage...
thread 'main' panicked at ~/.cargo/git/checkouts/snarkvm-36b60a54cfa1484f/a87ceb9/ledger/store/src/helpers/rocksdb/internal/mod.rs:159:67:
Couldn't open the database: Error { message: "IO error: While lock file: ~/.aleo/storage/ledger-2/LOCK: Resource temporarily unavailable" }
```

## Expected Behavior
There should be a more human-readable output, so that the user can easily figure out how to resolve the problem.

For example, it could print something like *"Please make sure no other snarkOS process is running"*

## Notes

This panic/unwrap happens in snarkVM and might require some changes there to properly have the error reach the CLI module.

ljedrz:
Partially closes https://github.com/ProvableHQ/snarkOS/issues/3641.

### DIFF
--- a/ledger/store/src/helpers/rocksdb/internal/mod.rs
+++ b/ledger/store/src/helpers/rocksdb/internal/mod.rs
@@ -138,45 +138,48 @@ impl Database for RocksDB {
 
         // Retrieve the database.
         let db_path = aleo_std_storage::aleo_ledger_dir(network_id, &storage);
-        let database = DATABASES
-            .lock()
-            .entry(db_path.clone())
-            .or_insert_with(|| {
-                // Customize database options.
-                let mut options = rocksdb::Options::default();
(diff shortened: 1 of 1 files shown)


kaimast (approved):
Awesome! I thought this would need a much more invasive change.

niklaslong (approved):

PR #2728, opened 2025-05-23
Clear cache on all `ConsenusVersion` changes

raychu86:
## Motivation

This PR clears the `partially_verified_transactions` on all the `ConsenusVersion` changes rather than just `ConsenusVersion::V4`.

All `ConsenusVersion`s change consensus rules and regardless if it is transaction related or not, we clear the `partially_verified_transactions` out of abundance of caution. This will also prevent us from missing this step in the future for subsequent consensus versions.

### DIFF
--- a/synthesizer/src/vm/mod.rs
+++ b/synthesizer/src/vm/mod.rs
@@ -405,8 +405,8 @@ impl<N: Network, C: ConsensusStorage<N>> VM<N, C> {
                 // Unpause the atomic writes, executing the ones queued from block insertion and finalization.
                 #[cfg(feature = "rocks")]
                 self.block_store().unpause_atomic_writes::<false>()?;
-                // If the block advances to `ConsensusVersion::V4`, clear the partial verification cache.
-                if N::CONSENSUS_HEIGHT(ConsensusVersion::V4).unwrap_or_default() == block.height() {
+                // If the block advances to a new consensus version, clear the partial verification cache.
+                if N::CONSENSUS_VERSION_HEIGHTS.iter().any(|(_, height)| height == &block.height()) {
                     self.partially_verified_transactions().write().clear();
                 }
                 Ok(())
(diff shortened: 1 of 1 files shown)


vicsn (approved):
Thank you🙏

synthesizer/src/vm/mod.rs
```diff
@@ -405,8 +405,8 @@ impl<N: Network, C: ConsensusStorage<N>> VM<N, C> {
                 // Unpause the atomic writes, executing the ones queued from block insertion and finalization.
                 #[cfg(feature = "rocks")]
                 self.block_store().unpause_atomic_writes::<false>()?;
-                // If the block advances to `ConsensusVersion::V4`, clear the partial verification cache.
-                if N::CONSENSUS_HEIGHT(ConsensusVersion::V4).unwrap_or_default() == block.height() {
+                // If the block advances to a new consensus version, clear the partial verification cache.
+                if N::CONSENSUS_VERSION_HEIGHTS.iter().any(|(_, height)| height == &block.height()) {
```
vicsn, line 411: Performance nit: we can reverse iter and break the loop if the block height is below CONSENSUS_VERSION_HEIGHTS

vicsn: Its a nit now but will be essential once we have a hundred versions

niklaslong (approved):

synthesizer/src/vm/mod.rs
niklaslong: Since this is in the hot path, we may even want a flag to signal the consensus version change so we don't check on each call to `finalize`. For now, I think I'll add a TODO and merge.

niklaslong:
Aha! We have a perf regression, it seems 🤔 https://github.com/ProvableHQ/snarkVM/actions/runs/15294537889/job/43020394758

PR #2729, opened 2025-05-27
[Perf] Cache the state root in the Query

ljedrz:
A `Query` may currently look up the current state root more than once. This PR instead caches it and also allows a `Query` to be created with the given state root.

### DIFF
--- a/ledger/query/src/query.rs
+++ b/ledger/query/src/query.rs
@@ -13,6 +13,8 @@
 // See the License for the specific language governing permissions and
 // limitations under the License.
 
+use std::sync::{Arc, OnceLock};
+
 use crate::QueryTrait;
 use console::{
     network::prelude::*,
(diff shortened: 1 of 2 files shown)


d0cd:
Great to see this one come in! 
It could also be nice to have a "static" variant of `QueryTarget` that didn't need any network calls. 
It could be a separate issue, so as to not block this one.

raychu86:
Must needed start to Query improvements in general, also solves the issue where a Tx that requests multiple StateRoots may have issues executing if there is a difference in received state roots.

2 issues I see with the current approach:
1.  If a user uses the same Query for a while, then it will no longer fetch the new state paths. This means it won't ever be able to spend records that were produced AFTER the first state root was cached.
2. The StatePath request should have the same state root as the one used in TX generation. This would break that model and invalidate tx generation. The approach would likely have to use the state root derived from the state path or checks to enforce equivalency of the cached state root and state path derived state root.

ljedrz:
@raychu86 how about introducing an extra `bool` into the `Query` that would determine whether it caches the state root or not? It could be `false` by default.

raychu86:
Discussed with @niklaslong about a possible solution for this with additional optimization benefits. Couple takeaways:

1. New snarkOS REST endpoint that fetches multiple state paths at once (given a vec of commitments).
2. Add support for the above endpoint in the Query trait
3. If there are records in the execution - Wuery the state roots for all the records being spent before starting the execution and then pull from that cache during execution
4. If there are no records in the execution - Use the cached state root

raychu86:
This is likely partially addressed with - https://github.com/ProvableHQ/snarkVM/pull/2884

PR #2730, opened 2025-05-29
Revert "Clear cache on all `ConsenusVersion` changes"

niklaslong:
Reverts ProvableHQ/snarkVM#2728, see https://github.com/ProvableHQ/snarkVM/pull/2728#issuecomment-2916190976.


niklaslong:
After further investigation, the perf regression seems unrelated to this change—closing for now but we'll keep an eye on perf for the next release.

PR #2731, opened 2025-05-29
Create evaluate_commit and evaluate_hash.

mikebenfield:
The primary purpose is to create evaluate_commit and evaluate_hash functions which allow evaluating commits and hashes without the full Aleo machinery of stacks and registers. This is for the Leo interpreter to call to make sure its hash and commit results are identical to those in snarkVM.

CommitVariant and HashVariant (formerly Committer and Hasher) are now public, as evaluate_commit and evaluate_hash need to accept a parameter specifying which variant to use.

As long as I was making largish changes to operation/commit.rs and operation/hash.rs ...

A secondary purpose was to extract out a common implementation of evaluate and execute instead of the copy+pasted duplication that exists currently. This is done through a combination of traits (see HashCore and CommitCore) and macros (see define_run_hash, define_run_commit, and define_evaluate_or_execute) because the Network and Aleo traits are not written to allow easily abstracting their common functionality into a trait without a decent bit of shim code. The implementation is split into two parts (`run_hash` or `run_commit` plus evaluate_or_execute) because the new evaluate_commit and evaluate_hash functions need the functionality without registers and stacks on its own.

A tertiary purpose was to replace the VARIANT u8
generic parameter of CommitInstruction and HashInstruction with a type parameter constrained by a sealed trait. This better exploits the type system to express the intent: the type parameters via the trait CommitVariantTag or HashVariantTag now directly correspond to variants of CommitVariant or HashVariant, and only valid values of this parameter exist, unlike with the u8 VARIANT where the connection is arbitrary and must constantly be matched against and checked for invalid values.

I also added missing `From<V>` implementations for the console and circuit versions of Literal, which allows much briefer code (see all the `.into()` in `define_run_hash`).

### DIFF
--- a/synthesizer/program/src/logic/instruction/operation/commit.rs
+++ b/synthesizer/program/src/logic/instruction/operation/commit.rs
@@ -20,24 +20,25 @@ use crate::{
 };
 use console::{
     network::prelude::*,
-    program::{Literal, LiteralType, Plaintext, PlaintextType, Register, RegisterType, Value},
+    program::{Literal, LiteralType, Plaintext, PlaintextType, Register, RegisterType, Scalar, Value},
 };
 
 /// BHP256 is a collision-resistant function that processes inputs in 256-bit chunks.
-pub type CommitBHP256<N> = CommitInstruction<N, { Committer::CommitBHP256 as u8 }>;
+pub type CommitBHP256<N> = CommitInstruction<N, { CommitVariant::CommitBHP256 as u8 }>;
 /// BHP512 is a collision-resistant function that processes inputs in 512-bit chunks.
-pub type CommitBHP512<N> = CommitInstruction<N, { Committer::CommitBHP512 as u8 }>;
+pub type CommitBHP512<N> = CommitInstruction<N, { CommitVariant::CommitBHP512 as u8 }>;
 /// BHP768 is a collision-resistant function that processes inputs in 768-bit chunks.
-pub type CommitBHP768<N> = CommitInstruction<N, { Committer::CommitBHP768 as u8 }>;
+pub type CommitBHP768<N> = CommitInstruction<N, { CommitVariant::CommitBHP768 as u8 }>;
 /// BHP1024 is a collision-resistant function that processes inputs in 1024-bit chunks.
-pub type CommitBHP1024<N> = CommitInstruction<N, { Committer::CommitBHP1024 as u8 }>;
+pub type CommitBHP1024<N> = CommitInstruction<N, { CommitVariant::CommitBHP1024 as u8 }>;
 
 /// Pedersen64 is a collision-resistant function that processes inputs in 64-bit chunks.
-pub type CommitPED64<N> = CommitInstruction<N, { Committer::CommitPED64 as u8 }>;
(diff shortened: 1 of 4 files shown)


mikebenfield:
If I should be asking for different reviewers someone let me know.

The changes I really need are just the addition of evaluate_commit and evaluate_hash. IMO the other changes are clear improvements, but if it's too tedious to review this larger diff I can make a somewhat less disruptive PR instead.

d0cd:
@mikebenfield I think this PR would be easier to review if it was split into:
1. The refactoring of the body of the `evaluate` methods into a separate public function, and making `Hasher` public.
2. The ergonomic improvements

mikebenfield:
I think I'll actually just roll back most of the changes other than a macro to reduce copy+pasted code, plus extracting out the functionality I need. I'll push a different commit shortly.

mikebenfield:
Alright here's a much smaller change instead. New commit message too.

synthesizer/program/src/logic/instruction/operation/commit.rs
```diff
@@ -110,6 +111,50 @@ impl<N: Network, const VARIANT: u8> CommitInstruction<N, VARIANT> {
     }
 }
 
+// This code is nearly identical in `execute` and `evaluate`; we
+// extract it here in a macro.
+//
+// The `$q` parameter allows us to wrap a value in `Result::Ok`, since
+// the `Aleo` functions don't return a `Result` but the `Network` ones do.
+macro_rules! docommit {
```
d0cd: Nit.
```suggestion
macro_rules! do_commit {
```

synthesizer/program/src/logic/instruction/operation/hash.rs
```diff
@@ -196,6 +197,72 @@ impl<N: Network, const VARIANT: u8> HashInstruction<N, VARIANT> {
     }
 }
 
+// This code is nearly identical in `execute` and `evaluate`; we
+// extract it here in a macro.
+//
+// The `$q` parameter allows us to wrap a value in `Result::Ok`, since
+// the `Aleo` functions don't return a `Result` but the `Network` ones do.
+macro_rules! dohash {
```
d0cd: Nit.
```suggestion
macro_rules! do_hash {
```

d0cd (approved):
LGTM overall! 

Left some very minor nits.
I read the macro to myself and thought "what in the world is a dohash? :)

synthesizer/program/src/logic/instruction/operation/commit.rs
```diff
@@ -110,6 +111,50 @@ impl<N: Network, const VARIANT: u8> CommitInstruction<N, VARIANT> {
...
+///
+/// This allows running the commit without the machinery of stacks and registers.
+/// This is necessary for the Leo interpeter.
+pub fn evaluate_commit<N: Network>(
+    variant: CommitVariant,
+    input: &Value<N>,
+    randomizer: &Scalar<N>,
+    destination_type: LiteralType,
+) -> Result<Literal<N>> {
+    evaluate_commit_private(variant as u8, input, randomizer, destination_type)
```
vicsn: Nit: in snarkVM we more often use the suffix `_internal`, because the word "private" is used for cryptography and alerts and fatigues when validating security issues.

vicsn (approved):

PR #2733, opened 2025-05-30
[Fix] Return unconfirmed transaction from fee ID

kaimast:
When given the transaction ID of a fee txn, `BlockStorage::get_unconfirmed_transaction` would not actually return the original/unconfirmed transaction but the free transaction.

This PR changes `BlockStorage` to behave as intended.

### DIFF
--- a/ledger/store/src/block/mod.rs
+++ b/ledger/store/src/block/mod.rs
@@ -881,7 +881,7 @@ pub trait BlockStorage<N: Network>: 'static + Clone + Send + Sync {
         }
     }
 
-    /// Returns the transaction for the given `transaction ID`.
+    /// Returns the transaction for the given `TransactionID`.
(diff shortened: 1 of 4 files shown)


kaimast:
Converting back to "draft" because the change does not work as expected.

kaimast:
Ready for review. I tested it end-to-end and added a unit test to the PR.

vicsn (approved):
Amazing Kai. I will test E2E one more time with a rejected tx in a running network, and then we can go live

ledger/store/src/block/mod.rs
```diff
@@ -1514,4 +1541,59 @@ mod tests {
             );
         }
     }
+
+    /// Test that we can look up a rejected transaction using the fee transaction ID.
+    #[test]
+    fn test_rejected_transaction() {
```
niklaslong, line 1547: Thanks for adding a test! ❤️

niklaslong (approved):

PR #2734, opened 2025-05-30
[ARC-0046] Staking for Puzzle Solution Submissions

raychu86:
## Motivation

This PR is implements the new features highlighted in the [ARC-0046 Proposal](https://github.com/AleoNet/ARCs/discussions/97).

This is done by:

1. Tracking the provers and the number of solutions they have landed in the current epoch.
2. Aborting solutions from provers that exceeds their solution limit.
3. Invalidating blocks that have solutions that exceed a prover's solution limit.
4. Implementing a stepwise function to calculate the number of solutions a prover can land in an epoch based on the prover stake and current block timestamp.

Note that there is no need for a new `ConsensusVersion` if the first enforcement timestamp is after all validators upgrade.

## Test Plan

Tests have been added to ensure solution limits are properly calculated based on timestamp.

### DIFF
--- a/ledger/src/advance.rs
+++ b/ledger/src/advance.rs
@@ -234,7 +234,7 @@ impl<N: Network, C: ConsensusStorage<N>> Ledger<N, C> {
                         let prover_address = solution.address();
                         let num_accepted_solutions = accepted_solutions.get(&prover_address).copied().unwrap_or(0);
                         // Determine the the prover has reached their solution limit.
-                        if self.has_reached_solution_limit(&prover_address, num_accepted_solutions) {
+                        if self.is_solution_limit_reached(&prover_address, num_accepted_solutions) {
                             return false;
                         }
                         // Check if the solution is valid.
--- a/ledger/src/advance.rs
+++ b/ledger/src/advance.rs
@@ -117,7 +117,8 @@ impl<N: Network, C: ConsensusStorage<N>> Ledger<N, C> {
             *self.current_committee.write() = Some(current_committee);
         }
 
-        // If the block is the start of a new epoch, or the epoch hash has not been set, update the current epoch hash.
+        // If the block is the start of a new epoch, or the epoch hash has not been set,
+        // update the current epoch hash and clear the epoch prover cache.
         if block.height() % N::NUM_BLOCKS_PER_EPOCH == 0 || self.current_epoch_hash.read().is_none() {
             // Update and log the current epoch hash.
             match self.get_epoch_hash(block.height()).ok() {
@@ -127,6 +128,16 @@ impl<N: Network, C: ConsensusStorage<N>> Ledger<N, C> {
                 }
                 None => {
                     error!("Failed to update the current epoch hash at block {}", block.height());
+                }
+            }
+            // Clear the epoch provers cache.
+            self.epoch_provers_cache.write().clear();
+        } else {
+            // If the block is not part of a new epoch, add the new provers to the epoch prover cache.
+            if let Some(solutions) = block.solutions().as_deref() {
+                let mut epoch_provers_cache = self.epoch_provers_cache.write();
+                for (_, s) in solutions.iter() {
+                    let _ = *epoch_provers_cache.entry(s.address()).and_modify(|e| *e += 1).or_insert(1);
                 }
             }
         }
@@ -139,11 +150,11 @@ impl<N: Network, C: ConsensusStorage<N>> Ledger<N, C> {
 pub fn split_candidate_solutions<T, F>(
     mut candidate_solutions: Vec<T>,
     max_solutions: usize,
-    verification_fn: F,
+    mut verification_fn: F,
 ) -> (Vec<T>, Vec<T>)
 where
     T: Sized + Copy,
-    F: Fn(&mut T) -> bool,
+    F: FnMut(&mut T) -> bool,
 {
     // Separate the candidate solutions into valid and aborted solutions.
     let mut valid_candidate_solutions = Vec::with_capacity(max_solutions);
@@ -217,9 +228,23 @@ impl<N: Network, C: ConsensusStorage<N>> Ledger<N, C> {
                 // Retrieve the latest proof target.
                 let latest_proof_target = self.latest_proof_target();
                 // Separate the candidate solutions into valid and aborted solutions.
+                let mut accepted_solutions: IndexMap<Address<N>, u64> = IndexMap::new();
                 let (valid_candidate_solutions, aborted_candidate_solutions) =
                     split_candidate_solutions(candidate_solutions, N::MAX_SOLUTIONS, |solution| {
-                        self.puzzle().check_solution_mut(solution, latest_epoch_hash, latest_proof_target).is_ok()
+                        let prover_address = solution.address();
+                        let num_accepted_solutions = accepted_solutions.get(&prover_address).copied().unwrap_or(0);
+                        // Determine the the prover has reached their solution limit.
+                        if self.has_reached_solution_limit(&prover_address, num_accepted_solutions) {
+                            return false;
+                        }
+                        // Check if the solution is valid.
+                        let is_valid =
+                            self.puzzle().check_solution_mut(solution, latest_epoch_hash, latest_proof_target).is_ok();
+                        // Add the solution to the accepted solutions if it is valid.
+                        if is_valid {
+                            *accepted_solutions.entry(prover_address).or_insert(0) += 1;
+                        }
+                        is_valid
                     });
 
                 // Check if there are any valid solutions.
--- a/ledger/src/advance.rs
+++ b/ledger/src/advance.rs
@@ -233,18 +233,20 @@ impl<N: Network, C: ConsensusStorage<N>> Ledger<N, C> {
                     split_candidate_solutions(candidate_solutions, N::MAX_SOLUTIONS, |solution| {
                         let prover_address = solution.address();
                         let num_accepted_solutions = accepted_solutions.get(&prover_address).copied().unwrap_or(0);
-                        // Determine the the prover has reached their solution limit.
+                        // Check if the prover has reached their solution limit.
                         if self.is_solution_limit_reached(&prover_address, num_accepted_solutions) {
                             return false;
                         }
-                        // Check if the solution is valid.
-                        let is_valid =
-                            self.puzzle().check_solution_mut(solution, latest_epoch_hash, latest_proof_target).is_ok();
-                        // Add the solution to the accepted solutions if it is valid.
-                        if is_valid {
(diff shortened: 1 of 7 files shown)


vicsn:
> Note that there is no need for a new ConsensusVersion if the first enforcement timestamp is after all validators upgrade.

Is it possible to use the ConsensusVersion or height anyway? It would be nice, for all future analysis, if all changes to validation logic are implemented in a uniform way. In the very unlikely event that height diverges from expected timestamp, we could update future heights.

raychu86:
> > Note that there is no need for a new ConsensusVersion if the first enforcement timestamp is after all validators upgrade.
> 
> Is it possible to use the ConsensusVersion or height anyway? It would be nice, for all future analysis, if all changes to validation logic are implemented in a uniform way. In the very unlikely event that height diverges from expected timestamp, we could update future heights.

We can definitely still have a new `ConsensusVersion` as a check while still maintaining the timestamp usage. This would be more procedural and organizational rather than required technically.

ledger/src/advance.rs
```diff
@@ -217,9 +228,23 @@ impl<N: Network, C: ConsensusStorage<N>> Ledger<N, C> {
...
                 let latest_proof_target = self.latest_proof_target();
                 // Separate the candidate solutions into valid and aborted solutions.
+                let mut accepted_solutions: IndexMap<Address<N>, u64> = IndexMap::new();
                 let (valid_candidate_solutions, aborted_candidate_solutions) =
                     split_candidate_solutions(candidate_solutions, N::MAX_SOLUTIONS, |solution| {
-                        self.puzzle().check_solution_mut(solution, latest_epoch_hash, latest_proof_target).is_ok()
+                        let prover_address = solution.address();
+                        let num_accepted_solutions = accepted_solutions.get(&prover_address).copied().unwrap_or(0);
+                        // Determine the the prover has reached their solution limit.
+                        if self.has_reached_solution_limit(&prover_address, num_accepted_solutions) {
```
vicsn: I think this might need to be `self.has_reached_solution_limit(&prover_address, num_accepted_solutions + 1)`

Let's say a prover is allowed to submit 2 solutions. The cache contains 1, `accepted_solutions` contains 1. This check will pass. And as a result, we will be adding **another** solution to `accepted_solutions`

vicsn (approved):
On second thought, defining consensus versions 2 years in advance is not ideal, so I'm happy to stick to the timestamps approach. 

Please keep in mind before this is released whether we want an appropriate message version bump so clients which didn't update halt with a clear error message.

Found one potential bug, otherwise LGTM

ledger/src/advance.rs
raychu86: It should be correct as is. In your example of 2 allowed solutions in an epoch, if cache has 1 already, then there is only 1 allocation left.

If the incoming block has 2 more solutions. The first one will be accepted, which sets num_accepted_solutions to 1. The second solution in the block will now return true for `self.has_reached_solution_limit(&prover_address, num_accepted_solutions)` and be ignored.

ledger/src/get.rs
```diff
@@ -87,6 +87,26 @@ impl<N: Network, C: ConsensusStorage<N>> Ledger<N, C> {
         Ok(epoch_hash)
     }
 
+    /// Returns the provers and the number of solutions they have submitted for the current epoch.
+    pub fn get_epoch_provers(&self) -> Arc<RwLock<IndexMap<Address<N>, u32>>> {
```
ljedrz: nit: ideally, I'd return just `RwLock<IndexMap<Address<N>, u32>>` from here, or even just `IndexMap<Address<N>, u32>`, for broader utility

ledger/src/get.rs
```diff
@@ -87,6 +87,26 @@ impl<N: Network, C: ConsensusStorage<N>> Ledger<N, C> {
...
+        let current_block_height = self.vm().block_store().current_block_height();
+        let start_of_epoch = current_block_height.saturating_sub(current_block_height % N::NUM_BLOCKS_PER_EPOCH);
+        let existing_epoch_blocks: Vec<_> = (start_of_epoch..=current_block_height).collect();
+
+        // Create the epoch provers cache.
+        let epoch_provers = Arc::new(RwLock::new(IndexMap::new()));
+        cfg_iter!(existing_epoch_blocks).for_each(|height| {
+            if let Ok(Some(solutions)) = self.get_solutions(*height).as_deref() {
+                for (_, s) in solutions.iter() {
+                    epoch_provers.write().entry(s.address()).and_modify(|e| *e += 1).or_insert(1);
```
ljedrz: such intense locking for writes could easily defeat the purpose of parallelism; I'd suggest collecting the results into an `IndexMap` without involving an `RwLock`

raychu86: Note this is a one time initialization at startup, and `existing_epoch_blocks` can have at most 360 items. 

I'd like to deserialize in parallel (`get_solution`), because I assume that is the bulk of this operation; do you think the locking overhead would be that impactful?


Also the lock could be done before the `solutions.iter()` rather on each solution.

ledger/src/advance.rs
vicsn: I stand corrected.

If you have the time and energy, a test for this with a couple different cases would be nice, e.g.:
- empty prover cache v.s. non-empty prover cache
- solution in block within limit, solution above limit

ledger/src/get.rs
ljedrz: if the sets of solutions are similar in size, they will heavily compete for write access; doing it for all solutions (for a given height) at once would result in less aggressive re-locking, but will also very likely lead to heavy contention, making it basically a sequential operation; if the deserialization is quite slow and it's a single initialization, I'd recommend collecting all the solutions into a vector, utilizing `rayon` there, and then populating `epoch_provers` (which wouldn't require a lock at that point)

raychu86: @ljedrz I implemented a map reduce approach for the collection [here](https://github.com/ProvableHQ/snarkVM/pull/2734/commits/01053def6f3d9ff569ec7a54811f3ed8c37198c6#diff-f314fc61f4d461eb4f0347950951e8c24dd6fbd48625116a77566807294bda87R244-R273) (It may be a bit overkill, as there can be at most 359 blocks with 4 solutions each)

ljedrz: Indeed, it's a bit complicated; I was thinking of something simpler, which uses parallelism where it's most beneficial, and sequential processing where it's not:
```
let addresses = cfg_iter!(existing_epoch_blocks)
    .flat_map(|height| {
        if let Ok(Some(solutions)) = self.get_solutions(*height).as_deref() {
            solutions.iter().map(|(_, s)| s.address())
        }
    })
    .collect::<IndexSet<_>>();

let mut epoch_provers = IndexMap::new();
for addr in addresses {
    epoch_provers.entry(addr).and_modify(|e| *e += 1).or_insert(1);
}
```

raychu86: Cleaned up with - [2314f01](https://github.com/ProvableHQ/snarkVM/pull/2734/commits/2314f01e7d71b87ad0e01b215860ff403d2d99f4)

ledger/src/advance.rs
raychu86: Additional integration tests have been added

peterpan0708:
Can we separate the prover address from the staking address?
The prover address is a hot address, while the staking address should be a cold address.
This feature would be very helpful for mining pool operations.

howardwu:
This feature already exists in staking. When you bond, please bond your stake using your `prover_address`, and provide a cold `withdrawal_address` to the call to `fn bond_public`.

peterpan0708:
> This feature already exists in staking. When you bond, please bond your stake using your `prover_address`, and provide a cold `withdrawal_address` to the call to `fn bond_public`.

Thanky for the reply, Which wallet & app implement this feature? I tried both aleo.info and staking.xyz,  but i don't have an option to set the withdrawal_address.

I found this document, https://medium.com/@Upnode/aleo-native-staking-e0218c7c0c98
At least i can implement this by using cli tool, is this the only way?

howardwu:
> > This feature already exists in staking. When you bond, please bond your stake using your `prover_address`, and provide a cold `withdrawal_address` to the call to `fn bond_public`.
> 
> Thanky for the reply, Which wallet & app implement this feature? I tried both aleo.info and staking.xyz, but i don't have an option to set the withdrawal_address.
> 
> I found this document, https://medium.com/@Upnode/aleo-native-staking-e0218c7c0c98 At least i can implement this by using cli tool, is this the only way?

Correct, those are liquid staking products (I believe). You will need to perform native staking.

You can use snarkOS CLI `snarkos developer execute` command to call `bond_public` to natively stake, or you can use the Provable Tools website (https://www.provable.tools/protocol) to call `bond_public` to natively stake.

ledger/src/advance.rs
```diff
@@ -129,6 +130,16 @@ impl<N: Network, C: ConsensusStorage<N>> Ledger<N, C> {
...
                 }
             }
+            // Clear the epoch provers cache.
+            self.epoch_provers_cache.write().clear();
+        } else {
+            // If the block is not part of a new epoch, add the new provers to the epoch prover cache.
+            if let Some(solutions) = block.solutions().as_deref() {
+                let mut epoch_provers_cache = self.epoch_provers_cache.write();
+                for (_, s) in solutions.iter() {
+                    let _ = *epoch_provers_cache.entry(s.address()).and_modify(|e| *e += 1).or_insert(1);
```
vicsn, line 140: It seems like the `epoch_provers_cache` can only ever increase during an epoch, specifically in this line of code (ignoring startup logic).

But the `STAKE_REQUIREMENTS_PER_SOLUTION` can change at any block (by their median validator timestamp). Is it possible that we might need to re-evaluate the allowance down based on the increased requirement? Especially because we want to ensure all validators have the same view on the `epoch_provers_cache`.

ljedrz (approved):
Did another pass; to the extent I understand these changes, LGTM :+1:.

ledger/src/advance.rs
raychu86: I can't think of a case where an increase in `STAKE_REQUIREMENTS_PER_SOLUTION` (which would cause a decrease in `solution_limit`) would cause validator deviation because as long as their block view is the same. "

If any validator is behind, they won't be processing the new solutions until they catch up on blocks anyways.

vicsn: Indeed on second thought I don't see a risk for deviation between validators.

I also realise that we cannot revoke valid solutions from previous blocks, so decrementing the cache would be silly. We should just be sure that whenever we validate correctness of solutions, we take into account that the `epoch_prover_cache` may have been filled up with a previous `STAKE_REQUIREMENT_PER_SOLUTION`. And because we only call `is_solution_limit_reached` on **new** solutions, this is fine.

PR #2735, opened 2025-06-02
fix(ledger/store): return unconfirmed txn from fee id

vicsn:
## Motivation

Copy of https://github.com/ProvableHQ/snarkVM/pull/2733 for mainnet


niklaslong (approved):

vicsn:
I should not make branches with prefix "mainnet" anymore, they are indestructible.

PR #2736, opened 2025-06-02
Use generic queries in the VM

ljedrz:
Caveats:
- `QueryTrait` now requires that the implementors are `Clone`
- if a query is not supplied (i.e. it's `None`), it gets recreated in functions that need to use it more than once; this is perfectly fine with the current `Query` object, or if we assume the query to typically be provided

The 1st commit is the actual change, and the related diff is small; the second commit, and the vast majority of the PR's diff, is the formatting fallout from adjusting the tests to disambiguate the `None` values.

### DIFF
--- a/ledger/benches/transaction.rs
+++ b/ledger/benches/transaction.rs
@@ -24,7 +24,8 @@ use console::{
     program::{Plaintext, Record, Value},
 };
 use ledger_block::Transition;
-use ledger_store::ConsensusStore;
+use ledger_query::Query;
+use ledger_store::{ConsensusStorage, ConsensusStore};
 use synthesizer::{VM, program::Program};
 
 use aleo_std::StorageMode;
@@ -83,11 +84,30 @@ function hello:
     .unwrap();
 
     c.bench_function("Transaction::Deploy", |b| {
-        b.iter(|| vm.deploy(&private_key, &program, Some(records[0].clone()), 600000, None, rng).unwrap())
+        b.iter(|| {
+            vm.deploy(
+                &private_key,
+                &program,
+                Some(records[0].clone()),
+                600000,
+                None::<Query<MainnetV0, <LedgerType<_> as ConsensusStorage<_>>::BlockStorage>>,
+                rng,
+            )
+            .unwrap()
+        })
     });
 
     c.bench_function("Transaction::Deploy - verify", |b| {
-        let transaction = vm.deploy(&private_key, &program, Some(records[0].clone()), 600000, None, rng).unwrap();
+        let transaction = vm
+            .deploy(
+                &private_key,
+                &program,
+                Some(records[0].clone()),
+                600000,
+                None::<Query<MainnetV0, <LedgerType<_> as ConsensusStorage<_>>::BlockStorage>>,
+                rng,
+            )
+            .unwrap();
         b.iter(|| vm.check_transaction(&transaction, None, rng).unwrap())
     });
 }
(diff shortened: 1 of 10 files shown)


niklaslong (approved):

vicsn (approved):
LGTM. This will break a ton of dependent software (tx-cannons, delegated provers). I'll make sure to notify everyone during the canary release.

synthesizer/src/vm/execute.rs
```diff
@@ -224,7 +234,11 @@ impl<N: Network, C: ConsensusStorage<N>> VM<N, C> {
                 lap!(timer, "Execute the call");
 
                 // Prepare the assignments.
-                cast_mut_ref!(trace as Trace<N>).prepare(query)?;
+                if let Some(query) = query {
```
kaimast: It would be a little nicer to use similar code to what you removed above:

```suggestion
let query = query.unwrap_or_else(|| Query::VM(self.block_store().clone());
cast_mut_ref!(trace as Trace<N>).prepare(query)?;
```

I am unfamiliar with the metrics code, but is it fine to remove `lap!(timer, "Prepare the query");` from this function?

ledger/benches/transaction.rs
```diff
@@ -122,15 +142,20 @@ fn execute(c: &mut Criterion) {
                 vm.execute_authorization(
                     execute_authorization.replicate(),
                     Some(fee_authorization.replicate()),
-                    None,
+                    None::<Query<MainnetV0, <LedgerType<_> as ConsensusStorage<_>>::BlockStorage>>,
```
kaimast: Should we add a helper function that returns `None` or the empty query from below?

kaimast (approved):
Looks good! I added two suggestions.

synthesizer/src/vm/execute.rs
ljedrz: The `unwrap_or_else` can't be used due to the clash between generic and concrete types. Trait objects can't be used either, as the trait is not object safe. The alternative would be a dedicated `enum`, but it would result in a greater diff.

I don't use the metrics stuff either, but it doesn't seem like it's meaningful, as the creation of the query takes an insignificant amount of time.

ledger/benches/transaction.rs
ljedrz: I'm not sure if that's possible (due to concrete/generic type as output), but this gave me the idea that I could introduce type aliases to greatly reduce the diff.

synthesizer/src/vm/execute.rs
```diff
@@ -74,7 +77,15 @@ impl<N: Network, C: ConsensusStorage<N>> VM<N, C> {
                     )?,
                 };
                 // Execute the fee.
-                Some(self.execute_fee_authorization_raw(authorization, Some(query), rng)?)
+                if query.is_some() {
+                    Some(self.execute_fee_authorization_raw(authorization, query, rng)?)
+                } else {
+                    Some(self.execute_fee_authorization_raw(
+                        authorization,
+                        Some(Query::VM(self.block_store().clone())),
```
vicsn: It seems a bit silly to unwrap and rewrap the Option based on whether it is None. Can and should we do this one or more levels higher up the call chain?

vicsn (changes requested):

synthesizer/src/vm/execute.rs
ljedrz: makes a lot of sense; should I apply it only in non-`pub` functions or anything more?

vicsn: Good point, only non-pub please.

ljedrz: Done in https://github.com/ProvableHQ/snarkVM/pull/2736/commits/c7d200657a30a55b705ff131c842b6c1f0b2be0e; I noticed that there's a merge conflict, though; I'll fix it once the CI is done.

synthesizer/src/vm/execute.rs
kaimast: Oh, okay. That makes sense.

vicsn (approved):

PR #2742, opened 2025-06-03
[Feat] Database checkpoints

### LINKED ISSUE #2707: [Feature] Ability to revert ledger db
## 🚀 Feature

Add ability to revert ledger database in case of corruption.

## Motivation

Currently, if the ledger database is corrupted, the only way to recover is to either resync from genesis or download the ledger snapshot. Resyncing is not really practical right now, as syncing 7 million blocks would take months even from CDN syncing; downloading the snapshot is largely limited by network - currently the snapshot is sized at 400+ GB, and I can currently download the file at only around 500-600 Mbps, which will take 2 hours to download the whole file. Either way, the downtime would be too long to be considered comfortable to potentially recover a downed service.

Most of the corruptions would happen around at the tip of the database (incorrect current state etc), so most of the data in the database would be actually fine, and it's very wasteful to discard them altogether with corrupted data. If we are able to revert the database to a previous point, we could avoid the costly recover methods we currently are using. 

## Implementation

The major issue with the database data would be the mapping storage (FinalizeStorage), as it represents the current state of the mapping storage with no history. To enable reverting, we would need to add periodic snapshots to the mapping storage, so we can directly go back to that point in time. For other data, it would still be tedious to implement the reverting, as we would need to find all data related to the block being reverted and remove them one by one, and we don't have foreign keys to automatically cascade the deletion. 

**Are you willing to open a pull request?** (See [CONTRIBUTING](../../CONTRIBUTING.md))
hmm

ljedrz:
See the linked issue for the rationale; this PR implements the _direct checkpoints_ approach [outlined there](https://github.com/ProvableHQ/snarkVM/issues/2707#issuecomment-2875678759).

Closes https://github.com/ProvableHQ/snarkVM/issues/2707.

Filing as a draft, pending community feedback wrt the scope.


ljedrz:
CC https://github.com/ProvableHQ/snarkOS/issues/3630

vicsn (approved):
LGTM! Please open for review

ledger/store/src/helpers/memory/block.rs
```diff
@@ -211,4 +211,10 @@ impl<N: Network> BlockStorage<N> for BlockMemory<N> {
     fn transaction_store(&self) -> &TransactionStore<N, Self::TransactionStorage> {
         &self.transaction_store
     }
+
+    #[cfg(feature = "rocks")]
+    #[allow(unreachable_code)]
```
kaimast: Is this allow(unreachable_code) still needed?

ledger/src/lib.rs
```diff
@@ -235,6 +235,11 @@ impl<N: Network, C: ConsensusStorage<N>> Ledger<N, C> {
         Ok(ledger)
     }
 
+    #[cfg(feature = "rocks")]
+    pub fn backup_database<P: AsRef<std::path::Path>>(&self, path: P) -> Result<(), String> {
```
kaimast: Can you add a small doc comment here? Something like "Creates a database checkpoint at the given path". 

It might be good to mention what happens if the path already exists, or if there are any other restrictions to where they can be created.

Also, it would be more consistent with the rest of the code to return `anyhow::Result<()>` here.

kaimast (changes requested):
Requested one addition of a doc comment. 

Also, do we explain somewhere how to use the checkpoints? I assume whoever runs the validator just manually replaces the local storage with the checkpoint.

vicsn (approved):

vicsn:
CI failed.

@kaimast the snarkOS PR has instructions which I'll convert into a doc for validators

kaimast:
Sounds good. I will approve once CI passes.

niklaslong (approved):

vicsn:
Plz resolve conflict with `origin/staging` and make tests pass

vicsn (approved):

ljedrz:
Not sure why we're still seeing the RocksDB build issue, it passes for me locally without the manual addition of `CXXFLAGS="$CXXFLAGS -include cstdint"` :thinking:.

vicsn (approved):

ledger/store/src/helpers/memory/block.rs
```diff
@@ -211,4 +211,9 @@ impl<N: Network> BlockStorage<N> for BlockMemory<N> {
     fn transaction_store(&self) -> &TransactionStore<N, Self::TransactionStorage> {
         &self.transaction_store
     }
+
+    #[cfg(feature = "rocks")]
```
kaimast, line 215: Do we need the feature gate if this is also "implemented" for memory mode?

ledger/store/src/helpers/memory/block.rs
```diff
@@ -211,4 +211,9 @@ impl<N: Network> BlockStorage<N> for BlockMemory<N> {
...
+    fn backup_database<P: AsRef<std::path::Path>>(&self, _path: P) -> Result<(), String> {
+        Err("Unavailable in memory-only mode".to_owned())
```
kaimast, line 217: Maybe just return `anyhow::Error` here? Then you don't have to convert it later.

kaimast (approved):
I just had some minor comments, but should be merge-able as is.

PR #2745, opened 2025-06-03
[Feature] Add API to pre-process blocks

kaimast:
## Motivation

The goal of this PR is to help with some of the block sync improvements and also remove code from snarkOS that duplicates functionality from snarkVM.
The snarkOS PR is not ready yet, but this change is already ready to be merged and does not break any existing functionality or APIs.

More concretely, the PR adds an API to *only* check a block's subDAG, not its content. This is needed, because during sync we can apply a block to the ledger only once a subsequent block contains enough votes. This prevents us from fully verifying the block because those in-depth checks require an updated state root in the ledger.
We also don't need to verify the block content to know if it is valid, because at sync we can determine the validity of a block from the DAG, i.e., when there are sufficient votes.

As of now, there is around [100 lines of code](https://github.com/ProvableHQ/snarkOS/blob/12a29592aef1d3621d2e3845c9a9b15dbb02dc7b/node/bft/src/sync/mod.rs#L495) in `snarkos-node-bft` that I hope we can remove after this change is made.
This will prevent future bugs where the DAG checks in VM and OS might not match.

The API allows checking a block's DAG against a chain of pending blocks. This is needed, because there can be a series of blocks that do not contain sufficient votes. While this does not happen much in practice, it is possible to have a sequence of dozens of pending blocks.

This new API also lay the groundwork for light clients, which would only verify the DAG and track validator stake. 

## API Design
One design decision is to introduce a new type `PendingBlock`, which represents a block that has its DAG checked but none of its contents (transactions, state, etc.).

The idea is to use the type system to force proper API usage. Either, you call `Ledger::check_next_block` like before, or you first call `Ledger::check_block_subdag` and then `Ledger::check_block_content`.
In the latter case, `Ledger::check_block_subdag` will return a `PendingBlock` and `Ledger::check_block_content` expects a `PendingBlock` as its input.

I similarly would like `Ledger::check_next_block` to return something like a `VerifiedBlock` and have `Ledger::advance_to_next_block` only accept such a `VerifiedBlock`, but did not want to break any existing API.

## Testing
A large chunk of this PR is adding tests for the pending block logic. It adds these new test as integration tests, because they only touch the public API and `ledger/src/tests.ts` is already over 3600LOC. It might be good to move more of the tests to integration in the future.

For this new feature, there is only one test. It builds a DAG with a series of blocks that did not receive sufficient votes and check that preprocessing a chain of pending blocks works as expected.


ledger/src/check_next_block.rs
```diff
@@ -17,29 +17,127 @@ use super::*;
...
-        if height != latest_block.height() + 1 {
-            bail!("Block height is {height}, but expected {}", latest_block.height() + 1);
-        }
-
         // Ensure the block hash does not already exist.
         if self.contains_block_hash(&block.hash())? {
             bail!("Block hash '{}' already exists in the ledger", block.hash())
         }
 
+        // First check that the pending set and block height are correct.
```
acoglio:
```suggestion
        // First check that the heights of the pending block sequence and of the new block are correct.
```

ledger/src/check_next_block.rs
```diff
@@ -17,29 +17,127 @@ use super::*;
...
+                expected_height += 1;
+            }
+
+            if height != expected_height {
+                bail!("Block has invalid height. Expected {expected_height}, but got {height}.");
+            }
+        }
+
+        // Ensure the block height does not already exist.
+        // TODO(kaimast): consider removing this as we already check the height beforehand.
```
acoglio: Yes, I would remove this check because it is an invariant that the ledger has blocks in increasing sequential heights, and we check that the new block is greater than the latest block height in the ledger. Although it may not affect performance much, extra code affects readability (at least for me).

ledger/src/check_next_block.rs
```diff
@@ -17,29 +17,127 @@ use super::*;
...
+    fn check_block_subdag_inner(&self, block: &Block<N>, pending_blocks: &[PendingBlock<N>]) -> Result<()> {
         let height = block.height();
         let latest_block = self.latest_block();
 
-        // Check that this is actually the next block.
-        if height != latest_block.height() + 1 {
-            bail!("Block height is {height}, but expected {}", latest_block.height() + 1);
-        }
-
         // Ensure the block hash does not already exist.
```
acoglio, line 59: Since hashes are "practically injective",  the mere fact that the height of the new block differs from all the other heights in the ledger (which we check below) implies that its hash must differ too (with overwhelming probability) — if the block hash depends on the block height.

ledger/tests/helpers/mod.rs
```diff
@@ -0,0 +1,290 @@
...
+
+        Self {
+            private_keys,
+            ledger,
+
+            genesis_block,
+            current_height: 1,
+            last_batch: Default::default(),
+            last_committed_batch: Default::default(),
+            last_block_round: 1,
```
acoglio: Should `last_block_round` be 2, since blocks are created at even blocks?

Actually, I'm thinking that initially it should be 0, because no blocks (except for the genesis block) have been created yet.

ledger/tests/helpers/mod.rs
```diff
@@ -0,0 +1,290 @@
...
+        rng: &mut TestRng,
+    ) -> Vec<Block<CurrentNetwork>> {
+        assert!(num_blocks > 0, "Need to build at least one block");
+
+        (0..num_blocks).map(|_| self.generate_block_with_opts(options, rng)).collect()
+    }
+
+    /// Create a new block, with a fully-connected DAG.
+    ///
+    /// This will "fill in " any gaps left in earlier rounds from non participating nodes.
```
acoglio:
```suggestion
    /// This will "fill in" any gaps left in earlier rounds from non participating nodes.
```

ledger/tests/helpers/mod.rs
```diff
@@ -0,0 +1,290 @@
...
+    last_block_round: u64,
+    /// The batch certificates of the last round we generated.
+    round_to_certificates: HashMap<u64, IndexMap<usize, BatchCertificate<CurrentNetwork>>>,
+    /// The batch certificate of the last leader (if any).
+    previous_leader_certificate: Option<BatchCertificate<CurrentNetwork>>,
+    /// The height of the last block we created.
+    current_height: u64,
+    /// The last batch for each committee member that was included in a block.
+    /// Maps the author's index to a round number.
+    last_batch: HashMap<usize, u64>,
```
acoglio: Should this be `last_batch_round`? It's a `u64` , i.e. round number.

ledger/tests/helpers/mod.rs
```diff
@@ -0,0 +1,290 @@
...
+    /// The last batch of a validator that was included in a block
+    last_committed_batch: HashMap<usize, u64>,
```
acoglio: What's the exact difference between `last_batch` and `last_committed_batch`?

ledger/tests/helpers/mod.rs
```diff
@@ -0,0 +1,290 @@
...
+            StorageMode::new_test(None),
+        )
+        .unwrap();
+
+        Self {
+            private_keys,
+            ledger,
+
+            genesis_block,
+            current_height: 1,
```
acoglio: Should `current_height` be initially 0 (= the height of the genesis block)?

ledger/tests/helpers/mod.rs
```diff
@@ -0,0 +1,290 @@
...
+        }
+    }
+
+    /// Create multiple blocks, with fully-connected DAGs.
+    #[allow(dead_code)]
+    pub fn generate_blocks(&mut self, num_blocks: usize, rng: &mut TestRng) -> Vec<Block<CurrentNetwork>> {
+        self.generate_blocks_with_opts(num_blocks, &BlockOptions::default(), rng)
+    }
+
+    /// Create multiple blocks, with additional parameters
```
acoglio:
```suggestion
    /// Create multiple blocks, with additional parameters.
```

ledger/tests/helpers/mod.rs
```diff
@@ -0,0 +1,290 @@
...
+    /// Create a new block, with a fully-connected DAG.
+    ///
+    /// This will "fill in " any gaps left in earlier rounds from non participating nodes.
+    pub fn generate_block(&mut self, rng: &mut TestRng) -> Block<CurrentNetwork> {
+        self.generate_block_with_opts(&BlockOptions::default(), rng)
+    }
+
+    /// Same as `generate_block` but with additional options/parameters.
+    pub fn generate_block_with_opts(&mut self, options: &BlockOptions, rng: &mut TestRng) -> Block<CurrentNetwork> {
+        assert!(self.current_height > 0);
```
acoglio: The height of the genesis block is 0.

ledger/tests/helpers/mod.rs
```diff
@@ -0,0 +1,290 @@
...
+        assert!(options.skip_nodes.len() * 3 < self.private_keys.len());
+
+        let block_height = self.current_height + 1;
+
+        // SubDAGs can be at most GC roudns long.
+        let mut round = if self.last_block_round < BatchHeader::<CurrentNetwork>::MAX_GC_ROUNDS as u64 {
+            1
+        } else {
+            self.last_block_round - BatchHeader::<CurrentNetwork>::MAX_GC_ROUNDS as u64 + 2
```
acoglio: Why the `+ 2`? When `self.last_block_round` is 100 (= `MAX_GC_ROUNDS`), `round` would be initialized to 2, but I think it still needs to be initialized to 1 in this case?

ledger/tests/helpers/mod.rs
```diff
@@ -0,0 +1,290 @@
...
+        };
+
+        // Create certificates for each round.
+        loop {
+            let mut created_anchor = false;
+
+            let previous_certificate_ids = if round <= 1 {
```
acoglio:
```suggestion
            let previous_certificate_ids = if round == 1 {
```
Since `round` is never initialized to 0, a few lines above in the code.

acoglio (approved):

ledger/tests/helpers/mod.rs
kaimast: The doc is incorrect and I will fix it. last_committed_batch is for batches included in a block, and last_batch for any certificate created by the validator.

ledger/tests/helpers/mod.rs
kaimast: Good point. I will rename those.

ledger/tests/helpers/mod.rs
kaimast: Good catch. I was actually able to remove current_height entirely. It's not used in the code, and that's why the incorrect value did not cause issues.

ledger/tests/helpers/mod.rs
kaimast: I think the idea was that you pick the next round for a block which is at least the previous round + 2. But I changed it for the case between 100 and 102

ledger/src/check_next_block.rs
kaimast: I will leave this for now. It's just another sanity check that local state didn't get corrupted somehow.

vicsn (approved):

PR #2749, opened 2025-06-04
ci: use release profile for tests

kaimast:
Just a test...


ljedrz:
This is a great idea as long as we can ensure that any debug assertions (which could be feature-gated with `feature = "test"`) etc. are still being triggered :+1:.

vicsn:
Closing because it didn't noticeably improve performance, we can revisit when we have faster/external build times

PR #2754, opened 2025-06-06
Use `QueryTrait` as a trait object.

mikebenfield:
Two reasons:

1. Significantly improves the ergonomics of using these functions. Otherwise the caller has to import multiple types just to define something like
`type NoQuery =
   Query<MainnetV0,
       <LedgerType as ConsensusStorage<MainnetV0>>::BlockStorage>;`
and use `None::<NoQuery>`.
just to specify to use a default implementation. Now we can just pass `None`.

2. Reduce code bloat. These functions have 3 type parameters already.

### DIFF
--- a/ledger/benches/transaction.rs
+++ b/ledger/benches/transaction.rs
@@ -24,8 +24,7 @@ use console::{
     program::{Plaintext, Record, Value},
 };
 use ledger_block::Transition;
-use ledger_query::Query;
-use ledger_store::{ConsensusStorage, ConsensusStore};
+use ledger_store::ConsensusStore;
 use synthesizer::{VM, program::Program};
 
 use aleo_std::StorageMode;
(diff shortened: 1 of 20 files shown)


ljedrz:
Cool, I recall seeing the clones and thought the trait was inherently not dyn-safe, but apparently not; the performance penalty should be negligible.

mikebenfield:
Yeah there actually is a _potential_ drawback to trait objects here: we use the `async_trait` crate to make this trait with async functions object safe, and if that crate ever went away and Rust never got real support for trait objects with async functions, this would stop working. But snarkvm depends on `async_trait` anyway, and in the unlikely event this comes to pass, we could just switch back.

vicsn (approved):
Thank you for saving the day. This makes a lot of sense, the code in question is not on the hot path.

kaimast:
Yay, for decluttering the code! I am not approving myself because I don't know that part of the code well, but it looks good to me.

I would not worry too much about the async_trait crate, as it widely used, so it won't go away anytime soon.

niklaslong (approved):
Definitely an ergonomics improvement, thank you for filing the PR! I don't think we'll notice a perf difference.

PR #2758, opened 2025-06-09
[Fix] Address feedback from program upgradability audits.

d0cd:
This PR addresses feedback from auditors that reviewed program upgradability.
Auditors used the following commit: [962d5fd](https://github.com/ProvableHQ/snarkVM/pull/2722/commits/962d5fd0f2055ca3fa1f73234c2eee381307a105) as the reference.


ledger/block/src/transaction/merkle.rs
```diff
@@ -261,7 +261,7 @@ impl<N: Network> Transaction<N> {
             bail!("Program owner is required for V2 deployment tree");
         };
         // Prepare the header for the hash.
-        let header = to_bits_le![program_checksum, program_owner];
+        let header = to_bits_le![program_checksum, program_owner, deployment.edition()];
```
rot256: This works, but increases the cost of computing the root additionally: we need to hash the deployment edition (admittedly only 16 bits) for every function in the program.

I would recommend simply adding the *whole* deployment to the `deployment_id` and a domain separator in-case the meaning of fields change in the future (the program_owner will sign the `deployment_id`, so if the semantics of preimage of the hash changes in the future we want to "break" all the previous signatures) e.g.

```rust
let header = sha3([V2_DEPLOYMENT_DOMAIN_SEPARATOR, deployment.to_bytes_le()])
```

Where `V2_DEPLOYMENT_DOMAIN_SEPARATOR` is a fixed-length constant.

This is a pretty fool-proof way of ensuring that the `deployment_id` uniquely identifies the whole deployment object and its type (which may change in the future). In case we don't need the tree over the functions, having:

```rust
let deployment_id = sha3([V2_DEPLOYMENT_DOMAIN_SEPARATOR, deployment.to_bytes_le()])
```

Also suffices and is much cheaper to compute.

d0cd: See https://github.com/ProvableHQ/snarkVM/pull/2758/commits/d13d8479a89f3ecefe1c93cf91f57dd32fa46f79

synthesizer/src/vm/finalize.rs
```diff
@@ -614,24 +618,26 @@ impl<N: Network, C: ConsensusStorage<N>> VM<N, C> {
...
-                            _ => return Err("Expected deploy transaction".to_string()),
+                            _ => return abort_with_error("Expected deploy transaction".to_string()),
                         };
                         // The finalize operation here involves appending the 'stack', and adding the program to the finalize tree.
                         match process.finalize_deployment(state, store, deployment, fee) {
                             // Ensure the finalize operations match the expected.
                             Ok((stack, finalize_operations)) => match finalize == &finalize_operations {
                                 // Store the stack.
-                                true => stacks.push(stack),
+                                true => process.stage_stack(stack),
```
randomsleep, line 635: I think we also need to update the stack in the `atomic_speculate` function right? Otherwise `atomic_speculate` and `atomic_finalize` may behave differently.

synthesizer/src/vm/finalize.rs
```diff
@@ -590,10 +589,15 @@ impl<N: Network, C: ConsensusStorage<N>> VM<N, C> {
             // Acquire the write lock on the process.
             // Note: Due to the highly-sensitive nature of processing all `finalize` calls,
             // we choose to acquire the write lock for the entire duration of this atomic batch.
-            let mut process = self.process.write();
+            let process = self.process.write();
 
-            // Initialize a list for the deployed stacks.
-            let mut stacks = Vec::new();
+            // A helper function abort an atomic batch with an error message while rolling back changes to the stacks in `Process`.
+            let abort_with_error = |error: String| -> Result<Vec<FinalizeOperation<N>>, String> {
```
randomsleep: This requires us to call this function on every error, which is error-prone. I think an alternative way is to use "defer" function to automatically call the `revert_stacks` function when exiting the current scope. Like this:

```rust
    defer! {
        // This code runs when the scope ends
        process.revert_stacks();
    }
```

If the scope succeeds, we call `process.commit_stacks()` like the current way and the above `revert_stacks` will do nothing.

ledger/block/src/transaction/merkle.rs
rot256: Perfect :)

synthesizer/src/vm/finalize.rs
d0cd: Yes, I believe that's now included (just not in the commit in the tracker).

randomsleep: I see it now :+1:

synthesizer/src/vm/finalize.rs
d0cd: This is pretty neat! I've never used this macro. 
Is there a crate you recommend? `defer`, `scopeguard` or `defer_lite`?
I'll check with the rest of the team on whether they're open to this.

synthesizer/process/src/lib.rs
```diff
@@ -84,6 +84,8 @@ pub struct Process<N: Network> {
     universal_srs: UniversalSRS<N>,
     /// The mapping of program IDs to stacks.
     stacks: Arc<RwLock<IndexMap<ProgramID<N>, Arc<Stack<N>>>>>,
+    /// The mapping of program IDs to old stacks.
+    old_stacks: Arc<RwLock<IndexMap<ProgramID<N>, Option<Arc<Stack<N>>>>>>,
```
d0cd, line 88: Open question on whether it's cleaner to stage stacks in `Process` or manage them out of band in `atomic_{finalize, speculate}`

synthesizer/src/vm/finalize.rs
randomsleep: The macro is basically defining a Struct in the current scope and execute the specified function when it is dropped (called RAII). You can implement a simple local version like this:

```rust
// Define a struct that holds a closure to be executed on drop
pub struct Defer<F: FnOnce()> {
    f: Option<F>,
}

impl<F: FnOnce()> Defer<F> {
    pub fn new(f: F) -> Self {
        Defer { f: Some(f) }
    }
}

impl<F: FnOnce()> Drop for Defer<F> {
    fn drop(&mut self) {
        if let Some(f) = self.f.take() {
            f();
        }
    }
}

// Macro to create a defer block with a unique guard variable
#[macro_export]
macro_rules! defer {
    ($($body:tt)*) => {
        let _defer_guard = $crate::Defer::new(|| { $($body)* });
    };
}
```

And use it:

```rust
fn main() {
    defer! {
        println!("execute defer");
    }

    println!("in main");
}
// Output:
// in main
// execute defer
```

d0cd: First of all big fan of this, I'll be borrowing this in a lot of my other code :)
Do you see a way defer can be done conditionally? Basically, if there's an error then `revert_stacks` otherwise `commit_stacks`?

randomsleep: For this case, if you already executed `commit_stacks`, then the `revert_stacks` (or called `revert_uncommitted_stack`) will do nothing right?

synthesizer/src/vm/verify.rs
```diff
@@ -136,17 +136,13 @@ impl<N: Network, C: ConsensusStorage<N>> VM<N, C> {
         // Construct the transaction checksum.
         let checksum = Data::<Transaction<N>>::Buffer(transaction.to_bytes_le()?.into()).to_checksum::<N>()?;
 
-        // We first acquire a read lock on the process to ensure that the programs are not updated while we are verifying the transaction.
-        // This lock is held for the duration of the transaction verification.
-        // Note: The calls to `check_deployment_internal` and `check_execution_internal` will also acquire a read lock on the process.
-        let process = self.process.read();
```
d0cd, line 142: Note to reviewers. Removing this read lock held over the duration of verification improved snarkOS node performance.

Furthermore, this is safe since editions are always sequentially incremented. If an edition is updated, the cache key will simply be invalid and the full verification will run again.

d0cd:

PR #2760, opened 2025-06-11
[Perf] Speed up authorize by evaluating instead of executing

vicsn:
## Motivation

This PR speeds up `authorize` by 40x for standard transfer calls to facilitate browser-based power users. Existing proofs of concepts end up waiting 30 seconds for authorizing a batch of transactions. This is done by letting `authorize` call `evaluate_function` instead of `execute_function`, while ensuring these functions have the same interface. Evaluation doesn't generate a circuit under the hood.

This does come at the cost of reduced UX: the user doesn't get feedback on whether they're creating a satisfying circuit. Note that there's no safety issue here; an authorization for an unsatisfied circuit will just lead to failing proof generation and verification. Moreover, for executions of common programs on common input, e.g. `credits.aleo/transfer_public`, the circuit is unlikely to be unsatisfied.

I added an `authorize_unchecked` function for clients/users who want to run the fast version.

## Testing
Existing unit tests should suffice, they use both the old and new version.


niklaslong:
On the API: do we want an `authorize_unchecked` instead of making it the default in `authorize`? This would also avoid the breaking change to callers.

vicsn:
> On the API: do we want an authorize_unchecked instead of making it the default in authorize? This would also avoid the breaking change to callers.

It's a good question which of the two versions should be the default. Either way, this PR doesn't touch any public API, its all internal. My opinion is that its fine to have the fast version be the default. I added more commentary in the PR description above.

synthesizer/src/vm/mod.rs
```diff
@@ -3036,7 +3036,8 @@ function check:
         let function_name = Identifier::<CurrentNetwork>::from_str("check").unwrap();
 
         // Generate a random Field for input
-        let input = Value::<CurrentNetwork>::from_str(&Field::<CurrentNetwork>::rand(rng).to_string()).unwrap();
```
d0cd, line 3112: Why did this need to be changed?

synthesizer/tests/test_process_execute.rs
```diff
@@ -39,7 +39,9 @@ fn test_process_execute() {
         // Run the test.
         let output = run_test(process.clone(), test);
         // Check against the expected output.
-        test.check(&output).unwrap();
+        if let Err(err) = test.check(&output) {
```
d0cd: This changes the behavior of the integration test framework a bit.
Should we be saving the output if check errors?

synthesizer/src/vm/mod.rs
vicsn: In `execute_function`, asserts are ignored. In `evaluate_function`, they are not ignored. To preserve the original succeeding test behaviour, we need to pass in exactly the field which this function expects, it is hardcoded a few lines above.

synthesizer/tests/test_process_execute.rs
vicsn: d6f9c7c14

raychu86:
Can you quantify the need for this change in terms of runtime difference? What are the major unlocks? Browser authorization generation?

I am always nervous when touching Stack operations, however the logic looks fairly straightforward.

synthesizer/process/src/stack/call/mod.rs
```diff
@@ -30,23 +30,33 @@ use synthesizer_program::{
     StackTrait,
 };
 
-pub trait CallTrait<N: Network> {
+pub trait CallTrait<N: Network, R: CryptoRng + Rng> {
```
raychu86: Why add the `R: CryptoRng + Rng` to the `CallTrait` and not directly in the `evaluate` function?

It was also originally in `execute`.

synthesizer/process/src/stack/evaluate.rs
```diff
@@ -97,15 +97,20 @@ impl<N: Network> Stack<N> {
...
         caller: Option<ProgramID<N>>,
+        root_tvk: Option<Field<N>>,
+        rng: &mut R,
     ) -> Result<Response<N>> {
         let timer = timer!("Stack::evaluate_function");
 
         // Retrieve the next request, based on the call stack mode.
-        let (request, call_stack) = match &call_stack {
+        let (request, call_stack) = match &mut call_stack {
+            CallStack::Authorize(ref mut requests, ..) => {
```
raychu86: The default match case error should be updated to include `Evaluate`, `Execute`, or `Authorize` now.

synthesizer/process/src/stack/evaluate.rs
```diff
@@ -97,15 +97,20 @@ impl<N: Network> Stack<N> {
...
+                (requests.pop().ok_or_else(|| anyhow!("No more requests in the call stack."))?, call_stack)
```
raychu86: Why not use `call_stack.pop()?;` like in `execute_function` to get the request?

synthesizer/process/src/stack/call/mod.rs
vicsn: No good reason: 2552566db

synthesizer/process/src/stack/evaluate.rs
vicsn: Good catch: 7258279cd

synthesizer/process/src/stack/evaluate.rs
vicsn: Ah much cleaner! 7258279cd

raychu86 (approved):

PR #2764, opened 2025-06-16
[Build] Support compilation w/o default features

### LINKED ISSUE #2721: [Bug] snarkVM does not build without default features
## 🐛 Bug Report

snarkVM requires some features to be enabled in order to build successfully.

We should either fix compilation without default features enabled, or remove feature gates for functionality that is essential to snarkVM (ledger, circuit, and synthesizer?).

## Steps to Reproduce

Run `cargo check --no-default-features`.

This will generate error messages similar to the ones in [this file](https://github.com/user-attachments/files/20358143/output.txt).

## Expected Behavior

snarkVM should build without problems.

kaimast:
## Motivation
Fixes #2721. This change enables us to build leaner test binaries by disabling unwanted features, and (hopefully) speeds up snarkOS compilation slightly.

## Overview
This PR introduces two new features, `package` and `file`, which are both enabled by default. This change more clearly defines dependencies for each feature. It also allows compiling without any default features or only the `file` feature.

Finally, the PR removes `cli` from the default features, as this flag depends on `ureq`, `self_update`, and a few other large dependencies. `cli` is only needed for the `snarkvm` binary, and it would be great if we could disable it in snarkOS.

## Testing
I manually tested each feature compiles by running `cargo check --no-default-features --features=<feature_name>`.

## Notes
We should consider removing features flags. snarkVM already has a lot of them, and it looks like `file` and its dependencies (`console` and `synthesizer`) should always be enabled.
I was not sure if removing existing features would break anything, so I did not make such a change.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -2686,6 +2686,7 @@ dependencies = [
  "rusty-hook",
  "self_update",
  "serde_json",
+ "snarkvm",
  "snarkvm-algorithms",
  "snarkvm-circuit",
  "snarkvm-console",
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -120,7 +120,6 @@ cli = [
   "package",
   "dep:clap",
   "dep:colored",
-  "dep:dotenvy",
   "dep:self_update",
   "dep:thiserror",
   "dep:ureq",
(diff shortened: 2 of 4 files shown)


vicsn (approved):

vicsn (changes requested):
CI fails

vicsn (approved):
Can you make a snarkOS PR to update to latest snarkVM rev? Just to make sure that still works.

PR #2766, opened 2025-06-17
[Perf] Reduce the number of allocs in Mul for LinearAssignment

ljedrz:
I found this one while heap-profiling a small binary performing `execute_authorization`. This refactoring resulted in the following perf improvements:

- runtime -1.5%
- all allocs -17.6%
- temp allocs -51%

PR #2769, opened 2025-06-18
[Perf] Reduce the number of allocs in Poseidon::permute

ljedrz:
Salvaged from https://github.com/ProvableHQ/snarkVM-fork/pull/1. The other change (related to `MulAssign`) is no longer significant with the upcoming https://github.com/ProvableHQ/snarkVM/pull/2766.

Perf improvements in a simple `execute_authorization` operation:
- total allocs -1%
- temp allocs -1.5%
- runtime -1%

While these are modest, the change is trivial and its effects would be more pronounced with more Poseidon hashing/permutations involved.

### DIFF
--- a/circuit/algorithms/src/poseidon/hash_many.rs
+++ b/circuit/algorithms/src/poseidon/hash_many.rs
@@ -164,8 +164,7 @@ impl<E: Environment, const RATE: usize> Poseidon<E, RATE> {
 
     /// Apply the Maximally Distance Separating (MDS) matrix in-place.
     #[inline]
-    fn apply_mds(&self, state: &mut [Field<E>]) {
-        let mut new_state = Vec::with_capacity(state.len());
+    fn apply_mds(&self, state: &mut [Field<E>], new_state: &mut Vec<Field<E>>) {
         for i in 0..state.len() {
             let mut accumulator = Field::zero();
             for (j, element) in state.iter().enumerate() {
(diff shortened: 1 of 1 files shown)


circuit/algorithms/src/poseidon/hash_many.rs
```diff
@@ -164,16 +164,15 @@ impl<E: Environment, const RATE: usize> Poseidon<E, RATE> {
...
+    fn apply_mds(&self, state: &mut [Field<E>], new_state: &mut Vec<Field<E>>) {
         for i in 0..state.len() {
             let mut accumulator = Field::zero();
             for (j, element) in state.iter().enumerate() {
                 accumulator += element * &self.mds[i][j];
             }
             new_state.push(accumulator);
         }
-        state.clone_from_slice(&new_state);
+        state.swap_with_slice(new_state);
```
raychu86, line 176: Can we guarantee that `new_state` and state` have the same length?

The `clone_from_slice` should never panic, but `swap_with_slice` might. 

I realize that calls to `apply_mds` via `permute` would guarantee this, but if something else calls `apply_mds` it may not

ljedrz: `clone_from_slice` [also panics](https://doc.rust-lang.org/std/primitive.slice.html#panics-32) when the lengths are not equal, that was the reason why I changed it to `swap_with_slice` without extra checks

raychu86: Got it. 

Nit - To preserve expected behavior of these functions, could you add a `new_state.clear` to the top of `apply_mds`?

ljedrz: Done.

vicsn (approved):

PR #2772, opened 2025-06-19
[Perf] Use a SmallVec in FromBits for Field

ljedrz:
Isolated from https://github.com/ProvableHQ/snarkVM-fork/pull/3.

`Field::from_bits_le` is an operation that's omnipresent in heap profiles; by using a `SmallVec` of the right capacity, we can get rid of _all_ those allocations. The exact scale of the perf improvement depends on the workload.


console/types/field/src/from_bits.rs
```diff
@@ -47,7 +49,7 @@ impl<E: Environment> FromBits for Field<E> {
             Ok(Field { field: E::Field::from_bigint(field).ok_or_else(|| anyhow!("Invalid field from bits"))? })
         } else {
             // Construct the sanitized list of bits padded with `false`
-            let mut sanitized_bits = vec![false; size_in_bits];
+            let mut sanitized_bits: SmallVec<[bool; 384]> = smallvec![false; size_in_bits];
```
raychu86, line 52: Why use the hardcoded value here? Is it not possible to fetch/use the size_in_bits constant?

ljedrz: Sadly, it is impossible, as it's not a constant value.

kaimast: Can't we make `Field::size_in_bits` a const function, or does the trait need to be `dyn` compatible?

ljedrz: It's a generic object, so you can't have a `Field::<N>::size_in_bits` that's `const`.

vicsn: What about using `E::Field::SIZE_IN_BITS`? Seems like `size_in_bits` just calls into that const.

ljedrz: Sadly, that's still generic:
```
error: generic parameters may not be used in const operations
  --> console/types/field/src/from_bits.rs:52:53
   |
52 | ...SmallVec<[bool; E::Field::SIZE_IN_BITS]> = smallvec![false; size_in_bits];
   |                    ^^^^^^^^^^^^^^^^^^^^^^ cannot perform const operation using `E`
   |
   = note: type parameters may not be used in const expressions
```

vicsn: Is it possible and a good idea to define a new const and add a unit test somewhere asserting the relevant instantiated values are the same?

ljedrz: hmm, as long as we knew what to test in order to find the greatest possible number of bits, we could define a `MAX_BITSIZE` (currently equal to `384`) and check if it's of that size

ljedrz:
Changed the magic number to a `const` and documented it. The benefit is still there: up to 2.4k allocs fewer per each maximum-length array plaintext, which translates to a perf improvement (and a lower overall strain on the allocator) on array-rich workloads.

vicsn (approved):

PR #2773, opened 2025-06-19
[Perf] Remove a redundant outer lock from the Process

ljedrz:
This is the `Process`:
```
#[derive(Clone)]
    /// The universal SRS.
}
```
and this is where it resides:
```
#[derive(Clone)]
pub struct VM<N: Network, C: ConsensusStorage<N>> {
    /// The process.
    process: Arc<RwLock<Process<N>>>,
    (...)
```
With `UniversalSRS` being immutable after its initialization, this setup basically leaves us with the following "meat" of the `Process` within the `VM`:
```
Arc<RwLock<Arc<RwLock< MAP_OF_STACKS >>>>
```
This means that the outer lock is redundant; the only place we ever use a write guard is finalization, which I retained in this PR by applying it to the underlying map of stacks (the only object within `Process` that can be altered).

### DIFF
--- a/ledger/benches/transaction.rs
+++ b/ledger/benches/transaction.rs
@@ -263,7 +263,7 @@ function main:
         let inputs = [Value::from_str("2group").unwrap()].into_iter();
 
         // Add the program to the VM.
-        vm.process().write().add_program(&program).unwrap();
+        vm.process().add_program(&program).unwrap();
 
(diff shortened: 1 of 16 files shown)


ljedrz:
Actually, I think there is an API issue here; the only reason why we have a lock over the `Process` is to restrict it during finalization, but it is the only entry point into the stacks. I need to tinker with it a bit more.

PR #2776, opened 2025-06-23
[Do Not Merge] Fall back to int deserialization for block metadata

vicsn:
## Motivation

This old PR: https://github.com/ProvableHQ/snarkVM/pull/2559 introduced a backwards-incompatible change. We just changed serialization of data. This PR tries both versions in the deserialization.

Alternatively to this PR, explorers and indexers re-index their data to use the new format. Keen to hear your feedback on the preferred way forward. CC @miazn @HarukaMa 

## Test Plan

- [ ] Test that deserialization and `snarkos developer scan` now succeeds.

### DIFF
--- a/ledger/block/src/header/metadata/serialize.rs
+++ b/ledger/block/src/header/metadata/serialize.rs
@@ -24,8 +24,8 @@ impl<N: Network> Serialize for Metadata<N> {
                 metadata.serialize_field("network", &self.network)?;
                 metadata.serialize_field("round", &self.round)?;
                 metadata.serialize_field("height", &self.height)?;
-                metadata.serialize_field("cumulative_weight", &self.cumulative_weight.to_string())?;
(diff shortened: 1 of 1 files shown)


HarukaMa:
Looks like this is only affecting JSON block format. AleoScan doesn't consume JSON format; it currently only accepts data in binary format and directly from local rocksdb, so it's not affected by this change. For example, the explorer won't directly support upcoming upgradable program, as the binary format of programs will be changed to include a new `constructor` element.

Internally AleoScan does have a JSON serializer to produce data for the frontend to consume, but it's not designed to be consumed back by snarkVM as it didn't store the subdag data from blocks, thus can't reproduce complete blocks from the database.

vicsn:
Also has been fixed for explorer.provable.com - thx @miazn !!!

PR #2777, opened 2025-06-24
Introduce the dev_println and dev_eprintln macros.

mikebenfield:
If the `dev-print` feature is enabled, these work like `println` and `eprintln`, respectively. This is to give users (in particular Leo) more control of snarkvm's output - currently some print statements are compiled in when `debug-assertions` is on, but we want to elide those print statements even with `debug-assertions`.

I also removed some superfluous `#[must_use]` attributes - this attribute does nothing when used on a function in a trait implementation.


vicsn (commented):
In principle LGTM!

It has been quite useful to read some of these logs during test runs, usually from CI.

Is there a way to enable the logs when the code is built/ran with `cargo test`? Is that desirable/acceptable for you? Or should we just enable the features in snarkVM's and snarkOS's CI?

mikebenfield:
This new commit uses the attribute `#[cfg(any(feature = "dev-print", test))]` so the printing always happens during tests, regardless of the status of the `dev-print` feature. I also made `dev-print` features in each crate that depends on `snarkvm-utilities`, each of which implies `snarkvm-utilities/dev-print`, so users can easily enable the feature when compiling any higher level crate.

vicsn (approved):

PR #2778, opened 2025-06-25
[Feature] introduce `pub authorize_request`

vicsn:
## Motivation

Offline hardware wallets need to be able to sign a Request without importing the entire VM object.

## Test Plan

- An E2E test can be found here: https://github.com/vicsn/credits


synthesizer/process/src/stack/authorize.rs
```diff
@@ -118,7 +118,37 @@ impl<N: Network> Stack<N> {
...
+        finish!(timer, "Construct the authorization from the function");
+
+        // Return the authorization.
+        Ok(authorization)
+    }
+
+    /// Authorizes calls to public credits.aleo functions for the given request.
+    /// Compared to `authorize`, no private key is needed, but this only works for single public requests.
+    #[inline]
+    pub fn authorize_request<A: circuit::Aleo<Network = N>, R: Rng + CryptoRng>(
```
niklaslong, line 133: Just wondering if this shouldn't be `authorize_credits` but perhaps best keep the naming generic in case other calls can be authorised without the sk in future.

niklaslong (approved):

synthesizer/process/src/stack/authorize.rs
vicsn: Actually that was the original name, but kept it general and indeed the comment is outdated :') Will fix in a next PR.

PR #2780, opened 2025-06-26
Add test_max_certificates feature

vicsn:
## Motivation

Enables running a mainnet network with low targets and many certificates (validators).

This is needed for our current stress-testing setup.

Opening as a draft, because alternatively to this, we could also want to fix our tx-cannon to support canary/testnet.

## Test Plan

- [ ] TODO


vicsn:
Closing, it really is safer if we don't risk lowering the targets on mainnet.

PR #2781, opened 2025-06-26
[Test] Update solution tests to support future staking requirements

raychu86:
## Motivation

This PR updates the tests to support future changes to stake requirements that come from ARC-0046. This is done by creating a dedicated prover address that has sufficient stake in order to test the other validation properties.


vicsn (approved):
Great catch

PR #2782, opened 2025-06-29
Refactor: consolidate and simplify some traits.

mikebenfield:
This PR cleans up a lot of traits and their uses across synthesizer-program and synthesizer-process. 

In some cases several traits are already meant to be implemented together, and are here consolidated into a single trait, or one trait is made a supertrait of another when appropriate. Thus a reference like this:

```
registers: &mut (impl RegistersSigner<N> + RegistersLoad<N> + RegistersStore<N>)
```

becomes simplified to this:

```
registers: &mut impl RegistersSigner<N>
```

Some of these traits are entirely unnecessary and can just be incorporated into concrete types, and we can often replace references to traits with references to a concrete type, so that

```
registers: &mut (
    impl RegistersCall<N>
    + RegistersSigner<N>
    + RegistersSignerCircuit<N, A>
    + RegistersLoadCircuit<N, A>
    + RegistersStoreCircuit<N, A>
),
```

becomes

```
registers: &mut Registers<N, A>,
```

and 
```
stack: &(
    impl StackEvaluate<N>
    + StackExecute<N>
    + StackMatches<N>
    + StackKeys<N>
    + StackProgram<N>
),
```

becomes 
```
stack: &Stack<N>,
```

This PR thus results in a net decrease of over 600 lines of code. 

The PR is in 5 separate commits which can be examined individually; the commit message for each contains more detail about its changes. 

Some likely inappropriate uses of `#[inline]` have been removed, scattered among the various commits. 

This PR is only intended as a refactor to make these components of snarkvm a bit easier to work with, and likely results in no difference to compile times or runtime performance. I'd like to gradually make similar changes throughout the code base though, some of which are likely to result in big compile time improvements.


vicsn:
> some of which are likely to result in big compile time improvements

Looking forward to it.

vicsn:
Are these refactors necessary for the path towards speeding up compile-times? ~Large portions of this code were written by Howard and Ray. If we can spare them the bandwidth to review and remember/explain why they wrote it like this, it would be a win, despite the ugliness.~

vicsn:
Can you rebase this one and subsequently https://github.com/ProvableHQ/snarkVM/pull/2783 ? I promise to prioritize it

synthesizer/process/src/stack/helpers/matches.rs
```diff
@@ -99,6 +99,223 @@ impl<N: Network> StackMatches<N> for Stack<N> {
     fn matches_future(&self, future: &Future<N>, locator: &Locator<N>) -> Result<()> {
         self.matches_future_internal(future, locator, 0)
     }
+
+    /// Returns `true` if the proving key for the given function name exists.
+    fn contains_proving_key(&self, function_name: &Identifier<N>) -> bool {
```
vicsn, line 104: All of these additions don't seem a fit for this filename.

vicsn (commented):
One nit, and CI fails on clippy :)

vicsn (approved):

PR #2783, opened 2025-06-30
Workspace dependencies

mikebenfield:
Move a bunch of dependency declarations into workspace declarations, including all crates in the workspace and external crates which are dependents of more than one crate. 

In a cargo workspace, dependencies should be workspace dependencies when possible so versions and paths can be managed conveniently in one location and there's no chance of divergence. 

There are two commits here which can be examined individually. The first commit simply removes some redundant macros. 

These commits build off my other PR so I'm making that branch the target for now.


console/program/src/data/plaintext/from_bits.rs
```diff
@@ -65,7 +65,9 @@ impl<N: Network> Plaintext<N> {
             let literal = Literal::from_bits_le(literal_variant, next_bits(literal_size as usize)?)?;
 
             // Cache the plaintext bits, and return the literal.
-            Ok(Self::Literal(literal, OnceCell::with_value(bits_le.to_vec())))
+            let lock = OnceLock::new();
+            lock.set(bits_le.to_vec()).unwrap();
```
ljedrz: nit: these could be shortened using [From\<T\> for OnceLock\<T\>](https://doc.rust-lang.org/std/sync/struct.OnceLock.html#impl-From%3CT%3E-for-OnceLock%3CT%3E)

ledger/store/src/block/mod.rs
```diff
@@ -627,21 +616,16 @@ pub trait BlockStorage<N: Network>: 'static + Clone + Send + Sync {
...
     /// Returns the block height that contains the given `solution ID`.
     fn find_block_height_from_solution_id(&self, solution_id: &SolutionID<N>) -> Result<Option<u32>> {
-        match self.solution_ids_map().get_confirmed(solution_id)? {
-            Some(block_height) => Ok(Some(cow_to_copied!(block_height))),
-            None => match self.aborted_solution_heights_map().get_confirmed(solution_id)? {
-                Some(block_height) => Ok(Some(cow_to_copied!(block_height))),
-                None => Ok(None),
-            },
+        Ok(match self.solution_ids_map().get_confirmed(solution_id)? {
+            some @ Some(..) => some,
```
vicsn, line 620: Today I Learned.

vicsn (approved):
LGTM! 

How much AI was used in this process?

A more serious question: will this PR impact any project which imports any snarkVM crate? If yes, it would be good to prepare some documentation and add it to the PR, so I can include it in the release notes. We have a big breaking change release coming up, so even outlining bash commands for the search and replace may be welcome.

mikebenfield:
Pushed another couple commits; see their commit messages. Notably I removed a bunch of busted feature flags so CI will hopefully pass. 

I used a decent bit of AI but I had to walk it through a couple crates at a time and do a lot of manual editing and shell commands too. Not sure if that was a skill issue on my part with my prompts. 

Surprisingly there are no breaking API changes (as testified by `cargo semver-checks`), but many feature flags have been removed. These include all the enable_console features and many features in snarkvm-algorithms. In most cases the crates wouldn't compile without those feature flags enabled directly or otherwise anyway, and no functionality has been removed, and I doubt anyone was getting that fine-grained with it, but it's possible somebody was actually manually enabling a flag which is now gone.

ledger/store/src/block/mod.rs
kaimast: Today I learned too...

algorithms/Cargo.toml
```diff
@@ -46,148 +46,96 @@ harness = false
...
-  "msm",
-  "polycommit_wasm",
-  "r1cs",
-  "snark",
-]
 cuda = [ "snarkvm-algorithms-cuda" ]
 profiler = [ "aleo-std/profiler" ]
-crypto_hash = [ ]
-fft = [ ]
-locktick = [ "snarkvm-parameters?/locktick" ]
```
kaimast, line 179: Are you sure we don't need to enable the locktick feature anymore?

kaimast:
Thanks for the PR! I have been slowly adding workspace dependencies, but was hesitating to do a massive change like this. 
Right now, there are way too many features, and we hardly ever test all possible feature combinations. So any reduction in feature complexity is appreciated. 

One issue I noticed with this PR is that it seems to always compile the wasm bindings. For example, I can trigger the following compile error:

```
~> cargo clippy --workspace --all-targets --no-default-features
    Checking snarkvm-console-network-environment v3.8.0 (/Users/kaimast/dev/snarkVM/console/network/environment)
    Checking snarkvm-parameters v3.8.0 (/Users/kaimast/dev/snarkVM/parameters)
warning: snarkvm-algorithms-cuda@3.8.0: warning: /Library/Developer/CommandLineTools/usr/bin/ranlib: archive library: /Users/kaimast/dev/snarkVM/target/debug/build/snarkvm-algorithms-cuda-340e6abf3ab6dad6/out/libblst_msm.a the table of contents is empty (no object file members in the library define global symbols)
    Checking snarkvm-wasm v3.8.0 (/Users/kaimast/dev/snarkVM/wasm)
    Checking snarkvm-fields v3.8.0 (/Users/kaimast/dev/snarkVM/fields)
    Checking snarkvm-curves v3.8.0 (/Users/kaimast/dev/snarkVM/curves)
error[E0433]: failed to resolve: use of undeclared crate or module `snarkvm_console`
  --> wasm/src/tests.rs:16:5
   |
16 | use snarkvm_console::{
   |     ^^^^^^^^^^^^^^^ use of undeclared crate or module `snarkvm_console`

error[E0432]: unresolved import `snarkvm_utilities`
  --> wasm/src/tests.rs:20:5
   |
20 | use snarkvm_utilities::TestRng;
   |     ^^^^^^^^^^^^^^^^^ use of undeclared crate or module `snarkvm_utilities`

warning: unused import: `core::str::FromStr`
  --> wasm/src/tests.rs:22:5
   |
22 | use core::str::FromStr;
   |     ^^^^^^^^^^^^^^^^^^
   |
   = note: `#[warn(unused_imports)]` on by default

Some errors have detailed explanations: E0432, E0433.
For more information about an error, try `rustc --explain E0432`.
warning: `snarkvm-wasm` (lib test) generated 1 warning
error: could not compile `snarkvm-wasm` (lib test) due to 2 previous errors; 1 warning emitted
warning: build failed, waiting for other jobs to finish...
```

There are seemingly two issues here. First, `snarkvm-wasm` has a `console` feature, but does not compile without it. Second, `snarkvm-wasm` should not be enabled without the `wasm` feature set.

Also, I am not sure how to test the proposed change thoroughly. Last time I made feature changes, I just tried compiling without default features and one specific feature enabled, for each snarkVM feature...

mikebenfield:
When I run your same command `cargo clippy --workspace --all-targets --no-default-features` on `staging` I get the same error, so I don't think either of the problems you mention were introduced by this PR. 

Actually, discovering that kind of thing (the fact that many crates won't compile with/without default features or with certain combinations of feature flags) was what led me to just throwing out tons of feature flags without trying to carefully figure them out.

algorithms/Cargo.toml
mikebenfield: No. Is it important that this be a feature flag, or can I just remove the feature flag and always enable it?

ljedrz: it should be an optional feature - it's used to identify deadlocks and profile locks in general, so it has some performance penalty when enabled.

mikebenfield: Cool; I restored this feature flag and usage here.

kaimast:
> When I run your same command `cargo clippy --workspace --all-targets --no-default-features` on `staging` I get the same error, so I don't think either of the problems you mention were introduced by this PR.

It turns out I only fixed it for the top-level crate, so `cargo clippy --all-targets --no-default-features` (without `--workspace`) works.

kaimast (dismissed):
I will approve this now, as I don't see anything that's obviously wrong, and the change will reduce future pain a lot.

It might still be a good idea to not merge until after 4.0.0, in case it does break something.

kaimast:
The synthesizer tests do not seem to build now

mikebenfield:
@kaimast Good with you to merge now?

vicsn:
Removed Kai as reviewer as he has a lot on his plate and already approved a recent version, but I'd be curious if @iamalwaysuncomfortable can still build the sdk, as I see this PR impacting some WASM stuff.

kaimast (approved):
Yes, this looks good. At least, I don't see anything that is obviously wrong.

@vicsn let's merge this soon before it conflicts with another PR and to have enough time to detect any problems that this could cause before the next release.

kaimast:
One thing I noticed just now is that this PR removed some no-std code in `utilities`. For example, `utilities/src/errors.rs` now always re-exports `std::error::Error` even when the `std` feature is not set.

Is this intentional? I am not sure if we ever build without std, but we should either revert that change or remove the `std` featuer entirely.

mikebenfield:
Yes, that was intentional. `snarkvm-utilities` wouldn't build without the `std` feature, so I removed all the no-std stuff from it. I intended to also remove the `std` feature flag, but apparently failed to do so. I can make another PR doing that tomorrow.

vicsn:
@trevor-crypto please advise if the above is fine for you. I'm glad you mentioned that after this PR, this was easier to pull off: https://github.com/ProvableHQ/snarkVM/pull/2791

trevor-crypto:
> @trevor-crypto please advise if the above is fine for you. I'm glad you mentioned that after this PR, this was easier to pull off: #2791

LGTM

Issue #2775, opened 2025-06-21
[Bug] `cfg_iter!` macros use features incorrectly


kaimast:
When using a more recent version of Rust (but still Rust 2021), I get many of the following errors.

```
warning: unexpected `cfg` condition value: `serial`
   --> node/bft/src/helpers/storage.rs:398:9
    |
398 |         cfg_sorted_by!(rounds.clone(), |a, _, b, _| a.cmp(b))
    |         ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
    |
    = note: expected values for `feature` are: `cuda`, `default`, `locktick`, `metrics`, `telemetry`, and `test`
    = note: using a cfg inside a macro will use the cfgs from the destination crate and not the ones from the defining crate
    = help: try referring to `cfg_sorted_by` crate for guidance on how handle this unexpected cfg
    = help: the macro `cfg_sorted_by` may come from an old version of the `snarkvm_utilities` crate, try updating your dependency with `cargo update -p snarkvm_utilities`
    = note: see <https://doc.rust-lang.org/nightly/rustc/check-cfg/cargo-specifics.html> for more information about checking conditional configuration
    = note: `#[warn(unexpected_cfgs)]` on by default
    = note: this warning originates in the macro `cfg_sorted_by` (in Nightly builds, run with -Z macro-backtrace for more info)
```

This is because utilities has the `serial` feature, but many crates that use it do not.

The best fix is to change these macros to inline functions. I will do that eventually, but it is low priority.

kaimast:
This is fixed now in snarkVM and snarkOS. It might still be an issue for other users of snarkVM, but there is nothing more we can do on the snarkVM side, as we do not want to switch to inline functions.

PR #2785, opened 2025-07-01
[Chore] Replace all `cfg_` macros with inline functions

kaimast:
This PR changes all macros in `snarkvm_utilities::parallel` to inline functions. I am marking this as a draft until after 4.0.0 is release. The change is needed to upgrade to Rust 2024 cleanly, as mentioned in #2775, and also allows clippy to inspect the code more thoroughly.

The change set is massive but reviewing it should be straightforward. All major changes are in `utilities/src/parallel.rs`. For other files, the changes mostly come down to something like `cfg_iter_mut!(foo)` being changed to `cfg_iter(&mut foo)` with a few exceptions.
One exception is that I implemented `IntoIter` for `&Transactions`, and others were required to make `cargo clippy` happy.


kaimast:
This should be good to merge now. I am happy to rebase / merge a few more times if needed.

raychu86:
@iamalwaysuncomfortable Wanted to flag here that this _might_ affect SDK related compilation due to the usage of the inlining of the `cfg...` iterators. This will now be defined on the defining crate rather than the calling crate, which might affect how the `wasm` feature is passed through.

vicsn:
Thx for flagging Ray. @iamalwaysuncomfortable please confirm. Should we add a wasm build in CI? I vaguely thought we had it but I don't see the job anymore.

kaimast:
> Thx for flagging Ray. @iamalwaysuncomfortable please confirm. Should we add a wasm build in CI? I vaguely thought we had it but I don't see the job anymore.

I can add that if you want.

raychu86:
The logic looks mostly correct, however I didn't go through every line or replacement of the cfg macros.

My concern is that usability of these macros is not as explicit and more confusing for the developer. In addition, all repos that import `cfg_iter` will be broken; this includes snarkOS, explorer clients, CDNs, wallets, etc.

Any chance there is a cleaner solution that could preserve backwards compatibility? @ljedrz might have some rust magic as well

kaimast:
My intention was to make these easier to use, as inline functions provide a type signature, and, at least on my setup, work better with clippy. But I get that that is somewhat subjective.
I did not realize these macros are used much outside of snarkOS and snarkVM. Before, I made a change to `cfg_find` in [this PR](https://github.com/ProvableHQ/snarkVM/pull/2743) without it causing any headaches that I am aware of, but maybe some of the other macros are used more widely. 

Another useful thing about this change is that enabling the `serial` feature on the `snarkvm` dependency will also enable it for all other crates that use it (because features in rust are additive), whereas before you had to make sure that every dependency had `serial` enabled if you want to avoid using rayon.

If we want to keep them as macros, we would need to make sure the `serial` feature exists in every crate that uses one of the macros, so that the warning does not fire. There were actually quite a few that did not have the feature in snarkOS, as you can see in [this commit](https://github.com/ProvableHQ/snarkOS/pull/3699/commits/98751890672b90b309da40eafe3df178b0b6a278#diff-f5e82e097aa5005ad9c92fe29abb9de43cddfbf974a5dcafcdad98e4cbb7b715).
Alternatively, we could add `#[allow(unexpected_cfgs)]` inside the macros to disable the warning.

ljedrz:
Considering the potential for downstream breakage and the fact that some of the replacement functions are rather complex, my recommendation would be to keep the current setup, and to fix the application of the `serial` feature instead.

That being said, in general I too prefer plain functions, and the `clippy` point is valid. In addition, the features are normally additive, so the `serial` feature would ideally be `parallel` instead (just like the convention is to have an `std` feature, as opposed to a `no_std` one).

kaimast:
Let me try to remove the commits of this PR from #2788 (the Rust 2024 upgrade) and see if causes any problems. If not, I will close this PR, and we can maybe revisit it in the future.

kaimast:
Closing this in favor of #2788.

PR #2789, opened 2025-07-03
[Perf] Faster block retrieval from storage

kaimast:
#2933 should be merged first.

### Deserialization Changes
The PR introduces a `FromBytesUnchecked` trait, that allow structs to implement deserialization that skips some consistency changes. This is only intended to be used for trusted sources, such when loading from disk and, maybe in the future, for workers communicating with the primary or core clients with their validator.

To make use of this new trait there is a new function in `snarkvm_ledger_utiltiies::bytes` called `unchecked_deserialize` that replaces `bincode::deserialize` in the storage logic.

### Benchmark Changes
There are two new benchmarks for the ledger: "dag" and "store". The former measures deserialization for certificates and other DAG structures, and the latter measures read/write performance of `snarkvm_ledger_store`.

I also changed the benchmark script to behave similar to the one snarkOS, where benchmark runs from branches other than staging are not added to the data visible on [the benchmark graphs](https://provablehq.github.io/snarkvm/dev/bench).
The benchmarks usually only run for staging commits, but runs can be manually triggered for other branches. For those runs, the results are still recorded in the workflow summary, but adding the results to the graphs would create too much noise.

### Test Code Changes
The PR is quite big partially due to improvements to the test code. There is now a `ledger::test_helpers` submodule that exports the `TestChainBuilder` and contents of the `ledger-test-helpers` crate.

The purpose of this to help with construction of DAG-based chains for benchmarking. I am also hoping to build more integration tests in the future, now that the test helpers are public.

### Performance Gains
This PR adds a very basic benchmarks that inserts 1000 blocks into the storage and measures how long retrieval takes. With the change we get about a 50% reduction in retrieval times (17.4 -> 8.8ms; measured on my Macbook).

Note that performance for fetching the block height is unchanged, as deserialization cost is minimal in that case.

#### Without deserialization checks (new)
```
block_store/BlockDB::get_block/4validators
                        time:   [8.7308 ms 8.8144 ms 8.9101 ms]
block_store/BlockDB::get_block_height/4validators
                        time:   [4.7294 µs 4.8710 µs 5.0252 µs]
```

#### With deserialization checks (old)
```
block_store/BlockDB::get_block/4validators
                        time:   [11.531 ms 17.404 ms 24.750 ms]
block_store/BlockDB::get_block_height/4validators
                        time:   [4.4960 µs 4.7083 µs 4.9456 µs]

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -440,7 +440,7 @@ version = "1"
 version = "1.0.0"
 
 [workspace.dependencies.criterion]
-version = "0.6"
+version = "0.7"
 
 [workspace.dependencies.colored]
 version = "3"
--- a/ledger/Cargo.toml
+++ b/ledger/Cargo.toml
@@ -179,8 +179,7 @@ features = [ "test" ]
 workspace = true
 
 [dev-dependencies.aleo-std-storage]
-version = "1"
-default-features = false
+workspace = true
 
 [dev-dependencies.serde_json]
 workspace = true
--- a/ledger/Cargo.toml
+++ b/ledger/Cargo.toml
@@ -20,11 +20,18 @@ edition = "2024"
 name = "block"
 path = "benches/block.rs"
 harness = false
+required-features = [ "test-helpers", "rocks" ]
 
 [[bench]]
 name = "bonded_mapping"
 path = "benches/bonded_mapping.rs"
 harness = false
+
+[[bench]]
+name = "store"
+path = "benches/store.rs"
+harness = false
+required-features = [ "test-helpers", "rocks" ]
 
 [[bench]]
 name = "dag"
@@ -72,7 +79,8 @@ test = [
 test-helpers = [
   "snarkvm-ledger-test-helpers",
   "snarkvm-ledger-committee/test-helpers",
-  "snarkvm-ledger-narwhal/test-helpers"
+  "snarkvm-ledger-narwhal/test-helpers",
+  "dep:snarkvm-circuit"
 ]
 test_targets = [
   "snarkvm-console/test_targets",
@@ -90,6 +98,10 @@ workspace = true
 
 [dependencies.snarkvm-ledger-committee]
 workspace = true
+
+[dependencies.snarkvm-circuit]
+workspace = true
+optional = true
 
 [dependencies.snarkvm-ledger-narwhal]
 workspace = true
@@ -155,12 +167,20 @@ workspace = true
 workspace = true
 features = [ "filesystem" ]
 
+[dev-dependencies.snarkvm-ledger]
+workspace = true
+features = [ "test-helpers" ]
+
 [dev-dependencies.snarkvm-ledger-block]
 workspace = true
 features = [ "test" ]
 
 [dev-dependencies.snarkvm-ledger-test-helpers]
 workspace = true
+
+[dev-dependencies.aleo-std-storage]
+version = "1"
+default-features = false
 
 [dev-dependencies.serde_json]
 workspace = true
--- a/ledger/Cargo.toml
+++ b/ledger/Cargo.toml
@@ -27,6 +27,12 @@ path = "benches/bonded_mapping.rs"
 harness = false
 
 [[bench]]
+name = "dag"
+path = "benches/dag.rs"
+harness = false
+required-features = ["test-helpers"]
+
+[[bench]]
 name = "transaction"
 path = "benches/transaction.rs"
 harness = false
--- a/ledger/authority/src/bytes.rs
+++ b/ledger/authority/src/bytes.rs
@@ -27,17 +27,15 @@ impl<N: Network> FromBytes for Authority<N> {
             2.. => Err(error("Invalid authority variant")),
         }
     }
-}
 
-impl<N: Network> FromBytesUnchecked for Authority<N> {
-    /// Reads the authority from the buffer.
+    /// Reads the authority from the buffer without performing any checks on the data.
     fn read_le_unchecked<R: Read>(mut reader: R) -> IoResult<Self> {
         // Read the variant.
         let variant = u8::read_le(&mut reader)?;
         // Match the variant.
         match variant {
             0 => Ok(Self::Beacon(FromBytes::read_le(&mut reader)?)),
-            1 => Ok(Self::Quorum(FromBytesUnchecked::read_le_unchecked(&mut reader)?)),
+            1 => Ok(Self::Quorum(FromBytes::read_le_unchecked(&mut reader)?)),
             2.. => Err(error("Invalid authority variant")),
         }
     }
--- a/ledger/authority/src/bytes.rs
+++ b/ledger/authority/src/bytes.rs
@@ -34,7 +34,7 @@ impl<N: Network> FromBytes for Authority<N> {
         let variant = u8::read_le(&mut reader)?;
         // Match the variant.
         match variant {
-            0 => Ok(Self::Beacon(FromBytes::read_le(&mut reader)?)),
+            0 => Ok(Self::Beacon(FromBytes::read_le_unchecked(&mut reader)?)),
             1 => Ok(Self::Quorum(FromBytes::read_le_unchecked(&mut reader)?)),
             2.. => Err(error("Invalid authority variant")),
         }
--- a/ledger/authority/src/bytes.rs
+++ b/ledger/authority/src/bytes.rs
@@ -29,6 +29,20 @@ impl<N: Network> FromBytes for Authority<N> {
     }
 }
 
+impl<N: Network> FromBytesUnchecked for Authority<N> {
+    /// Reads the authority from the buffer.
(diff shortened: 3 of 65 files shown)

### CODE REVIEW

File: Cargo.toml [hunk 1]
```diff
@@ -24,6 +24,12 @@ license = "Apache-2.0"
 edition = "2024"
 rust-version = "1.88.0" # Attention - Change the MSRV in rust-toolchain and in .circleci/config.yml as well
 
+[package.metadata.cargo-semver-checks.lints]
```
vicsn: See comment in other PR.

ledger/src/check_next_block.rs
```diff
@@ -17,22 +17,106 @@ use super::*;
...
 use crate::narwhal::BatchHeader;
 
+use anyhow::bail;
+
+/// Wrapper for a block that has a valid subDAG, but where the block header,
+/// solutions, and transmissions have not been verified yet.
+///
+/// This type is created by `Ledger::check_block_subdag` and consumed by `Ledger::check_block_content`.
+#[derive(Clone, PartialEq, Eq)]
+pub struct PendingBlock<N: Network>(Block<N>);
```
vicsn: I assume we'll want to remove all the PendingBlock changes from this PR. Or do you aim to get those merged first?

ledger/src/lib.rs
```diff
@@ -33,7 +33,7 @@ pub use snarkvm_ledger_store as store;
 pub mod test_helpers;
 
 mod helpers;
-use helpers::*;
+pub use helpers::*;
```
vicsn, line 36: Was this caught be semver-checks? If so, I'm already very happy that we added it.

utilities/src/bytes.rs
```diff
@@ -166,6 +199,22 @@ impl<'de, T: FromBytes> FromBytesDeserializer<T> {
...
 
+pub struct FromBytesUncheckedDeserializer<T: FromBytesUnchecked + FromBytes>(PhantomData<T>);
+
+impl<'de, T: FromBytesUnchecked + FromBytes> FromBytesUncheckedDeserializer<T> {
+    /// Deserializes a dynamically-sized byte array.
+    pub fn deserialize_with_size_encoding<D: Deserializer<'de>>(deserializer: D, name: &str) -> Result<T, D::Error> {
+        let mut buffer = Vec::with_capacity(32);
+        deserializer.deserialize_bytes(FromBytesVisitor::new(&mut buffer, name))?;
+
+        if UNCHECKED_DESERIALIZE.get() {
```
vicsn, line 224: Given the constraints at hand, seems like a good approach.

Does this mean that we don't need the `FromBytesUncheckedDeserializer`? Or is that still useful to ensure we don't have to implement `FromBytesUnchecked` on every object within snarkVM? Though that shouldn't be too hard, as they can always just call into `FromBytes`.

vicsn (changes requested):
Great to see the 2x speedup!

ledger/src/lib.rs
kaimast: Yes!

ledger/src/check_next_block.rs
kaimast: Done. I also removed the changes to semver CI.

Cargo.toml
kaimast: I removed the semver commits from this PR.

utilities/src/bytes.rs
kaimast: Done in [8edc791](https://github.com/ProvableHQ/snarkVM/pull/2789/commits/8edc791ab48715a7443aef2d483d3d1bb5e5ba58).

utilities/src/bytes.rs
```diff
@@ -28,6 +29,21 @@ use serde::{
 };
 use smol_str::SmolStr;
 
+thread_local! {
+    static UNCHECKED_DESERIALIZE: Cell<bool> = const { Cell::new(false) };
```
vicsn, line 33: @ljedrz any final comments on this approach?

ledger/Cargo.toml
```diff
@@ -149,13 +167,21 @@ workspace = true
...
+
 [dev-dependencies.snarkvm-ledger-block]
 workspace = true
 features = [ "test" ]
 
 [dev-dependencies.snarkvm-ledger-test-helpers]
 workspace = true
 
+[dev-dependencies.aleo-std-storage]
+version = "1"
```
ljedrz: is this not a workspace dep?

utilities/src/bytes.rs
ljedrz: should be a fine workaround for now; in the future, once we move to `bincode 2`, we might be able to utilize something like [decode_from_slice_with_context](https://docs.rs/bincode/latest/bincode/fn.decode_from_slice_with_context.html) instead

ledger/authority/src/bytes.rs
```diff
@@ -27,6 +27,18 @@ impl<N: Network> FromBytes for Authority<N> {
...
         }
     }
+
+    /// Reads the authority from the buffer without performing any checks on the data.
+    fn read_le_unchecked<R: Read>(mut reader: R) -> IoResult<Self> {
+        // Read the variant.
+        let variant = u8::read_le(&mut reader)?;
+        // Match the variant.
+        match variant {
+            0 => Ok(Self::Beacon(FromBytes::read_le(&mut reader)?)),
```
ljedrz: since `read_le_unchecked` defaults to `read_le`, shouldn't we always prefer it in case some underlying deserialization becomes unchecked in the future?

ledger/block/src/serialize.rs
```diff
@@ -40,31 +40,30 @@ impl<N: Network> Serialize for Block<N> {
...
-                    DeserializeExt::take_from_value::<D>(&mut block, "aborted_transaction_ids")?,
-                )
-                .map_err(de::Error::custom)?;
-
-                // Ensure the block hash matches.
-                match block_hash == block.hash() {
-                    true => Ok(block),
-                    false => Err(de::Error::custom(error("Mismatching block hash, possible data corruption"))),
-                }
+        if deserializer.is_human_readable() {
```
ljedrz, line 43: nit: I also prefer this way of writing if-else blocks, but I also prefer if the diff isn't inflated by whitespace changes which could obscure the "real" changes :grin:

ledger/src/advance.rs
```diff
@@ -104,13 +104,14 @@ impl<N: Network, C: ConsensusStorage<N>> Ledger<N, C> {
                 "The given block is not the direct successor of the latest block"
             );
         } else {
+            ensure!(block.height() > 0, "Non-genesis blocks cannot have height 0");
```
ljedrz:
```suggestion
            ensure!(block.height() != 0, "Non-genesis blocks cannot have height 0");
```
nit: some day, `clippy` might complain about this

ljedrz (approved):
LGTM, left a handful of nits

ledger/authority/src/bytes.rs
vicsn: I guess the current approach has the benefit of being more explicit about what we've previously optimized.

On the other hand, as you say it would also be cleaner and less work if we just use real_le_unchecked for all of our custom types.

I'm not sure if either is much better and would therefore default to leaving the code as is.

ledger/block/src/serialize.rs
vicsn: It's not worth the revert at this point, but 100% agree for the future, please avoid the gratitious style refactors in future PRs

utilities/src/bytes.rs
kaimast: Good idea! It looks like that is only supported when using bincode's `Decode` trait and not serde's `Deserialize` trait.

ledger/block/src/serialize.rs
kaimast: An earlier version of the PR had another branch here for unchecked deserialization. I might have forgotten to revert to the match statement.

ledger/src/advance.rs
kaimast: Good catch!

ledger/authority/src/bytes.rs
kaimast: There are some cases where I already did this. For example, `Transactions` calls `Transaction::read_le_unchecked` but the latter just calls `Transactions::read_le`.
So, I will change this one too to be consistent.

kaimast: Made a relatively small set of changes in [this commit](https://github.com/ProvableHQ/snarkVM/pull/2789/commits/6771fbb019a272fc7600ecc85ab26763d660799d).

ledger/Cargo.toml
kaimast: It is. Fixed!

kaimast:
All reviewer comments should be addressed now.

vicsn (approved):
LGTM!

PR #2791, opened 2025-07-04
chore: changes for snarkvm-console to build on sgx

trevor-crypto:
## Motivation
In order to get `snarkvm-console` to compile under the target x86_64-fortanix-unknown-sgx
it is required to reworked some features in various dependency crates.

## Test Plan

Did not change any major functionality.
Only checked a full project build with `cargo build` under linux,
and a separate build of only `snarkvm-console` under SGX:

`cargo +nightly b --target x86_64-fortanix-unknown-sgx -p snarkvm-console --no-default-features --features "account, algorithms, collections, program, types, network_min"`

**Going to use the CI to work through errors**

## Related PRs

https://github.com/ProvableHQ/snarkVM/pull/2778


trevor-crypto:
Hi @vicsn, please have a look when you get a chance. Thank you!

trevor-crypto:
One question I have right now is will this actually be able to sign a `Request` without having to pull data from the filesystem or remotely?

kaimast:
The changes look good to me. Excited to get TEE support for (parts of) snarkVM!

I am not super familiar with this part of the code, so maybe @vicsn can answer your remaining questions.

console/network/Cargo.toml
```diff
@@ -9,7 +9,11 @@ license = "Apache-2.0"
 edition = "2021"
 
 [features]
-default = [ "snarkvm-algorithms/polycommit_full" ]
+default = []
```
vicsn: I'm not sure what the impact will be for dependents if we change `default`.

Is it possible for your usecase to import the `console` crate with `default-features = false`?

Could we introduce a default enabled `remote = snarkvm-parameters/remote` feature in `console`? Same for `filesystem`.

parameters/src/errors/parameter.rs
```diff
@@ -34,9 +34,12 @@ pub enum ParameterError {
 
     #[error("{}", _0)]
     Wasm(String),
+
+    #[error("Filesystem access is disabled, enable compiler flag for feature")]
+    FilesystemDisabled,
 }
 
-#[cfg(not(feature = "wasm"))]
+#[cfg(all(not(feature = "wasm"), feature = "remote"))]
```
vicsn: I think we could just do `#[cfg(feature = "remote")]` here, as well as in the cases below. Any dependent can choose to use `default-features = false` to disable.

wasm wizard @iamalwaysuncomfortable would you agree?

parameters/src/macros.rs
```diff
@@ -169,80 +169,86 @@ macro_rules! impl_load_bytes_logic_local {
...
-                        Err(_) => {
-                            eprintln!(
-                                "\n❗ Error - Failed to store \"{}\" locally. Please download this file manually and ensure it is stored in {:?}.\n",
-                                $filename, file_path
-                            );
+                    // Construct the URL.
+                    let url = format!("{}/{}", $remote_url, $filename);
+
+                    // Load remote file
+                    cfg_if::cfg_if! {
```
vicsn: Just checking, is a nested `cfg_if!` really required? Makes checking equivalence of the code really cumbersome.

vicsn (changes requested):
Looks great so far! Couple questions

console/network/Cargo.toml
trevor-crypto: > I'm not sure what the impact will be for dependents if we change default.

This is the crux of the problem, I think. It spiderwebs through other crates (`snarkvm-console-account` and `snarkvm-console-program`) since they don't disable default features. Should I try to disable default features on those instead?

> Is it possible for your usecase to import the console crate with default-features = false?

Yes, I had planned on doing this already, so it should be fine.

> Could we introduce a default enabled remote = snarkvm-parameters/remote feature in console? Same for filesystem.

Got it. I will remove the "snarkvm-parameters/full" and make respective features for "remote" and "filesystem"

parameters/src/macros.rs
trevor-crypto: Oversight on my part. Will remove the nesting!

trevor-crypto:
Need to rebase against large workspace change that is on `staging`.

trevor-crypto:
@vicsn After rebasing to the latest workspace dep changes, the changeset seems a bit nicer and was easier to get to compile for SGX...magic? :magic_wand:

parameters/src/macros.rs
trevor-crypto: Actually, doesn't seem possible: https://github.com/rust-lang/cfg-if/issues/6

parameters/src/errors/parameter.rs
trevor-crypto: Changed it slightly and used `target_env = "sgx"` instead of having a "remote" feature

console/Cargo.toml
```diff
@@ -50,6 +51,7 @@ test = [
 account = [ "network", "snarkvm-console-account" ]
 algorithms = [ "snarkvm-console-algorithms" ]
 collections = [ "algorithms", "snarkvm-console-collections" ]
+filesystem = [ "snarkvm-console-network/filesystem" ]
```
raychu86, line 54: Why not add an `sgx` feature that is not a default feature rather than injecting a `filesystem` feature that already represents the current behavior?

You could also enforce that the `target_env` must be `sgx` if the feature is enabled as well.

vicsn: I'm more in favour of `filesystem`, better not to have custom feature flags for every possible machine target (of which there are many, even differences between sgx machines)

parameters/Cargo.toml
```diff
@@ -93,8 +93,8 @@ version = "0.3.64"
 features = [ "XmlHttpRequest" ]
 optional = true
 
-[target."cfg(not(target_family = \"wasm\"))".dependencies.curl]
-version = "0.4.43"
+[target.'cfg(all(not(target_family = "wasm"), not(target_env = "sgx")))'.dependencies]
+curl = "0.4.43"
```
iamalwaysuncomfortable:
```suggestion
[target.'cfg(all(not(target_family = "wasm"), not(target_env = "sgx")))'.dependencies.curl]
version = "0.4.43"
```

We should keep the standard dependency format.

parameters/src/macros.rs
```diff
@@ -166,80 +166,86 @@ macro_rules! impl_load_bytes_logic_local {
...
-                if #[cfg(not(feature = "wasm"))] {
-                    let mut buffer = vec![];
-                    Self::remote_fetch(&mut buffer, &url)?;
-
-                    // Ensure the checksum matches.
-                    let candidate_checksum = checksum!(&buffer);
-                    if $expected_checksum != candidate_checksum {
-                        return checksum_error!($expected_checksum, candidate_checksum)
+        cfg_if::cfg_if! {
+            if #[cfg(feature = "filesystem")] {
```
iamalwaysuncomfortable, line 170:
```suggestion
            if #[cfg(feature = "filesystem")] {
```

This creates the need to keep track of always including the filesystem feature in any SnarkVM crates that depend directly on parameters. But perhaps more damaging to the ecosystem is that sometimes external rust crates DO in fact depend on the parameters crate [example](https://github.com/ProvableHQ/sdk/blob/mainnet/wasm/Cargo.toml#L45-L47) in ways that don't simply assume the default features, and this could take down services upon an upgrade.

I agree with @vicsn that being able to specify whether you have a filesystem is a good idea, but I think what would be better since it's much larger subset of people who will be using filesystems than not, is to create a `no-filesystem` feature flag and use instead as it would have fewer downstream side effects.

iamalwaysuncomfortable (dismissed):
This creates the need to keep track of always including the filesystem feature in any SnarkVM crates that depend directly on parameters. But perhaps more damaging to the ecosystem is that sometimes external rust crates DO in fact depend on the parameters crate [example](https://github.com/ProvableHQ/sdk/blob/mainnet/wasm/Cargo.toml#L45-L47) in ways that don't simply assume the default features, and this could take down services upon an upgrade.

I agree with @vicsn that being able to specify whether you have a filesystem is a good idea, but I think what would be better since it's much larger subset of people who will be using filesystems than not, is to create a no-filesystem feature flag and use instead as it would have fewer downstream side effects. If this PR can be refactored to use that, I'd be much happier with these changes.

ljedrz:
I'm in favor of a `filesystem` feature (as opposed to sth like `no-filesystem`), as it follows the Rust convention of features adding functionalities, e.g. the ubiquitous `std` feature. Since we're expecting it to be the typical user scenario, it should just be a default feature.

vicsn (approved):

PR #2792, opened 2025-07-04
[Feat] Implement new Inclusion version

raychu86:
## Motivation

This PR introduces a new Inclusion proof version that adds additional checks against the index of the record commitment being proven. In addition, a migration process will be introduced to enforce that `credits.aleo` records are properly upgraded via a new `upgrade` function and adhere to the new Inclusion proof rules.

This is done with the following changes:

- Introduce new `InclusionAssignment`
- Introduce new `pool` mapping
- Add `credits.aleo/upgrade` function

### Migration 

1. After release but prior to migration
    - Old inclusion proof is still used
    - `credits.aleo/upgrade` function not valid
2. New `ConsensusVersion::V8` migration height reached.
    - New inclusion proof is now used
    - `credits.aleo/upgrade` function is now enabled in consensus

## Test Plan

Unit tests have been added to affirm the properties are as desired. Devnets and stress tests have also been run to ensure correctness.


vicsn (approved):
LGTM!

PR #2793, opened 2025-07-04
[Feat] Introduces Record Versioning (V1), Sender Ciphertext, & Fixes Record Commitment

raychu86:
## Changelog

This PR includes the following major changes:
- Introduces `Record::version` with Version 0 (old) and Version 1 (new)
- Introduces `Record::is_hiding` which is `true` on Version 1 and `false` on Version 0
- Updates `Record::to_bits` serialization to support record versioning
- Fixes `Record::to_commitment` to use BHP commitments on Version 1
- Adds a `sender_ciphertext` which encrypts the `signer` address under the `recipient`'s address
    - See `Sender Ciphertext` section below
- Adds `Output::decrypt_sender_ciphertext` to decrypt the `sender_ciphertext`

### Sender Ciphertext

This PR introduces a `sender_ciphertext` which allows the recipient to decrypt the sender address using their account view key.

The encryption scheme is as follows (where `vk` is the recipient's account view key, and `G^r` is the record nonce):
```
record_view_key := (G^vk) * r == (G^r) * vk
randomizer := Hash_PSD4( ENCRYPTION_DOMAIN || record_view_key || 1field )
sender_ciphertext := ToXCoordinate(signer_address) + randomizer
``` 

To perform decryption of the sender ciphertext, call `Output::decrypt_sender_ciphertext` with your `account_view_key`. If the given account view key does not belong to the given output record's owner, the function will return an error.

## API Changes

The following APIs are impacted by this PR:
- `Request::InputID` now includes for record inputs a ` { "record_view_key": record_view_key } ` entry
- `Transition::Output` now includes for record outputs a ` { "sender_ciphertext": sender_ciphertext } ` entry

## Parameter changes

This PR also adds support for credits parameters and v0 credits parameters. The circuit changes will break the existing `credits.aleo` VKs for functions that create or consume records.

When a VM is loaded, it determines which credits verifying keys to use:
- V0 if the VM height is behind `ConsensusVersion::V8`
- V1 if the VM height is on or after `ConsensusVersion::V8`
- V0 if the environment is a test or the `test` features is enabled.

Once the VM passes the `ConsensusVersion::V8` height, it will automatically swap the `credits.aleo` VKs to the latest one.


synthesizer/process/src/verify_execution.rs
```diff
@@ -104,13 +105,23 @@ impl<N: Network> Process<N> {
...
-                .iter()
-                .enumerate()
-                .any(|(index, output)| !output.verify(function_id, transition.tcm(), num_inputs + index))
-            {
-                bail!("Failed to verify a transition output")
+            for (index, output) in transition.outputs().iter().enumerate() {
+                // If the consensus version are before `ConsensusVersion::V8`, ensure the output record is on Version 0.
+                // if the consensus version is on or after `ConsensusVersion::V8`, ensure the output record is on Version 1.
+                if let Some((_, record)) = output.record() {
+                    if (ConsensusVersion::V1..=ConsensusVersion::V7).contains(&consensus_version) {
```
vicsn, line 112: Just double checking if there's not an off-by-one here. On V8, we still expect the old inclusion version, but new records? https://github.com/ProvableHQ/snarkVM/pull/2792/files#diff-c84253a64c2536d7171b14e7867f0224cd9beef2193a1fa6c87868fce5d07dfbR160

synthesizer/src/vm/mod.rs
```diff
@@ -125,6 +123,23 @@ impl<N: Network, C: ConsensusStorage<N>> VM<N, C> {
...
 
+        #[cfg(not(any(test, feature = "test")))]
+        let mut process = {
+            // Determine the latest block height.
+            let latest_block_height = block_store.current_block_height();
+            // Determine the consensus version.
+            let consensus_version = N::CONSENSUS_VERSION(latest_block_height)?; // TODO (raychu86): Record Commitment - Select the proper consensus version.
+            // Initialize a new process based on the consensus version.
+            if (ConsensusVersion::V1..=ConsensusVersion::V7).contains(&consensus_version) {
+                Process::load_v0()?
```
vicsn, line 134: I'll just note that it seems a bit odd that we make process loading during startup dependent on the current consensus version. Are we also updating whenever we reach that consensus version as we're progressing at tip?

vicsn (approved):
Just two comments to improve my understanding, can be tackled in a future PR if needed

synthesizer/process/src/verify_execution.rs
vicsn: I see now the logic is the same everywhere already. **after** block height X, `ConsensusVersion::V8` and all of its rules come into effect, so the first block height from which it is enforced is effectively `X+1`.

PR #2794, opened 2025-07-04
[Migration] One time program upgrade to support circuit changes

raychu86:
This PR adds support for a 1-time redeployment for existing programs, without requiring program upgradability, which will come in a subsequent release. This is necessary because the circuit and VK changes created in https://github.com/ProvableHQ/snarkVM/pull/2793.

Additional rules include:
- Rejection of new deployments until `ConsensusVersion::V8`. 
- Rejection of executions of programs with edition 0 after `ConsensusVersion::V8`. 
    - Existing programs will need to redeploy with the new circuits.
    - New programs will need to deploy twice to have an executable version.
    - Handling of both of these operations will be addrsesed by the ANF and protocol teams to limit disruption for users and developers.

## Test Plan

Unit, integration, and devnet tests have been added/run to check correctness of the re-deployment process.


synthesizer/program/src/lib.rs
```diff
@@ -92,10 +92,15 @@ use console::{
     },
     program::{Identifier, PlaintextType, ProgramID, RecordType, StructType},
 };
+use snarkvm_utilities::cfg_iter;
 
+use console::prelude::Itertools;
```
d0cd, line 97: Nit. Move up.

synthesizer/process/src/lib.rs
```diff
@@ -120,28 +123,73 @@ impl<N: Network> Process<N> {
     }
 
     /// Adds a new program to the process.
+    /// If the program exists, then the existing program is replaced and discarded.
```
d0cd, line 126:
```suggestion
    /// If the program exists, then the existing program stack is replaced and returned.
```

synthesizer/process/src/lib.rs
```diff
@@ -120,28 +123,73 @@ impl<N: Network> Process<N> {
...
         if program.id() != &credits_program_id {
-            self.add_stack(Stack::new(self, program)?);
+            return Ok(self.add_stack(Stack::new(self, program)?));
         }
-        Ok(())
+        Ok(None)
     }
 
     /// Adds a new stack to the process.
+    /// If the program exists, then the existing stack is replaced and discarded
```
d0cd, line 140:
```suggestion
    /// If the program exists, then the existing program stack is replaced and returned.
```

synthesizer/process/src/stack/deploy.rs
```diff
@@ -67,6 +68,11 @@ impl<N: Network> Stack<N> {
 
         // Ensure the deployment is ordered.
         deployment.check_is_ordered()?;
+        // Ensure that edition in the stack and deployment matches.
```
d0cd, line 71: While this check is not incorrect, it's not practical since a user may want to verify deployments without needing to use the exact edition.

The edition consistency check also happens at the VM level.

synthesizer/src/vm/verify.rs
```diff
@@ -134,6 +134,37 @@ impl<N: Network, C: ConsensusStorage<N>> VM<N, C> {
 
         lap!(timer, "Check for duplicate elements");
 
+        // Get the consensus version.
+        let consensus_version = N::CONSENSUS_VERSION(self.block_store().current_block_height())?;
+
+        // Acquire a read lock on the process.
+        let process = self.process.read();
```
d0cd, line 141: We can consider minimizing this lock scope.
This would be safe since editions are always incremented and a write lock is held when the process stacks are updated during finalization.

d0cd:
Profiled `VM::load` which now respects transaction ordering.
```
# Mainnet
Loaded the VM from storage in: 139.417958ms
Loaded the VM from storage in: 305.644959ms
Loaded the VM from storage in: 190.118833ms
Loaded the VM from storage in: 126.090833ms
Loaded the VM from storage in: 155.77575ms

# Testnet
Loaded the VM from storage in: 10.121918292s
Loaded the VM from storage in: 10.164576541s
Loaded the VM from storage in: 10.015556958s
Loaded the VM from storage in: 10.243633667s
Loaded the VM from storage in: 10.000356417s

# Canary
Loaded the VM from storage in: 523.799416ms
Loaded the VM from storage in: 636.695666ms
Loaded the VM from storage in: 517.52ms
Loaded the VM from storage in: 536.825125ms
Loaded the VM from storage in: 527.400834ms
```

synthesizer/src/vm/verify.rs
vicsn: I think we can leave it. Or if you insist, move it just inside of `if let Transaction::Execute(id, _, execution, _) = transaction {`. There's a minor efficiency gain from taking it just once for all transitions.

vicsn (approved):
@raychu86 want to tackle Pranav's comments in a separate PR?

PR #2796, opened 2025-07-04
Clean up Record V0 error

vicsn:
## Motivation

This PR introduces a cleaner error when a user tries to create an authorization which is not allowed using V0 records.

## Test Plan

This should not influence any deserialization or syncing, as signing is a client- or prover-side operation. Some unit tests might not be happy because we fail earlier than before: during authorization instead of execution, and it is not guarded by consensus versions.


raychu86:
Logic looks good, but this will break some test assumptions, so they will need to be updated

raychu86 (approved):

PR #2797, opened 2025-07-05
[Feat] Increase deployment constraint limit 

raychu86:
## Motivation

This PR doubles the Deployment variable limit and constraint limit from 1m to 2m in preparation for the `4.0.0` release.

This will allow developers to deploy programs with richer and more complex circuit logic.

Note: There is no need for migration logic for this increase for 2 reasons:
- 4.0.0 is intended to be a synchronous validator restart 
- This is a limit _increase_, which retains backwards compatibility where all previous deployments are still valid.


vicsn (approved):

vicsn:
CI failed

d0cd (approved):

Issue #2599, opened 2025-01-20
[Feature] Program parsing does not fail on invalid program


vicsn:
## 🐛 Bug Report

The following aleo instructions program does not fail to be parsed, even though it references unknown program `staker_v1_multisig_b.aleo`

```
        Program::<CurrentNetwork>::from_str(&format!(
            r"
    import credits.aleo;
    import staking_v1.aleo;
    import staker_v1_b.aleo;
    import multisig_v1.aleo;

    program staker_v2_multisig_b.aleo;

    struct Settings:
        unstake_wait as u32;
        stake_paused as boolean;
        global_paused as boolean;
        max_reward_per_notify as u64;
        protocol_fee as u16;
        fee_account as address;
        fixed_stakers as boolean;

    struct Operation:
        program_id as address;
        op_type as u8;
        params as field;
        op_salt as u64;
        delay as u32;

    struct Request:
        operation as Operation;
        multisig as address;

    struct SetOperatorParams:
        operator as address;
        flag as boolean;

    function dummy:

    function set_operator:
        input r0 as address.public;
        input r1 as boolean.public;
        input r2 as u64.public;
        input r3 as [address; 3u32].public;
        input r4 as [signature; 3u32].private;
        cast r0 r1 into r5 as SetOperatorParams;
        hash.bhp256 r5 into r6 as field;
        cast staker_v1_multisig_b.aleo 3u8 r6 r2 0u32 into r7 as Operation;
        call multisig_v1.aleo/new_request r7 true r3 r4 into r8;
        call staker_v1_b.aleo/set_operator r0 r1 into r9;
        async set_operator r8 r9 into r10;
        output r10 as staker_v1_multisig_b.aleo/set_operator.future;

    finalize set_operator:
        input r0 as multisig_v1.aleo/new_request.future;
        input r1 as staker_v1_b.aleo/set_operator.future;
        await r0;
        await r1;

    function add_admin:
        input r0 as address.public;
        input r1 as u64.public;
        hash.bhp256 r0 into r2 as field;
        cast staker_v1_multisig_b.aleo 1u8 r2 r1 1200u32 into r3 as Operation;
        call multisig_v1.aleo/execute r3 into r4;
        call staker_v1_b.aleo/set_admin r0 true into r5;
        async add_admin r4 r5 into r6;
        output r6 as staker_v1_multisig_b.aleo/add_admin.future;

    finalize add_admin:
        input r0 as multisig_v1.aleo/execute.future;
[...]

PR #2798, opened 2025-07-07
Allow external structs.

mikebenfield:
Specifically,

1. Introduce a PlaintextType::ExternalStruct, and allow referring to such a type in .aleo code by a Locator.

2. Allow casting to external struct types.

3. Let two struct types be considered the same type as long as they are structurally the same and have the same local names. This funky ruleset is effectively here for backwards compatibility. In more detail:

```
/// Equivalence of structs means they have the same local names (regardless of whether
/// they're local or external), and their members have the same names and equivalent
/// types in the same order, recursively.
///
/// Equivalence of arrays means they have the same length and their element types are
/// equivalent.
///
/// This definition of equivalence was chosen to balance these concerns:
///
/// 1. All programs from before the existence of external structs will continue to work -
/// thus it's necessary for a struct created from another program to be considered equivalent
/// to a local one with the same name and structure, as in practice that was the behavior.
/// 2. We don't want to allow a fork. Thus we do need to check names, not just structural
/// equivalence - otherwise we could get a program deployable to a node which is using
/// this check, but not deployable to a node running an earlier SnarkVM.
```

Also note this other usecase from a user:

```
currently we got in an issue where one mainnet deployed contract imports tokenRegistry, and the contract we are writing for A imports that contract and A contract. but, A contract and tokenRegistryContract have same struct names so we cannot compile that.If we cannot change struct name in A contract, we need to wait for snarkVM update, just to compile the contracts
```


mikebenfield:
Rebased.

vicsn (commented):
Can you add a check in `check_transaction` that **deployments** using an external struct are only enabled from `ConsensusVersion::V10` onwards. You can put in dummy heights for the new consensus version.

And while you're at it, consider refactoring ConsensusVersion-related checks into a separate function called by `check_transaction`?

mikebenfield:
I moved some of the consensus version checks into a new function, but some others are scattered throughout `check_transaction` so this may or may not be what you had in mind.

synthesizer/tests/tests/parser/program/external_struct.aleo
```diff
@@ -0,0 +1,5 @@
+program child.aleo;
+
+function child:
+    call parent.aleo/parent into r0;
+    output r0 as parent.aleo/S.public;
```
vicsn, line 5: Should we test the external struct in any other location (as element of a map, struct, function input)? Or are we convinced there is no functional difference because we already try those in the higher level new execute tests?

synthesizer/src/vm/verify.rs
```diff
@@ -456,6 +414,64 @@ impl<N: Network, C: ConsensusStorage<N>> VM<N, C> {
...
+    if consensus_version >= ConsensusVersion::V9 {
+        ensure!(
+            deployment.program_checksum().is_some(),
+            "Invalid deployment transaction '{id}' - missing program checksum"
+        );
+        ensure!(deployment.program_owner().is_some(), "Invalid deployment transaction '{id}' - missing program owner");
+    }
+    if consensus_version <= ConsensusVersion::V9 {
+        ensure!(
+            !deployment.program().contains_external_struct(),
```
vicsn: `check_transaction` may be called multiple times, could we compute this when creating `ProgramCore` and set a member from which `ConsensusVersion` onwards it is supposed to be valid? Same for `contains_v9_syntax`?

synthesizer/program/src/function/mod.rs
```diff
@@ -95,6 +95,13 @@ impl<N: Network> FunctionCore<N> {
     pub const fn finalize_logic(&self) -> Option<&FinalizeCore<N>> {
         self.finalize_logic.as_ref()
     }
+
+    /// Does this function refer to an external struct explicitly?
+    pub fn contains_external_struct(&self) -> bool {
```
vicsn, line 100: Should this also check `finalize_logic`?

synthesizer/program/src/logic/instruction/mod.rs
```diff
@@ -430,6 +430,16 @@ impl<N: Network> Instruction<N> {
...
         instruction!(self, |instruction| instruction.output_types(stack, input_types))
     }
+
+    /// Does this instruction refer to an external struct explicitly?
+    pub fn contains_external_struct(&self) -> bool {
+        // Only cast instructions may contain an explicit reference to an external struct.
+        // Calls may produce them, but they don't explicitly reference the type, and that's
+        // always been allowed.
+        matches!(self,
+            Instruction::Cast(instruction) if instruction.cast_type().contains_external_struct()
```
vicsn, line 598: What about `CastLossy`?

console/program/src/data_types/register_type/mod.rs
```diff
@@ -34,6 +34,33 @@ pub enum RegisterType<N: Network> {
...
+            _ => self == rhs,
+        }
+    }
+
+    // Make unqualified structs or records into external ones with the given `id`.
+    pub fn qualify(self, id: ProgramID<N>) -> Self {
+        match self {
+            RegisterType::Plaintext(plaintext_type) => RegisterType::Plaintext(plaintext_type.qualify(id)),
+            RegisterType::Record(name) => RegisterType::ExternalRecord(Locator::new(id, name)),
+            RegisterType::ExternalRecord(..) | RegisterType::Future(..) => self,
```
vicsn, line 44: I don't want to cause a bigger refactor, this is purely a security question: why don't we care to have an `ExternalFuture`?

synthesizer/program/src/logic/instruction/operation/call.rs
```diff
@@ -222,10 +219,16 @@ impl<N: Network> Call<N> {
                 bail!("Expected {} outputs, found {}", closure.outputs().len(), self.destinations.len())
             }
             // Return the output register types.
-            Ok(closure.outputs().iter().map(|output| output.register_type()).cloned().collect())
+            Ok(closure
+                .output_types()
+                .into_iter()
+                // If the function is an external program, we need to qualify its structs or records with
+                // the appropriate ProgramID.
+                .map(|output_type| if is_external { output_type.qualify(*program.id()) } else { output_type })
```
vicsn, line 228: If I'm looking correctly, `qualify` can handle whether `is_external` or not. So either `is_external` is superfluous, or `qualify` should return a `Result`.

Same for the other `is_external` below.

synthesizer/program/src/parse.rs
```diff
@@ -78,6 +78,18 @@ impl<N: Network> Parser for ProgramCore<N> {
                 return map_res(take(0usize), Err)(string);
             }
         };
+
+        // Add the imports (if any) to the program.
```
vicsn, line 82: Why was this moved up?

synthesizer/tests/tests/process/execute/external_struct.aleo
```diff
@@ -0,0 +1,118 @@
...
+
+struct T:
+    x as field;
+
+struct OuterT:
+    x as field;
+    s as T;
+
+struct Outer2:
+    s as parent.aleo/S;
```
vicsn: Can we have a negative test with a non-extisting external struct which we expect to fail somewhere? As we don't check this at the parsing level: https://github.com/ProvableHQ/snarkVM/issues/2599

synthesizer/program/src/logic/instruction/operation/cast.rs
```diff
@@ -381,57 +395,16 @@ impl<N: Network, const VARIANT: u8> CastOperation<N, VARIANT> {
...
+                let external_stack = stack.get_external_stack(locator.program_id())?;
+                let plaintext = self.execute_cast_to_struct(&*external_stack, *locator.resource(), inputs)?;
+                // Store the struct.
+                registers.store_circuit(stack, &self.destination, plaintext.into())
+            }
+            CastType::Plaintext(PlaintextType::Struct(struct_name)) => {
+                let plaintext = self.execute_cast_to_struct(stack, *struct_name, inputs)?;
                 // Store the struct.
-                registers.store_circuit(stack, &self.destination, circuit::Value::Plaintext(struct_))
+                registers.store_circuit(stack, &self.destination, plaintext.into())
```
vicsn, line 415: Please double check this logic. It is now possible to cast data to an external struct in our circuit. But because those external structs are static and unaffected by program upgradability, no issues can occur.

Double clicking on this because program upgradability audit found issues with the way we handled **values** from external programs in the circuit.

vicsn (changes requested):
Very nice.

I would appreciate a short walk through synthesizer/program/src/logic/instruction/operation/cast.rs during the next Leo Weekly.

I'll defer to @d0cd about the usage and implementation of `types_structurally_equivalent` and whether he seed potential impending issues related to dynamic dispatch.

synthesizer/program/src/logic/instruction/mod.rs
mikebenfield: cast.lossy can't cast to a struct; this check exists

```
        if VARIANT == CastVariant::CastLossy as u8 {
            ensure!(
                matches!(self.cast_type, CastType::Plaintext(PlaintextType::Literal(..))),
                "`cast.lossy` is only supported for casting to a literal type"
            )
        }
```

Although if you'd rather be cautious, maybe with a view to future changes, I can check `CastLossy` here too.

synthesizer/program/src/function/mod.rs
mikebenfield: Yes; fixing

synthesizer/program/src/logic/instruction/operation/call.rs
mikebenfield: That would result in a struct type local to this program becoming a `PlaintextType::ExternalStruct`

synthesizer/program/src/parse.rs
mikebenfield: Leftover from my alternative approach to this issue; I'll move it back

synthesizer/program/src/logic/instruction/operation/cast.rs
mikebenfield: That seems correct to me

console/program/src/data_types/register_type/mod.rs
mikebenfield: I think `Future` essentially already is `ExternalFuture` - like `ExternalRecord` and `ExternalStruct` it has a `Locator`, not just an `Identifier`

synthesizer/src/vm/verify.rs
mikebenfield: That's possible of course, but note that as the code stands in the PR now once we hit `ConsensusVersion::V10` the check will never happen again, while if I make the check happen when creating `ProgramCore` it will continue to happen forever every time a `ProgramCore` is made.

synthesizer/tests/tests/parser/program/external_struct.aleo
mikebenfield: I added some more uses of external structs in this parser test, and note that `synthesizer/tests/tests/process/execute/external_struct.aleo` has some too.

synthesizer/src/vm/verify.rs
vicsn: Good point. It can stay as is!

Also, I realise this should be guarded by a new `ConsensusVersion::V10`. I'll add support for it in a separate PR next week.

synthesizer/program/src/parse.rs
mikebenfield: Oh actually I lied; this is necessary so that when we get to checks like this one

```
                PlaintextType::ExternalStruct(locator) => {
                    if !self.imports.contains_key(locator.program_id()) {
                        bail!(
                            "External program {} referenced in struct '{struct_name}' does not exist",
                            locator.program_id()
                        );
                    }
                }

```

in `add_struct`, the imports have already been added.

mikebenfield:
Here's some updates. 

You also asked me to walk you through the changes in `cast.rs`, which I didn't do yesterday. It's actually pretty straightforward so I'll just list what I've done here. The point is just to pull out the code that previously just applied to `Plaintext::Struct` into separate functions so it can be called when finding either `Plaintext::Struct` or `Plaintext::ExternalStruct`:

* Previously the code to actually carry out casting to structs was carried out in two places: the `cast_to_struct` function handled console values, while the circuit values were handled inline in the `execute` function. I factored out that common code to a `cast_to_struct_common` macro, used that to define separate `execute_cast_to_struct` and `evaluate_cast_to_struct` functions, and then called those functions when evaluating or executing where the destination type is `Struct` or `ExternalStruct`.

* Previously the code did a bunch of checks inline in `output_types` when the dest type what a struct. I pulled those checks into a closure called `struct_checks`, and now call that when matching either a `Struct` or an `ExternalStruct`.

synthesizer/program/src/logic/instruction/mod.rs
vicsn: No all good!

vicsn:
Can you make a PR to add clarification to the docs:
- https://github.com/AleoNet/welcome/blob/master/documentation/guides/aleo/04_opcodes.md?plain=1#L218
- https://github.com/AleoNet/welcome/blob/master/documentation/guides/aleo/04_opcodes.md?plain=1#L251
- https://github.com/AleoNet/welcome/blob/master/documentation/guides/aleo/04_opcodes.md?plain=1#L284 - it may be best to just refer back to the `assert.eq` documentation
- https://github.com/AleoNet/welcome/blob/master/documentation/guides/aleo/04_opcodes.md?plain=1#L314
- https://github.com/AleoNet/welcome/blob/master/documentation/guides/aleo/04_opcodes.md?plain=1#L435 - you mention in a comment above that we can't cast.lossy to a struct
- https://github.com/AleoNet/welcome/blob/master/documentation/guides/aleo/04_opcodes.md?plain=1#L1246 - again may be best to just refer back to earlier documentation
- https://github.com/AleoNet/welcome/blob/master/documentation/guides/aleo/04_opcodes.md?plain=1#L1279

vicsn:
Can you use this to confirm that the R1CS for all known deployments remains unchanged? https://github.com/ProvableHQ/aleo-program-regressions/tree/master/r1cs-hasher 

EDIT: It's been a while since we touched it, so may need some minor refactoring. Specifically, I see it is unclear which deployments have been used for the previous hash, I assume the first 211 of either canary/testnet/mainnet because the hash file name ends with an inexplicable `_211`. I'd advise computing a hash from scratch with all programs, using `staging` and `struct-namespace` snarkVM versions.

mikebenfield:
> Can you make a PR to add clarification to the docs:

Sure.

mikebenfield:
> Can you use this to confirm that the R1CS for all known deployments remains unchanged?

I can't get this to work on recent snarkvm commits. Using snarkvm `staging` branch (after making a couple changes to main.rs to account for snarkvm API changes):

```
% RUST_BACKTRACE=1 ~/.cargo/bin/r1cs-hasher programs/mainnet external-struct-mainnet
Processed 282 deployments

Initializing 'TestRng' with seed '0'


thread '<unnamed>' panicked at /Users/michaelbenfield/Code/snarkVM/circuit/environment/src/circuit.rs:293:9:
Surpassed the variable limit (41859)
```

UPDATE written by Victor: the tests passed now.

mikebenfield:
Regarding the docs about all the commands/instructions that compare equality of structs. Here's the actual behavior (both before and after this PR; it's unchanged):

1. If you're trying to compare two struct values and their types have different names, the program is rejected and snarkvm doesn't even try to run it;
2. If you're trying to compare two struct values and their types have the same name but different structure (like different member names or types), the program begins running but there will be a runtime error (not a failed constraint);
3. If you're trying to compare two struct values and their types have the same name and structure, then the comparison will work as expected.

I think this is not really behavior we want users to think about, but instead just want to users to only try to compare two struct values of the same type. For that reason I'm inclined to leave the documentation of those instructions as is, but let me know if you feel otherwise.

vicsn:
> I think this is not really behavior we want users to think about, but instead just want to users to only try to compare two struct values of the same type. For that reason I'm inclined to leave the documentation of those instructions as is, but let me know if you feel otherwise.

Sure that sounds reasonable, thx for taking the time.

synthesizer/src/vm/verify.rs
vicsn: V10 has now been merged into `staging`

synthesizer/tests/tests/process/execute/external_struct.aleo
mikebenfield: Is the test `external_struct_fail.aleo` adequate?

synthesizer/src/vm/verify.rs
```diff
@@ -455,7 +455,7 @@ fn ensure_deployment_valid_for_consensus_version<N: Network>(
         );
         ensure!(deployment.program_owner().is_some(), "Invalid deployment transaction '{id}' - missing program owner");
     }
-    if consensus_version <= ConsensusVersion::V9 {
+    if consensus_version < ConsensusVersion::V10 {
```
vicsn: While double checking documentation, I see I made a minor mistake, can you update this to say `external structs` not `external records`: https://github.com/ProvableHQ/snarkVM/blob/staging/console/network/src/consensus_heights.rs#L40

vicsn (approved):
LGTM!

console/program/src/data_types/array_type/parse.rs
```diff
@@ -14,15 +14,19 @@
...
 
 impl<N: Network> Parser for ArrayType<N> {
     /// Parses a string into a literal type.
     #[inline]
     fn parse(string: &str) -> ParserResult<Self> {
         // A helper function to parse the innermost element type.
         fn parse_inner_element_type<N: Network>(string: &str) -> ParserResult<PlaintextType<N>> {
-            alt((map(LiteralType::parse, PlaintextType::from), map(Identifier::parse, PlaintextType::from)))(string)
+            alt((
+                map(Locator::parse, PlaintextType::from),
```
d0cd, line 27: Worth commenting that order of the parsers matter.

console/program/src/data_types/plaintext_type/mod.rs
```diff
@@ -26,14 +26,60 @@ pub enum PlaintextType<N: Network> {
     /// A literal type contains its type name.
     /// The format of the type is `<type_name>`.
     Literal(LiteralType),
-    /// An struct type contains its identifier.
+    /// A struct type contains its identifier.
     /// The format of the type is `<identifier>`.
     Struct(Identifier<N>),
+    /// An external struct type contains its locator.
```
d0cd: Nit. Can we change the ordering so that `ExternalStruct` comes that the end? And ditto for all of the enums.

This is to ensure that the serialization variants follow the enum declaration order.

console/program/src/data_types/value_type/mod.rs
```diff
@@ -52,6 +52,15 @@ impl<N: Network> ValueType<N> {
             ValueType::Future(..) => 5,
         }
     }
+
+    /// Does this type refer to an external struct explicitly?
```
d0cd:
```suggestion
    /// Returns whether this type references an external struct.
```

synthesizer/program/src/closure/mod.rs
```diff
@@ -69,6 +69,18 @@ impl<N: Network> ClosureCore<N> {
     pub const fn outputs(&self) -> &IndexSet<Output<N>> {
         &self.outputs
     }
+
+    /// Returns the closure output types.
+    pub fn output_types(&self) -> Vec<RegisterType<N>> {
+        self.outputs.iter().map(|output| output.register_type()).cloned().collect()
+    }
+
+    /// Does this closure refer to an external struct explicitly?
```
d0cd:
```suggestion
    /// Returns whether the closure refers to an external struct.
```

synthesizer/program/src/constructor/mod.rs
```diff
@@ -50,6 +50,16 @@ impl<N: Network> ConstructorCore<N> {
     pub const fn positions(&self) -> &HashMap<Identifier<N>, usize> {
         &self.positions
     }
+
+    /// Does this constructor refer to an external struct explicitly?
```
d0cd: Nit. Match commenting style for rest of codebase.

synthesizer/program/src/function/mod.rs
d0cd: Is the fix still needed?

d0cd (commented):
We will also want `vm_execute_and_finalize` tests.

It would also be good to test `MAX_DATA_DEPTH` bounds and whether or not they can be circumvented with external structs.

console/program/src/data_types/plaintext_type/mod.rs
```diff
@@ -26,14 +26,60 @@ pub enum PlaintextType<N: Network> {
...
+impl<N: Network> PlaintextType<N> {
+    /// Are the two types equivalent for the purposes of static checking?
+    ///
+    /// Since struct types are compared by structure, we can't determine equality
+    /// by only looking at their names.
+    pub fn equal_or_structs(&self, rhs: &Self) -> bool {
+        use PlaintextType::*;
+
+        match (self, rhs) {
+            (ExternalStruct(..) | Struct(..), ExternalStruct(..) | Struct(..)) => true,
```
d0cd: What if we have two `Struct` with the same `Identifier`?

mikebenfield: What do you mean? Why would this be a problem?

d0cd: They would not be equal right?

d0cd: Ah I read this pattern wrong. 
Follow on, do we want two structs with different `Identifier`s to be equal or not equal?

d0cd: Based on the description, if structs have the same data but different names, they are still the same.

synthesizer/program/src/lib.rs
d0cd, line 1: Do there need to checks analagous to `add_struct` in `add_mapping`?

synthesizer/program/src/logic/instruction/operation/call.rs
```diff
@@ -222,10 +219,16 @@ impl<N: Network> Call<N> {
                 bail!("Expected {} outputs, found {}", closure.outputs().len(), self.destinations.len())
             }
             // Return the output register types.
-            Ok(closure.outputs().iter().map(|output| output.register_type()).cloned().collect())
+            Ok(closure
+                .output_types()
+                .into_iter()
+                // If the function is an external program, we need to qualify its structs or records with
```
d0cd:
```suggestion
                // If the function is an external program, we need to qualify its structs
```
Prolly copy pasta, but a closure cannot return a record.

synthesizer/program/src/logic/instruction/operation/call.rs
```diff
@@ -181,16 +181,17 @@ impl<N: Network> Call<N> {
...
     /// Returns the output type from the given program and input types.
-    #[inline]
     pub fn output_types(
         &self,
         stack: &impl StackTrait<N>,
         input_types: &[RegisterType<N>],
     ) -> Result<Vec<RegisterType<N>>> {
-        // Retrieve the external stack, if needed, and the resource.
-        let (external_stack, resource) = match &self.operator {
+        let stack_value;
```
d0cd, line 190: Nit. Keep the old comment.

d0cd (commented):
Have regression tests been run on existing programs? @vicsn

synthesizer/program/src/logic/instruction/operation/cast.rs
```diff
@@ -740,7 +731,7 @@ impl<N: Network, const VARIANT: u8> CastOperation<N, VARIANT> {
                         // Ensure the plaintext type matches the member type.
                         RegisterType::Plaintext(plaintext_type) => {
                             ensure!(
-                                plaintext_type == array_type.next_element_type(),
+                                plaintext_type.equal_or_structs(array_type.next_element_type()),
```
d0cd: Does this eventually need to check for structural equality?

synthesizer/program/src/traits/stack_and_registers.rs
```diff
@@ -150,6 +151,84 @@ pub trait StackTrait<N: Network> {
     ) -> Result<Record<N, Plaintext<N>>>;
 }
 
+/// Are the two types either the same, or both structurally equivalent `PlaintextType`s?
+pub fn register_types_structurally_equivalent<N: Network>(
```
d0cd: Is it necessary to pass in two stacks or can we just use one and resolve external stacks via `get_external_stack`?
This would have the side-effect of checking imports are defined for any external stack you get.

Ditto for below.

synthesizer/src/vm/verify.rs
```diff
@@ -449,6 +407,64 @@ impl<N: Network, C: ConsensusStorage<N>> VM<N, C> {
...
+    //   - the deployment edition is zero.
+    // If the `CONSENSUS_VERSION` is less than `V9` ensure that
+    //   - the deployment edition is zero or one.
+    //   - the program checksum is **not** present in the deployment,
+    //   - the program owner is **not** present in the deployment
+    //   - the program does not use constructors, `Operand::Checksum`, `Operand::Edition`, or `Operand::ProgramOwner`
+    // If the `CONSENSUS_VERSION` is greater than or equal to `V9`, then verify that:
+    //   - the program checksum is present in the deployment
+    //   - the program owner is present in the deployment
+    // If the `CONSENSUS_VERSION` is less than or equal to `V9`, then verify that:
```
d0cd:
```suggestion
    // If the `CONSENSUS_VERSION` is less than `V10`, then verify that:
```

d0cd (commented):
Overall, I believe this approach is sound. There are still some nits and questions to be addressed.

Some analysis:
- snarkVM `Value::Plaintext::Struct` do not store their name. The value itself only contains the members of the struct. So supporting structural equality does not affect the concrete value that is stored in console or circuit.
- Additionally program upgradability allows for cyclic references between programs. However, it is not possible to create cyclic structs because upgradability does not allow a struct's definition to be modified.

Some open questions:
- Is there any place where two structurally equal structs are **not** equal. We should document such special cases, if any.
- Is there any place where "resolving the external struct" can lead to a stack overflow? Basically having something deeply nested enough that VM craps out.

@mikebenfield feel free to throw in any additional justification or analysis.

d0cd (changes requested):
Thank you for your work and your patience with the reviews thus far @mikebenfield!
Taking an extra paranoid approach since this is a super sensitive part of our stack.
Requesting changes on the above feedback.

synthesizer/program/src/lib.rs
mikebenfield: `add_mapping` already does not have any of these checks for its types - `self.structs.contains_key(identifier)` doesn't happen there.

synthesizer/program/src/logic/instruction/operation/cast.rs
mikebenfield: Yes, that happens later. Note that `==` already didn't check for structural equality - only if the structs had the same names.

synthesizer/program/src/traits/stack_and_registers.rs
mikebenfield: The recursive call requires two separate stacks, since the different members of the different structs may be structurally equivalent structs belonging to two different programs, neither of which imports the other. The function still does `.get_external_stack` so I think it'll still catch any missing imports.

mikebenfield:
> Is there any place where two structurally equal structs are not equal.

Yes, see [my comment above](https://github.com/ProvableHQ/snarkVM/pull/2798#issuecomment-3136901733).

It's kind of tricky to explain exactly how the typing of structs works, but in many places it's actually just straightforward static typing, because previously you weren't able to easily refer to struct types from other programs - you had to redefine your own. It really only gets weird, and you have this structural equivalence of types, when a call to an external program returns a struct - otherwise, until this external structs PR, that was the only way you'd get a struct of a type from another program. 

> Is there any place where "resolving the external struct" can lead to a stack overflow? Basically having something deeply nested enough that VM craps out.

`types_structurally_equivalent` will recurse as long as you've got structs with external struct members. If we're worried about structs nested thousands deep I can switch to an iterative implementation (with heap allocation).

mikebenfield:
> We will also want vm_execute_and_finalize tests.

I added a test, but basically just copying the other test I made. Is there anything else specifically you'd like to see there?

> It would also be good to test MAX_DATA_DEPTH bounds and whether or not they can be circumvented with external structs.

How do you mean?

vicsn:
> Have regression tests been run on existing programs? @vicsn

r1cs-hasher, compare-program-serde and check-deserialization all ran succesfully.

synthesizer/program/src/traits/stack_and_registers.rs
d0cd: Understood, sort of like
```
A    B
 \   /
   C
```
    
where `A` and `B` are structurally equivalent.
This would be an interesting test case to have.

d0cd:
> > Have regression tests been run on existing programs? @vicsn
> 
> r1cs-hasher, compare-program-serde and check-deserialization all ran succesfully.

I think you'll also need to run `profile_process_load_from_store` to check that all programs can be loaded successfully into the VM with the changes on this branch.

Unless `r1cs-hasher` implicitly loads all deployed programs on each of the networks into the `Process`.

synthesizer/program/src/lib.rs
d0cd: Is that an issue, or does the check happen elsewhere?

synthesizer/program/src/logic/instruction/operation/cast.rs
d0cd: That's right, but my understanding is that before this was okay because the name of the struct was used in equality. Does that match yours?

mikebenfield:
Alright, regarding our earlier discussion, I realized it's actually not possible to relax type checking of structs separately from this external structs PR. In current snarkVM, having a struct with the expected name in the current program is the only way to figure out what the structure of a struct is supposed to be. So here I've just replaced all the places where types were checked using `==` with calls to the function `types_structurally_equivalent`.

vicsn:
> Unless r1cs-hasher implicitly loads all deployed programs on each of the networks into the Process.

It does: https://github.com/ProvableHQ/aleo-program-regressions/blob/master/r1cs-hasher/src/main.rs#L161C14-L161C29

console/program/src/data_types/array_type/bytes.rs
```diff
@@ -23,7 +23,8 @@ impl<N: Network> FromBytes for ArrayType<N> {
         let element_type = match variant {
             0 => PlaintextType::Literal(LiteralType::read_le(&mut reader)?),
             1 => PlaintextType::Struct(Identifier::read_le(&mut reader)?),
-            2.. => return Err(error(format!("Failed to deserialize element type {variant}"))),
+            3 => PlaintextType::ExternalStruct(Locator::read_le(&mut reader)?),
```
d0cd: This is the internal data format for an `ArrayType`. Would it make sense to use `2` here?
Ditto for the `ToBytes` impl.

console/program/src/data_types/plaintext_type/parse.rs
```diff
@@ -60,6 +62,7 @@ impl<N: Network> Display for PlaintextType<N> {
             Self::Literal(literal) => Display::fmt(literal, f),
             // Prints the struct, i.e. signature
             Self::Struct(struct_) => Display::fmt(struct_, f),
+            Self::ExternalStruct(locator) => Display::fmt(locator, f),
```
d0cd, line 66:
```suggestion
            // Prints the external struct, i.e. foo.aleo/bar
            Self::ExternalStruct(locator) => Display::fmt(locator, f),
```

synthesizer/process/src/stack/finalize_types/initialize.rs
d0cd, line 1: Do we need a structurally equivalent check in `check_branch` and on `get_or_use` for the default? More generally, there are a few (in)equality checks between `FinalizeTypes`. Do any of them need a structural equivalence check on their variants?

d0cd: One way we could ensure that we haven't missed any equality checks is by removing all invocations of == or `!=` in this file and using defined functions. If we go this route, probably best to leave a comment so that future implementers know.

synthesizer/process/src/stack/finalize_types/mod.rs
```diff
@@ -157,6 +158,17 @@ impl<N: Network> FinalizeTypes<N> {
                         None => bail!("'{identifier}' does not exist in struct '{struct_name}'"),
                     }
                 }
+                // Access the member on the path to output the register type.
+                (FinalizeType::Plaintext(PlaintextType::ExternalStruct(locator)), Access::Member(identifier)) => {
+                    // Retrieve the member type from the struct and check that it exists.
+                    let external_stack = stack.get_external_stack(locator.program_id())?;
```
d0cd, line 164: Note. Suppose that first access returns a struct and then a second access goes to an external struct in that one. e.g `A -> B -> C`. It must be the case that program `A` also imports program `C`. This is fine, just something to keep in mind on the Leo side.

synthesizer/process/src/stack/finalize_types/initialize.rs
d0cd: Ditto for register types.

synthesizer/process/src/stack/register_types/matches.rs
d0cd, line 1: Do we need equivalent structural checks in `finalize_types/matches.rs`?

synthesizer/program/src/logic/instruction/operation/call.rs
```diff
@@ -239,18 +243,14 @@ impl<N: Network> Call<N> {
...
-                    (true, ValueType::Record(record_name)) => Ok(RegisterType::ExternalRecord(Locator::from_str(
-                        &format!("{}/{}", program.id(), record_name),
-                    )?)),
-                    // Else, return the register type.
-                    (_, output_type) => Ok(RegisterType::from(output_type)),
-                })
-                .collect::<Result<Vec<_>>>()
+                .map(RegisterType::from)
+                // If the function is an external program, we need to qualify its structs or records with
+                // the appropriate ProgramID.
```
d0cd:
```suggestion
                // the appropriate `ProgramID`.
```

synthesizer/program/src/logic/instruction/operation/call.rs
```diff
@@ -222,10 +220,16 @@ impl<N: Network> Call<N> {
                 bail!("Expected {} outputs, found {}", closure.outputs().len(), self.destinations.len())
             }
             // Return the output register types.
-            Ok(closure.outputs().iter().map(|output| output.register_type()).cloned().collect())
+            Ok(closure
+                .output_types()
+                .into_iter()
+                // If the function is an external program, we need to qualify its structs with
+                // the appropriate ProgramID.
```
d0cd:
```suggestion
                // the appropriate `ProgramID`.
```

synthesizer/src/vm/verify.rs
```diff
@@ -449,6 +407,64 @@ impl<N: Network, C: ConsensusStorage<N>> VM<N, C> {
...
+    }
+    if consensus_version >= ConsensusVersion::V9 {
+        ensure!(
+            deployment.program_checksum().is_some(),
+            "Invalid deployment transaction '{id}' - missing program checksum"
+        );
+        ensure!(deployment.program_owner().is_some(), "Invalid deployment transaction '{id}' - missing program owner");
+    }
+    if consensus_version < ConsensusVersion::V10 {
+        ensure!(
```
d0cd: Sanity check. Is there any program that was not deployable in V9 but is deployable in V10 but doesn't explicitly use external structs?

synthesizer/tests/tests/vm/execute_and_finalize/mapping_with_nonexistent_structs.aleo
d0cd, line 1: Sanity check. We are okay with this program being deployable.

synthesizer/tests/tests/process/execute/external_struct.aleo
```diff
@@ -0,0 +1,118 @@
...
+    output r0 as external_struct_parent.aleo/S.public;
+
+/////////////////////////////////////////////////
+
+import child.aleo;
+
+program child2.aleo;
+
+function call_make_outer2:
+    call child.aleo/make_outer2 into r0;
```
d0cd, line 121: Could be interesting to test calling `child.aleo/make_s` and then accessing its fields.

d0cd (changes requested):
We're getting closer! Thanks for the continued effort 💪

synthesizer/src/vm/verify.rs
vicsn: External structs is the only addition in terms of program structure.

Additionally, the new fees will be strictly lower so a deployment with too low fee will be accepted from V10, but that's checked elsewhere in the currently open fee PR.

vicsn: Actually, maybe this child program... https://github.com/ProvableHQ/snarkVM/pull/2798/files/832e38a3462fa76de6b78c0468357f8f3760291c#r2292636568

If indeed that program was not deployable before this PR, then to accurately avoid any risk of forking, here in vm/verify.rs we would have to only allow programs with non-existent structs from V10 onwards.

Is there any way for us to not allow deploying non-existent structs? Seems like a footgun. Can we check for this during deployment verification?

synthesizer/tests/tests/vm/execute_and_finalize/mapping_with_nonexistent_structs.aleo
vicsn: Hmm preferably not, see https://github.com/ProvableHQ/snarkVM/pull/2798#discussion_r2293583373

vicsn:
> Let two struct types be considered the same type as long as they are structurally the same.

Sorry to question this so late in the game but this does seem like bad UX to me. I was only focusing on security in earlier reviews, whereas I should have started with a UX analysis. If I define two structurally equivalent token structs representing a user's token balances, just with different names, then they should not be deemed equal.

I could see the following high level options going forward:
1. We ensure `is.eq r0 r1 into r2;` is indeed only true if the fully qualified name is the same.
2. We disable the equality operator on structs, not sure if that breaks previous programs. This is not ideal, but its an option if 1. is too difficult to achieve.

mikebenfield:
> Sorry to question this so late in the game but this does seem like bad UX to me. 

Why? To me it seems like a dramatic improvement over the current situation. 

 
> 1. We ensure `is.eq r0 r1 into r2;` is indeed only true if the fully qualified name is the same.
> 2. We disable the equality operator on structs, not sure if that breaks previous programs. This is not ideal, but its an option if 1. is too difficult to achieve.

Neither of these options is feasible while keeping backwards compatibility. In particular with the second one, we'd have to disable a lot more than the equality operator.

synthesizer/tests/tests/vm/execute_and_finalize/mapping_with_nonexistent_structs.aleo
mikebenfield: This program (`child.aleo`, with `parent.aleo` removed since it uses the external struct notation) appears to deploy just fine on `staging`

synthesizer/src/vm/verify.rs
mikebenfield: That program `child.aleo` will deploy before this PR. Would you like me to add a check to make it not deployable?

console/network/src/consensus_heights.rs
```diff
@@ -37,7 +37,7 @@ pub enum ConsensusVersion {
     V8 = 8,
     /// V9: Support for program upgradability.
     V9 = 9,
-    /// V10: Support for external records.
+    /// V10: Support for external structs.
```
vicsn: I'll make a PR to introduce a V11.

synthesizer/src/vm/verify.rs
vicsn: > That program child.aleo will deploy before this PR. Would you like me to add a check to make it not deployable?

Yes please. Same with non-existent records which are referenced if that was possible today.

mikebenfield: Don't think it's possible with records - this was only an issue with mapping keys/values, which can't be records.

I added the check to `Stack::initialize_and_check` since that's ultimately where the checks for nonexistent structs in functions, closures, etc, come from.

vicsn: Thank you! I think we may have to guard this additional check by the consensus version so it only applies to future programs. So this whole piece of logic should be **moved** (not copied) to being called somewhere from check_transaction:
```
for mapping in self.program.mappings().values() {
            // These calls make sure structs exist.
            RegisterTypes::check_plaintext_type(self, mapping.key().plaintext_type())?;
            RegisterTypes::check_plaintext_type(self, mapping.value().plaintext_type())?;
 }
```

console/network/src/consensus_heights.rs
vicsn: Merged into staging, please update

mikebenfield: updated

synthesizer/program/src/logic/instruction/operation/call.rs
vicsn: Can I veto against these types of nits :)

synthesizer/program/src/logic/instruction/operation/call.rs
vicsn: Can I veto against these types of nits :)

synthesizer/process/src/stack/finalize_types/initialize.rs
mikebenfield: I've got checks in `check_branch` and `get_or_use`. There don't appear to be any uses of `==` on `FinalizeTypes` in this file. 

There's a checkin `regster_types/initialize.rs` at what I think is the only place there needs to be.

synthesizer/process/src/stack/register_types/matches.rs
mikebenfield: I've got some checks in there and a comment explaining where it's not needed.

mikebenfield:
@d0cd I've checked your questions again; I added a couple comments but I think the code as of the last push had addressed all the issues you raised. Let me know what you think.

synthesizer/src/vm/verify.rs
vicsn: @mikebenfield we need to move this check out of `Stack::initialize_and_check` and into `ensure_deployment_valid_for_consensus_version` in order to ensure this is only invalidated from the new consensus version onwards.

mikebenfield: Right, somehow I misread what you wrote previously. Alright I've moved it now.

synthesizer/src/vm/verify.rs
```diff
@@ -188,50 +190,9 @@ impl<N: Network, C: ConsensusStorage<N>> VM<N, C> {
...
-                        deployment.program_checksum().is_some(),
-                        "Invalid deployment transaction '{id}' - missing program checksum"
-                    );
-                    ensure!(
-                        deployment.program_owner().is_some(),
-                        "Invalid deployment transaction '{id}' - missing program owner"
-                    );
-                }
+
+                let stack = self.process.read().get_stack(deployment.program_id())?;
```
d0cd: Is the stack guaranteed to exist at this point?

synthesizer/src/vm/verify.rs
```diff
@@ -481,6 +442,73 @@ impl<N: Network, C: ConsensusStorage<N>> VM<N, C> {
...
+    //   - the deployment edition is zero.
+    // If the `CONSENSUS_VERSION` is less than `V9` ensure that
+    //   - the deployment edition is zero or one.
+    //   - the program checksum is **not** present in the deployment,
+    //   - the program owner is **not** present in the deployment
+    //   - the program does not use constructors, `Operand::Checksum`, `Operand::Edition`, or `Operand::ProgramOwner`
+    // If the `CONSENSUS_VERSION` is greater than or equal to `V9`, then verify that:
+    //   - the program checksum is present in the deployment
+    //   - the program owner is present in the deployment
+    // If the `CONSENSUS_VERSION` is less than `V10`, then verify that:
```
d0cd:
```suggestion
    // If the `CONSENSUS_VERSION` is less than `V11`, then verify that:
```

synthesizer/src/vm/verify.rs
```diff
@@ -481,6 +442,73 @@ impl<N: Network, C: ConsensusStorage<N>> VM<N, C> {
...
+    //   - the program does not use the external struct syntax `some_program.aleo/StructT`
```
d0cd:
```suggestion
    //   - the program does not use the external struct syntax `some_program.aleo/StructT`
    //   - the program's mappings do not use non-existent structs.
```

synthesizer/src/vm/verify.rs
vicsn: Ooff good catch. No it's not. And we don't even need it, we just read the `program` from it. We should add some integration tests in a new `test_v11.rs` file. You can copy behaviour from `test_v10.rs` and `test_v9.rs`: e.g. simply create a transaction with external struct before the V11 height, should fail to verify, after the V11 height, should succeed to verify.

mikebenfield: Good grief I'm a mess; sorry. I actually already fixed this locally before I even pushed this but somehow pushed the wrong commit. I've now pushed the fix, along with the changes to comments. I'll add the `test_v11.rs` shortly.

synthesizer/program/src/logic/instruction/operation/is.rs
```diff
@@ -131,15 +131,15 @@ impl<N: Network, const VARIANT: u8> IsInstruction<N, VARIANT> {
...
+        stack: &impl StackTrait<N>,
         input_types: &[RegisterType<N>],
     ) -> Result<Vec<RegisterType<N>>> {
         // Ensure the number of input types is correct.
         if input_types.len() != 2 {
             bail!("Instruction '{}' expects 2 inputs, found {} inputs", Self::opcode(), input_types.len())
         }
         // Ensure the operands are of the same type.
-        if input_types[0] != input_types[1] {
+        if !register_types_structurally_equivalent(stack, &input_types[0], stack, &input_types[1])? {
```
vicsn: Forking risk, transaction validity changed

synthesizer/tests/tests/process/execute/external_struct_fail.aleo
```diff
@@ -0,0 +1,27 @@
...
+
+import external_struct_fail_parent.aleo;
+
+program external_struct_fail_child.aleo;
+
+struct T:
+    x as field;
+
+function make_s:
+    // This should fail as the struct doesn't exist.
```
vicsn, line 25: Could we add one more test here or in vm/execute_and_finalize, I think I didn't see this yet: `call external_struct_fail_parent.aleo/make_s as T`, this should fail right because the local names are not the same?

synthesizer/program/src/logic/instruction/operation/call.rs
vicsn: Following up on this, I guess what my recommendation should actually have been, is that we could consider only qualifying local types, and letting qualify return a Result. However, that just seems like a useless refactor. Noting it here in case I come to more bad ideas.

synthesizer/src/vm/verify.rs
```diff
@@ -479,6 +439,104 @@ impl<N: Network, C: ConsensusStorage<N>> VM<N, C> {
...
+                self.plaintext_exists(mapping.key().plaintext_type(), deployment.program())?;
+                self.plaintext_exists(mapping.value().plaintext_type(), deployment.program())?;
+            }
+        }
+
+        Ok(())
+    }
+
+    // If `type_` is a struct or an array containing a struct, ensure the struct type exists.
+    fn plaintext_exists(&self, type_: &PlaintextType<N>, program: &Program<N>) -> Result<()> {
```
vicsn: Can this be moved to a lower level part of the code where we define program-related syntax?

Perhaps including the for-loop calling it.

mikebenfield: I moved it over to Process; lmk if that's not what you had in mind.

synthesizer/tests/tests/process/execute/external_struct_fail.aleo
mikebenfield: I added another test

vicsn:
@mikebenfield one of the synthesizer tests halted with a mysterious `VM safely halted at /home/circleci/.cargo/registry/src/index.crates.io-1949cf8c6b5b557f/itertools-0.14.0/src/zip_eq_impl.rs:50:17: itertools: .zip_eq() reached end of one iterator before the other`. Would be great if you can try to reproduce.

Note that a small improvement to `test_vm_execute_and_finalize` tests along the way may both speed up your debugging and be useful for the future:
- would there have been any way for this custom test runner to log which test failed on this particular error, so that for reproducing you'd only have to run a single one?
- speaking of running a single test, it is currently not possible or easy to see how to choose a single sub-test to run in `test_vm_execute_and_finalize`

mikebenfield:
@vicsn Alright I fixed that test but unfortunately didn't enhance the test function. 

There's now a semver check failing, but I don't think that is related to this PR.

vicsn:
@mohammadfawaz can records contain structs and/or external structs? Do we have tests for this in place? CC @Antonio95 for his work on dynamic dispatch

mohammadfawaz:
> @mohammadfawaz can records contain structs and/or external structs? Do we have tests for this in place? CC @Antonio95 for his work on dynamic dispatch

This should be totally okay! I'll make sure there are tests in place for that.

Antonio95:
> @mohammadfawaz can records contain structs and/or external structs? Do we have tests for this in place? CC @Antonio95 for his work on dynamic dispatch

Apologies for the late reply (I thought I was tagged to review). @mohammadfawaz's reply is definitely better informed than mine, but [some of my recent tests](https://github.com/ProvableHQ/snarkVM/blob/6cade2fc6e0818a27249e4e4a3969b9fbd78f97d/synthesizer/process/src/trace/translation/tests.rs#L362-L365) from the translation circuit do include records with (nested) structures. In a concrete record itself (not its definition), structures (and their nested structures, if any) are specified explicitly field by field, so the in-program vs. extern distinction isn't relevant there (but it may be in the in-program definition).

mohammadfawaz:
Closing in favor of https://github.com/ProvableHQ/snarkVM/pull/3045

PR #2799, opened 2025-07-07
Enable setting custom consensus height for tests

vicsn:
## Motivation

We often find ourselves setting a custom `LATEST_CONSENSUS_HEIGHT` manually, now it can be set in an environment variable while building snarkOS.

## Test Plan

Tested manually that it works.


d0cd:
If we want more flexibility than just the last height, we can consider [this](https://github.com/ProvableHQ/snarkVM-staging/blob/b949a184badd00bd055b4434b2a21791c4ea9009/console/network/src/lib.rs#L256) approach.
In fact, this work was based off your original implementation when we were testing migrations.

vicsn:
> If we want more flexibility than just the last height, we can consider this approach.
> In fact, this work was based off your original implementation when we were testing migrations.

Thx for sharing, great suggestion. I added some of the modifications here, with a focus of ensuring we have to do as few manual adjustments as possible when introducing a new consensus version: c915addc5

console/network/src/lib.rs
```diff
@@ -228,35 +233,101 @@ pub trait Network:
...
+    /// Returns the list of test consensus versions.
+    #[allow(non_snake_case)]
+    #[cfg(any(test, feature = "test", feature = "test_consensus_heights"))]
+    fn CONSENSUS_VERSION_HEIGHTS() -> &'static [(ConsensusVersion, u32); NUM_CONSENSUS_VERSIONS] {
+        // NOTE: this closure may panic, as it is only called during startup.
+        CONSENSUS_VERSION_HEIGHTS.get_or_init(|| {
+            // Define consensus version heights used for testing.
+            let mut test_consensus_heights = Self::_CONSENSUS_VERSION_HEIGHTS;
+            // Assert that the genesis height is 0.
+            assert_eq!(test_consensus_heights[0].1, 0, "Genesis height must be 0.");
```
raychu86: Why do these checks need to be done for the test heights when it's not done for the real ones?

raychu86: Can there be a helper function that checks all of these for both cases? It would keep the Network trait a bit cleaner as well.

Additionally, this particular check is not done for the heights passed in via ENV variable.

vicsn: Yeah I went back and forth on this actually, cleaned it up further: 58334b36d let me know if this is what you had in mind.

console/network/src/lib.rs
```diff
@@ -228,35 +233,111 @@ pub trait Network:
...
+                    // Set the parsed heights in the test consensus heights.
+                    for (i, height) in parsed_test_consensus_heights.into_iter().enumerate() {
+                        test_consensus_heights[i] = (Self::TEST_CONSENSUS_VERSION_HEIGHTS[i].0, height);
+                    }
+                    // Verify and return the test consensus heights.
+                    verify_consensus_heights(&test_consensus_heights);
+                    test_consensus_heights
+                }
+                Err(_) => {
+                    // Set the first height after genesis to 10.
```
d0cd: Could we just use the `TEST_CONSENSUS_VERSION_HEIGHTS` directly, instead of setting and verifying?

vicsn: Thx for catching ea89281b0 - still verifying just in case.

raychu86 (approved):

PR #2800, opened 2025-07-07
Fail loudly when using the wrong program or record version

vicsn:
## Motivation

An execution service in the wild which is on the latest snarkVM version but either 1. isn't synced 2. cached outdated verifying keys 3. uses the wrong record at the wrong time, will find their transactions don't land without clear error.

## Test Plan

- [x] Should be E2E tested on latest snarkOS.

## Related PRs

https://github.com/ProvableHQ/snarkVM/pull/2796 introduces a similar protection, but we deprecate it in favour of this PR so we can still flexibly test any behaviour during testing.

### DIFF
--- a/synthesizer/process/src/stack/authorization/mod.rs
+++ b/synthesizer/process/src/stack/authorization/mod.rs
@@ -152,13 +152,11 @@ impl<N: Network> Authorization<N> {
     /// Checks whether the authorization is for a valid program edition.
     pub fn check_valid_edition(&self, process: &crate::Process<N>, consensus_version: ConsensusVersion) -> Result<()> {
         // Determine the root transition's program ID.
-        let root_program_id = {
-            let transitions = self.transitions.read();
-            *transitions
-                .first()
-                .map(|(_, t)| t.program_id())
-                .ok_or_else(|| anyhow!("No transitions found in the Authorization."))?
-        };
+        let transitions = self.transitions.read();
+        let root_program_id = *transitions
+            .first()
+            .map(|(_, t)| t.program_id())
+            .ok_or_else(|| anyhow!("No transitions found in the Authorization."))?;
         // There is only one credits.aleo edition, so we can safely skip this case.
         if root_program_id.to_string() != "credits.aleo" {
             // Get the program's current edition.
@@ -166,6 +164,52 @@ impl<N: Network> Authorization<N> {
             // If we're past ConsensusVersion::V8, ensure new stacks are not on edition 0.
             if consensus_version >= ConsensusVersion::V8 && program_edition == 0 {
                 bail!("Cannot execute {} on edition {program_edition}", root_program_id.to_string());
+            }
+        }
+        Ok(())
+    }
+
+    /// Checks whether the authorization creates valid records.
+    pub fn check_valid_records(&self, consensus_version: ConsensusVersion) -> Result<()> {
+        let transitions = self.transitions.read();
+        // Collect the transition's records.
+        let records = transitions.values().map(|transition| {
+            let program_id = transition.program_id();
+            let function_name = transition.function_name();
+            let input_records = transition.outputs().iter().filter_map(|output| output.record());
+            let output_records = transition.outputs().iter().filter_map(|output| output.record());
+            (program_id, function_name, input_records, output_records)
+        });
+        // Ensure the records are valid.
+        for (program_id, function_name, input_records, output_records) in records {
+            for (_, record) in output_records {
+                // If the consensus version are before `ConsensusVersion::V8`, ensure the output record is on Version 0.
+                // if the consensus version is on or after `ConsensusVersion::V8`, ensure the output record is on Version 1.
+                if (ConsensusVersion::V1..=ConsensusVersion::V7).contains(&consensus_version) {
+                    #[cfg(not(any(test, feature = "test")))]
+                    ensure!(record.version().is_zero(), "Output record must be Version 0 before Consensus V8");
+                    #[cfg(any(test, feature = "test"))]
+                    ensure!(record.version().is_one(), "Output record must be Version 1 before Consensus V8");
+                } else {
+                    ensure!(record.version().is_one(), "Output record must be Version 1 on or after Consensus V8");
+                }
+            }
+            for (_, record) in input_records {
+                // if the consensus version is on or after `ConsensusVersion::V8`,
+                // Ensure version 0 credits.aleo records can only be used when calling upgrade().
+                if consensus_version >= ConsensusVersion::V8
+                    && **record.version() == 0
+                    && program_id.to_string() == "credits.aleo"
+                    && function_name.to_string() != "upgrade"
+                {
+                    bail!(
+                        "Cannot input record version {} on consensus version {:?} for program {} and function {}",
+                        record.version(),
+                        consensus_version,
+                        program_id,
+                        function_name
+                    );
+                }
             }
         }
         Ok(())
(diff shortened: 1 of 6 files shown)


synthesizer/process/src/authorize.rs
```diff
@@ -26,8 +26,12 @@ impl<N: Network> Process<N> {
         inputs: impl ExactSizeIterator<Item = impl TryInto<Value<N>>>,
         rng: &mut R,
     ) -> Result<Authorization<N>> {
+        // Get the program's Stack.
+        let stack = self.get_stack(program_id)?;
+        // Ensure the stack is on program edition 1 before executing.
+        ensure!(*stack.program_edition() > 0, "Cannot execute on program edition {}", *stack.program_edition());
```
raychu86: What about `credits.aleo` that is always on edition 0? 

Also, edition 0 program executions should still work prior to consensus V8

vicsn: https://github.com/ProvableHQ/snarkVM/pull/2800/commits/33e725d5c6cec5718cbbdc740ceb249df5a110c4

synthesizer/process/src/stack/authorization/mod.rs
```diff
@@ -147,6 +148,73 @@ impl<N: Network> Authorization<N> {
...
+            .map(|(_, t)| t.program_id())
+            .ok_or_else(|| anyhow!("No transitions found in the Authorization."))?;
+        // There is only one credits.aleo edition, so we can safely skip this case.
+        if root_program_id.to_string() != "credits.aleo" {
+            // Get the program's current edition.
+            let _program_edition = *process.get_stack(root_program_id)?.program_edition();
+            // If we're past ConsensusVersion::V8, ensure new stacks are not on edition 0.
+            #[cfg(not(any(test, feature = "test")))]
+            if _consensus_version >= ConsensusVersion::V8 && _program_edition == 0 {
+                bail!("Cannot execute {} on edition {_program_edition}", root_program_id.to_string());
```
mikebenfield:
```suggestion
                bail!("Cannot execute {root_program_id} on edition {_program_edition}");
```
nit: don't need `to_string()`

synthesizer/process/src/verify_execution.rs
```diff
@@ -115,7 +115,7 @@ impl<N: Network> Process<N> {
                         #[cfg(any(test, feature = "test"))]
                         ensure!(record.version().is_one(), "Output record must be Version 1 on or after Consensus V8");
                     } else {
-                        ensure!(record.version().is_one(), "Output record must be Version 1 on or after Consensus V8");
+                        ensure!(record.version().is_one(), "Output record must be Version 1 before Consensus V8");
```
mikebenfield: I don't follow this change. Isn't this `else` branch in the "on or after Consensus V8" case? Or maybe you meant to change the `bail` message immediately prior.

mikebenfield (commented):
For what it's worth I tested this out via Leo and via my program and it does seem to work as intended in the cases I tried. 

Left a couple little comments. 

I imagine these messages likely aren't intended to be encountered by regular end users, but I'll mention that if they are they might seem pretty opaque.

synthesizer/process/src/verify_execution.rs
vicsn: Good catch 2df3dd7b7

synthesizer/process/src/stack/authorization/mod.rs
```diff
@@ -147,6 +148,73 @@ impl<N: Network> Authorization<N> {
             _ => false,
         }
     }
+
+    /// Checks whether the authorization is for a valid program edition.
+    pub fn check_valid_edition(&self, process: &crate::Process<N>, _consensus_version: ConsensusVersion) -> Result<()> {
```
raychu86, line 153: Nit: Why the underscore here? We usually use this naming pattern for unused variables

```suggestion
    pub fn check_valid_edition(&self, process: &crate::Process<N>, consensus_version: ConsensusVersion) -> Result<()> {
```

synthesizer/process/src/stack/authorization/mod.rs
```diff
@@ -147,6 +148,73 @@ impl<N: Network> Authorization<N> {
...
+            .first()
+            .map(|(_, t)| t.program_id())
+            .ok_or_else(|| anyhow!("No transitions found in the Authorization."))?;
+        // There is only one credits.aleo edition, so we can safely skip this case.
+        if root_program_id.to_string() != "credits.aleo" {
+            // Get the program's current edition.
+            let _program_edition = *process.get_stack(root_program_id)?.program_edition();
+            // If we're past ConsensusVersion::V8, ensure new stacks are not on edition 0.
+            #[cfg(not(any(test, feature = "test")))]
+            if _consensus_version >= ConsensusVersion::V8 && _program_edition == 0 {
```
raychu86: Please add documentation here so we are aware of this check when Upgradability will land.

Once upgradability lands, this check will be different. We can't just check the program edition anymore, it will need to be programs that are edition 0 with no constructor that can't be called.

synthesizer/process/src/stack/authorization/mod.rs
```diff
@@ -147,6 +148,73 @@ impl<N: Network> Authorization<N> {
...
+    }
+
+    /// Checks whether the authorization creates valid records.
+    pub fn check_valid_records(&self, consensus_version: ConsensusVersion) -> Result<()> {
+        let transitions = self.transitions.read();
+        // Collect the transition's records.
+        let records = transitions.values().map(|transition| {
+            let program_id = transition.program_id();
+            let function_name = transition.function_name();
+            let input_records = transition.outputs().iter().filter_map(|output| output.record());
```
raychu86: This line is incorrectly fetching the output records instead of the input ones.

synthesizer/process/src/stack/authorization/mod.rs
vicsn: Clippy doesn't recognize the variables are used past `#[cfg(not(any(test, feature = "test")))]`. 

A resolution would be to use the variables also under the opposite featureset. But that doesn't seem worth the trouble. It is awkward but harmless, and happens in a few places in the codebase.

synthesizer/process/src/stack/authorization/mod.rs
vicsn: 635bbf6bf

It'll undoubtedly come up during testing.

synthesizer/process/src/stack/authorization/mod.rs
vicsn: Ouch, thx: 17c058cd7

The network can't infer the version from the anonymous input record's serial numbers.

synthesizer/process/src/stack/authorization/mod.rs
```diff
@@ -147,6 +148,54 @@ impl<N: Network> Authorization<N> {
             _ => false,
         }
     }
+
+    /// Checks whether the authorization is for a valid program edition.
+    pub fn check_valid_edition(&self, process: &crate::Process<N>, _consensus_version: ConsensusVersion) -> Result<()> {
+        // Determine the root transition's program ID.
+        let transitions = self.transitions.read();
+        let root_program_id = *transitions
```
d0cd: Should we be checking more than the root program ID?
The corresponding check in `check_transaction` does it for all of the component transitions.

synthesizer/process/src/verify_execution.rs
d0cd: It could be worth commenting why we're allowing record version 1 before V8 in tests.
This confused me for a second.

vicsn: Yeah it looks a bit silly. At the same time, adding context on that it was a backwards incompatible feels verbose. Did a minor clarification... e7aa9599c

synthesizer/process/src/stack/authorization/mod.rs
vicsn: Thank you for catching. 589d0bbd2

raychu86 (approved):

PR #2803, opened 2025-07-11
Don't make invalid Scalars in Scalar::from_bits_le

### LINKED ISSUE #2801: [Bug] Deployment can fail depending on sampled values
## 🐛 Bug Report

When I try to deploy this program:

```
program test.aleo;

function address_cast:
    input r0 as address.private;
    cast r0 into r1 as scalar;
    output r1 as scalar.private;
```

Deployment fails with this message
```
Failed to eject scalar value: The scalar is greater than or equal to the modulus.
```

because `synthesize_key` has sampled an address that cannot be cast to a scalar. 

For a demo of this failure, see [this repo](https://github.com/mikebenfield/heights_demo), and run with `RUST_BACKTRACE=full cargo run --release`. (It's set up to use snarkvm 3.8.0 but it fails on `staging` also.)

mikebenfield:
Previously, if the bits represented a value
greater than or equal to the modulus, an invalid
Scalar would be constructed. This can be a problem for instance during key synthesis when deploying,
because `eject_value` will panic, leading to a
failed deployment.

Fixes #2801

### DIFF
--- a/circuit/account/src/signature/helpers/from_bits.rs
+++ b/circuit/account/src/signature/helpers/from_bits.rs
@@ -124,12 +124,12 @@ mod tests {
 
     #[test]
     fn test_from_bits_le_public() {
-        check_from_bits_le(Mode::Public, 9, 0, 1875, 1881);
+        check_from_bits_le(Mode::Public, 9, 0, 1877, 1883);
     }
 
     #[test]
     fn test_from_bits_le_private() {
-        check_from_bits_le(Mode::Private, 9, 0, 1875, 1881);
+        check_from_bits_le(Mode::Private, 9, 0, 1877, 1883);
     }
 
     #[test]
@@ -139,11 +139,11 @@ mod tests {
 
     #[test]
     fn test_from_bits_be_public() {
-        check_from_bits_be(Mode::Public, 9, 0, 1875, 1881);
+        check_from_bits_be(Mode::Public, 9, 0, 1877, 1883);
     }
 
     #[test]
     fn test_from_bits_be_private() {
-        check_from_bits_be(Mode::Private, 9, 0, 1875, 1881);
+        check_from_bits_be(Mode::Private, 9, 0, 1877, 1883);
     }
 }
--- a/circuit/program/src/data/literal/cast/field.rs
+++ b/circuit/program/src/data/literal/cast/field.rs
@@ -183,8 +183,8 @@ mod tests {
     #[test]
     fn test_field_to_scalar() {
         check_cast::<Scalar<Circuit>, console_root::types::Scalar<MainnetV0>>(Mode::Constant, count_is!(253, 0, 0, 0));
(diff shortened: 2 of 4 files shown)


vicsn:
Thank you for diving in!

One tricky part about this code is that under the hood, it generates a circuit (or R1CS). By changing this logic, we've now created a backwards incompatible change which invalidates any deployed function which used this functionality for syncing clients and users.

Moreover, I can imagine perhaps this issue also exists with some other opcodes...

I vaguely recall that in our main VM loop `execute_function`, we somehow disable `assert` aleo instructions, for the same reason. Randomly sampled inputs to the circuit will be unlikely to lead to satisfying inputs to the assert. However I can't find where we implemented this at the moment. Perhaps d0cd remembers when he's back.

mikebenfield:
(just force pushed a fixed up version; sorry but I figure for a 15 line change that nobody's looked carefully at it's OK)

Yeah I was afraid something like that would be the case. 

Regarding assertions, as far as I can tell though failing assertions will just lead to constraints failing, and that's OK for executions when just synthesizing keys. 

As for other types, I checked at least for from_bytes and I don't think we have this particular problem.

So what would you suggest to fix this - at the moment several Leo tests are failing due to this, and many programs I `leo deploy` fail to deploy. (I'm not sure why this is happening more frequently now than before - maybe the RNG I'm using is seeded differently for one reason or another.)

vicsn:
> (just force pushed a fixed up version; sorry but I figure for a 15 line change that nobody's looked carefully at it's OK)

Yes all good :)

> So what would you suggest to fix this - at the moment several Leo tests are failing due to this, and many programs I leo deploy fail to deploy. (I'm not sure why this is happening more frequently now than before - maybe the RNG I'm using is seeded differently for one reason or another.)

The "more frequently" is over how big of a period, roughly? But yes it could be different RNG state, especially if you re-use an RNG over multiple deployments or executions. A million times 1 is still a theoretically valid output of an RNG :^).

How often does casting as a scalar (from an address or other difficult types) happen on programs deployed to canary/testnet/mainnet? You can find a list of programs in the master or PR branches in the private `aleo-program-regressions` repo. If it's not a common thing, perhaps you can refactor the tests not to attempt it anymore.

mikebenfield:
> The "more frequently" is over how big of a period, roughly? 

Last week or two but now it happens very often. If I `leo deploy` a program like this

```
program test.aleo;

function address_cast:
    input r0 as address.private;
    cast r0 into r1 as scalar;
    output r1 as scalar.private;
 ```
 
 It'll likely happen. 
 
> How often does casting as a scalar (from an address or other difficult types)

I found two instances in that repo. 

> perhaps you can refactor the tests not to attempt it anymore.

and then just leave casting into a scalar permanently broken?

vicsn:
Closing as this will change circuit-logic, which will require a much bigger and careful refactor

PR #2804, opened 2025-07-11
Remove some misleading comments.

mikebenfield:
There are multiple comments scattered about
to the effect of it being safe to
reconstitute one type from another as there
are more bits available to represent the first.

Some of these comments were backwards, in that
we're creating the type with fewer bits.

### DIFF
--- a/circuit/program/src/data/literal/cast_lossy/scalar.rs
+++ b/circuit/program/src/data/literal/cast_lossy/scalar.rs
@@ -65,9 +65,6 @@ impl<E: Environment, I: IntegerType> CastLossy<Integer<E, I>> for Scalar<E> {
     /// Casts a `Scalar` to an `Integer`, with lossy truncation.
     #[inline]
     fn cast_lossy(&self) -> Integer<E, I> {
-        // Note: We are reconstituting the integer from the scalar field.
-        // This is safe as the number of bits in the integer is less than the scalar field modulus,
-        // and thus will always fit within a single scalar field element.
         debug_assert!(I::BITS < <console::Scalar<E::Network> as console::SizeInBits>::size_in_bits() as u64);
 
         // Truncate the field to the size of the integer domain.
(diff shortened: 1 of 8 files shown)


Antonio95 (approved):
Definitely, the comments seem to be the result of a mixup in the direction of the cast. Thanks a lot for the fixes.

1. I agree with the deletion of all the comments removed in the PR

2. I think I have found a couple more, do you agree @mikebenfield ?
 -  https://github.com/ProvableHQ/snarkVM/blob/cc11039db8681cedc64f7b3b9b355ee8ac41689a/circuit/types/scalar/src/helpers/from_bits.rs#L25-L27 (this is even more incorrect because one isn't casting from a field element, but from a list of bits (represented by the field elements 1 or 0); and it is enforced bits above the `Scalar::size_in_bits` are zero.
 - https://github.com/ProvableHQ/snarkVM/blob/cc11039db8681cedc64f7b3b9b355ee8ac41689a/circuit/types/scalar/src/helpers/to_bits.rs#L39-L41 I believe the comment is incorrect here too as, unless there is something I am missing, the representation of the scalar in bits is computed and then appended to a vector of bits; and the length of the latter doesn't seem to be bound by `BaseField::size_in_bits` (but I am less sure about this last claim)
 - There may be others in this vein, I haven't done a systematic search.

3. Large irrelevant but, while we are at it, would you mind ReplaceAll-ing "commonly-used" by "commonly used"? (it appears in the doc of many of the functions I've read while reviewing this PR)

mikebenfield:
> 2. I think I have found a couple more, do you agree @mikebenfield ?

Yes; here's another commit with those removed. 

> 3. Large irrelevant but, while we are at it, would you mind ReplaceAll-ing "commonly-used" by "commonly used"? (it appears in the doc of many of the functions I've read while reviewing this PR)

Sure; done in this new commit.

PR #2806, opened 2025-07-14
Introduce StaticQuery

meddle0x53:
## Motivation

After a sync with @vicsn we are moving to using `snarkos developer execute` instead of the cannon for testing. One of the steps is to pregenerate transactions using the genesis state root and on the latest consensus version.

This can be done by passing them as the `--query` parameter to the command. Here is an example:

```
snarkos developer execute \
  --query '{"state_root": "sr1287g435kqmlu6w8fzpp9dyggfwxyyaeymg89p3nzefq86xxst5pq9eht34", "height": 249}' \
  --private-key <PK> \
  credits.aleo transfer_public aleo1rhgdu77hgyqd3xjj8ucu3jj9r2krwz6mnzyd80gncr5fxcwlh5rsvzp9px 1000u64 \
  --network 1  \
  --store file.txt \
  --store-format string
```

This now works and these transaction can be sent later with:

```
curl http://localhost:3030/testnet/transaction/broadcast -X POST -H "Content-Type: application/json" -d @file.txt
```

We will be able to pre-generate many transfer_public (and not only) transactions that we can send later while running the tests.

As per the requested change here : https://github.com/ProvableHQ/snarkOS/pull/3723 am moving the StaticQuery part to this repository now.

## Test Plan

Wrote a small unit test for the new static query parsing logic. Ran all the CLI unit tests.

Can be tested through SnarkOS with curl too. (Will update the PR there to use the StaticQuery from this PR).

## Related PRs

[Wrote a small unit test for the new static query parsing logic. Ran all the CLI unit tests.](https://github.com/ProvableHQ/snarkOS/pull/3723)


ledger/query/src/static_query.rs
```diff
@@ -0,0 +1,109 @@
...
+    }
+}
+
+#[derive(Deserialize)]
+struct StaticQueryInput {
+    state_root: String,
+    height: u32,
+}
+
+impl<N: Network> TryFrom<&String> for StaticQuery<N> {
```
niklaslong: I'd implement [`FromStr`](https://doc.rust-lang.org/beta/std/str/trait.FromStr.html) here as it implies parsing more so than `TryFrom<&str>` ([see _the_ book](https://doc.rust-lang.org/nightly/book/ch04-03-slices.html#string-slices-as-parameters) on `&String` vs `&str`).

ledger/query/src/static_query.rs
```diff
@@ -0,0 +1,109 @@
...
+impl<N: Network> QueryTrait<N> for StaticQuery<N> {
+    /// Returns the current state root.
+    fn current_state_root(&self) -> Result<N::StateRoot> {
+        Ok(self.state_root)
+    }
+
+    /// Returns the current state root.
+    #[cfg(feature = "async")]
+    async fn current_state_root_async(&self) -> Result<N::StateRoot> {
+        unimplemented!();
```
niklaslong: VM currently doesn't use the asynchronous queries but it might be good to introduce a message alongside the panic if this changes in future?

ledger/query/src/static_query.rs
```diff
@@ -0,0 +1,109 @@
...
+
+use anyhow::{Result, anyhow};
+use serde::Deserialize;
+use std::{collections::HashMap, str::FromStr};
+
+#[derive(Clone)]
+pub struct StaticQuery<N: Network> {
+    block_height: u32,
+    state_root: N::StateRoot,
+    state_paths: HashMap<Field<N>, StatePath<N>>,
```
niklaslong, line 27: @vicsn just a note that we may need to support more expressive state root fetching in the `QueryTrait` in future (iirc). This because the same query object is currently used for both the execution and the fee transactions. If both proofs involve private state, we may need different state roots for each proof so they pass the `statepath.stateroot == stateroot` check.

niklaslong (commented):
LGTM overall, left a few suggestions 🚀

ledger/query/src/static_query.rs
vicsn: Is it possible for you to use this querytype from within delegated proving and refactor it as needed in the future? I assume that whatever the delegated prover needs, we'll need for testing too, so it makes sense to consolidate into snarkVM. We can make it a breaking change.

ledger/query/src/static_query.rs
meddle0x53: As it is a static query not like the other one, querying an API (so it can be async), I guess we can say something in the lines "Async calls are not supported by StaticQuery"?

niklaslong: yep, that seems fine 👍

ledger/query/src/static_query.rs
niklaslong: Yes, we have the same implementation at the moment 👍

niklaslong (approved):

vicsn (approved):

PR #2807, opened 2025-07-14
[Feature] Complete implementation of program upgradability.

d0cd:
This PR contains the complete implementation of program upgradability.
It subsumes the following PRs:
- https://github.com/ProvableHQ/snarkVM/pull/2653
- https://github.com/ProvableHQ/snarkVM/pull/2655
- https://github.com/ProvableHQ/snarkVM/pull/2717
- https://github.com/ProvableHQ/snarkVM/pull/2722
- https://github.com/ProvableHQ/snarkVM/pull/2758


console/network/src/canary_v0.rs
```diff
@@ -149,31 +149,11 @@ impl Network for CanaryV0 {
...
-        (ConsensusVersion::V1, 0),
-        (ConsensusVersion::V2, 2_900_000),
-        (ConsensusVersion::V3, 4_560_000),
-        (ConsensusVersion::V4, 5_730_000),
-        (ConsensusVersion::V5, 5_780_000),
-        (ConsensusVersion::V6, 6_240_000),
-        (ConsensusVersion::V7, 6_880_000),
-        (ConsensusVersion::V8, 7_565_000),
-    ];
+    const CONSENSUS_VERSION_HEIGHTS: [(ConsensusVersion, u32); 9] = CANARY_V0_CONSENSUS_VERSION_HEIGHTS;
```
vicsn: Reminder: we'll need to resolve after https://github.com/ProvableHQ/snarkVM/pull/2799 is merged, and we'll need to update preliminary new consensus version heights

We can calculate the heights from Thursday onwards, Canary nodes need to update to a stable snarkOS version first.

synthesizer/src/vm/verify.rs
```diff
@@ -90,8 +90,7 @@ impl<N: Network, C: ConsensusStorage<N>> VM<N, C> {
         }
 
         // Compute the Merkle root of the transaction.
-        // Debug-mode only, as the `Transaction` constructor recomputes the transaction ID at initialization.
-        #[cfg(debug_assertions)]
+        // Attention - This check is mandatory. This is the only way to ensure that the transaction ID is well-formed.
```
vicsn: Can we bring back `#[cfg(debug_assertions)]`

vicsn (changes requested):

ledger/block/src/transaction/deployment/mod.rs
```diff
@@ -22,31 +22,51 @@ mod string;
...
+    /// After the migration height where this feature is enabled, the checksum will be required.
+    program_checksum: Option<[U8<N>; 32]>,
+    /// An optional owner for the program.
+    /// This field creates a backwards-compatible implicit versioning mechanism for deployments.
+    /// Before the migration height where this feature is enabled, the owner will **not** be allowed.
+    /// After the migration height where this feature is enabled, the owner will be required.
+    program_owner: Option<Address<N>>,
 }
 
+impl<N: Network> PartialEq for Deployment<N> {
```
d0cd, line 51: Are there any concerns if we do not equate the `program_checksum` or `program_owner`?

ledger/block/src/transaction/deployment/mod.rs
```diff
@@ -181,6 +222,45 @@ impl<N: Network> Deployment<N> {
...
+    /// Note: This method is intended to be used by the synthesizer **only**, and should not be called by the user.
+    #[doc(hidden)]
+    pub fn set_program_owner_raw(&mut self, program_owner: Option<Address<N>>) {
+        self.program_owner = program_owner;
+    }
+
+    /// An internal function to return the implicit deployment version.
+    /// This function implicitly checks that the deployment checksum and owner is well-formed.
+    #[doc(hidden)]
+    pub(super) fn version(&self) -> Result<DeploymentVersion> {
```
d0cd, line 231: Can this be made public?

synthesizer/process/src/finalize.rs
```diff
@@ -86,8 +91,17 @@ impl<N: Network> Process<N> {
                 // Initialize the mapping.
                 finalize_operations.push(store.initialize_mapping(*program_id, *mapping.name())?);
             }
-            finish!(timer, "Initialize the program mappings");
+            lap!(timer, "Initialize the program mappings");
+
+            // If the program has a constructor, execute it and extend the finalize operations.
+            // This must happen after the mappings are initialized as the constructor may depend on them.
+            if deployment.program().contains_constructor() {
+                let operations = finalize_constructor(state, store, &stack, N::TransitionID::default())?;
```
d0cd, line 99: Verify that it is okay to use `N::TransitionID::default()`. Alternatively, we could use the fee transition ID.

vm/file/manifest.rs
```diff
@@ -49,7 +53,8 @@ impl<N: Network> Manifest<N> {
     "program": "{id}",
     "version": "0.0.0",
     "description": "",
-    "license": "MIT"
+    "license": "MIT",
```
d0cd, line 56: @reviewers
This is the way developers using `Package` can specify editions and issue upgrades.
They will have to manually set the appropriate editions.
If no editions are provided, then we default to 0.
Would love to hear your thoughts on this form factor.

synthesizer/process/src/tests/test_credits.rs
d0cd, line 1: Should we add a test with a constructor here?

synthesizer/process/src/stack/helpers/initialize.rs
```diff
@@ -34,96 +34,33 @@ impl<N: Network> Stack<N> {
...
+        let stack = Self::create_raw(process, program, edition)?;
+        // Initialize and check the stack's validity.
+        stack.check_and_initialize(process)?;
+        // Return the stack.
+        Ok(stack)
+    }
+
+    /// Create a new stack, given the process and program, without completely initializing or checking for validity.
+    #[inline]
+    pub(crate) fn create_raw(process: &Process<N>, program: &Program<N>, edition: u16) -> Result<Self> {
```
d0cd, line 47: @reviewers
This new API allows `Process`es to be more ergonomically initialized for programs with different editions.
Would love to get your thoughts on this design.

console/algorithms/benches/bhp.rs
```diff
@@ -66,10 +66,30 @@ fn bhp1024(c: &mut Criterion) {
     c.bench_function(&format!("BHP1024 Hash - input size {}", input.len()), |b| b.iter(|| hash.hash(&input)));
 }
 
+fn bhp1024_large(c: &mut Criterion) {
+    const SIZE_IN_BYTES: [usize; 4] =
+        [1_000 /* 1 kB */, 10_000 /* 10 KB */, 100_000 /* 100 KB */, 1_000_000 /* 1 MiB */];
```
ljedrz:
```suggestion
        [1_000 /* 1 kB */, 10_000 /* 10 kB */, 100_000 /* 100 kB */, 1_000_000 /* 1 MB */];
```
nit

console/network/src/consensus_heights.rs
```diff
@@ -0,0 +1,68 @@
...
+
+/// The consensus version height for `CanaryV0`.
+pub const CANARY_V0_CONSENSUS_VERSION_HEIGHTS: [(ConsensusVersion, u32); 9] = [
+    (ConsensusVersion::V1, 0),
+    (ConsensusVersion::V2, 2_900_000),
+    (ConsensusVersion::V3, 4_560_000),
+    (ConsensusVersion::V4, 5_730_000),
+    (ConsensusVersion::V5, 5_780_000),
+    (ConsensusVersion::V6, 6_240_000),
+    (ConsensusVersion::V7, 6_895_000),
```
ljedrz: didn't this value use to be `6_880_000`?

ledger/block/src/transaction/deployment/bytes.rs
```diff
@@ -70,6 +102,16 @@ impl<N: Network> ToBytes for Deployment<N> {
             // Write the certificate.
             certificate.write_le(&mut writer)?;
         }
+        // If the deployment version is 2, write the program checksum and program owner.
+        // Note: The unwraps are safe because `Deployment::version` only returns `V2` if both the checksum and owner is present.
+        if version == DeploymentVersion::V2 {
+            // Write the bytes of the checksum.
+            for byte in &self.program_checksum.unwrap() {
+                byte.write_le(&mut writer)?;
+            }
```
ljedrz, line 111: wouldn't it be preferable to use `self.program_checksum.unwrap().write_le(&mut writer)?;` here instead? or do we deliberately want to omit the length of the checksum?

ledger/block/src/transaction/deployment/mod.rs
```diff
@@ -57,12 +77,28 @@ impl<N: Network> Deployment<N> {
...
+                ensure!(
+                    program_checksum == self.program.to_checksum(),
+                    "The program checksum in the deployment does not match the computed checksum for '{program_id}'"
+                );
+            }
+            // If the program checksum is absent, then verify that the edition is zero or one.
+            // It must be the case that this program was deployed before upgradability was introduced.
+            None => {
+                ensure!(
+                    self.edition == 0 || self.edition == 1,
```
ljedrz:
```suggestion
                    self.edition <= 1,
```
nit

ledger/block/src/transaction/deployment/mod.rs
```diff
@@ -131,6 +162,16 @@ impl<N: Network> Deployment<N> {
...
     }
 
+    /// Returns the program checksum, if it was stored.
+    pub const fn program_checksum(&self) -> Option<&[U8<N>; 32]> {
+        self.program_checksum.as_ref()
+    }
+
+    /// Returns the program owner, if it was stored.
+    pub const fn program_owner(&self) -> Option<&Address<N>> {
+        self.program_owner.as_ref()
```
ljedrz:
```suggestion
    pub const fn program_owner(&self) -> Option<Address<N>> {
        self.program_owner
```
IIRC the `Address` is `Copy`

ljedrz: the same might apply to the checksum

edit: actually no, as `U8` is not `Copy`

ledger/block/src/transaction/merkle.rs
```diff
@@ -191,20 +188,28 @@ impl<N: Network> Transaction<N> {
...
         ensure!(
-            functions.len() == verifying_keys.len(),
-            "Number of functions ('{}') and verifying keys ('{}') do not match",
-            functions.len(),
+            num_functions == verifying_keys.len(),
+            "Number of functions ('{num_functions}') and verifying keys ('{}') do not match",
             verifying_keys.len()
         );
+        // Ensure there are functions.
+        ensure!(num_functions > 0, "Deployment must contain at least one function");
```
ljedrz:
```suggestion
        ensure!(num_functions != 0, "Deployment must contain at least one function");
```
`clippy` likes to complain about such comparisons, even if it doesn't do so _yet_

synthesizer/process/src/finalize.rs
```diff
@@ -201,6 +215,101 @@ fn finalize_fee_transition<N: Network, P: FinalizeStorage<N>>(
...
+            _ => {
+                let result = try_vm_runtime!(|| command.finalize(stack, store, &mut registers));
+                match result {
+                    // If the evaluation succeeds with an operation, add it to the list.
+                    Ok(Ok(Some(finalize_operation))) => finalize_operations.push(finalize_operation),
+                    // If the evaluation succeeds with no operation, continue.
+                    Ok(Ok(None)) => {}
+                    // If the evaluation fails, bail and return the error.
+                    Ok(Err(error)) => {
+                        println!("'constructor' failed to evaluate command ({command}): {error}");
```
ljedrz: wouldn't this result in a duplicate log? the same applies to the `Err(_)` branch

synthesizer/process/src/stack/deploy.rs
```diff
@@ -51,7 +51,13 @@ impl<N: Network> Stack<N> {
         finish!(timer);
 
         // Return the deployment.
-        Deployment::new(*self.program_edition(), self.program.clone(), verifying_keys)
+        Deployment::new(
+            *self.program_edition,
+            self.program.clone(),
+            verifying_keys,
+            Some(self.program_checksum),
+            Some(Address::zero()),
```
ljedrz: just making sure this "placeholder-looking" value was intentional

synthesizer/process/src/stack/helpers/check_upgrade.rs
```diff
@@ -0,0 +1,126 @@
...
+        // Ensure the program is not `credits.aleo`.
+        ensure!(program_id != &ProgramID::from_str("credits.aleo")?, "Cannot upgrade 'credits.aleo'");
+        // Ensure the program ID matches.
+        ensure!(old_program.id() == new_program.id(), "Cannot upgrade '{program_id}' with different program ID");
+        // Ensure that all of the imports in the old program exist in the new program.
+        for old_import in old_program.imports().keys() {
+            if !new_program.contains_import(old_import) {
+                bail!("Cannot upgrade '{program_id}' because it is missing the original import '{old_import}'");
+            }
+        }
```
ljedrz, line 52: this implies that the order of the original imports may change; we might want to mention this in the overview, and it applies to all items checked in that way, as opposed to e.g. using `zip` to compare them in an ordered fashion

synthesizer/process/src/stack/helpers/synthesize.rs
```diff
@@ -32,6 +32,11 @@ impl<N: Network> Stack<N> {
         let program_id = self.program_id();
         // Retrieve the function input types.
         let input_types = self.get_function(function_name)?.input_types();
+        // Retrieve the program checksum, if the program has a constructor.
+        let program_checksum = match self.program().contains_constructor() {
+            true => Some(self.program_checksum_as_field()?),
+            false => None,
+        };
```
ljedrz, line 39: I'm pretty sure I've seen this a few times by now, perhaps it could be turned into a method?

synthesizer/process/src/stack/mod.rs
```diff
@@ -215,54 +224,164 @@ pub struct Stack<N: Network> {
...
+            // Ensure the program imports all exist in the process already.
+            if !process.contains_program(import) {
+                bail!("Cannot add program '{}' because its import '{import}' must be added first", self.program.id())
+            }
+        }
+
+        // Add the constructor to the stack if it exists.
+        if let Some(constructor) = self.program.constructor() {
+            // Ensure that the constructor is not already added.
+            ensure!(self.constructor_types.read().is_none(), "Constructor already exists");
```
ljedrz: this _should_ be redundant due to line 276; otherwise, that write guard should be kept for longer

synthesizer/process/src/stack/mod.rs
```diff
@@ -215,54 +224,164 @@ pub struct Stack<N: Network> {
...
+        for closure in self.program.closures().values() {
+            // Retrieve the closure name.
+            let name = closure.name();
+            // Ensure the closure name is not already added.
+            ensure!(!self.register_types.read().contains_key(name), "Closure '{name}' already exists");
+
+            // Compute the register types.
+            let register_types = RegisterTypes::from_closure(self, closure)?;
+            // Add the closure name and register types to the stack.
+            self.register_types.write().insert(*name, register_types);
```
ljedrz: it would be faster to obtain a write guard before the loop, and to keep it live for its entire duration; this recommendation applies to all other spots like this where the collection is not expected to be read/modified from elsewhere in the meantime (which is most likely here, as we're just initializing it)

synthesizer/process/src/stack/mod.rs
```diff
@@ -309,3 +428,21 @@ impl<N: Network> Deref for StackRef<'_, N> {
...
+
+// A helper function to check that a program is well-formed.
+fn check_program_is_well_formed<N: Network>(program: &Program<N>) -> Result<()> {
+    // Ensure the program contains functions.
+    ensure!(!program.functions().is_empty(), "No functions present in the deployment for program '{}'", program.id());
+
+    // Serialize the program into bytes.
+    let program_bytes = program.to_bytes_le()?;
+    // Ensure the program deserializes from bytes correctly.
+    ensure!(program == &Program::from_bytes_le(&program_bytes)?, "Program byte serialization failed");
```
ljedrz, line 451: do we need this outside of debug/test scenarios?

ljedrz: the same applies to the string conversion below

synthesizer/program/src/finalize/mod.rs
```diff
@@ -115,8 +115,16 @@ impl<N: Network> FinalizeCore<N> {
         // Ensure the maximum number of commands has not been exceeded.
         ensure!(self.commands.len() < N::MAX_COMMANDS, "Cannot add more than {} commands", N::MAX_COMMANDS);
         // Ensure the number of write commands has not been exceeded.
-        ensure!(self.num_writes < N::MAX_WRITES, "Cannot add more than {} 'set' & 'remove' commands", N::MAX_WRITES);
+        if command.is_write() {
+            ensure!(
+                self.num_writes < N::MAX_WRITES,
```
ljedrz, line 120: shouldn't this be `<=`?

ljedrz: ah, no, the context is the addition of a new write :+1:

synthesizer/program/src/logic/command/contains.rs
```diff
@@ -26,8 +26,8 @@ use console::{
 pub struct Contains<N: Network> {
     /// The mapping name.
     mapping: CallOperator<N>,
-    /// The key to access the mapping.
-    key: Operand<N>,
+    /// The operands.
+    operands: [Operand<N>; 1],
```
ljedrz, line 30: why use a single-elem array instead of just the elem here? I'm guessing for "compatibility" with similar objects that have one or more operands, to always return a slice when accessing them?

synthesizer/program/src/logic/command/get.rs
```diff
@@ -35,7 +35,7 @@ impl<N: Network> PartialEq for Get<N> {
     /// Returns true if the two objects are equal.
     #[inline]
     fn eq(&self, other: &Self) -> bool {
-        self.mapping == other.mapping && self.key == other.key && self.destination == other.destination
+        self.mapping == other.mapping && self.key() == other.key() && self.destination == other.destination
```
ljedrz: nit: this is just `self == other` if both derived the equality traits

synthesizer/program/src/logic/command/get.rs
```diff
@@ -46,7 +46,7 @@ impl<N: Network> std::hash::Hash for Get<N> {
     #[inline]
     fn hash<H: std::hash::Hasher>(&self, state: &mut H) {
         self.mapping.hash(state);
-        self.key.hash(state);
+        self.key().hash(state);
         self.destination.hash(state);
     }
 }
```
ljedrz: this is just a handwritten `#[derive(Hash)]`, just without the slice (only its single element) in `self.key`

ledger/block/src/transaction/deployment/bytes.rs
d0cd: Yeah, we're deliberately omitting the length of the checksum since we know that its 32.

synthesizer/program/src/logic/command/contains.rs
ljedrz: yep, confirmed; also to avoid some allocations

synthesizer/program/src/constructor/parse.rs
```diff
@@ -0,0 +1,141 @@
...
+        let (string, _) = tag(":")(string)?;
+
+        // Parse the commands from the string.
+        let (string, commands) = many1(Command::parse)(string)?;
+
+        map_res(take(0usize), move |_| {
+            // Initialize a new constructor.
+            let mut constructor = Self { commands: Default::default(), num_writes: 0, positions: Default::default() };
+            if let Err(error) = commands.iter().cloned().try_for_each(|command| constructor.add_command(command)) {
+                eprintln!("{error}");
```
ljedrz, line 38: nit: potential for duplicate logs

synthesizer/process/src/stack/deploy.rs
d0cd: That's right, left a clarifying comment.

synthesizer/src/vm/verify.rs
```diff
@@ -137,40 +137,43 @@ impl<N: Network, C: ConsensusStorage<N>> VM<N, C> {
...
-                    if program_id != ProgramID::from_str("credits.aleo")? && stack.program_edition().is_zero() {
-                        drop(process); // Drop the process lock. 
-                        bail!(
-                            "Invalid execution transaction '{id}' - the program edition for '{program_id}' cannot be zero for `ConsensusVersion::V8` or greater. Please redeploy the program."
-                        );
-                    }
-                }
+        // Get the program editions from the transaction.
+
+        let mut program_editions = Vec::with_capacity(Transaction::<N>::MAX_TRANSITIONS);
```
ljedrz, line 145: this capacity would currently be `32`, right? that should be fine, but if we foresee it growing a lot more in the future, this could result in overallocating

synthesizer/src/vm/verify.rs
```diff
@@ -217,6 +266,14 @@ impl<N: Network, C: ConsensusStorage<N>> VM<N, C> {
                         ensure!(!is_program_in_storage, "Program ID '{}' is already deployed", deployment.program_id());
                         // Ensure the program does not already exist in the process.
                         ensure!(!is_program_in_process, "Program ID '{}' already exists", deployment.program_id());
+                        // Ensure that the program contains a constructor if the program is deployed after `ConsensusVersion::V9`.
+                        if consensus_version >= ConsensusVersion::V9 {
+                            // Check that the program contains a constructor.
```
ljedrz, line 270: here it is said that constructors belong to `V9` and above, but in line 158 it is expected from `V8` onward, or am I missing something :thinking:

synthesizer/process/src/stack/helpers/synthesize.rs
d0cd: It felt awkward having a method in `Program` to conditionally provide the checksum.
The one-liner also looked harder to read
```
let program_checksum = self.program.constructor().is_some().then(|| self.program_checksum_as_field()).transpose()?;
```
Curious if you see any other alternatives.

ljedrz:
note: `VM::check_transaction` has become _very_ long - we might want to consider splitting it into sub-checks

synthesizer/process/src/stack/helpers/synthesize.rs
ljedrz: I'm not particularly opinionated about this one, as it's still short; if the others don't mind it, it's no big deal

synthesizer/process/src/stack/mod.rs
d0cd: Good point, I opted to hold the write locks for the duration of the initialization.

ljedrz (commented):
LGTM engineering-wise :+1:

synthesizer/process/src/stack/mod.rs
d0cd: It's more of a sanity check.
What's nice is that if there was a bug in our (de)serializers, then the deployment would fail instead of passing and allowing this bug to come up later.

synthesizer/program/src/logic/command/contains.rs
d0cd: That's right, it was so that I can uniformly get the operands of the commands without cloning them.

synthesizer/src/vm/verify.rs
d0cd: That's right, also unlikely to grow by too much.
Even a doubling would be a lot.

synthesizer/src/vm/verify.rs
d0cd: The check on L158 was originally introduced in V8 and we also want it to apply for programs that contain constructors.

synthesizer/program/src/logic/command/get.rs
d0cd: Opted to derive.

synthesizer/program/src/logic/command/get.rs
d0cd: Opted to derive.

ledger/block/src/transaction/deployment/mod.rs
d0cd: I believe `console::U8` is copy.

ljedrz: ah, I may have checked the circuit variant

vm/file/manifest.rs
vicsn: I have never seen anything use `Package`, where is this used at the moment?

synthesizer/process/src/stack/helpers/initialize.rs
vicsn: LGTM!

synthesizer/process/src/stack/mod.rs
```diff
@@ -203,10 +210,12 @@ pub struct Stack<N: Network> {
     program: Program<N>,
     /// A reference to the global stack map.
     stacks: Weak<RwLock<IndexMap<ProgramID<N>, Arc<Stack<N>>>>>,
+    /// The register types for the program constructor, if it exists.
+    constructor_types: Arc<RwLock<Option<FinalizeTypes<N>>>>,
```
vicsn, line 214: Not super important, but can you explain to me why locks are needed?

ljedrz: I'm assuming it's because those values can't change between editions, so they don't need to differ between them in the stacks

d0cd: They're being used for multi-threaded interior mutability. So that we can defer initialization of `Stacks`s. This is useful in helping us initialize a set of stacks with cyclic imports (see `Process::add_programs_with_editions`)

d0cd: I specifically went with `RwLock` instead of `RefCell` to be consistent with the proving and verifying keys.

synthesizer/program/src/constructor/parse.rs
d0cd: Nom may not bubble errors up correctly.

vm/file/manifest.rs
d0cd: snarkOS developer

ledger/block/src/transaction/deployment/mod.rs
```diff
@@ -57,12 +77,28 @@ impl<N: Network> Deployment<N> {
...
-        ensure!(
-            self.edition == 0 || self.edition == 1,
-            "Deployed the wrong edition (expected 0 or 1, found '{}').",
-            self.edition
-        );
+        // Ensure that either the both the program checksum and owner are present, or both are absent.
+        // The call to `Deployment::version` implicitly performs this check.
+        self.version()?;
+        // Validate the deployment based on the program checksum.
+        match self.program_checksum {
```
raychu86: Do we need to check that if the `checksum` exists, that the `constructor` must also exist?

ledger/block/src/transaction/deployment/mod.rs
raychu86: Not particularly, as we don't do much Deployment equality checks. 

But why not include them in this?

ledger/block/src/transaction/deployment/mod.rs
raychu86: Can be public if there's a need to use this version.

Also, why are the `set_*_raw` functions `pub` instead of `pub(super)` if we want to restrict usage?

synthesizer/process/src/finalize.rs
raychu86: What is the transition_id used for outside of RandChaCha preimage?

ledger/block/src/transaction/deployment/mod.rs
d0cd: We don't enforce that since after V9, an upgrade for a program deployed during V8 may need to be made.
This also leaves room for additional migrations for non-V9 programs in the future.

synthesizer/src/vm/verify.rs
```diff
@@ -185,30 +188,75 @@ impl<N: Network, C: ConsensusStorage<N>> VM<N, C> {
...
+                        "The program owner in the deployment did not match the owner in the transaction\n('[{}]' != '[{}]')",
+                        given_owner,
+                        owner.address()
+                    );
+                }
+
                 // If the edition is zero, then check that:
                 //  - The program does not exist in the store or process.
-                // If the edition is one, then check that:
+                //  - The program contains a constructor.
```
d0cd, line 249:
```suggestion
                //  - The program contains a constructor after `ConsensusVersion::V9`.
```

raychu86 (approved):

raychu86:
Waiting on 1 change for Program Deployment defaults for checksum and owner

vicsn (approved):
Amazing work everyone, and especially @d0cd . I will update the consensus heights in a follow-up PR.

PR #2808, opened 2025-07-14
Remove std feature flag from snarkvm-utilities.

mikebenfield:
As mentioned in #2783, snarkvm-utilities was previously not building without this feature flag, and all the no-std parts were removed there. Since then this feature flag does nothing, so we now remove it.


kaimast (approved):
Thanks!

vicsn:
CC @trevor-crypto

PR #2810, opened 2025-07-15
[CI] Add a locktick check to the build script

ljedrz:
A pretty straightforward impl that doesn't involve heavy parsing. If this approach is fine, the snarkOS counterpart PR is going to look roughly the same.

Cc https://github.com/ProvableHQ/snarkOS/issues/3717.


vicsn (approved):
cool

PR #2811, opened 2025-07-15
 Postrelease merge of Testnet back into staging

vicsn:
## Motivation

^

### DIFF
--- a/console/program/src/request/sign.rs
+++ b/console/program/src/request/sign.rs
@@ -174,13 +174,6 @@ impl<N: Network> Request<N> {
                     };
                     // Ensure the record belongs to the signer.
                     ensure!(**record.owner() == signer, "Input record for '{program_id}' must belong to the signer");
-                    // Ensure version 0 credits.aleo Records are only allowed when calling upgrade().
-                    if program_id == ProgramID::from_str("credits.aleo")?
-                        && record.version() == &U8::zero()
-                        && function_name != Identifier::from_str("upgrade")?
-                    {
-                        bail!("Version 0 credits.aleo Records are only allowed when calling 'upgrade()'");
-                    }
                     // Compute the record view key.
                     let record_view_key = (*record.nonce() * *view_key).to_x_coordinate();
                     // Compute the record commitment.
--- a/synthesizer/process/src/stack/authorization/mod.rs
+++ b/synthesizer/process/src/stack/authorization/mod.rs
@@ -152,13 +152,11 @@ impl<N: Network> Authorization<N> {
     /// Checks whether the authorization is for a valid program edition.
     pub fn check_valid_edition(&self, process: &crate::Process<N>, consensus_version: ConsensusVersion) -> Result<()> {
         // Determine the root transition's program ID.
-        let root_program_id = {
-            let transitions = self.transitions.read();
-            *transitions
-                .first()
-                .map(|(_, t)| t.program_id())
-                .ok_or_else(|| anyhow!("No transitions found in the Authorization."))?
-        };
+        let transitions = self.transitions.read();
+        let root_program_id = *transitions
+            .first()
+            .map(|(_, t)| t.program_id())
+            .ok_or_else(|| anyhow!("No transitions found in the Authorization."))?;
         // There is only one credits.aleo edition, so we can safely skip this case.
         if root_program_id.to_string() != "credits.aleo" {
             // Get the program's current edition.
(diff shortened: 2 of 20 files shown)


console/program/src/request/sign.rs
```diff
@@ -174,13 +174,6 @@ impl<N: Network> Request<N> {
                     };
                     // Ensure the record belongs to the signer.
                     ensure!(**record.owner() == signer, "Input record for '{program_id}' must belong to the signer");
-                    // Ensure version 0 credits.aleo Records are only allowed when calling upgrade().
```
kaimast, line 177: I'm unfamiliar with this part of the codebase. Why was it safe to remove these checks?

synthesizer/process/src/traits/mod.rs
```diff
@@ -41,10 +41,12 @@ pub trait StackEvaluate<N: Network>: Clone {
     ///
     /// # Errors
     /// This method will halt if the given inputs are not the same length as the input statements.
-    fn evaluate_function<A: circuit::Aleo<Network = N>>(
+    fn evaluate_function<A: circuit::Aleo<Network = N>, R: CryptoRng + Rng>(
         &self,
         call_stack: CallStack<N>,
         caller: Option<ProgramID<N>>,
+        root_tvk: Option<Field<N>>,
```
kaimast, line 48: what does "tvk" mean? Is it a view key?

synthesizer/process/src/traits/mod.rs
```diff
@@ -41,10 +41,12 @@ pub trait StackEvaluate<N: Network>: Clone {
     ///
     /// # Errors
     /// This method will halt if the given inputs are not the same length as the input statements.
-    fn evaluate_function<A: circuit::Aleo<Network = N>>(
+    fn evaluate_function<A: circuit::Aleo<Network = N>, R: CryptoRng + Rng>(
```
kaimast, line 44: It might be good to document the arguments for this function. It has multiple optional arguments, and it is not clear to me when which one is needed.

synthesizer/process/src/traits/mod.rs
```diff
@@ -89,4 +91,7 @@ pub trait StackProgramTypes<N: Network> {
 pub trait RegistersCall<N: Network> {
     /// Returns the current call stack.
     fn call_stack(&self) -> CallStack<N>;
+
+    /// Returns a reference to the current call stack.
+    fn call_stack_ref(&self) -> &CallStack<N>;
```
kaimast, line 96: I assume this was due to backwards compatibility, but it would make more sense if call_stack() returns a reference and then the caller can clone the stack if needed, instead of having two functions.

synthesizer/process/src/traits/mod.rs
kaimast: oh, i see. "transition view key"

synthesizer/tests/utilities/tests/program_test.rs
```diff
@@ -68,6 +68,11 @@ impl ProgramTest {
     pub fn keys(&self) -> &[PrivateKey<CurrentNetwork>] {
         &self.keys
     }
+
+    /// Returns the path to the expectation file.
+    pub fn path(&self) -> &PathBuf {
```
kaimast, line 73: It is slightly nicer to return `&Path`. Returning `&PathBuf` is like returning `&String` (instead of `&str`).

To do that, you replace `&self.path` with `self.path.to_path()`.

synthesizer/process/src/stack/authorization/mod.rs
```diff
@@ -147,6 +148,53 @@ impl<N: Network> Authorization<N> {
             _ => false,
         }
     }
+
+    /// Checks whether the authorization is for a valid program edition.
+    pub fn check_valid_edition(&self, process: &crate::Process<N>, _consensus_version: ConsensusVersion) -> Result<()> {
```
kaimast, line 153: nit: change this to `consensus_version` and have something like this in the function body instead 
```rust
#[cfg(any(test, feature="test"))]
let _ = consensus_version;
```

kaimast (approved):
This all looks good to me. I left a few nitpicky comments, which we can either address here or maybe in a follow-up PR.

I am still not super familiar with `synthesizer` code, so another person should take a look at it.

console/program/src/request/sign.rs
vicsn: They were very recently added and we're moving them into `execute`, later in the transaction creation lifecycle: https://github.com/ProvableHQ/snarkVM/pull/2800

synthesizer/process/src/traits/mod.rs
vicsn: Indeed it would. Ideally we never even clone the CallStack at all. In need of some refactoring one day

synthesizer/process/src/stack/authorization/mod.rs
vicsn: Oe neat trick, will remember

vicsn:
Superseded by https://github.com/ProvableHQ/snarkVM/pull/2815 to resolve merge conflicts

PR #2813, opened 2025-07-16
[Feature] Improve Error Handling

kaimast:
## Motivation
Improve error handling in `snarkVM` to
* clearly differentiate user and I/O errors from internal errors that are bugs
* format `anyhow` error chains to show as much information as possible
* make testing and debugging easier (this PR serves a basis for #2789)
* provide helper functions for better error handling in snarkOS (see [snarkOS #3826](https://github.com/ProvableHQ/snarkOS/pull/3826))

### Top-level panic handling
This PR now install a panic handler at `vm/cli/main.rs` and shows a "pretty" output when encountering panics. See the example below.
```
$> snarkvm execute foo

⚠️ snarkVM encountered an unexpected error at vm/cli/commands/execute.rs:49:20:
attempt to divide by zero

This is most likely a bug!
Please report it to the snarkVM developers: https://github.com/ProvableHQ/snarkVM/issues/new?template=bug.md
```

### Top-level handling for user errors
`snarkvm` will now print the entire chain of anyhow errors. For example, when passing invalid arguments to `snarkvm run` the following output is generated.

```
$> snarkvm run main 0u32 0u16
⚠️ Running leo_test.aleo:main with inputs [0u32, 0u16] failed
     ↳ 'u32' is invalid: expected u32, found 0u16

Use `--help` for instructions on how to use this command
```

## Other changes
* A `ensure_equals` macro that compares two values and `bail`s they do not match. This is used in the block validation functions to print the actual values when something goes wrong.
* Improvements to `ledger_test_helpers`
* A utility function to log an `anyhow::Error`. It "flattens" the error by joining the messages with an em dash. This is similar to how snarkOS currently handles such errors
* Avoid printing an empty line when there are no updates for snarkVM
* Minor fixes to the README

## Notes
* REST/Query changes were moved to #2901.
* As @ljedrz pointed out, we cannot use `panic = "abort"`, which was the case in an earlier version of the PR, due to how program errors are handled in snarkVM. Instead, the CLI uses a nested try_catch.
* This change just adds a panic handler as another layer of checks, and does not change existing errors in the code to panics. We would still need to discuss whether to allow panics in production code.

### DIFF
--- a/ledger/block/src/genesis.rs
+++ b/ledger/block/src/genesis.rs
@@ -20,25 +20,32 @@ impl<N: Network> Block<N> {
     pub const NUM_GENESIS_TRANSACTIONS: usize = 4;
 
     /// Returns `true` if the block is a genesis block.
-    pub fn is_genesis(&self) -> bool {
-        // Ensure the previous block hash is zero.
-        self.previous_hash == N::BlockHash::default()
-            // Ensure the header is a genesis block header.
-            && self.header.is_genesis()
-            // Ensure the genesis authority is a beacon.
-            && self.authority.is_beacon()
-            // Ensure there is the correct number of ratification operations in the genesis block.
-            && self.ratifications.len() == 1
-            // Ensure there are no solutions in the genesis block.
-            && self.solutions.is_empty()
-            // Ensure there is the correct number of accepted transaction in the genesis block.
-            && self.transactions.num_accepted() == Self::NUM_GENESIS_TRANSACTIONS
-            // Ensure there is the correct number of rejected transaction in the genesis block.
-            && self.transactions.num_rejected() == 0
-            // Ensure there is the correct number of finalize operations in the genesis block.
-            && self.transactions.num_finalize() == 2 * Self::NUM_GENESIS_TRANSACTIONS
-            // Ensure there are no aborted transaction IDs in the genesis block.
-            && self.aborted_transaction_ids.is_empty()
+    pub fn is_genesis(&self) -> Result<bool> {
+        if !self.header.is_genesis()? {
+            return Ok(false);
+        }
+
+        ensure!(self.previous_hash == N::BlockHash::default(), "Invalid previous hash");
+        ensure!(self.authority.is_beacon(), "Invalid block authority");
+        ensure!(self.solutions.is_empty(), "Invalid solutins");
+        ensure!(self.transactions.num_rejected() == 0, "Invalid number of rejected transactions");
+        ensure!(self.aborted_transaction_ids.is_empty(), "Genesis block must not contain aborted transactions");
+
+        // Perform additional checks in production
+        #[cfg(not(any(test, feature = "test")))]
+        {
+            ensure!(self.ratifications.len() == 1, "Invalid number of ratifications");
+            ensure!(
+                self.transactions.num_accepted() == Self::NUM_GENESIS_TRANSACTIONS,
+                "Invalid number of accepted transactions"
+            );
+            ensure!(
+                self.transactions.num_finalize() == 2 * Self::NUM_GENESIS_TRANSACTIONS,
+                "Invalid number of finalized transactions"
+            );
+        }
+
+        Ok(true)
     }
 }
 
@@ -53,6 +60,6 @@ mod tests {
     fn test_genesis() {
         // Load the genesis block.
         let genesis_block = Block::<CurrentNetwork>::read_le(CurrentNetwork::genesis_bytes()).unwrap();
-        assert!(genesis_block.is_genesis());
+        assert!(genesis_block.is_genesis().unwrap());
     }
 }
--- a/ledger/block/src/header/metadata/genesis.rs
+++ b/ledger/block/src/header/metadata/genesis.rs
@@ -15,34 +15,29 @@
 
 use super::*;
 
+use snarkvm_utilities::ensure_equals;
+
+const GENESIS_ROUND: u64 = 0;
+const GENESIS_HEIGHT: u32 = 0;
+const GENESIS_CUMULATIVE_WEIGHT: u128 = 0;
+const GENESIS_CUMULATIVE_PROOF_TARGET: u128 = 0;
+
 impl<N: Network> Metadata<N> {
     /// Initializes the genesis metadata.
     pub fn genesis() -> Result<Self> {
-        // Prepare a genesis metadata.
-        let network = N::ID;
-        let round = 0;
-        let height = 0;
-        let cumulative_weight = 0;
-        let cumulative_proof_target = 0;
-        let coinbase_target = N::GENESIS_COINBASE_TARGET;
-        let proof_target = N::GENESIS_PROOF_TARGET;
-        let last_coinbase_target = N::GENESIS_COINBASE_TARGET;
-        let last_coinbase_timestamp = N::GENESIS_TIMESTAMP;
-        let timestamp = N::GENESIS_TIMESTAMP;
-
-        // Return the genesis metadata.
         Self::new(
-            network,
-            round,
-            height,
-            cumulative_weight,
-            cumulative_proof_target,
-            coinbase_target,
-            proof_target,
-            last_coinbase_target,
-            last_coinbase_timestamp,
-            timestamp,
+            N::ID,
+            GENESIS_ROUND,
+            GENESIS_HEIGHT,
+            GENESIS_CUMULATIVE_WEIGHT,
+            GENESIS_CUMULATIVE_PROOF_TARGET,
+            N::GENESIS_COINBASE_TARGET,
+            N::GENESIS_PROOF_TARGET,
+            N::GENESIS_COINBASE_TARGET,
+            N::GENESIS_TIMESTAMP,
+            N::GENESIS_TIMESTAMP,
         )
+        .with_context(|| "Failed to create genesis block")
     }
 
     /// Returns `true` if the metadata is a genesis metadata.i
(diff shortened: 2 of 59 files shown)


ledger/block/src/genesis.rs
```diff
@@ -20,25 +20,32 @@ impl<N: Network> Block<N> {
...
-            && self.solutions.is_empty()
-            // Ensure there is the correct number of accepted transaction in the genesis block.
-            && self.transactions.num_accepted() == Self::NUM_GENESIS_TRANSACTIONS
-            // Ensure there is the correct number of rejected transaction in the genesis block.
-            && self.transactions.num_rejected() == 0
-            // Ensure there is the correct number of finalize operations in the genesis block.
-            && self.transactions.num_finalize() == 2 * Self::NUM_GENESIS_TRANSACTIONS
-            // Ensure there are no aborted transaction IDs in the genesis block.
-            && self.aborted_transaction_ids.is_empty()
+    pub fn is_genesis(&self) -> Result<bool> {
```
ljedrz, line 23: this is a very welcome change for development - I remember having related issues in the past and not knowing which bit went wrong :+1:

ljedrz (commented):
I like the general direction; what I (and probably the other reviewers) would like to see is some examples for how the common errors are displayed (e.g. the VM halting), and to know whether you've checked that the `locktick` feature (which relies on backtraces) still works with the `panic = "abort"` change introduced here.

ledger/block/src/genesis.rs
kaimast: Yeah, this was the main reason I started this set of changes.

ledger/block/src/header/metadata/genesis.rs
```diff
@@ -15,58 +15,67 @@
...
-            && self.coinbase_target == N::GENESIS_COINBASE_TARGET
-            // Ensure the proof target in the genesis block is `GENESIS_PROOF_TARGET`.
-            && self.proof_target == N::GENESIS_PROOF_TARGET
-            // Ensure the last coinbase target in the genesis block is `GENESIS_COINBASE_TARGET`.
-            && self.last_coinbase_target == N::GENESIS_COINBASE_TARGET
-            // Ensure the last coinbase timestamp in the genesis block is `GENESIS_TIMESTAMP`.
-            && self.last_coinbase_timestamp == N::GENESIS_TIMESTAMP
-            // Ensure the timestamp in the genesis block is `GENESIS_TIMESTAMP`.
-            && self.timestamp == N::GENESIS_TIMESTAMP
+    /// Returns `true` if the metadata is a genesis metadata.i
```
ljedrz:
```suggestion
    /// Returns `true` if the metadata is a genesis metadata.
```

ledger/block/src/header/metadata/genesis.rs
```diff
@@ -15,58 +15,67 @@
...
+        ensure!(self.height == GENESIS_HEIGHT, "Genesis block not at genesis height");
+
+        ensure_equals!(self.cumulative_weight, GENESIS_CUMULATIVE_WEIGHT, "Invalid cumulative weight");
+        ensure_equals!(
+            self.cumulative_proof_target,
+            GENESIS_CUMULATIVE_PROOF_TARGET,
+            "Invalid cumulative proof target"
+        );
+        ensure_equals!(self.last_coinbase_timestamp, N::GENESIS_TIMESTAMP, "Invalid last coinbase timestamp");
+        ensure_equals!(self.last_coinbase_timestamp, N::GENESIS_TIMESTAMP, "Invalid last coinbase timestamp");
```
ljedrz, line 65:
```suggestion
```
duplicate; probably meant to be `self.timestamp`?

ledger/block/src/header/metadata/verify.rs
```diff
@@ -33,78 +35,50 @@ impl<N: Network> Metadata<N> {
...
-        );
-        // Ensure the cumulative weight is correct.
-        ensure!(
-            self.cumulative_weight == expected_cumulative_weight,
-            "Cumulative weight is incorrect in block {expected_height} (found '{}', expected '{}')",
+        if let Err(err) = self.check_validity() {
+            bail!("Metadata is malformed in block {expected_height}: {err}");
+        }
+
+        ensure_equals!(self.round, expected_round, "Round is incorrect in block {expected_height}");
```
ljedrz, line 42: this should be below the height check if we want to use the `expected_height` as authoritative, no?

ledger/query/LICENSE.md
ljedrz, line 1: is this intentional? I'd personally prefer to not have the copies of the license file everywhere, but it's the current convention

ljedrz (changes requested):
Could we split out the REST changes into a follow-up PR? While they build on the new abstractions from this one, they introduce some additional changes to that setup.

kaimast:
> Could we split out the REST changes into a follow-up PR? While they build on the new abstractions from this one, they introduce some additional changes to that setup.

Good point. I removed the commit for the REST changes from this PR and opened another draft PR for it (#2901).

ledger/block/src/header/metadata/verify.rs
kaimast: I am unsure. I just set the checks in the order the arguments are passed to the function.

kaimast: (It is also in the order that existed before this change)

ledger/query/LICENSE.md
kaimast: This change was moved to the other PR, but I would really prefer if we started to point to a single LICENSE file. This file was also formatted differently than the one in the root directory.

ljedrz (approved):
LGTM

ledger/block/src/header/metadata/genesis.rs
vicsn: This was marked as resolved but it still seems to be duplicate

ledger/block/src/header/metadata/mod.rs
```diff
@@ -80,40 +81,50 @@ impl<N: Network> Metadata<N> {
...
-                    && self.proof_target >= N::GENESIS_PROOF_TARGET
-                    // Ensure the coinbase target is larger than the proof target.
-                    && self.coinbase_target > self.proof_target
-                    // Ensure the last coinbase target is at or above the minimum.
-                    && self.last_coinbase_target >= N::GENESIS_COINBASE_TARGET
-                    // Ensure the last coinbase timestamp is after the genesis timestamp.
-                    && self.last_coinbase_timestamp >= N::GENESIS_TIMESTAMP
-                    // Ensure the timestamp in the block is after the genesis timestamp.
-                    && self.timestamp > N::GENESIS_TIMESTAMP
+    pub fn check_validity(&self) -> Result<()> {
```
vicsn, line 99: Why rename this function? Please keep refactors to a minimum, especially given that you keep the other `is_valid`'s intact

ledger/block/src/header/metadata/mod.rs
```diff
@@ -80,40 +81,50 @@ impl<N: Network> Metadata<N> {
...
+                min = N::GENESIS_COINBASE_TARGET,
+            );
+        }
+
+        if self.proof_target < N::GENESIS_PROOF_TARGET {
+            bail!(
+                "Invalid proof target: Was {actual} but expected {min} or greater.",
+                actual = self.proof_target,
+                min = N::GENESIS_PROOF_TARGET,
+            );
```
vicsn, line 130: Why not stick to the ensure! and >/>= convention? Makes it more consistent and easier to read.

ledger/block/src/header/metadata/verify.rs
vicsn: Yeah I guess neither is authoritative and the height is just shared in the error message as context. I think the refactor looks good as is.

ledger/block/src/header/mod.rs
```diff
@@ -72,29 +74,27 @@ impl<N: Network> Header<N> {
...
-                *self.previous_state_root != Field::zero()
-                    // Ensure the transactions root is nonzero.
-                    && self.transactions_root != Field::zero()
-                    // Ensure the finalize root is nonzero.
-                    && self.finalize_root != Field::zero()
-                    // Ensure the ratifications root is nonzero.
-                    && self.ratifications_root != Field::zero()
-                    // Ensure the metadata is valid.
-                    && self.metadata.is_valid()
+    pub fn check_validity(&self) -> Result<()> {
```
vicsn, line 89: Same, why go through the effort of changing the name? Mind you its also a pub function (like too many functions in snarkVM), so it has a chance of breaking some external utility.

ledger/src/advance.rs
```diff
@@ -71,7 +74,8 @@ impl<N: Network, C: ConsensusStorage<N>> Ledger<N, C> {
                 candidate_solutions,
                 candidate_transactions,
                 rng,
-            )?;
+            )
+            .with_context(|| "Failed to construct block template")?;
```
vicsn, line 78:
```suggestion
            .with_context(|| "Failed to construct block")?;
```

synthesizer/program/src/logic/instruction/operation/hash.rs
```diff
@@ -239,7 +239,7 @@ macro_rules! do_hash {
 /// Evaluate a hash operation.
 ///
 /// This allows running the hash without the machinery of stacks and registers.
-/// This is necessary for the Leo interpeter.
+/// This is necessary for the Leo interpreter.
```
vicsn, line 242: This is not worth anyone's time to fix and review.

utilities/src/vm_error.rs
```diff
@@ -23,22 +23,24 @@ macro_rules! try_vm_runtime {
...
         // indicate that the panic was expected and handled.
         panic::set_hook(Box::new(|e| {
-            let msg = e.to_string();
-            let msg = msg.split_ascii_whitespace().skip_while(|&word| word != "panicked").collect::<Vec<&str>>();
-            let mut msg = msg.join(" ");
-            msg = msg.replacen("panicked", "VM safely halted", 1);
             #[cfg(debug_assertions)]
-            eprintln!("{msg}");
+            {
+                let msg = e.to_string().replacen("panicked", "VM safely halted", 1);
```
vicsn: Can you explain why you removed these lines:
```
           let msg = msg.split_ascii_whitespace().skip_while(|&word| word != "panicked").collect::<Vec<&str>>();
            let mut msg = msg.join(" ");
```

utilities/src/vm_error.rs
```diff
@@ -23,22 +23,24 @@ macro_rules! try_vm_runtime {
...
+                eprintln!("{msg}");
```
vicsn, line 44: Shall we wrap it in `dev-eprint`? 

Deferring a discussion of also printing trace logs to here: https://github.com/ProvableHQ/snarkVM/issues/2849

vm/cli/main.rs
```diff
@@ -13,19 +13,53 @@
...
+    cli::{CLI, Updater},
+    utilities::display_error,
+};
 
 use clap::Parser;
 
-fn main() -> anyhow::Result<()> {
+use std::panic::catch_unwind;
+
+fn main() {
```
vicsn, line 25: Shall we remove the entire snarkVM CLI in a next PR? CC @d0cd 

It is a chore to keep updated. In the meantime:
- `execute` is already available in snarkOS or any other app importing snarkVM
- `new` creates an aleo instruction program to deploy, which is not useful by itself and not something we care about exposing to users. For testing, unit tests suffice until we have a clear need in place.

This can be a separate issue.

kaimast: Sounds good to me, but I would prefer doing this in a separate PR.

ledger/block/src/header/metadata/mod.rs
kaimast: I made this change so I can see the actual values, and not just `invalid coinbase target` as the error message.

ledger/block/src/header/metadata/genesis.rs
kaimast: I might have forgotten to push a commit. It is resolved now.

ledger/block/src/header/metadata/mod.rs
kaimast: The function signature is different. It returns `Result<()>` now, carrying more information on why the block is invalid.
I added `is_valid()` which simply calls `check_validity().is_ok()` for backwards compatibility.

ledger/block/src/header/mod.rs
kaimast: Same answer as to the other one. I kept `is_valid` for backwards compatibility.

ledger/src/advance.rs
kaimast: This is message is generated when `construct_block_template` fails. I think it is correct?

synthesizer/program/src/logic/instruction/operation/hash.rs
kaimast: Sorry, my editor shows typos as warnings, so I correct it. I will not fix unrelated typos in future PRs.

utilities/src/vm_error.rs
kaimast: I think this might also be a good thing to tackle in a separate PR?

utilities/src/vm_error.rs
kaimast: Must have been by mistake. I re-added the trimming.

synthesizer/program/src/logic/instruction/operation/hash.rs
vicsn: Ah thx for sharing, that explains :)

vicsn (approved):

PR #2814, opened 2025-07-16
[Feature] Enable safe external loading of inclusion prover for `wasm` targets

iamalwaysuncomfortable:
## Motivation

SnarkVM process stores the inclusion prover and verifier in static location in memory. It is initialized via `load_bytes` method defined by the `impl_remote!` macro in the `parameters` and lazily loaded into a static location the first time the `inclusion_proving_key` is called. If trying to build a transaction offline using a SnarkVM process running in userspace, this will first check if the inclusion prover is located at a specific location on disk, otherwise it will attempt to download it from the internet.

As a matter of security, wasm runtimes run in a sandboxed environment and generally do not have access to their host machine's file system directly. Thus, the normal approach of checking the local filesystem is not available, and the sole option for initializing the inclusion prover is download it.

If desiring to build a transaction on an offline machine that is using a `wasm` binary, the inclusion prover is unable to be loaded and offline transaction building will fail. This effectively means offline transaction creation is not available in a javascript environment. Many offline wallets, custodians, and other developers generally desire to write their applications in `node.js` and thus are currently blocked from doing so.

This PR introduces an option to insert the inclusion prover bytes from an external source, and verify that the checksum and byte buffer length match the expected values for the inclusion prover, allowing safe external insertion of the inclusion prover. 

## Test Plan

Add tests that ensure
* Bytes that do not match the expected inclusion prover bytes do not pass.
* The inclusion prover is successfully loaded from external sources.
* All functionality surrounding the inclusion prover that exists when the wasm flag is not present remains unaffected.

## Steps Before Marking PR as Ready for Review

* Write described tests.
* Code & style cleanup.
* Rebase onto the `staging` branch (this was branched from v3.8.0 tag in order to provide the SDK the ability to test offline transaction building against the latest stable `mainnet` tag)

### DIFF
--- a/synthesizer/process/src/trace/mod.rs
+++ b/synthesizer/process/src/trace/mod.rs
@@ -340,9 +340,20 @@ impl<N: Network> Trace<N> {
 
         if !batch_inclusions.is_empty() {
             // Fetch the inclusion proving key.
+            #[cfg(not(feature = "wasm"))]
             let proving_key = match inclusion_version {
                 Some(InclusionAssignmentWrapper::V0(..)) => ProvingKey::<N>::new(N::inclusion_v0_proving_key().clone()),
                 Some(InclusionAssignmentWrapper::V1(..)) => ProvingKey::<N>::new(N::inclusion_proving_key().clone()),
+                None => bail!("Invalid or missing inclusion version"),
+            };
+            #[cfg(feature = "wasm")]
+            let proving_key = match inclusion_version {
+                Some(InclusionAssignmentWrapper::V0(..)) => {
+                    ProvingKey::<N>::new(N::inclusion_v0_proving_key(None).clone())
+                }
+                Some(InclusionAssignmentWrapper::V1(..)) => {
+                    ProvingKey::<N>::new(N::inclusion_proving_key(None).clone())
+                }
                 None => bail!("Invalid or missing inclusion version"),
             };
             // Insert the inclusion proving key and assignments.
--- a/console/network/src/canary_v0.rs
+++ b/console/network/src/canary_v0.rs
@@ -241,10 +241,27 @@ impl Network for CanaryV0 {
         static INSTANCE: OnceLock<Arc<VarunaProvingKey<Console>>> = OnceLock::new();
         INSTANCE.get_or_init(|| {
             // Skipping the first byte, which is the encoded version.
-            Arc::new(
+            #[cfg(feature = "wasm")]
+            let inclusion_key = inclusion_key_bytes
+                .map(|bytes| {
+                    snarkvm_parameters::canary::InclusionProver::verify_bytes(&bytes)
+                        .expect("Bytes provided did not match expected inclusion checksum.");
+                    Arc::new(
+                        CircuitProvingKey::from_bytes_le(&bytes[1..]).expect("Failed to load inclusion proving key."),
+                    )
+                })
+                .unwrap_or_else(|| {
+                    Arc::new(
+                        CircuitProvingKey::from_bytes_le(&snarkvm_parameters::canary::INCLUSION_V0_PROVING_KEY[1..])
+                            .expect("Failed to load inclusion proving key."),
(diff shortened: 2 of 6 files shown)


vicsn:
> SnarkVM process stores the inclusion prover and verifier in static location in memory. It is initialized via load_bytes method defined by the impl_remote! macro in the parameters and lazily loaded into a static location the first time the inclusion_proving_key is called. If trying to build a transaction offline using a SnarkVM process running in userspace, this will first check if the inclusion prover is located at a specific location on disk, otherwise it will attempt to download it from the internet.

~I thought with this recent change this is not an issue anymore, `impl_remote!` on WASM will not assume existence of a filesystem anymore: https://github.com/ProvableHQ/snarkVM/tree/mainnet/parameters/src/macros.rs#L233~ Nevermind I'm wrong. You want to pass in the data optionally. 👍 But let's avoid the optionally compiled inputs.

> This PR introduces an option to insert the inclusion prover bytes from an external source, and verify that the checksum and byte buffer length match the expected values for the inclusion prover, allowing safe external insertion of the inclusion prover.

I did review the PR and the logic looks fine to me! Only concern I have is that this PR changes the interface of `fn inclusion_v0_proving_key`, which I assume is accessible in WASM (for why would you otherwise adjust it), therefore potentially requiring all wallets to update. So creating a new function might be better. But maybe because of the first point above we don't even need it anymore.

synthesizer/process/src/trace/mod.rs
```diff
@@ -342,8 +342,20 @@ impl<N: Network> Trace<N> {
         if !batch_inclusions.is_empty() {
             // Fetch the inclusion proving key.
             let proving_key = match inclusion_version {
-                Some(InclusionAssignmentWrapper::V0(..)) => ProvingKey::<N>::new(N::inclusion_v0_proving_key().clone()),
-                Some(InclusionAssignmentWrapper::V1(..)) => ProvingKey::<N>::new(N::inclusion_proving_key().clone()),
+                Some(InclusionAssignmentWrapper::V0(..)) => ProvingKey::<N>::new(
+                    N::inclusion_v0_proving_key(
+                        #[cfg(feature = "wasm")]
+                        None,
```
vicsn: Will this break, because it will go and try and use `INCLUSION_V0_PROVING_KEY` in this case?

iamalwaysuncomfortable: In the case that the inclusion version matches V0, It should have the following behavior:

* Crash the surrounding process due to the encapsulated `.expect` statements if the machine calling it is not online and hasn't been previously initialized with external bytes.
* Download the `InclusionV0` bytes if the machine is online if it hasn't been initialized.

iamalwaysuncomfortable:
> I did review the PR and the logic looks fine to me! Only concern I have is that this PR changes the interface of fn inclusion_v0_proving_key, which I assume is accessible in WASM (for why would you otherwise adjust it), therefore potentially requiring all wallets to update. So creating a new function might be better. But maybe because of the first point above we don't even need it anymore.

This won't require existing wallet impls to change because nobody calls this function manually in `wasm`. Currently this is first called in `prove_batch` function in `synthesizer/process/src/trace/mod.rs` which if the `wasm` environment is online will download the inclusion key as normal and initialize the static memory location. If the machines are offline and running `wasm` binaries, any calls to `prove_batch` just crash currently. 

All this does is enable clients in `wasm` to optionally call this function with `inclusion_proving_key` bytes the first time to initialize it if they're offline. This is necessary if custodians are working on airgapped machines and want to execute any function involving records.

Generally this is important because:
* Nobody wants to expose airgapped machines to the internet.
* Javascript processes are generally run as scripts that run logic and shutdown, and not as long running processes.

iamalwaysuncomfortable:
@vicsn this PR has been updated with your feedback and I responded to the concern in your last review. This PR is ready for re-review.

vicsn (approved):
LGTM!

PR #2816, opened 2025-07-16
build: use LLD for linking

kaimast:
The same change as [snarkos#3731](https://github.com/ProvableHQ/snarkOS/pull/3731), but for snarkVM.


kaimast:
I noticed `synthesizer-program-keccak` runs all tests for `synthesizer-program`, not just the keccak ones. That is also fixed in this PR.

vicsn (approved):
Thx!

PR #2817, opened 2025-07-17
Ureq default features to enable tls

vicsn:
## Motivation

`rusttls` is part of ureq's default features.
But because we're since recently enabling `json`, we need to manually enable `default-features`

## Test Plan

Tested `snarkos developer execute` on macOS.

CI issues:
- [x] `error: could not compile rustls (lib) due to 16 previous errors.   full command: cd "/home/circleci/project/wasm" && "cargo" "build" "--tests" "--target" "wasm32-unknown-unknown"`. Perhaps a newer ureq or a newer Rust version might help.
- [x] `Output record must be Version 0 before Consensus V8` - caused by some test innapropriately sampling record version 1 in combination with https://github.com/ProvableHQ/snarkVM/pull/2800

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -583,6 +583,15 @@ dependencies = [
 ]
 
 [[package]]
+name = "crc32fast"
+version = "1.5.0"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "9481c1c90cbf2ac953f07c8d4a58aa3945c425b7185c9154d67a65e4230da511"
+dependencies = [
+ "cfg-if",
+]
+
+[[package]]
 name = "criterion"
 version = "0.5.1"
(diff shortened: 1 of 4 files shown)


raychu86:
Can you explain why this was not an issue before? We've had the `json` trait enabled for a while

kaimast:
The default features are disabled for the workspace dependency. Should we just remove the `default-features=false` line there?

raychu86:
Looks like this breaks the WASM CI?

kaimast:
Does ureq support WASM? [This issue](https://github.com/algesten/ureq/issues/667) seems to indicate it does not. Maybe time is only needed for TLS and it works without that feature.

raychu86 (approved):

PR #2818, opened 2025-07-18
Tightened deserialization bounds

Antonio95:
Adds size checks for a few deserialised fields to prevent memory crashes. The modified structures are:
- [Authorization](https://github.com/ProvableHQ/snarkVM/blob/deserialization-bounds/synthesizer/process/src/stack/authorization/bytes.rs#L34)
- [ConfirmedTransaction](https://github.com/ProvableHQ/snarkVM/blob/deserialization-bounds/ledger/block/src/transactions/confirmed/bytes.rs#L31)
- [Deployment](https://github.com/ProvableHQ/snarkVM/blob/deserialization-bounds/ledger/block/src/transaction/deployment/bytes.rs#L36)
- [Execution](https://github.com/ProvableHQ/snarkVM/blob/deserialization-bounds/ledger/block/src/transaction/execution/bytes.rs#L34)
- [Request](https://github.com/ProvableHQ/snarkVM/blob/deserialization-bounds/console/program/src/request/bytes.rs#L40)
- [Transition](https://github.com/ProvableHQ/snarkVM/blob/deserialization-bounds/ledger/block/src/transition/bytes.rs#L38C1-L44C10)

I also have a couple of questions:
- In [Deployment](https://github.com/ProvableHQ/snarkVM/blob/deserialization-bounds/ledger/block/src/transaction/deployment/bytes.rs#L36), should it also be checked that the number of entries is > 0, or are deployments with no functions allowed?
- In [ConfirmedTransaction](https://github.com/ProvableHQ/snarkVM/blob/deserialization-bounds/ledger/block/src/transactions/confirmed/bytes.rs#L31), is `MAX_COMMANDS` the right bound to use for the number of finalize operations?
- It would be possible to try and add the bound `powers_len < MAX_NUM_POWERS` in the [ComitterKey](https://github.com/ProvableHQ/snarkVM/blob/deserialization-bounds/algorithms/src/polycommit/sonic_pc/data_structures.rs#L70) structure of the Sonic PCS. Should that constant be read from [this line in mainnet](https://github.com/ProvableHQ/snarkVM/blob/deserialization-bounds/parameters/src/mainnet/powers.rs#L62) or is that more specific than the Sonic code wants to be? In any case, it seems `CommitterKey` is never shared over the network and therefore deserialisation bounds may not be as relevant there (and are not in place for other fields such as `lagrange_bases_at_beta_len`).

### DIFF
--- a/curves/src/templates/bls12/g2.rs
+++ b/curves/src/templates/bls12/g2.rs
@@ -66,7 +66,7 @@ impl<P: Bls12Parameters> ToBytes for G2Prepared<P> {
 impl<P: Bls12Parameters> FromBytes for G2Prepared<P> {
     fn read_le<R: Read>(mut reader: R) -> IoResult<Self> {
         let ell_coeffs_len: u32 = FromBytes::read_le(&mut reader)?;
-        let mut ell_coeffs = Vec::new();
+        let mut ell_coeffs = Vec::with_capacity(ell_coeffs_len as usize);
         for _ in 0..ell_coeffs_len {
             let coeff_1: Fp2<P::Fp2Params> = FromBytes::read_le(&mut reader)?;
             let coeff_2: Fp2<P::Fp2Params> = FromBytes::read_le(&mut reader)?;
--- a/algorithms/src/polycommit/sonic_pc/data_structures.rs
+++ b/algorithms/src/polycommit/sonic_pc/data_structures.rs
@@ -17,6 +17,7 @@ use super::{LabeledPolynomial, PolynomialInfo};
 use crate::{crypto_hash::sha256::sha256, fft::EvaluationDomain, polycommit::kzg10};
 use snarkvm_curves::PairingEngine;
 use snarkvm_fields::{ConstraintFieldError, Field, PrimeField, ToConstraintField};
+use snarkvm_parameters::mainnet::MAX_NUM_POWERS;
 use snarkvm_utilities::{FromBytes, ToBytes, error, serialize::*};
 
 use hashbrown::HashMap;
@@ -68,6 +69,12 @@ impl<E: PairingEngine> FromBytes for CommitterKey<E> {
     fn read_le<R: Read>(mut reader: R) -> io::Result<Self> {
         // Deserialize `powers`.
         let powers_len: u32 = FromBytes::read_le(&mut reader)?;
+        // Ensure the number of powers is within bounds.
+        if powers_len as usize > MAX_NUM_POWERS {
+            return Err(error(format!(
+                "CommitterKey (from 'read_le') has too many points ({powers_len} > {MAX_NUM_POWERS})"
+            )));
+        }
         let mut powers_of_beta_g = Vec::with_capacity(powers_len as usize);
         for _ in 0..powers_len {
             let power: E::G1Affine = FromBytes::read_le(&mut reader)?;
--- a/console/network/src/lib.rs
+++ b/console/network/src/lib.rs
@@ -90,6 +90,15 @@ pub enum ConsensusVersion {
     V8 = 8,
 }
 
+/// The depth of the Merkle tree for transactions in a block.
+/// Note: The technical limit is 2^20 - 1 transactions, to allow compatibility with the
+/// finalize operations tree, which requires 1 leaf for the ratified finalize ID.
+pub const TRANSACTIONS_DEPTH: u8 = 20;
+/// The depth of the Merkle tree for the transaction.
+pub const TRANSACTION_DEPTH: u8 = 5;
+/// The depth of the Merkle tree for the transition.
+pub const TRANSITION_DEPTH: u8 = 5;
+
 pub trait Network:
     'static
     + Environment
@@ -211,6 +220,9 @@ pub trait Network:
     /// The maximum number of bytes in a transaction.
     // Note: This value must **not** be decreased as it would invalidate existing transactions.
     const MAX_TRANSACTION_SIZE: usize = 128_000; // 128 kB
+
+    /// The maximum number of transitions allowed in a transaction.
+    const MAX_TRANSITIONS: usize = usize::pow(2, TRANSACTION_DEPTH as u32);
 
     /// The state root type.
     type StateRoot: Bech32ID<Field<Self>>;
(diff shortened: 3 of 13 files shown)


curves/src/templates/bls12/g2.rs
```diff
@@ -66,7 +66,7 @@ impl<P: Bls12Parameters> ToBytes for G2Prepared<P> {
 impl<P: Bls12Parameters> FromBytes for G2Prepared<P> {
     fn read_le<R: Read>(mut reader: R) -> IoResult<Self> {
         let ell_coeffs_len: u32 = FromBytes::read_le(&mut reader)?;
-        let mut ell_coeffs = Vec::new();
+        let mut ell_coeffs = Vec::with_capacity(ell_coeffs_len as usize);
```
d0cd, line 69: Is there potential for over-allocation here?

d0cd:
Thanks for adding in the checks @Antonio95!
To your questions:
- `Deployment`s must have at least one function. I believe this is done in `Deployment::check_is_ordered`
- `MAX_COMMANDS` is a loose upper bound. Since `NumFinalizeSize` is a `u16` and `MAX_COMMANDS` is `u16::MAX`, the bound doesn't do much. `ConfirmedTransaction::accepted_deploy` does have a tighter bounds check. The rejected initializers should only be invoking the fee. The `accepted_execute` could be bounded by `MAX_WRITES`.
- I'll let @vicsn weigh in on the intention for the `CommitterKey`

vicsn:
> I'll let @vicsn weigh in on the intention for the CommitterKey

Yes https://github.com/ProvableHQ/snarkVM/blob/deserialization-bounds/parameters/src/mainnet/powers.rs#L62 is the globally relevant constant.

curves/src/templates/bls12/g2.rs
Antonio95: Yes, there is - but there was also before the change was made, as the lines immediately after (the former) `ell_coeffs = Vec::new();` filled the vector with that many elements anyway. So the change only makes allocation slightly more efficient in non-over-allocating cases: over-allocating cases are the same before an after.

I guess the more problematic issue is the lack of a bound for `ell_coefficients`, but that structure is not shared over the network (`G2Prepared<P>` is only read as part of reading `CommitterKey`, which I have been told isn't shared over the network (mentioned in the last point 3 of this PR's description)).

Antonio95:
> * `Deployment`s must have at least one function. I believe this is done in `Deployment::check_is_ordered`

@d0cd to be clear: do you mean there is no need to also add that `> 0` check at the deserialisation stage, in `read_le`?

Antonio95:
> * `MAX_COMMANDS` is a loose upper bound. Since `NumFinalizeSize` is a `u16` and `MAX_COMMANDS` is `u16::MAX`, the bound doesn't do much. `ConfirmedTransaction::accepted_deploy` does have a tighter bounds check. The rejected initializers should only be invoking the fee. The `accepted_execute` could be bounded by `MAX_WRITES`.

You're absolutely right that the current bound does nothing (except if `MAX_COMMANDS` shrinks in the future), thanks for catching that. Perhaps it's not the worst idea to leave the check in for future-proofing, but I am also happy to remove it (e. g. for the minimal efficiency gain) if you prefer.

So with the current code, the maximum amount of bytes [this part](https://github.com/ProvableHQ/snarkVM/blob/866ec8b6035c9a3d49d8a6e14100c2c62707cd63/ledger/block/src/transactions/confirmed/bytes.rs#L31-L32) can read is  `11 * 32 * (2^16 - 1) B = 22MB` ([the numbers come from the fact that a `FinalizeOperation` is  essentially 11 field elements](https://github.com/ProvableHQ/snarkVM/blob/866ec8b6035c9a3d49d8a6e14100c2c62707cd63/synthesizer/program/src/logic/finalize_operation/mod.rs#L25)). That seems assumable, doesn't it? Let me know if you want me to impose some other bound, though, such as one of the ones you mentioned.

Antonio95:
> > I'll let @vicsn weigh in on the intention for the CommitterKey
> 
> Yes https://github.com/ProvableHQ/snarkVM/blob/deserialization-bounds/parameters/src/mainnet/powers.rs#L62 is the globally relevant constant.

Thank you, I just [enforced the bound](https://github.com/ProvableHQ/snarkVM/blob/ea46564a7e6f65f5c0e7be4ac1aa6035ade8338b/algorithms/src/polycommit/sonic_pc/data_structures.rs#L73) at the cost of making the `MAX_NUM_POWERS` constant `pub` [here](https://github.com/ProvableHQ/snarkVM/blob/ea46564a7e6f65f5c0e7be4ac1aa6035ade8338b/parameters/src/mainnet/powers.rs#L62).

vicsn:
Logic LGTM, but as we're not in a rush for this one, I'll make sure we do an E2E sync test over this logic to ensure we don't regress.

Antonio95:
> Logic LGTM, but as we're not in a rush for this one, I'll make sure we do an E2E sync test over this logic to ensure we don't regress.

Good, although before merging a decision should be made on the `MAX_COMMANDS` bound check above because it is currently [only implemented in one of the arms of the affected function](https://github.com/ProvableHQ/snarkVM/blob/ea46564a7e6f65f5c0e7be4ac1aa6035ade8338b/ledger/block/src/transactions/confirmed/bytes.rs#L23-L87), so it should be either removed from that arm or implemented in the other ones.

d0cd:
> > * `Deployment`s must have at least one function. I believe this is done in `Deployment::check_is_ordered`
> 
> @d0cd to be clear: do you mean there is no need to also add that `> 0` check at the deserialisation stage, in `read_le`?

I don't believe so.

> You're absolutely right that the current bound does nothing (except if `MAX_COMMANDS` shrinks in the future), thanks for catching that. Perhaps it's not the worst idea to leave the check in for future-proofing, but I am also happy to remove it (e. g. for the minimal efficiency gain) if you prefer.

We can leave the bounds check in.

Antonio95:
Thank you very much, @d0cd . I've added the `num_finalize` check to the other arms so, in principle, everything is ready for the end-to-end test @vicsn mentioned.

raychu86 (approved):
Checks look reasonable

vicsn:
Going to merge this in now so we can just do a single sync test on staging before the next release

PR #2819, opened 2025-07-21
Conditionally abort failed scoped atomic batches

ljedrz:
This is a "tightening" for the behavior of scoped atomic write batches; if an error occurs outside of an atomic batch operation (which is unlikely, but technically possible), an `abort_atomic` is more appropriate, as there are no pending operations to roll back.

### DIFF
--- a/ledger/store/src/helpers/mod.rs
+++ b/ledger/store/src/helpers/mod.rs
@@ -61,7 +61,11 @@ macro_rules! atomic_batch_scope {
             },
             // Rewind this atomic batch scope.
             Err(err) => {
-                $self.atomic_rewind();
+                if is_atomic_in_progress {
+                    $self.atomic_rewind();
+                } else {
+                    $self.abort_atomic();
+                }
                 Err(err)
             }
         }
--- a/ledger/store/src/helpers/rocksdb/internal/map.rs
+++ b/ledger/store/src/helpers/rocksdb/internal/map.rs
@@ -642,6 +642,11 @@ mod tests {
             self.extra_maps.finish_atomic()
         }
 
+        fn abort_atomic(&self) {
+            self.own_map.abort_atomic();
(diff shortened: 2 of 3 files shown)


ledger/store/src/helpers/mod.rs
```diff
@@ -61,7 +61,11 @@ macro_rules! atomic_batch_scope {
             },
             // Rewind this atomic batch scope.
             Err(err) => {
-                $self.atomic_rewind();
+                if is_atomic_in_progress {
```
raychu86, line 64: Could you add documentation for the cases here?

raychu86:
Logically makes sense. 

Are the existing tests sufficient to test this case? 

Are there other implications from this change? Does this affect `snarkOS` transmission map operations that may be occurring at the same time?

ledger/store/src/helpers/mod.rs
raychu86: Also, how often is this case actual hit in our code?

When do we perform `atomic_batch_scope` without `atomic_finalize` being called first?

ljedrz:
> Are the existing tests sufficient to test this case?

Not really - actually, this would be indicative of something unexpected (which would reveal itself by the higher-level operation failing).
In order for these conditions to trigger, there would have to be some kind of a logic issue, in which case we'd rather do an extra full state cleanup just to be on the safe side.

> Are there other implications from this change? Does this affect snarkOS transmission map operations that may be occurring at the same time?

Hmm, actually the same question would apply right now in case of `abort_atomic` being called once the higher-level operation had failed and some other pending storage operation has started :thinking:. Then again, such conditions are very rare - the atomic operation is only ever concluded after calling `finish_atomic` (which happens at the very end of a successful `atomic_batch_scope`) or `abort_atomic`.

raychu86 (approved):

PR #2825, opened 2025-07-24
Do not determine consensus version based on user-submitted height

vicsn:
## Motivation

This PR fixes an issue with our consensus version determination. We need to determine it based on the verifier's current block height, not based on the user's submitted transaction state root height. This issue came up during testing the submission of transactions with old state roots. 

Because all we did was to **lower** the execution cost over time, it wasn't necessarily a vulnerability, but if we were to copy this type of logic over to other places it could create serious issues.

## Forking risk

In order to assure that 85564f9d3 does not cause a syncing node to fork , we would have to scan the entire history to see if an execution transaction was ever submitted with a V1 `state_root` to a V2 validator. This seems very likely to have occured just after the switch to `V2`. So 4c6d6ae20 just removes the usage of `execution_cost_v1` entirely.

## Test Plan

- [x] We should test our transaction pregeneration again with this. @meddle0x53 
- [ ] We can do a genesis-to-tip sync test with this PR just to be sure there is no forking issue.


synthesizer/src/vm/verify.rs
```diff
@@ -328,16 +328,11 @@ impl<N: Network, C: ConsensusStorage<N>> VM<N, C> {
...
-                            .unwrap_or_default();
-                        let consensus_version = N::CONSENSUS_VERSION(block_height)?;
-                        let (cost, (_, _)) = match consensus_version == ConsensusVersion::V1 {
-                            true => execution_cost_v1(&self.process().read(), execution)?,
-                            false => execution_cost_v2(&self.process().read(), execution)?,
-                        };
+                        // We are using execution_cost_v2 to compute the execution cost.
+                        // Technically, execution_cost_v2 only became active from ConsensusVersion::V2 onwards.
+                        // However, due to a bug in the transaction validation logic, some transactions may still have used execution_cost_v1.
+                        // execution_cost_v2 is less strict than execution_cost_v1, so we can safely use it here.
```
raychu86:
```suggestion
                        // Using `execution_cost_v2` is fine as a default because it is strictly cheaper than or equivalent to `execution_cost_v1`.
```

raychu86:
One of the tests we have that uses hard coded fee amounts to check verification is now breaking since the verification assumptions are now different.

vicsn:
Note to self, rebase this on https://github.com/ProvableHQ/snarkVM/pull/2831

raychu86 (approved):

raychu86:
Note that we can technically remove `execution_cost_v1` now since it will be unused. And in _**theory**_ we can remove `ConsensusVersion::V1`, and shift everything. But we shouldn't because all the public messaging and documentation will be off and incorrect.

vicsn:
Going to merge this in now so we can just do a single sync test on staging before the next release

PR #2829, opened 2025-07-26
Expose constants to they can be used elsewhere

vicsn:
## Motivation

For example in: https://github.com/ProvableHQ/snarkOS/pull/3753/commits/875a5d9ca227f10212e5ef9645ee65e67a87e66d

### DIFF
--- a/synthesizer/src/vm/verify.rs
+++ b/synthesizer/src/vm/verify.rs
@@ -33,7 +33,7 @@ macro_rules! ensure_is_unique {
 }
 
 impl<N: Network, C: ConsensusStorage<N>> VM<N, C> {
-    /// The maximum number of deployments to verify in parallel.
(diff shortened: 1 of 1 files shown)


meddle0x53 (approved):
LGTM

synthesizer/src/vm/verify.rs
```diff
@@ -34,9 +34,9 @@ macro_rules! ensure_is_unique {
 
 impl<N: Network, C: ConsensusStorage<N>> VM<N, C> {
     /// The maximum number of deployments to verify in parallel.
```
kaimast: The doc comment would be more descriptive like this.
```suggestion
    /// The maximum number of deployments that the VM can verify in parallel.
```

kaimast (dismissed):
It would be good to improve the doc comments for these constants, if they are public now.

raychu86 (approved):

PR #2828, opened 2025-07-26
Add ConsensusVersion::V10 and move logic into consensus_heights.rs

vicsn:
## Motivation

Add ConsensusVersion::V10
Move consensus version and height logic to a single file to make future editing again a little bit easier.

## Test Plan

unit tests pass

PR #2831, opened 2025-07-28
Feat implement arc 0005 fee reduction

Roee-87:
NOTE:  This PR should NOT be merged until ARC 0005 has been approved.

## Motivation

This PR implements [ARC 0005's](https://github.com/ProvableHQ/ARCs/discussions/92) fee reduction proposal.

## Test Plan

Tests for deployment using `deployment_cost_v2` have been added.  No tests for `execution_cost_v1` or `execution_cost_v2` are present...it's unclear whether I should add those for this PR.  I modified the `test_storage_cost_compute_correctly` test method to use `execution_cost_v3`.

### DIFF
--- a/synthesizer/process/src/cost.rs
+++ b/synthesizer/process/src/cost.rs
@@ -737,6 +737,30 @@ function dummy:",
                 deployment_cost(&process, &deployment_3).unwrap(),
                 (4186500, (943000, 603500, 1640000, 1000000))
             );
+
+            // Verify the deployment costs with deployment_cost_v2.
+            let mut deployment_0 = process.deploy::<A, _>(&program_0, rng).unwrap();
+            deployment_0.set_program_checksum_raw(Some(deployment_0.program().to_checksum()));
+            deployment_0.set_program_owner_raw(Some(Address::rand(rng)));
+            assert_eq!(deployment_cost_v2(&process, &deployment_0).unwrap(), (1053140, (879000, 24140, 50000, 1000000)));
+
+            let mut deployment_1 = process.deploy::<A, _>(&program_1, rng).unwrap();
+            deployment_1.set_program_checksum_raw(Some(deployment_1.program().to_checksum()));
+            deployment_1.set_program_owner_raw(Some(Address::rand(rng)));
+            assert_eq!(deployment_cost_v2(&process, &deployment_1).unwrap(), (1687660, (878000, 24140, 50000, 1000000)));
+
+            let mut deployment_2 = process.deploy::<A, _>(&program_2, rng).unwrap();
+            deployment_2.set_program_checksum_raw(Some(deployment_2.program().to_checksum()));
+            deployment_2.set_program_owner_raw(Some(Address::rand(rng)));
+            assert_eq!(deployment_cost_v2(&process, &deployment_2).unwrap(), (1217140, (911000, 24140, 182000, 1000000)));
+
+            let mut deployment_3 = process.deploy::<A, _>(&program_3, rng).unwrap();
+            deployment_3.set_program_checksum_raw(Some(deployment_3.program().to_checksum()));
+            deployment_3.set_program_owner_raw(Some(Address::rand(rng)));
+            assert_eq!(
+                deployment_cost_v2(&process, &deployment_3).unwrap(),
+                (2707140, (943000, 24140, 1640000, 1000000))
+            );
         }
 
         // Run the tests for all networks.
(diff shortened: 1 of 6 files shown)


synthesizer/src/vm/deploy.rs
```diff
@@ -56,7 +56,17 @@ impl<N: Network, C: ConsensusStorage<N>> VM<N, C> {
...
+        let (minimum_deployment_cost, _) = match N::CONSENSUS_VERSION(query.current_block_height()?)? {
+            ConsensusVersion::V1
+            | ConsensusVersion::V2
+            | ConsensusVersion::V3
+            | ConsensusVersion::V4
+            | ConsensusVersion::V5
+            | ConsensusVersion::V6
+            | ConsensusVersion::V7
+            | ConsensusVersion::V8 => deployment_cost_v1(&self.process().read(), &deployment)?,
+            _ => deployment_cost_v2(&self.process().read(), &deployment)?,
```
Roee-87: If I include `ConsensusVersion::V9` match to the old fee calculation method, then the synthesizer integration test passes locally.  However, this generates warnings about an "unreachable" case (since `V10` doesn't exist), which prohibits me from pushing that commit.

vicsn (changes requested):
Can you rebase on top of https://github.com/ProvableHQ/snarkVM/pull/2828

~@Roee-87 nevermind, if we'd merge this into testnet then we would only be at version 9.~

Nevermind the nevermind, we are deciding to use ConsensusVersion 10 again

synthesizer/src/vm/deploy.rs
vicsn: The versions are comparable as integers, you could use `<`. You can look for other uses of `N::CONSENSUS_VERSION` in the code for examples.

synthesizer/process/src/cost.rs
```diff
@@ -22,8 +22,54 @@ use console::{
 use snarkvm_ledger_block::{Deployment, Execution, Transaction};
 use snarkvm_synthesizer_program::{CastType, Command, Instruction, Operand};
 
+const COST_FACTOR: u64 = 25;
+
+/// Returns the *minimum* cost in microcredits to publish the given deployment using the reduced synthesis cost (total cost, (storage cost, synthesis cost, constructor cost, namespace cost)).
```
d0cd, line 28: With number of cost variants growing, it could be helpful to add comments on what differentiates each variant.

synthesizer/process/src/cost.rs
```diff
@@ -22,8 +22,54 @@ use console::{
...
+    let storage_cost = size_in_bytes
+        .checked_mul(N::DEPLOYMENT_FEE_MULTIPLIER)
+        .ok_or(anyhow!("The storage cost computation overflowed for a deployment"))?;
+
+    // Compute the synthesis cost in microcredits.
+    let synthesis_cost =
+        num_combined_variables.saturating_add(num_combined_constraints) * N::SYNTHESIS_FEE_MULTIPLIER / COST_FACTOR;
+
+    // Compute the constructor cost in microcredits.
+    let constructor_cost = constructor_cost_in_microcredits(&Stack::new(process, deployment.program())?)?;
```
d0cd, line 54: Do you want to also have a `constructor_cost_in_microcredits_v2`?

synthesizer/src/vm/deploy.rs
Roee-87: Ah I tried ConsensusVersion::V1..=ConsensunsVersion::V9, but that only works if I pattern matched a boolean (using `.contains(consensus_version)`) -- which didn't work for three cases.
I'll try comparison operators.

synthesizer/process/src/cost.rs
Roee-87: @vicsn

synthesizer/process/src/cost.rs
Roee-87: added:  [f8ed29a](https://github.com/ProvableHQ/snarkVM/pull/2831/commits/f8ed29af8cd3fc7c931b66c66659596f7769fb34)

synthesizer/process/src/cost.rs
vicsn: Yes good catch we should also lower the cost for the constructor

vicsn:
Closing this PR for now to clean up a bit, so a new PR will be easier for everyone to digest. We have a few weeks until the new Canary release, so there is no rush.

PR #2833, opened 2025-07-29
Make StaticQuery part of the Query Enum.

meddle0x53:
## Motivation

As per this comment from @vicsn https://github.com/ProvableHQ/snarkOS/pull/3723#discussion_r2239329029 I am making the StaticQuery part of the Query enum, so it is easier to just create it form string representations and not Box it and cast it (as it is done in the SnarkOs PR).

## Test Plan

Using the stress tests I'm testing the static queries for days and all works as expected.

(Write your test plan here)

## Related PRs

https://github.com/ProvableHQ/snarkOS/pull/3723

### DIFF
--- a/ledger/query/src/query.rs
+++ b/ledger/query/src/query.rs
@@ -13,7 +13,7 @@
 // See the License for the specific language governing permissions and
 // limitations under the License.
 
-use crate::QueryTrait;
+use crate::{QueryTrait, StaticQuery};
 use console::{
     network::prelude::*,
     program::{ProgramID, StatePath},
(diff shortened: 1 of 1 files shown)


ledger/query/src/query.rs
```diff
@@ -101,6 +114,7 @@ impl<N: Network, B: BlockStorage<N>> QueryTrait<N> for Query<N, B> {
                 }
                 _ => bail!("Unsupported network ID in inclusion query"),
             },
+            Self::STATIC(_query) => unimplemented!("Async calls are not supported by StaticQuery"),
```
vicsn: Can we change these `unimplemented` to `Err`? E.g. using `bail!`. Best to avoid any chance of nodes crashing.

meddle0x53: Makes sense, yes.

vicsn (approved):

PR #2837, opened 2025-07-30
Optimize consensus version height checking with early break

### LINKED ISSUE #2737: [Feature] [Perf] Speed up cache clearing on consensus versions
## 🚀 Feature

See: https://github.com/ProvableHQ/snarkVM/pull/2728#discussion_r2103858673

vicsn:
I am generated by an AI. Bleep Bloop.

Add reverse iteration with early break to reduce unnecessary iterations in consensus version height checking hot path.

Closes https://github.com/provableHQ/snarkVM/issues/2737

### DIFF
--- a/synthesizer/src/vm/mod.rs
+++ b/synthesizer/src/vm/mod.rs
@@ -456,9 +456,13 @@ impl<N: Network, C: ConsensusStorage<N>> VM<N, C> {
                 #[cfg(feature = "rocks")]
                 self.block_store().unpause_atomic_writes::<false>()?;
                 // If the block advances to a new consensus version, clear the partial verification cache.
-                // TODO: This may have performance implications if the version
-                // list grows large as it is in the hot path.
-                if N::CONSENSUS_VERSION_HEIGHTS().iter().any(|(_, height)| height == &block.height()) {
+                if N::CONSENSUS_VERSION_HEIGHTS().iter().rev().any(|(_, height)| {
+                    if block.height() < *height {
+                        // If the block height is less than the consensus version height, break early.
+                        return false;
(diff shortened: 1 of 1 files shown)


synthesizer/src/vm/mod.rs
```diff
@@ -456,9 +456,13 @@ impl<N: Network, C: ConsensusStorage<N>> VM<N, C> {
                 #[cfg(feature = "rocks")]
                 self.block_store().unpause_atomic_writes::<false>()?;
                 // If the block advances to a new consensus version, clear the partial verification cache.
-                // TODO: This may have performance implications if the version
-                // list grows large as it is in the hot path.
-                if N::CONSENSUS_VERSION_HEIGHTS().iter().any(|(_, height)| height == &block.height()) {
+                if N::CONSENSUS_VERSION_HEIGHTS().iter().rev().any(|(_, height)| {
```
raychu86, line 459: Maybe a `take_while` would be cleaner.

Something like

```
    .iter()
    .rev()
    .take_while(|(_, height)| block.height() >= *height)
    .any(|(_, height)| height == &block.height())
```

vicsn: Id rather not waste our time

niklaslong (approved):
Looks fine to me, if a bit verbose—I'm not sure we have enough instances of this around the codebase but would it be worth creating helpers for the height checks, I'm not super keen on that complexity living inline in the code? 🤔

vicsn:
Yeah it is a bit awkward - right now this is the only place in the code, happy to further extract it out as soon as we have more instances.

PR #2838, opened 2025-07-31
[Feature] Use http::Uri for query endpoints

kaimast:
The `http` crate provides a `Uri` type that can parse URLs and perform some sanity check for us on them. This PR changes the `Query` type to use `http::Uri` internally.

### Other changes
* Added `Query::get_public_balance` and `Query::get_unconfirmed_transaction` helper functions.
* The PR introduces a `Network::SHORT_NAME` constant that is either `testnet`, `mainnet`, or `canary`.
* The PR enables `rustls` and `gzip` features for `ureq` fixing #2842.
* There are now more unit tests for query parsing.

### Testing
Aside from the additional unit tests, this branch can be tested using [snarkOS PR #3759](https://github.com/ProvableHQ/snarkOS/pull/3759).

### Notes
* `Query::STATiC` should be `Query::Static` (it is not an acronym). I did not want to change that for this PR.
* `get_public_balance` is not supported by `Query::VM` yet, as it requires access to the `ConsensusStore`, not just the `BlockStore`. That is something we could change, but I did no, because I was worried about breaking things.


vicsn (commented):
We should test the snarkOS PR before un-drafting this.

ledger/query/src/query.rs
```diff
@@ -184,43 +178,140 @@ impl<N: Network, B: BlockStorage<N>> QueryTrait<N> for Query<N, B> {
...
+    pub fn get_public_balance(&self, address: &Address<N>) -> Result<u64> {
+        // Initialize the program id and account identifier.
+        let account_mapping = Identifier::<N>::from_str("account")?;
+        let credits = ProgramID::<N>::from_str("credits.aleo")?;
+
+        match self {
+            Self::VM(_block_store) => {
+                // TODO Query has only access to the block store, not the finalize store.
+                // this could easily be changed, but might break things downstream
+                todo!();
```
vicsn: We should not expose functions which allow the node or other applications to panic.

ledger/query/src/query.rs
```diff
@@ -184,43 +178,140 @@ impl<N: Network, B: BlockStorage<N>> QueryTrait<N> for Query<N, B> {
...
-            },
+            Self::REST(url) => Self::get_request_async(&format!("{url}{}/block/height/latest", N::SHORT_NAME)).await,
             Self::STATIC(_query) => bail!("Async calls are not supported by StaticQuery"),
         }
     }
 }
 
 impl<N: Network, B: BlockStorage<N>> Query<N, B> {
+    /// Gets the public account balance of an Aleo Address (in microcredits).
+    pub fn get_public_balance(&self, address: &Address<N>) -> Result<u64> {
```
vicsn: Can you remove Query::get_public_balance and Query::get_unconfirmed_transaction? We have a lot of endpoints, and it's not clear whether we want to support all or some of them at this level in the code.

vicsn (changes requested):
Great cleanup, minor adjustment required

ledger/query/src/query.rs
kaimast: Okay. I moved them back to `snarkos/cli` in [6ed6be7](https://github.com/ProvableHQ/snarkOS/pull/3759/commits/6ed6be705ba30a292926979190daf78f2c14395e) and removed them from this PR.

raychu86 (approved):

PR #2839, opened 2025-07-31
WIP: Add database read/write/delete timing metrics

### LINKED ISSUE #2367: [Feature] Adding DB usage metrics to SnarkVM
## 🚀 Feature
As discussed with @vicsn and @niklaslong , it might be useful to expose metrics on DB writes/reads from SnarkVM to be scraped by Prometheus

## Motivation

To better debug/diagnose latency in the network/nodes

## Implementation

Add metrics (most likely latency histograms) to the `ledger/store/src/helpers/rocksdb` Map atomic read/write methods

vicsn:
## Motivation

These will be instrumental in improving snarkOS node performance.

Closes https://github.com/provableHQ/snarkVM/issues/2367

- Add histogram metrics for database operations with timing distributions
- Instrument RocksDB and Memory map operations with timing measurements
- Add map type labels to distinguish different storage maps (block, transaction, etc.)
- Track cumulative time for read, write, and delete operations
- Add metrics registration for database histogram metrics
- Provide inline timing measurements for consistent metric collection

## Test Plan

- [ ] Should run a full devnet to assess performance regressions.

## Related PRs

https://github.com/ProvableHQ/snarkOS/pull/3769

### DIFF
--- a/ledger/store/src/helpers/memory/internal/map.rs
+++ b/ledger/store/src/helpers/memory/internal/map.rs
@@ -100,7 +100,16 @@ impl<
             }
             // Otherwise, insert the key-value pair directly into the map.
             false => {
-                self.map.write().insert(bincode::serialize(&key)?, value);
+                let serialized_key = bincode::serialize(&key)?;
+                let start = std::time::Instant::now();
+                self.map.write().insert(serialized_key, value);
+                let duration = start.elapsed().as_secs_f64();
+                snarkvm_metrics::histogram_label(
+                    snarkvm_metrics::database::WRITE_DURATION,
+                    "map_type",
+                    "memory".to_string(),
+                    duration,
+                );
             }
         }
 
@@ -119,7 +128,16 @@ impl<
             }
             // Otherwise, remove the key-value pair directly from the map.
             false => {
-                self.map.write().remove(&bincode::serialize(&key)?);
+                let serialized_key = bincode::serialize(&key)?;
+                let start = std::time::Instant::now();
+                self.map.write().remove(&serialized_key);
+                let duration = start.elapsed().as_secs_f64();
+                snarkvm_metrics::histogram_label(
+                    snarkvm_metrics::database::DELETE_DURATION,
+                    "map_type",
+                    "memory".to_string(),
+                    duration,
+                );
             }
         }
 
@@ -217,12 +235,24 @@ impl<
                 .collect::<Result<Vec<_>>>()?;
 
             // Perform all the queued operations.
+            let start = std::time::Instant::now();
             for (key, value) in prepared_operations {
                 match value {
-                    Some(value) => locked_map.insert(key, value),
-                    None => locked_map.remove(&key),
+                    Some(value) => {
+                        locked_map.insert(key, value);
+                    }
+                    None => {
+                        locked_map.remove(&key);
+                    }
                 };
             }
+            let duration = start.elapsed().as_secs_f64();
+            snarkvm_metrics::histogram_label(
+                snarkvm_metrics::database::WRITE_DURATION,
+                "map_type",
+                "memory_batch".to_string(),
+                duration,
+            );
         }
 
         // Clear the checkpoint stack.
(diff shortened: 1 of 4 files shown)


raychu86:
Might be a good opportunity to expand our metrics macro usage as well.

We are currently bound to a single type of `histogram` and `gauge` usage, despite the underlying `metrics` crate being richer in functionality - https://github.com/ProvableHQ/snarkOS/pull/3064.

ledger/store/src/helpers/memory/internal/map.rs
```diff
@@ -315,7 +345,17 @@ impl<
         K: Borrow<Q>,
         Q: PartialEq + Eq + Hash + Serialize + ?Sized,
     {
-        Ok(self.map.read().get(&bincode::serialize(key)?).cloned().map(Cow::Owned))
+        let serialized_key = bincode::serialize(key)?;
+        let start = std::time::Instant::now();
```
ljedrz, line 349: nit: with this many uses I'd import the `Instant`

ledger/store/src/helpers/memory/internal/map.rs
```diff
@@ -100,7 +100,16 @@ impl<
             }
             // Otherwise, insert the key-value pair directly into the map.
             false => {
-                self.map.write().insert(bincode::serialize(&key)?, value);
+                let serialized_key = bincode::serialize(&key)?;
+                let start = std::time::Instant::now();
```
ljedrz, line 104: shouldn't we feature-gate it (`metrics`)? obtaining the time is a perf penalty in itself

ljedrz: I think it would work best with a simple macro taking a block containing the section to be measured

metrics/src/lib.rs
```diff
@@ -16,16 +16,26 @@
...
+const HISTOGRAM_NAMES: [&str; 3] = [database::READ_DURATION, database::WRITE_DURATION, database::DELETE_DURATION];
 
 pub mod committee {
     pub const TOTAL_STAKE: &str = "snarkvm_ledger_committee_total_stake";
 }
 
+pub mod database {
+    pub const READ_DURATION: &str = "snarkvm_database_read_duration_seconds";
+    pub const WRITE_DURATION: &str = "snarkvm_database_write_duration_seconds";
+    pub const DELETE_DURATION: &str = "snarkvm_database_delete_duration_seconds";
```
ljedrz, line 28: are the numbers for deletion any different for the "plain" writes? we might just need `READ` and `WRITE`

metrics/src/lib.rs
```diff
@@ -103,3 +113,15 @@ pub fn histogram<V: Into<f64>>(name: &'static str, value: V) {
 pub fn histogram_label<V: Into<f64>>(name: &'static str, label_key: &'static str, label_value: String, value: V) {
     ::metrics::histogram!(name, label_key => label_value).record(value.into());
 }
+
+/// Times a database operation and records the duration with the given map type label.
+pub fn time_database_operation<F, R>(metric_name: &'static str, map_type: &str, operation: F) -> R
```
ljedrz, line 118: doesn't seem like it's used at the moment, but this (or a macro) would be preferable for readability and performance (feature-gating)

ledger/store/src/helpers/memory/internal/map.rs
kaimast, line 104: Does `Instant` have high precision enough for this? These calls might take less than a millisecond.

kaimast:
For the block sync work, I have been using tracing directly (see [this PR](https://github.com/ProvableHQ/snarkVM/pull/2887)). That might be another option if these calls happen too frequently to report them to Prometheus.

ledger/store/src/helpers/memory/internal/map.rs
ljedrz: yes, but those values are likely to be miniscule when running in memory only

PR #2840, opened 2025-07-31
Actually enable hashing of R1CS

vicsn:
## Motivation

Previous attempts to hash R1CS seems to have hashed empty vectors.

## Test Plan

- [ ] Test ongoing in https://github.com/ProvableHQ/aleo-program-regressions/tree/update-argument

### DIFF
--- a/circuit/environment/src/circuit.rs
+++ b/circuit/environment/src/circuit.rs
@@ -351,6 +351,9 @@ impl Environment for Circuit {
             assert_eq!(0, circuit.borrow().num_private());
             assert_eq!(1, circuit.borrow().num_variables());
             assert_eq!(0, circuit.borrow().num_constraints());
+            // Optionally save a hash of the R1CS.
+            #[cfg(feature = "save_r1cs_hashes")]
+            r1cs.save_hash();
             // Convert the R1CS instance to an assignment.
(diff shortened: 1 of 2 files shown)


circuit/environment/src/circuit.rs
```diff
@@ -351,6 +351,9 @@ impl Environment for Circuit {
             assert_eq!(0, circuit.borrow().num_private());
             assert_eq!(1, circuit.borrow().num_variables());
             assert_eq!(0, circuit.borrow().num_constraints());
+            // Optionally save a hash of the R1CS.
+            #[cfg(feature = "save_r1cs_hashes")]
+            r1cs.save_hash();
```
mikebenfield, line 356: Is this the right place to put this? Over in `r1cs-hasher` `R1CS_HASHES` is still empty.

ljedrz: yeah, since the current place where `save_hash` is called is basically in the very next line (it's the first operation in `Assignment::from(r1cs)`), wouldn't this effectively be the same end result?

vicsn: My bad, somehow missed the other `save_hash` call, it was late

PR #1999, opened 2023-09-19
Introduce compact block headers

raychu86:
## Motivation

This PR introduces `CompactBatchHeader`, `CompactBatchCertificate`, and `CompactSubdag` by updating the vec of transmission ids in the batch header into a vec of index references. This is safe, because these references point to transactions, ratifications, and solutions that should already be included in the block. This will help reduce the block size by 1 TransmissionID (32 bytes) per transmission.

PR #2841, opened 2025-08-01
[Perf] Compact subdag

ljedrz:
A revival of https://github.com/ProvableHQ/snarkVM/pull/2304.

Filing as a draft for now, as due to the age of the aforelinked PR it was a challenging rebase, and further adjustments may be needed.

# Design
The goal of this PR is to decrease the bandwidth usage during block sync. If we scale to large amounts of validators (e.g. 100), besides signatures, duplicate transmissions may take up a significant amount of space.

To achieve this:
- Prior transmission IDs and new transmissions are stored in a Block just once.
- CompactHeaders, CompactCertificates and Compact Subdags reference these with ordered u32 indices
- New Quorum blocks are created, stored and broadcasted as Compact.
- Recipients must convert back to full transmissions on the ocassions:
  - For every received block to recompute the batch_id, which in turn ensures the signatures are actually valid.
  - For validators who do at-tip Sync with full certificates or who boot up with full certificates. This is not needed in theory, but require a large refactor to change. Conversion to full certificates can be unchecked.

## TODOs
- [x] Replace bitset by index vec to retain ordering
- [x] Make sure Block contains Optional `{prior, aborted}_{transaction,solution}_transmission_ids`, and stores/retrieves them from disk appropriately
- [ ] Out of Scope: Consider re-use the `SubdagTransmissions` object (or something similar) to unify passing in transmissions for all the conversions to-from compact subdag. This should not stand in the way of passing iterators, we must avoid Cloning these large amounts of data wherever we can...

## Related past PRs
https://github.com/ProvableHQ/snarkVM/pull/1999, https://github.com/ProvableHQ/snarkOS/pull/3000.

### DIFF
--- a/ledger/block/src/bytes.rs
+++ b/ledger/block/src/bytes.rs
@@ -45,7 +45,10 @@ impl<N: Network> FromBytes for Block<N> {
         let (prior_solution_transmission_ids, aborted_solution_transmission_ids, aborted_solution_ids) = if version >= 2
         {
             // Read the number of prior solution transmission IDs.
-            let num_prior_solutions = u32::read_le(&mut reader)?; // TODO: what is a sane bound on this value?
+            let num_prior_solutions = u32::read_le(&mut reader)?;
+            if num_prior_solutions as usize > Solutions::<N>::max_aborted_solutions().map_err(error)? {
+                return Err(error("Invalid number of prior solutions IDs in the block"));
+            }
             // Read the prior solution transmission IDs.
             let mut prior_solution_transmission_ids = Vec::with_capacity(num_prior_solutions as usize);
             for _ in 0..num_prior_solutions {
@@ -61,7 +64,7 @@ impl<N: Network> FromBytes for Block<N> {
             for _ in 0..num_aborted_solutions {
                 aborted_solution_transmission_ids.push(FromBytes::read_le(&mut reader)?);
             }
-            (Some(prior_solution_transmission_ids), Some(aborted_solution_transmission_ids), vec![])
+            (Some(prior_solution_transmission_ids), Some(aborted_solution_transmission_ids), None)
         } else {
             // Read the number of aborted solution IDs.
             let num_aborted_solutions = u32::read_le(&mut reader)?;
@@ -74,7 +77,7 @@ impl<N: Network> FromBytes for Block<N> {
             for _ in 0..num_aborted_solutions {
                 aborted_solution_ids.push(FromBytes::read_le(&mut reader)?);
             }
-            (None, None, aborted_solution_ids)
+            (None, None, Some(aborted_solution_ids))
         };
 
         // Read the transactions.
@@ -99,7 +102,7 @@ impl<N: Network> FromBytes for Block<N> {
                 for _ in 0..num_aborted_transactions {
                     aborted_transaction_transmission_ids.push(FromBytes::read_le(&mut reader)?);
                 }
-                (Some(prior_transaction_transmission_ids), Some(aborted_transaction_transmission_ids), vec![])
+                (Some(prior_transaction_transmission_ids), Some(aborted_transaction_transmission_ids), None)
             } else {
                 // Read the number of aborted transaction IDs.
                 let num_aborted_transactions = u32::read_le(&mut reader)?;
@@ -112,7 +115,7 @@ impl<N: Network> FromBytes for Block<N> {
                 for _ in 0..num_aborted_transactions {
                     aborted_transaction_ids.push(FromBytes::read_le(&mut reader)?);
                 }
-                (None, None, aborted_transaction_ids)
+                (None, None, Some(aborted_transaction_ids))
             };
 
         // Construct the block.
@@ -168,8 +171,10 @@ impl<N: Network> ToBytes for Block<N> {
         match version {
             1 => {
                 // Write the aborted solution IDs.
-                (u32::try_from(self.aborted_solution_ids.len()).map_err(error))?.write_le(&mut writer)?;
-                self.aborted_solution_ids.write_le(&mut writer)?;
+                let aborted_solution_ids =
+                    self.aborted_solution_ids.as_ref().ok_or(error("missing aborted_solution_ids"))?;
+                (u32::try_from(aborted_solution_ids.len()).map_err(error))?.write_le(&mut writer)?;
+                aborted_solution_ids.write_le(&mut writer)?;
             }
             _ => {
                 // Write the prior solution transmission ids.
@@ -195,8 +200,10 @@ impl<N: Network> ToBytes for Block<N> {
         match version {
             1 => {
                 // Write the aborted transaction ids.
-                (u32::try_from(self.aborted_transaction_ids.len()).map_err(error))?.write_le(&mut writer)?;
-                self.aborted_transaction_ids.write_le(&mut writer)?;
+                let aborted_transaction_ids =
+                    self.aborted_transaction_ids.as_ref().ok_or(error("missing aborted_transaction_ids"))?;
+                (u32::try_from(aborted_transaction_ids.len()).map_err(error))?.write_le(&mut writer)?;
+                aborted_transaction_ids.write_le(&mut writer)?;
             }
             _ => {
                 // Write the prior transaction transmission ids.
--- a/ledger/block/src/bytes.rs
+++ b/ledger/block/src/bytes.rs
@@ -22,7 +22,7 @@ impl<N: Network> FromBytes for Block<N> {
         // Read the version.
         let version = u8::read_le(&mut reader)?;
         // Ensure the version is valid.
-        if version != 1 {
+        if ![1, 2].contains(&version) {
             return Err(error("Invalid block version"));
         }
 
@@ -42,32 +42,78 @@ impl<N: Network> FromBytes for Block<N> {
         // Read the solutions.
         let solutions: Solutions<N> = FromBytes::read_le(&mut reader)?;
 
-        // Read the number of aborted solution IDs.
-        let num_aborted_solutions = u32::read_le(&mut reader)?;
-        // Ensure the number of aborted solutions IDs is within bounds (this is an early safety check).
-        if num_aborted_solutions as usize > Solutions::<N>::max_aborted_solutions().map_err(error)? {
-            return Err(error("Invalid number of aborted solutions IDs in the block"));
-        }
-        // Read the aborted solution IDs.
-        let mut aborted_solution_ids = Vec::with_capacity(num_aborted_solutions as usize);
-        for _ in 0..num_aborted_solutions {
-            aborted_solution_ids.push(FromBytes::read_le(&mut reader)?);
-        }
+        let (prior_solution_transmission_ids, aborted_solution_transmission_ids, aborted_solution_ids) = if version >= 2
+        {
+            // Read the number of prior solution transmission IDs.
+            let num_prior_solutions = u32::read_le(&mut reader)?; // TODO: what is a sane bound on this value?
+            // Read the prior solution transmission IDs.
+            let mut prior_solution_transmission_ids = Vec::with_capacity(num_prior_solutions as usize);
+            for _ in 0..num_prior_solutions {
+                prior_solution_transmission_ids.push(FromBytes::read_le(&mut reader)?);
+            }
+            // Read the number of aborted solution transmission IDs.
+            let num_aborted_solutions = u32::read_le(&mut reader)?;
+            if num_aborted_solutions as usize > Solutions::<N>::max_aborted_solutions().map_err(error)? {
+                return Err(error("Invalid number of aborted solutions IDs in the block"));
+            }
+            // Read the aborted solution transmission IDs.
+            let mut aborted_solution_transmission_ids = Vec::with_capacity(num_aborted_solutions as usize);
+            for _ in 0..num_aborted_solutions {
+                aborted_solution_transmission_ids.push(FromBytes::read_le(&mut reader)?);
+            }
+            (Some(prior_solution_transmission_ids), Some(aborted_solution_transmission_ids), vec![])
+        } else {
+            // Read the number of aborted solution IDs.
+            let num_aborted_solutions = u32::read_le(&mut reader)?;
+            // Ensure the number of aborted solutions IDs is within bounds (this is an early safety check).
+            if num_aborted_solutions as usize > Solutions::<N>::max_aborted_solutions().map_err(error)? {
+                return Err(error("Invalid number of aborted solutions IDs in the block"));
+            }
+            // Read the aborted solution IDs.
+            let mut aborted_solution_ids = Vec::with_capacity(num_aborted_solutions as usize);
+            for _ in 0..num_aborted_solutions {
+                aborted_solution_ids.push(FromBytes::read_le(&mut reader)?);
+            }
+            (None, None, aborted_solution_ids)
+        };
 
         // Read the transactions.
         let transactions = FromBytes::read_le(&mut reader)?;
 
-        // Read the number of aborted transaction IDs.
-        let num_aborted_transactions = u32::read_le(&mut reader)?;
-        // Ensure the number of aborted transaction IDs is within bounds (this is an early safety check).
-        if num_aborted_transactions as usize > Transactions::<N>::max_aborted_transactions().map_err(error)? {
-            return Err(error("Invalid number of aborted transaction IDs in the block"));
-        }
-        // Read the aborted transaction IDs.
-        let mut aborted_transaction_ids = Vec::with_capacity(num_aborted_transactions as usize);
-        for _ in 0..num_aborted_transactions {
-            aborted_transaction_ids.push(FromBytes::read_le(&mut reader)?);
-        }
+        let (prior_transaction_transmission_ids, aborted_transaction_transmission_ids, aborted_transaction_ids) =
+            if version >= 2 {
+                // Read the number of prior transaction transmission IDs.
+                let num_prior_transactions = u32::read_le(&mut reader)?; // TODO: what is a sane bound on this value?
+                // Read the prior transaction transmission IDs.
+                let mut prior_transaction_transmission_ids = Vec::with_capacity(num_prior_transactions as usize);
+                for _ in 0..num_prior_transactions {
+                    prior_transaction_transmission_ids.push(FromBytes::read_le(&mut reader)?);
+                }
+                // Read the number of aborted transaction transmission IDs.
+                let num_aborted_transactions = u32::read_le(&mut reader)?;
+                if num_aborted_transactions as usize > Transactions::<N>::max_aborted_transactions().map_err(error)? {
+                    return Err(error("Invalid number of aborted transaction IDs in the block"));
+                }
+                // Read the aborted transaction transmission IDs.
+                let mut aborted_transaction_transmission_ids = Vec::with_capacity(num_aborted_transactions as usize);
+                for _ in 0..num_aborted_transactions {
+                    aborted_transaction_transmission_ids.push(FromBytes::read_le(&mut reader)?);
+                }
+                (Some(prior_transaction_transmission_ids), Some(aborted_transaction_transmission_ids), vec![])
+            } else {
+                // Read the number of aborted transaction IDs.
+                let num_aborted_transactions = u32::read_le(&mut reader)?;
+                // Ensure the number of aborted transaction IDs is within bounds (this is an early safety check).
+                if num_aborted_transactions as usize > Transactions::<N>::max_aborted_transactions().map_err(error)? {
+                    return Err(error("Invalid number of aborted transaction IDs in the block"));
+                }
+                // Read the aborted transaction IDs.
+                let mut aborted_transaction_ids = Vec::with_capacity(num_aborted_transactions as usize);
+                for _ in 0..num_aborted_transactions {
+                    aborted_transaction_ids.push(FromBytes::read_le(&mut reader)?);
+                }
+                (None, None, aborted_transaction_ids)
+            };
 
         // Construct the block.
         let block = Self::from(
@@ -76,8 +122,12 @@ impl<N: Network> FromBytes for Block<N> {
             authority,
             ratifications,
             solutions,
+            prior_solution_transmission_ids,
+            aborted_solution_transmission_ids,
             aborted_solution_ids,
             transactions,
+            prior_transaction_transmission_ids,
+            aborted_transaction_transmission_ids,
             aborted_transaction_ids,
         )
         .map_err(error)?;
@@ -95,7 +145,9 @@ impl<N: Network> ToBytes for Block<N> {
     #[inline]
     fn write_le<W: Write>(&self, mut writer: W) -> IoResult<()> {
         // Write the version.
-        1u8.write_le(&mut writer)?;
+        let version =
+            if N::CONSENSUS_VERSION(self.height()).map_err(error)? >= ConsensusVersion::V10 { 2u8 } else { 1 };
+        version.write_le(&mut writer)?;
 
         // Write the block hash.
         self.block_hash.write_le(&mut writer)?;
@@ -113,16 +165,57 @@ impl<N: Network> ToBytes for Block<N> {
         // Write the solutions.
         self.solutions.write_le(&mut writer)?;
 
-        // Write the aborted solution IDs.
-        (u32::try_from(self.aborted_solution_ids.len()).map_err(error))?.write_le(&mut writer)?;
-        self.aborted_solution_ids.write_le(&mut writer)?;
+        match version {
+            1 => {
+                // Write the aborted solution IDs.
+                (u32::try_from(self.aborted_solution_ids.len()).map_err(error))?.write_le(&mut writer)?;
+                self.aborted_solution_ids.write_le(&mut writer)?;
+            }
+            _ => {
+                // Write the prior solution transmission ids.
+                let prior_solution_transmission_ids = self
+                    .prior_solution_transmission_ids
+                    .as_ref()
+                    .ok_or(error("missing prior_solution_transmission_ids"))?;
+                (u32::try_from(prior_solution_transmission_ids.len()).map_err(error))?.write_le(&mut writer)?;
+                prior_solution_transmission_ids.write_le(&mut writer)?;
+                // Write the aborted solution transmission ids.
+                let aborted_solution_transmission_ids = self
+                    .aborted_solution_transmission_ids
+                    .as_ref()
+                    .ok_or(error("missing aborted_solution_transmission_ids"))?;
+                (u32::try_from(aborted_solution_transmission_ids.len()).map_err(error))?.write_le(&mut writer)?;
+                aborted_solution_transmission_ids.write_le(&mut writer)?;
+            }
+        }
 
         // Write the transactions.
         self.transactions.write_le(&mut writer)?;
 
-        // Write the aborted transaction IDs.
-        (u32::try_from(self.aborted_transaction_ids.len()).map_err(error))?.write_le(&mut writer)?;
-        self.aborted_transaction_ids.write_le(&mut writer)
+        match version {
+            1 => {
+                // Write the aborted transaction ids.
+                (u32::try_from(self.aborted_transaction_ids.len()).map_err(error))?.write_le(&mut writer)?;
+                self.aborted_transaction_ids.write_le(&mut writer)?;
+            }
+            _ => {
+                // Write the prior transaction transmission ids.
+                let prior_transaction_transmission_ids = self
+                    .prior_transaction_transmission_ids
+                    .as_ref()
+                    .ok_or(error("missing prior_transaction_transmission_ids"))?;
+                (u32::try_from(prior_transaction_transmission_ids.len()).map_err(error))?.write_le(&mut writer)?;
+                prior_transaction_transmission_ids.write_le(&mut writer)?;
+                // Write the aborted transaction transmission ids.
+                let aborted_transaction_transmission_ids = self
+                    .aborted_transaction_transmission_ids
+                    .as_ref()
+                    .ok_or(error("missing aborted_transaction_transmission_ids"))?;
+                (u32::try_from(aborted_transaction_transmission_ids.len()).map_err(error))?.write_le(&mut writer)?;
+                aborted_transaction_transmission_ids.write_le(&mut writer)?;
+            }
+        }
+        Ok(())
     }
 }
 
@@ -138,6 +231,7 @@ mod tests {
         let rng = &mut TestRng::default();
 
         for expected in [crate::test_helpers::sample_genesis_block(rng)].into_iter() {
+            // todo: sample compact too for the Block tests
             // Check the byte representation.
             let expected_bytes = expected.to_bytes_le()?;
             assert_eq!(expected, Block::read_le(&expected_bytes[..])?);
--- a/ledger/block/src/bytes.rs
+++ b/ledger/block/src/bytes.rs
@@ -86,7 +86,10 @@ impl<N: Network> FromBytes for Block<N> {
         let (prior_transaction_transmission_ids, aborted_transaction_transmission_ids, aborted_transaction_ids) =
             if version >= 2 {
                 // Read the number of prior transaction transmission IDs.
-                let num_prior_transactions = u32::read_le(&mut reader)?; // TODO: what is a sane bound on this value?
+                let num_prior_transactions = u32::read_le(&mut reader)?;
+                if num_prior_transactions as usize > Transactions::<N>::max_aborted_transactions().map_err(error)? {
+                    return Err(error("Invalid number of prior transaction IDs in the block"));
+                }
                 // Read the prior transaction transmission IDs.
                 let mut prior_transaction_transmission_ids = Vec::with_capacity(num_prior_transactions as usize);
                 for _ in 0..num_prior_transactions {
@@ -238,7 +241,6 @@ mod tests {
         let rng = &mut TestRng::default();
 
         for expected in [crate::test_helpers::sample_genesis_block(rng)].into_iter() {
-            // todo: sample compact too for the Block tests
             // Check the byte representation.
             let expected_bytes = expected.to_bytes_le()?;
             assert_eq!(expected, Block::read_le(&expected_bytes[..])?);
--- a/ledger/block/src/lib.rs
+++ b/ledger/block/src/lib.rs
@@ -63,6 +63,9 @@ use snarkvm_ledger_narwhal_subdag::Subdag;
 use snarkvm_ledger_narwhal_transmission_id::TransmissionID;
 use snarkvm_ledger_puzzle::{PuzzleSolutions, Solution, SolutionID};
 
+#[cfg(not(feature = "serial"))]
+use rayon::prelude::*;
+
 #[derive(Clone, PartialEq, Eq)]
 pub struct Block<N: Network> {
     /// The hash of this block.
@@ -77,10 +80,20 @@ pub struct Block<N: Network> {
     ratifications: Ratifications<N>,
     /// The solutions in the block.
     solutions: Solutions<N>,
+    /// The transmission IDs of both accepted and aborted solutions in the previous block.
+    /// Technically prior checksums can be retrieved from the ledger, but that is unwieldy.
+    prior_solution_transmission_ids: Option<Vec<TransmissionID<N>>>,
+    /// The aborted solution transmission IDs in this block.
+    aborted_solution_transmission_ids: Option<Vec<TransmissionID<N>>>,
     /// The aborted solution IDs in this block.
     aborted_solution_ids: Vec<SolutionID<N>>,
     /// The transactions in this block.
     transactions: Transactions<N>,
+    /// The transmission IDs of both accepted and aborted transactions in the previous block.
+    /// Technically prior checksums can be retrieved from the ledger, but that is unwieldy.
+    prior_transaction_transmission_ids: Option<Vec<TransmissionID<N>>>,
+    /// The aborted transaction transmission IDs in this block.
+    aborted_transaction_transmission_ids: Option<Vec<TransmissionID<N>>>,
     /// The aborted transaction IDs in this block.
     aborted_transaction_ids: Vec<N::TransactionID>,
 }
@@ -110,8 +123,12 @@ impl<N: Network> Block<N> {
             authority,
             ratifications,
             solutions,
+            None,
+            None,
             aborted_solution_ids,
             transactions,
+            None,
+            None,
             aborted_transaction_ids,
         )
     }
@@ -124,8 +141,12 @@ impl<N: Network> Block<N> {
         subdag: Subdag<N>,
         ratifications: Ratifications<N>,
         solutions: Solutions<N>,
+        prior_solution_transmission_ids: Option<Vec<TransmissionID<N>>>,
+        aborted_solution_transmission_ids: Option<Vec<TransmissionID<N>>>,
         aborted_solution_ids: Vec<SolutionID<N>>,
         transactions: Transactions<N>,
+        prior_transaction_transmission_ids: Option<Vec<TransmissionID<N>>>,
+        aborted_transaction_transmission_ids: Option<Vec<TransmissionID<N>>>,
         aborted_transaction_ids: Vec<N::TransactionID>,
     ) -> Result<Self> {
         // Construct the beacon authority.
@@ -137,8 +158,12 @@ impl<N: Network> Block<N> {
             authority,
             ratifications,
             solutions,
+            prior_solution_transmission_ids,
+            aborted_solution_transmission_ids,
             aborted_solution_ids,
             transactions,
+            prior_transaction_transmission_ids,
+            aborted_transaction_transmission_ids,
             aborted_transaction_ids,
         )
     }
@@ -151,8 +176,12 @@ impl<N: Network> Block<N> {
         authority: Authority<N>,
         ratifications: Ratifications<N>,
         solutions: Solutions<N>,
+        prior_solution_transmission_ids: Option<Vec<TransmissionID<N>>>,
+        aborted_solution_transmission_ids: Option<Vec<TransmissionID<N>>>,
         aborted_solution_ids: Vec<SolutionID<N>>,
         transactions: Transactions<N>,
+        prior_transaction_transmission_ids: Option<Vec<TransmissionID<N>>>,
+        aborted_transaction_transmission_ids: Option<Vec<TransmissionID<N>>>,
         aborted_transaction_ids: Vec<N::TransactionID>,
     ) -> Result<Self> {
         // Ensure the number of aborted solutions IDs is within the allowed range.
@@ -197,15 +226,9 @@ impl<N: Network> Block<N> {
                 // Ensure the signature is valid.
                 ensure!(signature.verify(&address, &[block_hash]), "Invalid signature for block {}", header.height());
             }
-            Authority::Quorum(subdag) => {
-                // Ensure the transmission IDs from the subdag correspond to the block.
-                Self::check_subdag_transmissions(
-                    subdag,
-                    &solutions,
-                    &aborted_solution_ids,
-                    &transactions,
-                    &aborted_transaction_ids,
-                )?;
+            Authority::Quorum(_subdag) => {
+                // We purposefully do not repeat expensive checks on the subdag,
+                // which are already performed in `fn verify_authority` in `verify.rs`.
             }
         }
 
@@ -231,14 +254,17 @@ impl<N: Network> Block<N> {
             authority,
             ratifications,
             solutions,
+            prior_solution_transmission_ids,
+            aborted_solution_transmission_ids,
             aborted_solution_ids,
             transactions,
+            prior_transaction_transmission_ids,
+            aborted_transaction_transmission_ids,
             aborted_transaction_ids,
         )
     }
 
-    /// Initializes a new block from the given block hash, previous block hash, block header,
-    /// authority, ratifications, solutions, transactions, and aborted transaction IDs.
+    /// Initializes a new block.
     /// This is only called by [`Block::from`], which validates the block components.
     fn from_unchecked(
         block_hash: N::BlockHash,
@@ -247,8 +273,12 @@ impl<N: Network> Block<N> {
         authority: Authority<N>,
         ratifications: Ratifications<N>,
         solutions: Solutions<N>,
+        prior_solution_transmission_ids: Option<Vec<TransmissionID<N>>>,
+        aborted_solution_transmission_ids: Option<Vec<TransmissionID<N>>>,
         aborted_solution_ids: Vec<SolutionID<N>>,
         transactions: Transactions<N>,
+        prior_transaction_transmission_ids: Option<Vec<TransmissionID<N>>>,
+        aborted_transaction_transmission_ids: Option<Vec<TransmissionID<N>>>,
         aborted_transaction_ids: Vec<N::TransactionID>,
     ) -> Result<Self> {
         // Return the block.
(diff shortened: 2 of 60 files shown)


ledger/block/src/bytes.rs
```diff
@@ -113,13 +135,25 @@ impl<N: Network> ToBytes for Block<N> {
         // Write the solutions.
         self.solutions.write_le(&mut writer)?;
 
+        // Write the prior solution ids.
```
vicsn: Serialization needs to be backwards compatible. So for sure we have to write the new components at the end. And only optionally atomically read the new components.

ledger/block/src/lib.rs
```diff
@@ -77,10 +77,14 @@ pub struct Block<N: Network> {
...
     /// The solutions in the block.
     solutions: Solutions<N>,
+    /// The prior solution ids in the block.
+    prior_solution_ids: Vec<SolutionID<N>>,
     /// The aborted solution IDs in this block.
     aborted_solution_ids: Vec<SolutionID<N>>,
     /// The transactions in this block.
     transactions: Transactions<N>,
+    /// The prior transaction ids in the block.
+    prior_transaction_ids: Vec<N::TransactionID>,
```
vicsn: So these are ids of transactions which originally landed or aborted in a previous block. Please document in the docblock. Same for solutions.


### brainstrom ramble

Proposals may reference them. By caching the prior ids in the Block, CompactHeaders can still be converted to full BatchHeaders and their signature can be verified. Alternatively: Perhaps we are ok with CompactProposals to never be converted? They are discarded after the 100 garbage collection round threshold anyway. But this seems like a more involved change.

ledger/block/src/lib.rs
```diff
@@ -259,11 +279,81 @@ impl<N: Network> Block<N> {
...
+
+        // Collect ratification IDs.
+        let ratification_ids = ratifications.ratification_ids().copied().collect_vec();
+        // Collect transaction IDs.
+        let transaction_ids = transactions.transaction_ids().copied().collect_vec();
+
+        // Convert Quorum authority to subdag with full batch certificates.
+        subdag.into_full(
+            ratification_ids,
+            solutions.as_puzzle_solutions().cloned(),
```
vicsn: To check: can/should we pass any of these items by reference/iterator to avoid the extra allocations?

ledger/block/src/lib.rs
```diff
@@ -259,11 +279,81 @@ impl<N: Network> Block<N> {
...
+            prior_solution_ids,
+            transaction_ids,
+            prior_transaction_ids,
+            aborted_transaction_ids,
+        )
+    }
+
+    /// Borrow the Block and return a Subdag with full batch certificates.
+    pub fn to_full_subdag(&self) -> Result<Subdag<N>> {
```
vicsn, line 309: Are these two different methods (taking self and &self) really needed? Would it be cleaner for one to call the other?

ledger/block/src/serialize.rs
```diff
@@ -27,8 +27,10 @@ impl<N: Network> Serialize for Block<N> {
                 block.serialize_field("authority", &self.authority)?;
                 block.serialize_field("ratifications", &self.ratifications)?;
                 block.serialize_field("solutions", &self.solutions)?;
+                block.serialize_field("prior_solution_ids", &self.prior_solution_ids)?;
```
vicsn: Serialization needs to be backwards compatible.

ledger/block/src/transactions/mod.rs
```diff
@@ -221,6 +221,14 @@ impl<N: Network> Transactions<N> {
         self.transactions.keys()
     }
 
+    /// Returns the unconfirmed transaction IDs, for all transactions in `self`.
+    pub fn unconfirmed_transaction_ids(&self) -> Result<Vec<N::TransactionID>> {
+        self.transactions
+            .values()
+            .map(|confirmed_tx| confirmed_tx.to_unconfirmed_transaction_id())
```
vicsn, line 228: Note: we have an issue open to make this call a lot more efficient by not having to recompute the merkle tree

ledger/narwhal/batch-header/src/lib.rs
```diff
@@ -24,6 +24,8 @@ mod serialize;
 mod string;
 mod to_id;
 
+use std::collections::BTreeSet;
```
vicsn: Move a few lines down to external dependencies

ledger/narwhal/batch-header/src/bytes.rs
```diff
@@ -46,7 +46,7 @@ impl<N: Network> FromBytes for BatchHeader<N> {
             )));
         }
         // Read the transmission IDs.
-        let mut transmission_ids = IndexSet::new();
+        let mut transmission_ids = BTreeSet::new();
```
vicsn: FYI I guess I changed this because I thought it was gonna be faster.

ledger/narwhal/compact-certificate/Cargo.toml
```diff
@@ -0,0 +1,67 @@
...
+workspace = true
+
+[dependencies.snarkvm-ledger-narwhal-traits]
+workspace = true
+
+[dependencies.snarkvm-ledger-puzzle]
+workspace = true
+
+[dependencies.bit-set]
+version = "0.5.3"
```
vicsn: How comfy do we feel with this dependency?

ledger/narwhal/compact-certificate/src/lib.rs
```diff
@@ -0,0 +1,286 @@
...
+        let compact_header = CompactHeader::new(
+            batch_certificate.batch_header(),
+            ratifications,
+            solutions,
+            prior_solutions,
+            transactions,
+            prior_transactions,
+            aborted_transactions,
+        )?;
+        let signatures = batch_certificate.signatures().cloned().collect();
```
vicsn: Destructure instead of clone

ledger/narwhal/compact-header/src/lib.rs
```diff
@@ -0,0 +1,394 @@
...
+            "Could not find all transmission_ids"
+        );
+        BatchHeader::from(
+            self.author,
+            self.round,
+            self.timestamp,
+            self.committee_id,
+            transmission_ids,
+            self.previous_certificate_ids,
+            // self.last_election_certificate_ids,
```
vicsn: Remove commented out code

ledger/narwhal/subdag/src/bytes.rs
```diff
@@ -43,37 +50,81 @@ impl<N: Network> FromBytes for Subdag<N> {
...
+        }
     }
 }
 
 impl<N: Network> ToBytes for Subdag<N> {
     /// Writes the subdag to the buffer.
     fn write_le<W: Write>(&self, mut writer: W) -> IoResult<()> {
         // Write the version.
-        1u8.write_le(&mut writer)?;
+        2u8.write_le(&mut writer)?;
```
vicsn: We should still support version 1 for backwards compatibility

ledger/narwhal/subdag/src/lib.rs
```diff
@@ -116,39 +200,79 @@ fn weighted_median(timestamps_and_stake: Vec<(i64, u64)>) -> i64 {
...
+
+    /// Initializes a new subdag.
+    pub fn from_compact(subdag: BTreeMap<u64, IndexSet<CompactCertificate<N>>>) -> Result<Self> {
+        // Perform sanity checks.
+        sanity_check_subdag!(&subdag)?;
+        // Construct the subdag.
+        Ok(Self::Compact { subdag })
+    }
+
+    // pub fn get_and_iter(&self, round: &u64) -> Option<Vec<LeaderCertificate<N>>> {
```
vicsn: Do we need this?

ledger/narwhal/subdag/src/lib.rs
```diff
@@ -215,23 +472,107 @@ impl<N: Network> Subdag<N> {
...
+        };
         // Compute the subdag root.
         Ok(*N::merkle_tree_bhp::<SUBDAG_CERTIFICATES_DEPTH>(&leaves)?.root())
     }
-}
 
-impl<N: Network> Deref for Subdag<N> {
-    type Target = BTreeMap<u64, IndexSet<BatchCertificate<N>>>;
+    /// Returns the subdag with compact certificates
+    pub fn into_compact(
```
vicsn, line 517: Now that we have the full PR, do we really need both the into_compact and into_full? Or is one-way sufficient (outside of testing)?

ledger/narwhal/subdag/src/serialize.rs
```diff
@@ -36,8 +43,17 @@ impl<'de, N: Network> Deserialize<'de> for Subdag<N> {
...
-                    .map_err(de::Error::custom)?)
+                // Check if a full Subdag field is present.
+                let subdag_is_full = match value.get("full_subdag") {
+                    Some(..) => true,
+                    None => false,
+                };
+                match subdag_is_full {
+                    true => Ok(Self::from_full(DeserializeExt::take_from_value::<D>(&mut value, "full_subdag")?)
+                        .map_err(de::Error::custom)?),
+                    false => Ok(Self::from_full(DeserializeExt::take_from_value::<D>(&mut value, "compact_subdag")?)
```
vicsn: This seems wrong. Should we add tests in this file?

ledger/src/advance.rs
```diff
@@ -20,19 +20,50 @@ impl<N: Network, C: ConsensusStorage<N>> Ledger<N, C> {
...
+        let SubdagTransmissions { prior_included_transmissions, aborted_transmissions, transmissions } =
+            subdag_transmissions;
+
         // Decouple the transmissions into ratifications, solutions, and transactions.
         let (ratifications, solutions, transactions) = decouple_transmissions(transmissions.into_iter())?;
+        // Decouple the prior_transmissions into ratifications, solutions, and transactions.
+        let (prior_ratifications, prior_solution_ids, prior_transaction_ids) =
+            decouple_transmission_ids(prior_included_transmissions)?;
+        // Decouple the aborted_transmissions into ratifications, solutions, and transactions.
+        // TODO: add support for aborted solution ids
```
vicsn: Is this still relevant?

ledger/src/advance.rs
```diff
@@ -20,19 +20,50 @@ impl<N: Network, C: ConsensusStorage<N>> Ledger<N, C> {
...
+            decouple_transmission_ids(aborted_transmissions)?;
         // Currently, we do not support ratifications from the memory pool.
-        ensure!(ratifications.is_empty(), "Ratifications are currently unsupported from the memory pool");
+        ensure!(
+            ratifications.is_empty() && prior_ratifications.is_empty() && aborted_ratifications.is_empty(),
+            "Ratifications are currently unsupported from the memory pool"
+        );
         // Construct the block template.
-        let (header, ratifications, solutions, aborted_solution_ids, transactions, aborted_transaction_ids) =
+        let (header, ratifications, solutions, aborted_solution_ids, transactions, aborted_transaction_ids_2) =
```
vicsn: Funny name

ledger/src/advance.rs
```diff
@@ -20,19 +20,50 @@ impl<N: Network, C: ConsensusStorage<N>> Ledger<N, C> {
...
+        // Construct Ratification IDs.
+        let ratification_ids = ratifications.ratification_ids().copied().collect_vec();
+        // Construct Transaction IDs.
+        let transaction_ids = transactions.unconfirmed_transaction_ids()?;
+        // Construct the aborted Transaction IDs.
+        let aborted_transaction_ids =
+            aborted_transaction_ids.into_iter().chain(aborted_transaction_ids_2.into_iter()).collect_vec();
+
+        // Construct the compact Subdag
+        let subdag = subdag.into_compact(
```
vicsn: We'll need to support compact and non-compact depending on the ConsensusVersion.

vicsn (changes requested):
Very nice, I think its close to finished.

In `check_next_block` we should disallow the Full subdag from ConsensusVersion::V10, and disallow Compact before it, to avoid forking and complexity.

Do you also want to draft the snarkOS PR? Make sure to do it on top of Kai's recent snarkOS PR which uses the new PendingBlock.

I'll give another security review after the current round of fixes.

ledger/block/src/bytes.rs
ljedrz: good point, this wasn't a concern back when this PR was first drafted; or perhaps we should consider bumping the version of the `Block` object? that would allow us to read them conditionally from any spot we want to

ledger/block/src/lib.rs
ljedrz: 2 can easily be avoided (I'll push it soon), the others are more involved; due to the large scope of the PR, I'd recommend tracking potential perf improvements in a separate issue, so they can be looked into as a follow-up

ledger/block/src/lib.rs
ljedrz: actually, neither is currently used, so they can both be removed

ledger/narwhal/batch-header/src/bytes.rs
ljedrz: is the order not meaningful here, then (since `IndexSet` preserves insertion order, and `BTreeSet` depends on `Ord`)? because a `HashSet` should be faster than either of those

ledger/narwhal/compact-certificate/Cargo.toml
ljedrz: it is widely used in the ecosystem, and has some prominent maintainers; I'd say comfy enough, though I see that the version can be bumped

ledger/narwhal/compact-certificate/src/lib.rs
ljedrz: this requires a visibility change, but the benefit is clear :+1:

ledger/narwhal/subdag/src/bytes.rs
ljedrz: version 1 is still supported for reads; don't we want the 2nd one to be the default one for writes?

ledger/narwhal/subdag/src/lib.rs
ljedrz: nope, it was a prototyping artifact during the rebase

ledger/narwhal/subdag/src/lib.rs
ljedrz: `into_compact` appears to suffice

ledger/narwhal/subdag/src/serialize.rs
ljedrz: indeed; yeah, I think we should add compact cases to `sample_subdag`, even if just via `into_compact`

ledger/src/advance.rs
ljedrz: seems to be in place now :+1:

ledger/src/advance.rs
ljedrz: indeed - any suggestions for how to rename this one? `aborted_transaction_ids_from_template`?

ledger/src/advance.rs
ljedrz: hmm so we intend to merge and release this feature before introducing an updated version of consensus? sure, we can do that, just let me know which versions to consider; I now understand the earlier remark on serialization compatibility for writes

ledger/block/src/bytes.rs
vicsn: Yeah that makes sense! You can copy the approach from how we serialize other objects

ledger/block/src/lib.rs
vicsn: Maybe the snarkOS PR uses them

ljedrz: ah, good point - I'll restore it if that's the case

ledger/narwhal/batch-header/src/bytes.rs
vicsn: Good point.

FIFO ordering is nice, but not critical. I guess a problem is that the bitset doesn't retain the ordering. So we can't go from compact back to full certificates if we use IndexMap. Let's await the snarkOS PR to see if the conversion is needed.

If the conversion is needed, we can still make do without indexmap. A hard to manipulate deterministic ordering (e.g. alphabetical) may be acceptable. To make manipulation for MEV harder, we can randomize the order before linearization (in the snarkOS PR).

ledger/narwhal/subdag/src/bytes.rs
vicsn: All nodes should switch to reading and writing the new version at a given ConsensusVersion height. Otherwise  if not all nodes restart at the same time the ones who do wont be compatible anymore and the chain halts.

vicsn: The switch will likely be defined in snarkOS

ledger/src/advance.rs
vicsn: You can use ConsensusVersion::V10, it is the next one

ledger/src/advance.rs
vicsn: I had another read, I think `aborted_transaction_ids_from_finalization` would be nice.

ledger/block/src/bytes.rs
```diff
@@ -42,6 +42,20 @@ impl<N: Network> FromBytes for Block<N> {
         // Read the solutions.
         let solutions: Solutions<N> = FromBytes::read_le(&mut reader)?;
 
+        let prior_solution_ids = if version >= 2 {
+            // Read the number of prior solution IDs.
+            let num_prior_solutions = u32::read_le(&mut reader)?;
+            // TODO: we might want to restrict this value
+            // Read the aborted transaction IDs.
```
raychu86:
```suggestion
            // Read the prior solution IDs.
```

An upper-bound check of `max_aborted_solutions` would be sufficient here for the TODO

ledger/block/src/lib.rs
```diff
@@ -77,10 +77,14 @@ pub struct Block<N: Network> {
...
     /// The solutions in the block.
     solutions: Solutions<N>,
+    /// The IDs of both accepted and aborted solutions in the previous block.
+    prior_solution_ids: Vec<SolutionID<N>>,
     /// The aborted solution IDs in this block.
     aborted_solution_ids: Vec<SolutionID<N>>,
     /// The transactions in this block.
     transactions: Transactions<N>,
+    /// The IDs of both accepted and aborted transactions in the previous block.
+    prior_transaction_ids: Vec<N::TransactionID>,
```
raychu86: I think `prior_*` is a bit too ambiguous. Our current usage of `aborted_or_existing_transaction_ids` is also quite verbose, but I'd still lean more towards verbosity for the sake of clarity.

Other options could be:
`existing_*`
`past_*`

ledger/block/src/lib.rs
raychu86: Hmm that's a good point. I don't think there is a reason to convert Compact to full, but I'd need to double check snarkOS usages

ledger/block/src/serialize.rs
```diff
@@ -20,7 +20,11 @@ impl<N: Network> Serialize for Block<N> {
     fn serialize<S: Serializer>(&self, serializer: S) -> Result<S::Ok, S::Error> {
         match serializer.is_human_readable() {
             true => {
-                let mut block = serializer.serialize_struct("Block", 9)?;
+                let mut block = serializer.serialize_struct("Block", 12)?;
```
raychu86: The number of elements here will depend on the version.

Same for `prior_solution_ids` and `prior_transaction_ids` - they should not exist for version 1 blocks

ledger/block/src/lib.rs
raychu86: Also for backwards compatibility, an `Option<Vec<...>>` would be better than just an empty Vec for 2 reasons.

1. Services (wallets and explorers) have already serialized blocks without this field and might break if this is a required field.
2. Previous blocks will have an empty vec here; however, in reality they might actually have transactions or solutions that belong there based on the subdag state.

ledger/narwhal/batch-header/src/bytes.rs
raychu86: I feel like the insertion ordering is significant here, because how are you able to linearize the subdag correctly if the ordering is now alphabetical? 

It also incentivizes users to generate transactions that are lexicographically ahead.

Maybe I'm not seeing something clearly, because I do see that the new `SubdagTransmissions` struct does preserve the IndexMap ordering.

raychu86: Ah, I see that the bitset is not ordered by insertion, so a conversion from compact to full may not be exact.

Is it possible to make the BitSet ordered somehow? Maybe with a wrapper and vec with ordering?

ledger/narwhal/subdag/src/lib.rs
raychu86: Huh, I was just wondering where `transaction_indices()` was used and I can't seem to find an instance.

Are there any checks in Block or Subdag verification that should be checking these new bitmaps?

(I can't think of a reason block production might need to call `into_full` for the compact subdag, but we should probably test correctness of the back and forth conversion somewhere)

ledger/narwhal/batch-header/src/bytes.rs
vicsn: Preliminary update: we now think there is no need to convert to full subdags, let's revisit this comment when the snarkOS PR is drafted and compiles.

ledger/narwhal/compact-header/src/lib.rs
```diff
@@ -0,0 +1,401 @@
...
+    batch_id: Field<N>,
+    /// The author of the batch.
+    author: Address<N>,
+    /// The round number.
+    round: u64,
+    /// The timestamp.
+    timestamp: i64,
+    /// The committee ID.
+    committee_id: Field<N>,
+    /// The set of transaction indices in a block.
```
kaimast: Could you expand this?

I assume it would be something like `The transactions included in this batch, stored compactly as indices in the set of all transactions of the associated block`

kaimast:
This looks good to me, but I will do another pass before approving. Hopefully, merging/rebasing to staging will make CI pass.

However, I am wondering if it makes sense to add three new crates for this? The `compact-header` crate is only 800LOC and the `compact-batch` even less. This functionality could be added to the existing `snarkvm-ledger-narwhal-batch-certificate` and `snarkvm-ledger-narwhal-batch-header` crates without problems, unless I am missing something.

I know this mostly boils down to preference, but I feel like these many small crates add a lot of bloat. The most extreme case is the `traits` crate which is <50 LOC but over 200 lines of Cargo.toml, README, and LICENSE files.
Are there any cases where someone would want to use the regular `header` crate but not the `compact-header` crate? If they are always used together, it makes a lot of sense to merge them, IMHO.

ledger/block/src/lib.rs
ljedrz: While I wouldn't mind adding `Option` wrappers here, there is an alternative approach that we could take, i.e. adding the `version` (which is already stored in the database and sent over the network) to the `Block` object, which IMO would make reasoning about the contents more explicit and futureproof.

vicsn:
> However, I am wondering if it makes sense to add three new crates for this?

If we can find a good reason for it, e.g. faster compile times, we can refactor the crates in the future.

console/network/src/consensus_heights.rs
```diff
@@ -103,7 +103,7 @@ pub const TEST_CONSENSUS_VERSION_HEIGHTS: [(ConsensusVersion, u32); NUM_CONSENSU
     (ConsensusVersion::V7, 15),
     (ConsensusVersion::V8, 16),
     (ConsensusVersion::V9, 17),
-    (ConsensusVersion::V10, 18),
+    (ConsensusVersion::V10, 51),
```
vicsn: Note: think about how the test can be fixed without requiring a larger default consensus height... Can we pass in a custom consensus version in the test...?

ledger/src/advance.rs
```diff
@@ -283,8 +283,14 @@ impl<N: Network, C: ConsensusStorage<N>> Ledger<N, C> {
                         if self.is_solution_limit_reached(&prover_address, num_accepted_solutions) {
                             return false;
                         }
+                        // From ConsensusVersion::V10 onwards, use `check_solution` instead of `check_solution_mut`.
+                        let res = if N::CONSENSUS_VERSION(self.latest_height()).unwrap() >= ConsensusVersion::V10 {
+                            self.puzzle().check_solution(solution, latest_epoch_hash, latest_proof_target)
```
raychu86, line 402: Does this not reintroduce the forking risk?

Where a malicious validator forcibly changes the target in some of the solutions and keeps the correct one in the original. Then at block production, some would accept and some would abort the solution.

@vicsn

vicsn: We introduced check_solutions_mut before we had transmission checksum. And I believe it was technically not needed anymore ever since we did introduce transmission checksums.

Reason for changing this now, is to make it clear why we're not carrying around accepted_solution_checksums in the CompactBlock: because we can recompute the checksum from solutions.

It does not hurt for us to add a unit test that a solution with false target will be aborted.

ledger/block/src/serialize.rs
```diff
@@ -20,16 +20,46 @@ impl<N: Network> Serialize for Block<N> {
     fn serialize<S: Serializer>(&self, serializer: S) -> Result<S::Ok, S::Error> {
         match serializer.is_human_readable() {
             true => {
-                let mut block = serializer.serialize_struct("Block", 9)?;
+                let current_consensus_version = N::CONSENSUS_VERSION(self.height()).unwrap(); // TODO: remove unwrap.
```
ljedrz: @vicsn this `unwrap` is fine, as the related collection is always populated

console/network/src/consensus_heights.rs
ljedrz: I don't think so, the way the blocks are constructed depends on the block height (which the consensus version is derived from).

vicsn: Can you revert this change so I can have a look at which tests require this?

We can either have tests which require this be run with a separate CI job, compiled with a custom `CONSENSUS_VERSION_HEIGHTS`. But first I'll see if we can consider refactoring or discarding these tests.

ljedrz: done in https://github.com/ProvableHQ/snarkVM/pull/2841/commits/a2ad9a728665e56c1eff40714329872b86317404

ledger/block/src/lib.rs
vicsn: > I think prior_* is a bit too ambiguous. Our current usage of aborted_or_existing_transaction_ids is also quite verbose, but I'd still lean more towards verbosity for the sake of clarity.
> Other options could be:
> existing_*
> past_*

I'm happy to make things verbose if it makes things more clear, and excuse this non-native speaker commenting on this, but prior is a synonym of previous and past right?

ledger/block/src/bytes.rs
```diff
@@ -42,32 +42,78 @@ impl<N: Network> FromBytes for Block<N> {
...
-        }
-        // Read the aborted solution IDs.
-        let mut aborted_solution_ids = Vec::with_capacity(num_aborted_solutions as usize);
-        for _ in 0..num_aborted_solutions {
-            aborted_solution_ids.push(FromBytes::read_le(&mut reader)?);
-        }
+        let (prior_solution_transmission_ids, aborted_solution_transmission_ids, aborted_solution_ids) = if version >= 2
+        {
+            // Read the number of prior solution transmission IDs.
+            let num_prior_solutions = u32::read_le(&mut reader)?; // TODO: what is a sane bound on this value?
```
vicsn: You can use `max_aborted_solutions` again

ledger/block/src/bytes.rs
```diff
@@ -138,6 +231,7 @@ mod tests {
         let rng = &mut TestRng::default();
 
         for expected in [crate::test_helpers::sample_genesis_block(rng)].into_iter() {
+            // todo: sample compact too for the Block tests
```
vicsn: Currently Quorum Blocks (which we use in tests) can only be created as Full Blocks.

So consider *converting* another quorum block to compact block so we can test both the full and compact serialization.

ledger/block/src/lib.rs
```diff
@@ -77,10 +80,20 @@ pub struct Block<N: Network> {
     ratifications: Ratifications<N>,
     /// The solutions in the block.
     solutions: Solutions<N>,
+    /// The transmission IDs of both accepted and aborted solutions in the previous block.
+    /// Technically prior checksums can be retrieved from the ledger, but that is unwieldy.
+    prior_solution_transmission_ids: Option<Vec<TransmissionID<N>>>,
+    /// The aborted solution transmission IDs in this block.
+    aborted_solution_transmission_ids: Option<Vec<TransmissionID<N>>>,
     /// The aborted solution IDs in this block.
     aborted_solution_ids: Vec<SolutionID<N>>,
```
vicsn: aborted_solution_ids and aborted_transaction_ids are now also optional (none from v2 onwards)

ledger/block/src/lib.rs
```diff
@@ -259,11 +289,73 @@ impl<N: Network> Block<N> {
...
+            .iter()
+            .map(|tx| {
+                let checksum = Data::<Transaction<N>>::Buffer(tx.to_bytes_le()?.into()).to_checksum::<N>()?;
+                Ok(TransmissionID::Transaction(tx.id(), checksum))
+            })
+            .collect::<Result<Vec<_>>>()?;
+        // Compute the solution transmission IDs.
+        let solution_transmission_ids = match solutions.as_puzzle_solutions() {
+            Some(solutions) => {
+                let mut transmission_ids = Vec::new(); //with_capacity(solutions.solution_ids().len());
```
vicsn: Preallocate properly

ledger/block/src/serialize.rs
```diff
@@ -43,18 +73,54 @@ impl<'de, N: Network> Deserialize<'de> for Block<N> {
...
                     DeserializeExt::take_from_value::<D>(&mut block, "solutions")?,
-                    DeserializeExt::take_from_value::<D>(&mut block, "aborted_solution_ids")?,
+                    prior_solution_transmission_ids,
+                    aborted_solution_transmission_ids,
+                    aborted_solution_ids,
                     DeserializeExt::take_from_value::<D>(&mut block, "transactions")?,
-                    DeserializeExt::take_from_value::<D>(&mut block, "aborted_transaction_ids")?,
+                    prior_transaction_transmission_ids,
+                    aborted_transaction_transmission_ids,
+                    aborted_transaction_ids,
```
vicsn, line 123: TODO: also add a compact block test below

ledger/block/src/verify.rs
```diff
@@ -225,16 +233,24 @@ impl<N: Network> Block<N> {
...
+                        subdag,
+                        &self.solutions,
+                        &self.aborted_solution_ids,
+                        &self.transactions,
+                        &self.aborted_transaction_ids,
+                    )?,
+                    Subdag::Compact { subdag } => Self::check_compact_subdag_transmissions(
+                        subdag,
+                        &self.solutions,
+                        self.prior_solution_transmission_ids.as_ref().unwrap(),
```
vicsn: Please replace all the transmission unwraps by clean error handling. We can keep the consensus version unwrap.

ledger/block/src/verify.rs
```diff
@@ -558,32 +563,40 @@ impl<N: Network> Block<N> {
...
+        let mut seen_solution_ids: HashSet<SolutionID<N>> = HashSet::new();
 
         // Initialize a set of aborted or already-existing solution IDs.
-        let mut aborted_or_existing_solution_ids = HashSet::new();
+        let mut aborted_or_existing_solution_ids: HashSet<SolutionID<N>> = HashSet::new();
         // Initialize a set of aborted or already-existing transaction IDs.
-        let mut aborted_or_existing_transaction_ids = HashSet::new();
+        let mut aborted_or_existing_transaction_ids: HashSet<N::TransactionID> = HashSet::new();
+
+        let transmission_ids = subdag
```
vicsn: We don't have to collect anymore, just iterate over the references.

ledger/block/src/verify.rs
```diff
@@ -558,32 +563,40 @@ impl<N: Network> Block<N> {
...
+            .values()
+            .flatten()
+            .flat_map(|cert| cert.transmission_ids())
+            .copied()
+            .collect::<Vec<TransmissionID<N>>>();
+
+        // Prepare an iterator over the solution IDs.
```
vicsn: Move back up like the original code.

ledger/narwhal/batch-certificate/src/lib.rs
```diff
@@ -42,7 +43,7 @@ pub struct BatchCertificate<N: Network> {
     /// The batch header.
     batch_header: BatchHeader<N>,
     /// The signatures for the batch ID from the committee.
-    signatures: IndexSet<Signature<N>>,
+    pub signatures: IndexSet<Signature<N>>,
```
vicsn: Can you make an `into_components`/`into_inner` function instead?

ledger/narwhal/compact-certificate/Cargo.toml
```diff
@@ -0,0 +1,64 @@
+[package]
+name = "snarkvm-ledger-narwhal-compact-certificate"
+version = "4.0.0"
```
vicsn: Bump the versions.

ledger/narwhal/compact-header/src/lib.rs
```diff
@@ -0,0 +1,435 @@
...
+    pub const fn previous_certificate_ids(&self) -> &IndexSet<Field<N>> {
+        &self.previous_certificate_ids
+    }
+
+    /// Returns the signature.
+    pub const fn signature(&self) -> &Signature<N> {
+        &self.signature
+    }
+
+    pub fn to_transmission_ids<'a>(
```
vicsn, line 232: Docblock

ledger/narwhal/compact-header/src/lib.rs
```diff
@@ -0,0 +1,435 @@
...
+        ensure!(
+            transmission_ids.len() == self.transaction_indices.len() + self.solution_indices.len(),
+            "Internal logic error: could not find all transmission_ids."
+        );
+
+        Ok(transmission_ids)
+    }
+
+    /// Convert compact header to batch header
+    /// NOTE: this also recomputes the batch_id and verifies the signature.
```
vicsn: Remove this line

ledger/narwhal/compact-header/src/lib.rs
```diff
@@ -0,0 +1,435 @@
...
+        let transmission_ids = self.to_transmission_ids(
+            solutions,
+            prior_solutions,
+            aborted_solutions,
+            transactions,
+            prior_transactions,
+            aborted_transactions,
+        )?;
+
+        BatchHeader::from(
```
vicsn: Use `from_unchecked`. We already verify the `batch_id` in `verify_authority` which is called in `check_next_block` which is called on every incoming block. We already verify the signatures when deserializing the compact certificate.

ledger/narwhal/subdag/src/bytes.rs
```diff
@@ -43,37 +50,84 @@ impl<N: Network> FromBytes for Subdag<N> {
...
     }
 }
 
 impl<N: Network> ToBytes for Subdag<N> {
     /// Writes the subdag to the buffer.
     fn write_le<W: Write>(&self, mut writer: W) -> IoResult<()> {
         // Write the version.
-        1u8.write_le(&mut writer)?;
+        let version = 1u8;
+        version.write_le(&mut writer)?;
```
vicsn: Seems superfluous

ledger/narwhal/subdag/src/bytes.rs
```diff
@@ -43,37 +50,84 @@ impl<N: Network> FromBytes for Subdag<N> {
...
+        // Get the subdag type
+        let (subdag_type, subdag_len) = match self {
```
vicsn: TODO: also try compact deserialization in test below

ledger/narwhal/subdag/src/lib.rs
```diff
@@ -160,53 +315,188 @@ impl<N: Network> Subdag<N> {
...
+            Self::Full { subdag } => {
+                subdag.iter().map(|(round, certificates)| (certificates.len(), *round)).collect_vec()
+            }
+            Self::Compact { subdag } => {
+                subdag.iter().map(|(round, certificates)| (certificates.len(), *round)).collect_vec()
+            }
+        }
     }
 
+    // /// Returns certificates in this subdag (from earliest round to latest round).
```
vicsn: Can this be removed?

ledger/narwhal/subdag/src/lib.rs
```diff
@@ -160,53 +315,188 @@ impl<N: Network> Subdag<N> {
...
+        &self,
+        round: u64,
+        certificate_id: &Field<N>,
+    ) -> Option<Cow<'_, CompactCertificate<N>>> {
+        match self {
+            Self::Compact { subdag } => {
+                match subdag.get(&round) {
+                    Some(certificates) => {
+                        // Retrieve the certificate for the given certificate ID.
+                        match certificates.iter().find(|certificate| certificate.id() == *certificate_id) {
```
vicsn: map instead of match

ledger/narwhal/subdag/src/serialize.rs
```diff
@@ -36,8 +43,19 @@ impl<'de, N: Network> Deserialize<'de> for Subdag<N> {
...
+                let subdag_is_full = match value.get("subdag") {
+                    Some(..) => true,
+                    None => false,
+                };
+                match subdag_is_full {
+                    true => Ok(Self::from_full(DeserializeExt::take_from_value::<D>(&mut value, "subdag")?)
+                        .map_err(de::Error::custom)?),
+                    false => {
+                        Ok(Self::from_compact(DeserializeExt::take_from_value::<D>(&mut value, "compact_subdag")?)
+                            .map_err(de::Error::custom)?)
```
vicsn, line 56: Test compact subdag deserialization at the bottom

ledger/src/advance.rs
```diff
@@ -19,35 +19,181 @@ impl<N: Network, C: ConsensusStorage<N>> Ledger<N, C> {
     /// Returns a candidate for the next block in the ledger, using a committed subdag and its transmissions.
     pub fn prepare_advance_to_next_quorum_block<R: Rng + CryptoRng>(
         &self,
-        subdag: Subdag<N>,
-        transmissions: IndexMap<TransmissionID<N>, Transmission<N>>,
+        mut subdag: Subdag<N>,
+        subdag_transmissions: SubdagTransmissions<N>,
```
vicsn, line 23: This function is hella ugly now but we can create a new issue for cleaning it up

ledger/src/advance.rs
```diff
@@ -19,35 +19,181 @@ impl<N: Network, C: ConsensusStorage<N>> Ledger<N, C> {
     /// Returns a candidate for the next block in the ledger, using a committed subdag and its transmissions.
     pub fn prepare_advance_to_next_quorum_block<R: Rng + CryptoRng>(
         &self,
-        subdag: Subdag<N>,
-        transmissions: IndexMap<TransmissionID<N>, Transmission<N>>,
+        mut subdag: Subdag<N>,
```
vicsn: Just for clarity: Let's not make it mutable, we can conditionally use the entire new object.

vicsn (changes requested):

ledger/narwhal/subdag/src/lib.rs
```diff
@@ -50,39 +108,64 @@ fn is_sequential<T>(map: &BTreeMap<u64, T>) -> bool {
...
-    }
-    // Return `true` if the subdag matches the commit.
-    &commit == subdag
+        // Return `true` if the subdag matches the commit.
+        &commit == $subdag
+    }};
+}
+
+/// Performs sanity checks on the subdag.
+macro_rules! sanity_check_subdag {
```
kaimast, line 149: Does this have to be a macro, or could we also write something like the following?

```suggestion
fn sanity_check_subdag<T>(subdag BTreeMap<u64, IndexSet<T>>) -> Result<()> {
```

I always try to avoid macros as they harder to read, at least for me.

ledger/narwhal/subdag/src/lib.rs
```diff
@@ -116,39 +199,111 @@ fn weighted_median(timestamps_and_stake: Vec<(i64, u64)>) -> i64 {
...
+        }
+
+        Ok(())
+    }
+
+    /// Returns the subdag with full batch certificates.
+    pub fn into_full(
+        self,
+        solutions: Vec<TransmissionID<N>>,
+        prior_solutions: Vec<TransmissionID<N>>,
```
kaimast: Can you pass slices for the argument here, e.g.,`&[TransmissionID<N>]`, or use `into_iter` later to avoid copying?

ledger/narwhal/compact-header/src/lib.rs
```diff
@@ -0,0 +1,435 @@
...
+        aborted_solutions: impl ExactSizeIterator<Item = &'a TransmissionID<N>>,
+        transactions: impl Iterator<Item = &'a TransmissionID<N>>,
+        prior_transactions: impl ExactSizeIterator<Item = &'a TransmissionID<N>>,
+        aborted_transactions: impl Iterator<Item = &'a TransmissionID<N>>,
+    ) -> Result<IndexSet<TransmissionID<N>>> {
+        // Insert the transactions into the transmission_ids.
+        let mut transmission_ids = IndexSet::new();
+        transactions.chain(aborted_transactions).chain(prior_transactions).enumerate().try_for_each(
+            |(index, transmission_id)| {
+                if self.transaction_indices.contains(&u32::try_from(index)?) {
```
kaimast, line 245: Does this preserve the order? To me, it reads like the transactions are added in the order they are in the block not the order of transaction_indices.

ledger/narwhal/compact-header/src/lib.rs
kaimast: It would be good to know which of these arguments are from the compact block vs. the compact certificate. 
My understanding is they are all from the block, but I am not 100% sure.

ledger/narwhal/compact-header/Cargo.toml
```diff
@@ -0,0 +1,54 @@
...
+  "zero-knowledge"
+]
+categories = [
+  "compilers",
+  "cryptography",
+  "mathematics",
+  "wasm",
+  "web-programming"
+]
+include = [ "Cargo.toml", "src", "README.md", "LICENSE.md" ]
```
kaimast, line 22: I am curious if we could do `../../../LICENSE.md` for these in the future

ledger/narwhal/Cargo.toml
```diff
@@ -27,37 +27,49 @@ edition = "2024"
 default = [
   "batch-certificate",
   "batch-header",
+  "compact-certificate",
```
kaimast, line 30: Does it make sense to have all these feature flags? Mabye @vicsn knows if there are instances where we don't use the default features?

ledger/narwhal/traits/src/lib.rs
```diff
@@ -0,0 +1,46 @@
...
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+use indexmap::IndexSet;
+use snarkvm_console::{
+    program::{Network, Signature},
+    types::{Address, Field},
+};
+
+pub trait NarwhalCertificate<N: Network>: Send + Sync {
```
kaimast, line 24: Add a doccomment

ledger/narwhal/traits/Cargo.toml
```diff
@@ -0,0 +1,30 @@
+[package]
+name = "snarkvm-ledger-narwhal-traits"
+version = "4.0.0"
+authors = [ "The Aleo Team <hello@aleo.org>" ]
+description = "A batch certificate for a Narwhal-style memory pool in a decentralized virtual machine"
+homepage = "https://aleo.org"
+repository = "https://github.com/AleoHQ/snarkVM"
```
kaimast: This should be `ProvableHQ`, right?

ledger/narwhal/traits/Cargo.toml
```diff
@@ -0,0 +1,30 @@
...
+]
+include = [ "Cargo.toml", "src", "README.md", "LICENSE.md" ]
+license = "Apache-2.0"
+edition = "2024"
+
+[dependencies.snarkvm-console]
+workspace = true
+
+[dependencies.indexmap]
+version = "2.0"
```
kaimast: Isn't there a workspace dependency for indexmap?

ledger/block/src/lib.rs
```diff
@@ -77,12 +80,22 @@ pub struct Block<N: Network> {
...
     /// The solutions in the block.
     solutions: Solutions<N>,
+    /// The transmission IDs of both accepted and aborted solutions in the previous block.
+    /// Technically prior checksums can be retrieved from the ledger, but that is unwieldy.
+    prior_solution_transmission_ids: Option<Vec<TransmissionID<N>>>,
+    /// The aborted solution transmission IDs in this block.
+    aborted_solution_transmission_ids: Option<Vec<TransmissionID<N>>>,
     /// The aborted solution IDs in this block.
-    aborted_solution_ids: Vec<SolutionID<N>>,
+    aborted_solution_ids: Option<Vec<SolutionID<N>>>,
```
kaimast, line 90: Can you document when this can be `None`? 

Also, is it always the case that if `aborted_solution_ids` is `Some` `aborted_solution_transmission_ids` is also Some? If so, maybe they should be contained in the same `Option`?

ledger/block/src/bytes.rs
ljedrz: This (and the same applies to the `serde` one) is tricky, as we currently don't sample random blocks, only create new ones based on a sample genesis, which is not quorum-based (which would involve a subdag). Swapping out the authority is unsupported by the API, and would also likely break due to genesis checks.

However, I don't think this serialization check is needed, since the authority deserializes without any special treatment - it goes straight to the implementations for the `Subdag`, and those now test both the compact and the full version.

If anything, it would be nice to be able to test the serialization involving the presence of the new optional fields, but that's also tricky due to both the required block height and how ("manually") the blocks are sampled.

vicsn: Sounds good, feel free to resolve and remove the todo

ledger/block/src/bytes.rs
```diff
@@ -42,32 +42,81 @@ impl<N: Network> FromBytes for Block<N> {
...
-        }
-        // Read the aborted transaction IDs.
-        let mut aborted_transaction_ids = Vec::with_capacity(num_aborted_transactions as usize);
-        for _ in 0..num_aborted_transactions {
-            aborted_transaction_ids.push(FromBytes::read_le(&mut reader)?);
-        }
+        let (prior_transaction_transmission_ids, aborted_transaction_transmission_ids, aborted_transaction_ids) =
+            if version >= 2 {
+                // Read the number of prior transaction transmission IDs.
+                let num_prior_transactions = u32::read_le(&mut reader)?; // TODO: what is a sane bound on this value?
```
vicsn: Lingering todo, you can use max_aborted_transactions

vicsn (approved):
One more nit, otherwise LGTM!

kaimast (changes requested):
There are a few comments I left a while ago. Hopefully we can merge this soon

ledger/narwhal/subdag/src/lib.rs
ljedrz: not my go-to either, but at least this one is rather simple; I tried to change it into a function again now, but wasn't able to

ledger/narwhal/subdag/src/lib.rs
ljedrz: good catch, seems perfectly doable :+1:

ledger/narwhal/compact-header/Cargo.toml
ljedrz: I've always been a fan of such an approach, but IANAL

ledger/narwhal/compact-header/src/lib.rs
ljedrz: @vicsn can you please advise here? I'm not sure

ledger/block/src/lib.rs
ljedrz: I added extra docs; as for improved structure, it would indeed be best to even have something like an
```
enum AssociatedIDs {
    FullSubdag { aborted_solution_ids, aborted_transaction_ids },
    CompactSubdag { prior_solution_transmission_ids, aborted_solution_transmission_ids, prior_transaction_transmission_ids, aborted_transaction_transmission_ids },
}
```
but I'd recommend leaving it for a follow-up, as it is functionally the same, and this PR is a bitrot minefield already.

ledger/narwhal/compact-header/src/lib.rs
vicsn: Hmm yeah I see that's a bit confusing. Though note how we're using the same logic in `CompactHeader::new` to *create* the indices. This is quite brittle/unclear, so I think after this PR we could consolidate a bit (see comment in PR message about a better/new `SubdagTransmissions` abstraction.

ledger/narwhal/Cargo.toml
vicsn: I think it would be a chore to figure out, and we can tackle that separate when we're focusing on improving the dependency/feature tree to improve build times, starting with what gives us the biggest bang for our buck, and maybe we'll find automated tools to do most of the work for us.

kaimast:
Can you run benchmarks on the PR once it is rebased (or staging has been merged into it)? In case you do not have permissions to trigger them from the "Actions" tab, ping me and I will do it.

kaimast:
After more discussion, we decided to not merge compact subDAGs right now. I will leave #3000 open as a draft and close this PR, and we can revisit it in the future.

PR #2843, opened 2025-08-04
feat: handle is_genesis edge case

### LINKED ISSUE #2795: [Bug] `Block::is_genesis` behaves unexpectedly on inconsistent blocks
## 🐛 Bug Report

In `ledger/block/src/genesis.rs`, the code checks for a block being the genesis block like this:
```rust
      self.previous_hash == N::BlockHash::default()
            // Ensure the header is a genesis block header.
            && self.header.is_genesis()
            // Ensure the genesis authority is a beacon.
            && self.authority.is_beacon()
            // Ensure there are no solutions in the genesis block.
            && self.solutions.is_empty()
            // Ensure there is the correct number of accepted transaction in the genesis block.
            && self.transactions.num_accepted() == Self::NUM_GENESIS_TRANSACTIONS
            // Ensure there is the correct number of rejected transaction in the genesis block.
            && self.transactions.num_rejected() == 0
            // Ensure there is the correct number of finalize operations in the genesis block.
            && self.transactions.num_finalize() == 2 * Self::NUM_GENESIS_TRANSACTIONS
            // Ensure there are no aborted transaction IDs in the genesis block.
            && self.aborted_transaction_ids.is_empty()
```

This means that a block with a genesis header can still return `false` for `Block::is_genesis`. That just seems wrong.

If the block is at height 0 but not the genesis block, the storage is most likely corrupted and snarkVM should not continue as usual. A good solution would be to return `Result<bool>` here instead.

varun-doshi:
## Motivation

Fixes #2795 

## Test Plan

Test `test_is_genesis_fallible` is added to check if the updated function returns an err correctly

### DIFF
--- a/ledger/block/src/genesis.rs
+++ b/ledger/block/src/genesis.rs
@@ -19,26 +19,46 @@ impl<N: Network> Block<N> {
     /// Specifies the number of genesis transactions.
     pub const NUM_GENESIS_TRANSACTIONS: usize = 4;
 
-    /// Returns `true` if the block is a genesis block.
-    pub fn is_genesis(&self) -> bool {
+    /// Returns `Ok(true)` if the block is a genesis block.
+    pub fn is_genesis(&self) -> Result<bool> {
         // Ensure the previous block hash is zero.
-        self.previous_hash == N::BlockHash::default()
-            // Ensure the header is a genesis block header.
-            && self.header.is_genesis()
-            // Ensure the genesis authority is a beacon.
-            && self.authority.is_beacon()
-            // Ensure there is the correct number of ratification operations in the genesis block.
-            && self.ratifications.len() == 1
-            // Ensure there are no solutions in the genesis block.
-            && self.solutions.is_empty()
-            // Ensure there is the correct number of accepted transaction in the genesis block.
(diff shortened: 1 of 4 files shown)


vicsn:
Thank you for the change! Another current PR in snarkVM is refactoring this code, so not merging this one in yet unfortunately

ljedrz:
Just for reference: https://github.com/ProvableHQ/snarkVM/pull/2813 is the aforementioned PR.

varun-doshi:
> Just for reference: #2813 is the aforementioned PR.

Thank you for clarifying
I can merge/rebase my PR once the refactor PR is in

kaimast:
I think this can be closed now. Feel free to compare your PR with the most recent changes to `staging` to verify. 

Again, thanks for your contribution, even if we (probably) won't merge it.

varun-doshi:
> I think this can be closed now. Feel free to compare your PR with the most recent changes to `staging` to verify.
> 
> Again, thanks for your contribution, even if we (probably) won't merge it.

Just to clarify
Should I rebase to get latest changes or is there something wrong with this pr?

vicsn:
The issue at hand is already fixed. New PRs tackling new issues are always welcome.

PR #2845, opened 2025-08-07
Reduce `canary` and `testnet` stake amounts

raychu86:
## Motivation

This PR reduces the stake requirements per solution for `canary` and `testnet` to be 1% of the `mainnet` requirements. The intention is to assist with testing on these networks without having to request large quantities of credits from faucets.

I'm not a fan of this change as it makes this code more confusing to parse and adds yet another surfaces that differentiate network values that we need track. However, I do see how it may be needed to alleviate extraneous operational strain. 

We may be able to further clean this up in the future by making these values live in each `Network` declaration

Also note that this change will require `canary` and `testnet` networks to perform a synchronous reset, otherwise there may be a forking risk.

### DIFF
--- a/ledger/src/is_solution_limit_reached.rs
+++ b/ledger/src/is_solution_limit_reached.rs
@@ -15,15 +15,8 @@
 
 use super::*;
 
-/// The stake required to land one solution per epoch at various points in time.
-///
-/// Each entry represents a threshold where, starting from the given timestamp,
-/// a prover must have at least the specified amount of stake (in microcredits) to land one solution.
-///
-/// A prover with `n * stake` may land up to `n` solutions per epoch.
-///
-/// Format: `(timestamp, stake_required_per_solution)`
-pub const STAKE_REQUIREMENTS_PER_SOLUTION: [(i64, u64); 9] = [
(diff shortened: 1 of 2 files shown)


vicsn (changes requested):
Thank you for the change!

Historically, the ecosystem keeps having problems with faucets. We cannot expect to distribute these amounts easily. Please lower the canary/testnet requirements to incrementing single digit microcredits (1,2,3, etc.). Otherwise this PR is useless and does not provide a tangible UX benefit for testing.

To devil's advocate myself: the benefits I see to use higher values are:
- DoS prevention, which is not important at this stage on canary/testnet, and there are other more pressing attack vectors.
- To conceptually stay close to the mainnet setup to increase assurance. But those efforts can be better placed on e.g. defense in depth monitoring on the network whether the expected amount of solutions land.

I like the idea of moving this to the Network trait, but don't feel too strongly about it because there's no urgent need, so it risks being a useless refactor which may be reverted in the future when an urgent need comes up.

vicsn (approved):
Approving because 10k credits is the minimum staking requirement

PR #2848, opened 2025-08-11
[Chore] Migrate benchmarks to newer rust-toolchain action

kaimast:
`actions-rs/toolchain` has been unmaintained for at least three years. See [this issue](https://github.com/actions-rs/toolchain/issues/216). 

This PR also adds `workflow_dispatch` as a trigger for this workflow, so we can run it, if needed, on branches like this.

### DIFF
--- a/.github/workflows/benchmarks.yml
+++ b/.github/workflows/benchmarks.yml
@@ -4,6 +4,8 @@ on:
   push:
     branches:
       - 'staging'
+      - 'bench/rust-toolchain'
+  workflow_dispatch:
 
 jobs:
   # Run benchmarks and stores the output to a file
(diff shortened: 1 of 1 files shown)


ljedrz (approved):
LGTM, though I'm not sure if we shouldn't restrict the number of benchmarks being run (there may be some restrictions).

kaimast:
> LGTM, though I'm not sure if we shouldn't restrict the number of benchmarks being run (there may be some restrictions).

Good point. I am not sure if there is a way to do this? We could maybe change it to trigger the benchmark runs once a day on staging instead of after every commit.

(Also, I wanted to remove the trigger for `bench/rust-toolchain` from the file before merging, but it should not cause any issues)

ljedrz:
Not sure either, there's mention of [scaling](https://docs.github.com/en/actions/reference/limits), but not really of on-demand toggling.

Issue #2849, opened 2025-08-12
[Proposal] Let dev_println use trace logs


vicsn:
## 💥 Proposal

Currently, `dev_println` enabled logs are just printed to stdout (and I assume dev_eprintln to stderr).

We could instead use the trace library, so such logs are easily uniformly retrievable in tmux history and log files.

Is there any downside to doing this?

kaimast:
I would prefer this too. Having `dev_println` enabled for `test_network` in snarkOS creates a lot of output that cannot be hidden.

My guess is the reason for a dedicated `dev_println` function is so the output shows up in frontend tools that do not use logging frameworks, such as `leo`.

PR #2850, opened 2025-08-12
Improved varuna logging

vicsn:
## Motivation

The initial logging in low level snarkVM was set up without having in mind the production snarkOS node of today. Over the past months I spent a decent amount of hours trying to figure out why things happened. As a guiding principle, I think logs should be both informative and unique (so it is easily linkable to a specific line of code).

While a few tactical cleanups could have sufficed, with the magic of AI I'd like to provide the following luxuries:

- 9a867b3951ab719e14409bb838eecfc7b567ff87 minor documentation fix (apologies for the messy commit)
- 8a12fc2fd0873104bcb2cfc8b7dfbd6b190a8f1d and 74226abc67486a01959e01ffe50ac7bc8e02e4d7 and 3950bff40 replace custom sometimes unclear errors by more verbose anyhow errors
- 99b7e182d05c3751f9b87c41f21ec19d2472b744 moves all debug printing under the same existing `dev_println` feature flag. In the future we should also replace these with `trace` statements: https://github.com/ProvableHQ/snarkVM/issues/2849
- 7fd0a430c remove unused dependency

## Test Plan

- [ ] Create a snarkOS PR which enabled `dev_println` if `dev_network` is enabled.


algorithms/src/polycommit/kzg10/mod.rs
```diff
@@ -89,7 +87,7 @@ pub struct KZG10<E: PairingEngine>(PhantomData<E>);
 impl<E: PairingEngine> KZG10<E> {
     /// Constructs public parameters when given as input the maximum degree
     /// `degree` for the polynomial commitment scheme.
-    pub fn load_srs(max_degree: usize) -> Result<UniversalParams<E>, PCError> {
+    pub fn load_srs(max_degree: usize) -> Result<UniversalParams<E>, anyhow::Error> {
```
Antonio95, line 90: Here and in the rest of his file: is there any reason why `Error`,  `bail!` and `anyhow!` are not imported with `use anyhow::...` and instead qualified with the full path (eg `anyhow::bail!`) at every use? I have tested the former and it does work.

algorithms/src/polycommit/kzg10/mod.rs
```diff
@@ -132,7 +130,8 @@ impl<E: PairingEngine> KZG10<E> {
 
         let mut randomness = KZGRandomness::empty();
         if let Some(hiding_degree) = hiding_bound {
-            let mut rng = rng.ok_or(PCError::MissingRng)?;
+            let mut rng =
+                rng.ok_or(anyhow::anyhow!("The provided polynomial was meant to be hiding, but `rng` was `None`"))?;
```
Antonio95, line 134: Nitpick: polynomials aren't hiding or otherwise, commitments are. Perhaps:

```suggestion
                rng.ok_or(anyhow::anyhow!("The polynomial commitment was meant to be hiding, but `rng` was `None`"))?;
```

algorithms/src/polycommit/kzg10/mod.rs
```diff
@@ -182,7 +184,8 @@ impl<E: PairingEngine> KZG10<E> {
...
```
Antonio95, line 188: Nitpick: Idem

```suggestion
                rng.ok_or(anyhow::anyhow!("The polynomial commitment was meant to be hiding, but `rng` was `None`"))?;
```

algorithms/src/polycommit/kzg10/mod.rs
```diff
@@ -432,20 +443,22 @@ impl<E: PairingEngine> KZG10<E> {
...
+                anyhow::bail!("The degree bound ({}) is not supported by the parameters", bound)
             } else if bound < p.degree() || bound > max_degree {
-                return Err(PCError::IncorrectDegreeBound {
-                    poly_degree: p.degree(),
-                    degree_bound: p.degree_bound().unwrap(),
-                    max_degree,
-                    label: p.label().to_string(),
-                });
+                anyhow::bail!(
+                    "The degree bound ({}) for the polynomial {} (having degree {}) is greater than the maximum degree ({})",
```
Antonio95, line 456: This message doesn't reflect the condition "bound < p.degree". Perhaps I'm missing something? If not, maybe: (or even better, splitting this `else` into two separate cases)
```suggestion
                    "The degree bound ({}) for the polynomial {} is less than its degree ({}) or greater than the maximum degree ({})",
```

algorithms/src/polycommit/sonic_pc/mod.rs
```diff
@@ -128,9 +128,11 @@ impl<E: PairingEngine, S: AlgebraicSponge<E::Fq, 2>> SonicKZG10<E, S> {
             .map(|(_k, v)| *v)
             .collect::<Vec<_>>();
         if powers_of_beta_times_gamma_g.len() != supported_hiding_bound + 2 {
-            return Err(
-                PCError::HidingBoundToolarge { hiding_poly_degree: supported_hiding_bound, num_powers: 0 }.into()
-            );
+            return Err(anyhow::anyhow!(
+                "The degree of the hiding poly ({}) is not less than the maximum number of powers in `Powers` ({})",
+                supported_hiding_bound,
+                0
```
Antonio95, line 134: Why is this 0?

algorithms/src/polycommit/sonic_pc/mod.rs
```diff
@@ -128,9 +128,11 @@ impl<E: PairingEngine, S: AlgebraicSponge<E::Fq, 2>> SonicKZG10<E, S> {
             .map(|(_k, v)| *v)
             .collect::<Vec<_>>();
         if powers_of_beta_times_gamma_g.len() != supported_hiding_bound + 2 {
-            return Err(
-                PCError::HidingBoundToolarge { hiding_poly_degree: supported_hiding_bound, num_powers: 0 }.into()
-            );
+            return Err(anyhow::anyhow!(
```
Antonio95, line 131: This syntax seems to work and is less verbose (the same applies to many places; perhaps a search of `return Err(anyhow` can find them all)
```suggestion
            anyhow::bail!(
```

algorithms/src/polycommit/kzg10/mod.rs
```diff
@@ -432,20 +443,22 @@ impl<E: PairingEngine> KZG10<E> {
         max_degree: usize,
         enforced_degree_bounds: Option<&[usize]>,
         p: impl Into<LabeledPolynomialWithBasis<'a, E::Fr>>,
-    ) -> Result<(), PCError> {
+    ) -> Result<()> {
         let p = p.into();
         if let Some(bound) = p.degree_bound() {
-            let enforced_degree_bounds = enforced_degree_bounds.ok_or(PCError::UnsupportedDegreeBound(bound))?;
+            let enforced_degree_bounds = enforced_degree_bounds
+                .ok_or_else(|| anyhow::anyhow!("The degree bound ({}) is not supported by the parameters", bound))?;
```
Antonio95, line 450: `.ok_or(anyhow::anyhow!` is more concise than `.ok_or_else(|| anyhow::anyhow!`. The latter does avoid executing the necessary code to produce the `Err` - is that why it is used? I have seen instances of the former, too. Is the guiding principle: "if the error message is a literal, use `ok_or`, if it involves formatting use `ok_or_else`"?. If so, why not switch to `ok_or_else` in the literal-only case too? Both for consistency and to avoid construct the `Err` object (which are minimal, but then again, so are probably the savings in the formatting case, and `ok_or_else` is used there; I could be wrong about this, though).

algorithms/src/polycommit/sonic_pc/mod.rs
```diff
@@ -246,7 +251,7 @@ impl<E: PairingEngine, S: AlgebraicSponge<E::Fq, 2>> SonicKZG10<E, S> {
                 Ok((LabeledCommitment::new(label.to_string(), comm, degree_bound), rand))
             });
         }
-        let results: Vec<Result<_, PCError>> = pool.execute_all();
+        let results: Vec<Result<_, anyhow::Error>> = pool.execute_all();
```
Antonio95, line 254: anyhow's Result is syntactic sugar precisely for this. The same likely applies to other places.
```suggestion
        let results: Vec<Result<_>> = pool.execute_all();
```

algorithms/src/polycommit/sonic_pc/mod.rs
```diff
@@ -340,10 +346,10 @@ impl<E: PairingEngine, S: AlgebraicSponge<E::Fq, 2>> SonicKZG10<E, S> {
                 proof
             });
         }
-        let batch_proof = pool.execute_all().into_iter().collect::<Result<_, _>>().map(BatchProof).map_err(Into::into);
+        let batch_proof = pool.execute_all().into_iter().collect::<Result<_, _>>().map(BatchProof)?;
```
Antonio95, line 349: Dramatic savings of three characters. Probably applies to other places too.
```suggestion
        let batch_proof = pool.execute_all().into_iter().collect::<Result<_>>().map(BatchProof)?;
```

algorithms/src/r1cs/test_constraint_checker.rs
```diff
@@ -73,9 +73,9 @@ impl<F: Field> TestConstraintChecker<F> {
 impl<F: Field> ConstraintSystem<F> for TestConstraintChecker<F> {
     type Root = Self;
 
-    fn alloc<Fn, A, AR>(&mut self, _annotation: A, f: Fn) -> Result<Variable, SynthesisError>
+    fn alloc<Fn, A, AR>(&mut self, _annotation: A, f: Fn) -> Result<Variable, anyhow::Error>
```
Antonio95, line 76: Conditioned on adding `use anyhow::Result;` to the file (the same applies to other places; all can be found with the **regexp** search for `Result<.*, anyhow::Error>,`)
```suggestion
    fn alloc<Fn, A, AR>(&mut self, _annotation: A, f: Fn) -> Result<Variable>
```

ljedrz:
Having multiple strongly-typed errors is quite useful in case the crate is primarily used as a library; do we know how much the more self-contained snarkVM crates are being used elsewhere? If the snarkOS is the only major user, then switching to `anyhow` makes a lot of sense.

Needless to say, this would technically call for a bump to the major version of the affected crates.

vicsn:
> Having multiple strongly-typed errors is quite useful in case the crate is primarily used as a library; do we know how much the more self-contained snarkVM crates are being used elsewhere? If the snarkOS is the only major user, then switching to anyhow makes a lot of sense.

Thank you for pointing this out.

There are a number of other users - principally wallets (via WASM) and their backend infra (delegated provers and record scanners). I doubt any of them special case on the strongly typed errors, but they will indeed need to migrate to handling anyhow.

A bigger concern I just thought of is that perhaps one day we'll want to expect certain errors in certain unit tests, something we have avoided doing until now. However, that should still be possible with anyhow's string errors - and after this initial painful migration, it'll arguably be easier to adjust error messages to sharpen the tests in the future.

> Needless to say, this would technically call for a bump to the major version of the affected crates.

👍 I think we can do this for snarkVM

algorithms/src/snark/varuna/ahp/indexer/indexer.rs
```diff
@@ -173,13 +173,19 @@ impl<F: PrimeField, SM: SNARKMode> AHPForR1CS<F, SM> {
...
 
-        let constraint_domain = EvaluationDomain::new(num_constraints).ok_or(SynthesisError::PolyTooLarge)?;
-        let variable_domain = EvaluationDomain::new(num_variables).ok_or(SynthesisError::PolyTooLarge)?;
-        let input_domain = EvaluationDomain::new(num_padded_public_variables).ok_or(SynthesisError::PolyTooLarge)?;
-
-        let non_zero_a_domain = EvaluationDomain::new(num_non_zero_a).ok_or(SynthesisError::PolyTooLarge)?;
-        let non_zero_b_domain = EvaluationDomain::new(num_non_zero_b).ok_or(SynthesisError::PolyTooLarge)?;
-        let non_zero_c_domain = EvaluationDomain::new(num_non_zero_c).ok_or(SynthesisError::PolyTooLarge)?;
+        let constraint_domain =
+            EvaluationDomain::new(num_constraints).ok_or_else(|| anyhow::anyhow!("Polynomial degree is too large"))?;
```
Antonio95, line 177: At several places in this file: another instance where `ok_or_else(|| anyhow::anyhow!` can be replaced by `ok_or(anyhow::anyhow!` (but cf. my earlier comment on this change).

algorithms/src/snark/varuna/ahp/prover/round_functions/fifth.rs
```diff
@@ -44,7 +44,7 @@ impl<F: PrimeField, SM: SNARKMode> AHPForR1CS<F, SM> {
         verifier_message: verifier::FourthMessage<F>,
         state: prover::State<'_, F, SM>,
         _r: &mut R,
-    ) -> Result<prover::FifthOracles<F>, AHPError> {
+    ) -> Result<prover::FifthOracles<F>, anyhow::Error> {
```
Antonio95, line 47: For consistency with other files, this should be replaced by `anyhow::Result<prover::FifthOracles<F>` (and then `use anyhow::Result` above so that the full qualification can be removed here). This applies to other files which can easily be found with a search - I will not comment of them each individually.

algorithms/src/snark/varuna/ahp/verifier/verifier.rs
```diff
@@ -93,29 +92,33 @@ impl<TargetField: PrimeField, SM: SNARKMode> AHPForR1CS<TargetField, SM> {
...
-            let variable_domain =
-                EvaluationDomain::new(circuit_info.num_public_and_private_variables).ok_or(AHPError::PolyTooLarge)?;
+            let variable_domain = EvaluationDomain::new(circuit_info.num_public_and_private_variables)
+                .ok_or_else(|| anyhow::anyhow!("Polynomial degree is too large"))?;
             end_timer!(variable_domain_time);
 
             let non_zero_a_time = start_timer!(|| format!("Constructing non-zero-a domain for {circuit_id}"));
-            let non_zero_a_domain = EvaluationDomain::new(circuit_info.num_non_zero_a).ok_or(AHPError::PolyTooLarge)?;
+            let non_zero_a_domain = EvaluationDomain::new(circuit_info.num_non_zero_a)
+                .ok_or_else(|| anyhow::anyhow!("Polynomial degree is too large"))?;
```
Antonio95, line 106: The original message is a bit misleading (and vague), since what's happening here is that the matrix has too many entries for it to be interpolated over one of the available domains. Also, and regardless of this, now that we're writing error messages directly it would be good if this case mentions the concrete matrix, which should be done in lines 111 and 116 too.
```suggestion
                .ok_or_else(|| anyhow::anyhow!("Matrix A has too many non-zero entries"))?;
```

algorithms/src/snark/varuna/ahp/verifier/verifier.rs
```diff
@@ -93,29 +92,33 @@ impl<TargetField: PrimeField, SM: SNARKMode> AHPForR1CS<TargetField, SM> {
...
             let non_zero_c_time = start_timer!(|| format!("Constructing non-zero-c domain for {circuit_id}"));
-            let non_zero_c_domain = EvaluationDomain::new(circuit_info.num_non_zero_c).ok_or(AHPError::PolyTooLarge)?;
+            let non_zero_c_domain = EvaluationDomain::new(circuit_info.num_non_zero_c)
+                .ok_or_else(|| anyhow::anyhow!("Polynomial degree is too large"))?;
             end_timer!(non_zero_c_time);
 
             let input_domain_time = start_timer!(|| format!("Constructing input domain {circuit_id}"));
-            let input_domain = EvaluationDomain::new(circuit_info.num_public_inputs).ok_or(AHPError::PolyTooLarge)?;
+            let input_domain = EvaluationDomain::new(circuit_info.num_public_inputs)
+                .ok_or_else(|| anyhow::anyhow!("Polynomial degree is too large"))?;
```
Antonio95, line 121: For the same reasons as the previous comment:
```suggestion
                .ok_or_else(|| anyhow::anyhow!("The R1CS has too many public inputs"))?;
```

algorithms/src/snark/varuna/ahp/ahp.rs
```diff
@@ -88,12 +87,12 @@ impl<F: PrimeField, SM: SNARKMode> AHPForR1CS<F, SM> {
...
     pub fn max_degree(num_constraints: usize, num_variables: usize, num_non_zero: usize) -> Result<usize> {
         let zk_bound = Self::zk_bound().unwrap_or(0);
-        let constraint_domain_size =
-            EvaluationDomain::<F>::compute_size_of_domain(num_constraints).ok_or(AHPError::PolyTooLarge)?;
-        let variable_domain_size =
-            EvaluationDomain::<F>::compute_size_of_domain(num_variables).ok_or(AHPError::PolyTooLarge)?;
-        let non_zero_domain_size =
-            EvaluationDomain::<F>::compute_size_of_domain(num_non_zero).ok_or(AHPError::PolyTooLarge)?;
+        let constraint_domain_size = EvaluationDomain::<F>::compute_size_of_domain(num_constraints)
+            .ok_or_else(|| anyhow::anyhow!("Polynomial degree is too large"))?;
```
Antonio95, line 91: Similarly:
```suggestion
            .ok_or_else(|| anyhow::anyhow!("The R1CS has too many constraints"))?;
```

algorithms/src/snark/varuna/ahp/ahp.rs
```diff
@@ -88,12 +87,12 @@ impl<F: PrimeField, SM: SNARKMode> AHPForR1CS<F, SM> {
...
+        let variable_domain_size = EvaluationDomain::<F>::compute_size_of_domain(num_variables)
```
Antonio95, line 92: Similarly:
```suggestion
            .ok_or_else(|| anyhow::anyhow!("The R1CS has too many variables"))?;
```

algorithms/src/snark/varuna/ahp/ahp.rs
```diff
@@ -88,12 +87,12 @@ impl<F: PrimeField, SM: SNARKMode> AHPForR1CS<F, SM> {
...
+            .ok_or_else(|| anyhow::anyhow!("Polynomial degree is too large"))?;
+        let non_zero_domain_size = EvaluationDomain::<F>::compute_size_of_domain(num_non_zero)
+            .ok_or_else(|| anyhow::anyhow!("Polynomial degree is too large"))?;
```
Antonio95, line 95:
```suggestion
            .ok_or_else(|| anyhow::anyhow!("The R1CS has too many non-zero entries"))?;
```

algorithms/src/snark/varuna/ahp/ahp.rs
```diff
@@ -117,20 +116,31 @@ impl<F: PrimeField, SM: SNARKMode> AHPForR1CS<F, SM> {
         let num_non_zero_b = info.num_non_zero_b;
         let num_non_zero_c = info.num_non_zero_c;
         Ok([
-            EvaluationDomain::<F>::compute_size_of_domain(num_variables).ok_or(SynthesisError::PolyTooLarge)? - 2,
-            EvaluationDomain::<F>::compute_size_of_domain(num_non_zero_a).ok_or(SynthesisError::PolyTooLarge)? - 2,
-            EvaluationDomain::<F>::compute_size_of_domain(num_non_zero_b).ok_or(SynthesisError::PolyTooLarge)? - 2,
-            EvaluationDomain::<F>::compute_size_of_domain(num_non_zero_c).ok_or(SynthesisError::PolyTooLarge)? - 2,
+            EvaluationDomain::<F>::compute_size_of_domain(num_variables)
+                .ok_or_else(|| anyhow::anyhow!("Polynomial degree is too large"))?
```
Antonio95, line 120:
```suggestion
                .ok_or_else(|| anyhow::anyhow!("The R1CS has too many variables"))?
```

algorithms/src/snark/varuna/ahp/ahp.rs
```diff
@@ -117,20 +116,31 @@ impl<F: PrimeField, SM: SNARKMode> AHPForR1CS<F, SM> {
...
+                - 2,
+            EvaluationDomain::<F>::compute_size_of_domain(num_non_zero_a)
+                .ok_or_else(|| anyhow::anyhow!("Polynomial degree is too large"))?
```
Antonio95, line 123:
```suggestion
                .ok_or_else(|| anyhow::anyhow!("The matrix A has too many non-zero entries"))?
```

algorithms/src/snark/varuna/ahp/ahp.rs
```diff
@@ -117,20 +116,31 @@ impl<F: PrimeField, SM: SNARKMode> AHPForR1CS<F, SM> {
...
+                - 2,
+            EvaluationDomain::<F>::compute_size_of_domain(num_non_zero_b)
+                .ok_or_else(|| anyhow::anyhow!("Polynomial degree is too large"))?
```
Antonio95, line 126:
```suggestion
                .ok_or_else(|| anyhow::anyhow!("The matrix B has too many non-zero entries"))?
```

algorithms/src/snark/varuna/ahp/ahp.rs
```diff
@@ -117,20 +116,31 @@ impl<F: PrimeField, SM: SNARKMode> AHPForR1CS<F, SM> {
...
+                - 2,
+            EvaluationDomain::<F>::compute_size_of_domain(num_non_zero_c)
+                .ok_or_else(|| anyhow::anyhow!("Polynomial degree is too large"))?
```
Antonio95, line 129:
```suggestion
                .ok_or_else(|| anyhow::anyhow!("The matrix C has too many non-zero entries"))?
```

algorithms/src/snark/varuna/ahp/ahp.rs
```diff
@@ -117,20 +116,31 @@ impl<F: PrimeField, SM: SNARKMode> AHPForR1CS<F, SM> {
...
 
     pub(crate) fn cmp_non_zero_domains(
         info: &CircuitInfo,
         max_candidate: Option<EvaluationDomain<F>>,
     ) -> Result<NonZeroDomains<F>> {
-        let domain_a = EvaluationDomain::new(info.num_non_zero_a).ok_or(SynthesisError::PolyTooLarge)?;
-        let domain_b = EvaluationDomain::new(info.num_non_zero_b).ok_or(SynthesisError::PolyTooLarge)?;
-        let domain_c = EvaluationDomain::new(info.num_non_zero_c).ok_or(SynthesisError::PolyTooLarge)?;
+        let domain_a = EvaluationDomain::new(info.num_non_zero_a)
+            .ok_or_else(|| anyhow::anyhow!("Polynomial degree is too large"))?;
```
Antonio95, line 139:
```suggestion
            .ok_or_else(|| anyhow::anyhow!("The matrix A has too many non-zero entries"))?;
```

algorithms/src/snark/varuna/ahp/ahp.rs
```diff
@@ -117,20 +116,31 @@ impl<F: PrimeField, SM: SNARKMode> AHPForR1CS<F, SM> {
...
+        let domain_b = EvaluationDomain::new(info.num_non_zero_b)
+            .ok_or_else(|| anyhow::anyhow!("Polynomial degree is too large"))?;
```
Antonio95, line 141:
```suggestion
            .ok_or_else(|| anyhow::anyhow!("The matrix B has too many non-zero entries"))?;
```

algorithms/src/snark/varuna/ahp/ahp.rs
```diff
@@ -117,20 +116,31 @@ impl<F: PrimeField, SM: SNARKMode> AHPForR1CS<F, SM> {
...
+        let domain_c = EvaluationDomain::new(info.num_non_zero_c)
+            .ok_or_else(|| anyhow::anyhow!("Polynomial degree is too large"))?;
```
Antonio95, line 143:
```suggestion
            .ok_or_else(|| anyhow::anyhow!("The matrix C has too many non-zero entries"))?;
```

algorithms/src/snark/varuna/data_structures/proof.rs
```diff
@@ -260,38 +260,50 @@ impl<E: PairingEngine> Proof<E> {
...
         let g_evals = [&self.evaluations.g_a_evals, &self.evaluations.g_b_evals, &self.evaluations.g_c_evals];
         for evals in g_evals {
             if evals.len() != self.batch_sizes.len() {
-                return Err(SNARKError::BatchSizeMismatch);
+                return Err(anyhow!("G evaluation size mismatch: {} != {}", evals.len(), self.batch_sizes.len()));
             }
         }
         if self.third_msg.sums.len() != self.batch_sizes.len() {
-            return Err(SNARKError::BatchSizeMismatch);
+            return Err(anyhow!(
```
Antonio95, line 290: `return Err(anyhow!(...` can be replaced by `bail!(...`. It is perhaps a matter of conciseness vs. readability, but `bail!` is used in other parts of the code, so probably one should be picked for consistency.

algorithms/src/snark/varuna/data_structures/test_circuit.rs
```diff
@@ -40,22 +41,24 @@ impl<F: Field> core::fmt::Debug for TestCircuit<F> {
 }
 
 impl<ConstraintF: Field> ConstraintSynthesizer<ConstraintF> for TestCircuit<ConstraintF> {
-    fn generate_constraints<CS: ConstraintSystem<ConstraintF>>(&self, cs: &mut CS) -> Result<(), SynthesisError> {
+    fn generate_constraints<CS: ConstraintSystem<ConstraintF>>(&self, cs: &mut CS) -> Result<(), anyhow::Error> {
```
Antonio95, line 44: I think I've mentioned this elsewhere, but this is the same as `Result<()>` (because the `Result` there is `anyhow::Result`, which, on one generic, has the error type set to `anyhow::Error`)

algorithms/src/snark/varuna/varuna.rs
```diff
@@ -716,12 +717,15 @@ where
...
 
             if public_inputs_i.is_empty() {
-                bail!(SNARKError::EmptyBatch);
+                bail!("Public input is empty.");
             }
 
-            if public_inputs_i.len() != batch_sizes_vec[i] {
-                bail!(SNARKError::BatchSizeMismatch);
-            }
+            ensure!(
```
Antonio95, line 723: I like this, but other parts of the code do this with `if (...) {bail!...`. One should be picked for consistency.

algorithms/src/snark/varuna/varuna.rs
```diff
@@ -783,11 +787,12 @@ where
         for (i, (vk, &batch_size)) in keys_to_inputs.keys().zip(batch_sizes.values()).enumerate() {
             inputs_and_batch_sizes.insert(vk.id, (batch_size, padded_public_vec[i].as_slice()));
         }
-        let max_constraint_domain =
-            EvaluationDomain::<E::Fr>::new(max_num_constraints).ok_or(SynthesisError::PolyTooLarge)?;
-        let max_variable_domain =
-            EvaluationDomain::<E::Fr>::new(max_num_variables).ok_or(SynthesisError::PolyTooLarge)?;
-        let max_non_zero_domain = max_non_zero_domain.ok_or(SynthesisError::PolyTooLarge)?;
+        let max_constraint_domain = EvaluationDomain::<E::Fr>::new(max_num_constraints)
+            .ok_or_else(|| anyhow::anyhow!("Polynomial degree is too large"))?;
```
Antonio95, line 791:
```suggestion
            .ok_or_else(|| anyhow::anyhow!("The R1CS has too many constraints"))?;
```

algorithms/src/snark/varuna/varuna.rs
```diff
@@ -783,11 +787,12 @@ where
...
+        let max_variable_domain = EvaluationDomain::<E::Fr>::new(max_num_variables)
+            .ok_or_else(|| anyhow::anyhow!("Polynomial degree is too large"))?;
```
Antonio95, line 793:
```suggestion
            .ok_or_else(|| anyhow::anyhow!("The R1CS has too many variables"))?;
```

algorithms/src/snark/varuna/varuna.rs
```diff
@@ -783,11 +787,12 @@ where
...
+        let max_non_zero_domain =
+            max_non_zero_domain.ok_or_else(|| anyhow::anyhow!("Polynomial degree is too large"))?;
```
Antonio95, line 795:
```suggestion
            max_non_zero_domain.ok_or_else(|| anyhow::anyhow!("The R1CS has too many non-zero elements"))?;
```

circuit/environment/src/helpers/assignment.rs
```diff
@@ -13,7 +13,7 @@
 // See the License for the specific language governing permissions and
 // limitations under the License.
 
-use crate::{Constraint, LinearCombination, Variable};
+use crate::{Constraint, LinearCombination, Variable, prelude::Result};
```
Antonio95, line 16: Is this imported this way for any particular reason here? It boils down to `anyhow::Result`, which is imported the second way across many files. The current line obscures that somewhat.

If there is a design reason, happy to leave it as it is.

circuit/environment/src/helpers/r1cs.rs
```diff
@@ -250,6 +250,7 @@ impl<F: PrimeField> R1CS<F> {
     /// it to the R1CS_HASHES collection.
     #[cfg(feature = "save_r1cs_hashes")]
     pub(crate) fn save_hash(&self) {
+        snarkvm_utilities::dev_println!("Saving hash of R1CS");
```
Antonio95, line 253: Cool macro :)

synthesizer/process/src/execute.rs
```diff
@@ -30,8 +30,7 @@ impl<N: Network> Process<N> {
         // Construct the locator.
         let locator = Locator::new(*request.program_id(), *request.function_name());
 
-        #[cfg(feature = "aleo-cli")]
-        println!("{}", format!(" • Executing '{locator}'...",).dimmed());
+        dev_println!("{}", format!(" • Executing '{locator}'...",));
```
Antonio95, line 33: This loses the `dimmed`, right?

Antonio95 (changes requested):
Many comments/suggestions for a PR with many changes!

- Most comments relate to the use of `anyhow` and, where I have seen dubious syntax, I have not repeated the same comment in other places where it occurred.
- A few of the suggestions are simply improved error messages.

I am happy to take over any of the changes I proposed, from none to all - just let me know. They should all be trivial to fix in any case.

I concur with the other comment in the PR that the switch to a single error type is good *if* this code is mostly used as an end application where the user's (or programmer's) main form of interaction with errors is seeing them displayed on a console or log. If this code is meant to be used as a dependency, however, custom error types could potentially allow for better error handling.

vicsn:
Replacing this by a more appetizable and directly useful PR: https://github.com/ProvableHQ/snarkVM/pull/2874

synthesizer/process/src/execute.rs
vicsn: Indeed it does. The reasoning for the dimmed was to really be sure no user would be alarmed by these messages, but I don't see a reason to gives these particular logs a different status than other informative logs.

FWIW I don't think anyone is using `aleo-cli` at the moment.

algorithms/src/polycommit/kzg10/mod.rs
ljedrz: I support the `anyhow::Error` notation due to it being more clear about what it is, but `anyhow!` or `bail!` are unmistakably from `anyhow`, so they could be imported

PR #2874, opened 2025-08-17
Improved varuna logging

vicsn:
## Motivation

- Clarifies some of the specific error logs which caused me to waste time debugging the past month.
- Unified the use of `dev_(e)println` over `(e)println`

## Test Plan

- [ ] I should run snarkOS once with the suggested changes to see if the spam is manageable.

## Related PRs

https://github.com/ProvableHQ/snarkOS/pull/3792

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -3772,6 +3772,8 @@ dependencies = [
  "snarkvm-algorithms",
  "snarkvm-circuit",
  "snarkvm-console",
+ "snarkvm-utilities",
+ "tracing",
 ]
 
 [[package]]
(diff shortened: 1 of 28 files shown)


raychu86 (approved):
LGTM, fairly innocent change.

Failing CI due to removal of `colorize` crate usage

kaimast:
The unused dependencies check fails, but otherwise it looks good to merge.

ljedrz (approved):
nit: some of these changes will cause the duration to be calculated, which should be fine (I don't think those are particularly hot paths), but could manifest as performance hits in related benchmarks

PR #2877, opened 2025-08-18
Update fee model (ARC-0005)

Roee-87:
## Motivation

This PR updates the fee calculation model in accordance with [ARC-0005](https://github.com/ProvableHQ/ARCs/discussions/92) by introducing a 25X cost reduction factor.
## Test Plan

Unit tests have been added to check that the new fee model computes the fee correctly.

ARC-0005 link:  https://github.com/ProvableHQ/ARCs/discussions/92


synthesizer/src/vm/deploy.rs
```diff
@@ -56,7 +56,13 @@ impl<N: Network, C: ConsensusStorage<N>> VM<N, C> {
         let owner = ProgramOwner::new(private_key, deployment_id, rng)?;
 
         // Compute the minimum deployment cost.
-        let (minimum_deployment_cost, _) = deployment_cost(&self.process().read(), &deployment)?;
+        // if consensus_version < ConsensusVersion::V9 {
```
Roee-87: I commented this section out because it was causing multiple tests to fail:
```bash
thread 'vm::test_helpers::test_multiple_deployments_and_multiple_executions' panicked at synthesizer/src/vm/mod.rs:1018:14:
called `Result::unwrap()` on an `Err` value: Deployments are not supported before consensus version V9
```

synthesizer/src/vm/execute.rs
```diff
@@ -1119,4 +1292,143 @@ finalize test:
         // Check that the finalize cost is equal to the expected cost.
         assert_eq!(finalize_cost, expected_cost);
     }
+
+    #[test]
+    #[ignore = "memory-intensive"]
+    fn test_deep_nested_execution_cost_v3() {
```
vicsn: You may remove this test.

synthesizer/src/vm/deploy.rs
vicsn: I'm not sure where these commented out lines even come from. You may remove.

synthesizer/src/vm/deploy.rs
```diff
@@ -56,7 +56,13 @@ impl<N: Network, C: ConsensusStorage<N>> VM<N, C> {
...
+        //     bail!("Deployments are not supported before consensus version V9");
+        // }
+        let (minimum_deployment_cost, _) = match consensus_version {
+            ConsensusVersion::V10 => deployment_cost_v2(&self.process().read(), &deployment)?,
+            _ => deployment_cost_v1(&self.process().read(), &deployment)?,
```
vicsn: This will only use `deployment_cost_v2` on V10, not for all versions afterwards.

Can you do:
```
                if (ConsensusVersion::V1..=ConsensusVersion::V9).contains(&consensus_version) { deployment_cost_v2(...) } else { deployment_cost_v2(...) } 
```

synthesizer/src/vm/execute.rs
```diff
@@ -80,9 +80,17 @@ impl<N: Network, C: ConsensusStorage<N>> VM<N, C> {
...
+                let (minimum_execution_cost, (_, _)) = match consensus_version {
+                    ConsensusVersion::V1 => execution_cost_v1(&self.process().read(), &execution)?,
+                    ConsensusVersion::V2
+                    | ConsensusVersion::V3
+                    | ConsensusVersion::V4
+                    | ConsensusVersion::V5
+                    | ConsensusVersion::V6
+                    | ConsensusVersion::V7
+                    | ConsensusVersion::V8
+                    | ConsensusVersion::V9 => execution_cost_v2(&self.process().read(), &execution)?,
```
vicsn: Nice we're encountering a variety we didn't hit before.

For now, can you try something like:
```
if consensus_version >= ConsensusVersion::V10 {
  // use cost_v3
} else if consensus_version >= ConsensusVersion::V2 {
  // use cost_v2
} else {
  // use cost_v1
}
```

In the future we'll introduce a utility function so there is just one way to standard way to specify things.

synthesizer/process/src/cost.rs
```diff
@@ -65,6 +113,26 @@ pub fn deployment_cost<N: Network>(
...
+pub fn execution_cost_v3<N: Network>(process: &Process<N>, execution: &Execution<N>) -> Result<(u64, (u64, u64))> {
+    // Compute the storage cost in microcredits.
+    let storage_cost = execution_storage_cost::<N>(execution.size_in_bytes()?);
+
+    // Get the root transition.
+    let transition = execution.peek()?;
+
+    // Get the finalize cost for the root transition.
+    let stack = process.get_stack(transition.program_id())?;
+    let finalize_cost = cost_in_microcredits_v2(&stack, transition.function_name())? / COST_REDUCTION_FACTOR;
```
vicsn: Since `cost_in_microcredits_v*` are very small helper functions, might be nice to create a new v3 one just for consistency and ease of future refactors.

Then within `cost_per_command` you can do:
```
    let mapping_base_cost = match consensus_fee_version {
        ConsensusFeeVersion::V1 => MAPPING_BASE_COST_V1,
        ConsensusFeeVersion::V2 | ConsensusFeeVersion::V3 => MAPPING_BASE_COST_V2,
    };
    let quotient = match consensus_fee_version {
        ConsensusFeeVersion::V1 | ConsensusFeeVersion::V2 => 1,
        ConsensusFeeVersion::V3 => COST_REDUCTION_FACTOR,
    };
```

synthesizer/process/src/cost.rs
```diff
@@ -22,8 +22,56 @@ use console::{
 use snarkvm_ledger_block::{Deployment, Execution, Transaction};
 use snarkvm_synthesizer_program::{CastType, Command, Instruction, Operand};
 
+// Cost reduction factor from ARC 0005.
+pub const COST_REDUCTION_FACTOR: u64 = 25;
```
vicsn:
```suggestion
pub const ARC_0005_COST_REDUCTION_FACTOR: u64 = 25;
```

synthesizer/src/vm/verify.rs
```diff
@@ -410,7 +414,18 @@ impl<N: Network, C: ConsensusStorage<N>> VM<N, C> {
                     if is_fee_required {
                         // We are using execution_cost_v2 to compute the execution cost.
                         // Using `execution_cost_v2` is fine as a default because it is strictly cheaper than or equivalent to `execution_cost_v1`.
-                        let (cost, (_, _)) = execution_cost_v2(&self.process().read(), execution)?;
+                        let (cost, (_, _)) = match consensus_version {
```
vicsn: See comment above about using an if-else block so its more readable.

synthesizer/src/vm/execute.rs
```diff
@@ -982,6 +996,165 @@ finalize test:
         assert_eq!(finalize_cost, expected_cost);
     }
 
+    #[test]
+    fn test_wide_nested_execution_cost_v3() {
```
vicsn: You can remove

vicsn (changes requested):

synthesizer/src/vm/execute.rs
Roee-87: done 
98d0530

synthesizer/src/vm/verify.rs
Roee-87: fixed: 
98d0530

synthesizer/src/vm/execute.rs
Roee-87: done 
98d0530

synthesizer/src/vm/deploy.rs
Roee-87: done 
98d0530

synthesizer/src/vm/execute.rs
Roee-87: removed: 
98d0530

synthesizer/src/vm/deploy.rs
Roee-87: removed 
98d0530

synthesizer/process/src/cost.rs
Roee-87: done [98d0530](https://github.com/ProvableHQ/snarkVM/pull/2877/commits/98d05302e5bbc3868175b6931724b0fcf8705f5a)

synthesizer/process/src/cost.rs
Roee-87: Would we divide every command cost by `quotient`?

```Rust
Command::Instruction(Instruction::Abs(_)) => Ok(500 / quotient),
Command::Instruction(Instruction::AbsWrapped(_)) => Ok(500 / quotient),
```
etc.. ?

vicsn: Ah sorry yeah that's too much work. You should just divide the sum somewhere in e.g. the return of `cost_in_microcredits_v*`

vicsn (approved):
LGTM!

vicsn:
@Roee-87 compilation needs to be fixed (the git commit hook doesn't catch this)

synthesizer/process/src/cost.rs
Roee-87: OK -- I included it in the `cost_in_microcredits` method.  [1aafeb9](https://github.com/ProvableHQ/snarkVM/pull/2877/commits/1aafeb91ab341cbefa2555179792985eefb8843e)

vicsn:
To fix the failing `test_upgrade_child_program_beyond_transaction_spend_limit` test, you'll have to increase the body of the upgraded child from `construct_finalize_body(75, 0)` to something higher to ensure it actually goes above the spend limit with the new lower costs.

To fix the failing `synthesizer-integration` test, you have to run it locally with `REWRITE_EXPECTATIONS=1`. I'll consider removing this from the main CI workflow in the future.

d0cd:
Are constructor costs intended to be left as is?
I see that in `deployment_cost_v2` it is scaled down.
However, the function`constructor_cost_in_microcredits` is unchanged.
This might make error reporting and dependent libraries show incorrect values.

d0cd:
Costs need to also be updated [here](https://github.com/ProvableHQ/snarkVM/blob/ddafdc19c398f04264ccfae97c7970cfbfde9bf8/synthesizer/process/src/stack/mod.rs#L274) otherwise these changes will not go into effect. 

A couple ideas for easier maintenance:
- add a comment `cost.rs` documenting the process for adding a new cost function and what needs to be updated
- each consensus version bump ideally should have a test file `test_v<>` with integration tests for that specific item. Even if the change isn't pertinent, having the test file and intentionally declaring it empty can make for cleaner practice.

Roee-87:
@vicsn Can you chime on this?  I think you had indicated that I could remove the unit tests for `cost_in_microcredits_v3` -- unless the intention is to add the tests in a separate module?

> each consensus version bump ideally should have a test file test_v<> with integration tests for that specific item. Even if the change isn't pertinent, having the test file and intentionally declaring it empty can make for cleaner practice.

vicsn:
> @vicsn Can you chime on this? I think you had indicated that I could remove the unit tests for cost_in_microcredits_v3 -- unless the intention is to add the tests in a separate module?

Yeah the tests you copied were not testing the specific migration Pranav is talking about. Sorry for adding yet another refactor, but I think this small redesign should make this PR more robust: https://github.com/ProvableHQ/snarkVM/issues/2886
I suggest that on this PR, you can make the existing cost functions private (not exposed), so we're sure the VM only uses your new wrapper. Then I think no new tests are needed at this point.
This is not backwards compatible, but it's a small change which I think is important in order to be a better library.

Also, new unit tests are failing...

raychu86 (changes requested):
Code changes look fine logically, but we need to ensure that the fee reductions do not introduce attack or DoS vectors. The previous fee amounts were selected very carefully to align with an economic defense (at around 100 credits per second of runtime), and each finalize command was benchmarked and evaluated.

We should be running these benchmarks and anaylsis again like we did for our original pricing determination and the subsequent change for mapping base cost reductions:
- https://github.com/ProvableHQ/snarkVM/pull/2281
- https://github.com/ProvableHQ/snarkVM/pull/2564

vicsn:
> Code changes look fine logically, but we need to ensure that the fee reductions do not introduce attack or DoS vectors. The previous fee amounts were selected very carefully to align with an economic defense (at around 100 credits per second of runtime), and each finalize command was benchmarked and evaluated.
> We should be running these benchmarks and anaylsis again like we did for our original pricing determination and the subsequent change for mapping base cost reductions:

Thank you for catching. Because this is a linear decrease, we should not have to rerun the benchmarks and analysis. However, we should:
- split the `TRANSACTION_SPEND_LIMIT` in a compute spend and storage spend limit.
- scale down the compute spend limit by `ARC_0005_COST_REDUCTION_FACTOR`
- review whether this is still backwards compatible or whether e.g. its relation to `BATCH_SPEND_LIMIT` will be messy

raychu86:
> > Code changes look fine logically, but we need to ensure that the fee reductions do not introduce attack or DoS vectors. The previous fee amounts were selected very carefully to align with an economic defense (at around 100 credits per second of runtime), and each finalize command was benchmarked and evaluated.
> > We should be running these benchmarks and anaylsis again like we did for our original pricing determination and the subsequent change for mapping base cost reductions:
> 
> Thank you for catching. Because this is a linear decrease, we should not have to rerun the benchmarks and analysis. However, we should:
> 
> * split the `TRANSACTION_SPEND_LIMIT` in a compute spend and storage spend limit.
> * scale down the compute spend limit by `ARC_0005_COST_REDUCTION_FACTOR`
> * review whether this is still backwards compatible or whether e.g. its relation to `BATCH_SPEND_LIMIT` will be messy


I'm not sure why a linear decrease wouldn't require a re-run of the analysis we did before. We likely do not have significant optimizations to the finalize command runtimes, so the benchmarks from the previous runs can be used, but we should evaluate them against the current pricing.

And agreed on the `BATCH_SPEND_LIMIT`. There may be some assumptions we need to think about again

raychu86:
I thought about the `BATCH_SPEND_LIMIT` again, and don't believe there will be an impact on backwards compatibility since the fee requirements are strictly decreasing. There is a consensus halting risk if the proposals at the time of upgrade are nearing the `BATCH_SPEND_LIMIT`, but should be avoided if validators upgrade together. (@niklaslong)

`ARC_0005_COST_REDUCTION_FACTOR` is already only factored on the finalize/compute portion of the costs, so storage remains independent. It would be nice to have different safety metrics for `compute` and `storage` separately, but for now I'm less concerned about it.

vicsn:
Closing in favour of: https://github.com/ProvableHQ/snarkVM/pull/2890

Your name is still in the commits!

PR #2884, opened 2025-08-19
[Feat] Implement batched state path fetching

raychu86:
## Motivation

This PR implements `get_state_paths_for_commitments` to retrieve a batch of state paths rather than one at a time in the query. This will help mitigate the `global state root to be the same across iterations` issues seen when users/provers are querying for state. 

### Testing
Existing tests should ensure that new batched approach works as intended since we have a variety of tests for record spending.

## Related PRs

Sister snarkOS PR - https://github.com/ProvableHQ/snarkOS/pull/3794

### DIFF
--- a/ledger/query/src/query.rs
+++ b/ledger/query/src/query.rs
@@ -161,6 +161,11 @@ impl<N: Network, B: BlockStorage<N>> QueryTrait<N> for Query<N, B> {
 
     /// Returns a list of state paths for the given list of `commitment`s.
     fn get_state_paths_for_commitments(&self, commitments: &[Field<N>]) -> Result<Vec<StatePath<N>>> {
+        // Return an empty vector if there are no commitments.
+        if commitments.is_empty() {
+            return Ok(vec![]);
+        }
+
         match self {
             Self::VM(block_store) => block_store.get_state_paths_for_commitments(commitments),
             Self::REST(url) => {
--- a/ledger/query/src/query.rs
+++ b/ledger/query/src/query.rs
@@ -159,6 +159,70 @@ impl<N: Network, B: BlockStorage<N>> QueryTrait<N> for Query<N, B> {
         }
     }
 
+    /// Returns a list of state paths for the given list of `commitment`s.
+    fn get_state_paths_for_commitments(&self, commitments: &[Field<N>]) -> Result<Vec<StatePath<N>>> {
+        match self {
+            Self::VM(block_store) => block_store.get_state_paths_for_commitments(commitments),
+            Self::REST(url) => {
+                // Construct the comma separated string of commitments.
+                let commitments_string = commitments.iter().map(|cm| cm.to_string()).collect::<Vec<_>>().join(",");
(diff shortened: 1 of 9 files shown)


ledger/query/src/query.rs
```diff
@@ -159,6 +159,70 @@ impl<N: Network, B: BlockStorage<N>> QueryTrait<N> for Query<N, B> {
         }
     }
 
+    /// Returns a list of state paths for the given list of `commitment`s.
+    fn get_state_paths_for_commitments(&self, commitments: &[Field<N>]) -> Result<Vec<StatePath<N>>> {
+        match self {
+            Self::VM(block_store) => block_store.get_state_paths_for_commitments(commitments),
+            Self::REST(url) => {
```
d0cd, line 176: This will create a malformed request when the commitments are empty.

d0cd:
Overall, this PR seems to fix the problem!
Vanilla Leo v3.1.0 fails to execute `burn16` on this program: https://testnet.explorer.provable.com/program/multirecord.aleo/0 with the error
```
Inclusion expected the global state root to be the same across iterations
```

The same program, deployed on a local devnet, with the fix, successfully executes.

ledger/query/src/query.rs
raychu86: Addressed in [ec850ca](https://github.com/ProvableHQ/snarkVM/pull/2884/commits/ec850ca1da8ea550027efb5e7294c2102bb85ae2)

raychu86: There is also a catch on the snarkOS side where the following `http://0.0.0.0:3030/testnet/statePaths?commitments=` is still fine.

ledger/store/src/block/mod.rs
```diff
@@ -708,6 +708,15 @@ pub trait BlockStorage<N: Network>: 'static + Clone + Send + Sync {
         ))
     }
 
+    /// Returns a list of state paths for the given list of `commitment`s.
+    fn get_state_paths_for_commitments(
+        &self,
+        commitments: &[Field<N>],
+        block_tree: &BlockTree<N>,
+    ) -> Result<Vec<StatePath<N>>> {
+        commitments.iter().map(|commitment| self.get_state_path_for_commitment(commitment, block_tree)).collect()
```
iamalwaysuncomfortable, line 721: The read lock locks the BlockTree. I know there's a limit placed on the API call that calls this method, do we want to put a limit here as well or do we want this to be an unrestricted utility for fetching state paths and trust anyone using this will enforce proper limits to its usage if they're exposing it to outside callers?

raychu86: Added a restriction to the snarkVM level as well - [fbd87c3](https://github.com/ProvableHQ/snarkVM/pull/2884/commits/fbd87c3bedc64d9d319d65bbc3be09c5b3282165).

This will restrict services from DoSing themselves if they try to fetch too many state paths at once.

vicsn (approved):

PR #2887, opened 2025-08-22
[Feature] Add `tracing::instrument` to allow generating flame graphs

kaimast:
## Motivation
Generating flamegraphs using stack-based profilers, such as `perftools` or `cargo flamegraph` is great to measure CPU-bound synchronous workloads. For asynchronous or I/O bound-code `tracing` allows emitting information ("spans") about when a certain async function was started and finished, even if that function started/stopped multiple times, was waiting for a system call, or moved between threads.

The tracing spans can then be passed to different backends. I am particular interested in flamegraph/flamechart generation for my use case, but it also works with, for example, real-time profilers, such as [`tracy`](https://docs.rs/tracing-tracy/latest/tracing_tracy/).

## Proposed Changes
This PR adds `tracing::instrument` to various performance-critical functions when the `instrumentation` feature is enabled.

This allows, for example, generating flame graphs as is done in [snarkOS PR #3916](https://github.com/ProvableHQ/snarkOS/pull/3916)

### DIFF
--- a/README.md
+++ b/README.md
@@ -60,7 +60,7 @@ The following is an (incomplete) list of features flags in the snarkVM crate.
 * **instrumentation** -
   Adds `tracing::instrument` statements to (some) functions. This is useful for profiling, but should not be used in production.
 * **serial** -
-  *Disables* paralle processing using `rayon`.
+  *Disables* parallel processing using `rayon`. This is useful for specialized profiling, but should not be used in production.
 * **algorithms** -
   Adds the `algorithms` crate to `snarkvm` (as `snarkvm::algorithms`)
 * **circuit** -
@@ -78,7 +78,7 @@ The following is an (incomplete) list of features flags in the snarkVM crate.
 
 ### 3. Building Guide
 
-You can also build snarkVM from source. Because snarkVM is a library, this follwing guide is only useful if you plan to make modifications to its source code.
+You can also build snarkVM from source. Because snarkVM is a library, this following guide is only useful if you plan to make modifications to its source code.
 
 ### 3.1 Install Rust
 
@@ -98,7 +98,7 @@ We recommend installing Rust using [rustup](https://www.rustup.rs/). You can ins
 - Windows (64-bit):
 
   Download the [Windows 64-bit executable](https://win.rustup.rs/x86_64) or
-  [Windows 32-bit executable](https://win.rustup.rs/i686) and follow the on-screen instructions.i
+  [Windows 32-bit executable](https://win.rustup.rs/i686) and follow the on-screen instructions.
 
 ### 3.2 Additional Dependencies
 
--- a/README.md
+++ b/README.md
@@ -13,9 +13,10 @@
 ## Table of Contents
 
 * [1. Overview](#1-overview)
-* [2. Build Guide](#2-build-guide)
-* [3. Contributors](#3-contributors)
-* [4. License](#4-license)
+* [2. Usage Guide](#2-usage-guide)
+* [3. Build Guide](#3-build-guide)
+* [4. Contributors](#4-contributors)
+* [5. License](#5-license)
 
 ## 1. Overview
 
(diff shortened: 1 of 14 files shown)


README.md
```diff
@@ -83,9 +91,40 @@ Now to use `snarkvm`, in your terminal, run:
...
+  Adds the `synthesizer` crate to `snarkvm` (as `snarkvm::synthesizer`)
+* **parameters** -
+  Adds the `parameters` crate to `snarkvm` (as `snarkvm::parameters`)
+* **wasm** -
+  Enables behavior specific for WebAssembly. This feature should only be enabled when compiling to the `wasm32` architecture.
+
 ## 3. Contributors
 
-Thank you for helping make snarkVM better!  
+Thank you for helping make snarkVM betkter!
```
ljedrz:
```suggestion
Thank you for helping make snarkVM better!  
```

README.md
```diff
@@ -83,9 +91,40 @@ Now to use `snarkvm`, in your terminal, run:
...
+
+* **cuda** -
+  Allows some operations to run on the (NVidia) GPU, instead of on the CPU.
+* **cli** -
+  Enables the command-line interface. Needed when installing the `snarkvm` binary.
+* **locktick** -
+  This feature turns on code for detecting deadlocks.
+* **test_targets** -
+  This feature allows the lowering of coinbase and proof targets for testing.
+* **tracing** -
```
ljedrz: I'd recommend the name `instrumentation` instead, as `tracing` might imply just the use of the `tracing` crate to someone who skipped reading the README

kaimast: That's a good idea. Will do!

ledger/block/src/verify.rs
```diff
@@ -27,6 +27,7 @@ use rayon::prelude::*;
 
 impl<N: Network> Block<N> {
     /// Ensures the block is correct.
+    #[cfg_attr(feature = "instrumentation", tracing::instrument(skip_all))]
```
ljedrz: would it be possible to include the block's hash in the instrumentation, so that we could immediately tell which blocks potentially caused a slowdown?

ledger/src/advance.rs
```diff
@@ -92,6 +92,7 @@ impl<N: Network, C: ConsensusStorage<N>> Ledger<N, C> {
     }
 
     /// Adds the given block as the next block in the ledger.
+    #[cfg_attr(feature = "instrumentation", tracing::instrument(skip_all))]
```
ljedrz: ditto; the same applies to any function where there might be an identifiable "culprit" for a slowdown, and where an individual slowdown is possible

README.md
```diff
@@ -35,9 +36,51 @@
...
+* **cli** -
+  Enables the command-line interface. Needed when installing the `snarkvm` binary.
+* **locktick** -
+  This feature turns on code for detecting deadlocks.
+* **test_targets** -
+  This feature allows the lowering of coinbase and proof targets for testing.
+* **instrumentation** -
+  Adds `tracing::instrument` statements to (some) functions. This is useful for profiling, but should not be used in production.
+* **serial** -
+  *Disables* paralle processing using `rayon`.
```
ljedrz:
```suggestion
  *Disables* parallel processing using `rayon`. This is useful for specialized profiling, but should not be used in production.
```

README.md
```diff
@@ -49,30 +92,34 @@ We recommend installing Rust using [rustup](https://www.rustup.rs/). You can ins
 - Windows (64-bit):
 
   Download the [Windows 64-bit executable](https://win.rustup.rs/x86_64) or
-  [Windows 32-bit executable](https://win.rustup.rs/i686) and follow the on-screen instructions.
+  [Windows 32-bit executable](https://win.rustup.rs/i686) and follow the on-screen instructions.i
```
ljedrz:
```suggestion
  [Windows 32-bit executable](https://win.rustup.rs/i686) and follow the on-screen instructions.
```

ljedrz (approved):
LGTM in general, left some nits and suggestions

ledger/src/advance.rs
kaimast: I added the block height. For the hash we would need to implement `tracing::field::Value` which is best in a future PR. 
Generally, we should add more tracing annotation as needed for specific benchmark. This PR is just intended to set up the initial infrastructure for it.

kaimast:
@vicsn I added a short motivation to the PR description.

vicsn:
> @vicsn I added a short motivation to the PR description.

Thank you, that seems valuable. Would this approach be able to replace https://github.com/ProvableHQ/snarkOS/pull/3930 and create a similar graph? Or will we get a different representation? I'm happy to delay merging snarkOS PR 3930 until we hone in on the specific method we want to support long-term

kaimast:
I think there are ways to generate similar graphs with tracing. However, I should be fine to keep both for now, and we can revise it later.

PR #2888, opened 2025-08-24
[Fix] Record register check for valid upgrades.

d0cd:
This PR:
- adds an additional restriction that record output registers must match across upgrades.
- adds a test verifying this behavior.

**Note. Exiting programs may depend on a program whose register index is updated.** If this is the case, dependent programs must also upgrade to re-sample keys.


synthesizer/src/vm/verify.rs
```diff
@@ -324,6 +326,28 @@ impl<N: Network, C: ConsensusStorage<N>> VM<N, C> {
...
                                     "Invalid deployment transaction '{id}' - the existing program does not have a constructor, but the deployment program does"
                                 );
+                                // If the consensus version is V10 or greater, then check that each function's **record** output registers match the existing program.
+                                if consensus_version >= ConsensusVersion::V10 {
+                                    for (id, function) in existing_program.functions() {
+                                        // Get the corresponding function in the new program.
+                                        let Ok(new_function) = deployment.program().get_function(id) else {
+                                            bail!("Invalid deployment transaction '{id}' - missing function '{id}'")
+                                        };
+                                        // Ensure the record output registers match.
```
vicsn, line 336: Can we simplify and generalize to check **all** input **and** output registers? This brings clarity for developers.

And instead of checking just the list of registers (which may always be in the right order), can you ensure the list of `({input, output}_type, register)` pairs match?

In the future we can loosen the requirement and ensure that `({input, output}_type, relative_{input, output}_register_index)` matches, meaning the absolute register index may increase, allowing developers to write more logic.

synthesizer/src/vm/tests/test_v10.rs
```diff
@@ -0,0 +1,411 @@
...
+function second:
+    input r0 as test_one.aleo/First.record;
+    call test_one.aleo/first r0 into r1;
+    cast r0.owner r0.data into r2 as Second.record;
+    output r2 as Second.record;
+
+function cannot_be_called_from_test_one:
+    input r0 as test_one.aleo/First.record;
+    call test_one.aleo/first r0 into r1;
+    output r1 as test_one.aleo/First.record;
```
vicsn, line 93: Would the following simplified version also not be callable from test.one for the same reasons? If yes, it would be cheap and nice to add as a test. If not, can you explain why?

```
function cannot_be_called_from_test_one:
    input r0 as test_one.aleo/First.record;
    output r0 as test_one.aleo/First.record;
```

synthesizer/src/vm/tests/test_v10.rs
```diff
@@ -0,0 +1,411 @@
...
+    owner as address.private;
+    data as field.private;
+
+constructor:
+    assert.eq true true;
+
+function second:
+    input r0 as test_one.aleo/First.record;
+    call test_one.aleo/first r0 into r1;
+    cast r0.owner r0.data into r2 as Second.record;
```
vicsn, line 87: By the way, this behaviour is also somewhat surprising to me: apparently we can read previously consumed record values, which kind of messes with the "pass by value" analogy. Do you think this is desired? I'm happy to open a new issue on snarkVM or our docs based on our assessment.

synthesizer/src/vm/tests/test_v10.rs
d0cd: I believe this is callable.
The rule is: when a record (not external record) is passed in as an input to a function, it is consumed. The `call test_one.aleo/first r0 into r1;` attempts to consume the record again, which is why it is not callable from test one.
If you remove the call, then the record is consumed once if the function is called from `test_one`.

synthesizer/src/vm/verify.rs
d0cd: The existing check is [here](https://github.com/ProvableHQ/snarkVM/blob/39c842d4b3846323bcc8d2a0825e8f2150e9d4eb/synthesizer/process/src/stack/helpers/check_upgrade.rs#L18). This ensures that input and output types match.

The `add_input` function ensures that inputs are registers. I believe the current check for type equality is sufficient. There is no way for someone to reorder input declarations without it being caught by the existing check.

I think checking **all** output registers makes for worse UX. It makes it harder to do edits.

synthesizer/src/vm/tests/test_v10.rs
d0cd: This seems consistent with our semantics for external records.

vicsn (approved):

raychu86 (approved):

PR #2893, opened 2025-08-26
Doc cleanups

vicsn:
## Motivation

Minor documentation cleanups I noticed from the most recent PR review.

### DIFF
--- a/console/network/src/consensus_heights.rs
+++ b/console/network/src/consensus_heights.rs
@@ -258,7 +258,7 @@ mod tests {
 
     /// Ensure that `MAX_CERTIFICATES` increases and is correctly defined.
     /// See the constant declaration for an explanation why.
-    fn max_certificates_and_transaction_spend_limit_increasing<N: Network>() {
+    fn max_certificates_increasing<N: Network>() {
         let mut previous_value = N::MAX_CERTIFICATES.first().unwrap().1;
         for (_, value) in N::MAX_CERTIFICATES.iter().skip(1) {
             assert!(*value >= previous_value);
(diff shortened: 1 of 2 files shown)


console/network/src/consensus_heights.rs
```diff
@@ -258,7 +258,7 @@ mod tests {
 
     /// Ensure that `MAX_CERTIFICATES` increases and is correctly defined.
     /// See the constant declaration for an explanation why.
-    fn max_certificates_and_transaction_spend_limit_increasing<N: Network>() {
+    fn max_certificates_increasing<N: Network>() {
         let mut previous_value = N::MAX_CERTIFICATES.first().unwrap().1;
```
Antonio95, line 262: What about the one liner
```
assert!(N::MAX_CERTIFICATES.windows(2).all(|pair| pair[0].0 <= pair[1].0));
```
?

The name change is good in any case 👌

Antonio95 (approved):
LGTM

Out of curiosity, is the fact that the spend limit increases with the consensus version 1) not correct/not relevant/not checked or 2) checked in other place than `max_certificates_increasing`, the function formerly known as `max_certificates_and_transaction_spend_limit_increasing`?

console/network/src/consensus_heights.rs
vicsn: The functional programmer is strong in you. Will consider in the future. I have trauma's from not being able to easily return errors from nested closures.

vicsn:
> Out of curiosity, is the fact that the spend limit increases with the consensus version 1) not correct/not relevant/not checked or 2) checked in other place than max_certificates_increasing, the function formerly known as max_certificates_and_transaction_spend_limit_increasing?

I made the mistake of thinking we were doing to increase the limit, whereas we actually needed to decrease the limit.
That particular test is just an extra reminder in case we do such things in the future, lowering limits always requires a bit more thought.
But given that we *have* lowered it, I'm not sure what such a reminding test should look like.

PR #2895, opened 2025-08-26
[Fix] Ensure endpoint base URLs that end on slashes work as expected

kaimast:
## Motivation
`http::Uri` adds a slash if there is no path after the authority (hostname+port) but does not append a slash to non-empty paths. For example, `test.com` gets converted into `test.com/`, but `test.com/foo` will remain unchanged.
All our tests assumed a simple hostname (+port) for the Query URL, but nodes might use a URL that includes some prefix, e.g. `example.com/mynode`. 

This PR changes `Query` to handle both cases as expected. It also adds a new `Query::build_endpoint` helper function that can be unit tested easier.

## Test Plan
In addition to the unit tests, I will also test manually using `snarkos developer`.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -292,9 +292,9 @@ checksum = "bef38d45163c2f1dde094a7dfd33ccf595c92905c8f8f4fdc18d06fb1037718a"
 
 [[package]]
 name = "bitflags"
-version = "2.9.2"
+version = "2.9.3"
 source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "6a65b545ab31d687cff52899d4890855fec459eb6afe0da6417b8a18da87aa29"
+checksum = "34efbcccd345379ca2868b2b2c9d3782e9cc58ba87bc7d79d5b53d9c9ae6f25d"
 
 [[package]]
 name = "blake2"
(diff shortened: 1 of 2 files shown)


vicsn:
CC @meshiest

Meshiest:
The paths we use with https://github.com/monadicus/snarkops are like this:

`https://endpoint.fqdn/api/v1/env/:env_id/cannons/default` -> load balanced query across all nodes in environment `:env_id`

Not sure why the trailing slash would be added to `test.com` but not `test.com/path/to/query` (unless the routes are properly joined in that case)

kaimast:
> Not sure why the trailing slash would be added to test.com but not test.com/path/to/query (unless the routes are properly joined in that case)

If I understand correctly, the developers of the `http` crate assumed that a URI is only valid if it starts with a slash, so an empty path is printed as `/` not ``. However, [the spec](https://datatracker.ietf.org/doc/html/rfc3986#section-3.3) allows for empty paths as well. 
If you are curious, there is a somewhat related issue [here](github.com/hyperium/http/issues/507) that talks about path handling issues in `http`.

Meshiest:
The :env_id in my example is a route parameter and is replaced with "mainnet" or "canary"
I don't actually use colons in paths besides the protocol/scheme specification (scheme://...)

vicsn (approved):

kaimast:
> The :env_id in my example is a route parameter and is replaced with "mainnet" or "canary" I don't actually use colons in paths besides the protocol/scheme specification (scheme://...)

I understand. It still seemed a good idea to have one test with a character that is not a letter or slash.

PR #2898, opened 2025-08-28
Migrate from encoding crate to encoding_rs

14MR:
## Motivation

encoding crate is not supported anymore, migrating to encoding_rs

## Test Plan

Relying on `cargo test` and CI runs


vicsn:
Closing for now, we will tackle this internally, thank you for your contribution!

PR #2900, opened 2025-08-29
[Build] Do not enable default features for snarkvm crates

kaimast:
A recent PR (#2783) explicitly enables `default` on snarkVM crates in many places. We should not do this to reduce bloat and compile times. 

The `filesystem` feature also causes issues on WASM. I added a check to ensure it is not enabled when compiling to the wasm32 arch. That way, the issue can be detected at compile time, not run time.

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -170,7 +170,7 @@ jobs:
       - run_test: # This runs a single test with profiler enabled
           workspace_member: snarkvm-algorithms
           flags: varuna::prove_and_verify_with_square_matrix --features profiler
-          cache_key_suffix: -profiler
+          cache_key_suffix: -profiler-test
   
   circuit:
     executor: rust-docker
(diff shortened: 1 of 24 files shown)


kaimast:
One test still fails with a `SIGILL` error. I am unclear what causes this and if it is related to these changes.

```
error: failed to run custom build command for `num-traits v0.2.19`

Caused by:
  process didn't exit successfully: `/home/circleci/project/target/debug/build/num-traits-cf77287d4c6301dd/build-script-build` (signal: 4, SIGILL: illegal instruction)
warning: build failed, waiting for other jobs to finish...
```

vicsn:
> One test still fails with a SIGILL error. I am unclear what causes this and if it is related to these changes.

~See: https://users.rust-lang.org/t/rustc-sigill-during-cargo-build/67328/4~
~If I SSH into the CI machine and `cargo clean` before rerunning the test, it builds succesfully~
~I tried updating the Cargo.lock by updating the criterion dependency to clear the cache, but it still fails.~

UPDATE: fixed with the last commit

PR #2901, opened 2025-08-29
[Feature] Use v2 REST API and errors

kaimast:
This PR changes `Query` to use the new REST API which contains serialized errors. Instead of simply error'ing with the status code, it will not return an `anyhow::Error` containing the full error chain provided by the node's endpoint.

It also moves all REST-specific code to a dedicated module, similar to how it is already done for static queries.


ledger/query/src/query.rs
```diff
@@ -25,16 +26,25 @@ use snarkvm_synthesizer_program::Program;
...
 pub enum Query<N: Network, B: BlockStorage<N>> {
-    /// The block store from the VM.
+    /// Query state in a local block store.
     VM(BlockStore<N, B>),
-    /// The base URL of the node.
-    REST(http::Uri),
-    /// The local state to query.
+    /// Query state using a node's REST API.
+    REST(RestQuery<N>),
+    /// Query state using a state root.
```
vicsn:
```suggestion
    /// Return static state for testing and performance.
```

ledger/query/src/query/rest.rs
```diff
@@ -0,0 +1,319 @@
...
+    /// Returns the program for the given program ID.
+    pub fn get_program(&self, program_id: &ProgramID<N>) -> Result<Program<N>> {
+        self.get_request(&format!("program/{program_id}"))
+    }
+
+    /// Returns the program for the given program ID.
+    #[cfg(feature = "async")]
+    pub async fn get_program_async(&self, program_id: &ProgramID<N>) -> Result<Program<N>> {
+        self.get_request_async(&format!("program/{program_id}")).await
+    }
```
vicsn, line 179: I think I noted in a previous PR that I don't see why we should replicate a subset of our API calls in Query. Moreover, this code contains a critical typo, which underscores the need to minimize our maintainable codebase. :)

vicsn (changes requested):
Thx! Two small changes

ledger/query/src/query/rest.rs
kaimast: None of these are new calls in `Query`.

`Query::get_program` and `Query::get_program_async` were added in 2023 according to `git blame`.  `Query::get_transaction` was added last month, but I can remove it with this PR if you want.

vicsn: Alright, sure we can keep it now it's there.

vicsn:
We can rebase this PR after mainnet has been merged back into staging

ljedrz (approved):

vicsn:
Will approve and merge after merging in origin/staging

kaimast:
I merged staging and also fixed some clippy warnings in a separate commit (again, not sure why I see those, but it is good to fix them anyway).

ledger/query/src/query.rs
```diff
@@ -25,16 +26,25 @@ use snarkvm_synthesizer_program::Program;
 
 use anyhow::{Context, Result};
 // ureq re-exports the `http` crate.
-use ureq::http::{self, uri};
+use ureq::http;
+
+mod static_;
+use static_::StaticQuery;
```
vicsn: A publicly-visible struct cannot be imported by its prior path. A `pub use` may have been removed, or the struct itself may have been renamed or removed entirely.

kaimast: Fixed! I made `RestQuery` public too for consistency.

vicsn (approved):

PR #2902, opened 2025-08-29
[Fix] Ensure endpoint base URLs that end on slashes work as expected (for staging)

### LINKED ISSUE #2899: [Bug] New `Query` URLs break downstream tools.
## 🐛 Bug Report

The new `Query` urls appear to tag on the network name directly without an intermediate slash.
The Leo CLI (not sure about other developer tools) pass in endpoints without a trailing slash.
Integrating the latest version of the code into Leo results in errors of the form
```
Failed to fetch from https://api.explorer.provable.com/v1testnet/block/height/latest
```
One suggestion is to use URL formatting libraries like `url` to "push" segments to the path.

kaimast:
Same as #2895, but with staging as its base. It seems like the PR was only merged into canary so far. Fixes #2899.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -292,9 +292,9 @@ checksum = "bef38d45163c2f1dde094a7dfd33ccf595c92905c8f8f4fdc18d06fb1037718a"
 
 [[package]]
 name = "bitflags"
-version = "2.9.2"
+version = "2.9.3"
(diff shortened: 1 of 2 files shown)


vicsn:
My intention was to merge canary back into staging after the wasm issue is fixed, as we have more hotfixes going on snarkVM/OS. Either approach works, but we have to choose one.

PR #2903, opened 2025-08-29
[Fix] Re-enable `wasm` proving

iamalwaysuncomfortable:
## Motivation

SnarkVM tag [Canary-v4.2.1](https://github.com/ProvableHQ/snarkVM/releases/tag/canary-v4.2.1) ended up introducing several issues for wasm + other environments that don't assume availability of filesystems.

These issues were as follows:
1. Feature flags were structured such that one was unable to disable the new `filesystem` feature.
2. When the `filesystem` feature was turned disabled, `wasm` [parameter fetching was disabled in parameter macros](https://github.com/ProvableHQ/snarkVM/compare/c885fb8551aa8027a88e066c04f1e87c49947bc2...7bc4c2867c6d1ba52d084907d161cc034fdf5822?utm_source=chatgpt.com#diff-6258319405e4f00f17bfa92ba4d83fb1dc841fbb985f3ed01537be03927b8bf6R169-R246).
3. `std::time::Instant::now` [was directly exposed](https://github.com/ProvableHQ/snarkVM/compare/c885fb8551aa8027a88e066c04f1e87c49947bc2...7bc4c2867c6d1ba52d084907d161cc034fdf5822?utm_source=chatgpt.com) to the main proving and verifying subroutines in the synthesizer crate. For environments like `wasm` and `sgx` which do not have access to this particular `std` library, this caused panics.

The `chore/no-default` fixed issue #1, this PR does the following to fix issues 2 & 3.
1. It re-enables `wasm` parameter fetching in the parameter macros.
2. It gates std::time invocations behind the `dev-print` feature.

## Test Plan

* Testing across SnarkVM dependent products should be added to ensure the `chore/no-default` branch does not affect any other major products. 
* Wasm tests should be added that execute program executions and parameter downloads directly in wasm should be developed and run in ci to ensure these pathways aren't broken in the future.

## Related PRs

#2900

### DIFF
--- a/parameters/src/macros.rs
+++ b/parameters/src/macros.rs
@@ -167,7 +167,7 @@ macro_rules! impl_load_bytes_logic_local {
 macro_rules! impl_load_bytes_logic_remote {
     ($remote_url: expr, $local_dir: expr, $filename: expr, $metadata: expr, $expected_checksum: expr, $expected_size: expr) => {
         cfg_if::cfg_if! {
-            if #[cfg(feature = "filesystem")] {
+            if #[cfg(all(feature = "filesystem", not(feature="wasm")))] {
                 // Compose the correct file path for the parameter file.
                 let mut file_path = aleo_std::aleo_dir();
                 file_path.push($local_dir);
@@ -188,12 +188,10 @@ macro_rules! impl_load_bytes_logic_remote {
                         );
                     }
 
-                    // Construct the URL.
-                    let url = format!("{}/{}", $remote_url, $filename);
-
                     // Load remote file
                     cfg_if::cfg_if!{
                         if #[cfg(all(not(feature = "wasm"), not(target_env = "sgx")))] {
+                            let url = format!("{}/{}", $remote_url, $filename);
                             let mut buffer = vec![];
                             Self::remote_fetch(&mut buffer, &url)?;
 
@@ -213,16 +211,6 @@ macro_rules! impl_load_bytes_logic_remote {
                                     buffer
                                 }
                             }
-                        } else if #[cfg(feature = "wasm")] {
-                            let buffer = Self::remote_fetch(&url)?;
-
-                            // Ensure the checksum matches.
-                            let candidate_checksum = checksum!(&buffer);
-                            if $expected_checksum != candidate_checksum {
-                                return checksum_error!($expected_checksum, candidate_checksum)
-                            }
-
-                            buffer
                         } else {
                             return Err($crate::errors::ParameterError::RemoteFetchDisabled);
                         }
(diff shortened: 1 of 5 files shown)


synthesizer/snark/src/certificate/mod.rs
```diff
@@ -37,6 +37,7 @@ impl<N: Network> Certificate<N> {
         proving_key: &ProvingKey<N>,
         verifying_key: &VerifyingKey<N>,
     ) -> Result<Certificate<N>> {
+        #[cfg(feature = "dev-print")]
```
kaimast, line 40: Is it possible to enable this feature on WASM? Maybe we should write the following just to be safe
```suggestion
        #[cfg(all(not(feature = "wasm"), feature = "dev-print"))]
```

parameters/src/macros.rs
```diff
@@ -241,9 +229,29 @@ macro_rules! impl_load_bytes_logic_remote {
                     return checksum_error!($expected_checksum, candidate_checksum)
                 }
                 return Ok(buffer);
-            }
-            else {
-                return Err($crate::errors::ParameterError::FilesystemDisabled);
+            } else {
```
kaimast, line 232: can you not just do `else if #[cfg(feature = "wasm")]` here?

kaimast (dismissed):
I left two small nits.

Also, you might need to run `cargo update` to get the one failing test to pass. I think we had the same problem on the other PR.

synthesizer/snark/src/certificate/mod.rs
vicsn: The dev-(e)println macros are already gated by the dev-print feature flag. They dont print anything if its not enabled. And if we did want to add that defense in depth, if it was needed, then we should do it at the declaration of the macros.

Resolving this comment for now!

parameters/src/macros.rs
```diff
@@ -167,7 +167,7 @@ macro_rules! impl_load_bytes_logic_local {
 macro_rules! impl_load_bytes_logic_remote {
     ($remote_url: expr, $local_dir: expr, $filename: expr, $metadata: expr, $expected_checksum: expr, $expected_size: expr) => {
         cfg_if::cfg_if! {
-            if #[cfg(feature = "filesystem")] {
+            if #[cfg(all(feature = "filesystem", not(feature="wasm")))] {
```
vicsn, line 170: Is this necessary? Ideally `filesystem` is not enabled when `wasm` is enabled. Keeping the interfaces tight and minimal will make for easier bug-fixing in the future.

parameters/src/macros.rs
```diff
@@ -241,9 +229,29 @@ macro_rules! impl_load_bytes_logic_remote {
...
-                return Err($crate::errors::ParameterError::FilesystemDisabled);
+            } else {
+                cfg_if::cfg_if! {
+                    if #[cfg(feature = "wasm")] {
+                        let url = format!("{}/{}", $remote_url, $filename);
+                        let buffer = Self::remote_fetch(&url)?;
+
+                        // Ensure the size matches.
+                        if $expected_size != buffer.len() {
+                            remove_file!(file_path);
```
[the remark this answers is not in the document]
vicsn: Turns out that the underlying function call is guarded by `#[cfg(not(feature = "wasm"))]` anyway... Perhaps this code can use a bit of cleanup.

synthesizer/snark/src/certificate/mod.rs
```diff
@@ -46,8 +47,11 @@ impl<N: Network> Certificate<N> {
         // Compute the certificate.
         let certificate = Varuna::<N>::prove_vk(universal_prover, fiat_shamir, verifying_key, proving_key)?;
 
-        let _elapsed = timer.elapsed().as_millis();
-        dev_println!(" • Certified '{_function_name}' (in {_elapsed} ms)");
+        #[cfg(feature = "dev-print")]
+        {
+            let _elapsed = timer.elapsed().as_millis();
+            dev_println!(" • Certified '{_function_name}' (in {_elapsed} ms)")
+        };
```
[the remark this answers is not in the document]
vicsn: 10 points to Gryffindor 7769cbb3b

PR #2904, opened 2025-09-01
Prevent unwrap panic in tx generation

vicsn:
## Motivation

This was reported in a HackerOne submission. The function requires 2 records.

I'm not aware of any actual users of this code, but it's part of the public interface so we might as well fix it.


ljedrz (approved):
The title was a bit misleading to me (considering the diff); this change avoids a certain `unwrap`-related _crash_ in case the number of records is too low.

In any case, LGTM :ok_hand:.

PR #2910, opened 2025-09-01
Use external Process in Package::deploy

vicsn:
## Motivation

Deployment of programs with imports was recently broken, because the deployment_cost function assumes the process contains all of the imports. This PR ensures we are able to re-use a Process across the Package::deploy scope.

## Test Plan

Manually tested success of nested deployments. In the future we'll have regression tests in CI on all known types of programs.

## Related PRs

https://github.com/ProvableHQ/snarkOS/pull/3819


ljedrz (approved):
LGTM :+1:

PR #2915, opened 2025-09-04
replace `encoding` with `encoding_rs`

thethorne48:
## Motivation

I wanted to contribute to the work this project is doing and this TODO felt like a good way to provide value and make a helpful but low-impact entry to the ecosystem. 

## Test Plan

This code is part of a macro that the implemented in nearly every remote call made by the system. Any tests that deal with encoding will leverage these changes already. All basic tests pass.

## Related PRs

(Link your related PRs here)

[RUSTSEC-2021-0153](https://osv.dev/vulnerability/RUSTSEC-2021-0153)

### DIFF
--- a/parameters/Cargo.toml
+++ b/parameters/Cargo.toml
@@ -25,7 +25,7 @@ edition = "2024"
 
 [features]
 filesystem = [ "aleo-std/storage" ]
-large_params = []
+large_params = [ ]
 locktick = [ "dep:locktick" ]
 no_std_out = [ ]
 rocks = [ "snarkvm-ledger-store/rocks" ]
--- a/parameters/Cargo.toml
+++ b/parameters/Cargo.toml
@@ -1,7 +1,7 @@
 [package]
 name = "snarkvm-parameters"
 version = "4.1.0"
-authors = ["The Aleo Team <hello@aleo.org>"]
+authors = [ "The Aleo Team <hello@aleo.org>" ]
 description = "Parameters for a decentralized virtual machine"
 homepage = "https://aleo.org"
 repository = "https://github.com/ProvableHQ/snarkVM"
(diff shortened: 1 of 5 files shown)


parameters/Cargo.toml
```diff
@@ -1,7 +1,7 @@
 [package]
 name = "snarkvm-parameters"
 version = "4.1.0"
-authors = [ "The Aleo Team <hello@aleo.org>" ]
```
kaimast: Could you revert those whitespace changes? It just adds noise to the PR.

parameters/Cargo.toml
```diff
@@ -10,27 +10,27 @@ keywords = [
...
   "decentralized",
-  "zero-knowledge"
+  "zero-knowledge",
 ]
 categories = [
   "compilers",
   "cryptography",
   "mathematics",
   "wasm",
-  "web-programming"
```
kaimast: Same with this comma. Both are legal TOML, so no need to change.

kaimast (changes requested):
Thank you for your contribution. I think this is a worthwhile change, as `encoding_rs` is well-maintained and used by big projects, such as Gecko/Firefox.

Could you address the minor nits I left in the comments?

parameters/src/macros.rs
```diff
@@ -135,10 +135,14 @@ macro_rules! impl_store_and_remote_fetch {
                     ))?;
 
                 // Re-encode the text back into bytes using the chosen encoding.
-                use encoding::Encoding;
-                encoding::all::ISO_8859_5
-                    .encode(&rust_text, encoding::EncoderTrap::Strict)
-                    .map_err(|_| $crate::errors::ParameterError::Wasm("Parameter decoding failed".to_string()))
```
kaimast, line 141: Maybe I don't understand this code enough by t why does it say`decoding` here, not `enocding` or `re-encoding`?

kaimast:
If possible, could you also add a doc comment to `fn remote_fetch` that explains what the returned bytes contain, and maybe also why the function signature is different from the non-WASM version?

parameters/Cargo.toml
thethorne48: No worries, had some auto-formatter include these and should have disabled that prior to committing. Reverted 👍

parameters/src/macros.rs
thethorne48: I was also confused by that terminology but elected to preserve the original error message rather than modify it to say `re-encoding`. This code is buried inside 2 levels of macros and used quite broadly so I wasn't sure enough to make that wording change.

thethorne48:
> If possible, could you also add a doc comment to `fn remote_fetch` that explains what the returned bytes contain, and maybe also why the function signature is different from the non-WASM version?

@kaimast I added a couple comments to help identify the differences. Does this make sense with the purpose of that codebase? https://github.com/ProvableHQ/snarkVM/pull/2915/commits/d9adacb244819969468c54bb5158775f119845b3

vicsn:
Closing for now, we will tackle this internally, thank you for your contribution!

PR #2924, opened 2025-09-12
Reduce the test consensus version heights

raychu86:
## Motivation

This PR minimizes the testing consensus block heights to reduce testing runtime and improve devnet testing. Currently, devnets need to wait until blocks have exceeded block 19 to be on the latest consensus logic, which is quite cumbersome.

We may also want to consider allowing devnets to start with the latest `ConsensusVersion` from block 0 or 1.

## Test Plan

Existing tests should be sufficient to evaluate correctnes


vicsn (approved):
LGTM! If tests pass.

Maybe with some refactoring setting them all at genesis height is possible, but I think that energy is better spent on Roy's test runner node

raychu86:
Had to readjust the heights due to failing tests. Some tests require a buffer between the first block and subsequent consensus versions due to verification caching, number of validators, etc.

vicsn (approved):

PR #2925, opened 2025-09-14
[Feature] Enable offline proving of public functions in `wasm`

iamalwaysuncomfortable:
## Motivation

If building transactions without an internet connection in `wasm`, the `CommiterKey` can't be initialized without powers of 15 and shifted powers of 16 being included in `wasm` binary. This PR ensures these powers are included with `wasm` binaries so that `wasm` provers can operate offline.


vicsn (approved):
LGTM! 

I'll note that you introduced these guards in the first place. But now that we have 100 MB keyboard apps, I guess adding a few of these necessary bytes don't matter.

Please make sure to run SDK and wallet tests with this new snarkVM version (I'm working on automating this, consider using https://github.com/ProvableHQ/snarkOS/pull/3814 already)

PR #2926, opened 2025-09-16
[Feature] Enable consensus heights to be set in `wasm`

iamalwaysuncomfortable:
## Motivation

Wasm has no access to its host system's userspace and thus any calls to `std::env` fail. This makes it such that any testing being done with the SDK against user devnets fails which seriously restricts the ability of nodejs and web apps to test against devnets.  This PR adds the ability for wasm environment to set the consensus heights manually.

## Related PRs

SDK [PR 1093](https://github.com/ProvableHQ/sdk/pull/1093) enables the SDK to set `ConsensusVersion` test heights through the same `CONSENSUS_VERSION_HEIGHTS` environment variable as SnarkVM. But it requires the change present within this PR.

### DIFF
--- a/console/network/src/consensus_heights.rs
+++ b/console/network/src/consensus_heights.rs
@@ -106,8 +106,13 @@ pub const TEST_CONSENSUS_VERSION_HEIGHTS: [(ConsensusVersion, u32); NUM_CONSENSU
     (ConsensusVersion::V10, 18),
 ];
 
-#[cfg(any(test, feature = "test", feature = "test_consensus_heights"))]
-pub fn load_test_consensus_heights() -> [(ConsensusVersion, u32); NUM_CONSENSUS_VERSIONS] {
+#[cfg(any(test, feature = "test", feature = "test_consensus_heights", feature = "wasm"))]
+pub fn load_test_consensus_heights(
+    #[cfg(feature = "wasm")] consensus_version_heights: Option<String>,
+) -> [(ConsensusVersion, u32); NUM_CONSENSUS_VERSIONS] {
+    #[cfg(not(feature = "wasm"))]
+    let consensus_version_heights = std::env::var("CONSENSUS_VERSION_HEIGHTS").ok();
+
     // Define a closure to verify the consensus heights.
     let verify_consensus_heights = |heights: &[(ConsensusVersion, u32); NUM_CONSENSUS_VERSIONS]| {
         // Assert that the genesis height is 0.
@@ -124,8 +129,8 @@ pub fn load_test_consensus_heights() -> [(ConsensusVersion, u32); NUM_CONSENSUS_
     let mut test_consensus_heights = TEST_CONSENSUS_VERSION_HEIGHTS;
 
     // Check if we can read the heights from an environment variable.
-    match std::env::var("CONSENSUS_VERSION_HEIGHTS") {
-        Ok(height_string) => {
+    match consensus_version_heights {
+        Some(height_string) => {
             let parsing_error = format!("Expected exactly {NUM_CONSENSUS_VERSIONS} ConsensusVersion heights.");
             // Parse the heights from the environment variable.
             let parsed_test_consensus_heights: [u32; NUM_CONSENSUS_VERSIONS] = height_string
@@ -143,7 +148,7 @@ pub fn load_test_consensus_heights() -> [(ConsensusVersion, u32); NUM_CONSENSUS_
             verify_consensus_heights(&test_consensus_heights);
             test_consensus_heights
         }
-        Err(_) => {
+        None => {
             // Verify and return the default test consensus heights.
             verify_consensus_heights(&test_consensus_heights);
             test_consensus_heights
--- a/console/network/src/consensus_heights.rs
+++ b/console/network/src/consensus_heights.rs
@@ -106,13 +106,16 @@ pub const TEST_CONSENSUS_VERSION_HEIGHTS: [(ConsensusVersion, u32); NUM_CONSENSU
     (ConsensusVersion::V10, 18),
 ];
 
-#[cfg(any(test, feature = "test", feature = "test_consensus_heights", feature = "wasm"))]
-pub fn load_test_consensus_heights(
-    #[cfg(feature = "wasm")] consensus_version_heights: Option<String>,
-) -> [(ConsensusVersion, u32); NUM_CONSENSUS_VERSIONS] {
-    #[cfg(not(feature = "wasm"))]
-    let consensus_version_heights = std::env::var("CONSENSUS_VERSION_HEIGHTS").ok();
+#[cfg(any(test, feature = "test", feature = "test_consensus_heights"))]
+pub fn load_test_consensus_heights() -> [(ConsensusVersion, u32); NUM_CONSENSUS_VERSIONS] {
+    // Attempt to read the test consensus heights from the environment variable.
+    load_test_consensus_heights_inner(std::env::var("CONSENSUS_VERSION_HEIGHTS").ok())
+}
 
+#[cfg(any(test, feature = "test", feature = "test_consensus_heights", feature = "wasm"))]
+pub(crate) fn load_test_consensus_heights_inner(
+    consensus_version_heights: Option<String>,
+) -> [(ConsensusVersion, u32); NUM_CONSENSUS_VERSIONS] {
     // Define a closure to verify the consensus heights.
     let verify_consensus_heights = |heights: &[(ConsensusVersion, u32); NUM_CONSENSUS_VERSIONS]| {
         // Assert that the genesis height is 0.
@@ -128,7 +131,7 @@ pub fn load_test_consensus_heights(
     // Define consensus version heights container used for testing.
     let mut test_consensus_heights = TEST_CONSENSUS_VERSION_HEIGHTS;
 
-    // Check if we can read the heights from an environment variable.
+    // If version heights have been specified, validate and then return them.
     match consensus_version_heights {
         Some(height_string) => {
             let parsing_error = format!("Expected exactly {NUM_CONSENSUS_VERSIONS} ConsensusVersion heights.");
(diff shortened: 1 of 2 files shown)


console/network/src/lib.rs
```diff
@@ -251,7 +253,11 @@ pub trait Network:
     #[cfg(any(test, feature = "test", feature = "test_consensus_heights"))]
     fn CONSENSUS_VERSION_HEIGHTS() -> &'static [(ConsensusVersion, u32); NUM_CONSENSUS_VERSIONS] {
         // NOTE: this function may panic, as it is only called during startup.
-        CONSENSUS_VERSION_HEIGHTS.get_or_init(load_test_consensus_heights)
+        #[cfg(not(feature = "wasm"))]
+        let consensus_version_heights = CONSENSUS_VERSION_HEIGHTS.get_or_init(load_test_consensus_heights);
+        #[cfg(feature = "wasm")]
+        let consensus_version_heights = CONSENSUS_VERSION_HEIGHTS.get_or_init(|| load_test_consensus_heights(None));
```
d0cd: To confirm my understanding. In a WASM context, if this is called before `set_consensus_version_test_heights`, `CONSENSUS_VERSION_HEIGHTS` will be set and `set_consensus_version_test_heights` will not work right?

d0cd: An alternate design could be to expect that this is always set in a WASM context.

console/network/src/consensus_heights.rs
```diff
@@ -124,8 +129,8 @@ pub fn load_test_consensus_heights() -> [(ConsensusVersion, u32); NUM_CONSENSUS_
     let mut test_consensus_heights = TEST_CONSENSUS_VERSION_HEIGHTS;
 
     // Check if we can read the heights from an environment variable.
```
d0cd: This comment should be moved to where the env var is read, and a more appropriate one can be added in.

console/network/src/lib.rs
```diff
@@ -496,3 +502,11 @@ pub trait Network:
         leaf: &Vec<Field<Self>>,
     ) -> bool;
 }
+
+#[cfg(feature = "wasm")]
+pub fn set_consensus_version_test_heights(
+    heights: Option<String>,
+) -> [(ConsensusVersion, u32); NUM_CONSENSUS_VERSIONS] {
+    let heights = load_test_consensus_heights(heights);
```
d0cd: Does it make more sense to just call the `verify_consensus_heights` function here?

console/network/src/consensus_heights.rs
```diff
@@ -106,8 +106,13 @@ pub const TEST_CONSENSUS_VERSION_HEIGHTS: [(ConsensusVersion, u32); NUM_CONSENSU
     (ConsensusVersion::V10, 18),
 ];
 
-#[cfg(any(test, feature = "test", feature = "test_consensus_heights"))]
-pub fn load_test_consensus_heights() -> [(ConsensusVersion, u32); NUM_CONSENSUS_VERSIONS] {
+#[cfg(any(test, feature = "test", feature = "test_consensus_heights", feature = "wasm"))]
```
d0cd, line 115: An alternate design could be to not support `load_consensus_test_heights` in WASM, but instead expect that `set_consensus_test_heights` is called before the VM is loaded.

vicsn: That sounds like the more ideal approach to me. W could also introduce a new wrapper function around VM initialization which is meant to be called from the WASM context, which then 1. calls `set_consensus_version_test_heights` 2. initializes the VM as usual.

console/network/src/lib.rs
iamalwaysuncomfortable: @d0cd when the SDK starts up, it initializes a single `wasm` instance that any SnarkVM code compiled to `wasm` lives within. Thus anything static gated behind a `OnceLock` or any other object that enables one-time interior immutability creates a static memory location there.

Thus whoever is the first caller of  `CONSENSUS_VERSION_HEIGHTS.get_or_init` sets it forevermore. 

> To confirm my understanding. In a WASM context, if this is called before set_consensus_version_test_heights, CONSENSUS_VERSION_HEIGHTS will be set and set_consensus_version_test_heights will not work right?

Correct. Which is by design. The design in the SDK is to mimic the behavior with the `envar` by setting test heights if they're set by the environment variable. However this environment variable NEEDS to be set in JS and then passed in externally to initialize the static variable.

[This file specifically](https://github.com/ProvableHQ/sdk/pull/1093/files#diff-b2c9e12afcd77c0d153af2fc77a00c93fef1390a5efe9ee2450bc7a700208422R2-R13) in the SDK is how it's proposed to be done. Basically, if you're in a node context, it reads the envar, and automatically sets what's in your envar into `wasm`.  Otherwise it will initialize these heights with the default heights in the `mainnet`/`testnet`/`canary` traits when executions are run.

> An alternate design could be to expect that this is always set in a WASM context.

This puts too much burden on users and outside SDKs to manage this and is extremely error prone for people who don't know how to manage this correctly.

How would they determine the correct consensus heights for testing? How would they know the correct consensus heights for production? In production, we should be getting these from the `Network` trait impls and not anywhere else.

Also, if one doesn't know to call this function immediately, we introduce a ton of unexpected behavior into the SDK.

console/network/src/lib.rs
iamalwaysuncomfortable: > Does it make more sense to just call the verify_consensus_heights function here? 

Maybe, but then we'd just have to duplicate the parsing `load_test_consensus_heights` uses prior to calling `verify_consensus_heights` (which expects the enum). We also want to use the default test consensus heights if no string is passed as many users will be clueless at what to insert for the string value. One may assume "they should know, or be guided to do this" but I can assure you most parties won't be properly guided, and it'll create more chaos in the ecosystem if we assume users will know the right thing to do rather than providing safe default values.

Extracting both the string parsing & default value setting seems like a lot of code duplication. But maybe we can refactor the core logic within `load_test_consensus_heights` into an inner function and just put it within this function and load_test_consensus_heights so we don't have to bother with using other feature flags in `load_test_consensus_heights` - thoughts?

console/network/src/consensus_heights.rs
iamalwaysuncomfortable: > An alternate design could be to not support load_consensus_test_heights in WASM, but instead expect that set_consensus_test_heights is called before the VM is loaded.

This is not a good idea because

1. You don't initialize the `VM` object in `wasm` at all because: 
A. You're trying to make the `wasm` binary small, and importing all the `Store` tooling in the VM object blows that up.
B. You're limited to 4GB in `wasm`, and using a singular VM struct tends to cause OOMs as it stores everything.  We abandoned using the `VM` struct because of this.

2. Expecting outer clients to call `set_consensus_version_test_heights` when `wasm` clients are in production and need to use the heights present in the `Mainnet`/`Testnet`/`Canary` basically requires them to carry a copy of the `ConsensusHeights` which is extremely fragile because this will inevitably fall behind what's in the traits.

What's done now is that `set_consensus_version_test_heights` is called in node when the `CONSENSUS_TEST_HEIGHTS` `envar` is set in the outer userspace (which `wasm` has no direct access to). Otherwise it sets the production consensus test heights when an execution called. Forcing `wasm` clients (including SDKs which we don't author) to carry copies of the correct test heights for production use is going to create a large amount of ecosystem tech debt.

iamalwaysuncomfortable: > 1. calls set_consensus_version_test_heights 

And how would this be set? Would we require all users to set the `CONSENSUS_TEST_HEIGHT` envar to either production or dev? How would they know what the correct production heights are? And what happens in browser environments? Environment variables don't exist in browser environments.

> 2. initializes the VM as usual.

SnarkVM functions are called often out of order, you don't initialize a VM object in the same way you do in SnarkVM or SnarkOS and use the VM object for everything you do. What you do is initialize a `wasm` runtime and use that runtime as you need to (i.e. you often just use to do perform hashing, or decrypt records, or inspect/decrypt transactions, etc.)

vicsn: Thank you for explaining. If making it _optional_ for wasm developers is easier, and there is no footgun of unexpected behaviour because we just default to the default consensus heights, then it's fine with me.

console/network/src/lib.rs
vicsn: > But maybe we can refactor the core logic within load_test_consensus_heights into an inner function and just put it within this function and load_test_consensus_heights so we don't have to bother with using other feature flags in load_test_consensus_heights - thoughts?

Yes I would prefer that! In any case, we don't use conditionally-compiled function **inputs** anywhere, let's not introduce it, just use conditionally-compiled **functions**.

console/network/src/lib.rs
vicsn: Another suggestion: `set_consensus_version_test_heights` should return an error if `CONSENSUS_VERSION_HEIGHTS` was set already. That means `set_consensus_version_test_heights` should not use `get_or_init` but https://doc.rust-lang.org/beta/std/cell/struct.OnceCell.html#method.set

console/network/src/lib.rs
iamalwaysuncomfortable: Excellent, I'll refactor to that and never again present a conditionally compiled input

console/network/src/consensus_heights.rs
iamalwaysuncomfortable: It is easier, it would create unexpected behavior on most `wasm` clients now to enforce its usage, as there'd be no way around having to make the call in JS during `wasm` load. This would mean that every SDK and user of an SDK would need to call it manually and correctly for their apps to function, which in all likelihood they'd fail to do.

console/network/src/lib.rs
iamalwaysuncomfortable: We should probably rename this function then as I'd prefer it return the Consensus Heights so the JS users are able to see what they are!

[Example](https://github.com/ProvableHQ/sdk/pull/1093/files#diff-e0680abe1fbb50813cd1649d154a4edc9e6c907d7e91a8e3e765593aa2fe636bR39-R45)

iamalwaysuncomfortable:
@vicsn @d0cd, comments addressed. This is ready for re-review.

vicsn (approved):
Beautiful

PR #2927, opened 2025-09-17
[Feature] Improved Panic Handling

kaimast:
This PR is a follow-up on #2813. 

## Context: Handling Fatal Errors in snarkVM

Generally, snarkVM and snarkOS aim to never panic if it is possible to avoid. However, there is always the possibility of bugs that were overlooked or ending up in a state where the code cannot recover safely.

If the code panics, there are two different scenarios:
- If the panic happens inside a program execution (i.e., inside `try_vm_runtime`) we can safely log it and continue operating normally. snarkVM will simply treat this as a failed execution.
- If the panic happens outside a program execution (i.e., anywhere else in snarkVM or in snarkOS), we most likely do not want to simply continue execution, and we definitely want to notify the user and developers about the error.
We had several instances where a thread or tokio task panicked, and, while information about the panic was printed to `stdout`, there was no mention of it in the logs.

While the former is already implemented. The latter was partially introduced by #2813, but this PR adds some important improvement and fixes. The behavior of **`try_vm_runtime` remains unchanged** by this, albeit its implementation changes slightly.

## Improved Panic Handling Mechanisms
Panic handling as implemented in #2813 (and also before but to a limited extent) has a major shortcoming: it did not work well with multi-threading. That is because the panic handler is set per-process but, while executing a snarkVM program another thread, unrelated to execution, might panic, and snarkVM may falsely log it as a VM error.

The proposed mechanism in this PR is to store backtraces and error messages in a thread-local variable. `try_vm_runtime` is changed to retrieve the error message from this thread-local variable.
Additionally, there is now a `catch_unwind` function that wraps around the function of the same name from the standard library and returns the error message and backtrace on error. The latter is useful when printing error messages about panics, for example, in snarkOS.

## Utilities for Task Management
`tokio` allows propagating panics through `JoinHandle`s. This PR also provides wrappers around tokio's `spawn` and `spawn_blocking` functions that contain the boilerplate for propagating such panics.
This replaces logic that before resided in `snarkos-node-bft` but will be useful to other crates.

### In snarkVM
snarkVM has one notable instance of spawning a background task, the sequential ops thread (see the changes in #2975). [80c81d8](https://github.com/ProvableHQ/snarkVM/pull/2927/commits/80c81d8b9693c6eae9f2e7aaf419da67f0c1a2f6) adds `catch_unwind` to that thread to ensure future panics in the storage thread are logged properly.

### In snarkOS
For snarkOS, the plan is to use `catch_unwind` for important background tasks such as block synchronization. Again, this is to ensure that a panic in block synchronization does not happen silently and block synchronization stops for no apparent reason.

https://github.com/ProvableHQ/snarkOS/pull/3874 contains the full set of proposed changes for snarkOS.

## Notes
* For panic handling, there is a potential overhead of storing backtraces that might not always be needed. It is my understanding that panics do not happen very frequently in production, but if this understanding is wrong, please let me know.
* Panic handling for "detached" tokio task (those without `JoinHandle`s) does not work well (yet). Future improvements to tokio will change this, but we also do not use detached tasks much.


utilities/src/errors.rs
```diff
@@ -115,17 +173,68 @@ impl<T> PrettyUnwrap for anyhow::Result<T> {
...
+            let _ = msg;
+        }
+    }
+
+    // Return the result, allowing regular error-handling.
+    result
+}
+
+/// `catch_unwind` calls the given closure `f` and, if `f` panics, returns the panic message and backtrace.
+pub fn catch_unwind<R, F: FnMut() -> R>(f: F) -> Result<R, (String, Backtrace)> {
```
raychu86, line 205: Outside of tests, who calls this particular `catch_unwind` vs the one from `std::panic::catch_unwind`?

utilities/src/errors.rs
```diff
@@ -87,6 +134,17 @@ pub trait PrettyUnwrap {
     fn pretty_unwrap(self) -> Self::Inner;
 }
 
+/// Set the global panic hook for process. Should be called exactly once.
+pub fn set_panic_hook() {
+    std::panic::set_hook(Box::new(|err| {
+        let msg = err.to_string();
+
+        println!("Hook: {msg}");
```
raychu86: Will this always print when an error is hit?

We had initially hidden all finalize panics/errors since it caused concerns and alerted node operators, despite being expected VM behavior

kaimast: This was a leftover debug message. It is removed now.

utilities/src/task.rs
```diff
@@ -0,0 +1,71 @@
...
+    fn poll(mut self: Pin<&mut Self>, cx: &mut Context<'_>) -> Poll<Self::Output> {
+        let Poll::Ready(result) = std::pin::pin!(&mut self.inner).poll(cx) else {
+            return Poll::Pending;
+        };
+
+        match result {
+            Ok(value) => Poll::Ready(value),
+            Err(err) => {
+                if err.is_panic() {
+                    // Resume the panic on the main task
```
vicsn, line 63: Cool approach, is there any prior art on this or are we inventing something new?

ledger/store/src/helpers/memory/internal/nested_map.rs
```diff
@@ -282,9 +283,9 @@ impl<
     ///
     fn contains_key_confirmed(&self, map: &M, key: &K) -> Result<bool> {
         // Serialize 'm'.
-        let m = bincode::serialize(map)?;
+        let m = bincode::serialize(map).with_context(|| "Failed to serialize map")?;
```
vicsn, line 290: I should have asked this probably a few PRs back, but it feels cumbersome to have to manually add context to basically every function call this way. These are for situations where there is no panic, just clean error handling right? Is there any other idiomatic way, maybe we just want a minimal function call backtrace?

utilities/src/errors.rs
```diff
@@ -15,7 +15,14 @@
...
 use colored::Colorize;
 
-use std::{borrow::Borrow, fmt::Display};
+use std::{any::Any, backtrace::Backtrace, borrow::Borrow, cell::Cell, fmt::Display, panic};
+
+thread_local! {
+/// The message backtrace of the last panic on this thread (if any).
+///
+/// We store this information here instead of directly processing it in a panic hook, because panic hooks are global whereas this can be processed on a per-thread basis.
+/// For example, one thread may execute a program where panics should *not* cause the entire process to terminate, while in another thread there is a panic due to a bug.
```
vicsn, line 31: This looks like quite a crucial and useful change which will save us a lot of headache in the future, thank you for the proposal. Same question as before, I'm curious to hear about prior art in the Rust world

ledger/store/src/helpers/memory/internal/nested_map.rs
kaimast: We could use [`anyhow::Error::backtrace`](https://docs.rs/anyhow/latest/anyhow/struct.Error.html#method.backtrace) instead. It would require setting `RUST_BACKTRACE=1` but I would prefer if validators did that anyway.

utilities/src/errors.rs
kaimast: The plan is to use this in the snarkOS CLI. You can see it [in this experimental branch](https://github.com/ProvableHQ/snarkOS/pull/3874/files#diff-d98d16a7220076e688d1e126cd7ede9637cdad3ad17629103cc971c87c9e1587R101).

utilities/src/errors.rs
kaimast: There is ongoing work to atomically update the panic handler (in [this RFC](github.com/rust-lang/rust/issues/92649)).

However, that still does not allow setting a handler per thread. For that, the common suggestion I found is essentially what I implemented.

ledger/store/src/helpers/memory/internal/nested_map.rs
vicsn: The documentation notes that the backtraces would incur overhead. Does that overhead stem just from the reporting, or from the mere presence of RUST_BACKTRACE=1?

In other words, it would be good to know if the overhead is acceptable for the case where we only have to print a few backtraces when we observe something we categorize as ERROR.

vicsn:
> and snarkVM may falsely log it as VM error.

Is this the main and only usecase for the thread local panic logic of this PR? I'm trying to better understand whether the refactor is worth it at the moment.

I thought we printed the error in the VM panic handler. Therefore:
1. if it were to catch a non-VM panic, I'd assume we would clearly observe this in the logs. Clumsy, but no harm done.
2. if it were to catch a VM panic of the wróng thread, would there be a risk of any of the following:
a. verification of the wrong transaction being *logged* as wrong?
b. verification of the wrong transaction being *returned* as Err?
c. the above being non-deterministic, and therefore creating a risk of a fork on the network?

Cargo.toml
```diff
@@ -506,6 +506,10 @@ default-features = false
 [workspace.dependencies.smallvec]
 version = "1.14"
 
+[workspace.dependencies.tokio]
+version = "1"
+default-features = false
```
ljedrz: nit: by default, there are no active features in `tokio`

ledger/store/src/helpers/memory/internal/nested_map.rs
ljedrz: adding context can be useful in places where the root cause itself may not be very informative on its own; it's not mandatory

ledger/store/src/helpers/rocksdb/internal/nested_map.rs
```diff
@@ -59,10 +60,11 @@ impl<M: Serialize + DeserializeOwned, K: Serialize + DeserializeOwned, V: Serial
     fn create_prefixed_map(&self, map: &M) -> Result<Vec<u8>> {
         let mut raw_map = self.context.clone();
 
-        let map_size: u32 = bincode::serialized_size(&map)?.try_into()?;
+        let map_size: u32 =
+            bincode::serialized_size(&map).with_context(|| "Failed to get size of serialize map")?.try_into()?;
```
ljedrz, line 65:
```suggestion
            bincode::serialized_size(&map).with_context(|| "Failed to get the size of a serialized map")?.try_into()?;
```

utilities/src/task.rs
```diff
@@ -0,0 +1,71 @@
...
+                    std::panic::resume_unwind(err.into_panic());
```
ljedrz, line 64: won't this cause us to shut down the entire application as opposed to "silently" panicking within a single worker thread? as outlined in the PR description, I don't think this happens often, but it needs to be clear that this makes the node as a whole more susceptible to crashes

kaimast: Yes, that was also the intention of the previous PR. It doesn't seem like a good idea to continue execution after a panic, especially when they happen in the storage layer.

Cargo.toml
kaimast: removed

ledger/store/src/helpers/rocksdb/internal/nested_map.rs
kaimast: Fixed

utilities/src/task.rs
vicsn: > It doesn't seem like a good idea to continue execution after a panic, especially when they happen in the storage layer.

Two questions for both of you:
1. If the panic happens within the worker thread, the targeted behaviour should be to catch the panic and log it. Does this PR maintain that?
2. I asked this before, but before this PR, is there a forking risk if the VM verifies transactions in parallel? Or will a transaction verification panic always bubble up to it's own thread's panic handler?

ljedrz: 1. It would now log the panic, but then resume it, shutting everything down (as opposed to just the worker thread).
2. I'm not sure why this would be a forking risk -`verify_transactions`, even if executed outside the context of adding a new block, is still ultimately performed again before block insertion

vicsn: 1. Alright so this PRs approach is not what we should be doing
2. Txs which are aborted by the parallel verification are not serially verified anymore. One thread's execution should not mess with another thread. Is this an issue currently on mainnet?

kaimast: 1. I very much think we should fail if there is a fatal error. It is a lot easier for the node to restart than to try to recover the state somehow. 
In the current design, it is very rare for the system to encounter a panic (outside of VM execution) and, if it happens, it is an indication that something happened that we did not anticipate. The goal for us should be to have a backtrace from the panic and as much logged information as possible.

2. In the current design, it seems like different threads set/unset the global panic handler. I am not sure if it causes a problem on mainnet, but the worst that could happen is that some VM panics are logged incorrectly.

vicsn: 1. A fundamental issue, besides the many lines of not perfectly audited unwraps, is that the VM has some badly designed `halt` / `or_halt_with` logic which panics. That would need to be cleaned up before we let panics be panics. Made an issue: https://github.com/ProvableHQ/snarkVM/issues/2941

2. Alright, @ljedrz do you agree?

ljedrz: 2. I agree with @kaimast - the worst case scenario is a changed panic (originating from anywhere in snarkOS/VM) message - after all, we only call `catch_unwind` for code wrapped with `try_vm_runtime`, with the panic hook only changing the message.

utilities/src/errors.rs
kaimast: I added a commit to this branch that also uses the new `catch_unwind` in the sequential ops thread. In case new database panics pop up, this will be useful for debugging.

kaimast:
@vicsn I extended the PR description

utilities/src/lib.rs
```diff
@@ -58,5 +55,24 @@ pub use serialize::*;
...
 /// Use old name for backward-compatibility.
 pub use errors::io_error as error;
+
+/// This macro provides a VM runtime environment which will safely halt
+/// without producing logs that look like unexpected behavior.
+/// In debug mode, it prints to stderr using the format: "VM safely halted at {location}: {halt message}".
+///
+/// It is more efficient to set the panic hook once and directly use `errors::try_vm_runtime`.
+#[macro_export]
+macro_rules! try_vm_runtime {
```
raychu86, line 73: Don't all the finalize instances still call the macro `try_vm_runtime!` instead of `errors::try_vm_runtime()`?

This would call `set_panic_hook` multiple times rather than the one time you claim "Should be called exactly once."

kaimast: I clarified the documentation that it is safe to call `set_panic_hook` multiple times. It will just not have any effect after the first invocation. The most recent commit also adds a boolean flag that is set once the handler is installed, to make sure there is no performance impact of calling `set_panic_hook` multiple times.

It would be great if we could replace that macro eventually, but that would require users of snarkVM to call `set_panic_hook` themselves.

raychu86 (approved):

raychu86:
Logic looks reasonable, would be nice to see more unit tests to ensure the top level (non-`try_vm_runtime` panics) are handled correctly.

eranrund:
I've spent some time looking at this PR and some of the issues and PRs referenced by it.

I think there are a few issues being discussed/addressed here:

----

### Capturing VM panics gracefully

The default panic handler is called when a panic happens and it prints the error (+ stacktrace). When running VM code we want to capture that error and alter how it gets printed. Right now this is accomplished by temporarily setting a panic hook, but since panic hooks are global across all threads, this can lead to unrelated panics being reported by the temporarily-installed hook (or a VM halt being reported by the wrong hook). This is explicitly described in a dedicated issue - https://github.com/ProvableHQ/snarkVM/issues/3327

To address this, the current PR switches from the existing set-and-restore to a panic hook that is installed once and never removed. The hook itself no longer prints anything; instead it stores the panic message and backtrace in a thread-local, and the code that caught the panic (try_vm_runtime, catch_unwind) retrieves it on a per-thread basis and decides how to report it.

### Not swallowing panics

When a Tokio task panics, the panic is only observed if someone awaits the task's `JoinHandle`, which we don't do for most tasks. That results in panics going unnoticed, leaving the process in a degraded state. We want to systematically address all these places, so that panics propagate to the parent task, and either gracefully shut down the node or immediately exit it. My intuition is that we should crash when a panic occurs, since otherwise the node is left in an unknown state, where some tasks are still running while others have failed.

The work to address this needs to happen mostly in snarkOS. We basically can't just `tokio::spawn` without either awaiting the returned `JoinHandle` (somewhere), or wrapping the task in a `catch_unwind` so we can tell a panic happened and act accordingly (e.g. log and terminate).

### Propagating panics

The PR provides some building blocks for propagating panics such that if one task awaits a child task and the child task panics, the parent task also panics (via `resume_unwind`). This is the stuff in `utilities/src/task.rs` - wrappers around `tokio::spawn`/`spawn_blocking` whose `JoinHandle` re-raises the child's panic when awaited. It could simplify some code once we started addressing the un-`await`ed spawned tasks's `JoinHandle`s.

### Capturing a stacktrace when a panic happens

`catch_unwind` does not let us obtain a stacktrace of where the panic happened; we can only get whatever was passed to `panic!()`. This PR uses a thread-local to capture this information in the panic hook, bundling it together with the panic message. This helps get more useful information when a propagated panic takes place - otherwise, once we propagate the panic (`resume_unwind`), the stacktrace is lost. This is because the stacktrace is really only available during panic hook execution, which is where it gets logged when the default panic hook is in place.


------

### Some thoughts

One consequence of the approach in this PR worth calling out: since the installed hook never prints and never delegates to the previously-installed hook, the panic message/backtrace only surfaces when the panicking code is wrapped in `try_vm_runtime`/`catch_unwind`. Any unwrapped panic - e.g. a thread or task we haven't gotten to dealing with yet - becomes completely silent, which is arguably a regression on the "panics going unnoticed" problem.

I understand the motivation for capturing the message/backtrace as data: it lets the catcher log the panic through the proper channels (e.g. https://github.com/ProvableHQ/snarkOS/blob/d8125c9cb2e3ce22cb1a3b19c4385415d5c02e04/snarkos/main.rs#L114) rather than the hook printing raw to stderr. But I think we can have both: install the hook once, gate the "VM safely halted" handling on a thread-local "in VM runtime" flag, and delegate to the previous hook for all other panics. That way unwrapped panics keep the default print/backtrace behavior, and any host-installed hook keeps working. I have a prototype of this in https://github.com/ProvableHQ/snarkVM/pull/3333.

The two approaches could compose - the chained hook could also store the `(message, backtrace)` thread-local so `catch_unwind` propagation keeps working as in this PR - but I don't think we need to capture the backtrace at all if the hook is responsible for logging it. Since `resume_unwind` doesn't re-invoke the hook, a panic gets reported exactly once at the panic site, and propagation stays purely a control-flow concern. As it stands, this PR mixes panic reporting into the propagation path.

#### Where to go from here?

How we proceed depends on whether the approach in https://github.com/ProvableHQ/snarkVM/pull/3333 is accepted. If it is, what I think remains, is deciding how to handle unhandled panics in Tokio tasks. That breaks down into two decisions:

1. **How to catch them** - e.g. a central supervisor that is handed the `JoinHandle`s of one-off tasks and awaits them, or a `spawn` wrapper with built-in `catch_unwind` (closer to what this PR proposes), or something along those lines.
2. **What to do when one happens** - my vote is log and abort: as mentioned in one of the comment threads above, restarting a node is easy, while diagnosing a node that is only partially running is challenging.

PR #2928, opened 2025-09-17
[Feature] Implement bit-related opcodes: `serialize`, `deserialize`, `hash.native.*`.

d0cd:
**NO NEW COMMITS ARE ALLOWED AFTER [41d32c5](https://github.com/ProvableHQ/snarkVM/pull/2928/commits/41d32c5f4b59a27c83f3e260b5f8b91d239a29c4)**

## Motivation
To further interoperability of AVM programs between EVM programs, this PR introduces:
- native variants of the `hash.keccak*` and `hash.sha3*` opcodes that allow the user to output the hash as a bit array. These opcodes do not implicitly perform the BHP hash on the result. 
- `serialize.*` and `deserialize.*` opcodes which allow encoding and decoding AVM values to and from bits respectively.

The list of new instructions are as follows:

- `hash.keccak256.native`
- `hash.keccak384.native`
- `hash.keccak512.native`
- `hash.sha3_256.native`
- `hash.sha3_384.native`
- `hash.sha3_512.native`
- `hash.keccak256.native.raw`
- `hash.keccak384.native.raw`
- `hash.keccak512.native.raw`
- `hash.sha3_256.native.raw`
- `hash.sha3_384.native.raw`
- `hash.sha3_512.native.raw`
- `serialize.bits`
- `serialize.bits.raw`
- `deserialize.bits`
- `deserialize.bits.raw`

Similar to the existing `.raw` instructions, "raw" indicates that the raw bits of the input value are used.

## Questions for Reviewers

- Is the encoding sound? All data types should be able to be encoded and decoded back into the original type for both that raw and standard variants.
- Is the circuit implementation of `deserialize` properly constrained? Is the circuit static?
- Should the opcodes resize to the number of bits or error when they do not match? **Please weigh in on [this](https://github.com/ProvableHQ/snarkVM/pull/2928#discussion_r2379936270) thread**


circuit/environment/src/traits/to_bits.rs
```diff
@@ -59,6 +59,36 @@ pub trait ToBitsRaw: ToBits {
...
 }
 
+pub trait ToBitArray: ToBits {
+    /// Returns `self` as a bit array in little-endian order, resizing to the given length.
+    fn to_bit_array_le(&self, length: u32) -> Vec<Self::Boolean> {
+        use crate::traits::Inject;
+        // Get the bits.
+        let mut bits = self.to_bits_le();
+        // Resize to the given length.
+        bits.resize(length as usize, Self::Boolean::constant(false));
```
vicsn: I think the padding with const is fine, as array sizes are fixed at deployment time.

I'm worried that **trimming** the bits might be unintended behaviour. Can you add an example program using the new serialization in test_v11.rs ? A developer can't easily see how many bits their types serialize to, so we could consider adding an ensure!(bits.len() <= length);

The same comment applies to other places where you resize.

console/program/src/data_types/plaintext_type/size_in_bits.rs
```diff
@@ -0,0 +1,145 @@
...
+    /// Returns the number of raw bits of a plaintext type.
+    pub fn plaintext_size_in_raw_bits<F>(&self, get_struct: &F) -> Result<usize>
+    where
+        F: Fn(&Identifier<N>) -> Result<StructType<N>>,
+    {
+        self.plaintext_size_in_raw_bits_internal(get_struct, 0)
+    }
+
+    // A helper function to determine the number of raw bits of a plaintext type, while tracking the depth of the data.
+    fn plaintext_size_in_raw_bits_internal<F>(&self, get_struct: &F, depth: usize) -> Result<usize>
```
vicsn, line 110: Could you maybe add some unit tests for this function (and the one above) to confirm their correctness?

synthesizer/program/src/logic/instruction/operation/deserialize.rs
```diff
@@ -0,0 +1,861 @@
...
+        PlaintextType::Literal(literal_type) => {
+            // Get the expected size of the literal.
+            let expected_size = literal_type.size_in_bits::<N>();
+
+            // If the variant is `FromBits`, check the variant and metadata.
+            if variant == (DeserializeVariant::FromBits as u8) {
+                let plaintext_variant = next_bits(2)?;
+                let plaintext_variant = [plaintext_variant[0], plaintext_variant[1]];
+                ensure!(
+                    plaintext_variant == [false, false],
```
vicsn: Nit: can this be constant(s) instead of magic types?

synthesizer/program/src/logic/instruction/operation/hash.rs
```diff
@@ -269,64 +328,75 @@ impl<N: Network, const VARIANT: u8> HashInstruction<N, VARIANT> {
...
+            (37, PlaintextType::Array(array_type)) => <$pt>::from_bit_array($q($N::hash_keccak512(&bits()))?, **array_type.length()),
+            (38, PlaintextType::Array(array_type)) => <$pt>::from_bit_array($q($N::hash_keccak512(&bits_raw()))?, **array_type.length()),
+            (39, PlaintextType::Array(array_type)) => <$pt>::from_bit_array($q($N::hash_sha3_256(&bits()))?, **array_type.length()),
+            (40, PlaintextType::Array(array_type)) => <$pt>::from_bit_array($q($N::hash_sha3_256(&bits_raw()))?, **array_type.length()),
+            (41, PlaintextType::Array(array_type)) => <$pt>::from_bit_array($q($N::hash_sha3_384(&bits()))?, **array_type.length()),
+            (42, PlaintextType::Array(array_type)) => <$pt>::from_bit_array($q($N::hash_sha3_384(&bits_raw()))?, **array_type.length()),
+            (43, PlaintextType::Array(array_type)) => <$pt>::from_bit_array($q($N::hash_sha3_512(&bits()))?, **array_type.length()),
+            (44, PlaintextType::Array(array_type)) => <$pt>::from_bit_array($q($N::hash_sha3_512(&bits_raw()))?, **array_type.length()),
+            (0..=44, _) => bail!("Invalid destination type for 'hash' variant: {}", $variant),
```
vicsn:
```suggestion
            (0..=44, _) => bail!("Invalid destination type for 'hash' variant: {}", $destination_type),
```

synthesizer/program/src/logic/instruction/operation/deserialize.rs
```diff
@@ -0,0 +1,861 @@
...
+            | LiteralType::I16
+            | LiteralType::I32
+            | LiteralType::I64
+            | LiteralType::I128
+            | LiteralType::U8
+            | LiteralType::U16
+            | LiteralType::U32
+            | LiteralType::U64
+            | LiteralType::U128
+            | LiteralType::Scalar => Ok(()),
```
vicsn, line 107: Any particular reason we don't support Signature? I guess it's not needed for the usecase at the moment. But it might also be a pain to add it in later with extra ConsensusVersion guards.

And do I see correctly below that we do allow an Array of Signatures as destination type?

synthesizer/program/src/logic/instruction/operation/serialize.rs
```diff
@@ -0,0 +1,504 @@
...
+    /// Initializes a new `serialize` instruction.
+    pub fn new(
+        operands: Vec<Operand<N>>,
+        operand_type: RegisterType<N>,
+        destination: Register<N>,
+        destination_type: ArrayType<N>,
+    ) -> Result<Self> {
+        // Sanity check the number of operands.
+        check_number_of_operands(VARIANT, operands.len())?;
+        println!("a");
```
vicsn: Jackson 5 making a comeback

vicsn (changes requested):
The cargo semver breakage seems innocent to me, feel free to update the BASELINE_REV whenever this is ready to be merged

synthesizer/program/src/logic/instruction/operation/deserialize.rs
d0cd: I didn't support signatures because it's size in bits exceeds the maximum array size. 
The current implementation can already support them, however a user would only get the partial bits.
It wouldn't be too hard to support it in the future since the syntactic consensus version guard is straightforward.

And good catch, the current check is broken.

circuit/environment/src/traits/to_bits.rs
d0cd: Fair point! I think it all boils down to the UX we want to give developers.
Do developers need to know the exact size of the data types they want to serialize or deserialize into?
When you introduce arrays and composite types, this can be fairly tricky to do manually.
But a compiler should be able to do this for you.
The trimming can be helpful if the data type is too large to fit into our max array size.
I don't have a strong preference, I'll leave to reviewers to arbitrate.

Is there a specific behavior you wanted to test? Like serializing into smaller bits?
There are [these](https://github.com/ProvableHQ/snarkVM/blob/feat/bytes/synthesizer/process/src/tests/test_serializers.rs) tests that check for serde equivalence

vicsn: I'm specifically worried about a case where both in the client/SDK and server/deployment, a developer aims to serialize a particular message to sign and e.g. some serialized struct fields are silently ignored because of the trimming. The only way out I see at the moment is to not do the trimming... 

If it is functionality we care about in the future, we could add yet another opcode or variant which does trim... However, the developer can always manipulate and test their input data to trim it appropriately, therefore intentionally skipping certain fields. And we can build utilities on the client/SDK side to make that easier.

> There are [these](https://github.com/ProvableHQ/snarkVM/blob/feat/bytes/synthesizer/process/src/tests/test_serializers.rs) tests that check for serde equivalence

Thx that seems to suffice! Do I understand correctly that the `evaluate_function` invocation tests the serde path?

d0cd: I see your point. I'm getting rid of these traits and impls in the `utilities` and `circuit` crates. 

**How do you feel about the opcodes `serialize.bits*` and `deserialize.bits*` doing trimming?**

The program that is run in `evaluate_function` and `execute_function` tests that a given input type can be serialized and deserialized equivalently in the AVM.

console/program/src/data_types/plaintext_type/size_in_bits.rs
d0cd: Added in https://github.com/ProvableHQ/snarkVM/pull/2928/commits/40e94879eafbe9786df2f08b5ad06d8840c5cab8

circuit/environment/src/traits/to_bits.rs
d0cd: For context,
```
serialize r0 (field) into r1 ([boolean; u64]);
````
Should it error or give you the lower 64 bits of the field? (edited) 

```
derialize r0 ([boolean; 4]) into r1 (field);
```
Should it error or pad appropriately to the right number of field bits?

```
deserialize r0 ([boolean; 512]) into r1 (field);
```
Should it error or use the lower 253 bits and construct a field element? (edited)

Comment (d0cd) [reply circuit/environment/src/traits/to_bits.rs]: The `native.*` variants of the `keccak` and `sha3` hashes are implemented with trimming and extension. Should they be restricted to the exact number of bits?

Comment (d0cd) [reply circuit/environment/src/traits/to_bits.rs]: From an offline discussion with @vicsn 
```
IMO for the first implementation the middle one should pad, first and third should error.
Because trimming can still be done manually at the application level for usecases people care about. But then its at least safe by default.
Similar to how someone has to explicitly use .raw to start making exceptions
```

d0cd: I opted for requiring the correct byte sizes for all opcodes for consistency.
This is implemented in this PR: https://github.com/ProvableHQ/snarkVM/pull/2939

PR #2933, opened 2025-09-19
[CI] Fix semver errors due to CLI removal

kaimast:
`cargo semver-checks` will complain about the CLI crate and feature missing. This PR ignores those warnings. It also fixes some clippy warnings in a separate commit (I am unclear why I get those, but they need to be resolved eventually anyway).

I also added a `semver-check.sh` script, so that you can easily run the command with the same arguments as CI will. It would be better to set the baseline revision and features for the semver checks in `Cargo.toml`, but I am not sure if is possible.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -24,6 +24,12 @@ license = "Apache-2.0"
 edition = "2024"
 rust-version = "1.88.0" # Attention - Change the MSRV in rust-toolchain and in .circleci/config.yml as well
 
+[package.metadata.cargo-semver-checks.lints]
+workspace = true
+# We ignore removed features as `cli` is now missing
+# TODO remove this exception on the next release
+feature_missing = "allow"
+
 [workspace]
 members = [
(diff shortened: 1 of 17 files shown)


Cargo.toml
```diff
@@ -24,6 +24,12 @@ license = "Apache-2.0"
 edition = "2024"
 rust-version = "1.88.0" # Attention - Change the MSRV in rust-toolchain and in .circleci/config.yml as well
 
+[package.metadata.cargo-semver-checks.lints]
```
vicsn, line 27: Thx for the fix! The approach I was aiming for was not to add exceptions like this, but to update the `BASELINE_REV` in CI.

To make this clearer, we could add a section to `PULL_REQUEST_TEMPLATE`:
```
## Backwards compatibility

<!--
    If this PR breaks backwards compatibility, please explain why and update the BASELINE_REV in CI.
-->

(Write your motivation here)
```

PR #2937, opened 2025-09-26
[Perf] Add option to cache the most recent blocks in the ledger

### LINKED ISSUE #2934: [Feature] Move block cache to a lower level to be useful on more fronts
## 🚀 Feature

Currently we only cache blocks at the snarkOS level.

Even if we cache blocks in snarkVM, we should keep the behaviour intact where we only cache the highest N blocks.

kaimast:
Fixes #2934. This PR relies on the changes in #2945 and #2946. There is a [companion PR](https://github.com/ProvableHQ/snarkOS/pull/3923) in snarkOS that uses this new feature.

### API Change
There is now a new way to create a ledger object, namely `Ledger::load_with_opts`. This function is the same as `Ledger::load` but also takes a `LedgerOptions` struct.
The idea is that we can add new options in the future without introducing more constructors for `Ledger`.
This struct currently has two settings; `disable_startup_checks` and `enable_block_caching`. The former is equivalent to `Ledger::load_unchecked`.

An alternative to this new option would be a feature flag, but there are already many features in snarkVM. 

### Minor Changes
* Uses the `test_consensus_heights` feature more consistently across the code base ([9547150](https://github.com/ProvableHQ/snarkVM/pull/2937/commits/9547150e5d16e4f7c491e8f0ef73066ff1cba4aa)) and adds a `test_max_certificates` flag ([eac0376](https://github.com/ProvableHQ/snarkVM/pull/2937/commits/eac0376516561e298df2758eacacdac6d224e619)). 
My hope is that we can eventually move away from the nondescript `test` feature, but it is still there for compatibility.
* Removed `#[rustfmt(skip)]` in ledger ([7e5a2a4](https://github.com/ProvableHQ/snarkVM/pull/2937/commits/7e5a2a4b0336938b3e52d2093b40a461b15c7c1e)). I assume this is a remnant from a previous version of the ledger storage, but does not seem to be needed anymore.
* Adds a new benchmark that measures block advancement ([919c9b4](https://github.com/ProvableHQ/snarkVM/pull/2937/commits/919c9b417ec6b201c432b8aff1f9fedac819647a)).

### Notes
* This is considered a breaking change as it introduces a new method, `new_with_cache` to the `ConsensusStorage` trait. However, I am not aware of anyone implementing this trait outside snarkVM.

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -631,7 +631,7 @@ jobs:
           # test_vm_execute_and_finalize can take over 10 minutes in the current setup
           no_output_timeout: 20m
           workspace_member: snarkvm-synthesizer
-          flags: --test '*' --features test
+          flags: --test '*' --features test -- --test-threads=4 
           cache_key_suffix: -integration
 
   synthesizer-process:
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -441,11 +441,11 @@ jobs:
 
   ledger-with-rocksdb:
     executor: rust-docker
-    resource_class: << pipeline.parameters.large >>
+    resource_class: << pipeline.parameters.xlarge >>
     steps:
       - run_test:
           workspace_member: snarkvm-ledger
-          flags: --release --features=rocks -- --test-threads=2
+          flags: --release --features=rocks -- --test-threads=4
           timeout: 20m
 
   ledger-with-valid-solutions:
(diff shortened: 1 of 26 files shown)


vicsn:
This PR is full of unnecessary changes. Looking forward to see one addressing https://github.com/ProvableHQ/snarkVM/issues/2934 in a minimal way and nothing else.

PR #2938, opened 2025-09-26
[Feature] ECDSA + Bytes

d0cd:
**DO NOT MERGE**

This PR contains this work in:
- #2913 
- #2928 

It has been made to provide a complete diff for audits.


circuit/environment/src/traits/to.rs
```diff
@@ -65,6 +65,12 @@ pub trait ToFields {
     fn to_fields(&self) -> Vec<Self::Field>;
 }
 
+/// Unary operator for converting to a list of base fields.
+pub trait ToFieldsRaw: ToFields {
+    /// Returns the circuit as a list of base field elements using the raw bits.
+    fn to_fields_raw(&self) -> Vec<Self::Field>;
+}
```
rot256, line 72: Nit: why does the interface for `ToFieldsRaw` look different from `ToBitsRaw`? Why no `fn write_fields(&self, vec: &mut Vec<Self::Boolean>);`

circuit/environment/src/traits/to_bits.rs
```diff
@@ -100,6 +121,58 @@ impl<C: ToBits<Boolean = B>, B: BooleanTrait> ToBits for &[C] {
...
+}
+
+impl<C: ToBitsRaw<Boolean = B>, B: BooleanTrait> ToBitsRaw for &[C] {
+    /// A helper method to return a concatenated list of little-endian bits without variant or identifier bits from the circuits.
+    #[inline]
+    fn write_bits_raw_le(&self, vec: &mut Vec<Self::Boolean>) {
+        // The slice is order-preserving, meaning the first circuit in is the first circuit bits out.
+        for elem in self.iter() {
+            elem.write_bits_raw_le(vec);
+        }
```
rot256, line 163: Is it assumed that all the types have fixed-sized and hence the concatenation uniquely encodes a value (given its type)? If so this is not satisfied:

```rust
Vec<Vec<C>>
```

Where `C: ToBitsRaw` does not have a unique serialization: even knowing all the types the bits cannot be serialized,e g. `vec![vec![a], vec[b, c]]` has the same bits as `vec![vec![a, b], vec![c]]`.

If you want to disallow this in the type system you can add a marker trait: `trait FixedSized: ToBitsRaw` for which these implementations apply.

Then mark all the primitive types.

circuit/program/src/data/plaintext/from_bits.rs
```diff
@@ -66,7 +66,7 @@ impl<A: Aleo> Plaintext<A> {
         let variant = [variant1, variant2];
 
         // Literal
-        if variant == [false, false] {
+        if variant == PlaintextType::<A::Network>::LITERAL_PREFIX_BITS {
```
rot256, line 69: Nit: this is a stylistic choice, but consider a "match" instead: then you know that you are e.g. not accidentally checking for the same value twice, or forgot one.

circuit/program/src/data/plaintext/to_bits_raw.rs
```diff
@@ -0,0 +1,66 @@
...
+
+use super::*;
+
+impl<A: Aleo> ToBitsRaw for Plaintext<A> {
+    /// Returns this plaintext as a list of raw **little-endian** bits.
+    fn write_bits_raw_le(&self, vec: &mut Vec<Boolean<A>>) {
+        match self {
+            Self::Literal(literal, _) => {
+                // Extend the vector with the bits of the literal.
+                vec.extend_from_slice(&literal.to_bits_le());
```
rot256, line 24: Q: why is a literal serialized *with* the type tag?

circuit/program/src/data/plaintext/to_bits_raw.rs
```diff
@@ -0,0 +1,66 @@
...
+            }
+            Self::Struct(members, _) => {
+                // Compute the bits of the struct.
+                let mut bits_be = Vec::new();
+                for (_, value) in members {
+                    vec.extend(value.to_bits_raw_be());
+                }
+                bits_be.shrink_to_fit();
+                // Extend the vector with the bits of the struct.
+                vec.extend_from_slice(&bits_be);
```
rot256, line 56: Here `bits_be` is an empty vector?

Looks like a left-over from an earlier version.

circuit/program/src/data/plaintext/to_bits_raw.rs
```diff
@@ -0,0 +1,66 @@
...
+            }
+            Self::Array(elements, _) => {
+                // Compute the bits of the array.
+                for value in elements {
+                    vec.extend(value.to_bits_raw_be());
+                }
+            }
```
rot256, line 63: Nit: maybe use the `value.write_bits_raw_be(vec)` and avoid the additional allocations.

console/algorithms/src/ecdsa/mod.rs
```diff
@@ -0,0 +1,342 @@
...
+
+    /// Recover the public key from `(r,s, recovery_id)` using *your* hasher on `message`.
+    pub fn recover_public_key_with_digest(&self, digest_bits: &[bool]) -> Result<VerifyingKey> {
+        // Convert the digest output to bytes.
+        let digest = bytes_from_bits_le(digest_bits);
+
+        // Recover the public key using the prehash.
+        VerifyingKey::recover_from_prehash(&digest, &self.signature, self.recovery_id())
+            .map_err(|e| anyhow!("Failed to recover public key: {e:?}"))
+    }
```
rot256, line 175: Nit: really small nit, but in Ethereum land this has been called `ecrecover`.

console/algorithms/src/ecdsa/mod.rs
```diff
@@ -0,0 +1,342 @@
...
+        verifying_key: &VerifyingKey,
+        hasher: &H,
+        message: &[H::Input],
+    ) -> Result<()> {
+        // Hash the message.
+        let hash_bits = hasher.hash(message)?;
+
+        // Verify the signature using the prehash.
+        self.verify_with_digest(verifying_key, &hash_bits)
+    }
```
rot256, line 189: When instantiated with e.g. `Keccak256` and `H::Input = Bool` this is not an unforgeable signature scheme. The problem is when the `message` is not a guaranteed to be a multiple of 8 bits:

Keccak256 and others pack the bits into bytes, when there is not a multiple of 8 bits, the remaining bits in the last byte is implicitly padded with zero. This means that a signature on:

- `message = [1]`

Is also a valid signature on the messages:

- `message = [1, 0]`
- `message = [1, 0, 0]`
- `message = [1, 0, 0, 0]`
- `message = [1, 0, 0, 0, 0]`
- `message = [1, 0, 0, 0, 0, 0]`
- `message = [1, 0, 0, 0, 0, 0, 0]`
- `message = [1, 0, 0, 0, 0, 0, 0, 0]`

console/algorithms/src/ecdsa/mod.rs
```diff
@@ -0,0 +1,342 @@
...
+        message: &[H::Input],
+    ) -> Result<()> {
+        // Hash the message.
+        let hash_bits = hasher.hash(message)?;
+
+        // Ensure that the derived Ethereum address matches the provided one.
+        self.verify_ethereum_with_digest(ethereum_address, &hash_bits)
+    }
+
+    /// Verify `(r,s)` against `verifying_key` using *your* hasher on `message`.
```
rot256, line 214: Nit: this comment is wrong, it does not take a hasher.

console/algorithms/src/ecdsa/mod.rs
```diff
@@ -0,0 +1,342 @@
...
+    pub fn verify_ethereum_with_digest(
+        &self,
+        ethereum_address: &[u8; Self::ETHEREUM_ADDRESS_SIZE_IN_BYTES],
+        digest_bits: &[bool],
```
rot256, line 218: Q: is the digest not always going to be 256 bits in the case of Ethereum?

console/algorithms/src/ecdsa/mod.rs
```diff
@@ -0,0 +1,342 @@
...
+        // Accept optional 0x prefix
+        if let Some(rest) = s.strip_prefix("0x").or_else(|| s.strip_prefix("0X")) {
+            s = rest;
+        }
+
+        // Decode the hex string into bytes.
+        let bytes = hex::decode(s)?;
+
+        // Construct the signature from the bytes.
+        Self::from_bytes_le(&bytes)
```
rot256, line 301: FYI, this *ignores* any additional bytes appended to the `ECDSASignature`:

The default implementation of `from_bytes_le`:

```rust
pub trait FromBytes {
    /// Reads `Self` from `reader` as little-endian bytes.
    fn read_le<R: Read>(reader: R) -> IoResult<Self>
    where
        Self: Sized;

    /// Returns `Self` from a byte array in little-endian order.
    fn from_bytes_le(bytes: &[u8]) -> anyhow::Result<Self>
    where
        Self: Sized,
    {
        Ok(Self::read_le(bytes)?)
    }
}
```

Just treats the bytes as a reader, but it does not check that the reader is EOF after. Is this the desired behavior or should we be strict and use a [Cursor](https://doc.rust-lang.org/std/io/struct.Cursor.html) and check that we are at the end after reading? Does not seem like Aleo depends on ignoring additional bytes anywhere.

console/program/src/data/plaintext/from_bits.rs
```diff
@@ -59,7 +59,7 @@ impl<N: Network> Plaintext<N> {
         let variant = [variant[0], variant[1]];
 
         // Literal
-        if variant == [false, false] {
+        if variant == PlaintextType::<N>::LITERAL_PREFIX_BITS {
             let literal_variant = u8::from_bits_le(next_bits(8)?)?;
             let literal_size = u16::from_bits_le(next_bits(16)?)?;
             let literal = Literal::from_bits_le(literal_variant, next_bits(literal_size as usize)?)?;
```
rot256, line 65: Nit: you might want to consider a "match" here.

console/program/src/data/plaintext/mod.rs
```diff
@@ -43,6 +45,34 @@ pub enum Plaintext<N: Network> {
     Array(Vec<Plaintext<N>>, OnceLock<Vec<bool>>),
 }
 
+impl<N: Network> Plaintext<N> {
+    /// Returns a new `Plaintext::Array` from `Vec<bool>`, resizing to the specified size.
+    pub fn from_bit_array(mut bits: Vec<bool>, size: u32) -> Self {
+        bits.resize(size as usize, false);
```
rot256, line 51: Nit: consider adding an assert that the `.resize` is not truncating `bits` but only pads it.

synthesizer/program/src/logic/instruction/operation/sign_verify.rs
```diff
@@ -43,7 +58,10 @@ impl<N: Network> SignVerify<N> {
     /// Returns the opcode.
     #[inline]
     pub const fn opcode() -> Opcode {
-        Opcode::Sign
+        match RAW {
+            false => Opcode::Sign("sign.verify"),
+            true => Opcode::Sign("sign.verify.raw"),
+        }
```
rot256, line 64: Do you *need* a `.raw` version of the native signature scheme? 

It is risky to mix different message encodings *without domain separation* (alternatively can we add domain separation?), especially because the native encoding is TLV meaning every type has a unique prefix and hence the implicit zero padding which happens in:

```rust
/// Verifies a signature for the given address and message (as bits).
pub fn verify_bits(&self, address: &Address<N>, message: &[bool]) -> bool {
      // Pack the bits into field elements.
      match message.chunks(Field::<N>::size_in_data_bits()).map(Field::from_bits_le).collect::<Result<Vec<_>>>() {
          Ok(fields) => self.verify(address, &fields),
          Err(error) => {
              eprintln!("Failed to verify signature: {error}");
              false
          }
      }
}
```

Does not cause collisions between different types in the system.

I can imagine scenarios where users are signing a message using `.raw` which can then be verified as the non-raw version, interpreting is as, for instance as a signed authorization to transfer funds.

Signing anything with `.raw` using your key is a significant source of footguns.

synthesizer/program/src/logic/instruction/operation/serialize.rs
```diff
@@ -0,0 +1,503 @@
...
+pub fn evaluate_serialize<N: Network>(
+    variant: SerializeVariant,
+    input: &Value<N>,
+    destination_type: &ArrayType<N>,
+) -> Result<Value<N>> {
+    evaluate_serialize_internal(variant as u8, input, destination_type)
+}
+
+fn evaluate_serialize_internal<N: Network>(
+    variant: u8,
```
rot256, line 176: Nit: could it take `SerializeVariant`?

synthesizer/program/src/logic/instruction/mod.rs
```diff
@@ -90,46 +90,144 @@ pub enum Instruction<N: Network> {
...
+    /// Performs a Keccak hash on the input's raw bits, outputting 512 bits, hashing the result with BHP512.
+    HashKeccak512Raw(HashKeccak512Raw<N>),
+    /// Performs a Keccak hash on the input, outputting 512 bits.
+    HashKeccak512Native(HashKeccak512Native<N>),
+    /// Performs a Keccak hash on the input's raw bits, outputting 512 bits.
+    HashKeccak512NativeRaw(HashKeccak512NativeRaw<N>),
     /// Performs a Pedersen hash on up to a 64-bit input.
     HashPED64(HashPED64<N>),
+    /// Performs a Pedersen hash on the input's raw bits up to a 64-bit input.
+    HashPED64Raw(HashPED64Raw<N>),
```
rot256, line 190: Question: are you sure you need raw versions of all these? (PSD, BHP and PED)

These hashes are pretty Aleo-specific, do you need to interface with other systems?

synthesizer/program/src/logic/instruction/mod.rs
```diff
@@ -222,6 +326,7 @@ macro_rules! instruction {
     // i.e. `instruction!(custom_macro, self, |instruction| { operation(instruction) })`.
     ($macro_:ident, $object:expr, |$input:ident| $operation:block) => {
         $macro_!{$object, |$input| $operation, {
+            // The original opcodes.
             Abs,
```
rot256, line 330: Nit: consider adding explicit variant encodings, i.e.

```rust
Abs = 0,
AbsWrapped = 1,
Add = 2,
...
```

synthesizer/program/src/logic/instruction/operation/deserialize.rs
```diff
@@ -0,0 +1,873 @@
...
+        _ => bail!(
+            "Instruction '{}' cannot take type '{destination_type}' as input",
+            DeserializeVariant::opcode(variant)
+        ),
+    }
+}
+
+/// Deserializes the operand into the declared type.
+#[derive(Clone, PartialEq, Eq, Hash)]
+pub struct DeserializeInstruction<N: Network, const VARIANT: u8> {
```
rot256, line 127: Really sucks that Rust does not allow enums as const params.

synthesizer/program/src/logic/instruction/operation/deserialize.rs
```diff
@@ -0,0 +1,873 @@
...
+        // Check if the range is within bounds.
+        if let Some(next_bits) = subslice {
+            // Move the starting index.
+            index += n;
+            // Return the subslice.
+            Ok(next_bits)
+        } else {
+            bail!("Insufficient bits");
+        }
+    };
```
rot256, line 253: Nit: seen this in a few places, might want to create a struct for it and reuse, e.g.


```rust
struct Slicer<'a, T> {
    index: usize
    bits: &'a [T],
}

impl <'a, T> Slicer<'a, T> {
    fn next_slice(&mut self, n: usize) -> Result<&'a [T], Error> {
        // Safely procure a subslice with the length `n` starting at `index`.
        let subslice = bits.get(self.index..self.index + n);
        // Check if the range is within bounds.
        if let Some(next_bits) = subslice {
            // Move the starting index.
            self.index += n;
            // Return the subslice.
            Ok(next_bits)
        } else {
            bail!("Insufficient bits");
        }
   }
}
```

synthesizer/program/src/logic/instruction/operation/deserialize.rs
```diff
@@ -0,0 +1,873 @@
...
+
+    match destination_type {
+        PlaintextType::Literal(literal_type) => {
+            // Get the expected size of the literal.
+            let expected_size = literal_type.size_in_bits::<N>();
+
+            // If the variant is `FromBits`, check the variant and metadata.
+            if variant == (DeserializeVariant::FromBits as u8) {
+                let plaintext_variant = next_bits(2)?;
+                let plaintext_variant = [plaintext_variant[0], plaintext_variant[1]];
```
rot256, line 263: Nit: If you introduce this struct you could add:

```rust
impl <'a, T: Copy> Slicer<'a, T> {
    fn next_array<const N: usize>(&mut self) -> Result<[bool; N], Error> {
        Ok(self.next_slice(N)?
               .try_into()
               .expect("next_slice should always return exactly N bits"))
    }
}
```

So you would not have to do this:

```rust
let plaintext_variant = next_bits(2)?;
let plaintext_variant = [plaintext_variant[0], plaintext_variant[1]];
```

Also, `u8::from_bits_le` and `u16::from_bits_le` could take these fixed length arrays.

synthesizer/program/src/logic/instruction/operation/deserialize.rs
```diff
@@ -0,0 +1,873 @@
...
+            DeserializeVariant::FromBitsRaw => plaintext_type.plaintext_size_in_raw_bits(&get_struct),
+        }
+    };
+
+    // Get the number of bits needed.
+    let num_bits = get_size_in_bits(destination_type)?;
+
+    // Resize the bits to the appropriate length.
+    let mut bits = bits.to_vec();
+    bits.resize(num_bits, false);
```
rot256, line 235: This truncation behavior contradicts the described behavior in https://github.com/ProvableHQ/snarkVM/pull/2928#discussion_r2378649851

synthesizer/program/src/logic/instruction/operation/deserialize.rs
```diff
@@ -0,0 +1,873 @@
...
+            DeserializeVariant::FromBitsRaw => plaintext_type.plaintext_size_in_raw_bits(get_struct),
+        }
+    };
+
+    // Get the number of bits needed.
+    let num_bits = get_size_in_bits(destination_type)?;
+
+    // Resize the bits to the appropriate length.
+    let mut bits = bits.to_vec();
+    bits.resize(num_bits, circuit::Boolean::<A>::constant(false));
```
rot256, line 426: Ditto about expected truncation behavior.

synthesizer/program/src/logic/instruction/operation/deserialize.rs
```diff
@@ -0,0 +1,873 @@
...
+        // Check if the range is within bounds.
+        if let Some(next_bits) = subslice {
+            // Move the starting index.
+            index += n;
+            // Return the subslice.
+            Ok(next_bits)
+        } else {
+            bail!("Insufficient bits");
+        }
+    };
```
rot256, line 444: Nit: Here you could just create an other instance of the proposed `Slicer`

synthesizer/program/src/logic/instruction/operation/ecdsa_verify.rs
```diff
@@ -0,0 +1,495 @@
...
+    /// Returns the destination register.
+    #[inline]
+    pub fn destinations(&self) -> Vec<Register<N>> {
+        vec![self.destination.clone()]
+    }
+}
+
+// Perform the ECDSA verification based on the variant.
+macro_rules! do_ecdsa_verification {
+    ($variant: expr, $signature: expr, $pub_key: expr, $message: expr) => {{
```
rot256, line 164: Nit: This macro is only used once, maybe just inline it?

synthesizer/program/src/logic/instruction/operation/ecdsa_verify.rs
```diff
@@ -0,0 +1,495 @@
...
+
+    /// Parses a string into an operation.
+    #[inline]
+    fn from_str(string: &str) -> Result<Self> {
+        match Self::parse(string) {
+            Ok((remainder, object)) => {
+                // Ensure the remainder is empty.
+                ensure!(remainder.is_empty(), "Failed to parse string. Found invalid character in: \"{remainder}\"");
+                // Return the object.
+                Ok(object)
```
rot256, line 383: This "remainder is empty" check here is implicitly omitted elsewhere when using `fn from_bytes_le(bytes: &[u8]) -> anyhow::Result<Self>`

synthesizer/program/src/logic/instruction/operation/hash.rs
```diff
@@ -449,40 +631,128 @@ impl<N: Network, const VARIANT: u8> ToBytes for HashInstruction<N, VARIANT> {
...
     }
 
-    #[test]
-    fn test_parse() {
-        for destination_type in valid_destination_types() {
-            let instruction = format!("hash.bhp512 r0 into r1 as {destination_type}");
-            let (string, hash) = HashBHP512::<CurrentNetwork>::parse(&instruction).unwrap();
+    // A helper function to run a test.
+    fn run_test<N: Network, const VARIANT: u8>() {
+        // Initialize the RNG.
```
rot256, line 676: Consideration: another option instead of this `Variant` back/forth to `u8` because Rust does not allow const generics of user defined types, could be to have a trait:

```rust
trait AleoHash {
    const Variant: HashVariant
}
```

Then make things generic over that.

rot256 (commented):
A bunch of nits and comments which arose when going through the code. 
Feel free to disregard anything marked a nit as you see fit.

A number of the comments we already discuss separately.

circuit/environment/src/traits/to_bits.rs
vicsn: Rephrasing, is this just about how raw serialization of data in different data types can collide? Because that is expected behaviour.

circuit/program/src/data/plaintext/to_bits_raw.rs
vicsn: Sorry I don't see where the type tag is serialized, I think only the data is.

circuit/environment/src/traits/to_bits.rs
rot256: Different *values* of the *same type* can have the same encoding: 
the `impl` recursively defines `ToRawBits` for arbitrarily nested arrays.

A Vec<C> where C has a fixed size can be uniquely decoded if you know what C is.
But if C does not a fixed size this is not possible.

circuit/program/src/data/plaintext/to_bits_raw.rs
rot256: My bad, when I looked at this start of last week I had the erroneous understanding that `to_bits_le()` always included the tag, forgot to remove this comment.

synthesizer/program/src/logic/instruction/operation/deserialize.rs
d0cd: Addressed in #2939
No truncation is allowed.
`serialize` and `deserialize` instructions must specify exactly the correct number of bits.

synthesizer/program/src/logic/instruction/operation/deserialize.rs
d0cd: Addressed in #2939
No truncation is allowed.
`serialize` and `deserialize` instructions must specify exactly the correct number of bits.

circuit/environment/src/traits/to_bits.rs
d0cd: > Is it assumed that all the types have fixed-sized and hence the concatenation uniquely encodes a value (given its type)?

Yes, the assumption is that the size of the array is part of the type definition and deserialization requires knowing that type beforehand. The AVM doesn't support dynamic objects like vectors. 

It could be a worthwhile exercise to remove these "utility" impls and make our encoding more explicit.

console/algorithms/src/ecdsa/mod.rs
d0cd: Addressed in https://github.com/ProvableHQ/snarkVM/pull/2939 where the `.raw` and `.eth` variants all require messages that are "byte-aligned"

synthesizer/program/src/logic/instruction/operation/sign_verify.rs
d0cd: Addressed in https://github.com/ProvableHQ/snarkVM/pull/2939, where `sign.verify.raw` has been removed.

synthesizer/program/src/logic/instruction/mod.rs
d0cd: There is no explicit need ATM. We figured we'd include them for symmetry. Is there an issue with keeping them around?

synthesizer/program/src/logic/instruction/operation/deserialize.rs
d0cd: Agreed!

console/program/src/data/plaintext/mod.rs
d0cd: Addressed in https://github.com/ProvableHQ/snarkVM/pull/2939, which checks that the number of `bits` match the `size`.

d0cd:
Thanks for the comments here @rot256! I'm deferring the open threads as future work. If you see any that need to get addressed, we can continue on the specific items.

console/algorithms/src/ecdsa/mod.rs
d0cd: We support 384 and 512 bits as well.

circuit/program/src/data/plaintext/to_bits_raw.rs
d0cd: Addressed in this commit: [9df91bc](https://github.com/ProvableHQ/snarkVM/pull/2939/commits/9df91bcd860dba501e8fa402e12cfec100eaa642)

synthesizer/program/src/logic/instruction/operation/deserialize.rs
```diff
@@ -0,0 +1,873 @@
...
+
+                let literal_variant = circuit::U8::<A>::from_bits_le(next_bits(8)?);
+                A::assert_eq(&literal_variant, circuit::U8::<A>::constant(U8::new(literal_type.type_id())));
+
+                let literal_size = circuit::U16::<A>::from_bits_le(next_bits(16)?);
+                A::assert_eq(&literal_size, circuit::U16::<A>::constant(U16::new(expected_size)));
+            };
+            // Deserialize the literal.
+            let literal = circuit::Literal::<A>::from_bits_le(
+                &circuit::U8::<A>::constant(U8::new(literal_type.type_id())),
```
vicsn, line 467: Nit, partially just to reaffirm my understanding: are we constructing `circuit::U8::<A>::constant(U8::new(literal_type.type_id()))` twice for no good reason?

PR #2942, opened 2025-10-02
avoid tx root derivation

### LINKED ISSUE #2706: [Feature] Avoid tx root derivation
## 🚀 Feature

Similar to https://github.com/ProvableHQ/snarkVM/pull/2577

We can adjust `to_unconfirmed_id` and potentially other calls to to_deployment_id and to_execution_id.

Note: `fee_transition` should not return an `Option`, it always returns a fee.

usagi32:
## Motivation

closes #2706 

Tx root derivation takes some time during block generation, but these can be cached when computed for the first time.

## Test Plan

No critical or logic changes. CI should catch any regressions

### DIFF
--- a/ledger/block/src/transaction/deployment/bytes.rs
+++ b/ledger/block/src/transaction/deployment/bytes.rs
@@ -18,6 +18,14 @@ use super::*;
 impl<N: Network> FromBytes for Deployment<N> {
     /// Reads the deployment from a buffer.
     fn read_le<R: Read>(mut reader: R) -> IoResult<Self> {
+        // Read the id variant.
+        let id_variant = u8::read_le(&mut reader)?;
+        // Read the deployment id.
+        let id = match id_variant {
+            0 => None,
+            1 => Some(DeploymentID::<N>::read_le(&mut reader)?),
+            _ => return Err(error(format!("Invalid id variant '{id_variant}'"))),
+        };
         // Read the version and ensure the version is valid.
         let version = match u8::read_le(&mut reader)? {
             1 => DeploymentVersion::V1,
@@ -82,7 +90,7 @@ impl<N: Network> FromBytes for Deployment<N> {
         };
 
         // Return the deployment.
-        Self::new(edition, program, verifying_keys, program_checksum, program_owner)
+        Self::new(id, edition, program, verifying_keys, program_checksum, program_owner)
             .map_err(|err| error(format!("{err}")))
     }
 }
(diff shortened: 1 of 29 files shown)


ledger/block/src/transaction/deployment/bytes.rs
```diff
@@ -18,6 +18,14 @@ use super::*;
 impl<N: Network> FromBytes for Deployment<N> {
     /// Reads the deployment from a buffer.
     fn read_le<R: Read>(mut reader: R) -> IoResult<Self> {
+        // Read the id variant.
+        let id_variant = u8::read_le(&mut reader)?;
```
vicsn, line 22: Can this be made backwards compatible?

ledger/block/src/transaction/deployment/serialize.rs
```diff
@@ -25,7 +25,11 @@ impl<N: Network> Serialize for Deployment<N> {
                     DeploymentVersion::V1 => 3,
                     DeploymentVersion::V2 => 5,
                 };
-                let mut deployment = serializer.serialize_struct("Deployment", len)?;
+                let mut deployment =
+                    serializer.serialize_struct("Deployment", len + self.id.get().is_some() as usize)?;
+                if let Some(id) = self.id.get() {
+                    deployment.serialize_field("id", id)?;
```
vicsn, line 31: Not backwards compatible.

ledger/block/src/transaction/execution/mod.rs
```diff
@@ -17,14 +17,18 @@ mod bytes;
 mod serialize;
 mod string;
 
-use crate::{Transaction, Transition};
+use std::sync::OnceLock;
```
vicsn, line 20: please move down to line 26 so external dependencies are in the same place

ledger/block/src/transaction/deployment/mod.rs
```diff
@@ -30,6 +32,8 @@ use snarkvm_synthesizer_snark::{Certificate, VerifyingKey};
 
 #[derive(Clone)]
 pub struct Deployment<N: Network> {
+    /// The deployment id
+    id: OnceLock<DeploymentID<N>>,
```
vicsn, line 36: Would a design be possible where this is just an `Option<...>` like we did for the previous similar caching specified in the issue?

vicsn (changes requested):
Thank you for your contribution

ledger/block/src/transaction/deployment/mod.rs
usagi32: That would require mutable constructs to deployment all over the code base as when we call `to_deployment_id` it's probably best to cache the id if not computed yet once, this also implies a redundant recursive chain of callers that will have to take an `&mut`. We could probably do a `RefCell` though?

PR #2945, opened 2025-10-06
[Chore] Update baseline revision for semver checks

kaimast:
This PR updates the baseline revision used for semantic versioning checks, so that no more errors are generated due to CLI removal.

Additional Changes:
* Added a new `./.circleci/semver-checks.sh` script, so that you can easily run the check with the same arguments as CI
* Added an option to set `no_output_timeout` in `run_test`. 10 minutes appear to be too short for `parameters-large`. ([3311c1e](https://github.com/ProvableHQ/snarkVM/pull/2945/commits/3311c1e1df019bcdd0dcf9acbbc6d497386ab920))

(This is an updated version of #2933)


.circleci/config.yml
```diff
@@ -610,8 +627,11 @@ jobs:
     resource_class: << pipeline.parameters.twoxlarge >>
     steps:
       - run_test:
+          timeout: 30m
+          # test_vm_execute_and_finalize can take over 10 minutes in the current setup
+          no_output_timeout: 20m
           workspace_member: snarkvm-synthesizer
-          flags: --test '*' --features test -- --test-threads=4
+          flags: --test '*' --features test
```
vicsn: This test runs out of memory sometimes, which is why the thread limit was put in place. It is not perfect, but it seemed to do the job without having to refactor that whole test suite.

The same is true for other tests which had a limited number of threads allocated.

Cargo.toml
```diff
@@ -24,6 +24,12 @@ license = "Apache-2.0"
 edition = "2024"
 rust-version = "1.88.0" # Attention - Change the MSRV in rust-toolchain and in .circleci/config.yml as well
 
+[package.metadata.cargo-semver-checks.lints]
```
vicsn: Please remove this section. We should not have to worry about two places for updating CI, only the BASELISE_REV should be needed to be updated. And if you disagree we should discuss.

vicsn (changes requested):

Cargo.toml
kaimast: Removed. This shouldn't be needed anymore with the updated baseline revision anyway.

vicsn (approved):

Issue #2943, opened 2025-10-06
[Feature] Profile prepare_advance_to_next_quorum_block with full block


vicsn:
## 🚀 Feature

We should figure out what are the biggest bottlenecks when calling `prepare_advance_to_next_quorum_block` with a full block.

For reference, this is WIP work you can build on for a canonical test chain generator: 
- https://github.com/ProvableHQ/snarkVM/pull/2946
- https://github.com/ProvableHQ/snarkVM/pull/2947

In order to figure out the biggest bottlenecks, we should
1. Adjust the test chain generator to be able to read transactions from disk to be included in validator's proposals
2. Adjust  the test chain generator to be able to not just export a ledger, but also an uncomitted subdag and transmissions. Generate a subdag consisting of 2 rounds, 40 validatrors, 50 transactions per proposal.

With the above in hand, you should then in a new test be able to read the subdag and transmissions from disk in order to profile `prepare_advance_to_next_quorum_block`. You will also either to read in an appropriate 40 validator dev genesis block. 

A previous hacky approach to the profiling:
```
#[test]
fn test_prepare_advance_to_next_quorum_block() -> anyhow::Result<()> {
    // Deserialize the block from a file.
    let genesis_bytes = std::fs::read("genesis.bin").unwrap();
    let genesis: Block<CurrentNetwork> = bincode::deserialize(&genesis_bytes).unwrap();
    // Initialize the ledger.
    let ledger = CurrentLedger::load(genesis.clone(), StorageMode::Production).unwrap();
    let core_ledger = Arc::new(CoreLedgerService::new(ledger.clone(), Default::default()));
    // Read subdag from file.
    let subdag_bytes = std::fs::read("subdag.bin").unwrap();
    let subdag: Subdag<CurrentNetwork> = bincode::deserialize(&subdag_bytes).unwrap();
    // Read transmissions from file.
    let transmissions_bytes = std::fs::read("transmissions.bin").unwrap();
    let transmissions: IndexMap<TransmissionID<CurrentNetwork>, Transmission<CurrentNetwork>> =
        bincode::deserialize(&transmissions_bytes).unwrap();
    // Deserialize the individual transmissions.
    let transmissions = transmissions
        .into_iter()
        .map(|(id, transmission)| {
            let new_transmission = match transmission {
                Transmission::Transaction(tx) => {
                    let deserialized_tx = tx.deserialize_blocking().expect("Failed to deserialize transaction");
[...]

PR #2946, opened 2025-10-07
[Feature] Add testchain generator

kaimast:
Adds a simple tool that can be used to generate DAGs for testing.

This PR depends on the changes in #2945 to pass CI.


vicsn (approved):
LGTM from a birds-eye view, though I didn't actually run the tool

ledger/bins/testchain_generator.rs
```diff
@@ -0,0 +1,109 @@
...
+        let mut batch = builder.generate_blocks(batch_size, &mut rng).with_context(|| "Failed to generate blocks")?;
+
+        println!("Generated {pos} of {num_blocks} blocks");
+        pos += batch_size;
+        blocks.append(&mut batch);
+    }
+
+    println!("Done. Storing blocks to disk.");
+    let ledger =
+        Ledger::<Network, ConsensusDB<Network>>::load(builder.genesis_block().clone(), StorageMode::Development(0))
```
ljedrz: could we make the storage location configurable? this could clash with the current/past `--dev 0` ledger, and there is a `StorageMode::Custom`

ledger/bins/testchain_generator.rs
```diff
@@ -0,0 +1,109 @@
...
+        Ok(args) => args,
+        Err(err) => {
+            eprintln!("{err:?}");
+            eprintln!();
+            eprintln!("Usage: `snarkvm-testchain-generator <NUM_VALIDATORS> <NUM_BLOCKS> [<GENESIS_PATH>]`");
+            std::process::exit(1);
+        }
+    };
+
+    let mut rng = TestRng::default();
```
ljedrz: we might want the `TestRng` seed to be configurable in order to e.g. restart a run with the same seed, but a longer chain (in case we detect any issues or want to extend a prior run with identical properties)

ledger/bins/testchain_generator.rs
```diff
@@ -0,0 +1,109 @@
...
+
+        println!("Remove existing ledger data at {path:?}");
+    }
+
+    Ok(())
+}
+
+fn main() -> Result<()> {
+    // Uncomment this to enable logging.
+    // tracing_subscriber::fmt::init();
```
ljedrz: maybe this could be decided by a runtime param?

ledger/src/test_helpers/chain_builder.rs
ljedrz, line 117: why not just reuse the existing `TestRng`?

ledger/src/test_helpers/chain_builder.rs
```diff
@@ -135,34 +137,39 @@ impl TestChainBuilder {
...
 
+    /// Initialize the builder with the specified genesis block..
+    /// Note: this function mirrors the way the private keys are sampled in snarkOS `fn parse_genesis`.
+    pub fn new_with_quorum_size_and_genesis_block(num_validators: usize, genesis_path: String) -> Result<Self> {
+        // Attempts to load the genesis block file.
+        let buffer = std::fs::read(genesis_path)?;
+        // Return the genesis block.
+        let genesis = Block::from_bytes_le(&buffer)?;
+        /// The development mode RNG seed.
+        pub const DEVELOPMENT_MODE_RNG_SEED: u64 = 1234567890u64;
```
ljedrz, line 148: I'm guessing this is the exact seed that will yield valid dev validator keys?

ledger/store/src/helpers/rocksdb/internal/mod.rs
```diff
@@ -176,10 +176,10 @@ impl Database for RocksDB {
 
         // Ensure that multiple database instances are possible only when using the test storage
         // mode, and that in such scenarios, all of the instances are only using the test mode.
-        if !matches!(storage, StorageMode::Test(_)) {
-            ensure!(databases.len() == 1);
```
ljedrz, line 180: I wouldn't remove a production safeguard to facilitate a test - this shouldn't be an issue anymore if `StorageMode::Custom` is used instead, though

ljedrz: no, wait, this is just a logic flip? I'd rather leave it as before to avoid (my) confusion 😁

ljedrz (commented):
Left some suggestions.

ledger/src/test_helpers/chain_builder.rs
vicsn: Yes it is. Ideally snarkOS imports it from snarkVM, but that's a lot of refactoring, I think for now documentation suffices (see the comment at the top of the function).

ledger/bins/testchain_generator.rs
kaimast: The reason for this is that there are no binary specific dependencies. So we would either make the testchain generator its own crate, or have the ledger depend on tracing_subscriber.

ledger/src/test_helpers/chain_builder.rs
kaimast: The gensis beacon function generates a committee keys using that Rng. Later (line 122) the code than re-generates the same keys using the same seed. 

There might be a nicer way to do this, but this is what the initial test code looked like that I turned into this more modular testchain builder.

ledger/bins/testchain_generator.rs
kaimast: For this, and also your previous comment, we would probably need to add clap as a dependency.

I just aimed for it to be a minimal tool, hence it is just a binary target of the `ledger` crate, but we can also move it to a dedicated crate that can pull in as many dependencies as it wants.

ledger/src/test_helpers/chain_builder.rs
ljedrz: In general, I would prefer it to be its own separate crate, but since it's test-/dev-oriented, the extra path parameter could be handled with the minimal `std` plumbing, like the other ones already are.

ledger/bins/testchain_generator.rs
ljedrz: then again, that's probably niche enough to not be useful as a param - we would know the seed at the end of a problematic run, and it can be substituted manually if need be

ledger/bins/testchain_generator.rs
vicsn: FWIW totally fine with trying to keep the CLI minimal without clap, but silent ledger overwrites can be very annoying to debug, so maybe we just add one more `ledger-path` argument?

ledger/src/test_helpers/chain_builder.rs
vicsn: ^ I guess Lukasz' comment is meant for another thread :^)

ljedrz: Indeed, how did that happen :see_no_evil:

ljedrz:
@kaimast I've been working on https://github.com/ProvableHQ/snarkVM/issues/2943 based on this code, and I would recommend moving this into its own crate even more now; that way we could enable tracing by default, include a richer set of parameters, and apply profiling without worrying about dependencies.

vicsn:
> @kaimast I've been working on https://github.com/ProvableHQ/snarkVM/issues/2943 based on this code, and I would recommend moving this into its own crate even more now; that way we could enable tracing by default, include a richer set of parameters, and apply profiling without worrying about dependencies.

I relent, if it's very useful for the follow-up profiling work, @kaimast do go ahead

ledger/bins/testchain_generator.rs
kaimast: I added the --storage-path argument. I also added a --force flag, and disabled removal by default.

kaimast:
I added the requested command-line options and moved the tool to a separate crate.

ledger/store/src/helpers/rocksdb/internal/mod.rs
kaimast: We don't have a style guide, but it generally makes the code more legible to write `if condition a() else b()` instead of `if !condition b() else a()`

ljedrz: that's fine; it was just not the focus of the PR, so a change here was surprising and I was looking for something more than a refactoring

ledger/testchain-generator/src/main.rs
```diff
@@ -0,0 +1,119 @@
...
+
+use anyhow::{Context, Result, bail};
+use clap::{Parser, builder::PossibleValuesParser};
+
+#[derive(Parser)]
+struct Args {
+    /// The number of validators active on the chain.
+    /// (also corresponds to the number of certificates per round)
+    num_validators: usize,
+    /// The number of blocks to generator.
```
ljedrz, line 29:
```suggestion
    /// The number of blocks to generate.
```

ledger/testchain-generator/src/main.rs
```diff
@@ -0,0 +1,119 @@
...
+    /// Load the genesis block for the chain from the specified path instead of generating it.
+    #[clap(long)]
+    genesis_path: Option<String>,
+    /// Set a custom storage path for the ledger.
+    /// By default, it will use the ledger of the first devnet validator.
+    #[clap(long, short = 'p')]
+    storage_path: Option<String>,
+    /// Remove existing ledger if it already exists.
+    #[clap(long, short = 'f')]
+    force: bool,
```
ljedrz, line 40:
```suggestion
    overwrite_storage: bool,
```
nit: a bit more obvious IMO

ledger/testchain-generator/src/main.rs
```diff
@@ -0,0 +1,119 @@
...
+    let mut rng = TestRng::default();
+    let storage_mode =
+        if let Some(path) = args.storage_path { StorageMode::Custom(path.into()) } else { StorageMode::Development(0) };
+
+    remove_ledger(N::ID, &storage_mode, args.force)?;
+
+    let num_validators = args.num_validators;
+    let num_blocks = args.num_blocks;
+
+    println!("Initializing test chain builder with {num_validators} validators");
```
ljedrz, line 86: nit: since we've enabled `tracing`, perhaps we should commit to its logs for coherence?

ledger/testchain-generator/src/main.rs
```diff
@@ -0,0 +1,119 @@
...
+    }
+    .with_context(|| "Failed to set up test chain builder")?;
+
+    println!("Generating {num_blocks} blocks");
+
+    let mut pos = 0;
+    let mut blocks = vec![];
+
+    while blocks.len() < num_blocks {
+        let batch_size = (num_blocks - blocks.len()).min(100);
```
ljedrz, line 99: what's the intent behind the batching? block creation isn't that fast, we might as well log after every one of them, possibly including some useful information like the number of transactions - when I used this utility to produce blocks with 2k txs each from an initial pool of 6k, I felt like we might want to later have a param like `txs_per_block` that - when paired with input txs - would populate each block with the desired number of transactions, so more granularity in logging could be useful

ljedrz (approved):
LGTM, left a few nits and suggestions

ledger/testchain-generator/src/main.rs
kaimast: Done. I'll keep the short version as `-f` for lack of a better alternative.

ledger/testchain-generator/src/main.rs
kaimast: For me, it only takes a few seconds to generate 100 blocks, but that does for sure vary depending on the number of transactions. 

The ChainBuilder also generates a message for every generated block, so maybe it is sufficient to just increase log verbosity in your case?

ljedrz: No worries, it's simple enough to adjust later on if we consider it useful.

PR #2949, opened 2025-10-07
Compute size of Execution proof without needing to compute the proof

Antonio95:
Currently `Execution` storage costs are computed by generating a proof for the `Authorization` and checking its size, which creates an undesirable double-proving workflow (because the storage cost is an input when computing the fee-transaction proof). This PR adds:
- A method which, given a list of batch sizes, computes the size of the corresponding (batched) Varuna proof.
- An `Authorization::proof_size` method which computes the size of the Varuna proof for `self` by calculating the expected batch sizes and then calling the method above.
- A consistency check to ensure the inclusion batch size from the method above coincides with that in `prepare_verifier_inputs` (for futureproofing)
- An update of existing execution tests to `VarunaVersion::V2` and the addition to those tests of a proof-size check (i. e. ensuring that the size of the serialized proof, once computed, coincides with the predicted size given the `Authorization`). In the tested cases, the predicted size is an exact match.

Since testing has happened only for a limited time of example `Authorization`s available in the code, any issues should be reported when this starts being tested in more complex, real-world `Authorization`s.


Antonio95:
(converted to draft due to likely conflicts with PR #2948 , fixing soon)

synthesizer/process/src/stack/authorization/mod.rs
```diff
@@ -285,6 +286,71 @@ impl<N: Network> PartialEq for Authorization<N> {
...
+                let mut batch_sizes: Vec<usize> = circuit_frequencies.values().cloned().collect();
+
+                // We now add the single inclusion circuit for input records, if any:
+                let n_input_records = Self::number_of_input_records(self.transitions().values());
+                if n_input_records > 0 {
+                    batch_sizes.push(n_input_records);
+                }
+
+                // Varuna is always ran in hiding (i. e. ZK) mode when proving
+                // Executions. The added 1 corresponds to the version number.
```
vicsn:
```suggestion
                // Executions. The added 1 corresponds to the version number written in Proof::to_bytes_le.
```

algorithms/src/snark/varuna/data_structures/proof.rs
```diff
@@ -378,6 +380,88 @@ impl<E: PairingEngine> FromBytes for Proof<E> {
...
+///  - `batch_sizes`: the batch sizes of the circuits and instances being
+///    proved.
+///  - `hiding`: indicates whether the proof system is run in ZK mode
+///  - `varuna_version`: the version of Varuna being used
+///
+/// *Returns*:
+///  - `Some(size)` for `VarunaVersion::V2`, where `size` is the size of the
+///    proof in bytes.
+///  - `None` for `VarunaVersion::V1`.
+pub fn proof_size<E: PairingEngine>(
```
vicsn, line 397: Can you document that this size mirrors the size of `Proof::serialize_compressed`

vicsn (approved):
LGTM besides the documentation nits

algorithms/src/snark/varuna/data_structures/proof.rs
Antonio95: Changed the start of the function's documentation to
```
/// Predicts the size in bytes of a Varuna proof as produced by
/// `Proof::serialize_compressed` without needing to receive the proof itself.
```
let me know if that sounds okay

algorithms/src/snark/varuna/data_structures/proof.rs
```diff
@@ -378,6 +380,88 @@ impl<E: PairingEngine> FromBytes for Proof<E> {
...
+    varuna_version: VarunaVersion,
+    hiding: bool,
+) -> Option<usize> {
+    let n_circuits: usize = batch_sizes.len();
+    let n_instances: usize = batch_sizes.iter().sum();
+
+    match varuna_version {
+        VarunaVersion::V1 => None,
+        VarunaVersion::V2 => {
+            // All fields are serialised in Compressed mode The breakdown is as
```
d0cd, line 408:
```suggestion
            // All fields are serialized in `Compressed` mode. The breakdown is as
```
As for the nits on spelling for "serialized" and quotation for code components, they're very minor. Feel free to take or leave.

algorithms/src/snark/varuna/data_structures/proof.rs
```diff
@@ -378,6 +380,88 @@ impl<E: PairingEngine> FromBytes for Proof<E> {
...
+            // follows:
+            // - batch sizes: (boils down to CanonicalSerialize for [usize])
+            //   + one u64 for the length (the number of circuits) followed that many usize
+            //     This contains the size information for the vectors in all other fields,
+            //     which are therefore serialised without their length
```
d0cd:
```suggestion
            // - batch sizes: one `usize` for each batch size plus one `u64` for the number of batches.
            //     This contains the size information for all other vectors in the `Proof`, allowing them to be serialized without a length prefix.
```

algorithms/src/snark/varuna/data_structures/proof.rs
```diff
@@ -378,6 +380,88 @@ impl<E: PairingEngine> FromBytes for Proof<E> {
...
+            Some(
+                n_bool * size_bool
+                    + n_u64 * size_u64
+                    + size_of_val(batch_sizes)
+                    + n_field_elements * size_field_element
+                    + n_commitments * size_commitment
+                    + size_pc_proof,
+            )
+
+            // NB: "n_usize * size_usize" must be computed as
```
d0cd:
```suggestion
            // Note. "n_usize * size_usize" must be computed as
```
Minor nit. Had to look up "NB" to know what this meant :)

algorithms/src/snark/varuna/data_structures/proof.rs
```diff
@@ -378,6 +380,88 @@ impl<E: PairingEngine> FromBytes for Proof<E> {
...
+            let size_u64 = size_of::<u64>();
+            // The next two can be hard-coded if performance becomes critical
+            // and they are considered fully stable. They are 32 and 48 bytes at
+            // the time of writing, respectively (commitments are affine points
+            // written in compressed form).
+            let size_field_element = E::Fr::one().compressed_size();
+            let size_commitment = KZGCommitment::<E>::empty().compressed_size();
+
+            let size_pc_proof =
+                size_of::<usize>() + 3 * (size_commitment + 1) + if hiding { size_field_element } else { 0 };
```
d0cd: Is it a concern that the size of the serialization depends on a `usize`. Wouldn't that mean that there is a possibility for divergence based on system architecture?

synthesizer/process/src/stack/authorization/mod.rs
```diff
@@ -285,6 +286,71 @@ impl<N: Network> PartialEq for Authorization<N> {
 
 impl<N: Network> Eq for Authorization<N> {}
 
+impl<N: Network> Authorization<N> {
+    /// Returns the (exact) predicted size of the Varuna proof of an Authorization
```
d0cd:
```suggestion
    /// Returns the size of the Varuna proof of an `Authorization`
```

synthesizer/process/src/stack/authorization/mod.rs
```diff
@@ -285,6 +286,71 @@ impl<N: Network> PartialEq for Authorization<N> {
...
+    ///
+    /// *Arguments*:
+    ///  - `varuna_version`: the version of Varuna to use. Only `VarunaVersion::V2` is supported.
+    ///
+    /// *Returns*:
+    ///  - `Some(size)` for `VarunaVersion::V2`, where `size` is the size of the proof in bytes.
+    ///  - `None` for `VarunaVersion::V1`.
+    ///
+    // The value returned coincides with `Proof<N: Network>::write_le()`, which
```
d0cd:
```suggestion
    // The value returned is equal to the number of bytes returned by `Proof<N: Network>::write_le()`, which
```

synthesizer/process/src/stack/authorization/mod.rs
```diff
@@ -285,6 +286,71 @@ impl<N: Network> PartialEq for Authorization<N> {
...
+                }
+
+                // Varuna is always ran in hiding (i. e. ZK) mode when proving
+                // Executions. The added 1 corresponds to the version number written in Proof::to_bytes_le.
+                proof_size::<N::PairingCurve>(&batch_sizes, VarunaVersion::V2, true).map(|size| 1 + size)
+            }
+        }
+    }
+
+    /// Total number of inputs to the passed `Transition`s that are of type
```
d0cd:
```suggestion
    /// Returns the total number of inputs to the passed `Transition`s that are of type
```

synthesizer/process/src/tests/test_execute.rs
d0cd, line 1: I wanted to sanity check that we have a test on proof sizes where the inputs are records, ideally of different counts and sizes. Is that the case?

synthesizer/process/src/trace/mod.rs
```diff
@@ -374,12 +376,21 @@ impl<N: Network> Trace<N> {
...
-        transitions: impl ExactSizeIterator<Item = &'a Transition<N>>,
+        transitions: impl ExactSizeIterator<Item = &'a Transition<N>> + Clone,
         proof: &Proof<N>,
     ) -> Result<()> {
         // Construct the batch of inclusion verifier inputs.
         let batch_inclusion_inputs =
-            Inclusion::prepare_verifier_inputs(global_state_root, inclusion_version, transitions)?;
+            Inclusion::prepare_verifier_inputs(global_state_root, inclusion_version, transitions.clone())?;
+
+        let expected_n_inclusions = Authorization::number_of_input_records(transitions.clone());
```
d0cd: You might not need the second clone of `transitions`.

synthesizer/process/src/stack/authorization/mod.rs
d0cd, line 1: Really neat work!
It could be helpful for downstream users to have a method that takes in an `Authorization` and computes the fee amount. An added benefit is that dependent libraries (SDK, DPS, Leo) don't have to calculate the fee themselves, potentially leading to divergent implementations.

algorithms/src/snark/varuna/data_structures/proof.rs
Antonio95: Oh no, this is an error on my part. Serialising the proof with `.write_le` (which is what `execution.size_in_bytes()` does) leads to  `CanonicalSerialize for Proof` where sizes are encoded as `u64`s here: https://github.com/ProvableHQ/snarkVM/blob/78a69b70387d917ed1c16a09741a5b6d16862f17/algorithms/src/snark/varuna/data_structures/proof.rs#L304
(and the size of the vector itself is also encoded as a `u64` elsewhere).

Thank you, this could have been a tricky thing to catch later on. I've fixed the function accordingly. Out of curiosity, I also checked how `CanonicalSerialize` deals with `usize`s and they are simply serialised and deserialised as `u64`s, so all good in terms of platform independence (in whichever structures use them - not the case in this PR): https://github.com/ProvableHQ/snarkVM/blob/78a69b70387d917ed1c16a09741a5b6d16862f17/utilities/src/serialize/impls.rs#L147

Antonio95:
Hi @d0cd! Quite a few things have changed in the last commits:

1. After some discussion with @vicsn, it seems the best interface for the envisioned use cases is
    ```
    pub fn execution_cost_for_authorization<N: Network>(
         process: &Process<N>,
         authorization: &Authorization<N>,
    )
    ```
    which has been implemented and placed at the same location and level as 
    ```
    pub fn execution_cost<N: Network>(
        process: &Process<N>,
        execution: &Execution<N>,
    )
    ```
    Let me know if that looks good or you think there might be a better design (@niklaslong  might also have an informed opinion on this).

2. The `Authorization::proof_size` introduced in the previous commit has been removed, as it is not particularly useful given the new interface from point 1.

3. For some reasons (*) it was convenient to create an auxiliary `execution_cost_with_size()` which behaved exactly like `execution_cost()` but received the size of the `Execution` instead of computing it internally. So this has been added next to `execution_cost()`, which now wraps around the new function but computes the size of the `Execution` first. This caused a prototype change in `execution_cost_v1()`, `execution_cost_v2()` and `execution_cost_v3()`, which have been fixed and made private (they were only used in tests). This has all been discussed with Victor.

    _(*) The reasons, feel free to skip: because the storage cost is not directly `execution.size_in_bytes()` but a transformation of that under `execution_storage_cost()`, the new `execution_cost_for_authorization` could not simply read the storage cost returned by `execution_cost()` and add the proof size (because the returned value may have been non-additively modified by `execution_storage_cost()`). Therefore, on top of calling `execution_cost(),` `execution_cost_for_authorization()` would need to also call `execution.size_in_bytes()` and serialisation would thus happen twice for the same `Execution`. `execution_cost_with_size()` was introduced in order to avoid this._

4. None of my previous fee-estimation tests involved `Transition`s with input records. Victor helped me find some test `Execution`s involving such `Transition`s in `test_vm_execute_and_finalize.rs`. Since in that test there is only direct access to the `Execution` object (not the `Authorization`), I have created a test-only method
    ```
    #[cfg(feature = "test")]
    impl<N: Network> Authorization<N> {
        pub fn from_unchecked((requests, transitions): (Vec<Request<N>>, Vec<Transition<N>>)) -> Self {
            ...
        }
    }
    ```
    in `synthesizer/process/src/stack/authorization/mod.rs` in order to avoid having to create the actual `Request`s corresponding to the `Transition`s in the `Execution`.

5. Small important plot twist to point 4: the only test with input records in `synthesizer/tests/tests/vm/execute_and_finalize/` is `mint_and_split` (cf. [here](https://github.com/ProvableHQ/snarkVM/blob/78a69b70387d917ed1c16a09741a5b6d16862f17/synthesizer/tests/tests/vm/execute_and_finalize/mint_and_split.aleo#L7-L17)), which never actually consumes the records: the two calls to split are meant to fail for different reasons (visible in the expected output [here](https://github.com/ProvableHQ/snarkVM/blob/78a69b70387d917ed1c16a09741a5b6d16862f17/synthesizer/tests/expectations/vm/execute_and_finalize/mint_and_split.out#L10-L11)). Therefore, a third call to `split` has been added to that test which correctly consumes a record.

With the above modifications, `execute_fee_for_authorization` always predicts the correct cost in all tests (including in the case where a record is consumed).

synthesizer/process/src/tests/test_execute.rs
Antonio95: ... no it wasn't! And addressing that has been surpisingly tricky. Cf. points 4 and 5 in my comment https://github.com/ProvableHQ/snarkVM/pull/2949#issuecomment-3410461968 below.

Even with those fixes, the current tests for fee estimation only involve consumption of one record: this is simply limited by what is covered by the test cases in `synthesizer/tests/tests/vm/execute_and_finalize/`. However, as soon as test cases where more records (possibly of different sizes) are added to that folder, fee estimation (i. e.) `execution_cost_for_authorization` will automatically also be tested on those (caveat: the consensus version of the new test cases must be >= V4).

I should say that, while no records vs some records does significantly impact the tested functionality (one more batch of Varuna instances is added for the inclusion proofs), one records vs. many records is likely not problematic (in the sense that, if `execution_cost_for_authorization` works for the former case, it very likely does for the latter); and record size should have no impact at all (it is unrelated to the size of the Varuna proof). However, I completely agree that "likely" is not good enough - that's precisely why we test things :)

synthesizer/process/src/cost.rs
```diff
@@ -40,16 +43,87 @@ pub fn execution_cost<N: Network>(
...
+    let n_input_records = Authorization::number_of_input_records(authorization.transitions().values());
+    if n_input_records > 0 {
+        batch_sizes.push(n_input_records);
+    }
+
+    // Varuna is always ran in hiding (i. e. ZK) mode when proving Executions.
+    // If future versions of Varuna are introduced, the correct version should
+    // be deduced from the consensus version. The extra 1 byte comes from the
+    // version number.
+    let expected_proof_size = u64::try_from(1 + proof_size::<N::PairingCurve>(&batch_sizes, VarunaVersion::V2, true)?)?;
```
d0cd: Not important for this PR, but we could consider having a map from `ConsensusVersion` to `VarunaVersion`.

d0cd (approved):
Looks good to me overall! 
There are a few minor nits on improving the clarity of comments. Your call on if you think these are worth addressing.

synthesizer/process/src/cost.rs
vicsn: Can you give the magic variable `true` a name? `let hiding_mode = true`;

Is it feasible in `synthesizer/snark/src/proof/` to give `struct Proof` a member function called `proof_size` which calls into `varuna::Proof::proof_size`? Then it's crystal clear which object we're talking about and where "the extra 1 byte" comes from.

vicsn:
Almost there! Besides fixing the last open comments can you merge in staging?

algorithms/src/snark/varuna/data_structures/proof.rs
Antonio95: Addressed

synthesizer/process/src/stack/authorization/mod.rs
Antonio95: (function gone in later commits)

synthesizer/process/src/stack/authorization/mod.rs
Antonio95: (addressed and also fixed other parts of the same documentation that had become outdated due to the interface change)

synthesizer/process/src/trace/mod.rs
Antonio95: Thank you, fixed

synthesizer/process/src/stack/authorization/mod.rs
Antonio95: (function gone in later commits)

algorithms/src/snark/varuna/data_structures/proof.rs
Antonio95: Ah, sorry! I usually write it like that without even thinking :) Will pay more attention when documenting in the future. Fortunately, the comment is gone because there are no longer usize computations in the function ^^

synthesizer/process/src/cost.rs
Antonio95: Absolutely, this is sth that came up when Victor and I were discussing these changes too 👍

Antonio95: Yes! Both things addressed (new function [here](https://github.com/ProvableHQ/snarkVM/blob/795df43e183ce8296c0c2e07739e0caef30b2498/synthesizer/snark/src/proof/mod.rs#L58), let me know if you'd eg prefer a different name). I have also separated the hard-coded Varuna version into a separate variable with its own comment [here](https://github.com/ProvableHQ/snarkVM/blob/795df43e183ce8296c0c2e07739e0caef30b2498/synthesizer/process/src/cost.rs#L128).

synthesizer/snark/src/proof/mod.rs
```diff
@@ -39,3 +41,25 @@ impl<N: Network> Deref for Proof<N> {
...
+///  - `batch_sizes`: the batch sizes of the circuits and instances being
+///    proved.
+///  - `varuna_version`: the version of Varuna being used
+///  - `hiding`: indicates whether the proof system is run in ZK mode
+///
+/// *Returns*:
+///  - `Ok(size)` for `VarunaVersion::V2`, where `size` is the size of the proof
+///    in bytes.
+///  - `Err` for `VarunaVersion::V1`.
+pub fn proof_size<N: Network>(
```
vicsn, line 58: FYI I was thinking having this function as an `impl Proof { pub fn proof_size }` would be even clearer - but this is fine.

vicsn:
I'm still seeing "This branch has conflicts that must be resolved". Can you try: `git fetch origin staging && git merge origin/staging`

Antonio95:
> I'm still seeing "This branch has conflicts that must be resolved". Can you try: `git fetch origin staging && git merge origin/staging`

Yes! It's taken some effort to merge (staging had quite a few changes on the same functions I touched in the `costs.rs` file), but it's done now. I think it's ready to go now!

vicsn (approved):
LGTM!

PR #2950, opened 2025-10-07
[Feat]: Add an 'in' instruction in the VM.

usagi32:
## Motivation

for this issue in leo https://github.com/ProvableHQ/leo/issues/28721

This PR adds a new 'in' instruction support at the synthesizer level. Without this developers would have to do a verbose unroll of the entire array and also instruction costs would increase for contracts for multiple checks, so this saves that.

Some notes: 
- The tests currently only checks for single dimensional literal arrays as for multi dimensional arrays and struct struct arrays, it would require a slight different infrastructure then the rest of the tests. But also there's no reason implementation won't work for all kinds of arrays.
- Maybe the cost function can be optimized more, for e.g. rather than a constant, some heuristic approximation or maybe even linear multiple of the recursive array length, considering all the the inner members in structs etc.

If this PR is approved, I would also add the support in leo as well.

## Test Plan
Added new tests

### DIFF
--- a/synthesizer/program/src/logic/instruction/operation/in_instruction.rs
+++ b/synthesizer/program/src/logic/instruction/operation/in_instruction.rs
@@ -1,0 +1,255 @@
+// Copyright (c) 2019-2025 Provable Inc.
+// This file is part of the snarkVM library.
+
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at:
+
+// http://www.apache.org/licenses/LICENSE-2.0
+
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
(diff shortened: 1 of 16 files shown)


synthesizer/program/src/logic/instruction/operation/in_instruction.rs
```diff
@@ -0,0 +1,256 @@
...
+            )
+        };
+
+        // Make sure the first operand is not an illegal type for this case.
+        let circuit::Value::Plaintext(val) = &input_a else {
+            bail!("Array cannot have records or futures as its elements.")
+        };
+
+        // Check if the array contains the value.
+        let output = (0..array.len()).any(|index| val.is_equal(&array[index]).eject_value());
```
d0cd: I do not believe that this is sufficiently constrained. Generally, `eject_value` needs to be used with care in a circuit. One safe way to implement this is to construct an "or" expression over all of the equality operations, e.g, `(val == arr[0] || val == arr[1] || ... || val == arr[k])`

usagi32: done

PR #2955, opened 2025-10-13
Increase deploy limits by 8x

vicsn:
## Motivation

Developers on Aleo are having trouble with the current deployment constraint and variable limits. These limits were put in place to ensure even if an attacker found an avenue to cheaply spam the network with deployments, they would not be able to increase blocktime by too much. However, they need to be increased to allow for a more expressive AleoVM, several developers noted that they are hitting the limits.

Instead of the current transaction-level limits, or proposal-level limits (like we did for [finalize spend](https://github.com/ProvableHQ/snarkOS/pull/3471)), this PR proposes a much higher block-level limit.

The main downside of this approach is that it hurts predictability. A deployment might be aborted last-minute because another user deployed at the same time. Note however:
- whereas the Aleo network processes many executions, there are only [a few hundred programs](https://explorer.provable.com/programs) deployed on mainnet. Simultaneous deployment is unlikely.
- a deployment's "end user" is always a developer, from whom we can expect more expertise. `leo deploy` can already check whether a deployment was accepted - we can also let it check whether a competing deployment was accepted recently.

A nice side-effect of this PR is that it makes DoS attacks with many deployments harder.

A large downside is that encouraging this will increase prover time.

## Appendix: how the original limit was set

Deployment verification consists of a mix of single-threaded logic (R1CS synthesis) and multi-threaded logic (polynomial interpolation and evaluation). The original price of 25 microcredits per constraint focused on just the single-threaded runtime - allowing us to be in line with the 10 credits/second across our finalize opcodes. We don't have immediately reproducible benchmarks available unfortunately, this PR does not aim to touch that "constraints per second" metric.

## Test Plan

- [ ] After the design is approved, I'll add unit tests which deploy below and above the limit.

### DIFF
--- a/console/network/src/consensus_heights.rs
+++ b/console/network/src/consensus_heights.rs
@@ -41,6 +41,8 @@ pub enum ConsensusVersion {
     V10 = 10,
     /// V11: Support for external structs.
     V11 = 11,
+    /// V12: Support for block-level deployment constraints and variables limits.
+    V12 = 12,
 }
 
 impl ConsensusVersion {
@@ -50,7 +52,7 @@ impl ConsensusVersion {
 }
 
 /// The number of consensus versions.
-pub(crate) const NUM_CONSENSUS_VERSIONS: usize = 11;
+pub(crate) const NUM_CONSENSUS_VERSIONS: usize = 12;
 
 /// The consensus version height for `CanaryV0`.
 pub const CANARY_V0_CONSENSUS_VERSION_HEIGHTS: [(ConsensusVersion, u32); NUM_CONSENSUS_VERSIONS] = [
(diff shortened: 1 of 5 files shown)


vicsn:
Converted to draft because developers in question may need to be encouraged to use more efficient hash functions and finalize scope operations

vicsn:
A note from me: this change is actually still desirable to address DoS attacks

A note from @Antonio95 : note that FFT takes O(N log N) so ideally we do run new benchmarks for verification of the largest allowed deployments

Antonio95:
A corrected note from me (cf. previous comment): the "interpolate over Lagrange bases + evaluate at a point" we do during synthesis is not an FFT but has a linear-time algorithm.

PR #2957, opened 2025-10-14
Update ConsensusVersion::V11 heights

vicsn:
## Motivation

Based on the current height and last 7 days network speed


niklaslong (approved):
LGTM pending confirmation on the canary block time.

vicsn:
> LGTM pending confirmation on the canary block time.

Average blocktime indeed is a surprisingly slow 7 seconds on Canary.
The current values assume 4 seconds, but that is only all the more cautious. Instead of the migration kicking in after 3 days, it will kick in in ~6 days.

PR #2960, opened 2025-10-15
Add test to prevent regression of ConsensusVersion::read_le

vicsn:
# Motivation

We have to manually extend the logic in ConsensusVersion::read_le

This adds a test that we don't forget to add new versions in there

### DIFF
--- a/console/network/src/consensus_heights.rs
+++ b/console/network/src/consensus_heights.rs
@@ -391,6 +391,11 @@ mod tests {
         let result = ConsensusVersion::from_bytes_le(&bytes).unwrap();
         assert_eq!(result, version);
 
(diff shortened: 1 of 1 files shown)


kaimast:
That's a good idea!

Side note: it is really awkward that there is no better way to implement deserialization of such a simple type. `ConsensusVersion` has a `u16` representation and you can just write `version as u16` to serialize it, but there seems to be no standard way to transform an integer back into an enum.
Initially, I hoped to do something like the code below to deserialize it. Some third-party crates implement this, but I didn't want to add another dependency. 

```rust
let version: ConsensusVersion = u16::read_le(&mut data)?.try_into()?
```

kaimast (approved):

PR #2962, opened 2025-10-15
Add nested array test

vicsn:
## Motivation

Adds a test for maximally nested arrays of maximum size.
Also in the previous commit tested a related program with as many arrays of maximum size as possible.

No wild memory usage or runtime was observed.

### DIFF
--- a/synthesizer/src/vm/mod.rs
+++ b/synthesizer/src/vm/mod.rs
@@ -3448,4 +3448,146 @@ function adder:
         // Check that the program was deployed.
         assert!(vm.process().read().contains_program(&ProgramID::from_str("adder_program.aleo").unwrap()));
     }
+
+    #[test]
+    #[ignore = "memory-intensive"]
+    fn test_deployment_array_allocation_limit() {
+        use crate::test_helpers::CurrentNetwork;
+        
+        let rng = &mut TestRng::default();
+
+        // Initialize a private key.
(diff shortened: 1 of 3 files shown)


synthesizer/src/vm/mod.rs
```diff
@@ -3448,4 +3448,232 @@ function adder:
...
+                None,
+                rng,
+            )
+            .unwrap();
+
+        // Verify the execution
+        vm.check_transaction(&transaction, None, rng).unwrap();
+
+        println!(
+            "Successfully deployed and executed program with array nesting depth {} and {} elements per dimension",
```
d0cd, line 3609: Overall looks great! A couple points:
- It could be helpful to assert the max depth here just so we "fix" the actual size that was tested.
- It could be helpful to test in a finalize context and assert the cost.
- Also as a standard pattern, we should assert that the execution transactions were in fact accepted.

d0cd (approved):
Thanks for the test! Left some minor feedback, feel free to either address or merge the PR and I'm happy to make tweaks in `test/ecdsa-bytes`

synthesizer/src/vm/mod.rs
vicsn: Great suggestions.

e7ce898e1

Only thing I didn't do is assert the cost, feels out of scope

Issue #2871, opened 2025-08-14
[Perf] Speed up batch instance proof verification


vicsn:
## 💥 Proposal

In November we'll want to land a feature which may require proof verification with a batch instance size of up to 7968.

Currently, verifying such large instance sizes takes 2.5 seconds. How much can we get this down?

A proof of concept can be found in snarkVM branch `bench_large_instance_verifier`.

You can run a test for the relevant `fn verify_batch` function with:
```
cd algorithms/src/snark/varuna && cargo test prove_and_verify_with_tall_matrix_big -- --nocapture
```

vicsn:
I did some initial analysis on this.

Attached is a flamegraph made by:
- checking out `perf/verify_batch`
- running `cargo test prove_and_verify_with_tall_matrix_big --features serial -- --nocapture`
- waiting for the pid to be printed, and then running `sudo flamegraph --pid <pid>` using the utility installed by `cargo flamegraph`

![Image](https://github.com/user-attachments/assets/ee319a8c-8e67-4f88-baad-c5c68e254459)

The entire cost is basically Poseidon, so we should try and get that down.

We have some unmerged prior work on this topic: https://github.com/ProvableHQ/snarkVM/pull/2155

vicsn:
Ideas, which all need to be guarded by a new `VarunaVersion`

## Parallelize hash computation every time we absorb.

During the calls to `init_sponge` and `absorb*`, do something like:

```
let circuit_commitments_hashes = keys_to_inputs.par_iter().map(|(vk, inputs)| {
    let sponge = Self::init_sponge()?;
    sponge.absorb_native_field_elements(vk.circuit_commitments.as_slice());
    sponge.absorb_native_field_elements(inputs);
    sponge.squeeze_nonnative_field_elements(1).pop().ok_or(anyhow!("Failed to squeeze random element"))?
}).collect_vec();
let mut sponge = Self::init_sponge(fs_parameters, &inputs_and_batch_sizes, circuit_commitments_hashes.clone());
```

## Properly batch the RatSumCheck

## Implement a faster ZK-friendly hash function

Poseidon2 might be a good contender if it is very similar to the Poseidon we use. Or we go for a non-ZK-friendly hash function...

## Implement a non-ZK friendly hash function outside of the circuit

vicsn:
@Antonio95 I'm also curious whether there are any ways to more efficiently batch verify **separate** proofs

Antonio95:
[...]

PR #2964, opened 2025-10-16
[Perf] Only calculate the hash of circuit commitments once for the VK

ljedrz:
This addresses the 1st part of [this comment](https://github.com/ProvableHQ/snarkVM/issues/2871#issuecomment-3302769764) on the linked issue.

CC https://github.com/ProvableHQ/snarkVM/issues/2871.

### DIFF
--- a/algorithms/benches/snark/varuna.rs
+++ b/algorithms/benches/snark/varuna.rs
@@ -107,6 +107,7 @@ fn snark_batch_prove(c: &mut Criterion) {
 
         let mut pks = Vec::with_capacity(circuit_batch_size);
         let mut all_circuits = Vec::with_capacity(circuit_batch_size);
+        #[allow(clippy::mutable_key_type)]
         let mut keys_to_constraints = BTreeMap::new();
 
         for i in 0..circuit_batch_size {
@@ -191,7 +192,9 @@ fn snark_batch_verify(c: &mut Criterion) {
         let mut vks = Vec::with_capacity(circuit_batch_size);
         let mut all_circuits = Vec::with_capacity(circuit_batch_size);
         let mut all_inputs = Vec::with_capacity(circuit_batch_size);
+        #[allow(clippy::mutable_key_type)]
         let mut keys_to_constraints = BTreeMap::new();
+        #[allow(clippy::mutable_key_type)]
         let mut keys_to_inputs = BTreeMap::new();
         for i in 0..circuit_batch_size {
             let num_constraints = num_constraints_base + i;
--- a/algorithms/src/snark/varuna/data_structures/circuit_verifying_key.rs
+++ b/algorithms/src/snark/varuna/data_structures/circuit_verifying_key.rs
@@ -13,7 +13,7 @@
 // See the License for the specific language governing permissions and
 // limitations under the License.
 
-use crate::{polycommit::sonic_pc, snark::varuna::ahp::indexer::*};
+use crate::{AlgebraicSponge, polycommit::sonic_pc, snark::varuna::ahp::indexer::*};
 use snarkvm_curves::PairingEngine;
 use snarkvm_utilities::{FromBytes, FromBytesDeserializer, ToBytes, ToBytesSerializer, into_io_error, serialize::*};
 
@@ -25,10 +25,11 @@ use std::{
     io::{self, Read, Write},
     str::FromStr,
     string::String,
+    sync::OnceLock,
 };
 
 /// Verification key for a specific index (i.e., R1CS matrices).
-#[derive(Debug, Clone, PartialEq, Eq, CanonicalSerialize, CanonicalDeserialize)]
+#[derive(Debug, Clone, PartialEq, Eq)]
 pub struct CircuitVerifyingKey<E: PairingEngine> {
     /// Stores information about the size of the circuit, as well as its defined
     /// field.
@@ -36,6 +37,21 @@ pub struct CircuitVerifyingKey<E: PairingEngine> {
     /// Commitments to the indexed polynomials.
     pub circuit_commitments: Vec<sonic_pc::Commitment<E>>,
     pub id: CircuitId,
+    pub circuit_commitments_hash: OnceLock<E::Fq>,
+}
+
+impl<E: PairingEngine> CircuitVerifyingKey<E> {
+    pub fn get_or_calculate_circuit_commitments_hash<FS: AlgebraicSponge<E::Fq, 2>>(
+        &self,
+        fs_parameters: &FS::Parameters,
+    ) -> &E::Fq {
+        self.circuit_commitments_hash.get_or_init(|| {
+            let mut sponge = FS::new_with_parameters(fs_parameters);
+            sponge.absorb_native_field_elements(&self.circuit_commitments);
+
+            sponge.squeeze_native_field_elements(1)[0]
+        })
+    }
 }
 
 impl<E: PairingEngine> FromBytes for CircuitVerifyingKey<E> {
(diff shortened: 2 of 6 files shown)


algorithms/benches/snark/varuna.rs
```diff
@@ -107,6 +107,7 @@ fn snark_batch_prove(c: &mut Criterion) {
 
         let mut pks = Vec::with_capacity(circuit_batch_size);
         let mut all_circuits = Vec::with_capacity(circuit_batch_size);
+        #[allow(clippy::mutable_key_type)]
```
ljedrz, line 110: note: this is needed, because the `CircuitVerifyingKey` has interior mutability now; however, in its case it is perfectly fine, as the new member has no impact on the `Ord` impl (which only considers the `id`), and the fact that the key implements it is the reason why the warning is raised; see the corresponding `clippy` lint [link](https://rust-lang.github.io/rust-clippy/master/index.html#mutable_key_type).

algorithms/src/snark/varuna/data_structures/circuit_verifying_key.rs
```diff
@@ -25,17 +25,33 @@ use std::{
...
-#[derive(Debug, Clone, PartialEq, Eq, CanonicalSerialize, CanonicalDeserialize)]
+#[derive(Debug, Clone, PartialEq, Eq)]
 pub struct CircuitVerifyingKey<E: PairingEngine> {
     /// Stores information about the size of the circuit, as well as its defined
     /// field.
     pub circuit_info: CircuitInfo,
     /// Commitments to the indexed polynomials.
     pub circuit_commitments: Vec<sonic_pc::Commitment<E>>,
     pub id: CircuitId,
+    pub circuit_commitments_hash: OnceLock<E::Fq>,
```
vicsn, line 40: I think this is expensive enough that we should store it to disk - and it would be great if we can get rid of the OnceLock (which obfuscates when initialization happens, making performance analysis harder). O:) 

You correctly observed that we don't have to transmit it over the wire though.

vicsn:
Can you report the performance improvements for the existing `snark_batch_verify` benchmark with:
- circuit_batch_size:1, instance_size:5
- circuit_batch_size:1, instance_size:1000

algorithms/src/snark/varuna/varuna.rs
```diff
@@ -134,10 +135,10 @@ impl<E: PairingEngine, FS: AlgebraicSponge<E::Fq, 2>, SM: SNARKMode> VarunaSNARK
         Ok(circuit_keys)
     }
 
-    fn init_sponge<'a>(
+    fn init_sponge(
         fs_parameters: &FS::Parameters,
         inputs_and_batch_sizes: &BTreeMap<CircuitId, (usize, &[Vec<E::Fr>])>,
-        circuit_commitments: impl Iterator<Item = &'a [crate::polycommit::sonic_pc::Commitment<E>]>,
+        circuit_commitments_hashes: Vec<E::Fq>,
```
vicsn, line 156: Can you make a 3rd VarunaVersion, and guard the changes by it to preserve backwards compatibility? You can peek at where we use VarunaVersion::V2 for inspiration.

algorithms/src/snark/varuna/data_structures/circuit_verifying_key.rs
ljedrz: The issue was that at the moment of creation, the `fs_params` is not available, and at later stages the VKs are immutable - hence the `OnceLock`.

Storing to the disk would probably work around this, but retrieving it would be expensive, perhaps to the point of offsetting any performance gains from caching it, unless the hashing is **really** computationally expensive.

vicsn: > Storing to the disk would probably work around this, but retrieving it would be expensive, perhaps to the point of offsetting any performance gains from caching it, unless the hashing is really computationally expensive.

To be clear, we would retrieve it from disk only when we retrieve the VK from disk. And the hashes are very expensive.

However, a big downside I do see is the sheer amount of work to adjust the database logic. So for the first version you can also compute it during construction.

> The issue was that at the moment of creation, the fs_params is not available

Looks to me it's always available in N::varuna_fs_parameters() ? Or is there a scoping issue? Have fun with that :")

ljedrz:
> circuit_batch_size:1, instance_size:5

```
snark_batch_verify      time:   [4.6137 ms 4.6239 ms 4.6343 ms]
                        change: [−5.4575% −5.1305% −4.8178%] (p = 0.00 < 0.05)
                        Performance has improved.
Found 1 outliers among 100 measurements (1.00%)
  1 (1.00%) high mild
```
> circuit_batch_size:1, instance_size:1000

```
snark_batch_verify      time:   [149.98 ms 150.09 ms 150.22 ms]
                        change: [−0.1857% −0.0640% +0.0545%] (p = 0.32 > 0.05)
                        No change in performance detected.
Found 4 outliers among 100 measurements (4.00%)
  2 (2.00%) high mild
  2 (2.00%) high severe
```

algorithms/src/snark/varuna/data_structures/circuit_verifying_key.rs
ljedrz: > Looks to me it's always available in N::varuna_fs_parameters()?

That's correct; however, there is no notion of the `Network` - or even `snarkvm-console` - in `algorithms`. Would it be acceptable to alter `SNARK::circuit_setup` to require `FSParameters`, like the other `VarunaSNARK` functions do?

vicsn:
Thank you for the benchmarks, looks like I made a mistake...

This PR is only consequential when we have a very large amount of different circuits - but in the current system and even with dynamic dispatch we're never planning to have more than 32 circuits to batch - and the most common case is just 3 circuits.

Can you do one more benchmark with circuit_size: 3, instance_size: 10, before and after this PR? If the difference is inconsequential we can close this PR and stop this line of work.

ljedrz:
sure:
```
snark_batch_verify      time:   [10.654 ms 10.669 ms 10.683 ms]
                        change: [−6.3521% −6.1469% −5.9447%] (p = 0.00 < 0.05)
                        Performance has improved.
```

vicsn:
> circuit_batch_size:1, instance_size:1000

Could you maybe run this comparison again? Perhaps even try instance_size 4000? 

If it is indeed a negligible difference, could you reproduce the single-threaded flamegraph I made to revisit whether initi_sponge hashing is really a bottleneck? https://github.com/ProvableHQ/snarkVM/compare/staging...perf/verify_batch#diff-9ae02456754b316d993bd81a0e311bff5e432156b69d28b867d5c98027d0b79bR137

ljedrz:
> Perhaps even try instance_size 4000?
```
snark_batch_verify      time:   [586.57 ms 586.99 ms 587.43 ms]
                        change: [+0.2159% +0.3150% +0.4199%] (p = 0.00 < 0.05)
```

ljedrz:
> If it is indeed a negligible difference, could you reproduce the single-threaded flamegraph I made to revisit whether initi_sponge hashing is really a bottleneck?

Done; it does show that `init_sponge` takes ~43%, but this PR doesn't seem to impact it (only ~0.4% decrease in that value).

vicsn:
TIL again that dynamic dispatch *does* create potentially lots of circuits, we can keep this open as a draft.

algorithms/src/snark/varuna/data_structures/circuit_verifying_key.rs
vicsn: > However, a big downside I do see is the sheer amount of work to adjust the database logic. So for the first version you can also compute it during construction.

For my future self reading this: we should just store to disk, but we can write out that logic when we have a definite timeline for landing this feature.

vicsn:
@ljedrz do you want to build and benchmark another feature to this draft. Currently we hash `inputs_and_batch_sizes` inside of `fn init_sponge` using the expensive Poseidon hash. Instead, the prover and verifier can hash those to a single field element using `hash_sha3_256`, and then pass that single value into `fn {prove_verify}_batch` and into `fn init_sponge` to be hashed by Poseidon

ljedrz:
The [most recent commit](https://github.com/ProvableHQ/snarkVM/pull/2964/commits/e316f803ff218944690b53611863d3f3dc0cbbe9) provides the following benchmark wins compared with the previous one:

> circuit_batch_size:1, instance_size:5
```
nark_batch_verify      time:   [7.2229 ms 7.2641 ms 7.3056 ms]
                        change: [−5.8052% −5.1281% −4.4218%] (p = 0.00 < 0.05)
                        Performance has improved.
```

> circuit_batch_size:1, instance_size:1000
```
snark_batch_verify      time:   [162.82 ms 163.37 ms 163.95 ms]
                        change: [−27.720% −27.366% −27.010%] (p = 0.00 < 0.05)
                        Performance has improved.
```

PR #2965, opened 2025-10-17
Update PR template

vicsn:
I'll edit snarkOS after this is approved.

### DIFF
--- a/.github/PULL_REQUEST_TEMPLATE.md
+++ b/.github/PULL_REQUEST_TEMPLATE.md
@@ -10,8 +10,8 @@ If you changed any code, please provide us with clear instructions on how you ve
 
 ## Documentation
 
-If this PR adds or changes functionality, please take some time to update any docs, e.g. on github.com/AleoNet/welcome.
+If this PR adds or changes functionality, consider clarifying which docs need to be updated, e.g. on [AleoNet/welcome](https://github.com/AleoNet/welcome).
 
 ## Backwards compatibility
 
-Please review backwards compatibility. Does any functionality need to be guarded by an existing or new ConsensusVersion?
+Please review backwards compatibility. Does any functionality need to be guarded by an existing or new `ConsensusVersion`?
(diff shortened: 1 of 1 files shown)


ljedrz (approved):
The link under **Documentation** could be a real link, and the type name under **Backwards compatibility** could be formatted as code, but that's cosmetics - LGTM.

.github/PULL_REQUEST_TEMPLATE.md
```diff
@@ -1,24 +1,17 @@
...
-    If you changed any code, please provide us with clear instructions on how you verified your changes work.
-    Bonus points for screenshots and videos!
--->
+If you changed any code, please provide us with clear instructions on how you verified your changes work. Bonus points for screenshots and videos!
 
-(Write your test plan here)
+## Documentation
 
-## Related PRs
+If this PR adds or changes functionality, please take some time to update any docs, e.g. on github.com/AleoNet/welcome.
```
Antonio95: Do we maybe want to tone this down or rephrase it?
- I think some people may be put off to make a contribution if they realise they also need to update "any docs", which may be a bit daunting for users not deep into the codebase
- To be fair, I guess many contributors will only discover this _after_ they have implemented the changes and that might make them a bit more willing to put in the extra work and look into documenting things.
- I also understand documentation definitely needs to be kept updated and if contributors don't do that, we will have to ourselves, which can be a non-trivial amount of work.

Just some thoughts.

.github/PULL_REQUEST_TEMPLATE.md
```diff
@@ -1,24 +1,17 @@
...
 
-<!--
-    If this PR adds or changes functionality, please take some time to
-    update the docs at https://github.com/ProvableHQ/snarkVM and https://github.com/AleoHQ/protocol-docs,
-    and link to your PR here.
--->
+## Backwards compatibility
 
-(Link your related PRs here)
+Please review backwards compatibility. Does any functionality need to be guarded by an existing or new ConsensusVersion?
```
Antonio95: Similar thing here: isn't that something the maintainers (us) are better equipped to assess?

Antonio95 (approved):
Two minor comments (written from the perspective that the message is also geared  towards external contributors, which may not be the case)

.github/PULL_REQUEST_TEMPLATE.md
vicsn: The PR template is both for internal and external contributors

.github/PULL_REQUEST_TEMPLATE.md
vicsn: toned it down

PR #2969, opened 2025-10-20
[Fix] Always return version 0 for credits.aleo

### LINKED ISSUE #3804: Consistent handling of credits.aleo edition in REST API
This is a place for a discussion related to https://github.com/ProvableHQ/snarkOS/pull/3803 that @vicsn asked me to create.

Basically the discussion should be how the "/program/credits.aleo/latest_edition" endpoint should behave (the case is specific for "credits.aleo" here).

The current hardcoded behaviour is to return status 500 because of this code in ledger/store/src/transaction/deployment.rs:

```
    fn get_latest_edition_for_program(&self, program_id: &ProgramID<N>) -> Result<Option<u16>> {
        // Check if the program ID is for 'credits.aleo'.
        // This case is handled separately, as it is a default program of the VM.
        // TODO (howardwu): After we update 'fee' rules and 'Ratify' in genesis, we can remove this.
        if program_id == &ProgramID::from_str("credits.aleo")? {
            return Ok(None);
        }

        Ok(self.edition_map().get_confirmed(program_id)?.map(|x| *x))
    }
```

and this code in ledger/src/get.rs

```
    pub fn get_latest_edition_for_program(&self, program_id: &ProgramID<N>) -> Result<u16> {
        match self.vm.block_store().get_latest_edition_for_program(program_id)? {
            Some(edition) => Ok(edition),
            None => bail!("Missing latest edition for program ID {program_id}"),
        }
    }
```

kaimast:
Aims to fix [snarkOS #3804](https://github.com/ProvableHQ/snarkOS/issues/3804).

**Side note:** I am not sure if I understand the TODO left by Howard, but it would be great if we did not have to construct the `ProgramID` for `credits.aleo` on every call to these functions.

### DIFF
--- a/ledger/src/get.rs
+++ b/ledger/src/get.rs
@@ -260,6 +260,12 @@ impl<N: Network, C: ConsensusStorage<N>> Ledger<N, C> {
         }
     }
 
+    /// Returns the program for the given `program ID` and `edition`,
+    /// or `None` if no program of this ID and edition exists.
+    pub fn try_get_program_for_edition(&self, program_id: ProgramID<N>, edition: u16) -> Result<Option<Program<N>>> {
+        self.vm.block_store().get_program_for_edition(&program_id, edition)
+    }
+
     /// Returns the block solutions for the given block height.
     pub fn get_solutions(&self, height: u32) -> Result<Solutions<N>> {
         // If the height is 0, return the genesis block solutions.
--- a/ledger/src/get.rs
+++ b/ledger/src/get.rs
@@ -254,7 +254,7 @@ impl<N: Network, C: ConsensusStorage<N>> Ledger<N, C> {
 
     /// Returns the program for the given `program ID` and `edition`.
     pub fn get_program_for_edition(&self, program_id: ProgramID<N>, edition: u16) -> Result<Program<N>> {
-        match self.vm.block_store().get_program_for_edition(&program_id, edition)? {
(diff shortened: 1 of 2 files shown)


kaimast:
@d0cd you recently made changes to how program editions are handled. Do you think this change would break anything upstream?

vicsn:
Thx. Just so we're all on the same page, can you list in the PR description which endpoints you called to test and what the expected behaviour is?

kaimast:
> Thx. Just so we're all on the same page, can you list in the PR description which endpoints you called to test and what the expected behaviour is?

I ran a node without that uses snarkVM staging and queries the latest edition endpoint, which gave me the output below.
```
~> curl http://localhost:3030/v2/testnet/program/credits.aleo/latest_edition
{"message":"Missing latest edition for program ID credits.aleo","error_type":"internal_server_error"}
```

Then, I ran with this branch and did the same. It now returns 0 as the latest edition.
```
~> curl http://localhost:3030/v2/testnet/program/credits.aleo/latest_edition
0
```

kaimast:
I changed the PR slightly. It now simply inserts `credits.aleo` into the deployment store on startup, and then does not treat it differently than other programs for `get_program_for_edition`, `get_latest_program_edition`, and `get_latest_program`.

vicsn (approved):

PR #2970, opened 2025-10-20
Properly log rejection reasons and critical error in VM

vicsn:
## Motivation

It is critical for users to understand why transactions were rejected. Doing this properly will require a longer discussion (see [this issue](https://github.com/ProvableHQ/snarkOS/issues/3934)) but currently we don't even surface the information locally.

This PR is as small as I could make it, adding a few trace! and error! logs - unfortunately it has some indenting changes.

Note that aborted tx reasons are already logged here:
```
                    warn!("Speculation safely aborted a transaction - {error} ({tx_id})");
```

## Test Plan

CI should suffice.

### DIFF
--- a/synthesizer/src/vm/finalize.rs
+++ b/synthesizer/src/vm/finalize.rs
@@ -16,7 +16,7 @@
 use super::*;
 
 use snarkvm_ledger_committee::{MAX_DELEGATORS, MIN_DELEGATOR_STAKE, MIN_VALIDATOR_SELF_STAKE};
-use snarkvm_utilities::{cfg_sort_by_cached_key, defer, dev_eprintln};
+use snarkvm_utilities::{cfg_sort_by_cached_key, defer};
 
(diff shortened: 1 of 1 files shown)


raychu86:
The logs here aren't necessarily explicit regarding if a TX was rejected. 

A generic `Failed to finalize deploy/execute tx {} - {error}` message will be displayed, however this can still lead to an aborted TX. 

So for the node operator to know if a TX was rejected, they need to check that a subsequent `Speculation safely aborted a transaction` message does not exist.

vicsn:
Thx, I think that it's fine for now, as I'm not aware of anyone who is interested in these logs being read out in any automated way, and neither is the explorer. But it at least gives operators the option to search for rejected txids in the logs, and then the right information will pop up.

Making the log truly unique to only rejected messages can be follow-up work but is also a lot more work so can merge with https://github.com/ProvableHQ/snarkOS/issues/3934

raychu86 (approved):

PR #2971, opened 2025-10-21
[Perf] Block Caching

kaimast:
Cleaned up version of #2937 to address #2934. There is a https://github.com/ProvableHQ/snarkOS/pull/3923 in snarkOS that uses this new feature.

### Overview
The new block cache is part of the `BlockStore` and caches the 10 most recent blocks. It is always enabled and there is no API change, except for the addition of `Ledger::get_cache_size`.

### DIFF
--- a/ledger/src/lib.rs
+++ b/ledger/src/lib.rs
@@ -70,7 +70,7 @@ use aleo_std::{
     StorageMode,
     prelude::{finish, lap, timer},
 };
-use anyhow::Result;
+use anyhow::{Context, Result};
 use core::ops::Range;
 use indexmap::IndexMap;
 #[cfg(feature = "locktick")]
@@ -224,11 +224,11 @@ impl<N: Network, C: ConsensusStorage<N>> Ledger<N, C> {
 
         // Retrieve the latest height.
         let latest_height =
-            ledger.vm.block_store().max_height().ok_or_else(|| anyhow!("Failed to load blocks from the ledger"))?;
+            ledger.vm.block_store().max_height().with_context(|| "Failed to load blocks from the ledger")?;
         // Fetch the latest block.
         let block = ledger
             .get_block(latest_height)
-            .map_err(|err| err.context("Failed to load block {latest_height} from the ledger"))?;
+            .with_context(|| format!("Failed to load block {latest_height} from the ledger"))?;
 
         // Set the current block.
         ledger.current_block = Arc::new(RwLock::new(block));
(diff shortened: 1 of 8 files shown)


vicsn:
Given the scope creep in this PR I'm wondering again what the exact goal is we're pursuing. Looking at the original issue, it is unclear what the exact benefit is. Which is my bad for not documenting it in the issue when we had the conversation. 

For which code paths is the snarkOS block cache insufficient? At a quick glance I only see `get_block_round` which would have been a 4-line change in snarkOS.

UPDATE: REST endpoint, and it's future proofing.

ledger/src/lib.rs
```diff
@@ -90,6 +90,33 @@ pub type RecordMap<N> = IndexMap<Field<N>, Record<N, Plaintext<N>>>;
...
+impl LedgerOptions {
+    /// Disables checks at startup.
+    /// This can increase performance, but is less safe.
+    pub fn disable_startup_checks(self) -> Self {
+        Self { startup_checks: false, block_cache: self.block_cache }
+    }
+
+    /// Enables caching of the most recent blocks.
+    /// This can increase performance, but consumes more memory.
+    pub fn enable_block_cache(self) -> Self {
```
vicsn: Why would we want to set this to false?

ledger/src/lib.rs
```diff
@@ -90,6 +90,33 @@ pub type RecordMap<N> = IndexMap<Field<N>, Record<N, Plaintext<N>>>;
 /// The capacity of the LRU cache holding the recently queried committees.
 const COMMITTEE_CACHE_SIZE: usize = 16;
 
+/// Options
+pub struct LedgerOptions {
```
vicsn: Can we get rid of this added abstraction? It makes the code harder to maintain

ledger/src/lib.rs
kaimast: I thought there were use cases of snarkVM where the cache is not needed. I would also prefer to always enable it. Should I go ahead with that?

vicsn: Yes please!

kaimast:
I reduced the PR to around 300 lines without tests.

ledger/store/src/block/cache.rs
```diff
@@ -0,0 +1,222 @@
...
+    /// Invariant: all entries in this vector (except the first) a height equal to h+1 (where h` is the previous block entry)
+    blocks: VecDeque<Block<N>>,
+}
+
+impl<N: Network> BlockCache<N> {
+    /// The maximum size of the cache in blocks.
+    pub(super) const BLOCK_CACHE_SIZE: u32 = 10;
+
+    /// Initialize the cache with the given blocks.
+    pub fn new(blocks: Vec<Block<N>>) -> Result<Self> {
```
vicsn, line 38: Should we ensure we receive max 10?

vicsn (approved):
LGTM! one nit. Please update the PR description

PR #2973, opened 2025-10-21
[CI] Update baseline revision for semver checks

kaimast:
This PR (once again) bumps the baseline revision for semver checks.

The current errors were:
* enum variant added for `ConsensusVersion::V12`
* changes in `log_warning`/`log_error`: not used outside snarkOS/VM AFAIK
* removal of the `tracing` feature in `ledger/store`: that dependency is now always enabled


vicsn (approved):
Thank you for your service!

Issue #2954, opened 2025-10-13
[Proposal] Reduce the reliance on locks in consensus


ljedrz:
## 💥 Proposal

We are currently relying on multiple locks (`atomic_block`, `block_lock`, and likely more related ones in snarkOS) in order to ensure that ledger updates are performed sequentially. The ledger itself also expects all its writes (even speculative ones) to occur sequentially, and will throw an error if any such operations are attempted concurrently. It's becoming tricky to introduce performance improvements while having to juggle multiple lock guards.

Specifically, we hit database asserts when trying to implement some of the improvements in: https://github.com/ProvableHQ/snarkOS/issues/3960

I propose that we introduce some form of a queue (most likely a new channel) that would allow us to enqueue pending storage operations from any entry point and to process them one after the other. That way, we could remove the `atomic_block`, `block_lock`, and perhaps even introduce some true exclusive mutability via `&mut` to the workflow.

Considerations:
- I've found 3 places where we currently perform writes to the storage (the entry points to `start_atomic`):
  - `try_advance_to_next_block/prepare_advance_to_next_quorum_block/speculate/atomic_speculate`
  - `try_advance_to_next_block/check_next_block/check_next_block/atomic_speculate`
  - `try_advance_to_next_block/advance_to_next_block/advance_to_next_block` (both directly and through `atomic_finalize`)

  Are there any more?
- at which level should these operations be introduced? `try_advance_to_next_block` is too broad, but `atomic_speculate` might be too granular (it might be practical to know if it's block-checking or quorum block preparation stage)
- how would these operations be represented? a dedicated beefy `enum` that would be able to hold all the items applicable to the operations of interest, and fed to a channel where we have a shallow clone of the `Ledger`
- where would the channel be spawned? would it be controlled by snarkOS or snarkVM?

vicsn:
CC @kaimast what do you think?

> It's becoming tricky to introduce performance improvements while having to juggle multiple lock guards.

Is there also a security component, would the rewrite make it easier to prevent "assertion failed: !already_paused" from happening or is that a different topic altogether?

kaimast:
Wouldn't a channel behave the same as a top-level lock? It would sequentially order all database operations and, thus, restrict them to one OS thread. Channels will introduce another layer of indirection we have to deal with when we have to debug or add new features in the future.

[...]

PR #2975, opened 2025-10-23
[Feat] Ensure sequential processing of potential Ledger writes

ljedrz:
This is a proposal tackling https://github.com/ProvableHQ/snarkVM/issues/2954. It introduces a background thread responsible for the sequential processing of storage-related operations that cannot happen concurrently, in order to allow us to remove some of the related locks, to be able to introduce more performance optimizations related to block syncing, and in general be able to reason about the storage with more confidence.

Most urgently, this PR should ensure we avoid hitting database asserts which we repeatedly encountered when implementing improvements to consensus listed here: https://github.com/ProvableHQ/snarkOS/issues/3960

During prototyping, the following open questions were resolved:
> Are there any more [ledger operations with potential writes]?

I have found only one more operation not listed in the issue, which is only triggered when a dev-mode node creates a genesis block. That being said, it followed one of the call paths detected previously, so it required no special treatment.

> at which level should these operations be introduced? `try_advance_to_next_block` is too broad, but `atomic_speculate` might be too granular (it might be practical to know if it's block-checking or quorum block preparation stage)

In the end I decided to isolate 2 `SequentialOperation`s:
- `AddNextBlock`
- `AtomicSpeculate`

There is also `atomic_finalize` which may not be triggered concurrently, but it is only called as part of `AddNextBlock`, so I decided to only introduce a new safeguard (`ensure_sequential_processing`) in it.

> how would these operations be represented? a dedicated beefy enum that would be able to hold all the items applicable to the operations of interest, and fed to a channel where we have a shallow clone of the `Ledger`

It's a new `enum` (plus some helper objects), and in the end it doesn't seem too bad size-wise; as for the channel, it actually contains a shallow clone of the `VM` instead, as that's the actual entry point into the storage.

> where would the channel be spawned? would it be controlled by snarkOS or snarkVM?

I decided to spawn the dedicated thread when creating the `VM` - it is owned by it, together with the sender to the channel for transferring the operations.

Future/follow-up considerations:
- while this proposal solves the general issue for the set of operations we currently use with a small diff and little impact on the existing APIs, it is not foolproof - we would have to manually ensure that any potential future operations (however unlikely) that may not be executed concurrently use this new setup; a better API might be possible, but I haven't come up with one yet
- the persistent storage could inherit the thread id from the `VM`, and it could be the one to perform the `ensure_sequential_processing` check, probably in `start_atomic`; it would future-proof this setup, as we would immediately know if any new write batch was introduced, but not being processed sequentially
- this new setup could technically allow us to remove multiple locks from the storage-related objects, which would be a performance improvement and a simplification (as fewer checks may be needed now)

Filing this as a draft, as I haven't run all the tests locally yet, and it would require a stress-test run on snarkOS side.


synthesizer/tests/test_vm_execute_and_finalize.rs
```diff
@@ -55,7 +54,7 @@ fn test_vm_execute_and_finalize() {
         load_tests::<_, ProgramTest>("./tests/vm/execute_and_finalize", "./expectations/vm/execute_and_finalize");
 
     // Run each test and compare it against its corresponding expectation.
-    tests.par_iter().for_each(|test| {
+    tests.iter().for_each(|test| {
```
ljedrz, line 57: note: this doesn't seem to slow down the related CI job at all

synthesizer/src/vm/helpers/sequential_op.rs
```diff
@@ -0,0 +1,133 @@
...
+impl<N: Network, C: ConsensusStorage<N>> VM<N, C> {
+    /// Launches a thread dedicated to the sequential processing of storage-related
+    /// operations.
+    pub fn start_sequential_queue(
+        &self,
+        request_rx: mpsc::Receiver<SequentialOperationRequest<N>>,
+    ) -> thread::JoinHandle<()> {
+        // Spawn a dedicated thread.
+        let vm = self.clone();
+        thread::spawn(move || {
```
kaimast, line 31: Do you know if this code is run in WASM? I am unsure if you can spawn threads there.

ljedrz: I know of a `wasm_thread` crate, though I haven't used it yet; an alternative would be to use a blocking task, but then the plumbing would need to be moved to snarkOS; good point, I'll think about it.

synthesizer/src/vm/helpers/sequential_op.rs
```diff
@@ -0,0 +1,133 @@
...
+            Some(response)
+        } else {
+            None
+        }
+    }
+
+    /// A safeguard used to ensure that the given operation is processed in the thread
+    /// enforcing sequential processing of operations.
+    pub fn ensure_sequential_processing(&self) {
+        assert_eq!(thread::current().id(), self.sequential_ops_thread.lock().as_ref().unwrap().thread().id());
```
kaimast, line 96: My understanding was that one of the goals of this PR is to avoid assertions like this.
Is there no straightforward way to ensure certain functions are only invoked from within the worker thread by leveraging the type system?

ljedrz: Indeed, such safeguards are suboptimal, even if more foolproof than before; as I noted in the PR description, I see no simple ways of introducing type safety here right now, but it would most likely be a heavy lift.

synthesizer/src/vm/helpers/sequential_op.rs
iamalwaysuncomfortable: Thanks for thinking of us Kai! We don't use `VM` object in wasm :). Most `wasm` usage is strictly for execution or using individual pieces of tooling in SnarkVM so I wouldn't worry about the `wasm` target at all if you're dealing with changes that ONLY affect the `VM` object.

ljedrz: Indeed, this is **only** related to the `VM` methods.

synthesizer/src/vm/helpers/sequential_op.rs
```diff
@@ -0,0 +1,133 @@
...
+        let (response_tx, response_rx) = oneshot::channel();
+        let request = SequentialOperationRequest { op, response_tx };
+
+        // This pattern match is infallible unless already shutting down the thread.
+        if let Some(tx) = &*self.sequential_ops_tx.read() {
+            // Send the operation to be processed sequentially.
+            let _ = tx.send(request);
+
+            // Wait for the result of the queued operation.
+            let Ok(response) = response_rx.blocking_recv() else {
```
kaimast, line 83: This panics when called from an async context. We need to document that somewhere.

ljedrz: Yes, this is expected; in production conditions we always run these operations within the context of blocking tasks. I'm not sure where to document it, though, other than perhaps the general storage documentation. That being said, I made a note on this at the callsite for now.

kaimast: I think we would need to document that operations that use it. For example, I had `check_next_block` panic. That should have been in a blocking task anyway, but it still a fairly easy mistake to make.

kaimast:
I pushed another commit ([c4f3d7e](https://github.com/ProvableHQ/snarkVM/pull/2975/commits/c4f3d7eb472c1452f049860726ee5476da70a372)) that adds documentation about functions panicking outside async contexts.

Otherwise, it looks good and should be good to merge.

kaimast (approved):

vicsn (approved):

PR #2976, opened 2025-10-23
Fix finalize print statement

vicsn:
## Motivation

When investigating the reasons for a rejected transaction, I surfaced these logs:
```
2025-10-22T16:19:30.218183Z TRACE snarkvm_synthesizer::vm::finalize: Failed to finalize execute tx at10w86clc8666yww60hzmfxekxa8hzfuaxramqac7ydwm54vlp9gzqw882zw - 'constructor' failed to evaluate command (sub r3 r2 into r4;)
2025-10-22T20:58:23.703188Z TRACE snarkvm_synthesizer::vm::finalize: Failed to finalize execute tx at182hjr96y8xdnttupnvfqqh8htur5t7v8ugl8tvxvkt2h06lgwqpsnd69qm - 'finalize' failed on 'credits.aleo/fee_public' - 'constructor' failed to evaluate command (sub r2 r1 into r3;)
```

Based on the respective commands which failed, it is clear these are 'transfer_public' and 'fee_public' calls which fail due to insufficient balance. Instead of logging these as constructor calls, they should be logged as the respective function call.

## Test Plan

Manually reviewed all instances of `registers.function_name` to catch potential issues.


synthesizer/process/src/finalize.rs
```diff
@@ -241,8 +240,17 @@ fn finalize_constructor<N: Network, P: FinalizeStorage<N>>(
     // Get the constructor types.
     let constructor_types = stack.get_constructor_types()?.clone();
 
+    // Set the FinalizeRegisters function name for constructors.
+    // Before ConsensusVersion::V12, this was wrongly using the program id as function name.
```
d0cd: Why is it "wrong" to use the program ID name vs. hard-coding constructor? Is there a security issue or is this purely for logging?

synthesizer/process/src/finalize.rs
```diff
@@ -241,8 +240,17 @@ fn finalize_constructor<N: Network, P: FinalizeStorage<N>>(
...
+    let consensus_version = N::CONSENSUS_VERSION(state.block_height())?;
+    let identifier = if consensus_version >= ConsensusVersion::V12 {
```
d0cd: If the purpose of this change is to improve the log message, can you pass in a flag to `finalize_command_expect_await` and set the message accordingly?

This avoids the permanent "overhead" of tracking this to a `ConsensusVersion`

synthesizer/process/src/finalize.rs
vicsn: Wrong was maybe strongly worded, but we are setting a program_id into a variable called function_name, which is semantically wrong. It's a piece of the code few people looked at, so added some documentation: f644a54e8

synthesizer/process/src/finalize.rs
vicsn: Agreed, it adds complexity and risks a fork. Changed it to just a logging adjustment: 7d92c1e

d0cd (approved):
LGTM! Option to consider adding program ID to scope name for more informative errors.

PR #2977, opened 2025-10-23
[Perf] Block advancement benchmarks

kaimast:
This PR serves a basis for the block caching and other future performance improvements and has two parts.

* It introduces a new block advancement benchmark and improves the existing storage benchmark
* It adds a new feature to testchain generator that allows storing blocks directly on disk. These generated blocks are then used in the benchmark to allow for much shorter runtimes

### Testchain block storage
When running the testchain generator with the `--no-ledger` flag, it will create a folder at the given `--storage-path` containing the genesis block as `genesis.data` and blocks as `block{{height}}.data`. In both cases, the data is binary encoded using `to_bytes_le`.

The block advancement and storage benchmarks then load the ledger using a new helper function `load_blocks` in `ledger/benches/common/mod.rs`. This avoids costly chain generation when running the benchmark, and also makes the benchmarks more consistent as they use the same ledger for every run.

### Notes
* This PR does not add a CI workflow for those new benchmarks yet, but that is something I would like to add in the future.
* I added a commit that bumps the baseline revision for semver checks, as this PR modifies the `PrettyUnwrap` trait.

### DIFF
--- a/ledger/benches/advance.rs
+++ b/ledger/benches/advance.rs
@@ -1,0 +1,137 @@
+// Copyright (c) 2019-2025 Provable Inc.
+// This file is part of the snarkVM library.
+
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at:
+
+// http://www.apache.org/licenses/LICENSE-2.0
+
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+use snarkvm_console::prelude::*;
+use snarkvm_ledger::{
+    Block,
+    store::{
+        ConsensusStorage,
+        helpers::{memory::ConsensusMemory, rocksdb::ConsensusDB},
+    },
+};
+use snarkvm_utilities::PrettyUnwrap;
+
+use criterion::{BenchmarkGroup, Criterion, criterion_group, criterion_main, measurement::WallTime};
+use std::time::Instant;
+
+mod common;
+use common::{CurrentNetwork, create_ledger, initialize_logging, load_blocks};
+
+/// Measures block advancement.
+fn bench_ledger_advancement<S: ConsensusStorage<CurrentNetwork>>(
+    name: &str,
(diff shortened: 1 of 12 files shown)


ledger/benches/common/mod.rs
```diff
@@ -0,0 +1,81 @@
...
+}
+
+/// Load blocks, which were generated using `snarkvm-testchain-generator --no-ledger [..]` from the given path.
+#[allow(dead_code)]
+pub fn load_blocks(path: &str) -> Result<(Block<CurrentNetwork>, Vec<Block<CurrentNetwork>>)> {
+    let mut current_height = 1;
+    let mut blocks = vec![];
+
+    let genesis_block = {
+        let path: PathBuf = format!("{path}/genesis.data").into();
```
ljedrz, line 53: I'd typically use [PathBuf::push](https://doc.rust-lang.org/std/path/struct.PathBuf.html#method.push) to piece together paths, but since this is only used in tests, it should be fine.

ledger/benches/advance.rs
```diff
@@ -0,0 +1,137 @@
...
+            let num_preloaded_blocks = blocks.len() - 1;
+            while (ledger.latest_height() as usize) < num_preloaded_blocks {
+                ledger.advance_to_next_block(blocks_iter.next().unwrap()).unwrap();
+            }
+
+            let last_block = blocks_iter.next().unwrap();
+
+            let start = Instant::now();
+            for _ in 0..num_ops {
+                ledger.check_next_block(last_block, rng).unwrap();
```
ljedrz, line 89: wouldn't it be preferable to run the numbers for all the blocks involved, instead of only the last one?

ledger/testchain-generator/src/main.rs
```diff
@@ -41,6 +41,12 @@ struct Args {
     /// The name of the network to generate the chain for.
     #[clap(long, value_parser=PossibleValuesParser::new(vec![CanaryV0::SHORT_NAME, TestnetV0::SHORT_NAME, MainnetV0::SHORT_NAME]), default_value=TestnetV0::SHORT_NAME)]
     network: String,
+    /// Set the seed to used to generate the chain.
+    #[clap(long)]
+    seed: Option<u64>,
```
ljedrz, line 46: Related: https://github.com/ProvableHQ/snarkVM/issues/2988.

ljedrz (approved):
LGTM in general, left one question.

vicsn (approved):
Not a high priority feature but looks clean so pushing it through

PR #2978, opened 2025-10-24
[Feat] Allow the testchain-generator to be fed with pregenerated txs

ljedrz:
This PR addresses [this request](https://github.com/ProvableHQ/snarkVM/issues/2943#issuecomment-3432746311); it allows the `testchain-generator` to:
1. introduce pre-generated transactions to the baked blocks
2. "skip" the first blocks in order to apply the transactions under the latest conensus version

The blocks are fed with the maximum number of possible transactions per block, and each validator produces the maximum number of transmissions per batch (as long as the number of txs and solutions is sufficient).

This extension feels a little bit "duct-tapey" to me, and perhaps some of the affected objects need a bit more adjustment in order for the overall setup to feel more "natural". @kaimast happy to hear your thoughts on this.

### DIFF
--- a/ledger/src/test_helpers/chain_builder.rs
+++ b/ledger/src/test_helpers/chain_builder.rs
@@ -36,7 +36,11 @@ use anyhow::{Context, Result};
 use indexmap::{IndexMap, IndexSet};
 use rand::SeedableRng;
 use rand_chacha::ChaChaRng;
-use std::collections::{BTreeMap, HashMap};
+use std::{
+    cmp,
+    collections::{BTreeMap, HashMap},
+    mem,
+};
 use time::OffsetDateTime;
 
 pub type CurrentNetwork = MainnetV0;
@@ -65,15 +69,36 @@ pub struct TestChainBuilder<N: Network> {
     last_committed_batch_round: HashMap<usize, u64>,
     /// The start of the test chain.
     genesis_block: Block<N>,
+    /// Preloaded transactions to populate the blocks with.
+    transactions: Vec<Transaction<N>>,
 }
 
 /// Additional options you can pass to the builder when generating a set of blocks.
-#[derive(Clone, Default)]
-pub struct GenerateBlocksOptions {
+#[derive(Clone)]
+pub struct GenerateBlocksOptions<N: Network> {
     /// Do not include votes to the previous leader certificate
     pub skip_votes: bool,
     /// Do not generate certificates for the specific node indices (to simulate a partition).
     pub skip_nodes: Vec<usize>,
+    /// A flag indicating that a number of initial "placeholder blocks" should be baked
+    /// wthout transactions in order to skip to the latest version of consensus.
+    pub skip_to_current_version: bool,
+    /// The number of validators.
+    pub num_validators: usize,
+    /// Preloaded transactions to populate the blocks with.
+    pub transactions: Vec<Transaction<N>>,
+}
+
+impl<N: Network> Default for GenerateBlocksOptions<N> {
+    fn default() -> Self {
+        Self {
+            skip_votes: false,
+            skip_nodes: Default::default(),
+            skip_to_current_version: false,
+            num_validators: 0,
+            transactions: Default::default(),
+        }
+    }
 }
 
 /// Additional options you can pass to the builder when generating a single block.
@@ -131,19 +156,27 @@ impl<N: Network> TestChainBuilder<N> {
     }
 
     /// Initialize the builder with the default quorum size.
-    pub fn new(rng: &mut TestRng) -> Result<Self> {
-        Self::new_with_quorum_size(4, rng)
+    pub fn new(rng: &mut TestRng, transactions: Vec<Transaction<N>>) -> Result<Self> {
+        Self::new_with_quorum_size(4, rng, transactions)
     }
 
     /// Initialize the builder with the specified quorum size.
-    pub fn new_with_quorum_size(num_validators: usize, rng: &mut TestRng) -> Result<Self> {
+    pub fn new_with_quorum_size(
+        num_validators: usize,
+        rng: &mut TestRng,
+        transactions: Vec<Transaction<N>>,
+    ) -> Result<Self> {
         let (private_keys, genesis) = Self::initialize_components(num_validators, rng)?;
-        Self::from_components(private_keys, genesis)
+        Self::from_components(private_keys, genesis, transactions)
     }
 
     /// Initialize the builder with the specified genesis block..
     /// Note: this function mirrors the way the private keys are sampled in snarkOS `fn parse_genesis`.
-    pub fn new_with_quorum_size_and_genesis_block(num_validators: usize, genesis_path: String) -> Result<Self> {
+    pub fn new_with_quorum_size_and_genesis_block(
+        num_validators: usize,
+        genesis_path: String,
+        transactions: Vec<Transaction<N>>,
+    ) -> Result<Self> {
         // Attempts to load the genesis block file.
         let buffer = std::fs::read(genesis_path)?;
         // Return the genesis block.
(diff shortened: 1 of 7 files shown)


ledger/src/test_helpers/chain_builder.rs
```diff
@@ -65,15 +69,36 @@ pub struct TestChainBuilder<N: Network> {
...
+    pub transactions: Vec<Transaction<N>>,
+}
+
+impl<N: Network> Default for GenerateBlocksOptions<N> {
+    fn default() -> Self {
+        Self {
+            skip_votes: false,
+            skip_nodes: Default::default(),
+            skip_to_current_version: false,
+            num_validators: 0,
```
ljedrz, line 92: This value is never used - it gets overwritten with the number of private keys in the `TestChainBuilder`.

ledger/testchain-generator/src/main.rs
```diff
@@ -99,8 +140,8 @@ fn generate_testchain<N: Network>(args: Args) -> Result<()> {
         let batch_size = (num_blocks - blocks.len()).min(100);
         let mut batch = builder.generate_blocks(batch_size, &mut rng).with_context(|| "Failed to generate blocks")?;
 
-        println!("Generated {pos} of {num_blocks} blocks");
         pos += batch_size;
+        println!("Generated {pos} of {num_blocks} blocks");
```
ljedrz, line 184: This is a drive-by fix; without it, `pos` is always outdated by a single iteration.

ledger/testchain-generator/Cargo.toml
```diff
@@ -22,6 +22,7 @@ workspace = true
 
 [dependencies.snarkvm-console]
 workspace = true
+features = [ "test_consensus_heights" ]
```
ljedrz, line 30: Pretty sure we want this by default, otherwise we would never be able to reach the current consensus version.

ljedrz:
The CI jobs failed due to some GitHub outage; I'll rerun them later.

ledger/testchain-generator/src/main.rs
```diff
@@ -38,6 +44,9 @@ struct Args {
     /// Remove existing ledger if it already exists.
     #[clap(long, short = 'f')]
     force: bool,
+    /// Load the transactions to be used with the generated blocks.
```
vicsn: Can you document the required format of the txs file?

ledger/testchain-generator/src/main.rs
```diff
@@ -80,13 +89,45 @@ fn generate_testchain<N: Network>(args: Args) -> Result<()> {
...
+                let entry = entry?;
+                let path = entry.path();
+
+                let mut file = File::open(path)?;
+                let _ = file.read_to_string(&mut buffer)?;
+                let tx = Transaction::<N>::from_str(&buffer)?;
+                txs.push(tx);
+                buffer.clear();
+
+                // TODO: remove
```
vicsn: Indeed plz remove

vicsn (changes requested):
Great

ledger/testchain-generator/src/main.rs
ljedrz: https://github.com/ProvableHQ/snarkVM/pull/2978/commits/eedb6caa2c748ee0bc2f73fdf68b47f666b82837

ledger/testchain-generator/src/main.rs
ljedrz: https://github.com/ProvableHQ/snarkVM/pull/2978/commits/eedb6caa2c748ee0bc2f73fdf68b47f666b82837

ledger/src/test_helpers/chain_builder.rs
```diff
@@ -181,32 +222,69 @@ impl<N: Network> TestChainBuilder<N> {
...
+            println!("Skipping {height} blocks to reach {version}");
+
+            for _ in 0..*height {
+                let options = GenerateBlockOptions {
+                    skip_votes: options.skip_votes,
+                    skip_nodes: options.skip_nodes.clone(),
+                    ..Default::default()
+                };
+
+                let block = self.generate_block_with_opts(options, rng)?;
```
kaimast: You could use `generate_blocks_with_opts` here to get rid of the for loop.

ledger/src/test_helpers/chain_builder.rs
```diff
@@ -127,19 +152,27 @@ impl<N: Network> TestChainBuilder<N> {
     }
 
     /// Initialize the builder with the default quorum size.
-    pub fn new(rng: &mut TestRng) -> Result<Self> {
-        Self::new_with_quorum_size(4, rng)
+    pub fn new(rng: &mut TestRng, transactions: Vec<Transaction<N>>) -> Result<Self> {
```
kaimast: It might be nicer to have this as a separate constructor and default to no transactions.

ledger/src/test_helpers/chain_builder.rs
```diff
@@ -181,32 +222,69 @@ impl<N: Network> TestChainBuilder<N> {
...
 
     /// Create multiple blocks, with fully-connected DAGs.
     pub fn generate_blocks(&mut self, num_blocks: usize, rng: &mut TestRng) -> Result<Vec<Block<N>>> {
-        self.generate_blocks_with_opts(num_blocks, GenerateBlocksOptions::default(), rng)
+        let num_validators = self.private_keys.len();
+        let transactions = mem::take(&mut self.transactions);
+
+        self.generate_blocks_with_opts(
+            num_blocks,
+            GenerateBlocksOptions { num_validators, transactions, ..Default::default() },
```
kaimast: What if not all transactions fit in the block(s)? Will the rest be stored back in `self.transactions`?

ledger/src/test_helpers/chain_builder.rs
```diff
@@ -181,32 +222,69 @@ impl<N: Network> TestChainBuilder<N> {
...
             previous_leader_certificate: Default::default(),
+            transactions,
         })
     }
 
     /// Create multiple blocks, with fully-connected DAGs.
     pub fn generate_blocks(&mut self, num_blocks: usize, rng: &mut TestRng) -> Result<Vec<Block<N>>> {
-        self.generate_blocks_with_opts(num_blocks, GenerateBlocksOptions::default(), rng)
+        let num_validators = self.private_keys.len();
+        let transactions = mem::take(&mut self.transactions);
```
kaimast: Is this the only purpose of setting `transactions`? If so, wouldn't it be cleaner to not have that member and just call `generate_block_with_opts` from the testchain generator?

kaimast (changes requested):
Looks good. I added two questions/comments.

ljedrz:
This PR can yield to https://github.com/ProvableHQ/snarkVM/pull/2977.

ledger/src/test_helpers/chain_builder.rs
ljedrz: I normally would, but wouldn't this just recurse? It's in the body of `generate_blocks_with_opts` after all.

ledger/src/test_helpers/chain_builder.rs
ljedrz: you mean like a `fn new_with_transactions` (following the existing convention)? or rather `fn with_transactions(&mut self)`, updating that one field?

ledger/src/test_helpers/chain_builder.rs
ljedrz: as it stands right now, `generate_blocks` consumes all the `transactions` (via `mem::take` right above); do we foresee this being used several times? if not, we could have it take `self` by value, as is usual for the `Builder` pattern

ledger/src/test_helpers/chain_builder.rs
ljedrz: we could - I just wanted to use the current setup as much as possible, in order to not mix things up; and as it stands now, _the_ pick of choice is `generate_blocks` (which is currently being batched "from the outside" - I think it would be preferable to move that batching into the `chain_builder` setup)

kaimast:
@ljedrz I rebased the PR and added a commit that reduces the change set ([67dfc4e](https://github.com/ProvableHQ/snarkVM/pull/2978/commits/67dfc4e86808ce1f866f73787336a5c66d91776b)). Let me know what you think!

kaimast (approved):

vicsn (approved):

kaimast:
Generated a testchain with it and used it to benchmark snarkOS. Will merge it now!

PR #2980, opened 2025-10-24
Track invalidated stake

vicsn:
## Motivation

When combing through mainnet logs looking for potential issues, the only anomaly left on my list was seeing these: "is not in the committee - skipping all stakers"

In order to determine whether this is actually an actionable problem we'd want to inform the community about, this PR proposes to log the total amount of stake subject to this condition.


raychu86 (approved):
LGTM.

Only nit is potentially not logging the total invalidated stake if the value is 0.

PR #2944, opened 2025-10-06
[Feature] LoggableError trait

kaimast:
This PR introduce the `LoggableError` trait which enables to easily log `anyhow::Error`s in snarkVM. This allows to have a more uniformed error logging mechanisms across snarkOS/snarkVM. 
The PR's changes are also included in #2927, but the latter is already very complex and the logging changes are self-contained.

To see how `LoggableError` can be used, see the changes in `ledger/store`  in [71a6665](https://github.com/ProvableHQ/snarkVM/pull/2944/commits/71a6665096ddb86d48edc84235123ce2c464d2da).

### Notes
* This changes some of the error logging functions introduced in #2813, which makes it a breaking change. As far as I know, these functions are not (yet) used outside snarkVM and snarkOS.

PR #2992, opened 2025-10-30
Revise error logging

kaimast:
This PR removes `LoggableError` and then `log_and replaces it with simpler macros: `elog_error`, `elog_warning`, and `elog_debug`.

### Details
The current approach has a significant limitation in that it does not log the affected module correctly. The `track_caller` macro does indeed forward the caller's location (filename and line), but there is no way to access the module name.
The only way to implement this properly is through a macro.

The new `elog_` macros also replace the previous `log_*` functions.

You can either invoke them with an error directly.
```
elog_warning!(myerror); //logs the error at the WARN level
```

Or you can add a context, as with `LoggableError`.
```
elog_debug!(myerror, "Sync error at height {height}");
```
### Changed Behavior
Before these changes. Certain log messages incorrectly listed `errors` module as their source if they used the `LoggableError` trait. The message below is an example.
```
2025-10-30T21:13:40.693297Z  WARN snarkvm_utilities::errors: Cannot sign a batch at round 297 from '127.0.0.1:5139' - Certificate for round 296 already exists in storage (gc = 194)
```

After the changes, the macro will correctly print the module where the error was thrown.
```
2025-10-30T21:13:40.693297Z  WARN snarkos_node_bft::primary: Cannot sign a batch at round 297 from '127.0.0.1:5139' - Certificate for round 296 already exists in storage (gc = 194)
```

This is a needed fix as the module is often used to filter or search for log messages.

### Testing
There is a [companion PR in snarkOS](https://github.com/ProvableHQ/snarkOS/pull/3981) that uses these changes, and that I am testing the PR with.

The PR also includes a new unit tests that verifies the log output using `tracing_test`.

### DIFF
--- a/ledger/src/find.rs
+++ b/ledger/src/find.rs
@@ -15,7 +15,7 @@
 
 use super::*;
 
-use snarkvm_utilities::LoggableError;
+use snarkvm_utilities::elog_warning;
 
 impl<N: Network, C: ConsensusStorage<N>> Ledger<N, C> {
     /// Returns the block height that contains the given `state root`.
@@ -131,7 +131,7 @@ impl<N: Network, C: ConsensusStorage<N>> Ledger<N, C> {
                 Ok(Some(commitment)) => Some((commitment, record)),
                 Ok(None) => None,
                 Err(err) => {
-                    err.log_warning(format!("Failed to process 'find_record_ciphertexts({filter:?})'"));
+                    elog_warning!(err, "Failed to process 'find_record_ciphertexts({filter:?})'");
                     None
                 }
             }
@@ -149,7 +149,7 @@ impl<N: Network, C: ConsensusStorage<N>> Ledger<N, C> {
             iter.flat_map(|(commitment, record)| match record.decrypt(view_key) {
                 Ok(record) => Some((commitment, record)),
                 Err(err) => {
-                    err.log_warning("Failed to decrypt record");
+                    elog_warning!(err, "Failed to decrypt record");
                     None
                 }
             })
--- a/ledger/store/Cargo.toml
+++ b/ledger/store/Cargo.toml
@@ -109,6 +109,9 @@ workspace = true
 features = [ "write" ]
 optional = true
 
+[dependencies.tracing]
+workspace = true
+
 [dev-dependencies.aleo-std]
 workspace = true
 
--- a/ledger/store/src/helpers/rocksdb/internal/map.rs
+++ b/ledger/store/src/helpers/rocksdb/internal/map.rs
@@ -18,7 +18,7 @@
 use super::*;
 use crate::helpers::{Map, MapRead};
 
-use snarkvm_utilities::{LoggableError, bytes::unchecked_deserialize};
+use snarkvm_utilities::{bytes::unchecked_deserialize, elog_error};
 
 use core::{fmt, fmt::Debug, hash::Hash, mem};
 use indexmap::IndexMap;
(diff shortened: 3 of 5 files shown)


vicsn:
Closing because no issue was created inviting discussion. Please make sure to give a **concrete** example of why logging the module is useful.

kaimast:
> Closing because no issue was created inviting discussion. Please make sure to give a **concrete** example of why logging the module is useful.

I wasn't fully done with the draft PR. I added a relevant example to the description now.

vicsn:
> I wasn't fully done with the draft PR. I added a relevant example to the description now.

A discussion first would be great, we should at least consider reverting to the original approach to logging instead of adding more complexity.

kaimast:
Generally, this PR removes complexity. It adds lines of code, but those are mostly documentation and testing.  Below is a brief overview over the different approaches.

## Comparison
### Original Approach to Logging

The original approach to logging is still used widely in the codebase and makes sense in some instances. However, it requires a significant amount of boilerplate code and introduces inconsistent formatting of error message. For example, some used a colon to separate error, while others used dashes. Similarly, some error messages have color coding (such as using `dimmed()`) while others do not.

```
// Error handling boilerplate to add context
result.map_err(|e| {
    error!("RocksDB Iter deserialize(key) error: {e}");
})
.ok()?;
```

A common issue was also that we logged the error message `{e}` instead of the full error `{e:?}`.  For example, I would get a sync or BFT error with a low-level error from snarkVM as its cause. In such a case, I would need to figure out which code path could have caused this particular message, by either manually looking through the code, or adding additional log messages and re-run the failed test.
The latter is particularly cumbersome if we are looking at logs provided by a third party. 

The code works around this sometimes by having multiple log messages (e.g., one in `tcp` and then another in `router` or `gateway`). That works to some extent, but it makes it hard to figure out which errors are directly connected and can add noise to the logs.

### `LoggableError`
A recent PR #2944 introduced the `LoggableError` trait that adds new methods to `anyhow::Error` that simplifies generating those logs.

```
err.log_error("RocksDB Iter deserialize(key) error");
```

However, this approach had the limitation of not logging the module of the line that was invoked, but the module of where `LoggableError::log_error` was defined. I falsely assumed that `#[track_caller]` would avoid this, and it was not caught during review.
The approach also had another limitation in that `log_error` only accepts string and required another call to `format!` to format log messages, such as in the example below:

```
err.log_error(format!("Failed to fetch block at height {height}"));
```

### `elog_*`
This PR simplifies the `LoggableError` logic by replacing it with three fairly simple macros. It also fixes the issue of the incorrect submodule being logged. The above example becomes the following:
```
elog_error!("Failed to fetch block at height {height}");
```

The remaining downside of the approach is that it might be too verbose in some cases. I personally would rather err on the side of having log messages that are too verbose, but we can always remove `with_context(..)` calls to reduce that.

Finally, this change does not change or break any existing error logging not using the new approach, nor am I planning to migrate the entire codebase to these macros.

vicsn:
> A common issue was also that we logged the error message {e} instead of the full error {e:?}

Thank you for explaining the main problem at hand. The design of `LoggableError` was fine, but we should limit the number of macros in the code because most developers can't quickly read/debug how they work. Is any other design possible? Or could we stick to manually writing `{e:?}`?

kaimast:
We could remove everything but the `flatten_error` function. It requires more boilerplate code for error logging, but it is a little easier to read.
I made a branch for it here: https://github.com/ProvableHQ/snarkVM/compare/fix/revise-error-logging...fix/revise-error-logging-minimal?expand=1

vicsn:
> We could remove everything but the flatten_error function. It requires more boilerplate code for error logging, but it is a little easier to read.

Looks good!

Issue #2981, opened 2025-10-24
[Feature] Adjust CI to report unit test timings


vicsn:
## 🚀 Feature

CI logic is defined in `.circleci/config.yml`

In order to see which tests take the most amount of time, we could test using `cargo +nightly test -- -Zunstable-options --report-time`.

I tried it for one test and because we're on nightly, it prints a lot of extra warnings. I assume we'll be able to ignore those - because the existing tests already emit warnings sometimes which we ignore, and we have a workspace-level non-nightly `clippy` command which we take serious.

vicsn:
If this works we can also do this for snarkOS

meddle0x53:
Done for the SnarkOS too.

PR #2995, opened 2025-10-31
Implement reporting unit test timings in CI

meddle0x53:
## Motivation

Implements https://github.com/ProvableHQ/snarkVM/issues/2981 and also stores the timing reports as CSVs so they can be analysed.

## Test Plan

The CI passes and the timings of the different tests could be viewed.

## Documentation

The PR updates the CI configuration and modifies how tests are ran.

## Backwards compatibility

Backwards compatible.

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -786,7 +786,7 @@ jobs:
           flags: >
             --lib --bins --features test
             --partition count:1/2
-            -- --test-threads 8
+            -- --test-threads 16
           cache_key_suffix: -test1
           timeout: 25m
 
@@ -799,7 +799,7 @@ jobs:
           flags: >
             --lib --bins --features test
             --partition count:2/2
-            -- --test-threads 8
+            -- --test-threads 16
           cache_key_suffix: -test2
           timeout: 25m
 
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -248,7 +248,7 @@ commands:
               fi
             fi
 
-            DEFAULT_JOBS=16
+            DEFAULT_JOBS=12
 
             # Default JOBS, overridden by TEST_THREADS if set
             JOBS="${NEXTEST_JOBS:-$DEFAULT_JOBS}"
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -237,6 +237,17 @@ commands:
               fi
             fi
 
+            # Detect special "cargo-style" nextest mode: run like cargo test, but via nextest.
+            # This is a pseudo-flag for the CI script and must NOT be passed to cargo/nextest.
+            NEXTEST_CARGO_STYLE=0
+            if printf '%s\n' "$BUILD_FLAGS" | grep -q -- '--nextest-cargo-style'; then
+              NEXTEST_CARGO_STYLE=1
+              # Strip the pseudo-flag from BUILD_FLAGS so cargo/nextest never see it.
+              BUILD_FLAGS="$(printf '%s\n' "$BUILD_FLAGS" \
+                | sed -E 's/(^|[[:space:]])--nextest-cargo-style([[:space:]]|$)/ /g' \
+                | tr -s ' ' | sed -E 's/^[[:space:]]+//; s/[[:space:]]+$//')"
+            fi
+
             # Detect and extract --test-threads value (used to set JOBS)
             TEST_THREADS=""
             if [ -n "$RIGHT_OF_SEP" ]; then
@@ -258,6 +269,22 @@ commands:
               JOBS="$TEST_THREADS"
             elif [ -n "${NEXTEST_JOBS:-}" ]; then
               JOBS="$NEXTEST_JOBS"
+            fi
+
+            # If --nextest-cargo-style was requested, force nextest to a single job.
+            # This mimics cargo test in terms of "only one big test binary active at a time".
+            if [ "$NEXTEST_CARGO_STYLE" -eq 1 ]; then
+              JOBS=1
+
+              # Auto-detect cores for --test-threads, fall back to 1.
+              if command -v nproc >/dev/null 2>&1; then
+                CORES="$(nproc)"
+              else
+                CORES=1
+              fi
+
+              TEST_THREADS="$CORES"       # for logging
+              LIBTEST_ARGS+=(--test-threads "$CORES")
             fi
 
             # Create nextest config (disable fail-fast, add slow-timeout + JUnit output)
@@ -296,13 +323,24 @@ commands:
               cargo test --package="<< parameters.workspace_member >>" $BUILD_FLAGS_SANITIZED | tee "$OUT"
               : > "$CSV"; : > "$TOP"
             else
-              cargo nextest run \
-                -p "<< parameters.workspace_member >>" $BUILD_FLAGS_SANITIZED \
-                --profile ci "${NEXT_JOBS_ARGS[@]}" "${CFG_ARGS[@]}" \
-                --status-level fail --hide-progress-bar \
-                --no-tests=pass \
-                ${FILTER_EXPR:+--filter-expr "$FILTER_EXPR"} \
-                2>&1 | tee "$OUT"
+              if [ "${#LIBTEST_ARGS[@]}" -gt 0 ]; then
+                cargo nextest run \
+                  -p "<< parameters.workspace_member >>" $BUILD_FLAGS_SANITIZED \
+                  --profile ci "${NEXT_JOBS_ARGS[@]}" "${CFG_ARGS[@]}" \
+                  --status-level fail --hide-progress-bar \
+                  --no-tests=pass \
+                  ${FILTER_EXPR:+--filter-expr "$FILTER_EXPR"} \
+                  -- "${LIBTEST_ARGS[@]}" \
+                  2>&1 | tee "$OUT"
+              else
+                cargo nextest run \
+                  -p "<< parameters.workspace_member >>" $BUILD_FLAGS_SANITIZED \
+                  --profile ci "${NEXT_JOBS_ARGS[@]}" "${CFG_ARGS[@]}" \
+                  --status-level fail --hide-progress-bar \
+                  --no-tests=pass \
+                  ${FILTER_EXPR:+--filter-expr "$FILTER_EXPR"} \
+                  2>&1 | tee "$OUT"
+              fi
 
               # Parse junit.xml to extract test timings
               if [ -f "$JUNIT" ]; then
@@ -792,7 +830,7 @@ jobs:
     steps:
       - run_test:
           workspace_member: snarkvm-synthesizer
-          flags: --lib --bins --features test
+          flags: --lib --bins --features test --nextest-cargo-style
           cache_key_suffix: -test
           timeout: 25m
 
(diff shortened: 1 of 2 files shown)


vicsn:
Had a chat with Lukasz, nightly does have a higher chance of inadvertently breaking things, could you use `cargo nextest` instead? That also has an option to report timings, but it may be a bit more work to update the circleci config.

Just for context: the reason we use limited test-threads is to prevent going out of memory.

.circleci/config.yml
```diff
@@ -149,14 +149,136 @@ commands:
...
           name: "Run Tests"
           timeout: << parameters.timeout >>
           no_output_timeout: << parameters.no_output_timeout >>
-          command: RUST_MIN_STACK=67108864 cargo test --package=<< parameters.workspace_member >> << parameters.flags >>
+          command: |
+            set -euo pipefail
+            export CARGO_TERM_COLOR=never
+            mkdir -p artifacts target/nextest/ci
+
+            OUT="artifacts/<< parameters.workspace_member >>-nextest.txt"
```
meddle0x53, line 170: The new way of testing measures time and stores the results as artifacts:
1. The `*nextest.txt` contains the test logs
2. The `-test-times.csv` contains all the tests with their times in a CSV table
3. The `-slow-tests-top30.csv` contains the 30 slowest sorted from slowest downwards.

.circleci/config.yml
```diff
@@ -149,14 +149,136 @@ commands:
...
+            CSV="artifacts/<< parameters.workspace_member >>-test-times.csv"
+            TOP="artifacts/<< parameters.workspace_member >>-slow-tests-top30.csv"
+            JUNIT="target/nextest/ci/junit.xml"
+
+            FLAGS_FULL="<< parameters.flags >>"
```
meddle0x53, line 176: The flags passed to nextest and the flags passed to the different cases shouldn't interfere so with a lot of testing I reworked this script to work with all possible types of flags being passed when ran with `cargo nextest` and its flags.

.circleci/config.yml
```diff
@@ -149,14 +149,136 @@ commands:
...
+            fi
+
+            BUILD_FLAGS_SANITIZED="$(printf '%s' "$BUILD_FLAGS" \
+              | sed -E "s/(^|[[:space:]])--test[[:space:]]+'[^']*'([[:space:]]|$)/ /g" \
+              | sed -E 's/(^|[[:space:]])--test[[:space:]]+\"[^\"]*\"([[:space:]]|$)/ /g' \
+              | sed -E 's/(^|[[:space:]])--test[[:space:]]+[^[:space:]]+([[:space:]]|$)/ /g' \
+              | tr -s ' ' | sed -E 's/^[[:space:]]+//; s/[[:space:]]+$//' \
+            )"
+
+            FILTER_EXPR=""
```
meddle0x53, line 206: The nextest syntax for filtering is a bit strange, so I did this here to work with the old test flags for filtering.

So when defining a new package tests we can just use the normal `--skip` flags, this little script handles converting them.

.circleci/config.yml
```diff
@@ -149,14 +149,136 @@ commands:
...
+                  fi
+                done < "$tmpf"
+                rm -f "$tmpf"
+                if [ -n "$EXPRS" ]; then
+                  FILTER_EXPR="not ($EXPRS)"
+                fi
+              fi
+            fi
+
+            JOBS="${NEXTEST_JOBS:-2}"
```
meddle0x53: Next test's number of JOBS are defined here and can be overridden by passing `environment` variables to the specific package test definitions. By default it is the minimal 2, but for some I adjusted more.

.circleci/config.yml
```diff
@@ -149,14 +149,136 @@ commands:
...
+              CFG_ARGS+=(--config-file ".config/nextest.toml")
+            fi
+            {
+              printf '%s\n' '[profile.ci]'
+              printf '%s\n' 'fail-fast = false'
+              printf '%s\n' "slow-timeout = \"${SLOW_PERIOD}\""
+              printf '%s\n' ''
+              printf '%s\n' '[profile.ci.junit]'
+              printf '%s\n' 'path = "junit.xml"'
+            } > "$TMPD/nextest-override.toml"
```
meddle0x53, line 273: Nextest needs configuration file, so this is creating it.

.circleci/config.yml
```diff
@@ -149,14 +149,136 @@ commands:
...
+            CFG_ARGS+=(--config-file "$TMPD/nextest-override.toml")
+
+            echo "==> nextest package: << parameters.workspace_member >>"
+            echo "==> build flags (sanitized): '${BUILD_FLAGS_SANITIZED}'"
+            echo "==> filter-expr: ${FILTER_EXPR:-<none>}  (right-of-sep: ${RIGHT_OF_SEP:-<none>})"
+            echo "==> jobs: ${JOBS}"
+
+            if printf '%s' " $BUILD_FLAGS_SANITIZED " | grep -q ' --no-run '; then
```
meddle0x53, line 285: `--no-run` breaks nextest, so if we are going to do a not run test, we just keep the old command.

.circleci/config.yml
```diff
@@ -149,14 +149,136 @@ commands:
...
+              cargo test --package="<< parameters.workspace_member >>" $BUILD_FLAGS_SANITIZED | tee "$OUT"
+              : > "$CSV"; : > "$TOP"
+            else
+              cargo nextest run \
```
meddle0x53: Here is the nextest comand working and fully tested.

.circleci/config.yml
```diff
@@ -149,14 +149,136 @@ commands:
...
+                -p "<< parameters.workspace_member >>" $BUILD_FLAGS_SANITIZED \
+                --profile ci -j "${JOBS}" "${CFG_ARGS[@]}" \
+                --status-level fail --hide-progress-bar \
+                --no-tests=pass \
+                ${FILTER_EXPR:+--filter-expr "$FILTER_EXPR"} \
+                2>&1 | tee "$OUT"
+
+              if [ -f "$JUNIT" ]; then
+                awk -v RS='<testcase ' '
```
meddle0x53, line 299: This is the generation of the artifacts from the test logs. In nextest timing is reported in `<time>`, so we `awk` it and the name of the test.

.circleci/config.yml
```diff
@@ -149,14 +149,136 @@ commands:
...
+                  NR>1 {
+                    name=""; classname=""; time="";
+                    start=index($0,"classname=\""); if(start>0){ start+=11; rest=substr($0,start); end=index(rest,"\""); if(end>0) classname=substr(rest,1,end-1) }
+                    start=index($0,"name=\"");      if(start>0){ start+=6;  rest=substr($0,start); end=index(rest,"\""); if(end>0) name=substr(rest,1,end-1) }
+                    start=index($0,"time=\"");      if(start>0){ start+=6;  rest=substr($0,start); end=index(rest,"\""); if(end>0) time=substr(rest,1,end-1) }
+                    if(name!="" && time!=""){ full=(classname!=""?classname"::"name:name); printf "%s,%s\n", full, time }
+                  }
+                ' "$JUNIT" > "$CSV" || true
+
+                sort -t, -k2,2nr "$CSV" | head -n 30 > "$TOP" || true
```
meddle0x53, line 310: Here is how the top 30 are sorted and created.

.circleci/config.yml
```diff
@@ -149,14 +149,136 @@ commands:
...
+              else
+                echo "WARN: $JUNIT not found; no timings extracted." >&2
+                : > "$CSV"; : > "$TOP"
+              fi
+            fi
+
+      - store_artifacts:
```
meddle0x53, line 317: This stores the new artifacts  and they can be viewed in the CI job.

.circleci/config.yml
```diff
@@ -149,14 +149,136 @@ commands:
...
+
+            OUT="artifacts/<< parameters.workspace_member >>-nextest.txt"
+            CSV="artifacts/<< parameters.workspace_member >>-test-times.csv"
+            TOP="artifacts/<< parameters.workspace_member >>-slow-tests-top30.csv"
+            JUNIT="target/nextest/ci/junit.xml"
+
+            FLAGS_FULL="<< parameters.flags >>"
+            FLAGS_TRIM="$(printf '%s' "$FLAGS_FULL" | sed -e 's/^[[:space:]]*//' -e 's/[[:space:]]*$//')"
+
+            RIGHT_OF_SEP=""
```
vicsn, line 180: Can you add a comment for each block what it does and why it is needed?

.circleci/config.yml
```diff
@@ -601,11 +728,14 @@ jobs:
...
           flags: --lib --bins
-          timeout: 20m
+          no_output_timeout: 20m
+          timeout: 30m
 
   synthesizer-test:
     executor: rust-docker
     resource_class: << pipeline.parameters.twoxlarge >>
+    environment:
+      NEXTEST_JOBS: "4"
```
vicsn: Does this replace `--test-threads=2`? Should we replace all instances of `--test-threads` by `NEXTEST_JOBS`?

Perhaps related: the synthesizer-workflow now takes 55 minutes, I recall it taking less in previous runs.

vicsn:
Looks great! Can you make a PR in snarkOS as well?

.circleci/config.yml
meddle0x53: I didn't want to add bash comments to the inlined script to not break something unintentionally in the logic, but will try!

.circleci/config.yml
meddle0x53: Ugh, yeah, so I'll try to make `--test-threads` be automatically converted to `NEXTEST_JOBS` so we don't have to change anything.

For the workflow - I think it may be from the cache, being invalidated with each of my changes or maybe related to not using the `--test-threads`. Let me do a commit to see.

meddle0x53:
@vicsn I tried a few things here (though am mostly trying to make the snarkos version work in the moment) and it seems nextest runs are slower.

So, because we are going to try to speed up things anyway either I can try to make `synthesizer-workflow` faster in this PR or in a next one - your decision.

.circleci/config.yml
vicsn: > @vicsn I tried a few things here (though am mostly trying to make the snarkos version work in the moment) and it seems nextest runs are slower.
> So, because we are going to try to speed up things anyway either I can try to make synthesizer-workflow faster in this PR or in a next one - your decision.

Let's not merge when it's twice as slow. And independent of the runtime, it would be great if we could consistently use `NEXTEST_JOBS` OR `--test-threads`, if those are conceptually the same thing

meddle0x53:
Actually looking around many workflows are passing for the same time, especially in the SnarkOS version. It is specific for the [synthesizer-workflow](https://github.com/ProvableHQ/snarkVM/pull/2995/checks?check_run_id=54468695960) and for some of the others a bit, but not so much, hmmm.

.circleci/config.yml
meddle0x53: I think I found the problem, debugging, checking and I may be able to fix it `--test-threads ` will be used only.

vicsn (approved):
Amazing!

vicsn:
We will surely hit some edge-cases the following days on new PRs, but happy that we have this data to start optimizing

PR #2997, opened 2025-10-31
Disable StringType from V12

vicsn:
## Motivation

This PR disables the broken StringType from V12.

It does not fix StringType sampling, to make it less likely any deployment with strings will land before V12.

This does not win the efficiency award, but we can prune all of this checking logic and the Entire StringType code in the future.

## Test Plan

Added unit tests showcasing behaviour.

### DIFF
--- a/synthesizer/program/src/lib.rs
+++ b/synthesizer/program/src/lib.rs
@@ -99,17 +99,7 @@ use console::{
             take,
         },
     },
-    program::{
-        FinalizeType,
-        Identifier,
-        Literal,
-        LiteralType,
-        PlaintextType,
-        ProgramID,
-        RecordType,
-        StructType,
-        ValueType,
-    },
+    program::{Identifier, PlaintextType, ProgramID, RecordType, StructType},
     types::U8,
 };
 use snarkvm_utilities::cfg_iter;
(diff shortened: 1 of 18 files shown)


raychu86 (approved):

synthesizer/program/src/lib.rs
```diff
@@ -966,6 +976,69 @@ impl<N: Network> ProgramCore<N> {
 
         function_contains || closure_contains || command_contains || array_size_exceeds
     }
+
+    /// Returns `true` if a program contains any string type.
+    /// Before ConsensusVersion::V12, variable-length string sampling when using them as inputs caused deployment synthesis to be inconsistent and abort with probability 63/64.
+    /// After ConsensusVersion::V12, string types are disallowed.
+    #[inline]
+    pub fn contains_string_type(&self) -> bool {
```
raychu86, line 974: Seems like i was preemptive in my approval. There are some potential missing checks - 

- [x] Function operands
- [x] Function inputs
- [x] Function Outputs
- [x] Closure operands
- [x] Closure inputs (missing but should be accounted for by function operands)
- [x] Closure outputs (missing but should be accounted for by closure oprands)
- [x] Finalize operands
- [x] Finalize inputs (missing but should be accounted for by function operands)
- [x] Finalize outputs (missing but should be accounted for by the finalize operands)
- [x] Structs (missing)
- [x] Records (missing)
- [x] Mapping types (missing)

raychu86 (changes requested):
Dismissing old approval, I believe there are still some missing checks

raychu86 (approved):

PR #2998, opened 2025-11-03
Skip proof verification if building with dev_skip_checks

vicsn:
## Motivation

In order to accelerate testing of Leo programs, we introduce a feature flag `dev_skip_checks` to skip proof verification.

To prevent accidental usage, a compilation guard has been added to make it depend on `test_consensus_heights`, which in turn snarkOS only sets when using `test_network`.

Also fixes a bug in Fee deserialization which disallowed submitting a fee without Proof object. This does not create any security vulnerability because the VM explicitly checks for existence and validity of fee Execution Proofs.

## Out of scope ideas for future work
- Note that this approach may give false positives in case a devnode and caller have different configurations (block height, consensus version, snarkVM commit hash). The only middle road way to get around that would be to receive an entire `R1CS` object and for the devnode to check that it is satisfied given it's own witness. This can be follow-up work, because regardless, we will want to land the proof-less transaction in the VM.
- We have a mix of dev_, dev- and test_ feature flags, we can consider creating a formal standard at some point.


raychu86:
The verification of the proofs should be fairly lightweight (<10ms).

Is this a preemptive approach to removing the proof generation entirely for these dev networks?

vicsn:
> The verification of the proofs should be fairly lightweight (<10ms).
> Is this a preemptive approach to removing the proof generation entirely for these dev networks?

This allows us to not have to generate proofs in E2E Leo tests. Several partners complained a few months ago that tests are taking up many minutes or even half an hour to run, and there is now a resolution: https://github.com/ProvableHQ/leo/tree/feat/leo-devnode

raychu86:
> > The verification of the proofs should be fairly lightweight (<10ms).
> > Is this a preemptive approach to removing the proof generation entirely for these dev networks?
> 
> This allows us to not have to generate proofs in E2E Leo tests. Several partners complained a few months ago that tests are taking up many minutes or even half an hour to run, and there is now a resolution: https://github.com/ProvableHQ/leo/tree/feat/leo-devnode

Ah, I thought proof skipping was going to be done in snarkVM via a feature-flag, but it seems like it's fairly easy to construct in leo only.

raychu86 (approved):

PR #3002, opened 2025-11-05
[Feat] Announce new blocks via IPC

ljedrz:
**Introduction**

This is a proposal for a new feature which would allow freshly inserted blocks to be streamed via an IPC socket.

**Use cases**

The expected users are those who seek to receive information on new blocks as soon as possible (e.g. block explorers), and potentially (if only transmitting further, or with some manual deserialization) without even having to depend on snarkVM. This feature can also be used in test-related tools which produce blocks, e.g. in tandem with the `testchain-generator`, for immediate consumption by other tools.

**Why not implement this in snarkOS?**

1. This feature could be used by tools which don't depend on snarkOS, custom nodes, etc.
2. Having this operation in the context of the sequential processing thread guarantees that an aggressively syncing node won't cause the block stream to become unordered.
3. This also ensures that the blocks can be streamed while syncing **via any means** (CDN, other peers), ensuring maximum stream performance after a potential downtime. This is _the_ ultimate point where block insertion happens (other than directly in the `Ledger` itself) so no other place requires the announcement code to be attached.
4. In snarkOS, blocks are normally inserted via `try_advance_to_next_block`, which then notifies other nodes via the `BlockLocators` mechanism, which means attaching this setup there would incur additional database queries and cloning (as the `Block` is not readily available).
5. snarkVM already implements the entire persistent storage, so introducing a feature-gated IPC doesn't "break the paradigm" - we're already well into I/O.

**Additional considerations**

- the blocks could be announced even before their insertion into the ledger - they could be streamed optimistically, with a tiny confirmation message being sent to the IPC once they are fully accepted
- `bincode` could be changed to some more common `serde-compatible` serialization format
- the stream currently also receives information on the baked block's height, in order to avoid having to deserialize the block; would any additional information (e.g. the hash) be practical by default?

### DIFF
--- a/synthesizer/src/vm/helpers/sequential_op.rs
+++ b/synthesizer/src/vm/helpers/sequential_op.rs
@@ -49,7 +49,11 @@ impl<N: Network, C: ConsensusStorage<N>> VM<N, C> {
                         let ret = vm.add_next_block_inner(block);
                         #[cfg(feature = "announce-blocks")]
                         if ret.is_ok() {
-                            let _ = announce_block(&mut stream, ipc_payload);
+                            if let Err(e) = announce_block(&mut stream, ipc_payload) {
+                                error!("IPC error: {e}");
+                                // Attempt to restart the stream.
+                                stream = start_block_announcement_stream();
+                            }
                         }
                         SequentialOperationResult::AddNextBlock(ret)
                     }
@@ -164,11 +168,11 @@ fn start_block_announcement_stream() -> Option<Stream> {
 
     match Stream::connect(path) {
         Ok(stream) => {
-            debug!("Successfully started the IPC stream for block announcements");
+            debug!("Successfully (re)started the IPC stream for block announcements");
             Some(stream)
         }
         Err(e) => {
-            warn!("Couldn't start the IPC stream for block announcements: {e}");
+            warn!("Couldn't (re)start the IPC stream for block announcements: {e}");
             None
         }
     }
--- a/synthesizer/src/vm/helpers/sequential_op.rs
+++ b/synthesizer/src/vm/helpers/sequential_op.rs
@@ -45,7 +45,7 @@ impl<N: Network, C: ConsensusStorage<N>> VM<N, C> {
                 let ret = match op {
                     SequentialOperation::AddNextBlock(block) => {
                         #[cfg(feature = "announce-blocks")]
-                        let ipc_payload = (block.height(), bincode::serialize(&block).unwrap()); // Infallible.
+                        let ipc_payload = (block.height(), bincode::serialize(&block));
                         let ret = vm.add_next_block_inner(block);
                         #[cfg(feature = "announce-blocks")]
                         if ret.is_ok() {
@@ -179,9 +179,13 @@ fn start_block_announcement_stream() -> Option<Stream> {
 }
 
 #[cfg(feature = "announce-blocks")]
-fn announce_block(stream: &mut Option<Stream>, payload: (u32, Vec<u8>)) -> Result<bool> {
+fn announce_block(
+    stream: &mut Option<Stream>,
+    payload: (u32, Result<Vec<u8>, Box<bincode::ErrorKind>>),
+) -> Result<bool> {
     if let Some(stream) = stream {
-        let (block_height, block_bytes) = payload;
+        let (block_height, serialized_block) = payload;
+        let block_bytes = serialized_block?;
         debug!("Announcing block {block_height} to the IPC stream");
         let payload_size = u32::try_from(4 + block_bytes.len()).unwrap(); // Safe - blocks are smaller than 4GiB.
         stream.write_all(&payload_size.to_le_bytes())?;
--- a/synthesizer/src/vm/helpers/sequential_op.rs
+++ b/synthesizer/src/vm/helpers/sequential_op.rs
@@ -187,7 +187,7 @@ fn announce_block(
         let (block_height, serialized_block) = payload;
         let block_bytes = serialized_block?;
         debug!("Announcing block {block_height} to the IPC stream");
-        let payload_size = u32::try_from(std::mem::size_of::<u32>() + block_bytes.len()).unwrap()); // Safe - blocks are smaller than 4GiB.
+        let payload_size = u32::try_from(std::mem::size_of::<u32>() + block_bytes.len()).unwrap(); // Safe - blocks are smaller than 4GiB.
         stream.write_all(&payload_size.to_le_bytes())?;
         stream.write_all(&block_height.to_le_bytes())?;
         stream.write_all(&block_bytes)?;
--- a/synthesizer/src/vm/helpers/sequential_op.rs
+++ b/synthesizer/src/vm/helpers/sequential_op.rs
@@ -16,8 +16,12 @@
 use crate::vm::*;
 use console::network::prelude::Network;
 
+#[cfg(feature = "announce-blocks")]
+use interprocess::local_socket::{self, Stream, prelude::*};
 use std::{fmt, thread};
 use tokio::sync::oneshot;
+#[cfg(feature = "announce-blocks")]
+use tracing::*;
 
 impl<N: Network, C: ConsensusStorage<N>> VM<N, C> {
     /// Launches a thread dedicated to the sequential processing of storage-related
@@ -26,6 +30,9 @@ impl<N: Network, C: ConsensusStorage<N>> VM<N, C> {
         &self,
         request_rx: mpsc::Receiver<SequentialOperationRequest<N>>,
     ) -> thread::JoinHandle<()> {
+        #[cfg(feature = "announce-blocks")]
+        let mut stream = start_block_announcement_stream();
+
         // Spawn a dedicated thread.
         let vm = self.clone();
         thread::spawn(move || {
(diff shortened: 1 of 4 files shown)


kaimast:
Have you thought about exposing a simple high-level function, such as `wait_for_next_blocik`, instead? That would avoid having to pick a specific communication mechanism for this feature.

> the blocks could be announced even before their insertion into the ledger - they could be streamed optimistically, with a tiny confirmation message being sent to the IPC once they are fully accepted

_Somewhat of a tangent:_ For validators, snarkOS sync does not insert blocks into the ledger until they are confirmed by the next block. The goal is for clients to eventually move to the same, safer, mechanism.
It would make sense to think about whether we can propagate unconfirmed blocks to reduce latency in the future, and if snarkVM should have a notion of unconfirmed blocks.

ljedrz:
> Have you thought about exposing a simple high-level function, such as wait_for_next_blocik, instead? That would avoid having to pick a specific communication mechanism for this feature.

I'm not sure if this is desirable, as other potential mechanisms (e.g. the network) could be too slow to introduce at this stage, as it could slow down the validator. This way we can ensure that the announcements are near-instantaneous.

kaimast:
I was thinking of something very simple, like adding a condition variable (or `tokio::sync::Notify`) to `Ledger`. There is already a lock around [`current_block`](https://github.com/ProvableHQ/snarkVM/blob/ce07a2c8edc5f5125b3cc6edac9f920b55b9ca5c/ledger/src/lib.rs#L139). Whenever `current_block is updated, you could call `notify` and then have a function like this.

```rust
async fn wait_for_block_height(&self, next_height: u32) {
       while self.current_block.read().height() < next_height {
                self.block_notify.notified().await;
       }
}
```

One could then spawn a task in snarkOS, or whatever uses snarkVM, that waits on this function and executes some custom logic (like writing to a UNIX socket).

ljedrz:
Part of the design is to not have to have a dependency on snarkVM at all, it's a large library.

ljedrz:
Rebased, now that the [former base](github.com/ProvableHQ/snarkVM/pull/2975) is in. It was clean, and there were no other changes.

synthesizer/src/vm/helpers/sequential_op.rs
```diff
@@ -37,7 +44,17 @@ impl<N: Network, C: ConsensusStorage<N>> VM<N, C> {
                 // Perform the queued operation.
                 let ret = match op {
                     SequentialOperation::AddNextBlock(block) => {
+                        #[cfg(feature = "announce-blocks")]
+                        let ipc_payload = (block.height(), bincode::serialize(&block).unwrap()); // Infallible.
                         let ret = vm.add_next_block_inner(block);
+                        #[cfg(feature = "announce-blocks")]
+                        if ret.is_ok() {
+                            if let Err(e) = announce_block(&mut stream, ipc_payload) {
```
niklaslong, line 52: Could blocks be skipped in this case if it takes longer than an iteration to restart the stream?

ljedrz: This happens implicitly; if the stream goes down, the block is not announced, and the process resumes only once it is recreated.

synthesizer/src/vm/helpers/sequential_op.rs
```diff
@@ -37,7 +44,17 @@ impl<N: Network, C: ConsensusStorage<N>> VM<N, C> {
                 // Perform the queued operation.
                 let ret = match op {
                     SequentialOperation::AddNextBlock(block) => {
+                        #[cfg(feature = "announce-blocks")]
+                        let ipc_payload = (block.height(), bincode::serialize(&block).unwrap()); // Infallible.
```
niklaslong: Could you add why this is infallible to the comment?

ljedrz: You mean the call to `bincode::serialize`? Sure, I can comment on it.

ljedrz: Actually, since it has the _potential_ to be faulty (in case a bug slips into block serialization code), I refactored it a bit so that it can fail gracefully.

niklaslong (approved):
LGTM! For visibility @ljedrz could you mention how your testing has impacted the validator in the following cases: 
1. the indexer is not connected to the validator but the feature is enabled
2. similarly the indexer goes down mid-stream
3. would a large block impact consensus at all (serialisation cost)?

ljedrz:
> the indexer is not connected to the validator but the feature is enabled

Every time a block is baked, a `WARN` log is displayed indicating that the feature is active, but the IPC channel is not available.

> similarly the indexer goes down mid-stream

The same thing happens, there's just an additional initial `ERROR` log indicating a socket write error. If the indexer is restarted, the publishing resumes on its own.

> would a large block impact consensus at all (serialisation cost)?

I've calculated the average serialization speed in a test run, and got ~857MiB/s (and this was in a backfilling run with parallel serialization, and small blocks, both of which were certain to decrease the calculated speed). This means that a 1MiB block would be serialized in ~1.1ms, and that's with a machine below validator specs.

synthesizer/src/vm/helpers/sequential_op.rs
```diff
@@ -138,3 +155,46 @@ pub enum SequentialOperationResult<N: Network> {
...
+#[cfg(feature = "announce-blocks")]
+fn announce_block(
+    stream: &mut Option<Stream>,
+    payload: (u32, Result<Vec<u8>, Box<bincode::ErrorKind>>),
+) -> Result<bool> {
+    if let Some(stream) = stream {
+        let (block_height, serialized_block) = payload;
+        let block_bytes = serialized_block?;
+        debug!("Announcing block {block_height} to the IPC stream");
+        let payload_size = u32::try_from(4 + block_bytes.len()).unwrap(); // Safe - blocks are smaller than 4GiB.
```
kaimast: nit: avoid magic number here

```suggestion
        let payload_size = u32::try_from(std::mem::size_of::<u32>() + block_bytes.len()).unwrap()); // Safe - blocks are smaller than 4GiB.
```

kaimast: Also, why do we panic here if the payload here is too large, but not, for example, if serialization of the block failed?

ljedrz: I don't think a 4GiB block is possible at all

synthesizer/src/vm/helpers/sequential_op.rs
```diff
@@ -138,3 +155,46 @@ pub enum SequentialOperationResult<N: Network> {
         )>,
     ),
 }
+
+#[cfg(feature = "announce-blocks")]
+fn start_block_announcement_stream() -> Option<Stream> {
+    let path = std::env::var("BLOCK_ANNOUNCE_PATH")
```
raychu86, line 161: Side note: We should probably have a unified source that documenst all the environment variables that are used.

kaimast:
I think we should document what functionality the new feature enables somewhere. You could use [this README change](https://github.com/ProvableHQ/snarkVM/blob/1d3147e1d5dae4e84cbad8157dc4ad2d69c1d2f8/README.md#21-feature-flags) from the `tracing` PR as a basis. 

You might want to add a note that this is an "unstable" feature and that the wire format can change in the future, just to be safe.

ljedrz:
I filed a [dedicated issue](https://github.com/ProvableHQ/snarkVM/issues/3031) to later document this feature, so as to not cause any conflict with the changes linked by @kaimast.

vicsn:
Can we close this in favour of https://github.com/ProvableHQ/snarkVM/pull/3190 ? If anyone is against let me know

ljedrz:
@vicsn I consider this approach to be much simpler than the plugin, but as long as the new plumbing is useful directly within snarkVM, I don't have objections. It's basically UNIX vs batteries-included, each with its own perks.

miazn:
> @vicsn I consider this approach to be much simpler than the plugin, but as long as the new plumbing is useful directly within snarkVM, I don't have objections. It's basically UNIX vs batteries-included, each with its own perks.

Not sure if it's a 1/1 comparison, i think the complexity is worth it for the mapping updates, which you can't do using block stream without an external snarkVM dependency. It would be worth it, imo, to add block streams to the plugin structure as well

ljedrz:
@miazn precisely - the approach in this PR would only provide blocks to be used by external tools :heavy_check_mark: .

iamalwaysuncomfortable (changes requested):
This would be fast, but it's currently too narrow to be useful for indexers or event rpc. Current indexers largely just use the REST API for getting blocks and that works currently.

Beyond that - indexers generally need to index more than just block information (i.e. some info contained in mappings that don't appear in blocks), and generally need to collate info in highly specific ways. 

That implies that for this to provide the information indexers or RPC infra are after:
- This would need to evolve into a bigger project supporting more datatypes.
- Multiple streams would need to be available OR some intense custom serialization (flags marking data types, byte flags marking termination, etc.) would need to be created to make a single stream consumable.

i.e. It'd be hard to avoid making this into a meta-protocol to provide info beyond blocks. 

I do appreciate the thought experiment here. I think it's wise to continue that kind of experimentation to help us build concrete mental models of the correct approaches to things like richer chain RPC by debating the merits of the possible approaches.

PR #3003, opened 2025-11-06
Translation circuit for dynamic dispatch

Antonio95:
This PR adds a `TranslationAssignment` that can be converted into a circuit which takes in a static record, a dynamic one and some auxiliary data and performs the necessary checks between them. A number of adjacent changes were necessary:

- Sister functions `console::Record::merkleize` and `circuit::Record::merkleize` which compute the data root as described in the design doc.
- A function to_id for `circuit::DynamicRecord` which mimics its preexisting console counterpart.
- A function `compare_constraints` that takes two assignments and checks that the underlying constraints (but not their assignment-specific values!) coincide. This is important to ensure the translation circuit depends on the things it should but doesn't depend on the ones it shouldn't (which is also tested).

Tests in
- `synthesizer/process/src/trace/translation/tests.rs`
- `circuit/program/src/data/record/merkleize/tests.rs`

(you can tell I had fun writing record test cases, although they got a bit drier after I removed all `StringType`s and entries like `target: "Earth".private` or `name: "warp_rider".private` had to go).

The one outstanding issue is the difference in how data is merkleized in `console::DynamicRecord::from_record` vs. console::Record::merkleize` and `circuit::Record::merkleize`, as discussed with @vicsn and @d0cd privately. For now, I am manually overriding the root of the `DynamicRecord` obtained with `from_record` in the tests: https://github.com/ProvableHQ/snarkVM/blob/a68660bbb3de45fc7ee999cf3d1370a0f4683bab/synthesizer/process/src/trace/translation/tests.rs#L49-L56
Although it's a minor issue, it should of course be addressed before things become stable. The question of merkleization might also be affected by our universal-translation-circuit exploration, so we'll see.


circuit/environment/src/helpers/assignment.rs
```diff
@@ -225,6 +227,72 @@ impl<F: PrimeField> snarkvm_algorithms::r1cs::ConstraintSynthesizer<F> for Assig
     }
 }
 
+/// Asserts whether the underlying constraints of the two assignments are equal
```
d0cd:
```suggestion
/// Asserts whether the underlying constraints of the two assignments are equivalent in structure
```
Is this a better description?

circuit/program/src/data/record/merkleize/mod.rs
```diff
@@ -0,0 +1,73 @@
...
+// limitations under the License.
+
+#[cfg(test)]
+pub mod tests;
+
+use super::*;
+
+impl<A: Aleo> Record<A, Plaintext<A>> {
+    /// Returns the root of the merkleized record data as defined in `DynamicRecord`
+    pub fn merkleize(
```
d0cd: `merkleize` should be implemented in the circuit version of `MerkleTree`.
There is a console version [here](https://github.com/ProvableHQ/snarkVM/blob/ce07a2c8edc5f5125b3cc6edac9f920b55b9ca5c/console/collections/src/merkle_tree/mod.rs#L55). The circuit and console versions should be tested for equivalence.

circuit/program/src/data/record/merkleize/tests.rs
d0cd, line 1: These tests can be moved along with the implementation of `merkleize` to `circuit/collections/src/merkle_tree`. Worth noting that the circuits and tests shouldn't be restricted to records.

console/program/src/data/record/merkleize.rs
```diff
@@ -0,0 +1,74 @@
...
+// limitations under the License.
+
+use snarkvm_console_algorithms::Poseidon2;
+use snarkvm_console_collections::merkle_tree::PathHash;
+
+use super::*;
+
+impl<N: Network> Record<N, Plaintext<N>> {
+    /// Returns the root of the merkleized record data as defined in `DynamicRecord`
+    pub fn merkleize(
```
d0cd: This can likely be removed since console merkle-ization is implemented [here](https://github.com/ProvableHQ/snarkVM/blob/ce07a2c8edc5f5125b3cc6edac9f920b55b9ca5c/console/collections/src/merkle_tree/mod.rs#L55)

circuit/program/src/data/dynamic/record/to_id.rs
d0cd, line 1: Minor point. It would be nice to have equivalence tests for the `console` and `circuit` versions of this function. See `circuit/function_id/mod.rs` in the latest version of the `feat/dynamic-dispatch` branch.

synthesizer/process/src/trace/translation/assignment.rs
```diff
@@ -0,0 +1,198 @@
...
+
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+use super::*;
+
+/// An assignment for the record translation circuit
```
d0cd:
```suggestion
/// An assignment for the record translation circuit.
```
Nit.

synthesizer/process/src/trace/translation/tests.rs
d0cd, line 1: Solid tests! A couple of suggestions
- you could also add arrays to the records to increase "coverage"
- for each of the tests, we want to assert the number of constants, public/private vars, and constraints are in each instance of the translation circuit. See `circuit/program/src/request/verify.rs` in `feat/dynamic-dispatch` for an example of those tests.

circuit/environment/src/helpers/assignment.rs
Antonio95: Not quite: the function really checks equality in a pretty dumb way (which is enough for our purposes). For instance, the two linear combinations
`[(Variable(1), 5)]` (i. e. `5 * x_1`)
and
`[(Variable(1), 3), (Variable(1), 2)]` (i. e. `3 * x_1 + 2 * x_1`)
are equivalent, but the function will regard them as different. However, if two different assigments are used to produce the same R1CS, the resulting linear combinations will be equal, i. e. of the same shape (which is stronger than being equivalent), and that is what we check. Checking for equivalence would involve some amount of compilation, which could be very expensive depending on what one checks: take into account that e. g. two R1CS with the same constraints ordered differently are equivalent; as are two R1CS where witness variables are permuted with respect to one another.

Antonio95: @d0cd Perhaps "Asserts whether _the representations of_ the constraints underlying the two assignments are equal" would be more fitting?

circuit/program/src/data/record/merkleize/mod.rs
Antonio95: Fixed (in 8825918) by switching to the new `circuit::MerkleTree` structure

circuit/program/src/data/record/merkleize/tests.rs
Antonio95: These tests were checking the result of `merkleize` as described in the design doc (with each level being padded with at most 1 node) by computing the expected root manually. I think analogous (manual) tests are not needed now that we've switched to `circuit::MerkleTree`, which mirrors `console::MerkleTree`. What I have now added are consistency tests between the two. How does that sound?

console/program/src/data/record/merkleize.rs
Antonio95: Indeed, removed 👍  Thanks

synthesizer/process/src/trace/translation/assignment.rs
Antonio95: Fixed, thanks!

circuit/program/src/data/dynamic/record/to_id.rs
Antonio95: Done!

synthesizer/process/src/trace/translation/tests.rs
Antonio95: Done! :)

Antonio95:
@d0cd Ready for final review :)

circuit/environment/src/helpers/assignment.rs
d0cd: That sounds good!

circuit/program/src/data/dynamic/record/to_id.rs
```diff
@@ -0,0 +1,90 @@
...
+            let console_id = console_record.to_id(function_id, tvk, index).unwrap();
+            assert_eq!(circuit_id.eject_value(), console_id);
+        }
+    }
+
+    #[test]
+    fn test_to_id() {
+        test_to_id_with_mode(Mode::Constant);
+        test_to_id_with_mode(Mode::Public);
+        test_to_id_with_mode(Mode::Private);
```
d0cd: Nit. Consider testing constraint counts. Feel free to leave a TODO.

circuit/collections/src/merkle_tree/mod.rs
```diff
@@ -67,16 +67,196 @@ impl<E: Environment, const DEPTH: u8> Eject for MerklePath<E, DEPTH> {
...
+    empty_hash: Field<E>,
+    /// The number of hashed leaves in the tree.
+    number_of_leaves: usize,
+}
+
+impl<E: Environment, LH: LeafHash<E, Hash = PH::Hash>, PH: PathHash<E, Hash = Field<E>>, const DEPTH: u8> MerkleTree<E, LH, PH, DEPTH> {
+
+    #[inline]
+    /// Initializes a new Merkle tree with the given leaves.
+    pub fn new(leaf_hasher: LH, path_hasher: PH, leaves: &[LH::Leaf]) -> Result<Self> {
```
d0cd, line 90: For my own understanding, are there any differences between this implementation and the console one?

circuit/collections/src/merkle_tree/mod.rs
```diff
@@ -134,4 +314,46 @@ mod tests {
...
+            }).collect_vec();
+
+            let circuit_tree = MerkleTree::<Circuit, CircuitLH, CircuitPH, DEPTH>::new(circuit_leaf_hasher, circuit_path_hasher, &circuit_leaves).unwrap();
+
+            assert_eq!(*console_tree.root(), circuit_tree.root().eject_value());
+        }
+    }
+
+    #[test]
+    fn test_merkle_tree_compatibility_circuit_console() {
```
d0cd, line 372: Nit. Consider testing constraint counts. Feel free to leave a TODO.

d0cd (approved):
Great work on this PR! Looks good to me modulo the nits, which you are welcome to take or leave.

circuit/collections/src/merkle_tree/mod.rs
Antonio95: About this and the analogous comment for `test_to_id` above: I have added the count checks and set them with the UPDATE_COUNT env variable, but I'm getting strange behaviour and the count assertions fail. The updated tests with count checks are pushed, but let me know if you can fix them easily and otherwise we can remove them and leave a TODO.

To give an example of the odd behaviour: for some reason, the first time I run a circuit I get a different (higher) number of constants than after all subsequent iterations. More concretely, this test:

Screenshot 2025-11-13 at 10 03 19

which calls this function:

Screenshot 2025-11-13 at 10 03 09

results in this:

Screenshot 2025-11-13 at 10 03 34

That causes the count assertion to fail. Let me know if you find the solution easily, otherwise we can comment out or replace by a TODO for now.

Incidentally, I've cleaned up the example from the above pics and removed the prints; reintroduced the `Mode::Constant` test case and replaced the `TestRng::with_seed` by `TestRng::default` (at some point I thought introducing random field elements as constants might be resulting in different numbers of bits and affecting the constants; but switching to a seeded rng and removing the constant test case didn't help). The failing count assertion is also still there.

Issue #2958, opened 2025-10-14
[Feature] Ensure Verification Keys note unpadded instance size


vicsn:
## 🚀 Feature

This is defense in depth.

Similar to `VarunaVersion`, we can make a `VarunaKeyVersion` enum.

We can enforce the use of the new version from a new consensus height onwards, again looking at how the `VarunaVersion` migration was implemented for inspiration.

PR #3004, opened 2025-11-07
[Feature] Implement `snark.verify` opcode

raychu86:
## Motivation

This PR implements the `snark.verify` and `snark.verify.batch` opcode, which are `finalize`-only operations that verifies a Varuna proof. In order to support this, the array size was also increased from 512 to 2048. These features will activate at `ConsenusVersion::V14`.

### Examples:
`snark.verify r0 r1 r2 into r3`

r0 is the verifying key (represented by a byte array)
r1 is the verifier inputs (represented by an array of field elements)
r2 is the proof (represented by a byte array)
r3 is the verification results (represented by a boolean) 

`snark.verify.batch r0 r1 r2 into r3`

r0 is the array of verifying keys (represented by a 2-D byte array)
r1 is the array of verifier inputs (represented by a 3-D array of field elements)
r2 is the batch proof (represented by a byte array)
r3 is the verification results (represented by a boolean) 

### Additional Notes
Currently this only supports the native Varuna proofs, but the opcode can be expanded to other proof systems in the future.

The user is required to add the default public input `Field::one()` into the verifier inputs.

### TODO

- [x] Determine proper fee pricing
- [x] Implement verification for batch proofs
- [x] Determine the required array size. 2048 is likely not sufficient (For batched proofs using the same circuit, proofs are 957 bytes + 144 bytes per additional proof in the batch)
- [x] V14 Height needs to be properly set for the networks
- [x] Determine if the first public "one" variable needs to be manually added by the user.

## Test Plan

Tests have been added to check the proofs are properly verified and/or rejected.


synthesizer/program/src/lib.rs
```diff
@@ -983,6 +983,41 @@ impl<N: Network> ProgramCore<N> {
...
+            .any(|instruction| has_op(*instruction.opcode()));
+
+        // Determine if any finalize commands or constructor commands contain the new syntax.
+        let command_contains = cfg_iter!(self.functions())
+            .flat_map(|(_, function)| function.finalize_logic().map(|finalize| finalize.commands()))
+            .flatten()
+            .chain(cfg_iter!(self.constructor).flat_map(|constructor| constructor.commands()))
+            .any(|command| matches!(command, Command::Instruction(instruction) if has_op(*instruction.opcode())));
+
+        // Determine if any of the array types exceed the previous maximum length of 32.
```
raychu86:
```suggestion
        // Determine if any of the array types exceed the previous maximum length of 512.
```

synthesizer/program/src/logic/instruction/operation/snark_verify.rs
```diff
@@ -0,0 +1,461 @@
...
+                let bytes = plaintext.as_byte_array()?;
+                // Deserialize the proof.
+                Proof::<N>::from_bytes_le(&bytes)
+            }
+            _ => bail!("Expected the third operand to be a byte array."),
+        };
+
+        match $variant {
+            SnarkVerifyVariant::Varuna      => verifying_key()?.verify($function_name, $varuna_version, &inputs()?, &varuna_proof()?),
+            SnarkVerifyVariant::VarunaBatch => VerifyingKey::verify_batch(
```
vicsn: The following is all benchmarked on a macOS M2 Max, but still representative for validators.

Verification time of a batch proofs increases ~linearly based on the **number** of batched assignments, and quickly grows larger than the desired blocktime. This is because we Poseidon hash all VK commitments and public inputs.

Based on the current Varuna proof system, we should set a limit of 32 circuits or 100 instances (spread across the circuits) - which is roughly what we allow now in Executions, and takes up to ~100ms to verify. A check in `fn check_transaction` might be cleanest.

Please let me know a.s.a.p. in case we need more than that, in which case we could prioritize the following - which both will require a proof system and therefore wallet migration:
- caching a VK commitments hash https://github.com/ProvableHQ/snarkVM/pull/2964
- precomputing public inputs hashes

| Number of circuits | Number of instances per circuit | verify_batch() time (ms) |   |   |
|--------------------|---------------------------------|--------------------------|---|---|
|                  1 |                               1 |                        8 |   |   |
|                  1 |                              10 |                       10 |   |   |
|                  1 |                             100 |                       40 |   |   |
|                  1 |                             200 |                       73 |   |   |
|                 10 |                               1 |                       27 |   |   |
|                 10 |                              10 |                       77 |   |   |
|                 10 |                             100 |                      599 |   |   |
|                 10 |                             200 |                     1180 |   |   |
|                 50 |                             100 |                     8200 |   |   |
|                100 |                               1 |                      482 |   |   |
|                100 |                              10 |                     3207 |   |   |
|                100 |                             100 |                    27730 |   |   |
|                100 |                             200 |                    54700 |   |   |
|                200 |                               1 |                     1490 |   |   |
|                200 |                              10 |                   105000 |   |   |

synthesizer/program/src/logic/instruction/operation/snark_verify.rs
```diff
@@ -0,0 +1,461 @@
...
+
+// Perform the snark verification based on the variant.
+#[rustfmt::skip]
+macro_rules! do_snark_verification {
+    ($variant: expr, $function_name: expr, $varuna_version: expr, $verifying_key: expr, $inputs: expr, $proof: expr) => {{
+        let verifying_key = || match $verifying_key {
+            Value::Plaintext(plaintext) => VerifyingKey::<N>::from_bytes_le(&plaintext.as_byte_array()?),
+            _ => bail!("Expected the first operand to be a byte array."),
+        };
+        let verifying_keys = || match $verifying_key {
```
vicsn, line 118: Should we bail on duplicate verifying keys? I don't recall if this will fail but it would be bad UX anyhow.

synthesizer/program/src/logic/instruction/operation/snark_verify.rs
```diff
@@ -0,0 +1,461 @@
...
+    pub fn destinations(&self) -> Vec<Register<N>> {
+        vec![self.destination.clone()]
+    }
+}
+
+// Perform the snark verification based on the variant.
+#[rustfmt::skip]
+macro_rules! do_snark_verification {
+    ($variant: expr, $function_name: expr, $varuna_version: expr, $verifying_key: expr, $inputs: expr, $proof: expr) => {{
+        let verifying_key = || match $verifying_key {
```
vicsn, line 114: This is already a minor issue before this PR, but in `CircuitVerifyingKey::read_le` we should check that `circuit_commitments` contains 12 elements. This reduces the change of OOM bombs.

synthesizer/program/src/logic/instruction/operation/snark_verify.rs
```diff
@@ -0,0 +1,461 @@
...
+                    .collect::<Result<Vec<VerifyingKey<N>>, _>>()
+            }
+            _ => bail!("Expected the first operand to be a two-dimensional byte array."),
+        };
+
+        let inputs = || match $inputs {
+            Value::Plaintext(plaintext) => Ok(plaintext.as_field_array()?.into_iter().map(|f| *f).collect::<Vec<_>>()),
+            _ => bail!("Expected the second operand to be an array of fields."),
+        };
+        let batch_inputs = || match $inputs {
```
vicsn, line 139: Directly exposing key uploads and proof verification risks a serious malleability issue unfortunately because we pad public inputs with 0s up to their next power of two.

I'm confident that our existing VM is not at risk because our existing verifier inputs are not variable-length (along with more reasons). Whether snark.verify is at risk depends on the application using it.

We can either fix this in the VerifyingKey (which requires a proof system and therefore wallet migration) - CC https://github.com/provableHQ/snarkVM/issues/2958

Or you can require these inputs in the CircuitVerifyingKey to be a multiple of 2. However:
- We will also want to double check, test and document whether the first public "one" variable is expected to be submitted at this level of abstraction. 
- At the application level you'll have to instantiate some random unused variables for this.

synthesizer/src/vm/tests/test_v13.rs
```diff
@@ -0,0 +1,412 @@
...
+program test_snark_verify.aleo;
+
+function verify_proof:
+    input r0 as [u8; 673u32].private;
+    input r1 as [field; 2u32].private;
+    input r2 as [u8; 957u32].private;
+    async verify_proof r0 r1 r2 into r3;
+    output r3 as test_snark_verify.aleo/verify_proof.future;
+finalize verify_proof:
+    input r0 as [u8; 673u32].public;
```
vicsn: Can you add a test where the verifying key is stored in a mapping, and later retrieved for verification? Can you add negative tests for weird edge cases?

vicsn (changes requested):
Do you have any recommendation for how explorers should display verifying keys (in e.g. mappings) to users/developers, for them to easily be able to verify their correctness? 

Similar to https://github.com/ProvableHQ/leo/issues/28664 which is on the roadmap for Q1, the explorer could verify and display the leo program source code for such stored verifying keys - which I assume we'll want to make effectively immutable.

CC @miazn

synthesizer/src/vm/tests/test_v13.rs
```diff
@@ -0,0 +1,412 @@
...
+
+#[test]
+fn test_snark_verify() {
+    // Define the verification program.
+    let program = Program::from_str(
+        r"
+program test_snark_verify.aleo;
+
+function verify_proof:
+    input r0 as [u8; 673u32].private;
```
vicsn: If we want others to be able to use this, and for easier readability, we can consider creating `varuna.v2.vk.size` and `varuna.v2.proof.size` literals - this can also be future work.

synthesizer/src/vm/tests/test_v13.rs
```diff
@@ -0,0 +1,412 @@
...
+
+#[test]
+fn test_snark_verify_batch() {
+    // Define the verification program.
+    let program = Program::from_str(
+        r"
+program test_snark_verify_batch.aleo;
+
+function verify_batch_proof:
+    input r0 as [[u8; 673u32]; 1u32].private;
```
vicsn: A backwards compatibility topics to think about: as some of the other comments note, we may add a field to the VK object soon™️ , changing the requires array size.

Ideally, `snark.verify` takes in a `varuna_version` argument so we don't have to deploy new opcodes.

Moreover, it's worth thinking about impacts on deploying and execution. For this particular `function verify_batch_proof` function, the developer would have to deploy a new function because the function signature is impacted. I can imagine it would be efficient to have a child program with various `function submit_vk_v1_into_mapping`, which may need to be redeployed, and some parent program with `function submit_proof`.

There are other things we can do if we diverge further away from the current VM design, but I'll stop here.

mwyatt896:
Super exciting stuff y'all!

synthesizer/program/src/logic/instruction/operation/snark_verify.rs
raychu86: I'd leave that to another PR as it's not exactly the same scope as this.

synthesizer/program/src/logic/instruction/operation/snark_verify.rs
raychu86: I ran a test and duplicate VKs will not work due to the following checks:

```
        let keys_to_inputs: BTreeMap<_, _> =
            inputs.iter().map(|(verifying_key, inputs)| (verifying_key.deref(), inputs.as_slice())).collect();
        ensure!(keys_to_inputs.len() == num_expected_keys, "Incorrect number of verifying keys for batch proof");
```

This instruction will fail during the `VerifyingKey::verify_batch` call if duplicate VKs are provided.

synthesizer/program/src/logic/instruction/operation/snark_verify.rs
raychu86: Currently unclear for the number of batched instances we need. But for now I will place a placeholder check with a bound that can be changed.

For now I've set the max num circuits to be `Transaction::<N>::MAX_TRANSITIONS` - [10ec834](https://github.com/ProvableHQ/snarkVM/pull/3004/commits/10ec834c634b03d3b21cf34805f2998036646837)

synthesizer/src/vm/tests/test_v13.rs
raychu86: This can also be a `Leo` level abstraction rather than an Aleo instructions one.

synthesizer/program/src/lib.rs
raychu86: [10ec834](https://github.com/ProvableHQ/snarkVM/pull/3004/commits/10ec834c634b03d3b21cf34805f2998036646837)

raychu86:
> Do you have any recommendation for how explorers should display verifying keys (in e.g. mappings) to users/developers, for them to easily be able to verify their correctness?

This should be up to the developer to parse manually since these are effectively generic byte arrays. We can in theory try to parse byte arrays of size `673u32` or create a custom VK type, but I don't see it being that useful. 

> Similar to https://github.com/ProvableHQ/leo/issues/28664 which is on the roadmap for Q1, the explorer could verify and display the leo program source code for such stored verifying keys - which I assume we'll want to make effectively immutable.

Immutability of the mapping should be up to the program developer. We also can't really verify the VK because we don't necessarily know the circuit assignment.

synthesizer/program/src/logic/instruction/operation/snark_verify.rs
raychu86: The cop-out answer is that the lengths are developer defined. Wouldn't the use of the underlying `VarunaSnark::verify_batch` also do the same padding and checks? 

> We will also want to double check, test and document whether the first public "one" variable is expected to be submitted at this level of abstraction.

Good call on the first public "one" variable. I will consider adding it by default

raychu86: I've decided to leave out the prepended "one" variable, as it is a special case for our proof system and we can enforce that the developer/user adds it themselves. This is helpful because they should be doing verification natively as well which will require creating verifier inputs that include the "one".

synthesizer/src/vm/tests/test_v13.rs
raychu86: Done in [8291d31](https://github.com/ProvableHQ/snarkVM/pull/3004/commits/8291d319bc1aca5470047747283eb0bcc1069002).

More tests to follow.

synthesizer/src/vm/tests/test_v13.rs
raychu86: VarunaVersion added in [8ee686a](https://github.com/ProvableHQ/snarkVM/pull/3004/commits/8ee686a8866ac8dfff0217aa95bada6bd178357a)

raychu86: I would say the array size differences is fine to require developers to upgrade the program, since the underlying assumptions will be different for them as well.

synthesizer/program/src/logic/instruction/operation/snark_verify.rs
```diff
@@ -27,6 +27,10 @@ pub type SnarkVerify<N> = SnarkVerification<N, { SnarkVerifyVariant::Varuna as u
 /// Computes whether a `batch_proof` is valid for the given `verifying_keys` and `public inputs`.
 pub type SnarkVerifyBatch<N> = SnarkVerification<N, { SnarkVerifyVariant::VarunaBatch as u8 }>;
 
+// TODO (raychu86): SnarkVerify - Consider increasing this limit in the future.
+/// The maximum number of `snark.verify` circuits supported in a batch verification.
+pub const MAX_SNARK_VERIFY_CIRCUITS: u32 = 32;
```
vicsn: Nit: I would just move MAX_TRANSITIONS inside of Network and use it directly, it'll save a future developer a few minutes of work and a CI roundtrip on the failing unit test.

synthesizer/program/src/logic/instruction/operation/snark_verify.rs
```diff
@@ -314,7 +332,20 @@ impl<N: Network, const VARIANT: u8> SnarkVerification<N, VARIANT> {
...
-        // Ensure the third operand is an array of bytes.
+        // Check the number of batched instances is correct.
+        ensure!(
+            num_circuits == num_vks,
+            "Instruction '{}' expects the number of circuits ({num_circuits}) to match the number of verifying keys ({num_vks}).",
+            Self::opcode()
+        );
+        // Check that the number of circuit instances is properly bound.
+        ensure!(
+            num_circuits < MAX_SNARK_VERIFY_CIRCUITS,
```
vicsn: We also need to ensure the total number of instances is at most 100.

This is fine:

    vk1 or circuit1: 10 instances
    vk2 or circuit2: 90 instances

This is not:

    vk1 or circuit1: 11 instances
    vk2 or circuit2: 90 instances

synthesizer/program/src/logic/instruction/operation/snark_verify.rs
vicsn: Ok! https://github.com/ProvableHQ/snarkVM/issues/3083

synthesizer/program/src/logic/instruction/operation/snark_verify.rs
vicsn: > Wouldn't the use of the underlying VarunaSnark::verify_batch also do the same padding and checks?

Not sure I'm following. We cannot add `ensure!(public_inputs.len().is_power_of_two()` inside of `verify_batch` because our *existing* function executions have inputs which are not a power of two.

> The cop-out answer is that the lengths are developer defined.

Hmm I realise now that you may be referring to the fact that the aleo instructions encodes the public inputs length as e.g. `[field; 8u32].public;`. That does mitigate the concern considerably. Do you mind documenting the concern for future developers not to waste time on this above `SnarkVerifyBatch`?

synthesizer/program/src/logic/instruction/operation/snark_verify.rs
raychu86: Done in [8ee686a](https://github.com/ProvableHQ/snarkVM/pull/3004/commits/8ee686a8866ac8dfff0217aa95bada6bd178357a)

synthesizer/program/src/logic/instruction/operation/snark_verify.rs
raychu86: Addressed with [4da0200](https://github.com/ProvableHQ/snarkVM/pull/3004/commits/4da0200f174c26fcf9325012211a480a51aec674).

The opcodes now support padding and including inputs that exceeds the number of public inputs defined by the vk. If the number of provided inputs exceeds `VK.circuit_info.num_public_inputs`, then `snark.verify.*` will check that all subsequent input values are zero and trim them.

synthesizer/program/src/logic/instruction/operation/snark_verify.rs
raychu86: Resolved with [b6335bd](https://github.com/ProvableHQ/snarkVM/pull/3004/commits/b6335bd621284fcf880016363dc273aee661295c). As it is not really a strong dependency

synthesizer/program/src/logic/instruction/operation/snark_verify.rs
```diff
@@ -29,7 +29,9 @@ pub type SnarkVerifyBatch<N> = SnarkVerification<N, { SnarkVerifyVariant::Varuna
 
 // TODO (raychu86): SnarkVerify - Consider increasing this limit in the future.
 /// The maximum number of `snark.verify` circuits supported in a batch verification.
-pub const MAX_SNARK_VERIFY_CIRCUITS: u32 = 32;
+pub const MAX_SNARK_VERIFY_CIRCUITS: u32 = 1 << 5;
+/// The maximum number of `snark.verify` instances supported in a batch verification.
+pub const MAX_SNARK_VERIFY_INSTANCES: u32 = 1 << 7;
```
vicsn: We may want to document here that an `Execution` itself may also come with this many circuits and instances to verify (transitions, record inclusions, record translations). However, that proof is verified earlier in the pipeline, and connecting the two through a common budget would be very cumbersome, so for now it's fine to have two separate budgets.

circuit/program/src/data/future/to_bits.rs
```diff
@@ -39,7 +39,11 @@ impl<A: Aleo> ToBits for Future<A> {
         for argument in &self.arguments {
             let argument_bits = argument.to_bits_le();
             // Write the size of the argument.
-            U16::constant(console::U16::new(argument_bits.len() as u16)).write_bits_le(vec);
+            let argument_length = argument_bits.len();
+            match argument_length <= u16::MAX as usize {
+                true => U16::constant(console::U16::new(argument_length as u16)).write_bits_le(vec),
+                false => U32::constant(console::U32::new(argument_length as u32)).write_bits_le(vec),
+            }
```
raychu86, line 46: @vicsn @d0cd This section in particular warrants a review.

I only updated the bit serialization rather than byte serialization because we do not have `version` variants in serialization that could be used (particularly in circuit).

The bit serialization is used in generating the field repr preimage of the hash. Had to do some custom muxxing, but should be safe in theory because these argument sizes need to be fixed.

raychu86: Also, do we need to increase the limit for the `Plaintext::toBits` value sizes (that are currently bound to u16::MAX)

synthesizer/program/src/logic/instruction/operation/snark_verify.rs
raychu86: [c3f26a3](https://github.com/ProvableHQ/snarkVM/pull/3004/commits/c3f26a352ab2b40e946b676fd7cf36b498f09c09)

circuit/program/src/data/future/to_bits.rs
vicsn: This seems safe indeed because argument sizes are fixed.

Not sure I have context to answer whether the size of other types need to increase. Can you follow the requirements of a motivating unit test?

algorithms/src/snark/varuna/mode.rs
```diff
@@ -42,3 +44,20 @@ pub enum VarunaVersion {
     V1 = 1,
     V2 = 2,
 }
+
+impl ToBytes for VarunaVersion {
+    fn write_le<W: io::Write>(&self, writer: W) -> io::Result<()> {
+        (*self as u8).write_le(writer)
```
d0cd, line 51: Do we need to `repr(u8)` above?

console/program/src/data/future/to_bits.rs
```diff
@@ -61,15 +69,23 @@ impl<N: Network> ToBits for Future<N> {
             .write_bits_be(vec);
         vec.extend_from_slice(&function_name_bits);
 
-        // Write the number of arguments.
+        // Write the number of arguments.s
```
d0cd:
```suggestion
        // Write the number of arguments.
```

console/program/src/data/future/mod.rs
```diff
@@ -65,4 +65,9 @@ impl<N: Network> Future<N> {
     pub fn arguments(&self) -> &[Argument<N>] {
         &self.arguments
     }
+
+    /// Returns true if the bit size of any argument exceeds `u16::MAX`.
+    pub fn argument_bit_size_exceeds_u16(&self) -> bool {
```
d0cd: Is this used anywhere?

console/program/src/data/plaintext/mod.rs
```diff
@@ -71,6 +71,40 @@ impl<N: Network> Plaintext<N> {
...
+                        _ => bail!("Expected a u8 array, found a non-u8 element."),
+                    }
+                }
+                Ok(bytes)
+            }
+            _ => bail!("Expected a u8 array, found a non-array plaintext."),
+        }
+    }
+
+    /// Returns the `Plaintext` as a `Vec<N::Field>`, if it is a u8 array.
```
d0cd:
```suggestion
    /// Returns the `Plaintext` as a `Vec<N::Field>`, if it is a field array.
```

circuit/program/src/data/future/to_bits.rs
d0cd: Increasing `Plaintext::ToBits` is an option, if you want expand scope.
By increasing the bound here we are supporting much larger nested `Future`s and **slightly** larger `Plaintext` arguments. 

On a separate note, if we care about recovery of `ToBits` (which we haven't paid much attention to in the past), we can consider introducing a "version" byte with a zero in front of an identifier. Similar to dynamic disptach.

console/program/src/data_types/finalize_type/size_in_bits.rs
d0cd: This needs to be updated to account for the `U32` length.

d0cd: Maybe we can leave some comments in the code base to help ensure that these impls stay in sync. 
We already have tests for existing cases, but this doesn't help check for expansions.

synthesizer/tests/expectations/vm/execute_and_finalize/test_rand.out
d0cd, line 1: Gotta regen but lowkey...
https://github.com/ProvableHQ/snarkVM/issues/3130

synthesizer/src/vm/verify.rs
```diff
@@ -256,21 +264,34 @@ impl<N: Network, C: ConsensusStorage<N>> VM<N, C> {
...
+                    // Check that all future argument bit sizes do not exceed the maximum allowed size of u16::MAX.
+                    let stack = Stack::new(&self.process().read(), deployment.program())?;
+                    check_future_argument_bit_size(deployment.program(), &stack, u16::MAX as usize)?;
+                }
+
+                // Determine if any of the array types exceed the maximum array elements.
+                let max_array_elements = consensus_config_value!(N, MAX_ARRAY_ELEMENTS, current_block_height)
+                    .ok_or(anyhow!("Failed to retrieve the maximum array elements"))?;
+                ensure!(
+                    !deployment.program().exceeds_max_array_size(u32::try_from(max_array_elements)?),
```
d0cd: As I understand it, we don't have to perform this when we're on the latest consensus version. However if we don't guard this check, it happens for all deployments.
Deployments don't happen frequently but still could be worth the savings.

synthesizer/program/src/logic/instruction/operation/snark_verify.rs
```diff
@@ -0,0 +1,579 @@
...
+    verifying_key: &Value<N>,
+    varuna_version: Value<N>,
+    inputs: &Value<N>,
+    proof: &Value<N>,
+) -> Result<bool> {
+    Ok(do_snark_verification!(variant, _function_name, verifying_key, varuna_version, inputs, proof))
+}
+
+// Helper function to check if a type is a N-dimensional array of a given base literal type.
+fn check_nd_array_type<N: Network>(
```
d0cd, line 249: Do we have the opportunity to dedup with the ECDSA work?

synthesizer/program/src/logic/instruction/operation/snark_verify.rs
```diff
@@ -0,0 +1,579 @@
...
+        }
+
+        // Retrieve the inputs.
+        let verifying_key = registers.load(stack, &self.operands[0])?;
+        let varuna_version = registers.load(stack, &self.operands[1])?;
+        let inputs = registers.load(stack, &self.operands[2])?;
+        let proof = registers.load(stack, &self.operands[3])?;
+
+        // Verify the signature.
+        let _function_name = "snark.verify";
```
d0cd, line 310: Is there any significance to the function name? Is there any concern that this is always the same?

synthesizer/src/vm/tests/test_v14.rs
d0cd, line 1: Some ideas for tests:
- load and store VKs to mappings and choose which one based on a selector
- use a VK in a deployment and generate a proof for it. Almost recursive :)

d0cd (approved):
Lefts some comments and nits, looks like its nearly there!

d0cd (changes requested):
Disregard the above approval, meant to select "Request Changes"

synthesizer/program/src/logic/instruction/operation/snark_verify.rs
```diff
@@ -0,0 +1,579 @@
...
+
+        // Check the number of batched instances is correct.
+        ensure!(
+            num_circuits == num_vks,
+            "Instruction '{}' expects the number of circuits ({num_circuits}) to match the number of verifying keys ({num_vks}).",
+            Self::opcode()
+        );
+        // Check that the number of circuit is properly bound.
+        ensure!(
+            num_circuits < MAX_SNARK_VERIFY_CIRCUITS,
```
d0cd: Is this off by 1? Ditto below.

synthesizer/program/src/logic/instruction/operation/snark_verify.rs
```diff
@@ -0,0 +1,579 @@
...
+                    num_instances,
+                )
+            }
+        };
+        if !result {
+            bail!(
+                "Instruction '{}' expects the third input to be {}. Found input of type '{}'",
+                Self::opcode(),
+                expected_type,
+                &input_types[1]
```
d0cd:
```suggestion
                &input_types[2]
```

synthesizer/program/src/logic/instruction/operation/snark_verify.rs
```diff
@@ -0,0 +1,579 @@
...
+        match &input_types[3] {
+            RegisterType::Plaintext(PlaintextType::Array(array_type))
+                if array_type.base_element_type() == &PlaintextType::Literal(LiteralType::U8) =>
+            {
+                // valid byte array
+            }
+            _ => bail!(
+                "Instruction '{}' expects the fourth input to be a byte array. Found input of type '{}'",
+                Self::opcode(),
+                input_types[2]
```
d0cd:
```suggestion
                input_types[3]
```

synthesizer/program/src/logic/instruction/operation/snark_verify.rs
raychu86: Right now it's only for Varuna proofs. 

In the future, we can do other proof systems with syntax like `snark.verify.groth16`

synthesizer/program/src/logic/instruction/operation/snark_verify.rs
raychu86: [9273488](https://github.com/ProvableHQ/snarkVM/pull/3004/commits/92734881b3c187151fb64095a7cc2a258f43eed2)

synthesizer/program/src/logic/instruction/operation/snark_verify.rs
raychu86: [9273488](https://github.com/ProvableHQ/snarkVM/pull/3004/commits/92734881b3c187151fb64095a7cc2a258f43eed2)

synthesizer/program/src/logic/instruction/operation/snark_verify.rs
raychu86: [9273488](https://github.com/ProvableHQ/snarkVM/pull/3004/commits/92734881b3c187151fb64095a7cc2a258f43eed2)

synthesizer/src/vm/verify.rs
raychu86: Addressed in [0563439](https://github.com/ProvableHQ/snarkVM/pull/3004/commits/0563439c9334bfac1afd8897330ea437de3710f6)

algorithms/src/snark/varuna/mode.rs
raychu86: [9273488](https://github.com/ProvableHQ/snarkVM/pull/3004/commits/92734881b3c187151fb64095a7cc2a258f43eed2)

console/program/src/data/future/to_bits.rs
raychu86: [9273488](https://github.com/ProvableHQ/snarkVM/pull/3004/commits/92734881b3c187151fb64095a7cc2a258f43eed2)

console/program/src/data/plaintext/mod.rs
raychu86: [9273488](https://github.com/ProvableHQ/snarkVM/pull/3004/commits/92734881b3c187151fb64095a7cc2a258f43eed2)

console/program/src/data/future/mod.rs
raychu86: [0563439](https://github.com/ProvableHQ/snarkVM/pull/3004/commits/0563439c9334bfac1afd8897330ea437de3710f6)

synthesizer/program/src/logic/instruction/operation/snark_verify.rs
raychu86: I'd leave for future refactoring, but I don't see a big overlap here between usage in the two opcodes.

console/program/src/data_types/finalize_type/size_in_bits.rs
raychu86: Done in [44727d3](https://github.com/ProvableHQ/snarkVM/pull/3004/commits/44727d395fe1b012ce5f041964518db054654442)

circuit/program/src/data/future/to_bits.rs
raychu86: I will be leaving `Plaintext::ToBits` for future work, as it is quite involved and requires updating the `from_bits` impls.

Currently a `Plaintext::Struct` with a byte array element can only have ~1300 elements in that array due to the element bit size limit of u16::MAX.

console/program/src/data_types/finalize_type/size_in_bits.rs
raychu86: Documentation added in [4046c49](https://github.com/ProvableHQ/snarkVM/pull/3004/commits/4046c49378ffb53d262d86ac6a5e2f2fee05d7d6)

console/network/src/consensus_heights.rs
```diff
@@ -51,6 +51,7 @@ pub enum ConsensusVersion {
     V13 = 13,
     /// V14: Increase the program size limit to 512 kB and transaction size limit to 540 kB.
     ///      Introduces `aleo::GENERATOR` and `aleo::GENERATOR_POWERS` opcodes.
+    ///      Increase array size limit to 2048 and introduce snark.verify opcode.
```
vicsn:
```suggestion
    ///      Increase array size limit to 2048, increase the Future argument size limit to 32 bits and introduce snark.verify opcode.
```

vicsn (changes requested):
Because we had so many issues with external struct backwards compatibility, I went through every instance of `the argument` in the code and found two places to extend (nothing too serious):

- `let mut length = (u16::try_from(result.len()).unwrap()).to_bytes_le().unwrap();`, IMO just a single u32 test case would already be sufficient in `test_deeply_nested_future`.
- `preimage.push(Field::from_u16(argument_locator.index()));`, I guess we can just use `from_u32`?

My gut says it's not worth the effort at this to make all of this more strongly typed and tied to each other, because we may make arbitrary changes to the VM, so we will have to continue to rely on a consistent code style + greps.

vicsn:
Can you also merge in `origin/staging`?

raychu86:
> Because we had so many issues with external struct backwards compatibility, I went through every instance of `the argument` in the code and found two places to extend (nothing too serious):
> 
> * `let mut length = (u16::try_from(result.len()).unwrap()).to_bytes_le().unwrap();`, IMO just a single u32 test case would already be sufficient in `test_deeply_nested_future`.
> * `preimage.push(Field::from_u16(argument_locator.index()));`, I guess we can just use `from_u32`?
> 
> My gut says it's not worth the effort at this to make all of this more strongly typed and tied to each other, because we may make arbitrary changes to the VM, so we will have to continue to rely on a consistent code style + greps.


Thank you for doing a pass on this front. Regarding the two findings:

1. We are now constrained by the u16 for bytes, so u16 here is correct for `let mut length = (u16::try_from(result.len()).unwrap()).to_bytes_le().unwrap();`, since this is for byte (de)serialization.
2. `preimage.push(Field::from_u16(argument_locator.index()));` isn't for Futures, rather it is used for Restrictions. "Argument" is an overloaded term here.

If you agree, I don't think these require changing.

console/network/src/consensus_heights.rs
raychu86: [1bff6b6](https://github.com/ProvableHQ/snarkVM/pull/3004/commits/1bff6b6720319460454cc1c2d00ec6245455956d)

vicsn (approved):

synthesizer/program/src/logic/instruction/operation/snark_verify.rs
```diff
@@ -0,0 +1,579 @@
...
+            bail!("Instruction '{}' expects 4 operands, found {} operands", Self::opcode(), self.operands.len())
+        }
+
+        // Retrieve the inputs.
+        let verifying_key = registers.load(stack, &self.operands[0])?;
+        let varuna_version = registers.load(stack, &self.operands[1])?;
+        let inputs = registers.load(stack, &self.operands[2])?;
+        let proof = registers.load(stack, &self.operands[3])?;
+
+        // Verify the signature.
```
d0cd, line 309:
```suggestion
        // Verify the proof.
```

synthesizer/program/src/logic/instruction/operation/snark_verify.rs
```diff
@@ -0,0 +1,579 @@
...
+    /// Returns the opcode.
+    #[inline]
+    pub const fn opcode() -> Opcode {
+        Opcode::Snark(SnarkVerifyVariant::new(VARIANT).opcode())
+    }
+
+    /// Returns the operands in the operation.
+    #[inline]
+    pub fn operands(&self) -> &[Operand<N>] {
+        // Sanity check that there are exactly three operands.
```
d0cd, line 91:
```suggestion
        // Sanity check that there are exactly four operands.
```

synthesizer/src/vm/verify.rs
```diff
@@ -141,7 +141,8 @@ impl<N: Network, C: ConsensusStorage<N>> VM<N, C> {
         lap!(timer, "Check for duplicate elements");
 
         // Get the consensus version.
-        let consensus_version = N::CONSENSUS_VERSION(current_block_height)?;
+        let current_block_height = self.block_store().current_block_height();
```
d0cd, line 144: `current_block_height` is available on L79

d0cd (approved):
LGTM, found some minor nits.

Issue #2996, opened 2025-10-31
[Proposal] Reduce CI time to max ~15 min. per run


vicsn:
## 💥 Proposal

Now that we know which tests take the most amount of time, we can aim to:
- reduce compile time
- change or refactor some of the tests
  - removing the synthesizer job will require some discussion: https://github.com/ProvableHQ/snarkVM/issues/2979
  - perhaps we can turn off proof generation to speed up testing
- https://github.com/ProvableHQ/snarkVM/issues/3089
- https://github.com/ProvableHQ/snarkVM/issues/3090

vicsn:
[all-tests-by-time.csv](https://github.com/user-attachments/files/24590991/all-tests-by-time.csv)

vicsn:
@meddle0x53 can you share the script/command you used to generate that CSV ^ ? CC @mohammadfawaz

meddle0x53:
Yes, here are the scripts I used to imitate the way the CI runs the tests and to time them:

https://github.com/meddle0x53/snarkVM/pull/2

PR #3005, opened 2025-11-07
Improve CircleCI setup

meddle0x53:
## Motivation

As part of https://github.com/ProvableHQ/snarkVM/issues/2996 , we are doing a few experiments with the CircleCI startup and provisioning logic as part of this PR. The idea is to speedup the setup part of the logic and remove some issues with hanging builds.

## Test Plan

Use the CI and its workflows to test that times are better compared to the other PRs at the time.

## Documentation

There are problems with hanging workflows, specifically the windows one, rust compilation is slow and there are repetitions, so trying to figure out how to fix these issues and we'll document the fixes which work.

* The windows workflow now doesn't get stuck (tested with over 30 tries)
* SSCache is now used (there are logs to guarantee that and it is used after a lot of small updates)
* Runtime is consistently under 25 min; Compilation on heavy libs like the synthesizes is now 2+ min faster (sscache hits).

To make it faster I see a few ways now (no other ideas):
1. Simplify and remove bad tests.
2. More partitioning.
3. Maybe bigger compilation optimisation with keeping the target folder, but this can lead to broken builds when there are big changes, so not a good idea.

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -885,20 +885,7 @@ jobs:
             --lib --bins --features test
             --partition count:2/2
             -- --test-threads 8
-          cache_key_suffix: -test2-sccache-debug
-          timeout: 25m
-
-  synthesizer-test-partition2:
-    executor: rust-docker
-    resource_class: << pipeline.parameters.twoxlarge >>
-    steps:
-      - run_test:
-          workspace_member: snarkvm-synthesizer
-          flags: >
-            --lib --bins --features test
-            --partition count:2/2
-            -- --test-threads 8
-          cache_key_suffix: -test2-sccache-debug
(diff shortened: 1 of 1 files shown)


kaimast:
That looks great! I investigated similar improvements at some point (e.g., https://github.com/ProvableHQ/snarkOS/pull/3680) but we never found time to finish and merge them. 

Feel free to tag me as a reviewer once it is ready.

meddle0x53:
@kaimast adding you. But am still figuring out why the sscache is not being hit properly... But looking at your PR, very similar changes, means I am on the right path!

vicsn (approved):
Thank you, this looks great

kaimast (approved):

PR #3007, opened 2025-11-07
[Fix] Remove LoggableError trait

kaimast:
Following the discussion in #2992, this PR removes the `LoggableError` trait. To get a similar behavior, one can still call `flatten_error` and pass it to a logging macro, such as `error!`.


vicsn (approved):
Thank you!

PR #3011, opened 2025-11-08
Skip proof and certificate generation and verification if building with dev_skip_checks

vicsn:
## Motivation

Follow-up to https://github.com/ProvableHQ/snarkVM/pull/2998

In order to accelerate testing of Leo programs, as well as extend prototyping in unit tests, we extend a feature flag `dev_skip_checks` to skip proof and certificate generation and verification.

NOTE: As developer, you will likely still want to confirm that your deployment is within R1CS constraint and variable limits, so we should build tooling for `leo synthesize` which synthesizes and checks all functions in parallel. 

## Test Plan

- [x] Integrate in `leo devnode`

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -175,7 +175,7 @@ dev_skip_checks = [
 save_r1cs_hashes = [ "snarkvm-circuit/save_r1cs_hashes" ]
 test_exports = [ "snarkvm-algorithms/test_exports" ]
 test_targets = [ "snarkvm-console/test_targets" ]
-test_consensus_heights = [ "snarkvm-console/test_consensus_heights" ]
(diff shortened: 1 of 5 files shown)


Roee-87 (approved):
Devnode successfully generates deployments and upgrade transactions using the verifying key and certificate defined in this PR.

raychu86:
Nit: Can you add the `feature = "test_consensus_heights")` additions to the existing execute proof verification checks that only have`= `dev_skip_checks`

vicsn:
> Nit: Can you add the feature = "test_consensus_heights") additions to the existing execute proof verification checks that only have= dev_skip_checks`

I did

raychu86 (approved):

PR #3014, opened 2025-11-10
[Feat] Facilitate RocksDB's secondary instances

ljedrz:
This PR enables support for RocksDB's [secondary instances](https://github.com/facebook/rocksdb/wiki/Read-only-and-Secondary-instances). This could be used by all sorts of tools that require an insight into the storage of a running node.

Filing as a draft, as it depends on a pending [`aleo-std` PR](https://github.com/ProvableHQ/aleo-std/pull/17).

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -426,11 +426,13 @@ path = "wasm"
 version = "=4.3.0"
 
 [workspace.dependencies.aleo-std]
-version = "1.0.1"
+git = "https://github.com/ljedrz/aleo-std.git"
+branch = "feat/secondary_storage"
+features = ["storage"]
 
 [workspace.dependencies.aleo-std-storage]
-default-features=false
-version = "1.0.1"
+git = "https://github.com/ljedrz/aleo-std.git"
+branch = "feat/secondary_storage"
 
 [workspace.dependencies.anyhow]
 version = "1.0.73"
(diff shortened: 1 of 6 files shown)


vicsn:
> secondary instances.

Very cool. Any evidence on how stable this is, or whether "read-only" instance historically had less bugs?

ljedrz:
> Any evidence on how stable this is

It has been stable since at least 2021.

> whether "read-only" instance historically had less bugs?

Compared to a read-only primary instance? I don't think either is any more stable, and I'm not sure if a read-only primary instance can be used while a primary read/write session is in progress. Secondary instances are also better suited for continued use (via [`try_catch_up_with_primary`](https://docs.rs/rocksdb/latest/rocksdb/struct.DBCommon.html#method.try_catch_up_with_primary)).

ljedrz:
The stability question reminded me that we're actually a bit behind in terms of the version of RocksDB we're using. I pushed a commit updating it to 0.24; I've also confirmed that snarkOS builds with it.

ljedrz:
Hmm, not sure why the CI jobs are failing with a `clang`-related build error, it works for me locally, even without `CXXFLAGS = "-include cstdint"` (commented out in `.cargo/config.toml`).

In any case, an update isn't required for this feature; if I don't find a solution, I'll revert back to `0.21` when updating the `aleo-std` branch.

PR #3015, opened 2025-11-10
[Refactor] simplify shutting down the sequential ops thread

kaimast:
The changes in #2975 may cause panics if a `Ledger` object is dropped because the underlying channel cannot be used in an async context. 

This PR changes the shutdown logic to not rely on a special `Shutdown` command, and, instead, closes the channel and waits for the background thread to terminate.

### DIFF
--- a/synthesizer/src/vm/mod.rs
+++ b/synthesizer/src/vm/mod.rs
@@ -602,13 +602,18 @@ impl<N: Network, C: ConsensusStorage<N>> VM<N, C> {
 
 impl<N: Network, C: ConsensusStorage<N>> Drop for VM<N, C> {
     fn drop(&mut self) {
-        // This check isn't perfect, but in practice it was only needed in order
-        // for tests to not leak memory.
-        if Arc::strong_count(&self.sequential_ops_tx) == 2 {
-            // Send a shutdown signal.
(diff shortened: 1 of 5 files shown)


synthesizer/src/vm/mod.rs
```diff
@@ -602,13 +602,18 @@ impl<N: Network, C: ConsensusStorage<N>> VM<N, C> {
...
-            self.sequential_ops_tx.write().take();
+        // Check if this the final external reference to `VM`.
+        if Arc::strong_count(&self.sequential_ops_tx) == 1 {
+            // If the background thread exists, shut it down.
+            if let Some(thread) = self.sequential_ops_thread.lock().take() {
+                // First, close the channel.
+                self.sequential_ops_tx.write().take();
+                // Wait for the thread to terminate.
+                trace!("Waiting for sequential ops thread to terminate");
+                thread.join().expect("Sequential ops thread had an error");
```
ljedrz, line 613: Hmm, I think it's fair to panic under such conditions, as it would indicate that atomic speculation or block insertion had panicked, and we're already shutting down.

ljedrz (approved):
This is an improvement; my first implementation also involved a check for strong count equal to `1`, I think I just neglected to `join` the sequential thread.

vicsn (approved):
Great cleanup

PR #3019, opened 2025-11-12
[Feat] Restore PendingBlocks API

kaimast:
Now that #2975 and #2971 are merged, we can also re-enable the API to pre-process blocks.

There is a companion PR in snarkOS (https://github.com/ProvableHQ/snarkOS/pull/3983) to test these changes.

### DIFF
--- a/.circleci/semver-checks.sh
+++ b/.circleci/semver-checks.sh
@@ -2,7 +2,7 @@
 
 set -x
 
-BASELINE_REV=5eb528f # UPDATE ME ON NECESSARY BREAKING CHANGES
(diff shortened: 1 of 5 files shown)


kaimast:
Semver checks are still failing, but that is something we can debug separately.

kaimast:
@vicsn stresstests for this PR pass. Maybe we can do another run and then merge?

vicsn (approved):
Two succesful stress-test runs in a row, let's do this. There will still be more stress-test runs before the release as we merge new features.

PR #3020, opened 2025-11-12
Revert "Properly log rejection reasons and critical error in VM"

howardwu:
Reverts ProvableHQ/snarkVM#2970

These error messages are used to debug failures in snarkVM testing. They should not be removed.


vicsn:
Proposing https://github.com/ProvableHQ/snarkVM/pull/3021 instead

PR #2999, opened 2025-11-03
Improve RESTQuery error

vicsn:
## Motivation

To help debug a validator's issue.

PR #3025, opened 2025-11-14
Update CircleCI config to exclude target directory

meddle0x53:
## Motivation

The semver checks fail too often, so maybe caching target is the reason for that. Removing this to try to fix the problem.

## Test Plan

Check if the CI passes for the semver checks.

## Documentation

Removed target directory from cache save paths.
This may fix the semver check fails.

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -1063,7 +1063,10 @@ jobs:
           name: Check for unused dependencies
           no_output_timeout: 10m
           command: |
-            cargo install cargo-machete@0.7.0
+            set -euo pipefail
+            if ! command -v cargo-machete >/dev/null 2>&1; then
+              cargo install cargo-machete@0.7.0 --locked
+            fi
             cargo machete
       - clear_environment:
           cache_key: v4.2.0-rust-1.88.0-machete-cache
@@ -1088,7 +1091,8 @@ jobs:
     executor: rust-docker
     resource_class: << pipeline.parameters.twoxlarge >>
     steps:
-      - checkout
+      - checkout:
+          method: full
       - setup_environment:
           cache_key: v4.2.0-rust-1.88.0-cargo-semver-checks-cache
       - run:
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -163,7 +163,7 @@ commands:
           paths:
             - /home/circleci/.cache/sccache
             - /home/circleci/.cargo/registry
-            - /home/circleci/.cargo/git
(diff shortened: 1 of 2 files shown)


vicsn:
Seems like the semver test still fails with "error: An object with id ffb3b2c89734c2c29ee87fd4db880fe3b46789dc could not be found".

I don't see the error message reported anywhere on github: https://github.com/obi1kenobi/cargo-semver-checks/issues

Can you try renaming the cache key so it clears:           cache_key: v4.2.0-rust-1.88.0-cargo-semver-checks-cache

kaimast:
Maybe this helps: the object is `.circleci/config.yml`. It is referenced in recent commits, including some on `staging`. My initial guess that it was due to some deleted branch, but that does not seem to be the case.

.circleci/config.yml
```diff
@@ -1088,7 +1091,12 @@ jobs:
     executor: rust-docker
     resource_class: << pipeline.parameters.twoxlarge >>
     steps:
-      - checkout
+      - checkout:
+          method: full
+      - run:
+          name: Drop cached git checkouts (semver)
+          command: |
+            rm -rf /home/circleci/.cargo/git
```
meddle0x53: The combination of the full checkout and this fixes the missing revision.

meddle0x53: I can try returning the cache of the git folder here to see if it will work with the full checkout only.

vicsn: sounds good, yes let's make the diff as small as possible

.circleci/config.yml
```diff
@@ -163,7 +163,7 @@ commands:
           paths:
             - /home/circleci/.cache/sccache
             - /home/circleci/.cargo/registry
-            - /home/circleci/.cargo/git
+            #- /home/circleci/.cargo/git
```
kaimast: Should we add a comment here why this is not cached?

meddle0x53: I think just removing it once fixes the revision problem, so will return it to see how it goes.

meddle0x53:
OK, so there were 2 problems:
1. The script was not checking out everything so we were getting the revision error. I fixed it by making sure the repo is fully cloned for this check.
2. Cargo machete was already there, because of the cache so I did a change to only install it if it is not there.

kaimast (approved):
LGTM!

Maybe we can squash the commits? 13 commits for an 6 line change is quite a lot.

kaimast:
I cherry-picked the most recent commit onto #2999 and will merge that one, once its tests have passed.

PR #3027, opened 2025-11-14
[Feature] Add Aleo generator access to programs.

raychu86:
## Motivation

This PR adds access to the Aleo generator `G` to program functions and finalize via a new operand, `aleo::GENERATOR`. 

This will allow us to perform `g_scalar_multiply` in programs, which enables various capabilities like address derivation.

### Example Usage:

The following program tests view key to address derivation in function and finalize. You can index the generator by passing in a `u32` index value (`aleo::GENERATOR[0u32]`). Or read the full generator object as a `[group; 251u32]` by using `aleo::GENERATOR` directly.
```
    function foo:
        input r0 as scalar.public;
        input r1 as address.public;
        mul aleo::GENERATOR[0u32] r0 into r2;
        cast r2 into r3 as address;
        assert.eq r1 r3;
        async foo r0 r1 into r4;
        output r4 as {program_id}/foo.future;
    finalize foo:
        input r0 as scalar.public;
        input r1 as address.public;
        mul aleo::GENERATOR[0u32] r0 into r2;
        cast r2 into r3 as address;
        assert.eq r1 r3;
```


synthesizer/program/src/lib.rs
```diff
@@ -983,6 +983,22 @@ impl<N: Network> ProgramCore<N> {
...
+    /// This includes `Operand::Generator`.
+    /// This is enforced to be `false` for programs before `ConsensusVersion::V13`.
+    #[inline]
+    pub fn contains_v13_syntax(&self) -> bool {
+        // Check each instruction and output in each function's finalize scope for the use of
+        // `Operand::Generator`.
+        cfg_iter!(self.functions()).any(|(_, function)| {
+            function.finalize_logic().is_some_and(|finalize_logic| {
+                cfg_iter!(finalize_logic.commands()).any(|command| {
+                    cfg_iter!(command.operands()).any(|operand| matches!(operand, Operand::Generator(_)))
```
vicsn: Nested parallel iteration may degrade performance

synthesizer/program/src/lib.rs
```diff
@@ -983,6 +983,22 @@ impl<N: Network> ProgramCore<N> {
         })
     }
 
+    /// Returns `true` if a program contains any V13 syntax.
+    /// This includes `Operand::Generator`.
+    /// This is enforced to be `false` for programs before `ConsensusVersion::V13`.
+    #[inline]
+    pub fn contains_v13_syntax(&self) -> bool {
+        // Check each instruction and output in each function's finalize scope for the use of
```
vicsn: Also function scope instructions. I'm aware of your other recent issue that we may want to improve on this approach in general.

synthesizer/program/src/logic/instruction/operand/parse.rs
```diff
@@ -29,6 +29,9 @@ impl<N: Network> Parser for Operand<N> {
             map(tag("block.height"), |_| Self::BlockHeight),
             map(tag("block.timestamp"), |_| Self::BlockTimestamp),
             map(tag("network.id"), |_| Self::NetworkID),
+            map(pair(tag("aleo::GENERATOR"), opt(delimited(tag("["), U32::<N>::parse, tag("]")))), |(_, index)| {
```
vicsn: I see we also have Group::gen already, is that the same generator as aleo::GENERATOR[0]? 

Do you have a specific usecase in mind for the generators in other indices?

synthesizer/src/vm/verify.rs
```diff
@@ -1645,6 +1653,156 @@ function compute:
         assert!(vm.block_store().get_confirmed_transaction(&valid_tx_id).unwrap().unwrap().is_accepted());
         assert!(vm.block_store().get_confirmed_transaction(&invalid_tx_id).unwrap().unwrap().is_rejected());
     }
+
+    #[cfg(feature = "test")]
+    #[test]
+    fn test_aleo_generators_migration() {
```
vicsn: May be more suitable in a test_v13.rs at this point.

Noting that I still hope to make time one day to figure out a better test setup, but it works at the moment.

vicsn (changes requested):

synthesizer/program/src/logic/instruction/operand/parse.rs
raychu86: They are not the same generator. 

The full access to the generator is a vector of groups. You can however just use the first index for `g_scalar_multiply`

synthesizer/program/src/lib.rs
raychu86: Good catch, addressed in [dcf3e86](https://github.com/ProvableHQ/snarkVM/pull/3027/commits/dcf3e867e8621faeab3f8d7245ae7681cb6491f8)

synthesizer/program/src/lib.rs
raychu86: I kept it consistent with the checks from previous consensus versions. 

In the future we can update this code to perform a singular scan of these objects rather than doing it once per consensus version change.

synthesizer/src/vm/verify.rs
raychu86: Done in [fd07c05](https://github.com/ProvableHQ/snarkVM/pull/3027/commits/fd07c05d726cf8e6fef0f30119909cc60b8c68a8)

synthesizer/program/src/logic/instruction/operand/parse.rs
vicsn: My only concern would be to rename `aleo::GENERATOR` to e.g. `aleo_group::GENERATOR`. Or to mention the specific curve... Happy to approve after you can confirmation on that (whether buy-in or push-back)

Antonio95:
I recently ran into this issue (precisely when writing a test that involves doing sth similar to key derivation inside an Aleo-instructions program). My current fix has been to get `CurrentNetwork::g_powers()[0]` in Rust and `format!` it into the Aleo-instruction-program's `str`ing.

I find it confusing that there's a point referred to as a "generator" in [the documentation](https://developer.aleo.org/guides/aleo/language/#group-elements) and obtainable through the literal `group::GEN` which is different from the generator `G = CurrentNetwork::g_powers()[0]` used throughout the protocol (e. g. via `vk * G = address`). To be clear, every non-zero point of the group those two points live in is also a generator. Does this other point `group::GEN` have any particular role in the protocol? (if not, why was it added as special literal? Perhaps `G =  CurrentNetwork::g_powers()[0]` was generated by multiplying the fixed, untrusted `group::GEN` by a random scalar?).

In any case, some better naming may be useful in distinguishing it from the actual, more protocol-relevant generator `CurrentNetwork::g_powers()[0]` to prevent errors of the flavour of `vk * group::GEN == address`. Also, perhaps the referenced paragraph of the Aleo Instructions documentation could clarify the distinction, although that update might make more sense once this PR is merged and Aleo programmers really have that interface to `CurrentNetwork::g_powers()[0]`.

synthesizer/program/src/logic/instruction/operand/parse.rs
raychu86: I kept it as `aleo` rather than `aleo_group` because we already have a `group::GEN` to avoid further confusing (it is already confusing). 

I also don't really want to mention the specific curve, because it may be too technical for users that want to do address operations related to Aleo without knowing the underlying curves.

raychu86: Open to other names tho, I don't particularly like how confusing it is already with the two different generators.

raychu86:
> I find it confusing that there's a point referred to as a "generator" in [the documentation](https://developer.aleo.org/guides/aleo/language/#group-elements) and obtainable through the literal `group::GEN` which is different from the generator `G = CurrentNetwork::g_powers()[0]` used throughout the protocol (e. g. via `vk * G = address`). To be clear, every non-zero point of the group those two points live in is also a generator. Does this other point `group::GEN` have any particular role in the protocol? (if not, why was it added as special literal? Perhaps `G = CurrentNetwork::g_powers()[0]` was generated by multiplying the fixed, untrusted `group::GEN` by a random scalar?).

@Antonio95 100% agree, this is quite confusing. Here is the breakdown/difference:

- `group::GEN` is the generator of the prime order subgroup.
    - I don't have an Aleo program use-case for this off the top of my head, but we do use this as one of the defaults during specific casting operations.

- `g_powers` is the "The group bases for the Aleo signature and encryption schemes" that is generated by calling `hash_to_curve` on the following string "AleoAccountEncryptionAndSignatureScheme0". 
    - This in particular is used specifically for Aleo accounts and signatures.
    - I opted for `aleo::GENERATOR` for some level of deviation, as explained above in the thread with Victor.

Antonio95:
@raychu86  I see, that does make sense. No complaints on my side, then. Let's just choose names that avoid confusion. 

I don't want to make the discussion even more complex (and it isn't my call to make), but this is what I'm thinking atm. `G_POWERS` could be good in that it matches some code on the Rust side (eg. `g_scalar_multiply`). However, having `aleo::G` and `aleo::GEN` might be confusing. Perhaps we can refer to it as the "base point", eg. in sth like `aleo::BASE_POINT_POWERS[...]`?

Whatever nomenclature you settle for in the end, I think it would be good to give the 0-th element of the array its own identifier (e. g. `aleo::BASE_POINT` for `aleo::BASE_POINT_POWERS[0u32]` in the above nomenclature), since it has a distinguished role and is probably the one developers will be accessing most often.

vicsn:
I like `group::BASE_POINT` and `group::BASE_POINT_POWERS`. @raychu86 can you make a spreadsheet to vote internally?

vicsn (approved):

vicsn:
@raychu86 a few tests are failing. Some spurious which we can ignore for now, but one expects a different generator name and synthesizer-integration may need to be ran with REWRITE_EXPECTATIONS due to the new consensus version.

Issue #3026, opened 2025-11-14
[Bug] Production startup time takes multiple minutes


vicsn:
## 🐛 Bug Report

Recent benchmarks show that startup time for production ledgers with thousands of deployments can take multiple minutes on solid `m7i.4xlarge` machines. Specific data will be added to this issue by @meddle0x53 

Loading the ledger happens via `Ledger::load` and then `VM::from`. Given the small surface area and heavy dependence on I/O, a few simple `timer!` and `lap!` additions should be sufficient to learn what is taking so much time. Likely contenders are:
- Serial calling of `load_deployment`, which creates `Stack`s - the VM's abstraction around deployed programs. If `Stack` loading is a bottleneck, this will mean that lazy loading them (as proposed by https://github.com/ProvableHQ/snarkVM/pull/2591 ) will create a large performance issue. 
  - Perhaps we can load deployments without any imports in parallel
  - Perhaps we can extract parallelizable components from `load_deployment`
- Initializing the `ConsensusStore`

PR #3030, opened 2025-11-18
[Perf] Facilitate the caching of the block tree

ljedrz:
This PR targets https://github.com/ProvableHQ/snarkVM/issues/3026.

I used `pprof` to analyze the loading of a full mainnet ledger, and the resulting flamegraph:

flamegraph

indicated that nearly 96% of the associated work is the creation of the block tree. I double-checked it, and indeed, around 90% of the loading time was spent on that operation.

These changes facilitate the caching of the block tree alongside the ledger, and automatically loading it on startup. A snarkOS counterpart PR will follow, where one of the methods introduced here will be automatically called on shutdown.

While I don't have the full mainnet ledger locally, I have verified that the one I do have loads more quickly when these changes are introduced.


ljedrz:
I ran the numbers on a full maninnet ledger after these changes, and it's a clear improvement:
```
before: 239.75s
after:   27.82s (down 88%)
```
Dumping the serialized block tree took 2.48s, and the resulting file was 965MiB.

For future reference, I'm also attaching the updated flamegraph of a ledger load (w/o `rayon` for increased clarity):

flamegraph_new

I'll analyze it to see if there's room for more improvement, but these would be follow-ups unrelated to the changes from this PR.

ledger/store/src/helpers/rocksdb/block.rs
```diff
@@ -219,9 +223,44 @@ impl<N: Network> BlockStorage<N> for BlockDB<N> {
...
+            // Construct the block tree.
+            N::merkle_tree_bhp(&hashes)
+        }
+
+        let mut path = aleo_ledger_dir(N::ID, self.storage_mode());
+        path.push("block_tree");
+
+        if let Ok(serialized_tree) = fs::read(&path) {
+            // Deserialize a ready block tree.
+            let ret = bincode::deserialize(&serialized_tree).or_else(|e| {
```
kaimast, line 256: Should there be additional checks here to ensure the block tree is up-to-date? I think you can just compare the most recent block hashes.

ledger/store/src/block/mod.rs
```diff
@@ -1200,10 +1195,23 @@ impl<N: Network, B: BlockStorage<N>> BlockStore<N, B> {
...
     }
+
+    /// Serializes and persists the current block tree.
+    #[cfg(feature = "rocks")]
+    pub fn cache_block_tree(&self) -> Result<()> {
+        let mut path = aleo_ledger_dir(N::ID, self.storage.storage_mode());
+        path.push("block_tree");
+
+        let serialized_tree = bincode::serialize(&&*self.tree.read())?;
+        fs::write(path, &serialized_tree)?;
```
kaimast: nit: it is slightly more efficient to do

```rust
let file = File::create(path)?;
bincode::serialize_into(file, &&*self.tree.read())?;
```

ledger/store/src/helpers/rocksdb/block.rs
ljedrz: note L259: the file is removed as soon as it's loaded, in order to ensure that it won't be reused, and the new one is only created on shutdown (on snarkOS side); it's a valid concern if we ever want to cache the block tree outside of shutdown conditions, but I don't think we do

kaimast:
Looks great to me! I left one comment about potentially loading an outdated cache, which I believe needs to addressed.

Additionally, I wonder if we could store the block tree on `Ledger::drop` and not have snarkOS counterpart. That would be cleaner and ensure it is only invoked during shutdown. snarkOS shutdown is a little messy as of now, though, so that might be not possible yet  (hence https://github.com/ProvableHQ/snarkOS/pull/3915).

ledger/store/src/block/mod.rs
ljedrz: hmm yeah, that's probably a bit faster; might need an explicit call to `File::sync_data` too, I'll check it out

vicsn:
PR looks great, will approve after Kai's comments are addressed

ljedrz:
> I wonder if we could store the block tree on Ledger::drop and not have snarkOS counterpart.

Yeah, it did work just fine, and I guess there's no reason not to have it happen implicitly; I'll remove the snarkOS-side PR if this change is accepted.

In other words, review comments addressed.

ljedrz:
After some additional tests I noticed that the `Ledger` is not internally `Arc`ed, so the `Drop` impl needs to be adjusted; I should have it ready soon.

ljedrz:
This is the right direction, but I'm not always seeing the `Drop` being triggered; I need to run some more snarkOS tests.

ljedrz:
@kaimast ok, I have a temporary shutdown solution for snarkOS to ensure that the `Drop` gets triggered. I'll propose it, and it can later be superseded by your changes.

Edit: filed as https://github.com/ProvableHQ/snarkOS/pull/4021.

vicsn (approved):
LGTM, but let's await Kai

ledger/store/src/helpers/rocksdb/block.rs
```diff
@@ -219,9 +223,44 @@ impl<N: Network> BlockStorage<N> for BlockDB<N> {
...
+            // Prepare an iterator over the block heights and prepare the leaves of the block tree.
+            let hashes = storage
+                .id_map()
+                .iter_confirmed()
+                .sorted_unstable_by(|(h1, _), (h2, _)| h1.cmp(h2))
+                .map(|(_, hash)| hash.to_bits_le())
+                .collect::<Vec<Vec<bool>>>();
+
+            // Construct the block tree.
+            N::merkle_tree_bhp(&hashes)
```
vicsn, line 246: Can a problem exist when the succesfully deserialized cached tree doesn't match or is smaller/bigger than the state of the node's ledger

ljedrz: this is part of an inner function which is used only if the cached tree is not available

ledger/store/src/block/mod.rs
kaimast: It should sync the file on drop. But performance doesn't really matter, so either is fine.

ledger/store/src/helpers/rocksdb/block.rs
kaimast: My worry was about clients moving ledger snapshots around and restoring an old file. Another, very unlikely, case is that the file gets corrupted but deserializes fine. 
Adding another sanity check, if not too expensive, could save us some headaches in the future.

ljedrz: sure, that's a fair concern; I'm not too savvy with the tree myself, do you have a suggestion how to easily sanity-check it? or @vicsn?

ledger/store/src/helpers/rocksdb/block.rs
vicsn: Continuing here: https://github.com/ProvableHQ/snarkVM/pull/3030#discussion_r2547087823

ledger/store/src/helpers/rocksdb/block.rs
vicsn: The merkle tree's leaves are the block hashes.

Inside of `struct MerkleTree`, we still store hashes of those leaves in:
```
    /// The internal hashes, from root to hashed leaves, of the full Merkle tree.
    tree: Vec<PH::Hash>,
```

Looking at the API which `MerkleTree` exposes, I see 3 options. We should pick one which balances simplicity and speed:
- use `pub const fn number_of_leaves` to see if it matches the number of blocks (not sure if this requires iterating over the entire db index). 
- use `pub fn prove` the final leave see if it returns without error (not very idiomatic)
- use `pub fn leaf_hashes` to match the last block hash (though note you have to manually hash the leave to see if it matches)

ljedrz:
I've introduced a very cheap (basically free) coherence check in `Ledger::load_unchecked`; as a bonus, we now call `BlockStore::max_height` - which traverses the entire reverse state root map (meaning it is somewhat expensive, even though it is optimized) - only once.

ledger/src/lib.rs
```diff
@@ -226,16 +226,25 @@ impl<N: Network, C: ConsensusStorage<N>> Ledger<N, C> {
...
-        }
+            0
+        };
         lap!(timer, "Initialize genesis");
 
-        // Retrieve the latest height.
-        let latest_height =
-            ledger.vm.block_store().max_height().with_context(|| "Failed to load blocks from the ledger")?;
+        // Ensure that the greatest stored height matches that of the block tree.
+        ensure!(
```
vicsn, line 248: Brilliant find - can you confirm that you manually tested that we get the expected error message? A simple way could be to run a network, remove a node's ledger but nót their block cache, and then to start the node again

ljedrz: I did test locally that it works, though I haven't swapped the cached tree to check the error; that being said, I initially used the value of `1` instead of `0` for the genesis scenario, and the node did fail to load (but I don't remember if the message was there). I can double-check with a swapped ledger as soon as I can.

ljedrz:
I've tested both possible cases:
1. when the block tree is above the storage, `BlockStore::open` now fails with `Block {height} exists in tree but not in storage; perhaps you used a wrong block tree cache file?`
2. when the block tree is below the storage, `Ledger::load_unchecked` now fails with `The stored height is different than the one in the block tree; please ensure that the cached block tree is valid or delete it`

vicsn (approved):

ledger/store/src/block/mod.rs
```diff
@@ -1009,17 +1015,7 @@ impl<N: Network, B: BlockStorage<N>> BlockStore<N, B> {
...
-            .id_map()
-            .iter_confirmed()
-            .sorted_unstable_by(|(h1, _), (h2, _)| h1.cmp(h2))
-            .map(|(_, hash)| hash.to_bits_le())
-            .collect::<Vec<Vec<bool>>>();
-
-        // Construct the block tree.
-        debug!("Found {} blocks on disk", hashes.len());
-        let tree = N::merkle_tree_bhp(&hashes)?;
+        let tree = storage.create_block_tree()?;
```
raychu86, line 1017: Can you add subsequent checks here to ensure that the root of the block tree matches the proper state root in the `state_root_map` for the expected block height that the tree is supposed to represent? 

If this fails, we would need to reconstruct the tree from scratch.

ledger/store/src/block/mod.rs
```diff
@@ -1200,10 +1198,35 @@ impl<N: Network, B: BlockStorage<N>> BlockStore<N, B> {
...
 
+    /// Stores a database backup at the given location.
     #[cfg(feature = "rocks")]
     pub fn backup_database<P: AsRef<std::path::Path>>(&self, path: P) -> Result<(), String> {
         self.storage.backup_database(path)
     }
+
+    /// Serializes and persists the current block tree.
+    #[cfg(feature = "rocks")]
+    pub fn cache_block_tree(&self) -> Result<()> {
```
raychu86, line 1208: Is there a lock we can add to enforce that this operation is properly performed and completed at shutdown?

raychu86 (changes requested):

ledger/store/src/block/mod.rs
ljedrz: This is executed as part of the `Drop` impl for the `Ledger`, which is a synchronous operation and always fully concludes on a clean shutdown, and in snarkOS, even in case of a panic (as it uses the default `panic = "unwind"` mode); using a lock isn't necessary, as the aforementioned impl has full exclusive ownership of the `Ledger` (via `&mut self`).

raychu86: Is this documented somewhere in the code? It'd be nice to indicate the atomic properties of the Drop impl

kaimast: Isn't the block store also dropped when dropping the ledger? Maybe we could not expose this function at all, and just have it happen on drop?

ljedrz: @raychu86 I extended the docs for the `Drop` impl.
@kaimast I think there is utility in having it as a method on `BlockStore` too, in case we want to use it for some test run or if we ever need to extend the `Drop` impl for the `Ledger` (which is the right impl to use, as all the stores only ever exist as part of the `Ledger`, so it's a clear central point for cleanups).

ledger/src/lib.rs
kaimast: nit: maybe print the filename here as well, so it is easy to locate

kaimast (approved):
LGTM! I think @raychu86 comments are all addressed as well.

ledger/store/src/block/mod.rs
```diff
@@ -46,15 +46,19 @@ use snarkvm_ledger_block::{
 use snarkvm_ledger_narwhal_batch_certificate::BatchCertificate;
 use snarkvm_ledger_puzzle::{Solution, SolutionID};
 use snarkvm_synthesizer_program::{FinalizeOperation, Program};
+use std::io::BufWriter;
```
raychu86: nit: move import down

raychu86 (approved):

vicsn:
@ljedrz can you merge in origin/staging?

ljedrz:
Merged `staging` and addressed the final suggestions.

vicsn (approved):

PR #3033, opened 2025-11-19
Update CircleCI config to run the semver workflow

meddle0x53:
## Motivation

The semver checks fail too often, so maybe caching target is the reason for that. Removing this to try to fix the problem.

## Test Plan

Check if the CI passes for the semver checks.

## Documentation

Removed target directory from cache save paths.
This may fix the semver check fails.

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -1063,7 +1063,10 @@ jobs:
           name: Check for unused dependencies
           no_output_timeout: 10m
           command: |
-            cargo install cargo-machete@0.7.0
+            set -euo pipefail
+            if ! command -v cargo-machete >/dev/null 2>&1; then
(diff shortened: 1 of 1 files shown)


vicsn:
@meddle0x53 @kaimast the target branch for this PR was not `staging`. Can you please double check if we need to re-open something on `staging`? Thx!

meddle0x53:
@vicsn Actually it is in staging : https://github.com/ProvableHQ/snarkVM/commit/063de51772750c557bff3668be493c84b34da76d

I made two PRs and one of them went to staging. Because I did it in two different ways it confused me too, but the fact is that is in the staging branch : https://github.com/ProvableHQ/snarkVM/blob/staging/.circleci/config.yml#L1095

PR #3034, opened 2025-11-20
[Perf] More unchecked deserialization from the DB

ljedrz:
This PR is a follow-up to https://github.com/ProvableHQ/snarkVM/pull/3030 building on its findings, but can be handled individually.

These changes are very similar to those from https://github.com/ProvableHQ/snarkVM/pull/2789, and may improve block retrieval performance even further.

Below are the figures for the loading of a modestly-sized (413'697 blocks) local ledger:
```
before: 2.90s
after:  2.55s (down 12%)
```

I'd expect the impact of these changes to be similar with larger ledgers.


kaimast (approved):
Looks good!

PR #3036, opened 2025-11-20
[Fix] Do not automatically append API version

kaimast:
We recently added backwards compatibility to REST API v1, which is also still the default. This enables us to not force "v2" anymore by appending it to the base URL of a query.

This pull request
* Removes the code that detects an API version in the base URL and appends it ([8b13264](https://github.com/ProvableHQ/snarkVM/pull/3036/commits/8b13264de6bf97c00103973245c8b54f91be75e9))
* Uses a constant for detecting the content type of response instead of a hard-coded string ([29e222d](https://github.com/ProvableHQ/snarkVM/pull/3036/commits/29e222d22ad89b7b4e116c4c50ad22da9f8bd011))

### DIFF
--- a/ledger/query/src/query/rest.rs
+++ b/ledger/query/src/query/rest.rs
@@ -226,16 +226,17 @@ impl<N: Network> RestQuery<N> {
         if response.status().is_success() {
             response.body_mut().read_json().with_context(|| format!("Failed to parse JSON response from {endpoint}"))
         } else {
-            let content_type = response
+            // v2 will return the error in JSON format.
+            let is_v2 = response
                 .headers()
-                .get("Content-Type")
-                .ok_or_else(|| anyhow!("Endpoint return error without ContentType"))?
-                .to_str()
-                .with_context(|| "Endpoint returned invalid ContentType")?;
+                .get(http::header::CONTENT_TYPE)
+                .and_then(|ct| ct.to_str().ok())
+                .map(|ct| ct.contains("json"))
+                .unwrap_or(false);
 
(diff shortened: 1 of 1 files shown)


ledger/query/src/query/rest.rs
```diff
@@ -226,16 +213,17 @@ impl<N: Network> RestQuery<N> {
         if response.status().is_success() {
             response.body_mut().read_json().with_context(|| format!("Failed to parse JSON response from {endpoint}"))
         } else {
-            let content_type = response
+            // v2 will return the error in JSON format.
+            let is_v2 = response
```
vicsn: Can we name the variable agnostic to any supposed version? Just switch based on content-type

ledger/query/src/query/rest.rs
```diff
@@ -263,12 +251,26 @@ impl<N: Network> RestQuery<N> {
...
             response.json().await.with_context(|| format!("Failed to parse JSON response from {endpoint}"))
         } else {
-            // Convert returned error into an `anyhow::Error`.
-            let error: RestError = response
-                .json()
-                .await
-                .with_context(|| format!("Failed to parse JSON error response from {endpoint}"))?;
-            Err(error.parse().context(format!("Failed to fetch from {endpoint}")))
+            // v2 will return the error in JSON format.
+            let is_v2 = response
```
vicsn: Same

vicsn (changes requested):

vicsn (approved):

PR #3041, opened 2025-11-26
Fixed `mint_and_split.aleo` test

Antonio95:
As far as I can tell, this is the only test in `test_vm_execute_and_finalize` which involves inclusion proofs (it consumes a record). One reason this is relevant is testing fee estimation, which involves proof sizes and therefore should also be tested on batch proofs that include inclusion proofs.

The test case which consumes a record was added in https://github.com/ProvableHQ/snarkVM/commit/d394f44de5a7ff23091867ea0b7308959c9f2a04 and was broken soon after. Here broken means that the transaction consuming the record now failed, but since expectations were regenerated (here: https://github.com/ProvableHQ/snarkVM/commit/fd6191d072adbeeb89b076645be4f3897b771d87), the test still passes.

In this commit, I updated the record to actually be consumed again, as well as the expectation.

For the future, we can think of a way to ensure this transaction actually consumes a record (which doesn't get overwritten by regenerated expectations; so perhaps hardcoded at the test-function level); or let it be and assume future updates may mean there are no tests involving inclusion proofs in that file.

### DIFF
--- a/synthesizer/tests/tests/vm/execute_and_finalize/mint_and_split.aleo
+++ b/synthesizer/tests/tests/vm/execute_and_finalize/mint_and_split.aleo
@@ -17,7 +17,7 @@ cases:
     private_key: APrivateKey1zkpJhviKDvvm7yu7SZuhSudVR7zjCRG2HznuAHwuGYc1xqN
   - program: mint_and_split.aleo
     function: split
-    inputs: [ "{ owner: aleo1j2hfs6yru47h2nvsjdefwtw6nwaj0y4zcl02juyy29txm7nt6y9qln7uhp.private, microcredits: 100u64.private, _nonce: 4745749250363947745807119124444606639539806979883854068889616498365610280135group.public, _version: 1u8.public }", 50u64]
+    inputs: [ "{ owner: aleo1j2hfs6yru47h2nvsjdefwtw6nwaj0y4zcl02juyy29txm7nt6y9qln7uhp.private, microcredits: 100u64.private, _nonce: 8328661033496413191179695268319516441428135407695959353783510885778770196235group.public, _version: 1u8.public }", 50u64]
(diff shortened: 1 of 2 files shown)


synthesizer/tests/tests/vm/execute_and_finalize/mint_and_split.aleo
```diff
@@ -17,7 +17,7 @@ cases:
     private_key: APrivateKey1zkpJhviKDvvm7yu7SZuhSudVR7zjCRG2HznuAHwuGYc1xqN
   - program: mint_and_split.aleo
     function: split
-    inputs: [ "{ owner: aleo1j2hfs6yru47h2nvsjdefwtw6nwaj0y4zcl02juyy29txm7nt6y9qln7uhp.private, microcredits: 100u64.private, _nonce: 4745749250363947745807119124444606639539806979883854068889616498365610280135group.public, _version: 1u8.public }", 50u64]
+    inputs: [ "{ owner: aleo1j2hfs6yru47h2nvsjdefwtw6nwaj0y4zcl02juyy29txm7nt6y9qln7uhp.private, microcredits: 100u64.private, _nonce: 8328661033496413191179695268319516441428135407695959353783510885778770196235group.public, _version: 1u8.public }", 50u64]
```
vicsn, line 20: So I guess the previous value became invalid because some internal RNG state of the test changed? Which will in all likelihood happen again soon.

As such, this PR has no value, it'll be broken again soon. We should probably move the testing of size and fee estimation with inclusion and translation to `test_v12.rs`.

Antonio95: Okay, added to my TODO list 👌

Antonio95:
Moved to a (still to be merged) separate test, as suggested: https://github.com/ProvableHQ/snarkVM/blob/3a67e6a1f2091500d2a0a4778bfa0c51495ae1fb/synthesizer/src/vm/tests/test_v12/mixed.rs#L25

PR #3042, opened 2025-11-26
[Feat] Limit the total supply of Aleo to 5B

raychu86:
## Motivation

This PR limits the total supply of Aleo that can be created to 5B credits. This is done by calculated the expected upper-bound block height based on the emission algorithms and bringing the block and coinbase rewards to 0 after that height.

Based on the current timelines, this limit height will be hit at block `263_527_685`, which is ~25 years from network launch. Because the changes will not occur for quite some time, there is flexibility for adjusting this limit height to be more accurate if there are major changes to network assumptions in the future.

Link to ARC - https://github.com/ProvableHQ/ARCs/discussions/101

**Do Not Merge until ARC is voted on and approved.**

### DIFF
--- a/console/network/src/lib.rs
+++ b/console/network/src/lib.rs
@@ -133,7 +133,12 @@ pub trait Network:
     /// This value represents the absolute upper bound on all ALEO created over the lifetime of the network.
     const MAX_SUPPLY: u64 = 5_000_000_000_000_000; // 5B credits
     /// The block height that upper bounds the total supply of Aleo credits to 5 billion.
+    #[cfg(not(any(test, feature = "test")))]
     const MAX_SUPPLY_LIMIT_HEIGHT: u32 = 263_527_685;
+    /// The block height that upper bounds the total supply of Aleo credits to 5 billion.
+    /// This is deliberately set to a low value (8) for testing purposes only.
+    #[cfg(any(test, feature = "test"))]
+    const MAX_SUPPLY_LIMIT_HEIGHT: u32 = 5;
     /// The cost in microcredits per byte for the deployment transaction.
     const DEPLOYMENT_FEE_MULTIPLIER: u64 = 1_000; // 1 millicredit per byte
     /// The multiplier in microcredits for each command in the constructor.
--- a/console/network/src/lib.rs
+++ b/console/network/src/lib.rs
@@ -133,11 +133,11 @@ pub trait Network:
     /// This value represents the absolute upper bound on all ALEO created over the lifetime of the network.
     const MAX_SUPPLY: u64 = 5_000_000_000_000_000; // 5B credits
     /// The block height that upper bounds the total supply of Aleo credits to 5 billion.
-    #[cfg(not(any(test, feature = "test")))]
+    #[cfg(not(feature = "test"))]
     const MAX_SUPPLY_LIMIT_HEIGHT: u32 = 263_527_685;
     /// The block height that upper bounds the total supply of Aleo credits to 5 billion.
     /// This is deliberately set to a low value for testing purposes only.
-    #[cfg(any(test, feature = "test"))]
+    #[cfg(feature = "test")]
     const MAX_SUPPLY_LIMIT_HEIGHT: u32 = 5;
     /// The cost in microcredits per byte for the deployment transaction.
     const DEPLOYMENT_FEE_MULTIPLIER: u64 = 1_000; // 1 millicredit per byte
--- a/console/network/src/lib.rs
+++ b/console/network/src/lib.rs
@@ -136,7 +136,7 @@ pub trait Network:
     #[cfg(not(any(test, feature = "test")))]
     const MAX_SUPPLY_LIMIT_HEIGHT: u32 = 263_527_685;
     /// The block height that upper bounds the total supply of Aleo credits to 5 billion.
-    /// This is deliberately set to a low value (8) for testing purposes only.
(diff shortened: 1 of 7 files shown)


console/network/src/lib.rs
```diff
@@ -129,6 +129,16 @@ pub trait Network:
...
     /// The starting supply of Aleo credits.
     const STARTING_SUPPLY: u64 = 1_500_000_000_000_000; // 1.5B credits
+    /// The maximum supply of Aleo credits.
+    /// This value represents the absolute upper bound on all ALEO created over the lifetime of the network.
+    const MAX_SUPPLY: u64 = 5_000_000_000_000_000; // 5B credits
+    /// The block height that upper bounds the total supply of Aleo credits to 5 billion.
+    #[cfg(not(any(test, feature = "test")))]
+    const MAX_SUPPLY_LIMIT_HEIGHT: u32 = 263_527_685;
+    /// The block height that upper bounds the total supply of Aleo credits to 5 billion.
+    /// This is deliberately set to a low value (8) for testing purposes only.
```
vicsn:
```suggestion
    /// This is deliberately set to a low value for testing purposes only.
```

vicsn (approved):
LGTM, let's await Kai's review too

kaimast:
One of the ledger tests seems to fail. We probably should look into that before merging.

console/network/src/lib.rs
```diff
@@ -129,6 +129,16 @@ pub trait Network:
 
     /// The starting supply of Aleo credits.
     const STARTING_SUPPLY: u64 = 1_500_000_000_000_000; // 1.5B credits
+    /// The maximum supply of Aleo credits.
+    /// This value represents the absolute upper bound on all ALEO created over the lifetime of the network.
+    const MAX_SUPPLY: u64 = 5_000_000_000_000_000; // 5B credits
+    /// The block height that upper bounds the total supply of Aleo credits to 5 billion.
+    #[cfg(not(any(test, feature = "test")))]
```
kaimast: Should this be using the `test` feature or the `test_consensus_heights` feature? I am still hoping we can phase out the former, as its name is not very descriptive.

kaimast (approved):
Looks good!

console/network/src/lib.rs
raychu86: We want this to be for tests only rather than `test_consensus_heights`, because `test_consensus_heights` can be used for devnets.

I've kept the feature flag consistent with the other network variables, so if we would like to cleanup, we should do in a subsequent PR.

raychu86:
Fixed tests and simplified logic with [d3fd59b](https://github.com/ProvableHQ/snarkVM/pull/3042/commits/d3fd59b134b0517f043b64cb5b5364d3b077c38f).

ledger/src/tests.rs
```diff
@@ -2569,8 +2581,21 @@ fn test_no_rewards_after_limit_height() {
...
+    assert!(!next_block.ratifications().is_empty());
+    let ratifications: Vec<_> = next_block.ratifications().iter().collect();
+    match ratifications[0] {
+        Ratify::BlockReward(block_reward) => {
+            assert_eq!(*block_reward, 0);
+        }
+        _ => panic!("Expected a block reward ratification"),
+    }
+    match ratifications[1] {
+        Ratify::PuzzleReward(puzzle_reward) => {
```
kaimast: I assume the overhead of keeping block rewards of 0 is minimal compared to not containing them at all?

ledger/block/src/verify.rs
```diff
@@ -430,7 +413,7 @@ impl<N: Network> Block<N> {
         // Calculate the time since last block.
         let time_since_last_block = timestamp.saturating_sub(previous_block.timestamp());
         // Compute the expected block reward.
-        let expected_block_reward = block_reward::<N>(
+        let mut expected_block_reward = block_reward::<N>(
```
kaimast: Wouldn't it be better for this function to return 0, so that we don't run into problems if it is used in other places?

vicsn: I agree that would be *slightly* cleaner, but only marginally so. If there is a good reason for the new approach, e.g. because it doesn't break a bunch of existing tests and speeds up our development, I'm fine with this more hacky new approach.

vicsn: Maybe worth expanding the docblock above `fn block_reward` though

ledger/src/tests.rs
raychu86: yea exactly, our `post-ratify` logic currently expects a `Ratify::BlockReward` object (even if it is 0) to update the committee mapping.

ledger/block/src/verify.rs
raychu86: Agreed, done in [c908858](https://github.com/ProvableHQ/snarkVM/pull/3042/commits/c908858a8bd368abab1e9b6ee84d28f9b24dccca) and [c026cc7](https://github.com/ProvableHQ/snarkVM/pull/3042/commits/c026cc7bf4d42722cb136bfabb543821ad8de51b)

kaimast (approved):

vicsn (approved):

kaimast:
@raychu86 it looks like the tests fail again

kaimast (approved):

vicsn (approved):
Still LGTM, and this will give us a further early peace of mind: https://github.com/ProvableHQ/snarkOS/pull/4029

PR #3045, opened 2025-11-27
Allow external structs.

mohammadfawaz:
Specifically,

1. Introduce a PlaintextType::ExternalStruct, and allow referring to such a type in .aleo code by a Locator.

2. Allow casting to external struct types.

3. Let two struct types be considered the same type as long as they are structurally the same and have the same local names. This funky ruleset is effectively here for backwards compatibility. In more detail:

```
/// Equivalence of structs means they have the same local names (regardless of whether
/// they're local or external), and their members have the same names and equivalent
/// types in the same order, recursively.
///
/// Equivalence of arrays means they have the same length and their element types are
/// equivalent.
///
/// This definition of equivalence was chosen to balance these concerns:
///
/// 1. All programs from before the existence of external structs will continue to work -
/// thus it's necessary for a struct created from another program to be considered equivalent
/// to a local one with the same name and structure, as in practice that was the behavior.
/// 2. We don't want to allow a fork. Thus we do need to check names, not just structural
/// equivalence - otherwise we could get a program deployable to a node which is using
/// this check, but not deployable to a node running an earlier SnarkVM.
```

Also note this other usecase from a user:

```
currently we got in an issue where one mainnet deployed contract imports tokenRegistry, and the contract we are writing for A imports that contract and A contract. but, A contract and tokenRegistryContract have same struct names so we cannot compile that.If we cannot change struct name in A contract, we need to wait for snarkVM update, just to compile the contracts
```


mohammadfawaz:
Rebase and squash https://github.com/ProvableHQ/snarkVM/pull/2798 and opened a new PR with Michael as the co-author of the single commit. This will make it easier to keep rebasing as we review this.

mohammadfawaz:
@vicsn I added a test that uses an external struct inside a record: https://github.com/ProvableHQ/snarkVM/pull/3045/commits/4fb2cd37cbe094aa6bbc425163bf7e7e08c6bad6

synthesizer/src/vm/tests/test_v13.rs
```diff
@@ -0,0 +1,105 @@
...
+    let caller_private_key = crate::vm::test_helpers::sample_genesis_private_key(rng);
+
+    // Initialize the VM at the correct height.
+    let v10_height = CurrentNetwork::CONSENSUS_HEIGHT(consensus_version).unwrap();
+    let vm = crate::vm::test_helpers::sample_vm_at_height(v10_height, rng);
+
+    // Define the first program with a record.
+    let program_one = Program::from_str(
+        r"
+program test_one.aleo;
```
vicsn, line 62: Can you also test that we can now reference the struct of a program deployed **before** the new consensus version - the easiest will be to use e.g. the hardcoded `credits.aleo/committee_state`

synthesizer/program/src/traits/stack_and_registers.rs
```diff
@@ -150,6 +151,109 @@ pub trait StackTrait<N: Network> {
...
+/// 1. All programs from before the existence of external structs will continue to work;
+///    thus it's necessary for a struct created from another program to be considered equivalent
+///    to a local one with the same name and structure, as in practice that was the behavior.
+/// 2. We don't want to allow a fork. Thus we do need to check names, not just structural
+///    equivalence - otherwise we could get a program deployable to a node which is using
+///    this check, but not deployable to a node running an earlier SnarkVM.
+///
+/// The stacks are passed because struct types need to access their stack to get their
+/// structure.
+pub fn types_equivalent<N: Network>(
```
vicsn, line 191: ChatGPT pointed me to the useful idea that we can easily add a couple obvious positive and negative tests for this central function.

synthesizer/src/vm/tests/test_v13.rs
vicsn: The above idea was mine, but another ChatGPT idea: test having `[external struct; N]` somewhere.

synthesizer/program/src/traits/stack_and_registers.rs
mohammadfawaz: Added a series of unit tests in `synthesizer/program/tests/types/equivalence.rs`

synthesizer/src/vm/tests/test_v13.rs
mohammadfawaz: Added two more e2e tests that use the two patterns above.

mohammadfawaz:
Review comments addressed in https://github.com/ProvableHQ/snarkVM/pull/3045/commits/3e0e03cc52cefe5cef6eaf577863b13b55fbb9b2

vicsn:
Please no more force pushing after reviews have started

synthesizer/program/src/traits/stack_and_registers.rs
```diff
@@ -242,11 +242,16 @@ pub fn types_equivalent<N: Network>(
             struct_compare(&*external_stack, struct_type0, stack1, struct_type1)
         }
         (Struct(id), ExternalStruct(loc)) => {
+            dbg!("HERE!");
```
vicsn: Now we know how the sausage is made.

vicsn (approved):
LGTM, one test seems to fail. If you run with `REWRITE_EXPECTATIONS=1`, please still check the new output makes sense.

synthesizer/program/src/traits/stack_and_registers.rs
mohammadfawaz: OH no... isn't `HERE` universal though? 😆

console/program/src/data_types/plaintext_type/bytes.rs
```diff
@@ -23,7 +23,8 @@ impl<N: Network> FromBytes for PlaintextType<N> {
             0 => Ok(Self::Literal(LiteralType::read_le(&mut reader)?)),
             1 => Ok(Self::Struct(Identifier::read_le(&mut reader)?)),
             2 => Ok(Self::Array(ArrayType::read_le(&mut reader)?)),
-            3.. => Err(error(format!("Failed to deserialize annotation variant {variant}"))),
+            3 => Ok(Self::ExternalStruct(Locator::read_le(&mut reader)?)),
+            4.. => Err(error(format!("Failed to deserialize annotation variant {variant}"))),
```
d0cd: Exists in current implementation, but it might be more clear to say
```suggestion
            4.. => Err(error(format!("Failed to deserialize plaintext type variant {variant}"))),
```

console/program/src/data_types/plaintext_type/mod.rs
```diff
@@ -18,7 +18,7 @@ mod parse;
 mod serialize;
 mod size_in_bits;
 
-use crate::{ArrayType, Identifier, LiteralType, StructType};
+use crate::{ArrayType, Identifier, LiteralType, Locator, ProgramID, StructType};
 use snarkvm_console_network::prelude::*;
 
 /// A `PlaintextType` defines the type parameter for a literal, struct, or array.
```
d0cd:
```suggestion
/// A `PlaintextType` defines the type parameter for a literal, struct, array, or external struct.
```

console/program/src/data_types/plaintext_type/size_in_bits.rs
```diff
@@ -67,7 +74,32 @@ impl<N: Network> PlaintextType<N> {
...
+                    total = total.checked_add(member_size).ok_or(anyhow!("`size_in_bits` overflowed"))?;
+                }
+
+                Ok(total)
+            }
+            PlaintextType::ExternalStruct(locator) => {
+                // Look up the struct
+                let struct_ = get_external_struct(locator)?;
+
+                // Account for the plaintext variant bits.
```
d0cd: Could consider unifying with the struct logic in the variant above, for consistency. 
Ditto for `size_in_bits_raw`.

synthesizer/process/src/stack/mod.rs
```diff
@@ -332,6 +332,12 @@ impl<N: Network> Stack<N> {
             }
         }
 
+        for mapping in self.program.mappings().values() {
```
d0cd: This check is performed in `verify.rs` gated by a consensus version. Does it also need to happen here?

synthesizer/program/src/constructor/mod.rs
```diff
@@ -51,6 +51,16 @@ impl<N: Network> ConstructorCore<N> {
         &self.positions
     }
 
+    /// Returns whether this constructor refers to an external struct.
+    pub fn contains_external_struct(&self) -> bool {
+        self.commands.iter().any(|command| {
```
d0cd: Should we also check the other command variants?

synthesizer/program/src/finalize/mod.rs
```diff
@@ -80,6 +80,12 @@ impl<N: Network> FinalizeCore<N> {
         &self.positions
     }
 
+    pub fn contains_external_struct(&self) -> bool {
+        self.commands
+            .iter()
+            .any(|command| matches!(command, Command::Instruction(inst) if inst.contains_external_struct()))
```
d0cd, line 86: Should we also check the other command variants?

synthesizer/program/src/logic/instruction/mod.rs
```diff
@@ -589,6 +589,16 @@ impl<N: Network> Instruction<N> {
         instruction!(self, |instruction| instruction.output_types(stack, input_types))
     }
 
+    /// Returns whether this instruction refers to an external struct.
+    pub fn contains_external_struct(&self) -> bool {
+        // Only cast instructions may contain an explicit reference to an external struct.
+        // Calls may produce them, but they don't explicitly reference the type, and that's
+        // always been allowed.
+        matches!(self,
```
d0cd: When a new instruction is added, we may forget to update this code.
For example, dynamic dispatch would in theory support `get.dynamic.record r0.microcredits into r1 as foo.aleo/data`, but it's easy for us to forget to update this rule.
This also holds true for some of the other checks that match on the instructions?
Is there a more reliable way we can enforce these properties?

synthesizer/program/src/logic/instruction/operation/deserialize.rs
```diff
@@ -535,6 +546,7 @@ where
             // Cache the plaintext bits, and return the struct.
             Ok(circuit::Plaintext::Struct(members, Default::default()))
         }
+        PlaintextType::ExternalStruct(_identifier) => todo!(),
```
d0cd: Bailing with an error message is fine if this is not supported.

synthesizer/program/src/logic/instruction/operation/deserialize.rs
```diff
@@ -349,6 +355,7 @@ where
             // Cache the plaintext bits, and return the struct.
             Ok(Plaintext::Struct(members, Default::default()))
         }
+        PlaintextType::ExternalStruct(_identifier) => todo!(),
```
d0cd: Bailing with an error message is fine if this is not supported.

synthesizer/program/tests/types/equivalence.rs
```diff
@@ -0,0 +1,236 @@
...
+type CurrentNetwork = MainnetV0;
+
+// ---------- Helper ----------
+fn sample_stack(program_text: &str) -> Result<Stack<CurrentNetwork>> {
+    let program = Program::from_str(program_text)?;
+    let stack = Stack::new(&Process::load()?, &program)?;
+    Ok(stack)
+}
+
+// ---------- 1️⃣ Literal equivalence ----------
```
d0cd: Nit. We should get rid of the numerical emojis in this file, to support our friends whose editors can't render :)

synthesizer/src/vm/tests/test_v13.rs
```diff
@@ -0,0 +1,105 @@
...
+    // when deploying the second program.
+    let block = deploy_programs(ConsensusVersion::V11);
+
+    assert_eq!(block.transactions().num_accepted(), 0);
+    assert_eq!(block.transactions().num_rejected(), 0);
+    assert_eq!(block.aborted_transaction_ids().len(), 1);
+}
+
+// This test verifies that a program with external structs can be deployed on
+// consensus version 12.
```
d0cd:
```suggestion
// consensus version 13.
```

synthesizer/src/vm/tests/test_v13.rs
```diff
@@ -0,0 +1,105 @@
...
+}
+
+fn deploy_programs(consensus_version: ConsensusVersion) -> Block<CurrentNetwork> {
+    let rng = &mut TestRng::default();
+
+    // Initialize a new caller.
+    let caller_private_key = crate::vm::test_helpers::sample_genesis_private_key(rng);
+
+    // Initialize the VM at the correct height.
+    let v10_height = CurrentNetwork::CONSENSUS_HEIGHT(consensus_version).unwrap();
```
d0cd:
```suggestion
    let height = CurrentNetwork::CONSENSUS_HEIGHT(consensus_version).unwrap();
```

synthesizer/src/vm/tests/test_v13.rs
```diff
@@ -0,0 +1,105 @@
...
+    let vm = crate::vm::test_helpers::sample_vm_at_height(v10_height, rng);
```
d0cd:
```suggestion
    let vm = crate::vm::test_helpers::sample_vm_at_height(height, rng);
```

synthesizer/src/vm/verify.rs
```diff
@@ -255,6 +256,21 @@ impl<N: Network, C: ConsensusStorage<N>> VM<N, C> {
...
                 }
 
+                // If the `CONSENSUS_VERSION` is less than `V13`, then verify that:
+                //   - the program does not use the external struct syntax `some_program.aleo/StructT`
+                // If the `CONSENSUS_VERSION` is greater than or equal to `V13`, then verify that:
+                //   - the program's mappings do not use non-existent structs.
+                if consensus_version < ConsensusVersion::V13 {
+                    ensure!(
+                        !deployment.program().contains_external_struct(),
+                        "Invalid deployment transaction '{id}' - external structs may only be used beginning with Consensus version 10"
```
d0cd:
```suggestion
                        "Invalid deployment transaction '{id}' - external structs may only be used beginning with `ConsensusVersion::V13`"
```

synthesizer/tests/tests/vm/execute_and_finalize/external_struct_in_array.aleo
```diff
@@ -0,0 +1,45 @@
...
+    assert.eq r0[1u32].commission  r1[1u32].commission;
+
+// Main function: constructs two identical arrays and checks them
+function main:
+    input r0 as boolean.private;
+    input r1 as u8.private;
+
+    // Build two identical committee_state structs
+    cast r0 r1 into r2 as credits.aleo/committee_state;
+    cast r0 r1 into r3 as credits.aleo/committee_state;
```
d0cd, line 36: Do you think it would be worthwhile to test two arrays one with external and one with local?

d0cd (commented):
It's getting close, I'm aligned with this notion of partial nominal and structural equivalence. Left a few suggestions.

A couple points/questions:
- What is the import rule? If a program A uses an external struct in B which itself uses an external struct in C, does A have to import B and C? It would be great to document and test the rule.
- Before merging, we should refresh the program regressions and run it to make sure that the VM can still be initialized with all of the new checks.

mohammadfawaz:
@d0cd I believe I addressed all your comments in https://github.com/ProvableHQ/snarkVM/pull/3045/commits/93e4e4807f6e2c04fa1d2e93f6c1f42f352dc28d except for 

> This check is performed in verify.rs gated by a consensus version. Does it also need to happen here?

I'm not sure why the check is gated by a consensus version check in `verify.rs`. Do you remember why Michael added this? Also, if we do want to added a consensus version check in `synthesizer/process/src/stack/mod.rs` too, how do we go about that? How do we get the consensus version in that file?

> What is the import rule? If a program A uses an external struct in B which itself uses an external struct in C, does A have to import B and C? It would be great to document and test the rule.

I added a test that showcases that transitive dependencies are not required at the top level as long as the top level does not explicitly use the external struct name directly. I think this rule matches how `import` is meant to work for all symbols.

> Before merging, we should refresh the program regressions and run it to make sure that the VM can still be initialized with all of the new checks.

What does this mean exactly?

vicsn:
> I'm not sure why the check is gated by a consensus version check in verify.rs. Do you remember why Michael added this? Also, if we do want to added a consensus version check in synthesizer/process/src/stack/mod.rs too, how do we go about that? How do we get the consensus version in that file?

The consensus version check in verify.rs was added to ensure all validators agree external structs are only enabled from the new consensus version onwards. I think this comment's point is that the logic in synthesizer/process/src/stack/mod.rs might be duplicate and superfluous.

d0cd:
> > Before merging, we should refresh the program regressions and run it to make sure that the VM can still be initialized with all of the new checks.
> 
> What does this mean exactly?

The regression repo is [here](github.com/ProvableHQ/aleo-program-regressions).
It contains a few utilities we can run to make sure changes are backwards compatible.

vicsn (approved):

vicsn:
Thank you everyone :)

PR #3047, opened 2025-12-02
[Feat] Facilitate record queries

ljedrz:
This PR introduces 2 new methods, `Ledger::{get_record_count, get_num_block_records}`, and a few helper "getter" methods that they require to work.

A snarkOS-side counterpart PR will follow.

### DIFF
--- a/ledger/src/get.rs
+++ b/ledger/src/get.rs
@@ -382,15 +382,15 @@ impl<N: Network, C: ConsensusStorage<N>> Ledger<N, C> {
     }
 
     /// Returns a tuple containing the number of all the input records and all the output records.
-    pub fn get_record_count(&self) -> (usize, usize) {
+    pub fn get_record_count(&self) -> RecordCount {
         let transition_store = self.vm.block_store().transition_store();
         let num_input_records = transition_store.input_store().record_map().len_confirmed();
         let num_output_records = transition_store.output_store().record_map().len_confirmed();
-        (num_input_records, num_output_records)
+        RecordCount { input: num_input_records, output: num_output_records }
     }
 
     /// Returns the list of input and output records applicable to the given block height.
-    pub fn get_num_block_records(&self, height: u32) -> Result<(usize, usize)> {
+    pub fn get_num_block_records(&self, height: u32) -> Result<RecordCount> {
         let block_store = self.vm.block_store();
         let block_hash = match block_store.get_block_hash(height)? {
             Some(block_hash) => block_hash,
@@ -463,6 +463,13 @@ impl<N: Network, C: ConsensusStorage<N>> Ledger<N, C> {
             }
         }
 
-        Ok((num_input_records, num_output_records))
+        Ok(RecordCount { input: num_input_records, output: num_output_records })
     }
+}
+
+/// An object representing the number of input and output records.
+#[derive(Debug, Clone, Copy, Default, serde::Deserialize, serde::Serialize)]
+pub struct RecordCount {
+    pub input: usize,
+    pub output: usize,
 }
--- a/ledger/src/get.rs
+++ b/ledger/src/get.rs
@@ -378,4 +378,12 @@ impl<N: Network, C: ConsensusStorage<N>> Ledger<N, C> {
             _ => bail!("Invalid bond_state in finalize storage."),
         }
     }
+
+    /// Returns a tuple containing the number of all the input records and all the output records.
+    pub fn get_record_count(&self) -> (usize, usize) {
+        let transition_store = self.vm.block_store().transition_store();
+        let num_input_records = transition_store.input_store().record_map().len_confirmed();
+        let num_output_records = transition_store.output_store().record_map().len_confirmed();
+        (num_input_records, num_output_records)
+    }
 }
--- a/ledger/src/get.rs
+++ b/ledger/src/get.rs
@@ -15,6 +15,8 @@
 
 use super::*;
 
+use crate::store::TransactionType;
+
 impl<N: Network, C: ConsensusStorage<N>> Ledger<N, C> {
     /// Returns the committee for the given `block height`.
     pub fn get_committee(&self, block_height: u32) -> Result<Option<Committee<N>>> {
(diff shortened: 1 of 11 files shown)


ledger/src/get.rs
```diff
@@ -378,4 +380,89 @@ impl<N: Network, C: ConsensusStorage<N>> Ledger<N, C> {
             _ => bail!("Invalid bond_state in finalize storage."),
         }
     }
+
+    /// Returns a tuple containing the number of all the input records and all the output records.
+    pub fn get_record_count(&self) -> (usize, usize) {
```
vicsn: Nit: a struct with named fields will save developers of dependents a lot of reasoning work - I still regret our `*_cost*` functions returning tuples.

ledger/src/get.rs
```diff
@@ -378,4 +380,89 @@ impl<N: Network, C: ConsensusStorage<N>> Ledger<N, C> {
...
+                for id in input_ids {
+                    if input_store.record_map().contains_key_confirmed(&id)? {
+                        num_input_records += 1;
+                    }
+                }
+                for id in output_ids {
+                    if output_store.record_map().contains_key_confirmed(&id)? {
+                        num_output_records += 1;
+                    }
+                }
```
vicsn, line 436: Consider unit testing or explicitly E2E testing the different cases

ledger/src/get.rs
```diff
@@ -378,4 +380,89 @@ impl<N: Network, C: ConsensusStorage<N>> Ledger<N, C> {
...
+        for (tx_id, tx_ty) in transaction_ids_with_type {
+            match *tx_ty {
+                TransactionType::Deploy => {
+                    continue;
+                }
+                TransactionType::Execute => {
+                    let Some(transition_ids_w_fee) = execution_store.id_map().get_confirmed(&tx_id)? else {
+                        continue;
+                    };
+                    let (transition_ids, _fee) = &*transition_ids_w_fee;
```
vicsn, line 451: Should we count I/O records in the _fee?

ljedrz: The fee records should be counted by the loop in L455 - the fee in this particular map is only an indicator if there was a fee associated with the related execution and its transitions.

ledger/src/get.rs
ljedrz: Good point, though the place for these structs would be snarkOS, no? Or is it ok if we include "target" objects in this module?

ledger/src/get.rs
ljedrz: I'd be happy to do so; is there any existing test that guarantees the existence of records, which I could repurpose for record counting checks?

vicsn: Actually ongoing work by @antonio95 :)

ledger/src/get.rs
vicsn: Well we define the functions here... So we need it here.

ledger/src/get.rs
vicsn: Hmm I'm not familiar with the database schema, but Fee transactions are different from Fee objects used in Execute or Deploy transactions.

ledger/src/get.rs
ljedrz: https://github.com/ProvableHQ/snarkVM/pull/3047/commits/657a046a03d89088f86c9f8dd3a8c762e94b7054

ledger/src/get.rs
Antonio95: Hi @ljedrz ! For some more context, what are your testing needs?

I recently added some functionality that I wanted to specifically test (among other cases) on transactions involving transitions which consumed records. (*Extra background: I went through the test cases in `synthesizer/tests/test_vm_execute_and_finalize.rs` and saw the only one with a chance of record consumption was `synthesizer/tests/tests/vm/execute_and_finalize/mint_and_split.aleo`. I modified the inputs to correctly consume one and my tested function did behave correctly; however, a new commit shortly after broke that so that my function was no longer being tested under record consumption. We suspect even if we fix the `mint_and_split.aleo` test case again, it will soon break.*) In summary, I couldn't find any high-level tests which reliably involved record consumption.

As part of the dynamic dispatch work, we are adding many tests involving record consumption. Most test cases in the subfolders of `https://github.com/ProvableHQ/snarkVM/blob/3a67e6a1f2091500d2a0a4778bfa0c51495ae1fb/synthesizer/src/vm/tests/test_v12` do (this is all happening in the `feat/dynamic-dispatch-extension-case1wip` branch). Cf. e. g. https://github.com/ProvableHQ/snarkVM/blob/3a67e6a1f2091500d2a0a4778bfa0c51495ae1fb/synthesizer/src/vm/tests/test_v12/mixed.rs#L25 for a specific example.

That code will likely not be merged for a few weeks, so if you want to add tests before that you can take inspiration from what I sent, or I'd also be happy to write a test case if you tell me what you need in a bit more detail :)

ljedrz: CC @niklaslong

ledger/src/get.rs
```diff
@@ -378,4 +380,96 @@ impl<N: Network, C: ConsensusStorage<N>> Ledger<N, C> {
             _ => bail!("Invalid bond_state in finalize storage."),
         }
     }
+
+    /// Returns a tuple containing the number of all the input records and all the output records.
+    pub fn get_record_count(&self) -> RecordCount {
+        let transition_store = self.vm.block_store().transition_store();
+        let num_input_records = transition_store.input_store().record_map().len_confirmed();
```
raychu86, line 387: These seem quite expensive, as it's a O(N) scan for each `len_confirmed` call.

raychu86:
Why is this feature necessary?

ljedrz:
> Why is this feature necessary?

@niklaslong would you like to provide the rationale?

ledger/src/get.rs
ljedrz: I'd call it _somewhat_ expensive - in all likelihood, it's the other call that's the more expensive one due to the number of queries involved; obtaining the number of records for a given map **is** O(N), but it's an optimized operation.

Issue #2983, opened 2025-10-24
[Feature] Evaluate whether to extend CI with static analysis


vicsn:
## 🚀 Feature

More specifically, tools which were suggested to us by auditors include:
- Semgrep
- Dylint
- cargo-edit
- cargo hack
- Zizmor

If we're happy with any of these, we can also extend to snarkOS

meddle0x53:
So I spent some time on checking `semgrep` and it is a fast static analysis tool that scans source code and configuration files using simple rule-based patterns, regexes mainly. Unlike compiler-level tools, it works purely on the structure of files, which makes it extremely quick (around 7-8 seconds for the entire SnarkVM repository in the moment, but I'll work on the specific rules that make sense for an example PR).

It is good for enforcing project-specific conventions.

It adds the following:

Additional safety layer on top of Clippy and tests.
Semgrep catches patterns that the Rust compiler and Clippy don’t cover - such as forbidden unwrap(), panic!, println!, unsafe patterns, and other conventions we choose to enforce. I will add examples on how to use these in the example PR; For this project, as we use panic!, todo!, etc I will add just an idea so we can, for example remove them in the future.

It doesn’t just analyze Rust code: Semgrep also inspects YAML, TOML, Markdown, JSON, and shell scripts, which allows us to detect issues in CI configs, Cargo metadata, documentation, and helper scripts.

Semgrep reliably flags leftover <<<<<<< HEAD / >>>>>>> branch markers in all supported file types, preventing accidental commits of broken files. We had commits with problems like that before. In source files the CI breaks but in some markdown or config files it may pass and break runtime (for the JSON, toml, yaml).

It allows custom repository-specific rules, I've been checking good practice-rules for rust and will add for yml, shell scripts, , json and markdown files. I did a `find . -type f -not -path "./target/*" -name '*.*'   | sed -E 's/.*\.([^.]+)$/\1/'   | tr 'A-Z' 'a-z'   | sort   | uniq -c   | sort -nr` and these file types make sense for this repository.

I think it won't make the CI slower and heavier and only has pros, so will make a PR with meaningful rules for the file types I mentioned.

meddle0x53:
## Dylint

Dylint is a tool for running custom Clippy-like lints without needing to modify the Rust compiler or Clippy itself. It’s essentially a framework that lets you write, test, and run your own lint crates that behave exactly like Clippy lints — but distributed as normal Rust crates and dynamically loaded at runtime.

[...]

PR #3049, opened 2025-12-03
Add Semgrep to the CI workflows

meddle0x53:
## Motivation

Following https://github.com/ProvableHQ/snarkVM/issues/2983 we are testing some of the tools suggested there and Semgrep seems to makes sense for additional analysis of the files.

This PR adds it and also a set of custom rules to be applied by it.

## Test Plan

The CI passes the new workflow.

## Documentation

The tool semgrep is a fast static analysis tool that scans source code and configuration files using simple rule-based patterns, regexes mainly. Unlike compiler-level tools, it works purely on the structure of files, which makes it extremely quick (4-5s with the custom rules in this PR)

It is good for enforcing project-specific conventions.

It adds the following:

Additional safety layer on top of Clippy and tests.
Semgrep catches patterns that the Rust compiler and Clippy don’t cover - such as forbidden unwrap(), panic!, println!, unsafe patterns, and other conventions we choose to enforce. 

It doesn’t just analyze Rust code: Semgrep also inspects YAML, TOML, Markdown, JSON, and shell scripts, which allows us to detect issues in CI configs, Cargo metadata, documentation, and helper scripts.

Semgrep reliably flags leftover <<<<<<< HEAD / >>>>>>> branch markers in all supported file types, preventing accidental commits of broken files. In source files the CI breaks but in some markdown or config files it may pass and break runtime (for the JSON, toml, yaml).

It allows custom repository-specific rules.

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -85,6 +85,29 @@ commands:
             - target
           key: cargo-cache-{{ arch }}-{{ checksum "Cargo.lock" }}
 
+  run_semgrep:
+    description: "Run Semgrep static analysis"
+    steps:
+      - checkout
+      - run:
+          name: Install Semgrep
+          command: |
+            set -euo pipefail
+
+            python3 -m pip install --user semgrep
+            echo 'export PATH="$HOME/.local/bin:$PATH"' >> "$BASH_ENV"
+      - run:
+          name: Run Semgrep (Rust + YAML/TOML + Markdown + Shell + JSON)
+          command: |
+            set -euo pipefail
+            semgrep \
+              --config semgrep-rules/rust.yml \
+              --config semgrep-rules/yaml-toml.yml \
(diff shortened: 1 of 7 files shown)


vicsn:
I'm happy you're testing, but seeing this, I think it's not worth the effort. :)

meddle0x53:
> I'm happy you're testing, but seeing this, I think it's not worth the effort. :)

OK, should I continue with checking the other tools from the issue https://github.com/ProvableHQ/snarkVM/issues/2983 or should I move to something else, what do you think?

vicsn:
I'd be keen to hear your thoughts on which are useful with examples

meddle0x53:
> I'd be keen to hear your thoughts on which are useful with examples

Ok, I'll do small researches (max an hour per tool) on all of them and add to the ticket.

PR #3050, opened 2025-12-04
[Feature] Add `CheckBlockError`

kaimast:
## Addition of `CheckBlockError`
This PR enables detecting race conditions when verifying a block. In particular, it returns a machine-readable `CheckBlockError` that contains an error type for the case when a block already exists in the ledger.
On the snarkOS side (see https://github.com/ProvableHQ/snarkOS/pull/4039), block sync will remove the affected pending blocks and retry when it encounters this error. This allows us to cut down on errors like `Unable to advance to the next block - Failed to speculate on transactions - Failed to post-ratify - Next round 2532 must be greater than current round 2532`.

One alternative to this approach would be to return `anyhow::Result<Option<PendingBlock>>`. I do not like this approach, as it is not really clear what `Ok(None)` means without reading the function's doc comment.
Another alternative would be to match the error message against a string, but that can easily break in the future and is generally considered bad practice.

Finally, it would be possible to use locks on the snarkOS side instead. I try to avoid locking there whenever possible to reduce the potential for deadlocks and to maintain good performance. The linked snarkOS PR still uses locks, but only when it is actually advancing the ledger.

The change is only made to `Ledger::check_block_subdag`, to ensure the existing `Ledger::check_next_block` still works as before. Maybe at some point we could consider cleaning up some of the block checking and advancement API, but now is not that time. 

### Atomic block checks
Additionally, the PR changes`Ledger::check_next_block` and `Ledger::check_block_subdag` so that it is guaranteed no concurrent block advancement happens.

The following is a new error message that appeared after the above changes, and is addressed by this:
```
ERROR snarkos_node_bft::sync: Block synchronization failed — Failed to check contents of pending block ab1y3l48teq38zm9kftznt60vpgl4ydq09f6qsx6mr269rw5qef7uqsjzd7l5 at height 342 — Failed to speculate over unconfirmed transactions — Invalid transaction found in the transactions list: Transaction 'at1zuc4zprjff66q8ftc8jtt6e9xa7ltym5ttg348jz37j6fnel3yyseqrhsc' already exists in the ledger
```

`check_next_block` and `check_block_subdag` now acquire a read-lock on `current_block`, which prevents the ledger advancing while the check is performed. In other words, they are no *atomic* operations.
On the snarkOS side, we still need to account for concurrent updates between the block check and the block advancement, but this atomicity guarantee makes it a lot easier and avoids messy errors, such as the one above.

### DIFF
--- a/ledger/src/check_next_block.rs
+++ b/ledger/src/check_next_block.rs
@@ -78,30 +78,30 @@ impl<N: Network, C: ConsensusStorage<N>> Ledger<N, C> {
     ///
     /// # Arguments
     /// * `block` - The block to check.
-    /// * `pending_block` - A sequence of blocks between the block to check and the current height of the ledger.
+    /// * `prefix` - A sequence of blocks between the block to check and the current height of the ledger.
+    ///
+    /// # Returns
+    /// * On success, a [`PendingBlock`] representing the block that was checked. Once the prefix of this block has been fully added to the ledger,
+    ///   the [`PendingBlock`] can then be passed to [`Self::check_block_content`] to fully verify it.
+    /// * On failure, a [`CheckBlockError`] describing the reason the block was rejected.
     ///
     /// # Notes
     /// * This does *not* check that the header of the block is correct or execute/verify any of the transmissions contained within it.
-    ///
     /// * In most cases, you want to use [`Self::check_next_block`] instead to perform a full verification.
-    ///
     /// * This will reject any blocks with a height <= the current height and any blocks with a height >= the current height + GC.
-    ///   For the former, a valid block already exists and,
-    ///   for the latter, the comittte is still unknown.
+    ///   For the former, a valid block already exists and,for the latter, the comittee is still unknown.
+    /// * This function executes atomically, in that there is guaranteed to be no concurrent updates to the ledger during its execution.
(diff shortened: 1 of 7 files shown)


ledger/src/check_next_block.rs
```diff
@@ -49,46 +80,60 @@ impl<N: Network, C: ConsensusStorage<N>> Ledger<N, C> {
...
         // The hash checks should be redundant, but we perform them out of extra caution.
         let mut expected_height = self.latest_height() + 1;
         for pending in pending_blocks {
             if pending.height() != expected_height {
-                bail!(
-                    "Pending block has invalid height. Expected {expected_height}, but got {actual}.",
-                    actual = pending.height()
-                );
+                return Err(CheckBlockError::InvalidPrefix {
+                    height: pending.height(),
```
vicsn: Slightly surprising to me that the height of the mismatched prefix is the same as this height. Maybe you want to use pending_tip.height() here (as well as for the other prefix errors)

vicsn (approved):

ledger/src/check_next_block.rs
kaimast: It doesn't make a lot of sense for this case, but for the on below it would tell you the specific block that had an incorrect hash.
Maybe it should just be the index of the slice of pending blocks?

kaimast: I changed it to return the index of the slice, instead.

vicsn (approved):

PR #3051, opened 2025-12-06
[Refactor] Output record register check from V10

d0cd:
This PR factors out the output record register check that was introduced in `ConsensusVersion::V10`. This allows downstream tools to check programs and surface better error messages to users.


mohammadfawaz (approved):
LGTM! Thanks. Yeah we can now use that in Leo codegen to emit an error and propose a workaround.

raychu86 (approved):

vicsn (approved):

PR #3054, opened 2025-12-08
[Bench] Remove old benchmark data

kaimast:
This PR cleans up the benchmark page, that has become hard to search through. To preview the changes, you can check out this branch (`dev/remove-benchmark-noise`) and then open `dev/bench/index.html`.

Changes:
* Removes all data from before 2024. Right now, some graphs have 1000s of data points, making it hard to read them and slows down render time. The old data is still in the git history, so no information is lost.
* Add a header element to the benchmark page, so that you can search for a specific graph using `Ctrl+F`
* Made the graphs slightly wider (`95%` instead of `1000px`)

### DIFF
--- a/dev/bench/index.html
+++ b/dev/bench/index.html
@@ -171,6 +171,10 @@
         function renderAllChars(dataSets) {
 
           function renderGraph(parent, name, dataset) {
+            let header = document.createElement('h3');
+            header.textContent = name;
+            parent.appendChild(header);
+
             const canvas = document.createElement('canvas');
             canvas.className = 'benchmark-chart';
             parent.appendChild(canvas);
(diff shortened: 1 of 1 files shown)


raychu86:
I do not like removing old data outright; I also consider that data only in git history is effectively lost because nobody will dig it up (and will not know that it even exists).

Is there a way to remove remove old benchmark data in a way that we still see the gist it? So we remove most runs, but keep a datapoint from a run that happened every month or so

raychu86 (approved):

kaimast:
@raychu86 I changed the PR to include a simple vibe-coded UI that lets you select a time range. This way, we do not need to remove any data.

PR #3058, opened 2025-12-10
Speed up ledger tests

meddle0x53:
## Motivation

Part of the work required for #2996

Testing some ideas with shared genesis block and some heavy calls removed or reworked.

## Test Plan

Monitor the CI for speedups, but even before that try locally to see if there is a visible speed up.


.config/nextest.toml
```diff
@@ -0,0 +1,107 @@
+# Make sure nextest is at least this version.
+nextest-version = "0.9.48"
+
+# 1) Define test groups at the TOP LEVEL (not inside a profile)
+# Each group has a max-threads concurrency.
+[test-groups]
+ledger-heavy = { max-threads = 1 }    # run these tests strictly one at a time
```
vicsn, line 7: Can you explain why you designed these to run one at a time?

circuit/types/group/src/add.rs
```diff
@@ -138,7 +138,7 @@ mod tests {
     use super::*;
     use snarkvm_circuit_environment::Circuit;
 
-    const ITERATIONS: u64 = 100;
+    const ITERATIONS: u64 = 90;
```
vicsn: Can you lower all of them  in the circuit crate to 10? Adding an explanation in the PR description for reviewers: "circuit test iterations are reduced because (1) we are running CI more and more often which are iterations by themselves (2)  they are essentially used as a dumb fuzzer which is unlikely to lead to new findings - in 2 years they didn't randomly break. Besides lowering these to match our actual compute capacity, any future spare compute can be put into a higher leverage fuzzer such as what we started and documented here https://provable.com/blog/fuzzing-the-Aleo-VM

curves/src/traits/tests_field.rs
```diff
@@ -330,7 +330,7 @@ pub fn random_sqrt_tonelli_tests<F: PrimeField + SquareRootField>(rng: &mut Test
 
 #[cfg(test)]
 pub(crate) fn bench_sqrt<F: PrimeField + SquareRootField>(rng: &mut TestRng) {
```
vicsn, line 332: Can you move this to an existing or new benchmark file? See other bench examples in this repo

ledger/block/src/header/merkle.rs
```diff
@@ -94,7 +94,7 @@ mod tests {
 
     type CurrentNetwork = MainnetV0;
 
-    const ITERATIONS: u64 = 1_000;
+    const ITERATIONS: u64 = 900;
```
vicsn: You can lower iterations in the ledger crate also to 10. Besides the other comment about it being a bad fuzzer, this one has the additional argument that we regularly run this in production (on every new block) so it won't break silently.

ledger/block/src/transactions/confirmed/bytes.rs
```diff
@@ -16,99 +16,84 @@
 use super::*;
 
 impl<N: Network> FromBytes for ConfirmedTransaction<N> {
-    /// Reads the confirmed transaction from a buffer.
     fn read_le<R: Read>(mut reader: R) -> IoResult<Self> {
+        fn read_finalize_ops<N: Network, R: Read>(
```
vicsn: Changing deserialization logic seems out of scope. We should keeo things easy for reviewers.

ledger/block/src/transactions/confirmed/mod.rs
```diff
@@ -363,9 +363,25 @@ impl<N: Network> Deref for ConfirmedTransaction<N> {
 #[cfg(test)]
 pub mod test_helpers {
     use super::*;
-    use console::network::MainnetV0;
 
-    type CurrentNetwork = MainnetV0;
+    type CurrentNetwork = console::network::MainnetV0;
```
vicsn: Was this AI?

ledger/block/src/transactions/confirmed/mod.rs
```diff
@@ -481,8 +507,68 @@ mod test {
...
 
+    use once_cell::sync::Lazy;
+
     type CurrentNetwork = console::network::MainnetV0;
 
+    static FIXTURES: Lazy<Fixtures> = Lazy::new(|| {
+        // Deterministic, so no flakes and stable perf.
+        let rng = &mut TestRng::fixed(123456789);
+
+        // Use the existing helpers that produce VALID finalize ops / rejected objects.
```
vicsn: A future reader won't know the distinction between existing and new code

ledger/block/src/transactions/confirmed/mod.rs
```diff
@@ -481,8 +507,68 @@ mod test {
     use super::*;
     use crate::transactions::confirmed::test_helpers;
 
+    use once_cell::sync::Lazy;
+
     type CurrentNetwork = console::network::MainnetV0;
 
+    static FIXTURES: Lazy<Fixtures> = Lazy::new(|| {
```
vicsn: The whole lazy loading approach feels suboptimal for performance (contention, duplicate work across runs) compared to reading from disk.

Note that if we do go the read from disk route, we also need a test that confirms the data on disk matches the expected generated data. And that test can be run with an environment variable REWRITE_EXPECTATIONS=1 to overwrite the values (we do this in one other snarkVM test suite already)

ledger/block/src/transactions/confirmed/serialize.rs
```diff
@@ -124,51 +124,103 @@ impl<'de, N: Network> Deserialize<'de> for ConfirmedTransaction<N> {
...
-        expected: T,
-    ) {
-        // Serialize
+    type CurrentNetwork = MainnetV0;
+
+    static SAMPLES: Lazy<Vec<ConfirmedTransaction<CurrentNetwork>>> =
+        Lazy::new(crate::transactions::confirmed::test_helpers::sample_confirmed_transactions_for_serde);
+
+    fn sample_limit() -> usize {
+        std::env::var("SNARKVM_SERDE_SAMPLES").ok().and_then(|v| v.parse::<usize>().ok()).unwrap_or(usize::MAX)
```
vicsn: Can we stick to the ITERATIONS syntax?

ledger/block/src/transactions/serialize.rs
```diff
@@ -67,211 +67,148 @@ impl<'de, N: Network> Deserialize<'de> for Transactions<N> {
...
 
-    const ITERATIONS: u32 = 6;
+    /// Default number of samples used by these tests.
+    /// Override with SNARKVM_TX_SERDE_SAMPLES (e.g. 1, 2, 4).
+    fn sample_limit() -> usize {
+        std::env::var("SNARKVM_TX_SERDE_SAMPLES").ok().and_then(|v| v.parse::<usize>().ok()).unwrap_or(2)
+    }
+
+    /// If we explicitly want heavy coverage (slow), run with:
+    /// SNARKVM_TX_SERDE_HEAVY=1
```
vicsn: I do like the idea of running heavier tests only on merge.

But I dont think we have to worry about this deserialization thát much, these tests are not fuzzers.

So we can just pick a single ITERATIONS value

ledger/block/src/transactions/serialize.rs
```diff
@@ -67,211 +67,148 @@ impl<'de, N: Network> Deserialize<'de> for Transactions<N> {
...
+    fn test_bytes() {
+        for txs in iter_sets() {
+            tobytes_roundtrip(txs);
+        }
+    }
 
-        // Check the serialization.
-        check_bincode(crate::transactions::test_helpers::sample_block_transactions(rng));
+    #[test]
+    fn test_display_fromstr_smoke() {
```
vicsn: We don't have any test with postfix _smoke

ledger/src/test_helpers.rs
```diff
@@ -62,14 +73,103 @@ pub struct TestEnv {
...
+    // Prefer the workspace target dir. If CARGO_TARGET_DIR is set, use it.
+    // Otherwise, assume snarkvm-ledger is at <workspace>/ledger and target is at <workspace>/target.
+    let target_dir = std::env::var_os("CARGO_TARGET_DIR")
+        .map(PathBuf::from)
+        .unwrap_or_else(|| PathBuf::from(env!("CARGO_MANIFEST_DIR")).join("..").join("target"));
+
+    let dir = target_dir.join("test-cache");
+    let _ = fs::create_dir_all(&dir);
+
+    // Versioned so you don’t reuse an incompatible cache after upgrades.
```
vicsn: Not needed, genesis should never change

ledger/src/test_helpers.rs
```diff
@@ -62,14 +73,103 @@ pub struct TestEnv {
...
 }
 
-pub fn sample_test_env(rng: &mut (impl Rng + CryptoRng)) -> TestEnv {
-    // Sample the genesis private key.
+pub struct SharedGenesis {
+    pub private_key: PrivateKey<CurrentNetwork>,
+    pub genesis: Block<CurrentNetwork>,
+}
+
+fn genesis_cache_path() -> PathBuf {
```
vicsn: Nice!

ledger/src/test_helpers.rs
```diff
@@ -62,14 +73,103 @@ pub struct TestEnv {
...
+
     let private_key = PrivateKey::<CurrentNetwork>::new(rng).unwrap();
+    let store = CurrentConsensusStore::open(StorageMode::new_test(None)).unwrap();
+    let genesis = VM::from(store).unwrap().genesis_beacon(&private_key, rng).unwrap();
+
+    SharedGenesis { private_key, genesis }
+}
+
+// Process-local cache, backed by the shared on-disk cache.
+// This is the piece that makes nextest multi-process runs benefit.
```
vicsn: Can you explain why the lazy is needed?

ledger/src/test_helpers.rs
```diff
@@ -62,14 +73,103 @@ pub struct TestEnv {
...
+pub static SHARED_GENESIS: Lazy<SharedGenesis> = Lazy::new(|| {
+    if let Some(shared) = read_shared_genesis_from_file() {
+        return shared;
+    }
+
+    let shared = compute_shared_genesis();
+
+    // If someone else won the race and wrote it while we computed, just use theirs.
```
vicsn: Funny setup. Can we use a more explicit REWRITE_EXPECTATIONS environment variable and commit the cache to disk?

ledger/store/src/helpers/rocksdb/internal/tests.rs
```diff
@@ -145,7 +145,7 @@ fn test_scalar_mul() {
 
     let rng = &mut TestRng::default();
 
-    const ITERATIONS: u32 = 1_000_000u32;
+    const ITERATIONS: u32 = 900_000u32;
```
vicsn: Depends on the runtime how low we can make this, would a second be a good target...? Do we just do 1000 if this really came up in your profiling?

synthesizer/process/src/tests/test_serializers.rs
```diff
@@ -35,7 +35,7 @@ type CurrentAleo = AleoV0;
 #[test]
 fn test_serialize_deserialize_equivalence() {
     // Number of iterations to run for each type.
-    const ITERATIONS: usize = 50;
+    const ITERATIONS: usize = 40;
```
vicsn, line 39: Synthesizer crate iterations can also go down to 10

synthesizer/program/tests/instruction/hash.rs
```diff
@@ -74,10 +75,54 @@ use snarkvm_synthesizer_program::{
...
+    static STACK_CACHE: RefCell<HashMap<StackKey, CachedStack>> = RefCell::new(HashMap::new());
+}
+
+thread_local! {
+    static HASH_RNG: RefCell<TestRng> = RefCell::new(TestRng::default());
+}
+
+const ITERATIONS: usize = 10;
+
+static PROCESS: Lazy<Process<CurrentNetwork>> = Lazy::new(|| Process::load().unwrap());
```
vicsn: Not sure if possible, but we may want to try just writing the ledger to disk, and loading it via load_ledger. How big would the cache be roughly?

synthesizer/program/tests/instruction/hash.rs
```diff
@@ -157,6 +204,16 @@ fn sample_valid_input_types<N: Network, R: CryptoRng + Rng>(
 /// **Attention**: When changing this, also update in `src/logic/instruction/hash.rs`.
 fn sample_valid_destination_types<N: Network>(variant: HashVariant) -> Vec<PlaintextType<N>> {
     match variant {
+        HashVariant::HashKeccak256Raw
```
vicsn: Did we miss these originally? Nice catch

synthesizer/program/tests/instruction/hash.rs
```diff
@@ -246,90 +303,81 @@ fn check_hash<const VARIANT: u8>(
...
     destination_type: &PlaintextType<CurrentNetwork>,
 ) {
-    println!("Checking '{opcode}' for '{input_type}.{mode}'");
-
-    // Initialize the stack.
-    let (stack, operands, destination) = sample_stack(opcode, input_type, *mode, destination_type).unwrap();
-
-    // Sample the input.
-    let input = stack.sample_plaintext(input_type, &mut TestRng::default()).unwrap();
+    if std::env::var("SNARKVM_HASH_TEST_LOG").is_ok() {
```
vicsn: The less new env vars the better

synthesizer/src/vm/mod.rs
```diff
@@ -629,6 +629,113 @@ pub(crate) mod test_helpers {
...
     pub(crate) type CurrentAleo = AleoV0;
 
+    use snarkvm_utilities::{FromBytes, ToBytes};
+    use std::{
+        fs,
+        io,
+        path::{Path, PathBuf},
+    };
+
+    const GENESIS_CACHE_MAGIC: &[u8] = b"SNARKVM_GENESIS_CACHE_V1\n";
```
vicsn: Haha, it is nice but please fix the name

vicsn: Note the genesis file may be unique per network (canary/testnet/mainnet)

utilities/src/biginteger/tests.rs
```diff
@@ -76,7 +76,7 @@ fn biginteger_bytes_test<B: BigInteger>(rng: &mut TestRng) {
 }
 
 fn biginteger_to_string_test<B: BigInteger>(rng: &mut TestRng) {
-    const ITERATIONS: u64 = 1_000_000;
+    const ITERATIONS: u64 = 900_000;
```
vicsn: 1000

utilities/src/bits.rs
```diff
@@ -449,7 +449,7 @@ mod tests {
     use anyhow::Result;
     use rand::{Rng, distributions::Alphanumeric};
 
-    const ITERATIONS: u64 = 10000;
+    const ITERATIONS: u64 = 8000;
```
vicsn, line 453: 1000

utilities/src/bytes.rs
```diff
@@ -582,7 +582,7 @@ mod test {
 
     use rand::Rng;
 
-    const ITERATIONS: usize = 1000;
+    const ITERATIONS: usize = 800;
```
vicsn: 10

vicsn (changes requested):

.config/nextest.toml
meddle0x53: Yes, I saw that if you run one of these with other tests in parallel it is so CPU heavy that it becomes slower and also slows down the tests being executed with it. So we keep them being executed in parallel to the not-so-heavy ones, just in new workflow and they heaviest are executed sequentially in that workflow, which actually makes them faster.

ledger/block/src/transactions/confirmed/bytes.rs
meddle0x53: I was trying many ideas as nothing was doing much of an improvement. So many of them can be stupid and will revert them. Some come from me consulting with ChatGPT without being deep in the domain too, so they can be really stupid.

ledger/block/src/transactions/confirmed/mod.rs
meddle0x53: I was using AI for ideas a lot, but this particular thing may be from me copying code that worked in one place to another place just to make the compiler happy and not thinking much to just trigger the next run... I don't think the AI is so stupid, I think it is me here.

ledger/block/src/transactions/confirmed/mod.rs
meddle0x53: This PR is a WIP it has these fixtures as one of the ideas, but in the end I don't think they are making much improvement and we can remove them (I will explain why they are like that and why we can remove them bellow).

The real improvement is:

1. The very CPU heavy tests to not be ran in parallel with the other in the same workflow, which makes all tests run slower.
2. The ITERATIONS lowering.

All other changes didn't improve things much. The lazy idea is - do this the first time it is called, all next test in this process (because of nextest, parallelism is with OS level processes) have it in the memory now. Here it can not be done not-lazy as it will be read on compile time.

If I remove the lazy and keep the FIXTURES they will be read from a file on every call. They have to be in a file so nextest shares between processes, if they are in-memory from the beginning the parallelism of nextest makes it so that the reuse is only between some of the tests that are being run sequentially in the current OS level process. So the whole lazy idea and file storage is because of nextest.

But in the end this didn't result in real speed up, so the whole idea can be scrapped.

ledger/block/src/transactions/confirmed/serialize.rs
meddle0x53: It is one idea to run them only in some circumstances as heavier and the normal CI to run them as lighter, but the ITERATIONS can do that, yeah. Just with the env var, the CI can decide when to run them as heavier and when to as lighter.

ledger/src/test_helpers.rs
meddle0x53: This is some AI idea, didn't lead to real improvements, so it can be removed.

ledger/src/test_helpers.rs
meddle0x53: I explained it in another comment. But:
1. Nextest uses OS process level parallelism, so to reuse things they can not be in-memory only, on-disk cache is used.
2. I want the state to be in a constant, but constants are compile-time, so to initialise them on runtime Lazy is used hile defining them.
3. Because of Lazy, the first time that constant is read the cache from disk is loaded in-memory, all subsequent reads to the constant in this OS-level process will not read from disk, they will read from its in-memory value now.

In the end it didn't improve speed much, though :/

ledger/src/test_helpers.rs
meddle0x53: The cache to disk is on disk only because of nextest, if all tests in the workflow were ran in threads not OS-level processes I wouldn't cache it on disk.

This "someone else winning the race" refers to another OS-level process of nextest running some of the tests.

ledger/store/src/helpers/rocksdb/internal/tests.rs
meddle0x53: It didn't really came up in profiling, I was just lowering all possible ITERATIONS to see if there are improvements. The profiling results of the heaviest tests are so time taking, but then I saw that if I ran these heaviest tests sequentially their separate run times are dramatically lower. So changing the ITERATIONS of all tests that have ITERATIONS is not based on the profiling it is a desperate move to see if something improves run time after days of ideas that didn't improve it much...

synthesizer/program/tests/instruction/hash.rs
meddle0x53: I am not sure all of this is worth it, as I didn't see much speedup from the caching at all. And you want to commit it (from another comment), but as I said all things cached to disk are because of how parallelism works in nextest, not because I wanted them on disc, really.

synthesizer/program/tests/instruction/hash.rs
meddle0x53: Yeah, this is from the idea to make tests simpler in the CI through it deciding when and where to make them simpler.

ledger/src/test_helpers.rs
vicsn: Makes sense!

I guess for some tests the genesis block alone is just not that big of a proportion then, maybe the result is better for the tests where you cache this entire process? Do you have any quantification of the impact? We can leave it if it's really insignificant, but since so many tests depend on a genesis block I'd hope that the benefits add up.

ledger/store/src/helpers/rocksdb/internal/tests.rs
vicsn: If this test doesn't have a significant runtime we can leave the iterations untouched

synthesizer/program/tests/instruction/hash.rs
vicsn: Ok we can defer

circuit/types/group/src/add.rs
meddle0x53: Should we try this? What is the final decision?

ledger/block/src/transactions/confirmed/mod.rs
meddle0x53: Yeah...

circuit/types/group/src/add.rs
vicsn: Yes can you open a new PR just lowering the iterations? You can build on top of this branch: https://github.com/ProvableHQ/snarkVM/pull/3078

PR #3059, opened 2025-12-10
[Fix] One-off loading in epoch cache

raychu86:
## Motivation

The existing code has a one-off issue with loading the `epoch_provers_cache`. The correct behavior when nodes are operating is that the cache should be cleared when the block height reaches an epoch boundary (`block_height % N::NUM_BLOCKS_PER_EPOCH == 0`). However the current logic when a node loads from DB and the block state is exactly at the epoch boundary will not account for this cleared cache. This can lead to deviation from expected values in the cache. 

This PR fixes this by making sure the ledger has a clear cache when loading from DB and the block state starts at exactly the epoch boundary.

### DIFF
--- a/ledger/src/lib.rs
+++ b/ledger/src/lib.rs
@@ -265,6 +265,14 @@ impl<N: Network, C: ConsensusStorage<N>> Ledger<N, C> {
     pub fn load_epoch_provers(&self) -> IndexMap<Address<N>, u32> {
         // Fetch the block heights that belong to the current epoch.
         let current_block_height = self.vm().block_store().current_block_height();
+
+        // If the current block height is at the start of an epoch, return an empty map.
+        // This is safe because it is once a ledger advances to the next epoch, then the cache should be empty
+        // to prepare for solutions on the next epoch.
+        if current_block_height % N::NUM_BLOCKS_PER_EPOCH == 0 {
+            return IndexMap::new();
+        }
+
         let start_of_epoch = current_block_height.saturating_sub(current_block_height % N::NUM_BLOCKS_PER_EPOCH);
         let existing_epoch_blocks: Vec<_> = (start_of_epoch..=current_block_height).collect();
 
--- a/ledger/src/lib.rs
+++ b/ledger/src/lib.rs
@@ -266,8 +266,9 @@ impl<N: Network, C: ConsensusStorage<N>> Ledger<N, C> {
         // Fetch the block heights that belong to the current epoch.
         let current_block_height = self.vm().block_store().current_block_height();
 
-        // Determine the first block to start checking. Note that the starting block of the epoch (where current_block_height % N::NUM_BLOCKS_PER_EPOCH == 0),
-        // will have solutions counted towards the previous epoch, so we start from the next block.
+        // Determine the first block to start checking.
+        // Note that the epoch boundary (where current_block_height % N::NUM_BLOCKS_PER_EPOCH == 0) can contain solutions
+        // for the previous epoch X. The subsequent block is the first block to contain solutions for the current epoch X+1.
         let start =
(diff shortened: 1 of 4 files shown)


ledger/src/lib.rs
```diff
@@ -265,8 +265,13 @@ impl<N: Network, C: ConsensusStorage<N>> Ledger<N, C> {
...
         // Fetch the block heights that belong to the current epoch.
         let current_block_height = self.vm().block_store().current_block_height();
-        let start_of_epoch = current_block_height.saturating_sub(current_block_height % N::NUM_BLOCKS_PER_EPOCH);
-        let existing_epoch_blocks: Vec<_> = (start_of_epoch..=current_block_height).collect();
+
+        // Determine the first block to start checking.
+        // Note that the epoch boundary (where current_block_height % N::NUM_BLOCKS_PER_EPOCH == 0) can contain solutions
+        // for the previous epoch X. The subsequent block is the first block to contain solutions for the current epoch X+1.
+        let start =
+            current_block_height.saturating_sub(current_block_height % N::NUM_BLOCKS_PER_EPOCH).saturating_add(1);
```
vicsn: If `current_block_height=NUM_BLOCKS_PER_EPOCH`, `start=NUM_BLOCKS_PER_EPOCH+1`, which seems odd if not wrong.

How about:
```
current_epoch_block_height = current_block_height - 1
start = current_epoch_block_height - (current_epoch_block_height % NUM_BLOCKS_PER_EPOCH) + 1
```

For `NUM_BLOCKS_PER_EPOCH=10`:
- current_block_height=1 -> start=1
- current_block_height=10 -> start=1
- current_block_height=11 -> start=11

vicsn (changes requested):

ledger/src/tests.rs
```diff
@@ -2743,23 +2729,139 @@ mod valid_solutions {
             // Set the latest block height.
             block_height = ledger.latest_height();
 
+            // Update the epoch solutions count.
+            if block_height % CurrentNetwork::NUM_BLOCKS_PER_EPOCH == 0 {
```
kaimast:
```suggestion
            if block_height.is_multiple_of (CurrentNetwork::NUM_BLOCKS_PER_EPOCH) {
```

Pretty sure this is a clippy warning in newer Rust versions, so it would be good to address these now.

ledger/src/lib.rs
kaimast: > If current_block_height=NUM_BLOCKS_PER_EPOCH, start=NUM_BLOCKS_PER_EPOCH+1, which seems odd if not wrong.

`(NUM_BLOCKS_PER_EPOCH % NUM_BLOCKS_PER_EPOCH) + 1 = 1`, right?

It may be good to add some helper functions for these conversions, though. Things like, `block_height_to_epoch` and `get_first_block_in_epoch`, could be unit tested and useful in other places.

raychu86: | `current_block_height=NUM_BLOCKS_PER_EPOCH`, `start=NUM_BLOCKS_PER_EPOCH+1`

The logic here is actually correct in that the loop doesn't execute, however I agree it's a bit confusing.

ledger/src/tests.rs
raychu86: Resolved with [6cca9ee](https://github.com/ProvableHQ/snarkVM/pull/3059/commits/6cca9eef8a0cd9fdacebfe34b69bba92f5681258)

ledger/src/lib.rs
raychu86: Cleaned up with [6cca9ee](https://github.com/ProvableHQ/snarkVM/pull/3059/commits/6cca9eef8a0cd9fdacebfe34b69bba92f5681258)

vicsn (approved):

Issue #3056, opened 2025-12-10
[Proposal] Progressively replace `anyhow` use in library crates with `thiserror`


mitchmindtree:
## 💥 Proposal

While beginning to address #2941 and investigating the feasibility of introducing `Result` in some locations, I've noticed that `anyhow::Result` is exposed from the `snarkvm_console_network_environment::prelude`, and is used throughout a lot of the library crates.

While `anyhow` can be convenient for not having to think about creating new error types, it is not designed for use within libraries or to be exposed in public APIs.

This is a proposal to remove `anyhow::Result` from the prelude, and to progressively replace the use of `anyhow` for error handling throughout the VM libraries with `thiserror`. I've noticed there is already some use of `thiserror` which is great!

## Motivation

The author of `thiserror` and `anyhow` distinguishes between the use-cases for the two crates briefly in the [`thiserror` README](https://github.com/dtolnay/thiserror?tab=readme-ov-file#comparison-to-anyhow):

> Use thiserror if you care about designing your own dedicated error type(s) so that the caller receives exactly the information that you choose in the event of failure. This most often applies to library-like code. Use [Anyhow](https://github.com/dtolnay/anyhow) if you don't care what error type your functions return, you just want it to be easy. This is common in application-like code.

Problems with `anyhow` in libraries:

- **Loss of type information:** Downstream code cannot match on specific error variants or exhaustively handle different failure modes. Accessing and handling underlying errors requires downcasting which can fail (e.g. can be a problem when refactoring).
- **Poor API boundaries:** `anyhow::Error` is essentially `Box<dyn Error>` - it's meant for applications that just propagate errors to users, not libraries that need to expose errors as part of their public API.
- **Documentation gaps:** Error variants aren't visible in generated docs, leaving users guessing about the set of possible failure modes. Ideally, a user can read the error type returned within the `Result` and see all possible failure modes for the API their calling.

Benefits of `thiserror` in libraries:

- Concrete error enums that can be matched and handled exhaustively in a type-safe manner (i.e. no downcasting required).
- Library users can see the full set of possible failure modes just by looking at the error type.
- Proper error composition through transparent wrapping of underlying errors.

Having nice error types would be great for consumers of the snarkvm libs (like `leo` tooling) so that we can handle different error variants accordingly (e.g. handling a VM halt vs some deployment failure).

## Implementation

This might look something like:

[...]

Issue #3055, opened 2025-12-09
[Feature] On `halt`, return an error (or some outcome type) with the failing Aleo instruction index


mitchmindtree:
## 🚀 Feature

In the case that execution halts early for some reason, the VM should return an error (or some outcome type) describing the instruction index at which the halt occurred.

## Motivation

This is required to provide nice feedback to users from `leo` tooling. Knowing the failing instruction index, we can use source mapping to provide nicer unit-testing errors.

Currently, the VM `panic!`s the host on `halt`, meaning we lose all type information and context about the error and risk conflating VM halts with actual implementation/logic bugs. See #2941.

Originally raised here:

- https://github.com/ProvableHQ/leo/issues/28992

## Implementation

- [ ] Address #2941.
- [ ] Ensure the returned `Error` (or similar exec outcome type) describes the Aleo instruction index.

Issue #2941, opened 2025-10-01
[Bug] Remove clear panic potentials in validator code paths


vicsn:
## 🐛 Bug Report

Search for `halt` in `console` and `circuit` to see the instances.

Potentially we also need to audit unwrap and other panicking code.

Only then can we remove/change try_vm_runtime.

mitchmindtree:
OK, I've been doing a deep dive into the situation here over the past day, and I am still gauging the scope of this.

## Problem: `halt`

The source of the vast majority of unnecessary `panic!`s of the host is `Environment::halt`.

https://github.com/ProvableHQ/snarkVM/blob/cf10c9df8092625b8cbc5163bc41ddf11b5992f7/console/network/environment/src/environment.rs#L60-L63

Every implementation of `Environment::halt` simply `panic!`s the host.

This is undesirable, as we lose all type-safety and context, and can only attempt to recover control flow using `catch_unwind` which was never intended to be used as an exception.

When we try to recover with `catch_unwind` we risk conflating VM halts with actual implementation/logic bugs. In some cases `panic!`s can occur in non-`UnwindSafe` locations, and so we cannot even recover with `catch_unwind`.

## Solution

Ideally, every callsite of `E::halt` would instead be returning a `Halt` error type or similar.

E.g. rather than `E::halt(msg)`, we likely want `return Err(Halt(msg));`.

This will require updating *many* function signatures that currently hide their fallibility to return a `Result<T, Halt>` instead. A quick grep shows hundreds of instances where we `E::halt` (panic!) throughout the VM.

Some fallible operations (like `Div` dividing by `0`) cannot return a `Result` as they are an implementation of an upstream trait (i.e. `std::ops::Div`). Ideally, we would never have been providing these implementations, and instead would provide a custom `CheckedDiv` trait or similar. However, as a temporary solution these implementations could simply continue to `panic!` directly for now.

## Implementation

I imagine this will look something like:

- [ ] Remove `Environment::halt`.
- [ ] Add a new `pub struct Halt(String);` error type to the `environment` mod.
- [ ] Replace calls to `E::halt(msg)` with `return Err(Halt(msg));`:
    - update the caller's function signature as necessary so that we can propagate the error.
    - If the call occurs within a upstream trait implementation whose fn signature cannot be changed, simply `panic!` directly instead.

@vicsn let me know your thoughts! I've begun digging into this, and will provide updates here as I run into other issues or steps we might need.

---

**`anyhow::Result`**

[...]

PR #3064, opened 2025-12-16
feat: On instruction failure, return an error with the failing aleo instruction index

mitchmindtree:
## Motivation

This partly addresses #3055, ensuring that we expose the failing Aleo instruction from the top-level VM methods.

This does _not_ address the halt=panic issue discussed in #2941, though should pave the way so that when we do introduce better error handling for halts, we have a clear path to propagate these errors up to the VM user.

## Overview

When an instruction fails, the new `StackExecError::Instruction` variant now captures:

1. Index in the function's instruction list
2. The formatted instruction (e.g., "add r0 r1 into r2" or "call foo.aleo/bar")
3. The failure reason

The error format is: "Instruction ({instruction_string}) at index {index} failed: {error}"

For nested calls, you get a "stack trace" of instruction contexts, e.g.

```
Stack execution failed:
  Instruction (call foo.aleo/bar) at index 3 failed:
    Call failed:
      Substack execution failed:
        Instruction (add r0 r1 into r2) at index 7 failed: {actual error}
```

This is propagated up through to the main public VM methods (authorize, execute, evaluate, deploy).

Where necessary, an `Anyhow` variant has been added to the new error types in order to keep the changes minimal and focused on our instruction context. Ideally, we can remove these catch-all `Anyhow` variants as we migrate errors over to dedicated types with `thiserror` (see #3056).

## Follow-up

- [ ] Address #2941, replacing `panic!` behaviour with error handling.

### DIFF
--- a/ledger/src/error.rs
+++ b/ledger/src/error.rs
@@ -1,0 +1,42 @@
+// Copyright (c) 2019-2025 Provable Inc.
+// This file is part of the snarkVM library.
+
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at:
+
+// http://www.apache.org/licenses/LICENSE-2.0
+
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+use snarkvm_synthesizer::{VmDeployError, VmExecError};
+use thiserror::Error;
+
+// NOTE: Many errors in this module temporarily contain `Anyhow` variants.
+// Remove these variants as we migrate errors to thiserror.
+
+/// Errors that may occur when creating a transfer transaction.
+#[derive(Debug, Error)]
+pub enum CreateTransferError {
+    /// VM execution failed.
+    #[error("VM execution failed: {0}")]
+    VmExec(#[from] VmExecError),
+    /// A temporary variant for type-erased anyhow errors.
+    #[error(transparent)]
+    Anyhow(#[from] anyhow::Error),
+}
+
+/// Errors that may occur when creating a deploy transaction.
+#[derive(Debug, Error)]
+pub enum CreateDeployError {
+    /// VM deployment failed.
+    #[error("VM deployment failed: {0}")]
+    VmDeploy(#[from] VmDeployError),
+    /// A temporary variant for type-erased anyhow errors.
+    #[error(transparent)]
+    Anyhow(#[from] anyhow::Error),
+}
--- a/synthesizer/process/src/error.rs
+++ b/synthesizer/process/src/error.rs
@@ -1,0 +1,177 @@
+// Copyright (c) 2019-2025 Provable Inc.
+// This file is part of the snarkVM library.
+
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at:
+
+// http://www.apache.org/licenses/LICENSE-2.0
+
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+use thiserror::Error;
+
+// NOTE: Many errors in this module temporarily contain `Anyhow` variants.
+// Remove these variants as we migrate errors to thiserror.
+
+/// Errors that may occur during process authorization.
+#[derive(Debug, Error)]
+pub enum ProcessAuthError {
+    /// Stack authorization failed.
+    #[error("Stack authorization failed: {0}")]
+    StackAuth(#[from] StackAuthError),
+    /// A temporary variant for type-erased anyhow errors.
+    #[error(transparent)]
+    Anyhow(#[from] anyhow::Error),
+}
+
+/// Errors that may occur during process evaluation.
+#[derive(Debug, Error)]
+pub enum ProcessEvalError {
+    /// Stack evaluation failed.
+    #[error("Stack evaluation failed: {0}")]
+    StackEval(#[from] StackEvalError),
+    /// A temporary variant for type-erased anyhow errors.
+    #[error(transparent)]
+    Anyhow(#[from] anyhow::Error),
+}
+
+/// Errors that may occur during process execution.
+#[derive(Debug, Error)]
+pub enum ProcessExecError {
+    /// Stack execution failed.
+    #[error("Stack execution failed: {0}")]
+    StackExec(#[from] StackExecError),
+    /// A temporary variant for type-erased anyhow errors.
+    #[error(transparent)]
+    Anyhow(#[from] anyhow::Error),
+}
+
+/// Errors that may occur during process deployment.
+#[derive(Debug, Error)]
+pub enum ProcessDeployError {
+    /// Stack execution failed during synthesis.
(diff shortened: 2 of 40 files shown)


mitchmindtree:
Ahh, looks like we might need to either pin `cargo-nextest` to an older version, or update our `rust-toolchain.toml` rust version.
```
info: syncing channel updates for '1.88-x86_64-unknown-linux-gnu'
info: latest update on 2025-06-26, rust version 1.88.0 (6b00bc388 2025-06-23)
info: downloading component 'cargo'
info: downloading component 'clippy'
info: downloading component 'rust-docs'
info: downloading component 'rust-std'
info: downloading component 'rustc'
info: downloading component 'rustfmt'
info: installing component 'cargo'
info: installing component 'clippy'
info: installing component 'rust-docs'
info: installing component 'rust-std'
info: installing component 'rustc'
info: installing component 'rustfmt'
    Updating crates.io index
error: cannot install package `cargo-nextest 0.9.115`, it requires rustc 1.89 or newer, while the currently active rustc version is 1.88.0
`cargo-nextest 0.9.114` supports rustc 1.88
```

vicsn:
> pin cargo-nextest to an older version

Pin it!

vicsn:
@mitchmindtree looks great, I'm very suprised that [semver-checks.sh](https://app.circleci.com/pipelines/github/ProvableHQ/snarkVM/15664/workflows/76dd586d-5387-4ada-8a66-d5a91c8dcb0d/jobs/781806?utm_campaign=vcs-integration-link&utm_medium=referral&utm_source=github-checks-link&utm_content=summary) in CI doesn't fail. Can you double check, are we unintendedly ignoring checking these parts of the code?

Could you also create draft PRs for the following if they're not too difficult, just to see if things still compile:
- SDK: https://github.com/ProvableHQ/sdk/blob/mainnet/wasm/Cargo.toml#L24
- snarkOS: https://github.com/ProvableHQ/snarkOS/blob/staging/Cargo.toml#L50
- Leo

mitchmindtree:
> Can you double check, are we unintendedly ignoring checking these parts of the code?

It looks like this is just one of the cases that cargo-semver-checks cannot currently catch. They list a few examples [here](https://github.com/obi1kenobi/cargo-semver-checks?tab=readme-ov-file#will-cargo-semver-checks-catch-every-semver-violation), one of which seems to be changes to generic types which is the main breaking change made in this PR (i.e. `anyhow::Result<T, anyhow::Error>` -> `anyhow::Result<T, SpecificError>`).

> Could you also create draft PRs for the following if they're not too difficult, just to see if things still compile

Will do!

mitchmindtree:
It looks like both the SDK and snarkOS are both behind staging already (the SDK moreso), and will require other unrelated changes aside from these as a part of updating their snarkVM dependency. We might want to handle those updates independently of this?

Leo will also need a dedicated PR to take advantage of these structured errors, though I think it makes sense to address #2941 first as this PR is barely useful to `leo` without also addressing #2941.

W.r.t. whether or not this is breaking: I think changing the function signatures like this is undoubtedly a breaking change for snarkVM as a library. Even if snarkOS, sdk and leo all happen to still compile due to anyhow's implicit error conversions, we could technically be breaking other downstream libs that depend on snarkVM that we don't know about, so it's probably safest to include this in a version bump.

vicsn:
snarkOS PR: https://github.com/ProvableHQ/snarkOS/pull/4051

As for SDK, @iamalwaysuncomfortable can you create a PR updating snarkVM to the latest staging branch?

ledger/src/error.rs
```diff
@@ -0,0 +1,42 @@
...
+
+/// Errors that may occur when creating a transfer transaction.
+#[derive(Debug, Error)]
+pub enum CreateTransferError {
+    /// VM execution failed.
+    #[error("VM execution failed: {0}")]
+    VmExec(#[from] VmExecError),
+    /// A temporary variant for type-erased anyhow errors.
+    #[error("{0}")]
+    Anyhow(#[from] anyhow::Error),
```
kaimast, line 30: Should we call this `Other` to indicate this a catch-all? I guess it doesn't really matter as it will hopefully be removed soon.

synthesizer/process/src/error.rs
```diff
@@ -0,0 +1,158 @@
...
+    #[error("{0}")]
+    Anyhow(#[from] anyhow::Error),
+}
+
+/// Errors that may occur during stack execution.
+#[derive(Debug, Error)]
+pub enum StackExecError {
+    /// Instruction at the given index failed.
+    #[error("Instruction ({1}) at index {0} failed: {2}")]
+    Instruction(usize, String, InstructionError),
```
kaimast: Should these be named fields, maybe?

kaimast:
```suggestion
    Instruction {
         index: usize,
         instruction: String,
         error: InstructionError,
     }
```

ledger/src/error.rs
```diff
@@ -0,0 +1,42 @@
...
+// Remove these variants as we migrate errors to thiserror.
+
+/// Errors that may occur when creating a transfer transaction.
+#[derive(Debug, Error)]
+pub enum CreateTransferError {
+    /// VM execution failed.
+    #[error("VM execution failed: {0}")]
+    VmExec(#[from] VmExecError),
+    /// A temporary variant for type-erased anyhow errors.
+    #[error("{0}")]
```
kaimast: This is a little cleaner:
```suggestion
    #[error(transparent)]
```

synthesizer/process/src/error.rs
mitchmindtree: Good call! I ended up adding a dedicated type as the variant appears in a couple places.

ledger/src/error.rs
mitchmindtree: Forgot about this attr, thanks!

mitchmindtree:
Thanks for the reviews! I think this should be good to go.

I don't think the CI errors here are related - one is the semver check, another is a circeci cargo-audit db bug.

How do we normally handle the semver check here? Do we normally bump the crate versions in the same PR? Or do we just use it as an indicator that a version bump is due in a follow-up PR?

vicsn:
We bump the semver commit rev in the .ci folder if we think a version bump is fine. Can you update it?

We version bump the crate before the actual release in a separate PR.

vicsn (approved):
LGTM! Delaying merge for a week so it goes into the subsequent release train

mitchmindtree:
OK I've rebased onto the latest `staging` and bumped `BASELINE_REV` :+1: 

> Delaying merge for a week so it goes into the subsequent release train

I'll let you land this when you're ready - in the meantime I'll open the follow-up error-handling PRs against this branch so they're easier to review.

kaimast:
One more suggestion I have is that should consider using the [`nonexhaustive attribute`](https://doc.rust-lang.org/reference/attributes/type_system.html#the-non_exhaustive-attribute) to avoid breaking the API when new error variants are introduced.

mitchmindtree:
Ah yep this did cross my mind too, though I do think there is a trade-off. It can be useful to exhaustively match against the error variants in some cases, and having the type-checker reveal new failure modes after an update can also be useful.

Perhaps we can selectively add it to the types where we foresee a lot of changes. E.g. In a WIP PR I plan to introduce an error type that covers the different possible `instruction.evaluate()` failures. This could be a good candidate as we know we plan to add new instructions going forward (like a verify op). Though I guess these kinds of changes would likely be breaking changes anyway :sweat_smile:

Maybe we can revisit the potential for `#[nonexhaustive]` when we get around to removing the temporary `Anyhow` variants, as we know this will require a technically breaking change anyway?

mitchmindtree:
> in the meantime I'll open the follow-up error-handling PRs against this branch so they're easier to review.

I just realised I cannot do this as I originally opened this PR from my fork (before I had push permissions to this repo) :woozy_face: 

I'm going to close this PR, then re-open it from a branch under this repo. This will make it easier to base the follow-up PRs on this branch.

mitchmindtree:
Closing in favour of #3081

PR #3066, opened 2025-12-29
fix: public index check in R1CS converter

SashaMalysehko:
The mapping between the environment R1CS and snarkvm_algorithms::r1cs already treats public indices consistently: index 0 is the constant one, and real public inputs use the same indices 1..n in both systems. In R1CS::generate_constraints we were asserting Index::Public(index + 1) when translating linear combinations, while the converter map and the Assignment-based converter both use Index::Public(index) directly. This off-by-one check was incorrect and could cause spurious assertion failures when public variables appear in constraint terms. The change removes the +1 so that the assertion matches the actual mapping and stays consistent with the Assignment converter.

### DIFF
--- a/circuit/environment/src/helpers/converter.rs
+++ b/circuit/environment/src/helpers/converter.rs
@@ -147,7 +147,7 @@ impl<F: PrimeField> R1CS<F> {
                                 let (index, _value) = index_value.as_ref();
                                 let gadget = converter.public.get(index).unwrap();
                                 assert_eq!(
-                                    snarkvm_algorithms::r1cs::Index::Public((index + 1) as usize),
+                                    snarkvm_algorithms::r1cs::Index::Public(*index as usize),
                                     gadget.get_unchecked(),
                                     "Failed during constraint translation. The public variable in the second system must match the first system (with an off-by-1 for the public case)"
                                 );


Antonio95:
Hi @SashaMalysehko! Thanks a lot for posting this, it does show some fine attention to detail 👌 

I've had a look and tentatively think the current code is correct, but I'm happy to discuss things with you. My thinking:

1. We're inside a loop `for (i, public) in self.to_public_variables().iter().skip(1).enumerate()`. `self.to_public_variables()` starts as (with some notational liberties) `[Variable::Public((0, F::1))]` (you can see this in `circuit/environment/src/helpers/r1cs.rs::new`) and subsequently gets other variables appended to it. Suppose it has the form `[Variable::Public((0, F::1)), Variable::Public(1, F::17)]` by the time we start the loop.
2. Because the loop does `enumerate` *after* `skip(1)`, the first tuple it iterates over is `(0, Variable::Public(1, F::17))`. This causes the first `assert_eq` (the one with `(i + 1) as u64`) to pass.
3. When the loop gets to `let gadget = cs.alloc_input(|| format!("Public {i}"), || Ok(*value))?`, `cs` simply creates a new public variable with the next available index. Since `cs` always starts with a public variable of index 0 (cf. for example the first check in the test), in the first iteration of the loop, the obtained gadget will be a variable with public index 1. This coincides with the `*index` in the second `assert_eq`, which is read from `(0, Variable::Public(1, F::17))`.

A more heuristic way to convince oneself everything is okay: surround the `assert_eq` the PR refers to (without your changes) with the following prints:
```
    println!(" - gadget.get_unchecked(): {:?}; i: {i}; index: {index}", gadget.get_unchecked());
    
    assert_eq!(
        snarkvm_algorithms::r1cs::Index::Public(*index as usize),
        gadget.get_unchecked(),
        "Public variables in the second system must match the first system (with an off-by-1 for the public case)"
    );
    
    println!("   assert_eq passes");
```

Then run `cargo test helpers::converter::tests::test_constraint_converter -- --nocapture` from the directory `snarkVM/circuit/environment`. It passes as expected.

What do you think? Am I missing anything?

Thanks again for delving into things

PR #3068, opened 2026-01-05
Ensure acyclic call graph

vicsn:
## Motivation

Ensures that cycles are not permitted in call graphs, as defense against any possible re-entrancy attack.

Targets the dynamic dispatch feature branch and release. Because cycles are not possible until this feature lands, technically the consensus version guard is perhaps not needed, but we will need such guards anyhow in the future if we decide to enable cycles.

## Test Plan

Added a few special case unit tests and running the cycle check against many VM tests.


synthesizer/process/src/verify_execution.rs
```diff
@@ -69,6 +70,15 @@ impl<N: Network> Process<N> {
 
         // Construct the call graph of the execution.
         let call_graph = self.construct_call_graph(execution.transitions())?;
+        // Ensure the constructed call graph is acyclic from ConsensusVersion::V14 onwards.
+        if consensus_version >= ConsensusVersion::V14 {
```
d0cd, line 74: Should this check be moved to `check_transaction` before the caching the verification status?

synthesizer/process/src/trace/mod.rs
```diff
@@ -192,6 +193,17 @@ impl<N: Network> Trace<N> {
         // Store the global state root.
         self.global_state_root.set(global_state_root).map_err(|_| anyhow!("Failed to set global state root"))?;
 
+        // Ensure the constructed call graph is acyclic from ConsensusVersion::V14 onwards.
```
d0cd, line 196: Do we also need to check for cycles at this point?

vicsn: This is for the prover to get an early error so they don't have to wonder why their transaction doesn't get accepted.

Another option is to add this in `authorize` but that's designed to be stateless (i.e. not to query for the block height) so we can't guard this or future changes by consensus version.

synthesizer/process/src/verify_execution.rs
vicsn: Had to think about this for a second. 

We clear the `partially_verified_transaction` whenever we move up a new consensus version so we should be fine.

given that this part of the code already relies on `consensus_version` and is the only place where the verifier is already constructing the `call_graph`, it seems fine.

synthesizer/process/src/verify_execution.rs
```diff
@@ -703,4 +713,48 @@ impl<N: Network> Process<N> {
...
+        tid_to_locator: &IndexMap<N::TransitionID, Locator<N>>,
+    ) -> Result<()> {
+        // Determine the root of the call graph (the last transition in the execution, by construction).
+        let (root_tid, root_locator) = tid_to_locator.iter().last().ok_or_else(|| anyhow!("No root found."))?;
+        // Track the transition child indices to explore next.
+        let mut stack: Vec<(N::TransitionID, usize)> = vec![(*root_tid, 0)];
+        // Track the locators in the *current call path* (push on descent, remove on backtrack).
+        let mut active_locators: HashSet<Locator<N>> = HashSet::from([*root_locator]);
+        // Perform an explicit-stack DFS; encountering an already-active locator indicates a cycle.
+        while let Some((node, next_child_index)) = stack.pop() {
```
d0cd, line 735: It could be nice to implement this as a bounded DFS by the maximum number of transitions so that we can have tighter guarantees on termination.

synthesizer/tests/test_vm_execute_and_finalize.rs
```diff
@@ -237,25 +237,24 @@ fn run_test(test: &ProgramTest) -> serde_yaml::Mapping {
...
-                }
+            // Test cost computation for Authorization
+            if consensus_version >= ConsensusVersion::V4 {
+                let actual_cost = execution_cost(&vm.process().read(), execution, consensus_version).unwrap();
+                let authorization = Authorization::from_unchecked((vec![], execution.transitions().cloned().collect()));
+                let expected_cost =
+                    execution_cost_for_authorization(&vm.process().read(), &authorization, consensus_version).unwrap();
+                assert_eq!(actual_cost, expected_cost);
             }
+            //  Test that the call graph is valid.
```
d0cd: If the call graph check is placed in `check_transaction`, do we need to check it again in this test code?

synthesizer/process/src/verify_execution.rs
vicsn: Not an expert on different DFS algorithms, but I did a quick extra review and it seems like serialization of an `Execution` already enforces that we don't receive more than `MAX_TRANSITIONS` and that `construct_call_graph` maintains that invariant!

synthesizer/tests/test_vm_execute_and_finalize.rs
vicsn: Good catch! Apologies I force pushed this away but I recently learned there's a nift "Compare" button bottom right in the Github UI.

synthesizer/process/src/verify_execution.rs
d0cd: That's true! I tend to prefer the bounded iteration pattern for defense in depth. I'll leave it up to you.

d0cd:
Looks mostly good to me, thanks @vicsn!

I was thinking about use cases for re-entrancy and wanted to float for consideration.
Consider a DEX that allows users to register custom pre-swap hooks:
```
function swap:
      input r0 as address.public;    
      input r1 as u64.public;     
      input r2 as field.public;       
      input r3 as field.public;       
      
      // Call user's hook before swap executes
      call.dynamic r2 'aleo' r3 with r0 r1 (as address.public u64.public) into r4 (as dynamic.future);
      // Pass parameters to async context.
      // Output future.
 ```
Suppose that the user's hook was a function `hook` that liquidates another existing position in this dex by dynamically calling `swap` again with a no-op hook.
This wouldn't be allowed if the call graph must be acyclic.

This example is an interesting one that would be possible in Uniswap V4 but is not critical to basic DEX functionality.
That said, we should spend some time convincing ourselves that don't have any use for recursion.

File: synthesizer/process/src/tests/test_call_graph.rs [hunk 5]
```diff
@@ -0,0 +1,184 @@
...
+#[test]
+fn test_ensure_acyclic_call_graph_bails_on_cycle() {
+    // A -> B -> B
+    // This represents a recursive call cycle by function identity (even though transition IDs differ).
+    let a = tid(1);
+    let b_0 = tid(2);
+    let b_1 = tid(3);
+
+    let call_graph: HashMap<_, Vec<_>> = HashMap::from([(a, vec![b_0]), (b_0, vec![b_1]), (b_1, vec![])]);
+    let tid_to_locator: IndexMap<_, _> = IndexMap::from([
```
Antonio95, line 58: I think the order should be `b_1`, `b_0`, `a`: we do post-order graph exploration (children subtrees from left to right, then root), right?

This doesn't affect the test because the only aspect of the order of `tid_to_locator` in `ensure_acyclic_call_graph` is the fact that the root is last.

synthesizer/process/src/tests/test_call_graph.rs
```diff
@@ -0,0 +1,184 @@
...
+        (a, locator("foo.aleo", "a")),
+    ]);
+
+    let err =
+        Process::<CurrentNetwork>::ensure_acyclic_call_graph(&call_graph, &tid_to_locator).unwrap_err().to_string();
+    assert!(err.contains("Cycle detected"), "Unexpected error: {err}");
+}
+
+#[test]
+fn test_ensure_acyclic_call_graph_accepts_single_node() {
```
Antonio95, line 70: Would a test for the "no root found" case (which is indeed controlled) also be relevant?

synthesizer/process/src/tests/test_call_graph.rs
```diff
@@ -0,0 +1,184 @@
...
+    // A -> {B, C}; B -> D; C -> D
+    let a = tid(1);
+    let b = tid(2);
+    let c = tid(3);
+    let d_0 = tid(4);
+    let d_1 = tid(5);
+
+    let call_graph: HashMap<_, Vec<_>> =
+        HashMap::from([(a, vec![b, c]), (b, vec![d_0]), (c, vec![d_1]), (d_0, vec![]), (d_1, vec![])]);
+    let tid_to_locator: IndexMap<_, _> = IndexMap::from([
```
Antonio95, line 88: Here the order is: `d_0`, `b`, `d_1`, `c`, `a` (but cf. earlier comment)

synthesizer/process/src/tests/test_call_graph.rs
```diff
@@ -0,0 +1,184 @@
...
+
+    let call_graph: HashMap<_, Vec<_>> = HashMap::from([
+        (a, vec![b, c]),
+        (b, vec![d_0]),
+        (c, vec![d_1]),
+        (d_1, vec![d_2]),
+        (d_0, vec![]),
+        (d_2, vec![]),
+    ]);
+    let tid_to_locator: IndexMap<_, _> = IndexMap::from([
```
Antonio95, line 175: Order: `d_0`, `b`, `d_2`, `d_1`, `c`, `a`

synthesizer/process/src/tests/test_call_graph.rs
```diff
@@ -0,0 +1,184 @@
...
+        (d_2, locator("foo.aleo", "d")),
+        (c, locator("foo.aleo", "c")),
+        (b, locator("foo.aleo", "b")),
+        (a, locator("foo.aleo", "a")),
+    ]);
+
+    let err =
+        Process::<CurrentNetwork>::ensure_acyclic_call_graph(&call_graph, &tid_to_locator).unwrap_err().to_string();
+    assert!(err.contains("Cycle detected"), "Unexpected error: {err}");
+}
```
Antonio95, line 187: Here are two missing test cases which I think are interesting

1) Check that the function does not blindly "flatten" the call graph to a mapping of locators, for instance: A -> (B, C); B -> C, C -> B should indeed pass as cycle-free (even though a "flattened" mapping would think there is a cycle B -> C -> B)

2) Check that the function detects cycles of length > 1

Here's the code for both of them. They pass:

```
#[test]
fn test_ensure_acyclic_call_graph_distinguishes_paths() {
    // A -> {B, C}; B -> C; C -> B
    let a = tid(1);
    let b_0 = tid(2);
    let c_0 = tid(3);
    let b_1 = tid(4);
    let c_1 = tid(5);

    let call_graph: HashMap<_, Vec<_>> = HashMap::from([(a, vec![b_0, c_0]), (b_0, vec![c_1]), (c_0, vec![b_1]), (b_1, vec![]), (c_1, vec![])]);

    let tid_to_locator: IndexMap<_, _> = IndexMap::from([
        (c_1, locator("foo.aleo", "c")),
        (b_0, locator("foo.aleo", "b")),
        (b_1, locator("foo.aleo", "b")),
        (c_0, locator("foo.aleo", "c")),        
        (a, locator("foo.aleo", "a")),
    ]);

    Process::<CurrentNetwork>::ensure_acyclic_call_graph(&call_graph, &tid_to_locator).unwrap();
}

#[test]
fn test_ensure_acyclic_call_graph_detects_longer_cycles() {
    // A1 -> {B0, C0}; B0 -> (C1, D0); D0 -> A1; C0 -> B1
    let a_0 = tid(1);
    let a_1 = tid(2);
    let b_0 = tid(3);
    let b_1 = tid(4);
    let c_0 = tid(5);
    let c_1 = tid(6);
    let d_0 = tid(7);

    let call_graph: HashMap<_, Vec<_>> = HashMap::from(
        [
            (a_0, vec![b_0, c_0]),
            (b_0, vec![c_1, d_0]),
            (b_1, vec![]),
            (c_0, vec![b_1]),
            (c_1, vec![]),
            (d_0, vec![a_1]),
            (a_1, vec![]),
        ]
    );

    let tid_to_locator: IndexMap<_, _> = IndexMap::from([
        (c_1, locator("foo.aleo", "c")),
        (a_1, locator("foo.aleo", "a")),
        (d_0, locator("foo.aleo", "d")),
        (b_0, locator("foo.aleo", "b")),
        (b_1, locator("foo.aleo", "b")),
        (c_0, locator("foo.aleo", "c")),        
        (a_0, locator("foo.aleo", "a")),
    ]);

    let err = Process::<CurrentNetwork>::ensure_acyclic_call_graph(&call_graph, &tid_to_locator).unwrap_err().to_string();
    assert!(err.contains("Cycle detected"), "Unexpected error: {err}");
}
```

synthesizer/process/src/verify_execution.rs
```diff
@@ -703,4 +713,48 @@ impl<N: Network> Process<N> {
...
+            let node_locator = tid_to_locator.get(&node).ok_or_else(|| anyhow!("Missing locator for '{node}'"))?;
```
Antonio95: This is only used in the `else` arm, I would move it there for clarity

Antonio95 (approved):
Excellent stuff! I've had a close look at the algorithm and it seems solid. Left only a couple of minor comments and a couple of extra test cases.

vicsn:
@copilot are you capable of fixing all of Antonio's review comments?

Antonio95 (approved):

Antonio95:
I've seen one of the latest commits added a comment to the `mod recursion` line in `test_v14.rs`. This PR should delete that module (since it disallows the tested functionality), correct?

rot256 (approved):
Looks correct. Played around with an alternative implementation, but to be honest, don't know if it ended up more readable:

```rust
pub fn ensure_acyclic_call_graph(
    call_graph: &HashMap<N::TransitionID, Vec<N::TransitionID>>,
    tid_to_locator: &IndexMap<N::TransitionID, Locator<N>>,
) -> Result<()> {
	enum Op<N: Network> {
		Visit(N::TransitionID),
		Remove(Locator<N>),
	}
	// Determine the root of the call graph (the last transition in the execution, by construction).
	let (root_tid, _) = tid_to_locator.iter().last().ok_or_else(|| anyhow!("No root found."))?;
	// Track the transition child indices to explore next.
	let mut stack: Vec<Op<N>> = vec![Op::Visit(*root_tid)];
	// Track the locators in the *current call path* (push on descent, remove on backtrack).
	let mut active_locators: HashSet<Locator<N>> = HashSet::new();
	// Perform an explicit-stack DFS; encountering an already-active locator indicates a cycle.
	while let Some(op) = stack.pop() {
		match op {
			Op::Visit(node) => {
				let loc = *tid_to_locator.get(&node).ok_or_else(|| anyhow!("Missing locator for '{node}'"))?;
				if !active_locators.insert(loc) {
					return Err(anyhow!("Cycle detected at '{loc}'"));
				}
				let children = call_graph.get(&node).ok_or_else(|| anyhow!("Missing node '{node}'"))?;
				stack.push(Op::Remove(loc));
				stack.extend(children.iter().map(|&child| Op::Visit(child)));
			}
			Op::Remove(loc) => {
				assert!(active_locators.remove(&loc));
			}
		}
	}
	Ok(())
}
```

vicsn:
> This PR should delete that module (since it disallows the tested functionality), correct?

I commented out the import, so it's effectively dead code

PR #3070, opened 2026-01-05
[Fix] Make `prepare_next_quorum_block` atomic

kaimast:
This will avoid certain race conditions when preparing a block template. 

Concretely, in the current design the following is possible:
* There is a thread t that calls `Ledger::prepare_next_beacon_block` which reads the value of `current_block`
* There is another thread t' that calls `Ledger::advance_to_next_block`, which updates `current_block` and the state root.
* If t now continues execution and reads the state root, it will return a block and not an error, but the block will be invalid, i.e. the state root is incorrect.

This is problematic for two reasons:
* `Ledger` implements `Sync`, which implies concurrent calls to its functions should be executed atomically and not read partial/inconsistent data.
* Such invalid blocks are harder to catch on the snarkOS side, as they appear to have the correct height and inconsistencies in the state root will be detected only later when the block is executed.

While this PR will ensure calls to the `Ledger` are atomic (formally, the Ledger being _linearizable_), it does not ensure that multiple calls to these functions do not interleave in unintended ways (formally, _serializable_). The latter is implemented in snarkOS and out of the scope of snarkVM, while the former makes the implementation for in snarkOS a lot easier. 

P.S.: There is this [great blog post by Peter Bailis](http://www.bailis.org/blog/linearizability-versus-serializability/) describing the two formal guarantees mentioned in the previous paragraph, in case you want to learn more about the formalism behind it. 

### Other minor changes
The PR also reduces log verbosity for sequential operations, as they add a decent amount of noise otherwise.

### DIFF
--- a/ledger/src/check_next_block.rs
+++ b/ledger/src/check_next_block.rs
@@ -49,6 +49,8 @@ pub enum CheckBlockError<N: Network> {
     BlockAlreadyExists { hash: N::BlockHash },
     #[error("Block has invalid height. Expected {expected}, but got {actual}")]
     InvalidHeight { expected: u32, actual: u32 },
+    #[error("Block has invalid round. Was {new}, but must be greater than previous round ({previous})")]
+    InvalidRound { new: u64, previous: u64 },
     #[error("Block has invalid hash")]
     InvalidHash,
     /// An error related to the given prefix of pending blocks.
@@ -206,6 +208,10 @@ impl<N: Network, C: ConsensusStorage<N>> Ledger<N, C> {
         // Ensure, again, that the ledger has not advanced yet. This prevents cryptic errors form appearing during the block check.
         if block.height() != latest_block.height() + 1 {
             return Err(CheckBlockError::InvalidHeight { expected: latest_block.height() + 1, actual: block.height() });
+        }
+        // Also ensure the round is valid, otherwise speculation on transactions will fail with a cryptic error.
+        if block.round() <= latest_block.round() {
+            return Err(CheckBlockError::InvalidRound { new: block.round(), previous: latest_block.round() });
         }
 
         // Ensure the solutions do not already exist.
--- a/ledger/src/check_next_block.rs
+++ b/ledger/src/check_next_block.rs
@@ -148,7 +148,7 @@ impl<N: Network, C: ConsensusStorage<N>> Ledger<N, C> {
         self.check_block_subdag_quorum(block)?;
 
         // Determine if the block subdag is correctly constructed and is not a combination of multiple subdags.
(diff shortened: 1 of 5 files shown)


vicsn:
Can you explain in the PR README with which test setup you saw what issue come up and how you validated it is resolved?

Can you try to run a network while calling `GET /<network>/block/latest` on repeat?

All existing calls to `latest_block()` in snarkVM and snarkOS seem to happen outside of the block generation path, so may incur some delay but no obvious deadlock.

ljedrz (commented):
It's not clear to me how this fixes things, as `prepare_advance_to_next_quorum_block` appears to only reuse the block obtained at the beginning.

If I'm not mistaken, the specific intent is to make it impossible for concurrent calls to `try_advance_to_next_block` to have a different `previous_block`; however, I'd argue that those calls should not happen concurrently at all, which this change doesn't prevent, as multiple concurrent readers are permitted.

kaimast:
> If I'm not mistaken, the specific intent is to make it impossible for concurrent calls to `try_advance_to_next_block` to have a different `previous_block`; however, I'd argue that those calls should not happen concurrently at all, which this change doesn't prevent, as multiple concurrent readers are permitted.

That is actually not the goal. I simply do not want any concurrent updates or reads to the ledger *while* the function is executing. I added some more comments to the code and will update the PR description in a second.
At a high-level my reasoning is that `Ledger` is `Sync`, which implies that it can be called to from concurrent threads, without ending up in a corrupted state, but it is possible to create invalid block templates if there are certain concurrent invocations in the current design.

ljedrz (approved):
Ah, I initially missed the broader context.

LGTM; a read guard should suffice to provide the desired guarantees, while not being too restrictive for any concurrent work. In the future we might want to reorganize the related logic so that mistakes are less likely to happen due to the structure itself, but it's a step in the right direction, and it's well-documented.

vicsn:
Let's first finish https://github.com/ProvableHQ/snarkOS/pull/4069 before merging

ledger/src/check_next_block.rs
raychu86: Is there a deadlock condition here? `latest_timestamp` also tries to get the read lock. It should be handled the same way `latest_epoch_hash` is fetched.

raychu86: `latest_timestamp()` is also used in `Ledger::num_remaining_solutions`, which may have a conflicting .read() as well.

kaimast:
All tests for https://github.com/ProvableHQ/snarkOS/pull/4096 pass, so I am merging this PR.

ledger/src/check_next_block.rs
kaimast: Fixed

PR #3071, opened 2026-01-06
[Feature] Fallback URL for parameter files

kaimast:
The PR adds another URL to fetch parameters from. It also includes a few small commits to fix CI. 

### Testing

1. Rename the first entry of `REMOTE_URLS` in `parameters/src/canary.rs` to something like `broken.provable.com`.
2. Remove any cached parameter files `rm -rf ~/.aleo/resources/*` (double check that this is the correct location for your setup)
3. Run the test without output capture: `cargo test --package=snarkvm-parameters canary::test_credits_bond_public_prover -- --nocapture`

You will get the following output:
```
running 1 test

⚠️  "bond_public.prover.e2c9caa" does not exist. Downloading and storing it (in "/Users/kaimast/.aleo/resources/bond_public.prover.e2c9caa").

   Installation - Downloading "https://broken.provable.com/canary/bond_public.prover.e2c9caa"
        Warning - curl::error: Couldn't resolve host name
        Warning - Failed to fetch from "https://broken.provable.com/canary". Trying next source...
   Installation - Downloading "https://s3.us-west-1.amazonaws.com/canary.parameters/bond_public.prover.e2c9caa"
   Installation - 100.00% complete (27 MB total)   Installation - Storing file in "/Users/kaimast/.aleo/resources/bond_public.prover.e2c9caa"
test canary::test_credits_bond_public_prover ... ok

test result: ok. 1 passed; 0 failed; 0 ignored; 0 measured; 220 filtered out; finished in 1.46s
```

### DIFF
--- a/parameters/src/macros.rs
+++ b/parameters/src/macros.rs
@@ -83,6 +83,10 @@ macro_rules! impl_store_and_remote_fetch {
 
                 easy.progress(true)?;
                 easy.progress_function(|total_download, current_download, _, _| {
+                    // avoid division by zero.
+                    if total_download == 0.0 {
+                        return true;
+                    }
                     let percent = (current_download / total_download) * 100.0;
                     let size_in_megabytes = total_download as u64 / 1_048_576;
                     let output =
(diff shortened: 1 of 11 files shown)


parameters/src/macros.rs
```diff
@@ -232,21 +274,45 @@ macro_rules! impl_load_bytes_logic_remote {
...
+                        }
+
+                        // If all URLs failed, return the last error.
+                        if let Some(e) = last_error {
+                            return Err(e);
+                        }
 
                         // Ensure the size matches.
                         if $expected_size != buffer.len() {
-                            remove_file!(file_path);
```
vicsn, line 240: Did you remove this on purpose? (I'll note I've always found it weird)

vicsn: Ah I see it's because of WASM and the macro also has a WASM guard

vicsn (approved):

PR #3075, opened 2026-01-07
[Feature] Introduce a new variant of `Deployment`s to provide a pathway to update program VKs.

d0cd:
## TODO
- [x] `snarkOS PR` with new API endpoints. [link](https://github.com/ProvableHQ/snarkOS/pull/4067)
- [ ] Documentation (including API).
- [ ] Deployment cannon. [link](https://github.com/ProvableHQ/snarkOS-staging/pull/78). Needs to support record translation keys.

## To Reviewers
- Consider whether this mechanism is sufficient for other forms of VK upgrades. Could it have been used for the Varuna V2 change? Could it have been used for the record commitment change?
- Is it sufficient to use `deployment_tree_v2` for V3 deployments?

## Motivation

This PR introduces `DeploymentVersion::V3` (amendment deployments), a new deployment variant that allows updating verifying keys (VKs) for an already-deployed program, without updating the program edition and owner or running the constructor. These amendment deployments should be allowed to be submitted by anyone.

This capability is essential for the "dynamic dispatch" feature, which requires that existing programs update their VKs to include record translation keys. In the future, this could be used to swap out VKs if the circuit or proof system is updated.

#### New Deployment Version (V3)
- **V3 deployments** have a program checksum but **no program owner** in the `Deployment` struct
- V3 deployments are only allowed starting from `ConsensusVersion::V14`

#### Storage Layer Changes (`ledger/store/src/transaction/deployment.rs`)
Added 6 new maps to track amendments:
- `amendment_count_map`: Tracks how many amendments exist for each (program_id, edition)
- `amendment_id_map`: Maps (program_id, edition, amendment_index) to transaction_id
- `reverse_amendment_id_map`: Maps transaction_id to (program_id, edition, amendment_index)
- `amendment_verifying_key_map`: Stores VKs for amendments
- `amendment_certificate_map`: Stores certificates for amendments
- `amendment_owner_map`: Stores the program owner (from the `Transaction` not the `Deployment`) who submitted each amendment

#### Store Consistency Guarantees
- **LIFO Amendment Removal**: Amendments must be removed in reverse order (latest first)
- **Base Protection**: Base deployments cannot be removed while amendments exist

#### Verification Changes (`synthesizer/src/vm/verify.rs`)
- V3 deployments are rejected before `ConsensusVersion::V14`
- V3 deployments must target an existing program with matching program, checksum, and edition
- `credits.aleo` cannot be amended (security restriction)

#### Finalization Changes (`synthesizer/process/src/finalize.rs`)
- V3 deployments reuse the existing stack's program and edition
- V3 deployments update the VKs in the stack but don't initialize new mappings or run the constructor
- Fee finalization is properly handled for V3 deployments

#### Process Changes (`synthesizer/process/src/verify_deployment.rs`)
- V3 amendment verification uses `Stack::new_raw()` to preserve the existing edition rather than incrementing it

#### Block Changes (`ledger/block/src/transaction/deployment/mod.rs`)
- Added `set_edition_raw()` method for setting deployment edition during V3 creation

### New Methods on `Ledger`

| Method | Description |
|--------|-------------|
| `find_original_transaction_id_from_program_id_and_edition(program_id, edition)` | Returns the base deployment transaction ID (not amendments) |
| `find_latest_transaction_id_from_program_id_and_edition(program_id, edition)` | Returns the latest amendment transaction ID if amendments exist, otherwise the original deployment |
| `find_transaction_id_from_program_id_edition_and_amendment(program_id, edition, amendment_index)` | Returns the transaction ID for a specific amendment |

### New Methods on `TransactionStore` / `DeploymentStore`

| Method | Description |
|--------|-------------|
| `get_base_verifying_key(program_id, function_name, edition)` | Returns the original deployment VK (ignores amendments) |
| `get_base_certificate(program_id, function_name, edition)` | Returns the original deployment certificate (ignores amendments) |
| `get_amendment_count(program_id, edition)` | Returns the number of amendments for a program/edition |
| `is_amendment(transaction_id)` | Checks if a transaction is a V3 amendment |
| `get_amendment_info(transaction_id)` | Returns (program_id, edition, amendment_index) for an amendment |
| `get_verifying_key_for_amendment(program_id, function_name, edition, amendment_index)` | Returns the VK for a specific amendment |
| `get_certificate_for_amendment(program_id, function_name, edition, amendment_index)` | Returns the certificate for a specific amendment |
| `get_deployment_for_amendment(program_id, edition, amendment_index)` | Returns the deployment for a specific amendment |
| `amendment_transaction_ids(program_id, edition)` | Iterator over all amendment transaction IDs |
| `amendment_verifying_keys(program_id, edition)` | Iterator over all amendment VKs |
| `amendment_certificates(program_id, edition)` | Iterator over all amendment certificates |
| `find_original_transaction_id_from_program_id_and_edition(program_id, edition)` | Returns base deployment transaction ID |
| `find_latest_transaction_id_from_program_id_and_edition(program_id, edition)` | Returns latest amendment or base deployment transaction ID |
| `find_transaction_id_from_program_id_edition_and_amendment(program_id, edition, amendment_index)` | Returns the transaction ID for a specific amendment |

### Updated Behavior

| Method | Old Behavior | New Behavior |
|--------|--------------|--------------|
| `get_verifying_key(program_id, function_name, edition)` | Returns deployed VK | Returns latest amendment VK if exists, otherwise base VK |
| `get_certificate(program_id, function_name, edition)` | Returns deployed certificate | Returns latest amendment certificate if exists, otherwise base |

## Test Plan

Run the deployment store tests:
```bash
cargo test -p snarkvm-ledger-store test_insert_get_remove
cargo test -p snarkvm-ledger-store test_amendment_insert_get_remove
cargo test -p snarkvm-ledger-store test_multiple_sequential_amendments
cargo test -p snarkvm-ledger-store test_find_transaction_id
cargo test -p snarkvm-ledger-store test_get_latest_vk_with_amendments
cargo test -p snarkvm-ledger-store test_cannot_remove_base_with_amendments
cargo test -p snarkvm-ledger-store test_cannot_remove_non_latest_amendment
```

Run serialization tests:
```bash
cargo test -p snarkvm-ledger-block deployment::bytes
cargo test -p snarkvm-ledger-block deployment::serialize
```

Run the integration tests:
```bash
cargo test -p snarkvm-synthesizer test_v14:: --features test
```

### Integration Test Coverage (`test_v14.rs`)

| Test | Description |
|------|-------------|
| `test_v3_deployment_basic` | Basic V3 amendment creation and verification |
| `test_v3_deployment_verify_and_finalize` | Full verification and finalization flow |
| `test_v3_deployment_multiple_amendments` | Sequential amendments on same program |
| `test_v3_deployment_remove_amendments` | Amendment removal lifecycle |
| `test_v3_deployment_execution_uses_latest_vk` | Execution uses amended VK |
| `test_v3_deployment_requires_existing_program` | Validation rejects amendments for non-existent programs |

## Backwards Compatibility

| Scenario | Behavior |
|----------|----------|
| Existing V1 deployments | Unchanged - continue to work |
| Existing V2 deployments | Unchanged - continue to work |
| V3 before ConsensusVersion::V14 | Rejected during verification |
| V3 at ConsensusVersion::V14+ | Allowed for non-credits programs |
| `credits.aleo` V3 amendment | Always rejected (security) |

## Migration Details
We plan to update existing programs by deploying amendments. Below are some stats on the new deployments that need to be issued for each network.

| Network | Programs | Max Size | Max Constraints | Max Variables | Max (C+V) | Total Cost |
|---------|----------|----------|-----------------|---------------|-----------|------------|
| Mainnet | 74 | 119,635 bytes | 1,056,261 | 1,276,349 | 2,332,610 | 1,358,565,071 microcredits |
| Testnet | 1,058 | 119,635 bytes | 1,152,369 | 1,430,362 | 2,582,731 | 9,029,801,711 microcredits |
| Canary | 2 | 28,499 bytes | 400,684 | 563,752 | 964,161 | 43,215,115 microcredits |


ledger/block/src/transaction/deployment/mod.rs
```diff
@@ -225,28 +225,29 @@ impl<N: Network> Deployment<N> {
...
-// An internal enum to represent the deployment version.
+/// The deployment version.
 #[derive(Copy, Clone, Eq, PartialEq)]
-pub(super) enum DeploymentVersion {
+pub enum DeploymentVersion {
+    /// A deployment without a program checksum or program owner.
     V1 = 1,
+    /// A deployment with both a program checksum and program owner.
     V2 = 2,
+    /// A deployment with a program checksum but without a program owner (for amendments).
```
vicsn: Neat!

ledger/store/src/helpers/memory/transaction.rs
```diff
@@ -111,6 +111,18 @@ pub struct DeploymentMemory<N: Network> {
...
+    amendment_count_map: MemoryMap<(ProgramID<N>, u16), u64>,
+    /// The amendment ID map.
+    amendment_id_map: MemoryMap<(ProgramID<N>, u16, u64), N::TransactionID>,
+    /// The reverse amendment ID map.
+    reverse_amendment_id_map: MemoryMap<N::TransactionID, (ProgramID<N>, u16, u64)>,
+    /// The amendment verifying key map.
+    amendment_verifying_key_map: MemoryMap<(ProgramID<N>, Identifier<N>, u16, u64), VerifyingKey<N>>,
+    /// The amendment certificate map.
+    amendment_certificate_map: MemoryMap<(ProgramID<N>, Identifier<N>, u16, u64), Certificate<N>>,
+    /// The amendment owner map.
```
vicsn: What is this used for?

vicsn (approved):
Marvellous, did not find any issues.

I think deployment_tree_v2 is fine.

VarunaVersion::V2 did not change the VK, so would not apply to this setup. The update to the credits program could have used this setup, but then again it is a single set of VKs hardcoded (and I think was fetched from CDN?)

ledger/block/src/transaction/deployment/mod.rs
```diff
@@ -211,42 +211,47 @@ impl<N: Network> Deployment<N> {
...
+    /// Returns the deployment version.
+    /// This function checks that the deployment checksum and owner is well-formed.
+    pub fn version(&self) -> Result<DeploymentVersion> {
         match (self.program_checksum.is_some(), self.program_owner.is_some()) {
             (false, false) => Ok(DeploymentVersion::V1),
             (true, true) => Ok(DeploymentVersion::V2),
-            (true, false) => {
-                bail!("The program checksum is present, but the program owner is absent.")
-            }
+            (true, false) => Ok(DeploymentVersion::V3),
```
raychu86, line 341: Would it be better to actually store the `DeploymentVersion` enum in the `Deployment` itself? Then we don't have to mux like this. Especially if there are more deployment versions in the future.

(We don't do this for `VarunaVersion`, but just a thought).

ledger/store/src/transaction/deployment.rs
```diff
@@ -65,6 +65,18 @@ pub trait DeploymentStorage<N: Network>: Clone + Send + Sync {
     type CertificateMap: for<'a> Map<'a, (ProgramID<N>, Identifier<N>, u16), Certificate<N>>;
     /// The fee storage.
     type FeeStorage: FeeStorage<N>;
+    /// The mapping of `(program ID, edition)` to the number of amendments.
+    type AmendmentCountMap: for<'a> Map<'a, (ProgramID<N>, u16), u64>;
+    /// The mapping of `(program ID, edition, amendment index)` to `transaction ID`.
+    type AmendmentIDMap: for<'a> Map<'a, (ProgramID<N>, u16, u64), N::TransactionID>;
```
raychu86: Nit: Amendment count vs Amendment index.

Maybe `AmendmentIndexesMap` or something so that we know that the "count" is the latest value to use for "amendment index" in the `AmendmentIDMap`

synthesizer/src/vm/verify.rs
```diff
@@ -282,39 +284,140 @@ impl<N: Network, C: ConsensusStorage<N>> VM<N, C> {
...
+                        // Note: This is safe to check for programs deployed before `ConsensusVersion::V8` because `credits.aleo/upgrade` was not yet introduced.
+                        deployment.program().check_external_calls_to_credits_upgrade()?;
                     }
-                    new_edition => {
+                    DeploymentVersion::V3 => {
+                        // For `V3` (amendment) deployments, check that:
+                        // - The program is not `credits.aleo`.
+                        // - The program already exists in the store and process.
+                        // - The existing program, checksum, and edition matches the one in the deployment.
+
```
raychu86: How can we prevent DOS from spamming amendments? In theory, we are protected by fee payments, but now any malicious party can add a billion amendments to every program. 

It's not exactly like spamming the network with new deployments, because this affects on-chain programs already. 

Ideally we have some sort of gate. Some options:
1. Only valid if it creates or updates VKs compared to the previous edition
2. Only certain signers can use V3 deployments
3. Only valid for specific allowed programs

vicsn: 1. Could be implemented by comparing the vk.id. Only if the vk is valid and yet the id is different, it is a valid amendment. This does incur the expensive synthesis, and would constitute a similar DoS risk as attackers specifying a wrong number of variables inside of their vk. Long term we plan to close this hole by taking Fees from such deployments.

ledger/store/src/helpers/memory/transaction.rs
```diff
@@ -111,6 +111,18 @@ pub struct DeploymentMemory<N: Network> {
...
+    amendment_owner_map: MemoryMap<(ProgramID<N>, u16, u64), ProgramOwner<N>>,
```
raychu86: Might be cleaner to do this via a dedicated `AmendmentStore`.

ledger/store/src/helpers/rocksdb/transaction.rs
```diff
@@ -121,6 +121,18 @@ pub struct DeploymentDB<N: Network> {
...
+    /// The amendment ID map.
+    amendment_id_map: DataMap<(ProgramID<N>, u16, u64), N::TransactionID>,
+    /// The reverse amendment ID map.
+    reverse_amendment_id_map: DataMap<N::TransactionID, (ProgramID<N>, u16, u64)>,
+    /// The amendment verifying key map.
+    amendment_verifying_key_map: DataMap<(ProgramID<N>, Identifier<N>, u16, u64), VerifyingKey<N>>,
+    /// The amendment certificate map.
+    amendment_certificate_map: DataMap<(ProgramID<N>, Identifier<N>, u16, u64), Certificate<N>>,
+    /// The amendment owner map.
+    amendment_owner_map: DataMap<(ProgramID<N>, u16, u64), ProgramOwner<N>>,
```
raychu86, line 135: I realize the `ProgramOwner` here is for the `Transaction::Deployment(.., owner, ...)` owner rather than the `program_owner` inside `Deployment`. 

But, this is a bit confusing that it uses `ProgramOwner`; because it's technically not the actual program owner, but the creator of the amendment. 

Logic is safe as is right now because it's similar to the V1 deployments, where the actual address can be hijacked, but can't be used in any way.

ledger/store/src/transaction/deployment.rs
```diff
@@ -208,133 +274,243 @@ pub trait DeploymentStorage<N: Network>: Clone + Send + Sync {
...
         let program_id = *program.id();
         // Retrieve the checksum.
         let checksum = deployment.program_checksum();
+        // Check if this is an amendment.
+        let is_amendment = deployment.version()? == DeploymentVersion::V3;
+
+        // Handle amendments separately from regular deployments.
+        if is_amendment {
+            // Amendments must target the latest edition of the program.
+            let Some(latest_edition) = self.edition_map().get_confirmed(&program_id)?.map(|e| *e) else {
```
raychu86, line 297: Should these be `get_confirmed` or `get_speculative`?

In theory, there could be multiple amendments to the same program in the same block

raychu86: Actually we may be safe here because we abort subsequent deployments that use the same programID in the same block.

ledger/store/src/transaction/deployment.rs
d0cd: How about `AmendmentNextIndexMap`?

synthesizer/src/vm/verify.rs
d0cd: Seems reasonable @vicsn.
We do anyways check that the deployment keys are valid via synthesis. 
And we do require fees.
So an earlier part of `check_transaction` we could make a DB call and check that the VKs are indeed different.
Any concerns?

ledger/store/src/helpers/memory/transaction.rs
d0cd: It's certainly an option.
One of the motivating reasons for doing it here is so that we didn't have to introduce too much logic.
A separate amendment store would have its own getters and getters over the Deployment + Amendment store.
If we do go the Amendment store route, it might make sense to even introduce separate `Transaction::Amendment`.
Do y'all see it the same way?

ledger/store/src/helpers/rocksdb/transaction.rs
d0cd: Yeah its a bit of a misnomer.
Ideally the `ProgramOwner` struct would be more like `DeploymentSigner`.
We may be able to rename it.

ledger/block/src/transaction/deployment/mod.rs
d0cd: It's certainly possible. 
Would you want it to also be reflected in the JSON serialization and explicitly stored in the DB?
Also its worth noting that the `version` function is used a few times in this code base to implicitly check that the deployment is well-formed, but we can find a way to keep that check around.

ledger/store/src/helpers/memory/transaction.rs
vicsn: IMO the existing PR is clean enough

synthesizer/src/vm/verify.rs
vicsn: Sounds great!

ledger/store/src/helpers/memory/transaction.rs
d0cd: We had a chance to discuss in person and we came to the conclusion that this PR is fine.
In the future, we can do "Ratification"s as a separate `Transaction` variant with its own store.

synthesizer/src/vm/verify.rs
d0cd: Addressed in [30a4ba5](https://github.com/ProvableHQ/snarkVM/pull/3075/commits/30a4ba5dcee6f01ffab28d2f41ca179f8d610eac).

@raychu86 @vicsn this restriction currently causes tests to fail since there are no changes to the VKs at this time. My recommendation would be to pause review on this PR, wait until dynamic dispatch is merged, and then come back to this.

ledger/block/src/transaction/deployment/bytes.rs
```diff
@@ -72,18 +73,49 @@ impl<N: Network> FromBytes for Deployment<N> {
...
+            DeploymentVersion::V2 | DeploymentVersion::V3 => {
                 // Read the program owner.
                 let owner = Address::<N>::read_le(&mut reader)?;
                 Some(owner)
             }
         };
 
+        // If the deployment version is V3, read the translation verifying keys.
+        let translation_verifying_keys = match version {
+            DeploymentVersion::V1 | DeploymentVersion::V2 | DeploymentVersion::V4 => None,
```
raychu86: This seems wrong. DeploymentVersion::V4 should have translation verifying keys.

d0cd: Agreed, still need to work through some changes to account for the no translation keys case.

ledger/block/src/transaction/deployment/bytes.rs
```diff
@@ -141,20 +143,25 @@ impl<N: Network> ToBytes for Deployment<N> {
...
         }
-        // If the deployment version is 3, write the translation verifying keys.
-        // Note: The unwrap is safe because `Deployment::version` only returns `V3` if the translation verifying keys are present.
-        if matches!(version, DeploymentVersion::V3) {
+        // If the deployment version is V3 or V4, write the translation verifying keys.
+        // V3 always has translation verifying keys (may be empty vec for programs without records).
+        // V4 always has translation verifying keys (may be empty vec for programs without records).
+        if matches!(version, DeploymentVersion::V3 | DeploymentVersion::V4) {
             // Get the translation verifying keys.
+            // Note: The unwrap is safe because V3 and V4 always have translation verifying keys.
```
d0cd:
```suggestion
            // Note: The unwrap is safe because `Deployment::version` guarantees that `self.translation_verifying_keys` is some for V3 and V4.
```

ledger/block/src/transaction/deployment/CLAUDE.md
d0cd, line 1: Remove this file.

ledger/block/src/transaction/deployment/mod.rs
```diff
@@ -327,30 +329,36 @@ impl<N: Network> Deployment<N> {
...
-    // Inactive after consensus version >= V9.
+pub enum DeploymentVersion {
+    /// A deployment without a program checksum or program owner (V1).
+    /// Inactive after consensus version >= V9.
     V1 = 1,
-    // Active after consensus version >= V9.
+    /// A deployment with both a program checksum and program owner (V2).
+    /// Active after consensus version >= V9.
     V2 = 2,
+    /// A deployment with a program checksum but without a program owner (V3).
```
d0cd, line 359: This should say, `Active after consensus version >= V14`

ledger/block/src/transaction/deployment/mod.rs
```diff
@@ -495,4 +503,47 @@ function compute:
         )
         .unwrap()
     }
+
+    /// Samples a V3 deployment (amendment) for the same program as V2.
+    /// V3 = checksum + no owner.
+    pub(crate) fn sample_deployment_v3(edition: u16, rng: &mut TestRng) -> Deployment<CurrentNetwork> {
```
d0cd, line 509: You could also run `sample_v2_deployment_with_translation_keys` and remove the owner. And then set the appropriate edition.

ledger/block/src/transaction/merkle.rs
```diff
@@ -146,6 +146,9 @@ impl<N: Network> Transaction<N> {
         match deployment.version() {
             Ok(DeploymentVersion::V1) => Self::deployment_tree_v1(deployment),
             Ok(DeploymentVersion::V2) => Self::deployment_tree_v2(deployment),
+            // Note: We use the same method for computing the deployment tree for V2 and V3.
+            // This is safe because the tree root contains a hash of all bytes in the deployment.
```
d0cd: "And V3 is guaranteed to have no owner"

ledger/test-helpers/src/lib.rs
```diff
@@ -292,6 +292,50 @@ function compute:
     .unwrap()
 }
 
+/// Samples a V3 deployment (amendment) for the same program as V2.
+/// The edition must match an existing deployment's edition.
+/// V3 = checksum + no owner.
+pub fn sample_deployment_v3(edition: u16, rng: &mut TestRng) -> Deployment<CurrentNetwork> {
```
d0cd, line 298: You could sample a `v2_deployment_with_translation_keys` and remove the owner.

synthesizer/process/src/deploy.rs
```diff
@@ -23,20 +22,19 @@ impl<N: Network> Process<N> {
         &self,
         program: &Program<N>,
         rng: &mut R,
-    ) -> Result<Deployment<N>, ProcessDeployError> {
+    ) -> Result<Deployment<N>> {
```
d0cd: We should bring back the existing error system.

ledger/block/src/transaction/merkle.rs
```diff
@@ -146,6 +146,10 @@ impl<N: Network> Transaction<N> {
...
+            // And V3 is guaranteed to have no owner.
```
d0cd:
```suggestion
            // This is safe because the tree root contains a hash of all bytes in the deployment and V3 is guaranteed to have no owner.
```

ledger/store/src/helpers/rocksdb/internal/id.rs
```diff
@@ -324,6 +330,17 @@ enum DataID {
     // Track dynamic IDs for outputs.
     OutputDynamicIDMap,
 
+    // Deprecated: was DeploymentContainsTranslationKeysMap.
```
d0cd: Delete this entry, it was never released.

ledger/store/src/transaction/mod.rs
d0cd, line 1: Are there any other functions whose comments should mention amendments and whose names should get renamed to account for "original" vs "latest"? Ditto for `block/src/transaction/find.rs`

d0cd, line 1: Take a holistic look at all of the DB methods? After adding amendments, are we consistent? Are there any we want to add? They should be named consistently across the codebase.

ledger/test-helpers/src/lib.rs
```diff
@@ -292,6 +292,16 @@ function compute:
...
```
d0cd, line 298: Is this correct? Shouldn't we sampling with translation keys?

ledger/test-helpers/src/lib.rs
```diff
@@ -518,6 +528,10 @@ pub fn sample_deployment_transaction(
             deployment.set_program_owner_raw(Some(Address::try_from(&private_key).unwrap()));
             deployment
         }
+        (3, _) => {
+            // V3 is an amendment - uses the same program as V2 but with new VKs and no program_owner.
```
d0cd:
```suggestion
            // V3 is an amendment - uses the same program as V2 but with additional translation VKs and no program_owner.
```

ledger/test-helpers/src/lib.rs
d0cd: Make sure all the helpers are consitent in this behavior.

synthesizer/process/src/stack/mod.rs
```diff
@@ -295,7 +295,7 @@ impl<N: Network> Stack<N> {
     }
 
     /// Partially initializes a new stack, given the process and the program, without checking for validity.
-    /// Note. This method should **NOT** be used by the on-chain VM to add new program, use `Stack::new` instead.
+    /// Note. This method should be used with caution as it does not completely check for validity.
```
d0cd:
```suggestion
    /// Note. This method should be used with caution as it does not completely check for validity.
    /// A `Stack` created by this method should also call `intialize_and_check`.
```

synthesizer/process/src/deploy.rs
```diff
@@ -45,21 +45,48 @@ impl<N: Network> Process<N> {
...
+
+                // Compute a new stack with the same program and edition.
+                let mut stack = Stack::new_raw(self, deployment.program(), *existing_stack.program_edition())?;
+                stack.initialize_and_check(self)?;
+                lap!(timer, "Compute the stack");
+
+                // Set the program owner to the existing owner.
+                stack.set_program_owner(*existing_stack.program_owner());
+
+                // Insert all verifying keys (unified: functions + records).
```
d0cd: We could dedup the key insertion logic outside of the loop.

synthesizer/process/src/deploy.rs
```diff
@@ -45,21 +45,48 @@ impl<N: Network> Process<N> {
...
+                lap!(timer, "Insert the verifying keys");
+
+                stack
+            }
+            DeploymentVersion::V3 => {
+                // V3 is an amendment — get the existing stack.
+                let existing_stack = self.get_stack(deployment.program_id())?;
+
+                // Compute a new stack with the same program and edition.
+                let mut stack = Stack::new_raw(self, deployment.program(), *existing_stack.program_edition())?;
```
d0cd, line 70: Make a note on why we use `Stack::new_raw` here and why it is safe.

synthesizer/process/src/finalize.rs
```diff
@@ -34,77 +34,123 @@ impl<N: Network> Process<N> {
...
+
+        // Finalize the deployment based on its version.
+        match version {
+            DeploymentVersion::V1 | DeploymentVersion::V2 => {
+                // Compute the program stack.
+                let mut stack = Stack::new(self, deployment.program())?;
+                lap!(timer, "Compute the stack");
+
+                // Set the program owner.
+                // Note: The program owner is only enforced to be `Some` after `ConsensusVersion::V9`
```
d0cd:
```suggestion
```

synthesizer/process/src/finalize.rs
```diff
@@ -34,77 +34,123 @@ impl<N: Network> Process<N> {
...
-                new_mappings
-            }
-        };
-        lap!(timer, "Retrieve the mappings to initialize");
 
-        // Initialize the mappings, and store their finalize operations.
-        atomic_batch_scope!(store, {
-            // Initialize a list for the finalize operations.
-            let mut finalize_operations = Vec::with_capacity(deployment.program().mappings().len());
+                // Compute a new stack with the same program and edition.
```
d0cd, line 123: Make a note on why `Stack::new_raw` is needed and why it is safe.

synthesizer/process/src/verify_deployment.rs
```diff
@@ -28,23 +28,42 @@ impl<N: Network> Process<N> {
...
         }
 
         // Ensure the program is well-formed, by computing the stack.
         // Note: The program owner is intentionally not set, since `program_owner` is an operand
         //   that is only available in a finalize scope.
-        let stack = Stack::new(self, deployment.program())?;
+        let stack = if is_amendment {
+            // For amendments, use the existing edition instead of incrementing.
+            let existing_stack = self.get_stack(program_id)?;
+            let stack = Stack::new_raw(self, deployment.program(), *existing_stack.program_edition())?;
```
d0cd, line 63: Make a note on why `Stack::new_raw` is needed and why it is safe.

synthesizer/src/vm/tests/test_v14/amendments.rs
d0cd, line 1: Review all of the tests in these files. Are any of them redundant? Dedup but maintain coverage.

Also do we test the behavior of dynamic calls for programs that were deployed before v14 with records but haven't been amended yet? To do this, we'd need to set up a 2-VM test (one for the local prover and one for the main VM) to make sure that translation keys for the main VM are not synthesized on the fly. The expected behavior is that if the programs haven't been amended, they shouldn't be dynamically callable with record translations.

synthesizer/src/vm/verify.rs
```diff
@@ -196,86 +196,14 @@ impl<N: Network, C: ConsensusStorage<N>> VM<N, C> {
...
-                    );
-                }
+                // Get the deployment version.
+                // This implicitly checks that the program checksum and program owner fields are set correctly.
+                let version = deployment.version()?;
+
+                // Validate deployment constraints based on consensus version.
+                validate_deployment_for_consensus_version(deployment, version, consensus_version, *id)?;
+
+                // If the consensus version is V13 or later, verify the program's mappings do not use non-existent structs.
```
d0cd: Should these checks below also be refactored into `validate_deployment_for_consensus_version`?

d0cd: Checks that can be done purely on the deployment, should go in there.

ledger/store/src/transaction/deployment.rs
```diff
@@ -208,140 +274,258 @@ pub trait DeploymentStorage<N: Network>: Clone + Send + Sync {
...
+                edition == latest_edition,
+                "Failed to insert amendment transaction '{transaction_id}' for program '{program_id}': expected edition {latest_edition}, found edition {edition}"
+            );
+
+            // Get the current amendment count, or initialize to 0 if this is the first amendment.
+            let amendment_index =
+                self.amendment_next_index_map().get_confirmed(&(program_id, edition))?.map(|c| *c).unwrap_or(0);
+
+            atomic_batch_scope!(self, {
+                // Store the program ID and edition for faster lookups.
```
d0cd:
```suggestion
                // Store the program ID and edition.
```

ledger/store/src/transaction/deployment.rs
```diff
@@ -208,140 +274,258 @@ pub trait DeploymentStorage<N: Network>: Clone + Send + Sync {
...
+                self.id_map().insert(*transaction_id, program_id)?;
+                self.id_edition_map().insert(*transaction_id, edition)?;
```
d0cd, line 317: Add comment here.

ledger/store/src/transaction/deployment.rs
```diff
@@ -208,140 +274,258 @@ pub trait DeploymentStorage<N: Network>: Clone + Send + Sync {
...
-            //  `ConsensusVersion::V9` introduces upgradability which allows editions to be incremented up to `u16::MAX`
-            self.id_edition_map().insert(*transaction_id, edition)?;
-
-            // If the checksum exists, then store it into the `ChecksumMap`.
-            if let Some(checksum) = checksum {
-                self.checksum_map().insert((program_id, edition), checksum)?;
-            }
+                // Increment the amendment count.
+                let next_index = amendment_index.checked_add(1).ok_or_else(|| {
+                    anyhow!("Amendment index overflow for program '{program_id}' (edition {edition})")
```
d0cd, line 342: Does bailing here create any issues? Do we need to also guard against this in `check_transaction` to ensure safe failure early?

ledger/store/src/transaction/deployment.rs
```diff
@@ -357,15 +541,31 @@ pub trait DeploymentStorage<N: Network>: Clone + Send + Sync {
...
+            let count = *amendment_count;
+            if count > 0 {
+                // Return the latest amendment transaction ID.
+                let latest_index = count.saturating_sub(1);
+                if let Some(transaction_id) =
+                    self.amendment_id_map().get_confirmed(&(*program_id, edition, latest_index))?
+                {
+                    return Ok(Some(*transaction_id));
+                }
+            }
```
d0cd: Does the other case need to be handled? Or should we leave a comment here.

ledger/store/src/transaction/deployment.rs
```diff
@@ -462,7 +719,29 @@ pub trait DeploymentStorage<N: Network>: Clone + Send + Sync {
...
+                    edition,
+                    latest_index,
+                ))?
+                else {
+                    bail!(
+                        "Failed to get the amendment verifying key for '{program_id}/{resource_name}' (edition {edition}, amendment {latest_index})"
+                    );
+                };
+                return Ok(Some(verifying_key.into_owned()));
+            }
```
d0cd: Does the other case need to be handled or do leave a comment?

d0cd: Ditto for similar patterns in the rest of the file.

d0cd: Could the pattern be refactored? Seems to create a lot of code.

ledger/store/src/transaction/deployment.rs
```diff
@@ -555,16 +1018,16 @@ pub trait DeploymentStorage<N: Network>: Clone + Send + Sync {
...
             bail!("Failed to get the deployed program '{program_id}' (edition {edition})");
         };
+
         // Retrieve the checksum.
         let program_checksum =
             self.checksum_map().get_confirmed(&(program_id, edition))?.map(|checksum| checksum.into_owned());
-        // If the checksum is present, then retrieve the owner address.
-        // Note: This is done to ensure that `Deployment` is consistent. Both the checksum and owner must be present or absent.
-        // This invariant is also enforced in `check_transaction`.
+        // If the checksum is present, retrieve the owner address (both must be present or absent).
```
d0cd: Is this comment still accurate?

ledger/store/src/transaction/deployment.rs
```diff
@@ -1164,4 +1908,296 @@ mod tests {
...
+        assert!(deployment_store.get_transaction(&amendment_0_id).unwrap().is_some());
+
+        // Remove amendment 1 (latest) - should succeed.
+        deployment_store.remove(&amendment_1_id).unwrap();
+        assert_eq!(Some(1), deployment_store.get_amendment_count(&original_program_id, 0).unwrap());
+
+        // Now removing amendment 0 should succeed (it's now the latest).
+        deployment_store.remove(&amendment_0_id).unwrap();
+        assert_eq!(None, deployment_store.get_amendment_count(&original_program_id, 0).unwrap());
+    }
```
d0cd, line 2293: Can you think of any other tests worth running?

synthesizer/src/vm/verify.rs
```diff
@@ -196,137 +196,29 @@ impl<N: Network, C: ConsensusStorage<N>> VM<N, C> {
...
-                        "Invalid deployment transaction '{id}' - program uses string type after `ConsensusVersion::V12`"
-                    );
-                }
-                if consensus_version < ConsensusVersion::V13 {
-                    ensure!(
-                        !deployment.program().contains_external_struct(),
-                        "Invalid deployment transaction '{id}' - external structs may only be used beginning with `ConsensusVersion::V13`"
-                    );
-                }
+                // Get the deployment version.
```
d0cd, line 196: Revert the refactor to deployment verification, it should just be its own PR.

ledger/store/src/transaction/deployment.rs
```diff
@@ -190,10 +257,16 @@ pub trait DeploymentStorage<N: Network>: Clone + Send + Sync {
         self.checksum_map().finish_atomic()?;
         self.verifying_key_map().finish_atomic()?;
         self.certificate_map().finish_atomic()?;
+        self.amendment_next_index_map().finish_atomic()?;
+        self.amendment_id_map().finish_atomic()?;
+        self.reverse_amendment_id_map().finish_atomic()?;
+        self.amendment_verifying_key_map().finish_atomic()?;
+        self.amendment_certificate_map().finish_atomic()?;
+        self.amendment_owner_map().finish_atomic()?;
         self.fee_store().finish_atomic()
```
vicsn: In all the other functions in this file `fee_store` doesn't come last.

vicsn (approved):
Still LGTM! Just one nit. The many checks and functions are somewhat lengthy and some might prefer refactoring, but I'm happy to move forward.

ledger/store/src/transaction/deployment.rs
d0cd: Addressed in [0a0340f](https://github.com/ProvableHQ/snarkVM/pull/3075/commits/0a0340f14117e91f5161b51e7ac265836c03bbc7)

synthesizer/src/vm/verify.rs
```diff
@@ -338,115 +348,186 @@ impl<N: Network, C: ConsensusStorage<N>> VM<N, C> {
...
+                        *stack.program_edition() == deployment.edition(),
+                        "Invalid deployment transaction '{id}' - program edition does not match the existing program edition"
+                    );
+
+                    // Ensure at least one VK (function or translation) has changed or been added.
+                    // This prevents spam amendments that don't actually modify anything.
+                    // Note: `deployment.verifying_keys()` returns the unified vector of
+                    // function and record (translation) VKs, so this covers both.
+                    let mut has_vk_change = false;
+                    for (name, (new_vk, _)) in deployment.verifying_keys() {
```
raychu86, line 397: Nit: We may want to parallelize this since it is's a full DB read and VK deserialization for each check.

@ljedrz may have more opinions on optimizations

synthesizer/src/vm/verify.rs
```diff
@@ -338,115 +348,186 @@ impl<N: Network, C: ConsensusStorage<N>> VM<N, C> {
...
+                    // Check that the program does not make any calls to `credits.aleo/upgrade`.
+                    // Note: This is safe to check for programs deployed before `ConsensusVersion::V8` because `credits.aleo/upgrade` was not yet introduced.
+                    deployment.program().check_external_calls_to_credits_upgrade()?;
                 }
-                // Check that the program does not make any calls to `credits.aleo/upgrade`.
-                // Note: This is safe to check for programs deployed before `ConsensusVersion::V8` because `credits.aleo/upgrade` was not yet introduced.
-                deployment.program().check_external_calls_to_credits_upgrade()?;
 
                 // Verify the deployment if it has not been verified before.
                 if !is_partially_verified {
```
raychu86, line 533: I thought a while about if amendments can get past `is_partially_verified` cache since it's based on Transaction ID `root(deployment_tree + fee)`. The deployment trees are the same despite the VKs being different, so a malicious party could in theory inject an invalid VK and it would pass.

But I think it's safe because we enforce that the Fee must be unique.


Would like a sanity-check here.

raychu86: I also thought about the reseting of the cache on each new ConsensusVersion would also help with safety. But I don't think it will since we don't have a rule that says Amendments are only valid on deployments that were deployed before V14 (no translation keys).

raychu86 (approved):
Left some nits/questions, but looking good overall.

raychu86:
@d0cd During one of our in-person review sessions, I recall some sort of issue with deployment(?) loading from storage where we were unable to determine which version properly without an extra map. 

I don't see the issue now, so I assume you fixed it. But wanted to double check, since I feel uneasy forgetting that issue was.

synthesizer/src/vm/verify.rs
ljedrz: I think it's just 2 identical method names conflated here - `DeploymentStore::verifying_keys` would indeed potentially be a performance issue, but `Deployment::verifying_keys` (which I believe is the one called here) works on in-memory data that's already deserialized.

synthesizer/src/vm/verify.rs
d0cd: The deployment trees won't be the same if the VKs are different because the root includes a hash of the entire deployment.

synthesizer/src/vm/verify.rs
d0cd: That's right. Here we are just checking against the existing VK in the stack, which should be that of the latest amendment (if one exists).

d0cd:
> @d0cd During one of our in-person review sessions, I recall some sort of issue with deployment(?) loading from storage where we were unable to determine which version properly without an extra map.

In previous versions of the design, we had:
- V2 for existing deployments
- V3 for new deployments with translation keys
- V4 for amendment deployments

In this design, it wasn't possible to know whether a deployment for a program without records is V2 or V3 without an extra map indicating it.

In the current design we use V2 for both existing deployments and upcoming deployments with translation keys. 
We also reuse the exiting DB maps to store the translation keys since record and function names can't clash.
So there's no need to distinguish between the two:
- a deployment without records has and will be V2
- a deployment with records but made before V14 will be V2, but won't have translation keys
- a deployment with records made after V14 will be V2, and will have translation keys

synthesizer/src/vm/verify.rs
raychu86: Ah, that's right. This shouldn't be an issue.

PR #3078, opened 2026-01-07
Reduce circleci jobs

vicsn:
## Motivation

To reduce costs, which are starting to get out of hand. 

Please review whether removing the noconsole jobs are fine, they seem:
(1) inaccurate, because we dó use Console stuff in Circuit crates
(2) not even unique because the two crates in question don't have custom features?

Two concurrent efforts are:
- migration to free but less performant github actions, this will take a lot longer to verify: https://github.com/ProvableHQ/snarkVM/pull/3074
- making tests more efficient: https://github.com/ProvableHQ/snarkVM/pull/3058

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -1205,16 +1205,15 @@ workflows:
       - synthesizer-process 
       - synthesizer-integration
       - synthesizer-program
-      - synthesizer-program-integration
       - synthesizer-snark
 
   snarkvm-workflow:
     jobs:
-      - snarkvm
       - algorithms
       - curves
       - fields
       - parameters
+      - snarkvm
       - utilities
       - utilities-derives
       - wasm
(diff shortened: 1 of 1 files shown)


ljedrz (approved):

.circleci/config.yml
```diff
@@ -1229,31 +1204,38 @@ workflows:
...
       - curves
       - fields
       - parameters
-      - parameters-large
-      - parameters-uncached
       - utilities
       - utilities-derives
       - wasm
+  
+  # Enable to run additional checks before releases
```
raychu86: Is there a way we can enforce that we do not forget to run these prior to release? This seems like an easy thing to forget.

vicsn: Changed and tested it so they only run on merges instead of on every commit

vicsn: Added a new `release-workflow`

raychu86 (approved):

meddle0x53:
@vicsn I my PR the synthetizer-heavy run, should it also go under the merge workflow? What do you think?

PR #3081, opened 2026-01-08
feat: On instruction failure, return an error with the failing aleo instruction index (re-opened)

mitchmindtree:
***NOTE: This is PR #3064 but from a branch on this repo rather than my fork. See [this comment](https://github.com/ProvableHQ/snarkVM/pull/3064#issuecomment-3721734520) for context.***

## Motivation

This partly addresses #3055, ensuring that we expose the failing Aleo instruction from the top-level VM methods.

This does _not_ address the halt=panic issue discussed in #2941, though should pave the way so that when we do introduce better error handling for halts, we have a clear path to propagate these errors up to the VM user.

## Overview

When an instruction fails, the new `StackExecError::Instruction` variant now captures:

1. Index in the function's instruction list
2. The formatted instruction (e.g., "add r0 r1 into r2" or "call foo.aleo/bar")
3. The failure reason

The error format is: "Instruction ({instruction_string}) at index {index} failed: {error}"

For nested calls, you get a "stack trace" of instruction contexts, e.g.

```
Stack execution failed:
  Instruction (call foo.aleo/bar) at index 3 failed:
    Call failed:
      Substack execution failed:
        Instruction (add r0 r1 into r2) at index 7 failed: {actual error}
```

This is propagated up through to the main public VM methods (authorize, execute, evaluate, deploy).

Where necessary, an `Anyhow` variant has been added to the new error types in order to keep the changes minimal and focused on our instruction context. Ideally, we can remove these catch-all `Anyhow` variants as we migrate errors over to dedicated types with `thiserror` (see #3056).

## Follow-up

- [ ] Address #2941, replacing `panic!` behaviour with error handling.

### DIFF
--- a/.circleci/semver-checks.sh
+++ b/.circleci/semver-checks.sh
@@ -2,7 +2,7 @@
 
 set -x
 
-BASELINE_REV=acd55ad100550 # UPDATE ME ON NECESSARY BREAKING CHANGES
+BASELINE_REV=1d47c130fa9c2 # UPDATE ME ON NECESSARY BREAKING CHANGES
 
 # Ensure that the command is installed.
 cargo install cargo-semver-checks@0.43.0 --locked
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -3687,6 +3687,7 @@ dependencies = [
  "snarkvm-synthesizer-snark",
  "snarkvm-utilities",
  "tempfile",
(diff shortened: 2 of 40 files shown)


mitchmindtree:
CI error appears unrelated, addressed by #3080.

vicsn (approved):

PR #3082, opened 2026-01-08
feat: Use `Result` for constraint enforcement and assertions

mitchmindtree:
## Motivation

This is a follow-up to #3081 and continues to address #2941 and #3055, i.e. replacing `panic!`s with proper error handling.

To keep the changes minimal, this PR specifically focuses on evaluation and execution of the assert operations (a common `panic!` pain-point for leo testing). This has helped to guide the remaining error propagation from the instruction methods (`evaluate`, `execute` and `finalize`) up to the `IndexedInstructionError` introduced in #3081.

This does *not* add error handling for all instructions, but should pave a clear path for doing so.

## Summary

- Make circuit `E::enforce` return `Result<(), ConstraintUnsatisfied>` instead of panicking.
- Add structured error types for assertion instruction evaluation, finalization, and execution (`EvalError`, `FinalizeError`, `ExecError`).
- Propagate constraint errors through the synthesizer stack.

## Details

**Circuit layer**: `E::enforce`, `E::assert`, and `E::assert_eq` now return `Result<(), ConstraintUnsatisfied>`, where previously they would `panic!`. The new `ConstraintUnsatisfied` error captures the constraint values (a, b, c) where the check a * b == c failed.

**Synthesizer layer**: New error types in `synthesizer/program/src/logic/instruction/error.rs`:

- `EvalError` - errors during instruction evaluation.
- `FinalizeError` - errors during finalization.
- `ExecError` - errors during instruction execution.
- `AssertError` - assert instruction failures (the first instruction-associated eval error). 

**Error propagation**: `StackExecError`, `CallExecError`, and `InstructionExecError` now include variants to propagate `ConstraintUnsatisfied` and `ExecError` from the circuit/instruction layers.

**Note:** Many `E::assert`call sites throughout the circuit crates now use `.expect()` to maintain the existing panicking behavior. This is temporary—these sites will be addressed in follow-up PRs. For now, only the assertion instructions (assert.eq, assert.neq) have been fully updated to return `Result` for unsatisfied constraints across all three paths (evaluate, finalize, execute).

## Follow-up

- [ ] Continue on #2941.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -2943,6 +2943,7 @@ dependencies = [
  "snarkvm-curves",
  "snarkvm-fields",
  "snarkvm-utilities",
+ "thiserror 2.0.17",
 ]
 
 [[package]]
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -3744,6 +3744,7 @@ dependencies = [
  "snarkvm-synthesizer-process",
  "snarkvm-synthesizer-snark",
  "snarkvm-utilities",
+ "thiserror 2.0.17",
  "tiny-keccak",
 ]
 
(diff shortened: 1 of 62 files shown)


vicsn (approved):
LGTM!

vicsn:
Can you fix clippy in a follow-up PR?

mitchmindtree:
> Can you fix clippy in a follow-up PR?

Opened #3085

PR #3086, opened 2026-01-09
[Perf] Speed up sum_of_products

ljedrz:
This operation is non-negligible when executing transactions; a small optimization results in the following benchmark improvements:
```
Transaction::Execute(transfer_public) - verify
                        time:   [41.303 µs 41.309 µs 41.317 µs]
                        change: [−2.1724% −2.1356% −2.1002%] (p = 0.00 < 0.05)
                        Performance has improved.

Transaction::Execute(transfer_private) - verify
                        time:   [47.150 µs 47.159 µs 47.179 µs]
                        change: [−1.5528% −1.4960% −1.4362%] (p = 0.00 < 0.05)
                        Performance has improved.
```
I've checked all `varuna` and `transaction` benchmarks - the rest are either unaffected or the performance improvement is minimal.

CC @niklaslong

### DIFF
--- a/algorithms/src/crypto_hash/poseidon.rs
+++ b/algorithms/src/crypto_hash/poseidon.rs
@@ -233,8 +233,9 @@ impl<F: PrimeField, const RATE: usize> PoseidonSponge<F, RATE, 1> {
     #[inline]
     fn apply_mds(&mut self) {
         let mut new_state = State::default();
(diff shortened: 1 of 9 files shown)


vicsn (approved):
IMO not worth the effort anymore, we have plenty of backlog to create much better proving times by changing the proof system or lower level operations.

PR #3087, opened 2026-01-09
[Feat] Increase max program size to 512kB and max writes per finalize to 32

raychu86:
## Motivation

This PR increases the program size limit from 100 kB to 512 kB. This will allow developers to write much more expressive programs without having to use a nested import structure. To support this, the `MAX_TRANSACTION_SIZE` has also been increased to 540 kB.

In addition, this PR also increases the `MAX_WRITES` in each finalize scope/constructor from 16 to 32.

## Test Plan

Tests have been added to ensure that the new program size bound is increased at the new `ConsensusVersion`.

## Backwards compatibility

A new `ConsensusVersion` is required to ensure nodes accept the new rule at the same time to prevent forking. The rules  will be enforced during deployment verification.

snarkOS sister PR: https://github.com/ProvableHQ/snarkOS/pull/4086


console/network/src/lib.rs
```diff
@@ -201,7 +201,7 @@ pub trait Network:
     const MAX_RECORD_ENTRIES: usize = Self::MIN_RECORD_ENTRIES.saturating_add(Self::MAX_DATA_ENTRIES);
 
     /// The maximum program size by number of characters.
-    const MAX_PROGRAM_SIZE: usize = 100_000; // 100 kB
+    const MAX_PROGRAM_SIZE: usize = 256_000; // 256 kB
```
vicsn: In previous changes to constants, we changed this to an array with (value, consensus_height) pairs. Besides being more consistent, this also forces dependents to be forced to reason about the change at compile time

vicsn (changes requested):
One downside worth calling out is that this makes DoS prevention more complicated, as the minimum message size must now be at least 256kb (ignoring BlockResponse, for which we're building a different pathway).

That said, I'm not too worried about it, validators are well-firewalled anyway, and at this stage we should prioritize building apps.

console/network/src/lib.rs
raychu86: Good callout. Fixed in [429b237](https://github.com/ProvableHQ/snarkVM/pull/3087/commits/429b2370831afdde1e3999d59165678d42bd973e)

vicsn (approved):
Core logic LGTM, CI is failing

d0cd (dismissed):
I believe the `MAX_TRANSACTION_SIZE` also needs to be bumped up.

vicsn:
Hmm good catch, and make sure to add a unit test comparing the two constants so we don't forget in the future.

console/network/src/lib.rs
```diff
@@ -308,6 +308,20 @@ pub trait Network:
         Self::MAX_CERTIFICATES.last().map_or(Err(anyhow!("No MAX_CERTIFICATES defined.")), |(_, value)| Ok(*value))
     }
 
+    /// Returns the last `MAX_PROGRAM_SIZE` value.
+    #[allow(non_snake_case)]
+    fn LATEST_MAX_PROGRAM_SIZE() -> Result<usize> {
```
d0cd: Are we okay with changing this interface to `usize` and unwrapping here?

console/network/src/mainnet_v0.rs
```diff
@@ -185,6 +185,16 @@ impl Network for MainnetV0 {
...
     ];
+    /// The maximum program size by number of characters.
+    const MAX_PROGRAM_SIZE: [(ConsensusVersion, usize); 2] = [
+        (ConsensusVersion::V1, 100_000),  // 100 kB
+        (ConsensusVersion::V14, 512_000), // 512 kB
+    ];
+    /// The maximum transaction size in bytes.
+    const MAX_TRANSACTION_SIZE: [(ConsensusVersion, usize); 2] = [
+        (ConsensusVersion::V1, 128_000),  // 128 kB
+        (ConsensusVersion::V14, 540_000), // 540 kB
```
raychu86: Why `540` specifically?

console/network/src/lib.rs
raychu86: Yea I'm fine with that (using `.expect`s). But have this in other places as well, for instance, `LATEST_MAX_CERTIFICATES`.

console/network/src/mainnet_v0.rs
d0cd: The previous delta was 28kb so I stuck with it.

raychu86: I assume that was because it was set to fairly round numbers of 100kb and 128kb rather than explicitly specifically 28kb around the data. Though, I could be wrong.

vicsn:
CI is failing

console/network/src/mainnet_v0.rs
vicsn: Can we set it to max_program_size + ((vk_size + certificate_size)*max_functions) + remaining constant size fields from a Deploy Transaction - and document the derivation in a comment?

raychu86: Did a breakdown of the maximum transaction size:

| Name | Size (bytes) | Number | Total (bytes) |
|---|---:|---:|---:|
| tx_id | 32 | 1 | 32 |
| deployment_id | 32 | 1 | 32 |
| ProgramOwner | 161 | 1 | 161 |
| edition | 2 | 1 | 2 |
| program | MAX_PROGRAM_SIZE | 1 | MAX_PROGRAM_SIZE |
| verifying_key | 673 | X | 673·X |
| verifying_key_cert | 58 | X | 58·X |
| program_checksum | 32 | 1 | 32 |
| program_owner | 32 | 1 | 32 |
| fee_private | 2076 | 1 | 2076 |
| **TOTAL** | — | — | **2367 + MAX_PROGRAM_SIZE + (673+58)·X** |


Max number of verifying keys is 31 (MAX_FUNCTIONS) + 310 (MAX_RECORDS) = 341.

So the theoretical maximum transaction size = 2,367 + MAX_PROGRAM_SIZE + (673+58)·X = 251,638 + MAX_PROGRAM_SIZE.

d0cd: So 1 MB for a clean number?

raychu86: I'll set it to 768 kB for now, since that is within bounds. @vicsn we'll want to check implications of this in snarkOS, because the message deserialization is done with 100 kB assumptions at the moment.

vicsn:
Ci still failing

vicsn (approved):

console/network/src/mainnet_v0.rs
vicsn: Can you make an associated snarkOS PR?

Validators which didn't upgrade won't be able to deserialize and use txs above 100kb. So similar to the spend_limit checks we added, we can add an out of bounds check on the transmission size. You'll have to make sure both the **proposer** and **signer** use the same block height to base the check on.

I would leave pruning of the spend_limit check to a future date, but maybe you can move these types of checks all to a separate file in snarkOS so it doesn't further blow up primary.rs

synthesizer/program/src/lib.rs
```diff
@@ -1044,6 +1045,48 @@ impl<N: Network> ProgramCore<N> {
...
+            "Program size of {program_size} bytes exceeds the maximum allowed size of {maximum_allowed_program_size} bytes for the current height {block_height} (consensus version {}).",
+            N::CONSENSUS_VERSION(block_height)?
+        );
+
+        Ok(())
+    }
+
+    /// Checks that the program writes size does not exceed the maximum allowed size for the given block height.
+    pub fn check_program_writes(&self, block_height: u32) -> Result<()> {
+        // Determine the maximum allowed program size for the current consensus version.
```
d0cd, line 1067:
```suggestion
        // Determine the maximum allowed writes for the current consensus version.
```

synthesizer/program/src/lib.rs
```diff
@@ -1044,6 +1045,48 @@ impl<N: Network> ProgramCore<N> {
...
+        let max_num_writes = consensus_config_value!(N, MAX_WRITES, block_height)
+            .ok_or(anyhow!("Failed to fetch maximum program size"))?;
```
d0cd, line 1069:
```suggestion
            .ok_or(anyhow!("Failed to fetch maximum number of writes"))?;
```

vicsn: It's not ideal but leaving the nit to a future fix to speed up merging

console/network/src/consensus_heights.rs
```diff
@@ -49,7 +49,8 @@ pub enum ConsensusVersion {
     V12 = 12,
     /// V13: Introduces external structs.
     V13 = 13,
-    /// V14: Introduces `aleo::GENERATOR` and `aleo::GENERATOR_POWERS` opcodes.
+    /// V14: Increase the program size limit to 512 kB and transaction size limit to 540 kB.
```
d0cd, line 52:
```suggestion
    /// V14: Increase the program size limit to 512 kB and transaction size limit to 768 kB.
```

console/network/src/consensus_heights.rs
```diff
@@ -366,12 +391,41 @@ mod tests {
         }
     }
 
+    /// Ensure that `MAX_TRANSACTION_SIZE` is at least 28KB greater than `MAX_PROGRAM_SIZE` for all consensus versions.
+    /// This overhead accounts for proofs, signatures, and other transaction metadata.
+    fn transaction_size_exceeds_program_size<N: Network>() {
```
d0cd, line 396: This test is fine, but perhaps not effective now that we have an analytical expression for determining the minimum transaction size w.r.t to the program size.

console/network/src/lib.rs
```diff
@@ -220,7 +222,7 @@ pub trait Network:
     /// The maximum number of commands in finalize.
     const MAX_COMMANDS: usize = u16::MAX as usize;
     /// The maximum number of write commands in finalize.
-    const MAX_WRITES: u16 = 16;
+    const MAX_WRITES: [(ConsensusVersion, u16); 2] = [(ConsensusVersion::V1, 16), (ConsensusVersion::V14, 32)];
```
d0cd, line 225: Consider adding tests for `MAX_WRITES` in `consensus_heights.rs`.

synthesizer/process/src/cost.rs
```diff
@@ -996,7 +996,8 @@ function over_five_thousand:
     // Cost for a program +1 byte above the threshold.
     const STORAGE_COST_ABOVE_THRESHOLD: u64 = 5002;
     // Storage cost for an execution transaction at the maximum transaction size.
-    const STORAGE_COST_MAX: u64 = 3_276_800;
+    const V1_STORAGE_COST_MAX: u64 = 3_276_800;
+    const V14_STORAGE_COST_MAX: u64 = 117_964_800;
```
d0cd, line 1000: Do we want to also test deployment storage cost?

d0cd:
@raychu86 @vicsn do we need to consider changes to `FINALIZE_ID_DEPTH`? It was previously set to +4 for 16 finalize operations.

synthesizer/program/src/lib.rs
```diff
@@ -1044,6 +1045,48 @@ impl<N: Network> ProgramCore<N> {
...
+
+        // Check if the constructor exceeds the maximum number of writes.
```
d0cd, line 1071: It could be nice to bail at the specific violation so that Leo could use this to give users good message. If we want the parallelism, then we could collect the names.

synthesizer/src/vm/tests/test_v14.rs
```diff
@@ -17,11 +17,154 @@ use super::*;
...
 use snarkvm_utilities::TestRng;
 
-use console::account::ViewKey;
+// This test verifies that a large program that is over the previous size limit can be deployed after V14.
+#[test]
+fn test_deploy_large_program_v14() {
+    // Initialize an RNG.
+    let rng = &mut TestRng::default();
+
+    let large_program = Program::from_str(include_str!("./resources/large_program.aleo")).unwrap();
```
d0cd, line 30: If we want to reduce the diff, we can use the generator below.

synthesizer/program/src/lib.rs
```diff
@@ -1044,6 +1045,48 @@ impl<N: Network> ProgramCore<N> {
         })
     }
 
+    /// Checks that the program size does not exceed the maximum allowed size for the given block height.
+    pub fn check_program_size(&self, block_height: u32) -> Result<()> {
```
d0cd, line 1049: Slight nit on the API, but it would be nice if we didn't have to introduce the notion of `block_height` to the `synthesizer/program` crate. I realize it's already a bit blurry with the consensus version, so not super important.

synthesizer/src/vm/tests/test_v14.rs
```diff
@@ -171,3 +314,80 @@ fn test_aleo_generators_migration() {
...
+
+    // Ensure that the deployment is valid after ConsensusVersion::V14.
+    assert!(vm.check_transaction(&deployment, None, rng).is_ok());
+
+    // Deploy the program.
+    let next_block = sample_next_block(&vm, &private_key, &[deployment], rng).unwrap();
+    vm.add_next_block(&next_block).unwrap();
+
+    // Ensure that the valid transaction was accepted and the invalid one was rejected.
+    assert_eq!(next_block.transactions().num_accepted(), 1);
```
d0cd, line 392: We should also test execution.

PR #3088, opened 2026-01-12
Update link to README banner

raychu86:
## Motivation

This PR updates the link to the banner in the README. The previous `.png` link is now dead. 

We should consider updating the host site as well.


vicsn (approved):
Thx, we still need to change to Provable right

PR #3094, opened 2026-01-13
Introduce heavy test workflow

meddle0x53:
## Motivation

Move the heavy tests that actually get very slow when they are ran in parallel with other tests in their own workflow that is sequential. This actually makes the test runs faster which can be seen in this PR : https://github.com/ProvableHQ/snarkVM/pull/3058

## Test Plan

The test runtimes from the CI runs are improved.


.config/nextest.toml
```diff
@@ -0,0 +1,107 @@
...
+[[profile.ci.overrides]]
+filter = 'test("test_block_cache_during_advancement")'
+test-group = "ledger-heavy"
+
+[[profile.ci.overrides]]
+filter = 'test("test_block_cache_eviction")'
+test-group = "ledger-heavy"
+
+[[profile.ci.overrides]]
+filter = "test(/^snarkvm-synthesizer::vm::finalize::tests::test_/) or test(/^snarkvm-synthesizer::vm::verify::tests::test_/)"
```
vicsn: Why 4 test-threads here? Can we document why this edge case on top of the edge case is needed?

.circleci/config.yml
```diff
@@ -280,7 +278,14 @@ commands:
...
-              FILTER_EXPR='test(/test_command_parse/)+test(/test_instruction_parse/)+test(/test_process_execute/)+test(/test_program_parse/)+test(/test_vm_execute_and_finalize/)'
+              FILTER_EXPR='test(/test_command_parse/)+test(/test_instruction_parse/)+test(/test_process_execute/)+test(/test_program_parse/)'
+            fi
+            HEAVY_EXPR='test(/vm::finalize::tests::test_atomic_finalize_many/) or test(/vm::finalize::tests::test_excess_transactions_should_be_aborted/) or test(/vm::finalize::tests::test_finalize_catch_halt/) or test(/vm::finalize::tests::test_rejected_transaction_should_not_update_storage/) or test(/vm::test_helpers::test_load_deployments_with_imports/) or test(/vm::test_helpers::test_multiple_deployments_and_multiple_executions/) or test(/vm::verify::tests::test_verify_deploy_and_execute/)'
+            if [ "$WSM" = "snarkvm-synthesizer" ] && [ "${SYNTHESIZER_HEAVY:-0}" = "1" ]; then
+              FILTER_EXPR="$HEAVY_EXPR"
+            fi
+            if [ "$WSM" = "snarkvm-synthesizer" ] && [ "${SYNTHESIZER_HEAVY:-0}" != "1" ] && [ "$HAS_TEST_WILDCARD" -ne 1 ]; then
+              FILTER_EXPR="not ($HEAVY_EXPR)"
```
vicsn, line 291: Can you explain the meaning of this code?

I see we're manually specifying test names both here and in the nextest TOML. Is it possible to only have it in a single place?

meddle0x53: These are form my experimental PR, I think that the toml magic was not working that's why I did this to make it work, so when it passes and I see that it is an improvement just this run, I'll remove the toml to see if it had any worth  at all here.

vicsn: It needs better comments

vicsn: I think we need something clearer.

New idea: we can manually compile-guard certain tests with `#[cfg(feature = "prerelease")]`, and then we make one or more `jobs` to run those on the `release-workflow`. Maybe you can make a preliminary attempt based on the top 100 from `all-tests-by-time.csv`. For example:
- all the tests in `credits_migration_tests` or in `test_v*.rs` are about old migrations which (1) new functionality is unlikely to impact (2) typically won't impact security in production that much because they test historic migration behaviour.
- all the tests which mention `exhaustive` should only have to be run on releases.
- we have two `memory-intensive` tests which instead we can turn into a serial `prerelease` test.

meddle0x53:
Closing as other ideas were proposed.

PR #3103, opened 2026-01-15
Introduce pre-release test grouping

meddle0x53:
## Motivation

Grouping very slow  tests in a prerelease group, so they are ran only when necessary and are separated should make the normal PR-releated CI runs faster and cheaper.

The very slow tests are ignored by default with the `#[ignore]` tag, but are explicitly ran in special workflow that are only executed on merge in branches like staging/mainnet, etc. All workspaces that don't have `#[ignore]` tests explicitly run all tests, including the ignored ones, so if an ignored test is added to them it won't be skipped.
The workspaces with slow ignored tests are run without the ignored tests by default and only with the ignored tests in the special slower jobs.

## Test Plan

Run the CI and check if all passes as expected.

## Documentation

This PR uses a combination of `feature` - so the test are compiled only when they should be ran and `ignore` so they are never run, except when a specific option is passed. Special workflow is built for them.


vicsn:
I think the `/target/` folder can also become quite huge. Just consider how many GB you clean up when you cargo clean. We should probably not be caching it in circleci and only rely on sccache

meddle0x53:
Right, makes sense, let's see how much the compilation takes without it.

.circleci/config.yml
```diff
@@ -190,10 +188,61 @@ commands:
...
+                cache_key: v4.2.0-rust-1.88.0-<< parameters.workspace_member >><< parameters.cache_key_suffix >>-cache
+
+      - unless:
+          condition: << parameters.use_cache >>
+          steps:
+
+            - run:
+                name: Ensure clean build dirs (no cache mode)
+                command: |
+                  rm -rf target ~/.cache/sccache ~/.cargo/registry ~/.cargo/git || true
```
vicsn: I would think only this is needed when `!use_cache`. The other installations are generally useful or apply to when we `use_cache`

synthesizer/program/tests/instruction/hash.rs
```diff
@@ -394,29 +394,29 @@ test_hash!(hash_psd2_raw, HashPSD2Raw, ITERATIONS);
...
+test_hash!(hash_keccak256_native_raw, HashKeccak256NativeRaw, 2);
+test_hash!(hash_keccak384_native_raw, HashKeccak384NativeRaw, 1);
+test_hash!(hash_keccak512_native_raw, HashKeccak512NativeRaw, 1);
 
-test_hash!(hash_sha3_256_native_raw, HashSha3_256NativeRaw, 5);
-test_hash!(hash_sha3_384_native_raw, HashSha3_384NativeRaw, 5);
-test_hash!(hash_sha3_512_native_raw, HashSha3_512NativeRaw, 5);
+test_hash!(hash_sha3_256_native_raw, HashSha3_256NativeRaw, 2);
+test_hash!(hash_sha3_384_native_raw, HashSha3_384NativeRaw, 1);
+test_hash!(hash_sha3_512_native_raw, HashSha3_512NativeRaw, 1);
```
vicsn, line 419: @raychu86 you had mentioned last week that a single iteration of `<test _ $name _ is _ consistent>` foregoes it's purpose. However, that doesn't sound correct to me because in each iteration we sample fresh values. The array type is large, so we're effectively performing a lot of consistency checks, so we can do less iterations.

synthesizer/src/vm/tests/test_v9.rs
```diff
@@ -3104,6 +3104,8 @@ constructor:
 //  - parent programs that exceed the depth exist, but fail to execute.
 //  - the VM can be loaded from the store.
 #[test]
+#[cfg(feature = "prerelease")]
```
vicsn: I think we can more simply feature-guard:
```
#[cfg(feature = "test")]
#[cfg(feature = "prerelease")]
mod test_v8;

#[cfg(feature = "test")]
#[cfg(feature = "prerelease")]
mod test_v9;

#[cfg(feature = "test")]
#[cfg(feature = "prerelease")]
mod test_v10;

#[cfg(feature = "test")]
#[cfg(feature = "prerelease")]
mod test_v11;
```

vicsn (changes requested):

vicsn (changes requested):
`ledger-block` and `ledger` run 9 and 7 minutes longer respectively than any other job. Can we reduce or move some of it into prerelease as well? That would allow us to bring CI-time down to only 20 minutes.

.circleci/config.yml
meddle0x53: Actually to not change the shared code for running the tests it has to install sscache and use it as wrapper here too. It is done to keep the test run code unchanged and clear from `if use_cache` checks.

I think by just adding these two steps it is the cleanest way to support `use_cache/!use_cache` logic.

synthesizer/program/tests/instruction/hash.rs
raychu86: Obviously more iterations is "safer", but this is fine

raychu86:
There are so many special cases now with `ignore`s, feature flags, pre-releases. IMO I'm confused if the test I am writing is even run in CI. Is there a way to make it clear or at least check that we run each test at least one in some capacity?

meddle0x53:
@raychu86

If we use `ignore`:
1. When running non-ignored tests we will run only non-ignored tests - that's good
2. When we run `only` ignored tests (prerelease/merge workflows), only ignored tests are ran.

That way the slow tests are ignored by default and are only ran in specific branches.

vicsn:
> Is there a way to make it clear or at least check that we run each test at least one in some capacity?

@meddle0x53 for the jobs where we don't run any `--ignored` tests, could we do something silly in bash like `assert (cargo nextest --ignored --ignore-only --dry-run --list | wc -l) == 0` to ensure we didn't miss any ignored tests?

The other option is for **all** jobs without ignore tests to just run with `--ignored`, but this makes it harder to read the CircleCI config.

synthesizer/src/vm/mod.rs
```diff
@@ -1812,105 +1816,6 @@ function do:
         vm.add_next_block(&block).unwrap();
     }
 
-    #[test]
-    #[ignore]
-    fn test_deployment_memory_overload() {
```
vicsn, line 1817: FYI this was removed because it is duplicate with `test_deep_nested_execution_cost`

.circleci/config.yml
```diff
@@ -1,6 +1,9 @@
 version: 2.1
 
 parameters:
+  excluded_run_ignored_all_workspaces:
```
vicsn: What is this, why is it needed?

vicsn (commented):
Can you keep the PR README up to date?

.circleci/config.yml
vicsn: Is it possible to re-use the `setup_environment` step, perhaps passing the `use_cache` parameter to it? Or otherwise to create a separate `setup_environment_no_cache` step for consistency?

.circleci/config.yml
meddle0x53: These are the workspaces that are expected to have ignored tests so they are not ran with `all` by default (to not run their ignored tests by default), then their ignored tests are ran in the slow workflows.

I spent some time thinking how to make it so every workspace is run with both the ignored and non ignored tests by default except the ones we ignore the tests intentionally and the idea was to extract these in a place (at the top) they are visible. One of my local implementations had them hard-coded in the bash that runs the tests, but I thought this is better in the end.

synthesizer/src/vm/mod.rs
meddle0x53: I remember they were duplicated, yes.

.circleci/config.yml
vicsn: Thank you for explaining! I'm slightly against the added complexity but let's see what Ray likes best.

meddle0x53: Updated the README with a short description of this idea.

But it is not complex:
* For all workspaces always run all the tests, including the ignored ones (this way no test is missed)
* For workspaces with intentionally ignored test, exclude them from this "run all" logic, run all without the ignored tests by default.

This is the whole thing. I extracted these workspaces in this parameter to be very visible.

raychu86 (commented):
Generally fine with these changes, but it would be nice to have a better breakdown of the CI tests that are run and under what conditions. 

A README or table to track tests and have a high level overview would be much easier to track than reading the `config.yml` file. It would also bring clarity for where to write tests

vicsn:
@raychu86 

- simplified setup: d4b9fd6d4
- added README: 38c6a7fc6

CI.md
```diff
@@ -0,0 +1,19 @@
...
+are targeting a 15 minute max runtime.
+- `cargo-workflow`
+- `circuit-workflow`
+- `console-workflow`
+- `ledger-workflow`
+- `synthesizer-workflow`
+
+When a PR is merged, `merge-workflow` runs additional expensive or slow tests,
+e.g. those marked as `ignore` or which download paramters. Moreover, benchmarks
+are run from github actions.s
```
raychu86, line 16:
```suggestion
are run from github actions.
```

raychu86 (approved):

vicsn (approved):

PR #3106, opened 2026-01-16
Test job run costs

meddle0x53:
## Motivation

The idea is to measure the cost/time of running the same job in different types of containers, starting with `small`.

## Test Plan

Do a few pushes with different sizes of the same only-job, but with different container types.


vicsn:
Test succesful. `2xlarge` is twice as expensive as `xlarge`, but only completes the tests 25% faster.

PR #3107, opened 2026-01-16
A few fixes to external structs in records and futures

mohammadfawaz:
1. When checking that an external record matches its type, we need to use the external stack of the external program that contains the record. This is important because if the external record uses an external struct, we want to make sure that the external struct can be found. Previously, it didn't matter because, other than records, all types were visible everywhere. Now, that's changed.

2. Similar issue occurs with future arguments where, before comparing them when they are `Plaintext`, we have to select the right stack.

3. One more issue related to futures in `RegisterTypes::get_type` where finalize inputs need to be qualified with the program name when they are `Plaintext` from another program.

### DIFF
--- a/synthesizer/process/src/stack/finalize_types/mod.rs
+++ b/synthesizer/process/src/stack/finalize_types/mod.rs
@@ -161,7 +161,7 @@ impl<N: Network> FinalizeTypes<N> {
                 }
                 // Access the member on the path to output the register type.
                 (FinalizeType::Plaintext(PlaintextType::ExternalStruct(locator)), Access::Member(identifier)) => {
-                    // Retrieve the member type from the external struct.
+                    // Retrieve the member type from the external struct and check that it exists.
                     let external_stack = stack.get_external_stack(locator.program_id())?;
                     match external_stack.program().get_struct(locator.resource())?.members().get(identifier) {
                         // Qualify local struct references so subsequent accesses use the correct stack.
--- a/synthesizer/process/src/stack/finalize_types/mod.rs
+++ b/synthesizer/process/src/stack/finalize_types/mod.rs
@@ -161,11 +161,14 @@ impl<N: Network> FinalizeTypes<N> {
                 }
                 // Access the member on the path to output the register type.
                 (FinalizeType::Plaintext(PlaintextType::ExternalStruct(locator)), Access::Member(identifier)) => {
-                    // Retrieve the member type from the struct and check that it exists.
+                    // Retrieve the member type from the external struct.
                     let external_stack = stack.get_external_stack(locator.program_id())?;
                     match external_stack.program().get_struct(locator.resource())?.members().get(identifier) {
-                        // Retrieve the member and update `finalize_type` for the next iteration.
-                        Some(member_type) => finalize_type = FinalizeType::Plaintext(member_type.clone()),
+                        // Qualify local struct references so subsequent accesses use the correct stack.
+                        Some(member_type) => {
+                            let qualified = member_type.clone().qualify(*locator.program_id());
+                            finalize_type = FinalizeType::Plaintext(qualified);
+                        }
                         // Halts if the member does not exist.
                         None => bail!("'{identifier}' does not exist in struct '{locator}'"),
                     }
--- a/synthesizer/process/src/stack/helpers/stack_trait.rs
+++ b/synthesizer/process/src/stack/helpers/stack_trait.rs
@@ -578,14 +578,19 @@ impl<N: Network> Stack<N> {
         ensure!(future.arguments().len() == inputs.len(), "Future arguments do not match");
 
         // Check that the arguments match the inputs.
+        // Use the external stack if the future is from an external program.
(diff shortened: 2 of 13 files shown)


synthesizer/process/src/stack/finalize_types/mod.rs
```diff
@@ -161,11 +161,14 @@ impl<N: Network> FinalizeTypes<N> {
                 }
                 // Access the member on the path to output the register type.
                 (FinalizeType::Plaintext(PlaintextType::ExternalStruct(locator)), Access::Member(identifier)) => {
-                    // Retrieve the member type from the struct and check that it exists.
+                    // Retrieve the member type from the external struct.
```
d0cd:
```suggestion
                    // Retrieve the member type from the external struct and check that it exists.
```

synthesizer/process/src/stack/helpers/stack_trait.rs
```diff
@@ -578,14 +578,19 @@ impl<N: Network> Stack<N> {
         ensure!(future.arguments().len() == inputs.len(), "Future arguments do not match");
 
         // Check that the arguments match the inputs.
+        // Use the external stack if the future is from an external program.
```
d0cd, line 581: Is there a forking risk in fixing this implementation? 
Specifically, is there a chance an execution that would once fail would now be accepted by nodes that update to v4.5.0?

mohammadfawaz: I think this goes back to global namespacing of structs before we added external structs. Was that truly the case? From the examples I've seen, it looks like we were able to use the main program stack to access a struct defined in another program as long as that program is imported in our main program. We should confirm this though in 4.4.0. I will do that today.

vicsn: > Specifically, is there a chance an execution that would once fail would now be accepted by nodes that update to v4.5.0?

I'd like this PR to go into v4.5.0
We can accept any forking risk arising from going from `canary-v4.5.0` to `canary-v4.5.1`
And if we miss a corner case it is not the end of the world, there is a forking risk but it is in a very limited timewindow.

When reviewing `contains_external_struct` I noticed some potential deficiencies. It never uses a `Process` for resolution, and always returns `false` for `PlaintextType::Struct` and `RegisterType::{Record, ExternalRecord, Future}`. Traversing our full AST is just very cumbersome.

mohammadfawaz: What I'm thinking is I'll change all `contains_external_struct` functions to accept a `get_struct` closure such as:
```
    pub fn contains_external_struct<F>(&self, get_struct: &F) -> bool
    where
        F: Fn(&Identifier<N>) -> Result<&StructType<N>>,
 ```
 and then I'll eventually pass in `ProgramCore::get_struct` from the top level. `get_struct` will then give me access to the struct members which I'll recursively check.

Does this make sense?

Comment (vicsn) [reply synthesizer/process/src/stack/helpers/stack_trait.rs:581]: I think so yes! I think we use a somewhat similar pattern in other places. At least worth experimenting, and it also sets a better precedent for future snarkVM AST code analysis (is AST even the right word?)

Comment (mohammadfawaz) [reply synthesizer/process/src/stack/helpers/stack_trait.rs:581]: Ok so I spent some time doing this and it works fine.. However, it looks like what we had before was also safe due to this:
```rust
impl<N: Network> ProgramCore<N> {
    ...
    pub fn contains_external_struct(&self) -> bool {
        self.mappings.values().any(|mapping| mapping.contains_external_struct())
            || self
                .structs
                .values()
                .flat_map(|struct_| struct_.members().values())
                .any(|plaintext_type| plaintext_type.contains_external_struct())
            || self
                .records
                .values()
                .flat_map(|record| record.entries().values())
                .any(|entry| entry.plaintext_type().contains_external_struct())
            || self.closures.values().any(|closure| closure.contains_external_struct())
            || self.functions.values().any(|function| function.contains_external_struct())
            || self.constructor.iter().any(|constructor| constructor.contains_external_struct())
    }
    ...
}
```
Local struct members are already checked. So even if encounter a reference to a local struct with external struct members, and decide to ignore it like we currently do, we will still error out due to the check above.

Now, to be 100% correct, yes `contains_external_struct` currently may return incorrect result when called on a `PlaintextType`. However, because all `contains_external_struct` functions currently only serve the single top level `ProgramCore::contains_external_struct` function, it really doesn't matter.

vicsn: What about a function returning an `ExternalRecord` whose member contains an `ExternalRecord`?

mohammadfawaz: Right so you're thinking 
program A depends on program B and program B uses an external struct from C. In that case, if we're on consensus V11, then we will fail to deploy B, right? So A cannot be deployed. Or is there a way to bypass some check to make this work somehow?

vicsn (approved):

PR #3110, opened 2026-01-20
Resolve dynamic dispatch todos

vicsn:
## Motivation

Resolves a few of the outstanding dynamic dispatch TODOs, can be reviewed commit-per commit.


synthesizer/process/src/verify_execution.rs
```diff
@@ -410,8 +421,19 @@ impl<N: Network> Process<N> {
                 }
 
                 caller_inputs
+                    .iter()
+                    .zip(child_transition.inputs().iter())
+                    .map(|(caller_input, child_input)| {
+                        match (caller_input, child_input) {
+                            // Equivalence between caller and callee is assured through the translation circuit.
+                            (Input::DynamicRecord(_), Input::Record(..) | Input::ExternalRecord(_)) => caller_input,
```
vicsn: I will change this to use `call_dynamic_instruction_opt.clone().unwrap().operand_types()` instead of the untrusted `caller_input`for this comparison.

vicsn: 141318443

synthesizer/process/src/trace/mod.rs
```diff
@@ -504,11 +504,10 @@ impl<N: Network> Trace<N> {
...
 
         let expected_n_inclusions = Authorization::number_of_input_records(transitions);
+        let actual_n_inclusions = batch_inclusion_inputs.len();
         ensure!(
-            batch_inclusion_inputs.len() == expected_n_inclusions,
-            "Unexpected number of inclusion inputs: {} instead of {}",
-            batch_inclusion_inputs.len(),
-            expected_n_inclusions
+            actual_n_inclusions == expected_n_inclusions,
+            "Unexpected inclusion inputs: {actual_n_inclusions} v.s. {expected_n_inclusions}"
```
Antonio95: "Unexpected inclusion inputs" sounds like the values of the inputs were unexpected; or perhaps like there were more "inclusion inputs" (whatever that could mean) than expected, as opposed to either too many or too few, as this is checking.

Perhaps "Unexpected number of inclusion inputs" or "Unexpected number of inclusion instances"?

synthesizer/process/src/verify_execution.rs
```diff
@@ -413,8 +416,24 @@ impl<N: Network> Process<N> {
...
+                            // Equivalence between caller and callee is assured through the translation circuit.
+                            (
+                                Input::DynamicRecord(_),
+                                ValueType::DynamicRecord,
+                                Input::Record(..) | Input::ExternalRecord(_),
+                            ) => caller_input,
+                            // Equivalence between caller and callee is assured by using the child_transition inputs directly.
+                            _ => child_input,
+                        }
+                    })
```
Antonio95: This change is great, good that you caught the issue (of just taking `caller_inputs` as we were before).

I hope the reviewers of the main DD PR will weigh in on whether to keep the duplicate entries in `CallerMetadata` or not: as the second `match` arm here shows, they are irrelevant and, as usual, redundancy can lead to inconsistency. In any case, this check/selection is the best we can do atm.

Antonio95: To iterate on the comment above: once/if we remove the duplicate entries, the zip with `caller_inputs` will go away; the first arm here will translate into popping from `caller_inputs` (or advancing a counter); and at the end we'll have to check that `caller_inputs` is empty (or that the counter points to its last entry). The same ideas as above also go for the outputs case below.

synthesizer/process/src/verify_execution.rs
```diff
@@ -413,8 +416,24 @@ impl<N: Network> Process<N> {
...
+                    .collect::<Vec<_>>()
             } else {
-                child_transition.inputs()
+                child_transition.inputs().iter().collect::<Vec<_>>()
```
Antonio95: Been messing with this for 5 mins to try and make it look prettier to no avail... well played

synthesizer/process/src/verify_execution.rs
```diff
@@ -413,8 +416,24 @@ impl<N: Network> Process<N> {
...
+                    .zip(child_transition.inputs().iter())
+                    .map(|((caller_input, operand_type), child_input)| {
+                        match (caller_input, operand_type, child_input) {
+                            // Equivalence between caller and callee is assured through the translation circuit.
+                            (
+                                Input::DynamicRecord(_),
+                                ValueType::DynamicRecord,
+                                Input::Record(..) | Input::ExternalRecord(_),
+                            ) => caller_input,
+                            // Equivalence between caller and callee is assured by using the child_transition inputs directly.
```
Antonio95: Given this comment, I think we can make an explicit check. The changes are minimal: the `map` block is replaced by:

```
.map(|((caller_input, operand_type), child_input)| {
    match (caller_input, operand_type, child_input) {
        // Equivalence between caller and callee is assured through the translation circuit.
        (
            Input::DynamicRecord(_),
            ValueType::DynamicRecord,
            Input::Record(..) | Input::ExternalRecord(_),
        ) => Ok(caller_input),
        // Equivalence between caller and callee checked explicitly.
        _ => {
            ensure!(
                caller_input == child_input,
                "Caller and callee input in dynamic call to {} do not match: {caller_input} != {child_input}",
                transition.function_name()
            );
            Ok(child_input)
        },
    }
})
.collect::<Result<Vec<_>>>()?
```
Happy to add it and/or add the output counterpart (which also has to cover the separately treated dynamic-future case). Also okay if you disregard it: taking the callee inputs is the right thing to do anyway.

Antonio95 (approved):
Left a couple of minor comments. Happy to give this the green light even if those are disregarded.

synthesizer/process/src/verify_execution.rs
d0cd: I'm currently working on an alternate design that gets rid of `CallerMetadata` and keeps the dynamic IDs next to the `Input` and `Output`

vicsn: Thx guys! Let me know when that other approach is finalized/rejected and after I'll continue this PR.

d0cd: @vicsn The other approach has been approved, feel free to continue off the new changes.

synthesizer/process/src/verify_execution.rs
```diff
@@ -232,8 +238,15 @@ impl<N: Network> Process<N> {
                 self.get_stack(program_id).and_then(|stack| stack.get_translation_verifying_key(record_name))
             },
         )?;
-
-        // TODO (@vicsn): bring appropriate new measurement functions from execution_cost_for_authorization to here.
+        // Ensure `Authorization::translation_batches` matches the number of translations.
+        // We only compare the totals because preserving order is complex.
```
d0cd: Should we also mention that proof verification checks the order because the prover and verifier agree on the `translation_index`?

Feel free to phrase better.

d0cd (approved):
LGTM, minor nit

synthesizer/process/src/verify_execution.rs
vicsn: a1d6022a1

vicsn:
@d0cd pushed another set of commits to re-enable recursion, which I recall was still on our wishlist.

d0cd (approved):
LGTM!

PR #3117, opened 2026-01-21
[Feature] Add `Ledger::try_get_*` for transmissions

kaimast:
### Motivation
During sync, we want to differentiate between a missing transmission and some kind of storage corruption. As of now, these two outcomes are lumped together in `anyhow::Result`, which is not machine-readable.  Similar to #2974, but for transmissions. 

###  Proposed Changes
The PR adds `try_get_transaction`, `try_get_confirmed_transaction`, `try_get_unconfirmed_transaction`  and `try_get_solution` to `Ledger`. The existing getters are not changed to avoid API breakage.
It builds on top of #3070 as both are needed for sync improvements.

The PR also adds `#[deny(missing_docs)]` to some parts of `Ledger`. Ideally the entire `snarkvm-ledger` crate should be documented, but this is a good start to ensure new getters, like the ones added by this PR, are documented.

### DIFF
--- a/ledger/query/src/query.rs
+++ b/ledger/query/src/query.rs
@@ -201,10 +201,9 @@ impl<N: Network, B: BlockStorage<N>> Query<N, B> {
     /// Returns the transaction for the given transaction ID.
     pub fn get_transaction(&self, transaction_id: &N::TransactionID) -> Result<Transaction<N>> {
         match self {
-            Self::VM(block_store) => {
-                let txn = block_store.get_transaction(transaction_id)?;
-                txn.ok_or_else(|| anyhow!("Transaction {transaction_id} not in local storage"))
-            }
+            Self::VM(block_store) => block_store
+                .get_transaction(transaction_id)?
+                .ok_or_else(|| anyhow!("Missing transaction '{transaction_id}' in block storage")),
             Self::REST(query) => query.get_transaction(transaction_id),
             Self::STATIC(_query) => bail!("get_transaction is not supported by StaticQuery"),
         }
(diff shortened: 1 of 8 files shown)


ledger/query/src/query.rs
```diff
@@ -201,10 +201,9 @@ impl<N: Network, B: BlockStorage<N>> Query<N, B> {
     /// Returns the transaction for the given transaction ID.
     pub fn get_transaction(&self, transaction_id: &N::TransactionID) -> Result<Transaction<N>> {
         match self {
-            Self::VM(block_store) => {
-                let txn = block_store.get_transaction(transaction_id)?;
-                txn.ok_or_else(|| anyhow!("Transaction {transaction_id} not in local storage"))
-            }
+            Self::VM(block_store) => block_store
+                .get_transaction(transaction_id)?
+                .ok_or_else(|| anyhow!("Missing transaction '{transaction_id}' in block storage")),
```
vicsn, line 206: I asked many times before, please stop frivolous changes, such as log message rewording, it is a waste of everyone's time. That said, don't bother reverting it either.

ledger/src/advance.rs
```diff
@@ -17,6 +17,7 @@ use super::*;
 
 use anyhow::Context;
 
+#[deny(missing_docs)]
```
vicsn: This is out of scope for this PR and has no agreed upon strategy. Please remove.

vicsn (changes requested):
Core logic LGTM! The doc compile guards need a new issue.

ledger/src/advance.rs
kaimast: Done. We already do this, for example, in `snarkos-node-tcp`, but, you are right, it should be a separate issue.

vicsn (approved):

PR #3122, opened 2026-01-26
[Refactor] Isolate a synthesizer-error crate

ljedrz:
This PR facilitates concrete error use in crates which would otherwise run into circular dependencies (e.g. https://github.com/ProvableHQ/snarkVM/pull/3101).

The only "internal" dependencies of the new crate are `circuit-environment` and `console-network`, and they should remain minimal to ensure broad applicability.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -3344,6 +3344,7 @@ dependencies = [
  "snarkvm-ledger-store",
  "snarkvm-ledger-test-helpers",
  "snarkvm-synthesizer",
+ "snarkvm-synthesizer-error",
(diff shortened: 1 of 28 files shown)


vicsn (approved):
@raychu86 is your release process robust against new crates? I recall it skipped over a new one once.

raychu86:
The release process needs to be updated manually if there are new scripts. I had intended to add automated checks, but never got around to it since crate additions are fairly rare.

mitchmindtree (approved):
Apologies for the delay, LGTM :shipit:

PR #3123, opened 2026-01-26
Consensus version check in `Process` and `Stack`

mohammadfawaz:
Introduces consensus check guards on changes made in https://github.com/ProvableHQ/snarkVM/pull/3107 which do not strictly depend on external structs.

The consensus version is stored as an `AtomicU16` in the `VM` object which is later referenced using a `Weak` reference from `Process` and `Stack`. The atomic is updated every time a block is added. `Stack` now has access to the current consensus version by dereferencing the weak pointer.

### DIFF
--- a/console/network/src/consensus_heights.rs
+++ b/console/network/src/consensus_heights.rs
@@ -15,6 +15,7 @@
 
 use crate::{FromBytes, ToBytes, io_error};
 
+use anyhow::anyhow;
 use enum_iterator::{Sequence, last};
 use std::io;
 
@@ -79,9 +80,36 @@ impl FromBytes for ConsensusVersion {
     }
 }
 
+impl TryFrom<u16> for ConsensusVersion {
+    type Error = anyhow::Error;
+
+    fn try_from(value: u16) -> anyhow::Result<Self> {
+        match value {
+            1 => Ok(ConsensusVersion::V1),
+            2 => Ok(ConsensusVersion::V2),
+            3 => Ok(ConsensusVersion::V3),
+            4 => Ok(ConsensusVersion::V4),
+            5 => Ok(ConsensusVersion::V5),
+            6 => Ok(ConsensusVersion::V6),
+            7 => Ok(ConsensusVersion::V7),
+            8 => Ok(ConsensusVersion::V8),
+            9 => Ok(ConsensusVersion::V9),
+            10 => Ok(ConsensusVersion::V10),
+            11 => Ok(ConsensusVersion::V11),
+            12 => Ok(ConsensusVersion::V12),
+            13 => Ok(ConsensusVersion::V13),
+            _ => Err(anyhow!("Invalid consensus version discriminant: {value}")),
+        }
+    }
+}
+
 impl ConsensusVersion {
     pub fn latest() -> Self {
         last::<ConsensusVersion>().expect("At least one ConsensusVersion should be defined.")
+    }
+
+    pub fn to_u16(self) -> u16 {
+        self as u16
     }
 }
 
--- a/parameters/examples/setup.rs
+++ b/parameters/examples/setup.rs
@@ -15,7 +15,13 @@
 
 use snarkvm_algorithms::crypto_hash::sha256::sha256;
 use snarkvm_circuit::Aleo;
-use snarkvm_console::network::{CanaryV0, MainnetV0, Network, TestnetV0, prelude::ToBytes};
+use snarkvm_console::network::{
+    CanaryV0,
+    MainnetV0,
+    Network,
+    TestnetV0,
+    prelude::{ConsensusVersion, ToBytes},
+};
 use snarkvm_synthesizer::{Process, Program};
 
 use anyhow::Result;
@@ -89,7 +95,7 @@ pub fn credits_program<N: Network, A: Aleo<Network = N>>() -> Result<()> {
     // Initialize an RNG.
     let rng = &mut snarkvm_utilities::TestRng::fixed(1245897092);
     // Initialize the process.
-    let process = Process::setup::<A, _>(rng)?;
+    let process = Process::setup::<A, _>(std::sync::Arc::new(|| Ok(ConsensusVersion::latest())), rng)?;
     // Initialize the program.
     let program = Program::<N>::credits()?;
     let program_id = program.id();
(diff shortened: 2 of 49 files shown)


synthesizer/process/src/lib.rs
```diff
@@ -86,6 +86,7 @@ use std::{collections::HashMap, sync::Arc};
 
 #[derive(Clone)]
 pub struct Process<N: Network> {
+    get_consensus_version: Arc<dyn Fn() -> anyhow::Result<ConsensusVersion> + Send + Sync>,
```
d0cd: In practice, `get_consensus_version` will need to make a query to the block store. This could result in a fair bit of overhead. What are the tradeoffs with an imperative design that explicitly stores the consensus version?

mohammadfawaz: Yeah that is a bit of an issue I agree. The challenge would be: how do we update the consensus version inside each `Stack` when blocks advance? Would that also be an expensive operation?

d0cd: Perhaps the weak reference approach, similar to how stacks can refer back to the map of stacks in the process.

mohammadfawaz: Alright so I switched to using an atomic that lives in the VM. That atomic gets updated in `add_next_block_inner`. `Process` and `Stack` have a weak reference to it.
1. Is updating this value in `add_next_block_inner` enough?
2. `AtomicU16` should perform a lot better than `RwLock<ConsensusVersion>`.

console/network/src/consensus_heights.rs
```diff
@@ -79,10 +80,37 @@ impl FromBytes for ConsensusVersion {
...
+        }
+    }
+}
+
 impl ConsensusVersion {
     pub fn latest() -> Self {
         last::<ConsensusVersion>().expect("At least one ConsensusVersion should be defined.")
     }
+
+    pub fn to_u16(self) -> u16 {
```
d0cd, line 111: Do you need this if it already implements `repr(u16)`?

d0cd:
Discussed with @raychu86 and the feedback was that this approach breaks the existing layers of abstraction. Process doesn't have a notion of "block height" and introducing it here allows for there to be inconsistencies. 

@mohammadfawaz is there a way we could design a static check for the forking conditions and restrict them until `V14`. 
Something like: for every operand in the program, for every access in the operand, if the access is to a struct local to an external program, it is not allowed.

raychu86 (changes requested):
Couple reasons I don't believe ConsensusVersion should exist in Process:

1. Process has always been stateless and not have the concept of Ledgers and block heights

2. Now custom logic is in different places (Process AND TX verification)process.finalize_execution requires passing in expected state behavior, but now what happens if the FinalizeGlobalState block height differs from what Process has for ConsensusVersion? Which is ground truth? Which rules are used

3. More process initialization pathways makes it much more confusing, especially for devs and service providers. For instance delegated proving.

Ideally to keep with the existing custom, we do the new rule check in vm.check_transaction and Process just supports the latest features, just state added is gated via verification logic from the VM side.

parameters/examples/setup.rs
```diff
@@ -89,7 +96,7 @@ pub fn credits_program<N: Network, A: Aleo<Network = N>>() -> Result<()> {
     // Initialize an RNG.
     let rng = &mut snarkvm_utilities::TestRng::fixed(1245897092);
     // Initialize the process.
-    let process = Process::setup::<A, _>(rng)?;
+    let process = Process::setup::<A, _>(Arc::clone(&Arc::new(AtomicU16::new(ConsensusVersion::latest().to_u16()))), rng)?;
```
ljedrz, line 99: this is very weird; why not just `Arc::new(AtomicU16::new(ConsensusVersion::latest() as u16))`?

synthesizer/process/src/lib.rs
```diff
@@ -82,10 +82,14 @@ use indexmap::IndexMap;
...
 use parking_lot::RwLock;
-use std::{collections::HashMap, sync::Arc};
+use std::{
+    collections::HashMap,
+    sync::{Arc, Weak, atomic::AtomicU16},
+};
 
 #[derive(Clone)]
 pub struct Process<N: Network> {
+    consensus_version: Weak<AtomicU16>,
```
ljedrz, line 92: since this is a `Weak`, then who is the holder of the `Arc`?

synthesizer/process/src/lib.rs
```diff
@@ -262,14 +278,25 @@ impl<N: Network> Process<N> {
         Ok(process)
     }
 
+    /// Initializes a new process.
+    #[inline]
+    pub fn load() -> Result<Self> {
+        Self::load_with_consensus_version(Arc::new(AtomicU16::new(ConsensusVersion::latest().to_u16())))
```
ljedrz, line 284: this would cause the `Weak` in the `Process` to never resolve to the actual value (the source `Arc` will immediately be destroyed)

parameters/examples/setup.rs
ljedrz: also, I now see that this will cause the `Process` to never produce the consensus version - `Weak::upgrade` would always resolve to `None`

synthesizer/process/src/tests/test_execute.rs
```diff
@@ -2485,6 +2485,9 @@ fn test_process_deploy_credits_program() {
 
     // Initialize an empty process without the `credits` program.
     let empty_process = Process {
+        consensus_version: Arc::downgrade(&Arc::new(std::sync::atomic::AtomicU16::new(
```
ljedrz, line 2488: this is wrong - the `Weak` would never produce a value

ljedrz (changes requested):
There is some invalid use of `Weak` involved (see this [playground example](https://play.rust-lang.org/?version=stable&mode=debug&edition=2024&gist=ae2806b9964f3ab434b453dfe64720c6) for a minimal demonstration); whenever there is a `Weak`, there **must** be some live `Arc` available elsewhere, otherwise it will immediately be useless, and whenever there can't be a link between separate objects, there shouldn't be shared ownership between them at all.

niklaslong:
+1 on not coupling the consensus version to the `Process`, this would make it harder to maintain dependents who rely on this structure directly with the assumption that its internal behaviour can't change based on block height (and the behaviour that does is explicit in its method inputs).

mohammadfawaz:
Thank you all. This all makes sense to me. @d0cd and I discussed a new direction that involves purely static checks.

PR #3127, opened 2026-01-28
[Perf] A revival of 1955 & 2145

ljedrz:
This PR is a fresh implementation of https://github.com/ProvableHQ/snarkVM/pull/1955 combined with https://github.com/ProvableHQ/snarkVM/pull/2145, which built on it. Both were contained in a single commit each, and a few adjustments were necessary in order to work with the current codebase, but nothing too complex.

The affected benchmarks are:
```
Transaction::Execute(transfer_public) - verify
                        time:   [41.792 µs 41.808 µs 41.823 µs]
                        change: [−0.8837% −0.8428% −0.8003%] (p = 0.00 < 0.05)
// note: there were some outliers, but it would likely have been marked as an improvement without them

Transaction::Execute(transfer_private) - verify
                        time:   [46.640 µs 46.680 µs 46.718 µs]
                        change: [−2.3414% −2.2890% −2.2178%] (p = 0.00 < 0.05)
                        Performance has improved.

snark_certificate_prove_100000
                        time:   [109.25 ms 109.72 ms 110.19 ms]
                        change: [−3.3583% −2.7577% −2.1394%] (p = 0.00 < 0.05)
                        Performance has improved.

snark_certificate_verify_100000
                        time:   [93.865 ms 94.126 ms 94.391 ms]
                        change: [−2.3859% −1.9702% −1.5179%] (p = 0.00 < 0.05)
                        Performance has improved.
```

Filing as a draft, as backward compatibility (at least some serialization) still needs to be resolved.

CC @niklaslong

### DIFF
--- a/algorithms/src/polycommit/kzg10/degree_info.rs
+++ b/algorithms/src/polycommit/kzg10/degree_info.rs
@@ -13,21 +13,20 @@
 // See the License for the specific language governing permissions and
 // limitations under the License.
 
-use std::collections::HashSet;
+use std::collections::BTreeSet;
 
-#[derive(Clone)]
+#[derive(Clone, Debug, Default)]
 pub struct DegreeInfo {
     /// The maximum degree of the required SRS to commit to the polynomials.
     pub max_degree: usize,
     /// The maximum IOP poly degree used for (i)fft_precomputation.
     pub max_fft_size: usize,
-    /// TODO (howardwu): This should never be happening... Use a BTreeSet.
     /// The degree bounds on IOP polynomials.
-    pub degree_bounds: Option<HashSet<usize>>,
+    pub degree_bounds: Option<BTreeSet<usize>>,
     /// The hiding bound for polynomial queries.
     pub hiding_bound: usize,
     /// The supported sizes for the lagrange-basis SRS.
-    pub lagrange_sizes: Option<HashSet<usize>>,
+    pub lagrange_sizes: Option<BTreeSet<usize>>,
 }
 
 impl DegreeInfo {
@@ -35,9 +34,9 @@ impl DegreeInfo {
     pub const fn new(
         max_degree: usize,
         max_fft_size: usize,
-        degree_bounds: Option<HashSet<usize>>,
+        degree_bounds: Option<BTreeSet<usize>>,
         hiding_bound: usize,
-        lagrange_sizes: Option<HashSet<usize>>,
+        lagrange_sizes: Option<BTreeSet<usize>>,
     ) -> Self {
         Self { max_degree, max_fft_size, degree_bounds, hiding_bound, lagrange_sizes }
     }
--- a/algorithms/src/polycommit/kzg10/degree_info.rs
+++ b/algorithms/src/polycommit/kzg10/degree_info.rs
@@ -1,0 +1,63 @@
+// Copyright (c) 2019-2026 Provable Inc.
+// This file is part of the snarkVM library.
+
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at:
+
+// http://www.apache.org/licenses/LICENSE-2.0
+
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+use std::collections::HashSet;
+
+#[derive(Clone)]
+pub struct DegreeInfo {
+    /// The maximum degree of the required SRS to commit to the polynomials.
+    pub max_degree: usize,
+    /// The maximum IOP poly degree used for (i)fft_precomputation.
+    pub max_fft_size: usize,
+    /// TODO (howardwu): This should never be happening... Use a BTreeSet.
+    /// The degree bounds on IOP polynomials.
+    pub degree_bounds: Option<HashSet<usize>>,
+    /// The hiding bound for polynomial queries.
+    pub hiding_bound: usize,
+    /// The supported sizes for the lagrange-basis SRS.
+    pub lagrange_sizes: Option<HashSet<usize>>,
+}
+
+impl DegreeInfo {
+    /// Initializes a new degree info.
+    pub const fn new(
+        max_degree: usize,
+        max_fft_size: usize,
+        degree_bounds: Option<HashSet<usize>>,
+        hiding_bound: usize,
+        lagrange_sizes: Option<HashSet<usize>>,
+    ) -> Self {
+        Self { max_degree, max_fft_size, degree_bounds, hiding_bound, lagrange_sizes }
+    }
+}
+
+impl DegreeInfo {
+    pub fn union(self, other: &Self) -> Self {
+        let max_degree = self.max_degree.max(other.max_degree);
+        let max_fft_size = self.max_fft_size.max(other.max_fft_size);
+        let degree_bounds = match (&self.degree_bounds, &other.degree_bounds) {
+            (Some(a), Some(b)) => Some(a | b),
+            (Some(a), None) | (None, Some(a)) => Some(a.clone()),
+            (None, None) => None,
+        };
+        let hiding_bound = self.hiding_bound.max(other.hiding_bound);
+        let lagrange_sizes = match (&self.lagrange_sizes, &other.lagrange_sizes) {
+            (Some(a), Some(b)) => Some(a | b),
+            (Some(a), None) | (None, Some(a)) => Some(a.clone()),
+            (None, None) => None,
+        };
+        Self::new(max_degree, max_fft_size, degree_bounds, hiding_bound, lagrange_sizes)
+    }
+}
--- a/algorithms/src/polycommit/sonic_pc/data_structures.rs
+++ b/algorithms/src/polycommit/sonic_pc/data_structures.rs
@@ -91,9 +91,9 @@ impl<E: PairingEngine> FromBytes for CommitterKey<E> {
         }
 
         // Deserialize `lagrange_basis_at_beta`.
-        let has_lagrange_bases: bool = FromBytes::read_le(&mut reader)?;
+        let has_lagrange_bases = bool::read_le(&mut reader);
         let lagrange_bases_at_beta_g = match has_lagrange_bases {
-            true => {
+            Ok(true) => {
                 let lagrange_bases_at_beta_len: u32 = FromBytes::read_le(&mut reader)?;
                 let mut lagrange_bases_at_beta_g = BTreeMap::new();
                 for _ in 0..lagrange_bases_at_beta_len {
@@ -107,7 +107,9 @@ impl<E: PairingEngine> FromBytes for CommitterKey<E> {
                 }
                 Some(lagrange_bases_at_beta_g)
             }
-            false => None,
+            Ok(false) => None,
+            Err(e) if e.kind() == io::ErrorKind::UnexpectedEof => None,
+            Err(e) => return Err(e),
         };
 
         // Deserialize `powers_of_beta_times_gamma_g`.
--- a/algorithms/src/polycommit/sonic_pc/data_structures.rs
+++ b/algorithms/src/polycommit/sonic_pc/data_structures.rs
@@ -34,6 +34,7 @@ use std::{
     collections::{BTreeMap, BTreeSet},
     fmt,
     io,
+    mem,
     ops::{AddAssign, MulAssign, SubAssign},
 };
 
@@ -54,8 +55,8 @@ pub struct CommitterKey<E: PairingEngine> {
     pub powers_of_beta_g: Vec<E::G1Affine>,
 
     /// The key used to commit to polynomials in Lagrange basis.
-    /// This is `None` if `self` does not support lagrange bases
-    pub lagrange_bases_at_beta_g: Option<BTreeMap<usize, Vec<E::G1Affine>>>,
+    /// This is empty if `self` does not support lagrange bases
+    pub lagrange_bases_at_beta_g: BTreeMap<usize, Vec<E::G1Affine>>,
 
     /// The key used to commit to hiding polynomials.
     pub powers_of_beta_times_gamma_g: Vec<E::G1Affine>,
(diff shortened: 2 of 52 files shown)


algorithms/src/polycommit/sonic_pc/data_structures.rs
```diff
@@ -46,7 +54,8 @@ pub struct CommitterKey<E: PairingEngine> {
     pub powers_of_beta_g: Vec<E::G1Affine>,
 
     /// The key used to commit to polynomials in Lagrange basis.
-    pub lagrange_bases_at_beta_g: BTreeMap<usize, Vec<E::G1Affine>>,
+    /// This is `None` if `self` does not support lagrange bases
```
iamalwaysuncomfortable:
```suggestion
    /// This is `None` if `self` does not support lagrange bases.
```

algorithms/src/polycommit/sonic_pc/data_structures.rs
```diff
@@ -275,44 +296,171 @@ impl<E: PairingEngine> ToBytes for CommitterKey<E> {
...
-    pub powers_of_beta_times_gamma_g: Option<&'a Vec<E::G1Affine>>,
+    /// Update the committer_key
+    /// The `powers_of_beta_g` and `shifted_powers_of_beta_g` may grow or shrink
+    /// as per the new supported_degree The `enforced_degree_bounds` and
+    /// `shifted_powers_of_beta_times_gamma_g` will adjust to the new
+    /// degree_bounds The `lagrange_bases_at_beta_g` will optionally adjust
+    /// to the new degree_bounds The `powers_of_beta_times_gamma_g` is only
+    /// dependent on the `hiding_bound` so doesn't change This only works if
+    /// the SRS max_degree is fixed across specializations Note that this
+    /// implementation is not atomic. If specialize fails halfway through,
```
iamalwaysuncomfortable, line 309: This comment should be completed.

algorithms/src/polycommit/sonic_pc/data_structures.rs
```diff
@@ -275,44 +296,171 @@ impl<E: PairingEngine> ToBytes for CommitterKey<E> {
...
+    pub fn update(&mut self, srs: &UniversalParams<E>, degree_info: &DegreeInfo) -> Result<()> {
+        let trim_time = start_timer!(|| "Trimming public parameters");
+
+        // The maximum degree required by the circuit polynomials
```
iamalwaysuncomfortable, line 313: @copilot please add a commit that adds periods to all doc comments.

algorithms/src/polycommit/kzg10/degree_info.rs
```diff
@@ -0,0 +1,62 @@
...
+
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+use std::collections::BTreeSet;
+
+#[derive(Clone, Debug, Default)]
```
iamalwaysuncomfortable, line 18:
```suggestion
/// Degree info specifies information needed to determine which powers to use to build the CommitterKey from the SRS.
#[derive(Clone, Debug, Default)]
```

algorithms/src/polycommit/sonic_pc/data_structures.rs
```diff
@@ -275,44 +296,171 @@ impl<E: PairingEngine> ToBytes for CommitterKey<E> {
...
+                        let lowest_shift_degree = max_degree - new_highest_degree_bound;
+                        let highest_shift_degree = max_degree + 1 - highest_degree_bound;
+                        let mut new_powers = srs.powers_of_beta_g(lowest_shift_degree, highest_shift_degree)?;
+                        new_powers.extend_from_slice(&shifted_powers_of_beta_g);
+                        shifted_powers_of_beta_g = new_powers;
+                    }
+                    Ordering::Equal => (), // No need to adjust shifted_powers_of_beta_g
+                    Ordering::Less => {
+                        let degree_diff = highest_degree_bound - new_highest_degree_bound;
+                        shifted_powers_of_beta_g = shifted_powers_of_beta_g.drain(degree_diff - 1..).collect_vec();
```
iamalwaysuncomfortable, line 362: We may want to disable this in wasm and/or making trimming a manual call as we'll want to keep the SRS in place up to the max degree of the function in storage.

algorithms/src/polycommit/sonic_pc/data_structures.rs
```diff
@@ -275,44 +296,171 @@ impl<E: PairingEngine> ToBytes for CommitterKey<E> {
...
+                        Some(found_powers) => {
+                            new_shifted_powers_of_beta_times_gamma_g.insert(*degree_bound, found_powers);
+                        }
+                        None => {
+                            let shift_degree = max_degree - *degree_bound;
+                            let mut powers_for_degree_bound =
+                                Vec::with_capacity((max_degree + 2).saturating_sub(shift_degree));
+                            for i in 0..=supported_hiding_bound
+                                .checked_add(1)
+                                .ok_or(anyhow!("Overflow supported_hiding_bound"))?
```
iamalwaysuncomfortable, line 383: From Pratuysh's comment from long ago, he seemed concerned that if the hiding bound is close to the number of powers downloaded, this could lead removing the gauruntees hiding provides. How much of a concern do you think this still is?

algorithms/src/snark/varuna/ahp/indexer/circuit_info.rs
```diff
@@ -44,12 +48,51 @@ pub struct CircuitInfo {
...
     }
+
+    /// The maximum size of polynomials we need to do (i)fft for
+    pub fn max_fft_size<F: PrimeField>(&self) -> usize {
+        let size = [
+            2 * self.num_constraints,                  // zerocheck poly degree
+            2 * self.num_public_and_private_variables, // lineval sumcheck poly degree
+            2 * self.num_non_zero_a,                   // matrix sumcheck poly degree
+            2 * self.num_non_zero_b,                   // matrix sumcheck poly degree
+            2 * self.num_non_zero_c,                   // matrix sumcheck poly degree
```
iamalwaysuncomfortable, line 75: @vicsn this I think was possibly done before the changes in Arithmetization? Are these still valid?

algorithms/src/polycommit/sonic_pc/data_structures.rs
```diff
@@ -275,44 +296,171 @@ impl<E: PairingEngine> ToBytes for CommitterKey<E> {
...
 
-    /// The powers used to commit to shifted polynomials.
-    /// This is `None` if `self` does not support enforcing any degree bounds.
-    pub shifted_powers_of_beta_g: Option<&'a Vec<E::G1Affine>>,
+                let shifted_ck_time = start_timer!(|| "Constructing `shifted_powers_of_beta_g`");
+                match new_highest_degree_bound.cmp(&highest_degree_bound) {
+                    Ordering::Greater => {
+                        let lowest_shift_degree = max_degree - new_highest_degree_bound;
+                        let highest_shift_degree = max_degree + 1 - highest_degree_bound;
+                        let mut new_powers = srs.powers_of_beta_g(lowest_shift_degree, highest_shift_degree)?;
```
iamalwaysuncomfortable, line 355: Although this PR helps a ton by reducing key sizes, this becomes the new pain in the butt for those without filesystems because at the very best, it invokes synchronous XmlHttpRequest and that works (browsers are ending support for it) but it may block the app during the i/o. At worst, this fails and there's no access to the parameters.

There are three options to fix this:
1. Cache the pieces of the SRS fetched into memory somewhere and modify this logic slightly to fetch from cache.
2. Provide an `extern` function is implementable by outside callers which allows custom fetching logic, but validates param integrity with a checksum.
3. Find some way to do this asynchronously at an earlier point. The SDK is a custom executor, so it could potentially call parameter updates from degree bounds stated within pk/vks so that the SRS is properly trimmed by this point.

We'll have to do some current testing to ensure that XmlHttpRequest is still working in most browsers, otherwise we lose the ability to do local Proving if we DO have proving keys.

algorithms/src/snark/varuna/ahp/prover/round_functions/third.rs
```diff
@@ -154,12 +154,15 @@ impl<F: PrimeField, SM: SNARKMode> AHPForR1CS<F, SM> {
...
+        let ifft_precomputation = &state.ifft_precomputation;
+
         // Compute lineval sumcheck witnesses
         let mut job_pool = ExecutionPool::with_capacity(total_instances * 3);
         // Iterate for each circuit in the batch.
-        for ((((circuit, circuit_specific_state), batch_combiner), assignments_i), matrix_transposes_i) in state
+        for (((circuit_specific_state, batch_combiner), assignments_i), matrix_transposes_i) in state
             .circuit_specific_states
-            .iter_mut()
+            .values_mut()
```
iamalwaysuncomfortable, line 165: @ljedrz is this safe?

algorithms/src/snark/varuna/data_structures/circuit_proving_key.rs
```diff
@@ -49,7 +55,12 @@ impl<E: PairingEngine, SM: SNARKMode> FromBytes for CircuitProvingKey<E, SM> {
     fn read_le<R: Read>(mut reader: R) -> io::Result<Self> {
         let circuit_verifying_key = CanonicalDeserialize::deserialize_compressed(&mut reader)?;
         let circuit = CanonicalDeserialize::deserialize_compressed(&mut reader)?;
-        let committer_key = Arc::new(FromBytes::read_le(&mut reader)?);
+        let committer_key = match bool::read_le(&mut reader) {
```
iamalwaysuncomfortable: This serialization isn't strictly backwards compatible, this will break for old proving keys because there's no flag supporting it in the original version. A serialization helper needs to be developed to handle legacy proving keys and thus this needs a bit more work to be fully backwards compatible.

@vicsn @ljedrz what's the bandwidth in protocol for this? Should start working on addressing our own comments?

algorithms/src/snark/varuna/varuna.rs
```diff
@@ -67,56 +62,50 @@ impl<E: PairingEngine, FS: AlgebraicSponge<E::Fq, 2>, SM: SNARKMode> VarunaSNARK
...
+            degree_info = match degree_info {
+                Some(degree_info) => Some(degree_info.union(&degree_info_i)),
+                None => Some(degree_info_i),
+            };
+            indexed_circuits.push(indexed_circuit);
+        }
+        let degree_info = degree_info.unwrap();
 
-            let ck = CommitterUnionKey::union(std::iter::once(&committer_key));
+        universal_prover.update(srs, degree_info)?;
```
iamalwaysuncomfortable, line 90: This is a piece in `wasm` that we may have to call externally, which will necessarily mean we need the vks easily to crossreference all this info.

Could degree info potentially be gleaned from an authorization similarly to how the execution cost was? If so that'd give us a natural stopping point in the case where a wallet doesn't have all proving keys loaded (highly possible).

iamalwaysuncomfortable (commented):
@vicsn made several comments. 

Probably the most important is that for `wasm` we'll also need to build some tooling to adjust the parameters ahead of time. This requires some thought how to do correctly.

I'd like to proceed with this, would you suggest that Konstantin and I take on the role of carrying this to the finish line or do you recommend someone in protocol do it. Either way, I think we can get this in.

algorithms/src/snark/varuna/ahp/indexer/circuit_info.rs
vicsn: Yeah I should double check this, thx for the ping

algorithms/src/snark/varuna/data_structures/circuit_proving_key.rs
vicsn: Thx for the ping. Lukasz can make it backwards compatible.

algorithms/src/snark/varuna/varuna.rs
vicsn: The degree info / CircuitInfo is in the synthesized ProvingKey/VerifyingKey. I think there's no way to quickly get it otherwise.

vicsn:
> I'd like to proceed with this, would you suggest that Konstantin and I take on the role of carrying this to the finish line

@iamalwaysuncomfortable @kpandl sounds good, we'll await your instructions for next steps, as you have context on the SDK side of things.

algorithms/src/snark/varuna/ahp/prover/round_functions/third.rs
ljedrz: what aspect of safety do you have in mind? mutating values of a `BTreeMap` is perfectly fine, only the keys may not be modified

algorithms/src/snark/varuna/data_structures/circuit_proving_key.rs
ljedrz: I actually already had it, just needed to push; it's in https://github.com/ProvableHQ/snarkVM/pull/3127/commits/8083480322b888fd91523a655c7608859016fda8.

algorithms/src/snark/varuna/varuna.rs
iamalwaysuncomfortable: Okay that implies clients will need to collect the pk/vks, that's acceptable.

iamalwaysuncomfortable (commented):
> @iamalwaysuncomfortable @kpandl sounds good, we'll await your instructions for next steps, as you have context on the SDK side of things.

Next week I'll ensure we address two separate concerns that preclude usage of this PR by the wallet.

### Issue 1: Power Downloads

1. **No caching or external storage available without file systems:** No external storage exists in `wasm` and certain other systems like secure enclaves. This means both that the SRS can only be stored in memory currently and the only way to do this is with an internet connection (which is not always available). This means for EVERY Aleo web app, it must download hundreds of MB potentially for each usage of the web app.
2. **Deprecrated Synchronous XmlHttpDownloads Required:** Currently all parameter downloads are through this deprecated function (which is at potential risk of being unsupported entirely in future browsers) 2X the size of any downloads to be allocated due to the need to first change to a text-based encoding of the bytes first.

### Issue 2: CommitterKey Population
Because this PR shifts the concern of SRS updates to the `update` function instead of importing SRS pieces from the Proving key. Any offline transactions on the web or environments where Sync XmlHttpRequests fail to work will lose the ability to prove.

### Potential solutions: 
I see the following solutions solve the above two problems, we'll make these PRs hopefully over the next two weeks.

1. **Power Download Estimation (+ Optional Download):** Functions that take in a group of proving key and verifying keys that specify the powers which must be inserted into the SRS and whether they currently exist in the SRS. These should optionally download the parameters asynchronously for clients who desire it.
2. **External parameter insertion:** Functions that allow parameters to be inserted externally so that after estimation, the executor can download the parameters (or fetch them from local storage) and insert them. This will ensure the wallet does not have continually download powers. These should be verified by checksums when inserted.
3. **Constraint Estimation for Functions in Development:** In the case where a function that has NOT yet been synthesized (i.e. developers making new functions)  and no vk is available, a function that builds the circuit once and gets the circuit info as a first step should be developed.

Once we have these in place, we can carry this PR the rest of the way.

iamalwaysuncomfortable:
A secondary note on what the result of the above proposal + this PR for the wallet.

Once this is in, wallets will be able to do the following:
* Store 100s of proving keys instead of only 10-20.
* Store the required powers locally and load them locally.
* Avoid OOM issues that appear as a result of the sync XmlHttpRequest downloads.

Currently most wallets synthesize keys every time and the synthesis + download i/o is a significant portion of the time. This PR will significantly boost the local proving story.

ljedrz:
Closing, as the codebase has diverged significantly.

PR #3128, opened 2026-01-28
Static checks for invalid external records and futures pre V13

mohammadfawaz:
## Summary

This PR introduces a new static program-level check, `violates_pre_v13_external_record_and_future_rules`, which enforces restrictions on the usage of external records and external function outputs containing structs for consensus versions **≤ V12**.  

The check is intended to gate programs that rely on external record and future types disallowed under legacy rules, while allowing the same programs to pass under consensus versions **≥ V13**, where these restrictions are lifted.

The PR also fixes another bug due to an incorrect stack being used when moving through a list of accesses such as `r0[0u32].x;`

---

## Cases Considered

Violation considered are the following:
1. An external record as a function input that uses a nonlocal struct.
2. An external record returned from an external function call that uses a nonlocal struct.
3. An external future returned from an external function call that uses a nonlocal struct.
---

## Assumptions

There are **no other ways for external records or futures to appear** in a program beyond the instruction outputs and input types considered.


synthesizer/process/src/authorize.rs
```diff
@@ -17,6 +17,8 @@ use super::*;
 
 impl<N: Network> Process<N> {
     /// Authorizes a call to the program function for the given inputs.
+    ///
+    // This method is meant to be completely stateless. It has no notion of block height or consensus version.
```
d0cd, line 21:
```suggestion
```

synthesizer/process/src/lib.rs
```diff
@@ -84,6 +84,8 @@ use locktick::parking_lot::RwLock;
 use parking_lot::RwLock;
 use std::{collections::HashMap, sync::Arc};
 
+// Note: a `Process` and all of its fields are meant to be completely stateless. They have no
+// notion of block height or consensus version.
```
d0cd, line 88:
```suggestion
```

synthesizer/program/src/lib.rs
```diff
@@ -974,6 +984,105 @@ impl<N: Network> ProgramCore<N> {
             || self.constructor.iter().any(|constructor| constructor.contains_external_struct())
     }
 
+    /// Returns `true` if this program violates pre-V13 rules for external records
+    /// or futures by referencing non-local struct types across program boundaries.
+    ///
+    /// Notes:
+    /// 1. We only need to check functions because closures and constructors cannot refernece
+    ///    external records or future in any way
```
d0cd:
```suggestion
    ///    external records or futures.
```

synthesizer/src/vm/verify.rs
d0cd, line 197: Re-organize this comment section to follow the re-organization of the conditionals.
For the less than section, make note that these do not have long term implications on verification time because they are skipped by the time we get to the latest consensus version.
And then for the GTE section, make note the opposite.

synthesizer/process/src/stack/finalize_types/mod.rs
```diff
@@ -189,7 +198,11 @@ impl<N: Network> FinalizeTypes<N> {
                         true => None,
                         // Attention - This method must fail here and early return if the external program is missing.
                         // Otherwise, this method will proceed to look for the requested function in its own program.
-                        false => Some(stack.get_external_stack(locator.program_id())?),
+                        false => {
+                            let external_stack = stack.get_external_stack(locator.program_id())?;
```
d0cd: Can this be inlined in L204?

raychu86:
ConsensusVersion::V13 has already passed. Can you explain why different rules here will be safe for nodes? Is there no forking risk and no issues with nodes syncing blocks that were produced on the previous rules?

synthesizer/program/src/lib.rs
```diff
@@ -974,6 +984,105 @@ impl<N: Network> ProgramCore<N> {
...
+        F3: Fn(&Identifier<N>) -> bool,
+    {
+        // Helper: does a record contain any struct not defined locally?
+        let record_uses_external_struct = |record: &RecordType<N>| {
+            record.entries().iter().any(|(_, member)| {
+                matches!(
+                    member,
+                    EntryType::Constant(PlaintextType::Struct(name))
+                        | EntryType::Private(PlaintextType::Struct(name))
+                        | EntryType::Public(PlaintextType::Struct(name))
```
d0cd: Should this also check arrays?

vicsn:
> ConsensusVersion::V13 has already passed. Can you explain why different rules here will be safe for nodes? Is there no forking risk and no issues with nodes syncing blocks that were produced on the previous rules?

Just for canary, where IMO we can accept the forking risk.

vicsn (approved):
Approved if other comments are resolved

synthesizer/program/src/lib.rs
mohammadfawaz: Ah yes of course. Handled for both external records and external futures.

synthesizer/process/src/stack/finalize_types/mod.rs
```diff
@@ -204,7 +204,19 @@ impl<N: Network> FinalizeTypes<N> {
                     // Check that the index is in bounds.
                     match finalize_inputs.get_index(**index as usize) {
                         // Retrieve the input type and update `finalize_type` for the next iteration.
-                        Some(input) => finalize_type = input.finalize_type().clone(),
+                        Some(input) => {
+                            finalize_type = match input.finalize_type() {
```
d0cd, line 208: Lets leave a comment here on why we're qualifying. 
Make a note that this was added in V13, and a consensus version check was added to safely transition the program. 

Ditto for the same fix in `RegisteryTypes`

synthesizer/src/vm/tests/test_v13.rs
```diff
@@ -521,3 +439,434 @@ constructor:
...
+function second:
+    call test_one.aleo/make_s into r0;
+    output r0 as test_one.aleo/S.public;
+",
+    )
+    .unwrap();
+
+    // Use V11 rather than V12 to make sure we still won't be on V13
+    for consensus_version in [ConsensusVersion::V11, ConsensusVersion::V13] {
+        let block = run_deploy_test(consensus_version, &program_one, &program_two);
```
d0cd: For the last iteration, we should add the final block to VM.
And it would be nice to have an execution test now that the program is successfully deployed.

synthesizer/src/vm/tests/test_v13.rs
```diff
@@ -521,3 +439,434 @@ constructor:
...
+    assert_eq!(block.transactions().num_accepted(), 1);
+    assert_eq!(block.transactions().num_rejected(), 0);
+    assert_eq!(block.aborted_transaction_ids().len(), 0);
+
+    vm.add_next_block(&block).unwrap();
+
+    // Deploy second program (behavior depends on consensus version).
+    let deployment_two = vm.deploy(&caller_private_key, program_two, None, 0, None, rng).unwrap();
+
+    sample_next_block(&vm, &caller_private_key, &[deployment_two], rng).unwrap()
```
d0cd: It's worth documenting that final block doesn't get added to the VM.
Confused me for a bit in the beginning.

synthesizer/src/vm/verify.rs
```diff
@@ -249,24 +246,78 @@ impl<N: Network, C: ConsensusStorage<N>> VM<N, C> {
...
+                    let is_local_struct = |identifier: &Identifier<N>| program.structs().contains_key(identifier);
+
+                    ensure!(
+                        !program.violates_pre_v13_external_record_and_future_rules(
+                            &get_external_record,
+                            &get_external_function,
+                            &get_external_future,
+                            &is_local_struct,
+                        ),
+                        "Invalid external record",
```
d0cd: Does this error message need to change?

synthesizer/program/src/lib.rs
```diff
@@ -974,6 +984,112 @@ impl<N: Network> ProgramCore<N> {
             || self.constructor.iter().any(|constructor| constructor.contains_external_struct())
     }
 
+    /// Returns `true` if this program violates pre-V13 rules for external records
+    /// or futures by referencing non-local struct types across program boundaries.
+    ///
+    /// Notes:
+    /// 1. We only need to check functions because closures and constructors cannot refernece
```
d0cd:
```suggestion
    /// 1. We only need to check functions because closures and constructors cannot reference
```

synthesizer/program/src/lib.rs
```diff
@@ -974,6 +984,112 @@ impl<N: Network> ProgramCore<N> {
...
+    ///    external records or futures.
+    /// 2. We only need to check function inputs because if function outputs have a violation then
+    ///    either the inputs or the body of the function must violate first.
```
d0cd:
```suggestion
    ///    either the inputs or the body of the function must have a violation.
```

synthesizer/program/src/lib.rs
```diff
@@ -974,6 +984,112 @@ impl<N: Network> ProgramCore<N> {
             || self.constructor.iter().any(|constructor| constructor.contains_external_struct())
     }
 
+    /// Returns `true` if this program violates pre-V13 rules for external records
+    /// or futures by referencing non-local struct types across program boundaries.
```
d0cd:
```suggestion
    /// or futures by containing registers with non-local struct types across program boundaries.
```

synthesizer/program/src/lib.rs
```diff
@@ -974,6 +984,112 @@ impl<N: Network> ProgramCore<N> {
...
+    ///
+    /// Notes:
+    /// 1. We only need to check functions because closures and constructors cannot refernece
+    ///    external records or futures.
+    /// 2. We only need to check function inputs because if function outputs have a violation then
```
d0cd:
```suggestion
    /// 2. We need to check function inputs. 
```

d0cd: Move the "No need to check function outputs, because it has already been checked in either the inputs or call instruction" as point #4.

synthesizer/program/src/lib.rs
```diff
@@ -974,6 +984,112 @@ impl<N: Network> ProgramCore<N> {
...
+            // Scan instructions for calls to external programs.
+            for instruction in function.instructions() {
+                let Instruction::Call(call) = instruction else { continue };
+                let CallOperator::Locator(locator) = call.operator() else { continue };
+                let Ok(external_function) = get_external_function(locator) else {
+                    continue;
+                };
+
+                // Check if the outputs of the external function reference a struct that is not
+                // locally available. That's a violation.
```
d0cd:
```suggestion
                // locally available. 
```

synthesizer/program/src/lib.rs
```diff
@@ -974,6 +984,112 @@ impl<N: Network> ProgramCore<N> {
...
+                                return true;
+                            }
+                        }
+
+                        ValueType::Future(loc) => {
+                            let Ok(future) = get_external_future(loc) else {
+                                continue;
+                            };
+                            for input in future.input_types() {
+                                let FinalizeType::Plaintext(ty) = input else {
```
d0cd, line 1075: Do we need to look at `FinalizeType::Future` as well?

mohammadfawaz: Added a comment explaining why this isn't needed.

synthesizer/process/src/stack/register_types/mod.rs
```diff
@@ -271,7 +271,17 @@ impl<N: Network> RegisterTypes<N> {
                             register_type = match input.finalize_type() {
                                 FinalizeType::Plaintext(plaintext_type) => {
                                     let plaintext = match external_stack {
-                                        Some(ref stack) => plaintext_type.clone().qualify(*stack.program_id()),
+                                        Some(ref external_stack) => {
+                                            // Qualify the finalize input type with the external program ID so that any
```
d0cd, line 275:
```suggestion
                                            // Qualify the register type with the external program ID so that any
```

synthesizer/process/src/stack/finalize_types/mod.rs
```diff
@@ -204,7 +204,29 @@ impl<N: Network> FinalizeTypes<N> {
...
+                                FinalizeType::Plaintext(plaintext_type) => {
+                                    let plaintext = match external_stack {
+                                        Some(ref external_stack) => {
+                                            // Qualify the finalize input type with the external program ID so that any
```
d0cd, line 212:
```suggestion
                                            // Qualify the finalize type with the external program ID so that any
```

synthesizer/src/vm/tests/test_v13.rs
d0cd: @mohammadfawaz flagging this improvement and we should be good.

d0cd (approved):
Looks good @mohammadfawaz! Nice work diving deep into this issue.
I can confirm that the test cases in this PR fail to deploy on v4.4.0 and continue to be aborted until `ConsensusVersion::V13`.
The regression tests were also run confirming that a `Process` instance can be built off of this branch for all 3 networks.

Left some nits and minor final comments.

synthesizer/src/vm/tests/test_v13.rs
mohammadfawaz: Done!

PR #3138, opened 2026-02-02
Fix external structs access in external mapping;

mohammadfawaz:
### Summary

This PR addresses qualification handling for external structs in mapping operations and fixes a consensus-version–specific execution issue affecting external mappings in pre-V13.

### Changes

1. **Qualify external struct operands in mapping instructions**

   When external structs are used as operands to mapping instructions, operand types are now properly **qualified** if the mapping itself is external. This ensures that:
   - Mapping key and value types correctly reflect external struct usage.
   - Any structs nested within mapping key or value types are also qualified as external.
   - The type returned from the `get` instruction is qualified when required.

2. **Restore pre-V123behavior for external mapping `get`**

   In consensus version pre-V13 a `get` from an external mapping is valid as long as the key does not contain an external struct, even if the value type does. However, such programs fail to execute on v4.4.0.

   While this issue is fixed by (1), the fix must not apply to pre-V13. Therefore:
   - Any types that were newly qualified for external mappings are **unqualified** when targeting consensus versions prior to V13.
   - This restores correct pre-V13 semantics while preserving the fix for V13.

### Testing

- Added a comprehensive test covering various mapping operations involving external structs.
- Added two targeted tests to ensure that type *unqualification* in pre-V12 is applied correctly.

### A Quick Note

Across PRs #3107, #3128, and #3138, the recurring pattern is that a primary program performs an external call or instruction (e.g. get) that returns an external type (external record, struct, future, etc.), but type resolution is attempted as if the type were local. Under V12, this causes resolution to fail because the returned type is not qualified from the perspective of the primary program.

In V13, we fix this by qualifying externally-owned types at the boundary, using the program that owns the type, so resolution is explicit and unambiguous. This allows otherwise valid programs to pass.

We cannot safely apply the same fix in V12, because changing how external return types are resolved would alter which programs are accepted, creating forking risk. Therefore, these corrections are intentionally scoped to V13 consensus via static checks or execution-time checks.

### DIFF
--- a/synthesizer/process/src/stack/finalize_registers/registers_trait.rs
+++ b/synthesizer/process/src/stack/finalize_registers/registers_trait.rs
@@ -135,7 +135,13 @@ impl<N: Network> RegistersTrait<N> for FinalizeRegisters<N> {
                 // Ensure the type of the register is valid.
                 match (self.finalize_types.get_type(stack, register), &stack_value) {
                     // Ensure the plaintext value matches the plaintext type.
-                    (Ok(FinalizeType::Plaintext(plaintext_type)), Value::Plaintext(plaintext_value)) => {
+                    (Ok(FinalizeType::Plaintext(mut plaintext_type)), Value::Plaintext(plaintext_value)) => {
+                        if N::CONSENSUS_VERSION(self.state().block_height())? < ConsensusVersion::V13 {
+                            // Pre-V13, the `ExternalStruct` type did not exist. If the type happens to be qualified in
+                            // `get_type` above, we need to unqualify it (i.e. `ExternalStruct` becomes `Struct`) in order
+                            // to obtain the same behavior we had pre-V13.
+                            plaintext_type = plaintext_type.unqualify();
+                        }
                         stack.matches_plaintext(plaintext_value, &plaintext_type)?
                     }
                     // Ensure the future value matches the future type.
--- a/synthesizer/process/src/stack/finalize_registers/registers_trait.rs
+++ b/synthesizer/process/src/stack/finalize_registers/registers_trait.rs
@@ -140,6 +140,9 @@ impl<N: Network> RegistersTrait<N> for FinalizeRegisters<N> {
                             // Pre-V13, the `ExternalStruct` type did not exist. If the type happens to be qualified in
                             // `get_type` above, we need to unqualify it (i.e. `ExternalStruct` becomes `Struct`) in order
                             // to obtain the same behavior we had pre-V13.
+                            //
+                            // Note: we don't need to rescursively check `Struct` definitions since external structs
+                            // were not allowed before V13.
                             plaintext_type = plaintext_type.unqualify();
                         }
                         stack.matches_plaintext(plaintext_value, &plaintext_type)?
--- a/synthesizer/src/vm/tests/test_v13.rs
+++ b/synthesizer/src/vm/tests/test_v13.rs
@@ -1039,24 +1039,23 @@ constructor:
     assert_eq!(block.aborted_transaction_ids().len(), 0);
 }
 
+/// This test verifies that using the `get` command to read from an external mapping fails on V12
+/// when the mapping value contains an external struct (i.e. a struct that is not local to the program
+/// containing the `get`). The same scenario should succeed on V13, where the underlying bug has
+/// been fixed.
 #[test]
 fn test_external_mapping_external_struct_runtime_pre_post_v13() {
-    // Run the same scenario under a pre-V13 and post-V13 consensus version.
-    // The behavior difference is expected only at execution time.
-    //
-    // Start at V9 to make sure we're still pre-V13 by the time we execute the last transaction
-    for consensus_version in [ConsensusVersion::V9, ConsensusVersion::V13] {
-        let rng = &mut TestRng::default();
+    let rng = &mut TestRng::default();
+    let private_key = crate::vm::test_helpers::sample_genesis_private_key(rng);
+    // Start the VM at consensus version V9 to ensure we're not at V13 by the time we get to the
+    // execution transaction that calls the second program.
+    let height = CurrentNetwork::CONSENSUS_HEIGHT(ConsensusVersion::V9).unwrap();
+    let vm = crate::vm::test_helpers::sample_vm_at_height(height, rng);
 
-        let caller_private_key = crate::vm::test_helpers::sample_genesis_private_key(rng);
-
-        let height = CurrentNetwork::CONSENSUS_HEIGHT(consensus_version).unwrap();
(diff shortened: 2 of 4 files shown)


synthesizer/src/vm/tests/test_v13.rs
```diff
@@ -473,7 +473,7 @@ struct ExecutionTest<'a> {
 /// # Notes
 /// This helper abstracts VM initialization, block creation, deployment, and optional execution logic
 /// to reduce boilerplate in multiple tests and ensure consistent pre-/post-V13 behavior.
-fn deploy_two_programs_and_execute_v132(
+fn deploy_two_programs_and_execute_v13(
```
d0cd, line 476: This helper's gotten pretty complex. 
Probably worth elaborating in the comment the pathways that it tests and its expected behavior.

synthesizer/process/src/stack/finalize_registers/registers_trait.rs
```diff
@@ -135,7 +135,13 @@ impl<N: Network> RegistersTrait<N> for FinalizeRegisters<N> {
                 // Ensure the type of the register is valid.
                 match (self.finalize_types.get_type(stack, register), &stack_value) {
                     // Ensure the plaintext value matches the plaintext type.
-                    (Ok(FinalizeType::Plaintext(plaintext_type)), Value::Plaintext(plaintext_value)) => {
+                    (Ok(FinalizeType::Plaintext(mut plaintext_type)), Value::Plaintext(plaintext_value)) => {
+                        if N::CONSENSUS_VERSION(self.state().block_height())? < ConsensusVersion::V13 {
+                            // Pre-V13, the `ExternalStruct` type did not exist. If the type happens to be qualified in
```
d0cd, line 140: Note for reviewers. We don't need to recursively check `Struct` definitions since external structs were not allowed before V13. Had to convince myself of that.

synthesizer/src/vm/tests/test_v13.rs
```diff
@@ -923,10 +923,367 @@ constructor:
...
+                None,
+                0,
+                None,
+                rng,
+            )
+            .unwrap();
+
+        let block = sample_next_block(&vm, &caller_private_key, &[execution], rng).unwrap();
+
+        match consensus_version {
```
d0cd: We should modify this so that we explicitly check that an execution is rejected before consensus version V13, advance past V13, and check that it is accepted.

d0cd: Ditto for the test below.

d0cd (approved):
Requested some improvements to the tests, but otherwise looks good.

Small request, would you be able to add a few sentences on the overall pattern we've discovered between this PR and the prior set of fixes?

mohammadfawaz:
> Requested some improvements to the tests, but otherwise looks good.
> 
> Small request, would you be able to add a few sentences on the overall pattern we've discovered between this PR and the prior set of fixes?

Thanks! Where is the best place to add that? In the PR description you mean? or somewhere maybe in `test_v13.rs`?

synthesizer/src/vm/tests/test_v13.rs
mohammadfawaz: There's already a pretty extensive doc comment for it. So does the helper `assert_pre_post_v13`. It's currently only used for the pattern where the second deployment fails with pre-V13. The new tests I added do not use it.

Is there anything in those comments that need more clarifications?

d0cd: Ah didn't see that, thanks!

d0cd:
In the PR description is fine!

synthesizer/process/src/stack/finalize_registers/registers_trait.rs
```diff
@@ -135,7 +135,16 @@ impl<N: Network> RegistersTrait<N> for FinalizeRegisters<N> {
                 // Ensure the type of the register is valid.
                 match (self.finalize_types.get_type(stack, register), &stack_value) {
                     // Ensure the plaintext value matches the plaintext type.
-                    (Ok(FinalizeType::Plaintext(plaintext_type)), Value::Plaintext(plaintext_value)) => {
+                    (Ok(FinalizeType::Plaintext(mut plaintext_type)), Value::Plaintext(plaintext_value)) => {
+                        if N::CONSENSUS_VERSION(self.state().block_height())? < ConsensusVersion::V13 {
```
raychu86, line 139: In an attempt to reduce the number of `ConsensusVersion` checks everywhere, is there a way this can be done or checked externally?

mohammadfawaz: There are probably some static checks we could add, but it’s not obvious to me where they would live. Unlike the checks introduced in https://github.com/ProvableHQ/snarkVM/pull/3128, this issue cannot be caught purely at deploy-time (in `vm/verify.rs`). The offending program should deploy successfully under V12 and only fails during execution, because the problem arises when resolving externally-owned types at runtime.

mohammadfawaz: I'm going to try to add a static check in `check_execution_internal` in `vm/verify.rs` and see if that works. I think it should.

synthesizer/src/vm/tests/test_v13.rs
```diff
@@ -923,10 +923,407 @@ constructor:
...
+                0,
+                None,
+                rng,
+            )
+            .unwrap();
+
+        let block = sample_next_block(&vm, &caller_private_key, &[execution], rng).unwrap();
+        vm.add_next_block(&block).unwrap();
+
+        match consensus_version {
```
d0cd: The match here is a bit confusing to understand.
Could we instead assert that the transaction is rejected, add the block and advance until after V13, execute, and assert acceptance?
Ditto for the test below.

mohammadfawaz: Makes sense! This means we can get rid of the loop as well. Addressed here: https://github.com/ProvableHQ/snarkVM/pull/3138/commits/24e0b1f08f964a0eaf309be4d7ce9639c760899c

synthesizer/process/src/stack/finalize_registers/registers_trait.rs
mohammadfawaz: On second thought, and after discussing this with Pranav, a static check would cause the execution transaction to be aborted, while the current behavior on V12 would cause the execution transaction to be rejected. Since rejected transactions are still stored on-chain with fee IDs, there is a fork risk involved. So really the safest approach is "unfix" the bug on V12 in `fn store(..)` so that we continue to reject the transaction, which is what the PR currently does.

vicsn (approved):

PR #3143, opened 2026-02-10
[Perf] Revamp historical mapping storage

ljedrz:
The history of mappings is currently stored (on demand, via `--features=history`) in JSON format, which is inefficient. This PR removes that implementation in favor of a storage-based one.

The summary of main changes:
- `Plaintext` receives a `Hash` impl
- `MemoryMap` and `DataMap` no longer require the keys to be `Copy` (which is negligible; the alternative would be for the `Plaintext` to be `Copy`, which would be problematic)
- `Finalize{DB/Memory}` are optionally extended with 2 new storage maps and the current block height
- 2 new `DataID` variants are introduced, and they are deliberately **not** feature-gated
- `FinalizeStorage` is extended with 2 new associated types and a few methods
- whenever `FinalizeStorage::{insert_key_value, update_key_value}` is called with `--history` enabled, the 2 new storage maps are populated

This was tested by enhancing the existing `test_bond_and_unbond_validator` case from `snarkvm-ledger`.


ledger/src/tests.rs
```diff
@@ -734,19 +785,51 @@ fn test_bond_and_unbond_validator() {
...
+
+        let initial_mapping_value = ledger
+            .vm()
+            .finalize_store()
+            .get_historical_mapping_value(program_id, metadata_mapping_name, metadata_mapping_key.clone(), 3)
+            .unwrap()
+            .unwrap();
+        assert_eq!(&*initial_mapping_value, &Value::<CurrentNetwork>::try_from("4u32").unwrap());
+
+        let initial_mapping_heights = ledger
```
vicsn: Nit: I think many instances need a different variable name than "initial_*"

vicsn (changes requested):
Say we updated at height 2 and 4. Do I see correctly querying at height 3 will fail, and we leave it up to the user of the library to find the value at height 2? Is it worth it moving such logic into snarkVM?

ljedrz:
> Say we updated at height 2 and 4. Do I see correctly querying at height 3 will fail, and we leave it up to the user of the library to find the value at height 2?

Only if the user-facing query uses only `get_historical_mapping_value`; I assumed this would be handled on snarkOS-side, as it's a UX concern.

>Is it worth it moving such logic into snarkVM?

Your call, can be either.

ledger/store/src/program/finalize.rs
```diff
@@ -683,7 +683,22 @@ impl<N: Network, P: FinalizeStorage<N>> FinalizeStore<N, P> {
...
+        // First, obtain the heights at which updates have happened.
+        let Some(update_heights) = self.get_mapping_update_heights(program_id, mapping_name, mapping_key.clone())?
+        else {
+            return Ok(None);
+        };
+
+        // Find the updated height which matches the desired point in history.
+        let Some(applicable_height) = (match update_heights.binary_search(&height) {
+            Ok(_) => Some(height),
+            Err(idx) => update_heights.get(idx - 1).copied(),
```
vicsn, line 696: `idx` might be 0 right? So `idx - 1` might not be right?

In `macro_rules! consensus_config_value` we use:
```
                    if index == 0 {
                        None
                    } else {
                        Some($network::$constant[index - 1].1)
                    }
```

ljedrz: good call; fixed in https://github.com/ProvableHQ/snarkVM/pull/3143/commits/261e2f2d7481bbeacbc19461d5ae5e590bb260e0

synthesizer/src/vm/finalize.rs
```diff
@@ -1391,36 +1405,6 @@ impl<N: Network, C: ConsensusStorage<N>> VM<N, C> {
...
 
-                    #[cfg(all(feature = "history", feature = "rocks"))]
-                    {
-                        // When finalizing in `FinalizeMode::RealRun`, store the delegated and bonded mappings in history.
-                        if IS_FINALIZE {
-                            // Load a `History` object.
-                            let history = History::new(N::ID, store.storage_mode());
-
-                            // Write the delegated mapping as JSON.
-                            history.store_mapping(state.block_height(), MappingName::Delegated, &next_delegated_map)?;
```
miazn, line 1402: I'm assuming that these mappings get updated in the core logic that updates the key value store now?

ljedrz: yes, both the mappings and their historical values get updated as part of the existing `FinalizeStorage` plumbing

console/program/src/data/plaintext/mod.rs
```diff
@@ -121,6 +124,34 @@ impl<N: Network> From<&Literal<N>> for Plaintext<N> {
     }
 }
 
+impl<N: Network> Hash for Plaintext<N> {
+    fn hash<H: Hasher>(&self, state: &mut H) {
+        match self {
+            Self::Literal(literal, bits) => {
```
vicsn: Are these bits guaranteed to exist?

As these hashes influence the rocksdb key, we need to worry about collissions.

A smart contract's mapping's key type can never be upgraded or changed and has a fixed size. But we should be sure these bits can't be targeted to make it variable size.

If we're not sure we could always hash a bit of extra TLV encoding

ljedrz: we can just not hash those bits, actually; as far as I can tell, they are a performance optimization only

ljedrz: done in https://github.com/ProvableHQ/snarkVM/pull/3143/commits/1a8bb6e8217a80f449c04e53ddbaa1f6afdfcf19

ledger/store/src/program/finalize.rs
```diff
@@ -228,6 +308,25 @@ pub trait FinalizeStorage<N: Network>: 'static + Clone + Send + Sync {
...
+                let current_height = self.current_block_height().load(Ordering::SeqCst);
+
+                // Register the updated value at the current height.
+                self.mapping_update_map()
+                    .insert((program_id, mapping_name, key.clone(), current_height), value.clone())?;
+
+                // Obtain the list of past update heights.
+                let key = (program_id, mapping_name, key.clone());
+                let update_heights =
+                    self.mapping_update_heights_map().get_confirmed(&key)?.map(|list| list.into_owned());
```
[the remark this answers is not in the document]
ljedrz: not applicable, we're looking for **past** updates

vicsn (approved):

vicsn:
Re-opening to run CI

PR #3144, opened 2026-02-10
Fix the implementation of `unqualify`

mohammadfawaz:
The implementation of `unqualify` was wrong because it's using `.name()` instead of `.resource()` to access the name of the struct.

- Added an e2e test that ensures it's working correctly now.
- Added some unit tests to directly test `unqualify`.

It's unfortunate this wasn't caught in testing before. I was focused on checking that the pattern we're looking for in https://github.com/ProvableHQ/snarkVM/pull/3138 is failing on V12 but didn't check that some related patterns still pass. That's what the new test in `test_v13` now checks.

### DIFF
--- a/console/program/src/data_types/plaintext_type/mod.rs
+++ b/console/program/src/data_types/plaintext_type/mod.rs
@@ -69,7 +69,7 @@ impl<N: Network> PlaintextType<N> {
             PlaintextType::Literal(..) | PlaintextType::Struct(..) => self,
 
             // Drop the program qualification unconditionally
-            PlaintextType::ExternalStruct(locator) => PlaintextType::Struct(*locator.name()),
+            PlaintextType::ExternalStruct(locator) => PlaintextType::Struct(*locator.resource()),
 
             // Recurse into arrays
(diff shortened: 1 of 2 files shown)


raychu86:
I believe Canary has this issue too, so we'll want to make sure Canary gets the patch, then push up to Testnet. @vicsn

d0cd (approved):
LGTM! Are there another similar patterns we need coverage for?

mohammadfawaz:
> LGTM! Are there another similar patterns we need coverage for?

Can't think of anything at the moment that isn't a variation of what we already have. We basically now have positive a negative tests for the pre-V13 logic we added in https://github.com/ProvableHQ/snarkVM/pull/3138. If you can think of any exotic cases, please let me know!

vicsn (approved):
Confirmed that before this fix, an E2E sync test failed within an hour. After this fix, we can mostly sync genesis to tip testnet without issue (to be more precise, I tested blocks 0 to 9 million, and blocks 12 million to tip. We have a few snapshots to add to our testing repertoire.

Will merge into Canary too and roll it out to Canary first.

PR #3149, opened 2026-02-12
Report mismatching finalized transactions on failure to add block

vicsn:
## Motivation

Currently when using CDN sync, when adding a block fails due to mismatching finalization, we get a very opaque error message. To help debugging during testing, this PR proposes to dump the finalized ids and transactions - so they can be compared to the finalized block on the functional production network.

The debug message could be huge, but this is also expected to happen very infrequently - only on faulty nodes.

### DIFF
--- a/synthesizer/src/vm/finalize.rs
+++ b/synthesizer/src/vm/finalize.rs
@@ -154,8 +154,11 @@ impl<N: Network, C: ConsensusStorage<N>> VM<N, C> {
             bail!("The ratifications after speculation do not match the ratifications in the block");
         }
         // Ensure the transactions after speculation match.
-        if transactions != &confirmed_transactions.into_iter().collect() {
-            bail!("The transactions after speculation do not match the transactions in the block");
+        let confirmed_transactions = confirmed_transactions.into_iter().collect();
+        if transactions != &confirmed_transactions {
+            bail!(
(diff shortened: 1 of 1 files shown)


synthesizer/src/vm/finalize.rs
```diff
@@ -154,8 +154,12 @@ impl<N: Network, C: ConsensusStorage<N>> VM<N, C> {
             bail!("The ratifications after speculation do not match the ratifications in the block");
         }
         // Ensure the transactions after speculation match.
-        if transactions != &confirmed_transactions.into_iter().collect() {
-            bail!("The transactions after speculation do not match the transactions in the block");
+        let confirmed_transactions = confirmed_transactions.into_iter().collect();
+        if transactions != &confirmed_transactions {
+            let confirmed_transaction_ids = confirmed_transactions.transaction_ids().collect::<Vec<_>>();
+            bail!(
```
raychu86, line 160: You could reduce the size of the logs by only printing the transactions that were different with something like:

```
    let diffs: Vec<_> = transactions
        .iter()
        .zip(&confirmed_transactions)
        .enumerate()
        .filter_map(|(i, (expected, actual))| {
            if expected != actual {
                Some((i, expected, actual))
            } else {
                None
            }
        })
        .collect();
```

Thought it may not account for one-offs that well.

raychu86 (approved):

PR #3152, opened 2026-02-13
[Feat] Store the number of consensus heights and use it in CI

### LINKED ISSUE #3150: [Feature] Add a CI job to run VM integration tests on the penultimate consensus version
## 🚀 Feature

We could potentially have caught https://github.com/ProvableHQ/snarkVM/pull/3144 if we had run our VM integration tests on the penultimate consensus version.

Specifically, `test_vm_execute_and_finalize` from this job runs a bunch of deployments and executions through the VM:
```
  synthesizer-integration:
    executor: rust-docker
    steps:
      - run_test:
          timeout: 30m # test_vm_execute_and_finalize can take over 10 minutes in the current setup
          no_output_timeout: 20m
          flags: --test '*' --features test -- --test-threads=4
          cache_key_suffix: -integration
```

We should make a copy of the CI job with alternative configuration where the latest consensus version is changed to a `u32::MAX`. I see two approaches, there may be more:
- We could set `CONSENSUS_VERSION_HEIGHTS` before running the test, but to set those properly we'd have to:
  - manually parse e.g. the number of versions in `enum ConsensusVersion`. Feels brittle and ugly.
  - we could introduce a new file which contains just `NUM_CONSENSUS_VERSIONS`, and we verify in `build.rs` that it's correct. Less brittle but more complexity.
- We don't could introduce a `test_consensus_heights_penultimate` feature, which sets the heights like in `load_test_consensus_heights_inner` but with `u32::MAX`. I guess it's not compatible with the `test` feature.

I like the last idea the most but open to other ideas.

ljedrz:
Extends the build script to store the number of consensus versions in a dedicated file in order to be able to use it in CI jobs and for other dev purposes - in this case, in a modified `synthesizer-integration` job.

See the [2nd commit](https://github.com/ProvableHQ/snarkVM/commit/d1d7d483bc85ff924e970dab4fa607375036a923) for the CI job code.

Filing as a draft in order to fine-tune the details before testing the job, as it's a heavy one.

Closes https://github.com/ProvableHQ/snarkVM/issues/3150.


build.rs
```diff
@@ -178,10 +182,70 @@ fn check_file_licenses<P: AsRef<Path>>(path: P) {
...
+    });
+
+    // Traverse the list of variants and count them.
+    let mut num_consensus_versions = 0;
+    for line in line_iter {
+        let line = line.expect("Can't process the consensus heights file");
+        if line.contains("];") {
+            break;
+        } else {
+            num_consensus_versions += 1;
```
vicsn: This will be incorrect if we ever introduce lines of comments in that part of the code. Could we instead parse what is the largest `V*` in `enum ConsensusVersion`?

.circleci/config.yml
```diff
@@ -935,6 +935,28 @@ jobs:
...
+          # Generate the value (u32::MAX times the number of versions)
+          VAR_VALUE=$(yes 4294967295 | head -n "$COUNT" | paste -sd "," -)
+
+          # Export to BASH_ENV so 'run_test' can see it
+          echo "export CONSENSUS_VERSION_HEIGHTS=\"$VAR_VALUE\"" >> "$BASH_ENV"
+      - run_test:
+          timeout: 30m # test_vm_execute_and_finalize can take over 10 minutes in the current setup
+          no_output_timeout: 20m
+          workspace_member: snarkvm-synthesizer
+          flags: --test '*' --features test -- --test-threads=4
```
vicsn: Just checking: Does the main test we care about automatically fast forward to the latest consensus version? We should avoid that as it will otherwise run forever. :)

ljedrz: The inner workings of `test_vm_execute_and_finalize` are an enigma to me, but there are a few things worth noting:
1. when I run the test locally and adjust the only spot where the consensus version is used to display it, I immediately see version 14
2. there is some specific logic checking whether the version is greater or equal to `V4`
3. when I applied the proposed `CONSENSUS_VERSION_HEIGHTS`, I first got a panic, as some code requires the 1st height to be `0` (due to it being genesis, makes sense)
4. when I accounted for point `3`, the test does, in fact, hang

vicsn: Here is the problem:
```
        // If the `start_height` field is present in the config, parse it as a `u32`.
        // Otherwise use the latest consensus height as the default.
        let start_height = test_config
            .get("start_height")
            .map(|value| value.as_u64().expect("`start_height` must be a u32"))
            .unwrap_or(
                CurrentNetwork::CONSENSUS_HEIGHT(ConsensusVersion::latest()).expect("Expected consensus height") as u64
            ) as u32;
```

Can we give the test a flag which determines whether to use `ConsensusVersion::latest()` or `ConsensusVersion::latest() - 1` as a default?

And can we set `CONSENSUS_VERSION_HEIGHTS=0,1,..,ConsesusVersion::latest() - 1, u32::max()`

vicsn:
Can you also add this into `config.yml` - didn't want to run CI yet another time for such a small change:
```
diff --git a/.circleci/config.yml b/.circleci/config.yml
index 62f9373a5..07447c0d7 100644
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -974,7 +974,7 @@ jobs:
     steps:
       - run_test:
           workspace_member: snarkvm-synthesizer-program
-          flags: keccak
+          flags: keccak -- --test-threads=4
           timeout: 30m
           no_output_timeout: 20m
           cache_key_suffix: -integration-keccak
@@ -994,7 +994,7 @@ jobs:
     steps:
       - run_test:
           workspace_member: snarkvm-synthesizer-program
-          flags: sha
+          flags: sha -- --test-threads=4
           timeout: 30m
           no_output_timeout: 20m
           cache_key_suffix: -integration-sha
```

build.rs
ljedrz: can do :+1:

.circleci/config.yml
ljedrz: > Can we give the test a flag which determines whether to use ConsensusVersion::latest() or ConsensusVersion::latest() - 1 as a default?

Perhaps an additional env variable like `CONSENSUS_VERSION_DEFAULT_HEIGHT`? A bit of a mouthful, but it's a dev-facing feature.

> And can we set `CONSENSUS_VERSION_HEIGHTS=0,1,..,ConsesusVersion::latest() - 1, u32::max()`

can do :+1:

ljedrz: > can do

Well, almost - it is expected that there are at most `NUM_CONSENSUS_VERSIONS` in use, so it can only be
```
CONSENSUS_VERSION_HEIGHTS=0,1,..,ConsesusVersion::latest() - 2, u32::max()
```

vicsn: Sounds good. I hope it'l work now, feel free to enable CI if it works for you locally.

ljedrz:
Hmm, let me reopen in an attempt to start the CI.

PR #3153, opened 2026-02-16
[Feat] Store the number of consensus heights and use it in CI

### LINKED ISSUE #3150: [Feature] Add a CI job to run VM integration tests on the penultimate consensus version

ljedrz:
A reopening of https://github.com/ProvableHQ/snarkVM/pull/3152.

Extends the build script to store the number of consensus versions in a dedicated file in order to be able to use it in CI jobs and for other dev purposes - in this case, in a modified synthesizer-integration job.

Closes https://github.com/ProvableHQ/snarkVM/issues/3150.


vicsn:
I just went over the test output, and your previous messages, and I'm sad to say that it did not catch the regression in https://github.com/ProvableHQ/snarkVM/pull/3144 , which is what I had hoped. At least that also means we don't have to make use of the complexity we just created.

The `build.rs` changes could be useful for more tests, but mainly on the snarkOS side. Can snarkOS' `build.rs` read snarkVM source code files?

ljedrz:
> Can snarkOS' build.rs read snarkVM source code files?

Sadly no, _but_ snarkVM can be a build dependency, meaning we could just derive the number of consensus versions/heights from the relevant `enum`.

PR #3154, opened 2026-02-17
[CI] Run latest canary release

vicsn:


raychu86:
Note: Testnet + Mainnet will likely need an update to ConsensusVersion::V13 height. Canary _probably_ won't because the migration height has already passed

PR #3158, opened 2026-02-20
[Perf] Reuse past Merkle trees

ljedrz:
Block insertions have become slower over time, which especially impacts block sync. CPU profiling using a large (~700GB) ledger and a nontrivial number (~1500) of subsequent blocks have revealed that the related memory operations are the culprit:

image

The above flamegraph clearly indicates that `MerkleTree::prepare_append` is the (high-level) problematic operation, which is unsurprising, as for every new block, the tree is currently fully recreated. This, of course, involves several allocations and reallocations, which become slower with size. It is very tricky to work around too, because the append operation is atomic (any related failure causes the original tree to be retained), and we may not use `unsafe` code due to the `#![forbid(unsafe_code)]` attribute applicable to the entire crate. `rayon` would not help here either, as allocations themselves (as opposed to inserting values) are the bottleneck.

The solution is to save and reuse the previous Merkle tree whenever we call `prepare_append`, which allows us to retain large allocations. They are very likely to be larger than immediately needed (due to amortized growth), improving performance for significantly longer than a single operation (which only uses a single new "slot"). In order to fully preserve the existing API, the new `MerkleTree` member is wrapped in a `Mutex`, facilitating isolated updates.

This change resulted in a **~45.5%** reduction in block insertion time for the aforementioned group of blocks, and had a clear impact on the breakdown of CPU work:

image

As for the impact on allocations, it is also apparent and beneficial:
- despite the reduction in the number of allocations being only ~0.01%, it resulted in the average allocation size dropping by **99.6%**
- despite the reduction in the number of reallocations being only ~0.04%, it resulted in the average reallocation size dropping by **99.7%**

These effects would be more pronounced with larger ledgers, and less noticeable with smaller ones (but still beneficial).

The only downside is that maintaining a copy of the Merkle tree at all times increases the amount of memory used, but no more than the amount currently needed to perform this operation (just permanently).


ljedrz:
Pretty sure the one failed test is unrelated, but I won't restart the job just yet, as there may be some large internet infrastructure outage going on right now (I'm seeing failures in several different services).

vicsn (commented):
Very nice. 

What are your thought:
- adding an `fn prepare_append_mut` to do the update in place could be easier to follow, but would require duplicating way more lines of code...
- could we have a new `StorageMerkleTree` wrapper class which wraps setting of the usage of the `previous_tree`?

And do I see we are still recomputing the tree? Is that a potential bottleneck now that the superfluous allocations are resolved?

ljedrz:
> adding an `fn prepare_append_mut` to do the update in place could be easier to follow, but would require duplicating way more lines of code...

I've considered it, but yeah, it would add plenty of extra lines, and we would lose the atomicity, which I recall being an important feature. That being said, once the allocations are preserved, its cost is minimal.

> could we have a new `StorageMerkleTree` wrapper class which wraps setting of the usage of the previous_tree?

What would be its purpose? To facilitate the preservation of tree allocations in other Merkle tree implementations?

> And do I see we are still recomputing the tree? Is that a potential bottleneck now that the superfluous allocations are resolved?

We are; it is likely possible to avoid some of that work now that we have a copy available, but I'd recommend doing it separately; the same goes for optimizing the use of `rayon`, which - for my liking - is spending too much time doing work stealing (~8% of the profile, which should be reducible to <3%).

vicsn:
You probably saw it but surprisingly just a single CI test is failing with "The given Merkle leaf does not match the one in the Merkle tree". Do you also want to enable to `merge-workflow` tests in `.circleci/config.yml`?

> > could we have a new StorageMerkleTree wrapper class which wraps setting of the usage of the previous_tree?
> What would be its purpose? To facilitate the preservation of tree allocations in other Merkle tree implementations?

My suggestion was that we have *some* kind of new merkle tree abstraction without adding too many extra lines of merkle tree business logic so that in `ledger/store/src/block/mod.rs` we would only have to do one call to `self.tree.write().prepare_append_mut(...)`, and which internally calls:
```
let updated_tree = self.prepare_append(&[block.hash().to_bits_le()])?;
updated_tree.update_previous_tree(&mut self);
self = updated_tree
```
Not sure if this is possible or desirable, just wanted to mention it as an option.

> We are; it is likely possible to avoid some of that work now that we have a copy available, but I'd recommend doing it separately; the same goes for optimizing the use of rayon, which - for my liking - is spending too much time doing work stealing (~8% of the profile, which should be reducible to <3%).

Sounds good we can make separate issues.

ljedrz:
> Not sure if this is possible or desirable, just wanted to mention it as an option.

Hmm unless it takes a closure, we would lose some of the atomicity (potential rollback when the height is incorrect or the block insertion fails); however, that closure might not be usable, since the console collections are in a pretty "low level" crate, which doesn't know about the existence of blocks or the storage layer. However, the block insertion is the only place we ever use `MerkleTree::prepare_append`, and it's strictly a perf optimization, so I wouldn't worry about it being written a bit more verbosely.

> Sounds good we can make separate issues.

Will do; btw, the one failing test case should pass now.

vicsn (approved):

console/collections/src/merkle_tree/mod.rs
```diff
@@ -159,8 +183,10 @@ impl<E: Environment, LH: LeafHash<Hash = PH::Hash>, PH: PathHash<Hash = Field<E>
         // Compute the number of padded levels.
         let padding_depth = DEPTH - tree_depth;
 
-        // Initialize the Merkle tree.
-        let mut tree = vec![self.empty_hash; num_nodes];
+        // Reuse the previous Merkle tree, or initialize it if missing.
+        let mut tree = self.previous_tree.lock().take().unwrap_or_else(|| vec![self.empty_hash; num_nodes]);
```
raychu86: It wasn't clear to me at first that the state of `previous_tree` doesn't matter and is written over later (we only care that the VEC was allocated).


Might be nicer to have a lazy_static or OnceCell for pre-allocated Vecs (of each required power of two) that contains just the empty hashes.

ljedrz: while elegant and would remove the write perf penalty (the cost of populating the new trees with empty hashes), it would increase RAM use, and still require full allocations (clones of the zeroed trees) for every tree extension, which would still be a performance issue

raychu86: I see your point. 

Then, I'd say the naming of `previous_tree` could be improved to clarify that it doesn't actually require the exact previous tree, but just a vec that has already been allocated.

ljedrz:
[Addressed](https://github.com/ProvableHQ/snarkVM/pull/3158/commits/16785e6932b3e782e33102b181766e88c6065bfc) the review comments.

raychu86 (approved):

PR #3159, opened 2026-02-21
[Feat] Implement `RejectedReason`

raychu86:
## Motivation

This PR introduces `RejectedReason`s that stores why a transaction was rejected in the `Rejected` object. Depending on the reason, there may be a locator, index, and command stored as well. This will help with developer debugging as they can better trace where their errors are coming from.

The indexing is done on the Aleo Instruction level, so we may eventually want a tool that points to where in Leo code the error originates from.

Currently the 3 types are:
- DuplicateProgramID (Duplicate program ID deployment in the same block)
- Finalize (Finalize/constructor command failed)
- NonFinalize (VM error such as stack failure or bond_public restrictions)

Alternative to #3101.

## Test Plan

Tests have been added to ensure that the RejectedReasons are properly constructed and only occur after the desired consensus version.

## Backwards compatibility

This functionality is guarded with ConsensusVersion::V14, where rejected reasons are only stored after V14, and are invalid before then.

## TODO

- [x] Contemplate adding the arguments that caused the error to the RejectedReason (This can cause state bloat and is a DOS vector)
- [ ] Consider introducing an AbortedReason as well.


ljedrz (commented):
The general approach looks reasonable and maintainable to me.

Storage/serialization-wise, we need to ensure that the error values are constructed in a way that ensures that the database compression does its job well, which is even more important than keeping them short. For instance, a value like `xxxxxxxxxxxxVARyyyyyyyyyy` is strictly worse than `xxxxxxxxxxxxyyyyyyyyyyVAR` (where `x`s and `y`s are constant, and `VAR` varies) for dictionary-based compression that's used by rocksDB+LZ4.

We may also want to create some new test cases ensuring that the constant parts of the error strings do not change (I saw some new test cases, but I think they were concentrated on ensuring that the right errors are returned, as opposed to their string values).

synthesizer/process/src/finalize.rs
```diff
@@ -117,35 +128,44 @@ impl<N: Network> Process<N> {
...
-        let number_of_calls = stack.get_number_of_calls(transition.function_name())?;
-        ensure!(
-            number_of_calls == execution.len(),
-            "The number of transitions in the execution is incorrect. Expected {number_of_calls}, but found {}",
-            execution.len()
-        );
+        let number_of_calls = stack
+            .get_number_of_calls(transition.function_name())
+            .into_indexed(transition_locator, None::<(usize, &str)>)?;
+        if !number_of_calls == execution.len() {
```
d0cd:
```suggestion
        if number_of_calls != execution.len() {
```

raychu86: [bff20aa](https://github.com/ProvableHQ/snarkVM/pull/3159/commits/bff20aa4a96b9e980a619898f05c4303b120df86)

raychu86:
> Storage/serialization-wise, we need to ensure that the error values are constructed in a way that ensures that the database compression does its job well, which is even more important than keeping them short. For instance, a value like `xxxxxxxxxxxxVARyyyyyyyyyy` is strictly worse than `xxxxxxxxxxxxyyyyyyyyyyVAR` (where `x`s and `y`s are constant, and `VAR` varies) for dictionary-based compression that's used by rocksDB+LZ4.
> 
> We may also want to create some new test cases ensuring that the constant parts of the error strings do not change (I saw some new test cases, but I think they were concentrated on ensuring that the right errors are returned, as opposed to their string values).

@ljedrz I've refactored `RejectedReason` to store concrete objects rather than strings, this should help alleviate the compression concerns - [990183e](https://github.com/ProvableHQ/snarkVM/pull/3159/commits/990183e68e230dec5bb01d86559f904ec1eee6dd).

ledger/block/src/transactions/rejected/mod.rs
```diff
@@ -52,32 +56,40 @@ impl<N: Network> Rejected<N> {
...
         match self {
-            Self::Deployment(_, _) => None,
-            Self::Execution(execution) => Some(execution),
+            Self::Deployment(_, _, _) => None,
+            Self::Execution(execution, _) => Some(execution),
+        }
+    }
+
+    /// Returns the rejected reason.
+    pub fn rejected_reason(&self) -> Option<RejectedReason<N>> {
```
d0cd: Perhaps worth having a referenced option.

ledger/block/src/transactions/rejected/serialize.rs
```diff
@@ -59,14 +71,24 @@ impl<'de, N: Network> Deserialize<'de> for Rejected<N> {
                         // Parse the deployment.
                         let deployment: Deployment<N> =
                             DeserializeExt::take_from_value::<D>(&mut object, "deployment")?;
+                        // Parse the optional rejected reason.
+                        let rejected_reason: Option<RejectedReason<N>> = match object.get("rejected_reason") {
```
d0cd: I believe this could be more concisely written as:  `DeserializeExt::take_from_value::<D>(&mut object, "rejected_reason").ok()`

ledger/block/src/transactions/rejected_reason/mod.rs
```diff
@@ -0,0 +1,83 @@
...
+/// The reason a transaction was rejected.
+#[derive(Clone, PartialEq, Eq)]
+pub enum RejectedReason<N: Network> {
+    /// The transaction was rejected due to a duplicate program ID deployment in the same block.
+    DuplicateProgramID(ProgramID<N>),
+
+    /// The transaction was rejected due to a failed finalize command. (program ID, resource, index, command).
+    /// Note: We do not log the actual error message from the finalize command, as it may contain
+    /// sensitive information or lead to DOS vectors by storing string representations of large structs.
+    Finalize(ProgramID<N>, Identifier<N>, usize, Command<N>),
```
d0cd: An erroring `assert.eq` command can be up to 2KB. 
It's not that large, but it's worth considering removing `Command<N>` from the enum variant. 
The other fields provide enough information for tools to index and find the right command.
Some benefits include smaller enums and a simpler implementation for `from_indexed_finalize_error`.

One added note, we should probably also track the program edition.

ledger/block/src/transactions/rejected/serialize.rs
d0cd: Ditto for similar patterns in other files.

ledger/block/src/transactions/confirmed/mod.rs
```diff
@@ -418,7 +429,13 @@ pub mod test_helpers {
         };
 
         // Extract the rejected deployment.
-        let rejected = crate::rejected::test_helpers::sample_rejected_deployment(version, edition, is_fee_private, rng);
+        let rejected = crate::rejected::test_helpers::sample_rejected_deployment(
+            version,
+            edition,
+            is_fee_private,
+            has_rejected_reason,
```
d0cd, line 436: Should a greater percentage of tests set this to `true` since this is the new default behavior?

synthesizer/error/src/process.rs
```diff
@@ -189,3 +193,121 @@ impl<E> IndexedInstructionError<E> {
...
+        error: FinalizeError,
+    ) -> Self {
+        Self { program_id, resource, command: command.map(Box::new), error }
+    }
+}
+
+/// A helper macro to bail with an `IndexedFinalizeError`.
+///
+/// Two forms:
+///   - `indexed_finalize_bail!(program_id, resource, index, command, "message {}", args)` — with command context.
```
d0cd: Could this documentation be made more clear? I was confused when I didn't see "message {}" in the macro.

ledger/block/src/transactions/rejected/mod.rs
raychu86: [ca9b3c0](https://github.com/ProvableHQ/snarkVM/pull/3159/commits/ca9b3c02c2fa530955ea5eaaf694fd92c860e8c5)

ledger/block/src/transactions/rejected/serialize.rs
raychu86: [ca9b3c0](https://github.com/ProvableHQ/snarkVM/pull/3159/commits/ca9b3c02c2fa530955ea5eaaf694fd92c860e8c5)

synthesizer/error/src/process.rs
raychu86: [ca9b3c0](https://github.com/ProvableHQ/snarkVM/pull/3159/commits/ca9b3c02c2fa530955ea5eaaf694fd92c860e8c5)

ledger/block/src/transactions/confirmed/mod.rs
raychu86: [ca9b3c0](https://github.com/ProvableHQ/snarkVM/pull/3159/commits/ca9b3c02c2fa530955ea5eaaf694fd92c860e8c5)

ledger/block/src/transactions/rejected_reason/mod.rs
raychu86: Added edition here - [70bb92e](https://github.com/ProvableHQ/snarkVM/pull/3159/commits/70bb92ead86c450eaa17967a4ddb1b61b5591d78).

raychu86: Is the 2KB maximum cause for concern? It is a DOS vector, but not sure if it's worth saving on the abstraction of requiring state and doing the query.

raychu86: A bandaid solution for the future: we can consider disabling the serialization if the bloat gets out of hand.

Though it would be best to have a proper decision now.

d0cd: Personally, I'd lean towards removing `Command<N>`. 
Dapps and client code can easily cache programs and do the lookups on their own.
That said, it's probably worth getting the opinions of those who'd use this feature (partners and the Explorer/SDK team).

PR #3169, opened 2026-03-02
@fed/reqwest impl parameters

fedpre:
## Motivation

Updated the remote fetching implementation to use reqwest instead of curl to avoid issues with CA certificate. This changes uses the `native-tsl` from `reqwuest` without any additional step. 

## Test Plan

Tested on both iOS and Android devices and the change works.

## Backwards compatibility

Please review backwards compatibility. Does any functionality need to be guarded by an existing or new `ConsensusVersion`?


ledger/query/Cargo.toml
```diff
@@ -67,7 +67,8 @@ features = [ "preserve_order" ]
 
 [dependencies.reqwest]
 version = "0.13"
-features = [ "json" ]
+default-features = false
```
vicsn, line 70: Is this needed? If not can you revert (i.e. remove this line)

fedpre: It is and the reason is that with version 0.13 it installs by default some `aws-ls-*` libraries that break android. In this way, we avoid that. I tried several ways but android keeps crashing.

vicsn: Is it possible to make this a target-dependent exception like we have in`parameters/Cargo.toml`? It's hard to test on all possible platforms and we want to avoid breaking changes

fedpre: I will take a look if it is possible to do

vicsn: Update: this seems impossible: https://github.com/rust-lang/cargo/issues/1197

Let's take the risk, if some client software depending on snarkVM breaks then we can figure out a resolution

vicsn (approved):

ledger/query/Cargo.toml
fedpre: Ok sounds good. Let's wait on @iamalwaysuncomfortable review and then I guess we can proceed

iamalwaysuncomfortable:
@vicsn is this still mergeable?

ledger/query/Cargo.toml
iamalwaysuncomfortable: > Update: this seems impossible: rust-lang/cargo#1197 Let's take the risk, if some client software depending on snarkVM breaks then we can figure out a resolution

@vicsn @fedpre   It's impossible not with resolver v2, we do it on the very next line of this Cargo.toml.

```toml
[target.'cfg(target_arch="wasm32")'.dependencies.ureq]
workspace = true
features = [ "json", "gzip" ]
optional = true

[target.'cfg(not(target_arch="wasm32"))'.dependencies.ureq]
workspace = true
features = [ "json", "gzip", "rustls" ]
optional = true
```

Therefore this should work:
```toml
[target.'cfg(any(target_os = "android", target_os = "ios"))'.dependencies.reqwest]
version = "0.13"
default-features = false
features = [ "json", "native-tls" ]
optional = true

[target.'cfg(not(any(target_os = "android", target_os = "ios")))'.dependencies.reqwest]
version = "0.13"
default-features = false
features = [ "json" ]
optional = true
```

ledger/query/Cargo.toml
```diff
@@ -67,7 +67,8 @@ features = [ "preserve_order" ]
...
+features = [ "json", "native-tls" ]
```
iamalwaysuncomfortable: Native-tls will break `wasm` targets and have easily the capability of breaking multiple other Rust implementations who are using `rust-tls` as this will enforce `native-tls`.

To avoid downstream breakage, the change should be this:

```rust
[target.'cfg(any(target_os = "android", target_os = "ios"))'.dependencies.reqwest]
version = "0.13"
default-features = false
features = [ "json", "native-tls" ]
optional = true

[target.'cfg(not(any(target_os = "android", target_os = "ios")))'.dependencies.reqwest]
version = "0.13"
default-features = false
features = [ "json" ]
optional = true
```

parameters/Cargo.toml
```diff
@@ -93,8 +93,10 @@ version = "0.3.64"
 features = [ "XmlHttpRequest" ]
 optional = true
 
-[target.'cfg(all(not(target_family = "wasm"), not(target_env = "sgx")))'.dependencies.curl]
-version = "0.4.43"
+[target.'cfg(all(not(target_family = "wasm"), not(target_env = "sgx")))'.dependencies.reqwest]
+version = "0.13"
+default-features = false
+features = [ "blocking", "native-tls" ]
```
iamalwaysuncomfortable: Again `native-tls` would break wasm and easily break other could tls stacks as it'd enforce native-tls usage. Just target the mobile platforms specifically.

```suggestion
[target.'cfg(all(
    any(target_os = "android", target_os = "ios"),
    not(target_family = "wasm"),
    not(target_env = "sgx")
))'.dependencies.reqwest]
version = "0.13"
default-features = false
features = [ "blocking", "native-tls" ]

[target.'cfg(all(
    not(any(target_os = "android", target_os = "ios")),
    not(target_family = "wasm"),
    not(target_env = "sgx")
))'.dependencies.reqwest]
version = "0.13"
default-features = false
features = [ "blocking" ]
```

iamalwaysuncomfortable (dismissed):

fedpre:
@iamalwaysuncomfortable I have made the requested changes and it finally works on both iOS and Android - please check thanks

iamalwaysuncomfortable (commented):
This looks like the right solution. Moving to reqwest is a definite a portability upgrade and is likely to be maintained better than curl in future.

One final nit, why did the toml files get reformatted? Otherwise this is approved.

fedpre:
> This looks like the right solution. Moving to reqwest is a definite a portability upgrade and is likely to be maintained better than curl in future.
> 
> One final nit, why did the toml files get reformatted? Otherwise this is approved.

It is because I have installed `even better toml` on Cursor and it automatically formats it correctly. If you want, I can revert those changes.

vicsn (approved):

PR #3171, opened 2026-03-03
[Feat] Store historical rewards

### LINKED ISSUE #3163: [Feature] Staking Rewards Storage
## 🚀 Feature

Store changes in rewards per staker in RocksDB, queryable by REST API
- Add a new RocksDB map keyed by (staker_address, block_height) → StakingRewardEntry { validator, reward, new_stake }
- In atomic_post_ratify, after computing staking rewards via staking_rewards(), persist the per-staker reward amounts
- Expose new REST endpoints gated behind the feature flag:
  - GET /staking/rewards/{address}/{height} - reward at specific height

ljedrz:
Closes https://github.com/ProvableHQ/snarkVM/issues/3163.

### DIFF
--- a/synthesizer/Cargo.toml
+++ b/synthesizer/Cargo.toml
@@ -34,6 +34,7 @@ locktick = [
 async = [ "snarkvm-ledger-query/async", "snarkvm-synthesizer-process/async" ]
 cuda = [ "snarkvm-algorithms/cuda" ]
 history = [ "snarkvm-ledger-store/history" ]
+history-staking-rewards = [ "snarkvm-ledger-store/history-staking-rewards" ]
 rocks = [ "snarkvm-ledger-store/rocks" ]
 serial = [
   "snarkvm-console/serial",
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -128,6 +128,7 @@ package = [
 async = [ "snarkvm-ledger/async", "snarkvm-synthesizer/async" ]
 cuda = [ "snarkvm-algorithms/cuda" ]
 history = [ "snarkvm-synthesizer/history" ]
+history-staking-rewards = [ "snarkvm-synthesizer/history-staking-rewards" ]
 parameters_no_std_out = [ "snarkvm-parameters/no_std_out" ]
 locktick = [
   "snarkvm-console?/locktick",
--- a/ledger/Cargo.toml
+++ b/ledger/Cargo.toml
@@ -66,6 +66,10 @@ history = [
   "snarkvm-ledger-store/history",
   "snarkvm-synthesizer/history",
 ]
+history-staking-rewards = [
+  "snarkvm-ledger-store/history-staking-rewards",
+  "snarkvm-synthesizer/history-staking-rewards",
+]
 locktick = [
   "dep:locktick",
   "snarkvm-ledger-puzzle/locktick",
--- a/ledger/src/tests.rs
+++ b/ledger/src/tests.rs
@@ -38,6 +38,10 @@ use snarkvm_ledger_block::{Block, ConfirmedTransaction, Execution, Ratify, Rejec
 use snarkvm_ledger_committee::{Committee, MIN_VALIDATOR_STAKE};
 use snarkvm_ledger_narwhal::{BatchHeader, Data, Subdag, Transmission, TransmissionID};
 use snarkvm_ledger_store::ConsensusStore;
+#[cfg(feature = "history-staking-rewards")]
+use snarkvm_ledger_store::helpers::MapRead;
+#[cfg(feature = "history-staking-rewards")]
+use snarkvm_synthesizer::bonded_map_into_stakers;
 use snarkvm_synthesizer::{
     program::{Program, StackTrait},
     vm::VM,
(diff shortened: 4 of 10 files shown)


synthesizer/Cargo.toml
```diff
@@ -34,6 +34,7 @@ locktick = [
 async = [ "snarkvm-ledger-query/async", "snarkvm-synthesizer-process/async" ]
 cuda = [ "snarkvm-algorithms/cuda" ]
 history = [ "snarkvm-ledger-store/history" ]
+history-staking-rewards = [ "snarkvm-ledger-store/history-staking-rewards" ]
```
dgrkotsonis, line 37: I think this needs to be bubbled up to the top level Cargo.toml file, currently having issues compiling this with the history-staking-rewards flag for testing on a node

dgrkotsonis:
NOTE: Testing this and encountering issues querying certain mappings beyond staking (e.g. if you query one of test instances at:`http://34.145.46.115:3030/mainnet/program/credits.aleo/mapping/bonded/aleo1qy4qufq03wcph05fdf5aj09ez67vcmmlrzqf0zza352qwaq43gyqt3wdf6/history/5000` should return `"{\n  validator: aleo1vfukg8ky2mhfprw63s0k0hl4vvd8573s6fkn8cv9y0ca6q27eq8qwdnxls,\n  microcredits: 140666067980u64\n}"` but it returns null). Continuing to look into this now but just flagging

ljedrz:
@dgrkotsonis in the end this feature is separate from `history`, so this is expected without it in force.

dgrkotsonis:
> @dgrkotsonis in the end this feature is separate from `history`, so this is expected without it in force.

Sorry just to confirm this was compiling it with both `history` and the new `history-staking-rewards` flags, I was just running through a checklist of endpoints and noticed that the metadata was working but the bonded wasn't, looking for others now

ljedrz:
@dgrkotsonis this PR should have no impact on the preexisting `history` feature; are you sure it's the right address?

dgrkotsonis:
@ljedrz Yeah I noticed a few odd things I'm still trying to figure out but these are some observations:

1. The metadata mappings work as normal so the history flag is enabled and at least partially working, but for other mappings, they return null for the first couple hundred thousand blocks and then are seemingly returning values for blocks the ledger hasn't hit yet (e.g. my mainnet node I'm testing these changes with has synced up to block ~[2.8 million](http://34.145.46.115:3030/mainnet/block/height/latest) but when I request mappings from block 5 million for example, it returns non-null values - see [here](http://34.145.46.115:3030/mainnet/program/credits.aleo/mapping/bonded/aleo1qy4qufq03wcph05fdf5aj09ez67vcmmlrzqf0zza352qwaq43gyqt3wdf6/history/5000000). This could also just be because if we request a block higher than the ledger height snarkos/vm just default to the latest block? Is that the desired behavior? (CC: @vicsn this question may be more up your alley)

2. The mappings values it returns (apart from metadata mappings it seems) are incorrect. E.g. if you take a look at the bonded values at block 800,000 for the key `aleo1qy4qufq03wcph05fdf5aj09ez67vcmmlrzqf0zza352qwaq43gyqt3wdf6` on an older historical staking node (Can see [here](http://34.127.80.245:3030/mainnet/block/800000/history/bonded)) it shows `"{\n  validator: aleo1vfukg8ky2mhfprw63s0k0hl4vvd8573s6fkn8cv9y0ca6q27eq8qwdnxls,\n  microcredits: 140647951613u64\n}"` But when querying my node with the latest changes from this branch + the new history and staking flags it shows `"{\n  validator: aleo1vfukg8ky2mhfprw63s0k0hl4vvd8573s6fkn8cv9y0ca6q27eq8qwdnxls,\n  microcredits: 140624568486u64\n}"` (see [here](http://34.145.46.115:3030/mainnet/program/credits.aleo/mapping/bonded/aleo1qy4qufq03wcph05fdf5aj09ez67vcmmlrzqf0zza352qwaq43gyqt3wdf6/history/800000)). Is it possible there's an issue with how we handle the keys for retrieval?

3. Do we have a way to retrieve staking mappings from the vm store at the moment? E.g. something comparable to [get_historical_mapping_value()](https://github.com/ProvableHQ/snarkVM/blob/staging/ledger/store/src/program/finalize.rs#L696-L704) At the moment not sure how I can query staking mappings at certain blocks

ljedrz:
@dgrkotsonis 

1. The basic logic is that for every mapping update, we register both the individual update (the program id, mapping, key, and value) at the specific height, and all the heights at which all updates of the mapping have happened. Then, the applicable REST route uses [this logic](https://github.com/ProvableHQ/snarkVM/blob/staging/ledger/store/src/program/finalize.rs#L698) to return the value from the specific update at the height where it became "binding" ([playground demonstration](https://play.rust-lang.org/?version=stable&mode=debug&edition=2024&gist=3eb77f58706bb182bb2b165ef19d23c7)).

2. Could you provide a "minified" (something that's apparent at a low height) example of a mapping (or multiple) which isn't retrieved correctly in a local devnet or the current testnet/mainnet, and the expected values at specific queried heights? I could then double-check the exact logic that's being used.

3. Not yet (you would currently need to use `store.staking_rewards_map().get_confirmed(&(staker, height))`, but I can prepare a draft shortly.

dgrkotsonis:
@ljedrz

1. I think [this](https://github.com/ProvableHQ/snarkVM/blame/staging/ledger/store/src/program/finalize.rs#L715) line is causing it to return whatever the mapping value is for the latest block if you request a block that doesn't exist yet. I would think that we'd only want to get values for exact blocks so as not to cause confusion - @vicsn does this sound right?

2. My bad, I forgot that you are probably unable to query my nodes due to the firewalls, sure as an example:
(Node A is my existing history node with the older logic Mia and I added, Node B is my new node with these latest changes)

I'll be using block number 4 for these (chosen arbitrarily)

- Metadata: For metadata, Node A and Node B seem consistent. For the mapping key `aleo1qgqqqqqqqqqqqqqqqqqqqqqqqqqqqqqqqqqqqqqqqqqqqqqqqqqqanmpl0`, both show "93u32"
- Bonded: For the key `aleo1qy4qufq03wcph05fdf5aj09ez67vcmmlrzqf0zza352qwaq43gyqt3wdf6`, node A shows `"{\n  validator: aleo1vfukg8ky2mhfprw63s0k0hl4vvd8573s6fkn8cv9y0ca6q27eq8qwdnxls,\n  microcredits: 140625037214u64\n}"` whereas node B returns `null`
- Delegated: For the key `aleo1vfukg8ky2mhfprw63s0k0hl4vvd8573s6fkn8cv9y0ca6q27eq8qwdnxls`, Node A shows "106334771267012u64" but Node B returns `null`
- Withdraw: For the key `aleo1q3gtqtd03fs7chhjdr8c4hf8vkwt96pf3vw28uytsdrnwt4hrs9sg7c62j`, we get `aleo13kxtmlp3u37m9np4zf465rapd2h8gxhxv5wmp8q852hxhnnfac8s9769tn` from Node A but `null` from Node B

(Let me know if any of this doesn't make sense, I'm happy to set up a quick call to just show you what I see on my end)

3. Sounds great, I see you have a PR open I can take a look thanks!

dgrkotsonis:
@ljedrz I wiped my nodes and restarted and am no longer seeing the error. Will try to recreate but it could have just been a strange issue due to me starting from a stale ledger? Either way I'm no longer seeing issue #2 I raised, though issue #1 is still occurring for historical mappings (but not staking rewards)

ljedrz:
Indeed, starting from a preexisting ledger would explain the lack of some of the historical records. As for issue 1 (returning the value from the latest block when asking for a future height), it's a simple fix (just return null for future heights), I can do it first thing tomorrow.

dgrkotsonis:
Sounds great! I can go ahead and add that myself if it's late for you, thanks @ljedrz !

dgrkotsonis (approved):

ljedrz:
@dgrkotsonis actually, getting the previous historical value is the desired behavior - it just shouldn't happen if the given height is greater than the current height.

@vicsn please correct me if this goal has changed

vicsn:
Agreed with @ljedrz latest statement. A user should be able to query a historic value at any valid block height IMO

ljedrz:
Adjusted. The reasoning is that, for instance, at a current height of 10, if a mapping value had changed at height 5 and had no further changes, its value was the same as at height 5 for heights 6-10. The latest commit restores this behavior, but will now return nothing if a future height is given (as it's not a historical value, and the actual value at that height might still change).

vicsn (approved):

PR #3109, opened 2026-01-18
[WIP] Ensure records are linked

vicsn:
Closes https://github.com/ProvableHQ/snarkVM/issues/2667

## Impacted Programs
These programs have functions that violate the "consumption" requirement. Consequently those specific function will not be executable after `ConsensusVersion::V14`.

Analysis code can be found here: https://github.com/ProvableHQ/aleo-program-regressions/pull/32

### Mainnet
  - squash_v001.aleo/mint
  - squash_v001.aleo/water
  - squash_v001.aleo/water_and_level_up
  - squash_v2.aleo/migrate_from_v1
  - squash_v2.aleo/mint
  - squash_v2.aleo/water
  - squash_v2.aleo/water_and_level_up
  - squash_v3.aleo/migrate_from_v2
  - squash_v3.aleo/mint
  - squash_v3.aleo/water
  - squash_v3.aleo/water_and_level_up
  - squash_v4.aleo/migrate_from_v3

### Testnet
  - amm_interface_v006.aleo
  - authssnverify.aleo
  - collateral.aleo
  - consumer_program_xyz.aleo
  - eclipse_lending_usda_vault.aleo
  - house_dealer.aleo
  - just_take_record2.aleo
  - mana_collateral.aleo
  - nft_collections_v3.aleo
  - nft_registry_v3.aleo
  - otc_derivate_4.aleo
  - par_store_v5.aleo
  - privacypay_dao_v0005.aleo
  - pusdc_28541.aleo
  - pusdc_28542.aleo
  - trust_trust_trust.aleo
  - wheres_alex_v019.aleo
  - zerosecure_governance_manager.aleo
  - zerosecure_transfer_manager.aleo
  - zkescrow_testing.aleo/wrap
  - zmarket_marketplace_v3.aleo
  - zmarket_private_marketplace.aleo
  - zmarket_royalties_v3.aleo

PR #2936, opened 2025-09-25
[Fix] Set console caller and signer when executing closures.

vicsn:
## Motivation

Revival of: https://github.com/ProvableHQ/snarkVM/pull/2525

I don't believe this exposed a soundness issue in the VM, just a lack of functionality.

## Test Plan

Added a new test

- [x] Fix the test. Currently it fails with: "thread '<unnamed>' panicked at synthesizer/tests/test_vm_execute_and_finalize.rs:158:9: assertion failed: aborted_transaction_ids.is_empty()"

PR #3173, opened 2026-03-04
Ensure records exist

Antonio95:
This PR implements work that was initiated in #3109.

This PR adds a function `ensure_records_exist` which, given an execution, ensures that every `Transition` `t` in it satisfies the following:

> **(Guarantee)** Every non-static record (`DynamicRecord` or `ExternalRecord`) input to `t` or received by `t` from a *function* callee _corresponds to_ a static `Record` which exists on the ledger at the end of the full execution (whether consumed or not).

Here "corresponds to" refers to the relation between two record-like (`Record`, `DynamicRecord` or `ExternalRecord`)  registers that arises a) when they coincide at a *function or closure* input or output boundary (caller  register vs. callee register), or b) when they are connected by a cast-to-dynamic instruction.

Among other things, this makes read operations on `DynamicRecord` or `ExternalRecord` more meaningful.

At the same time, the following is required of every transition `t` in the execution:
> **(Contract)** Assuming (Guarantee), every record-like (`Record`, `DynamicRecord` or `ExternalRecord`) register output by `t` or passed by `t` to a *function* callee (but not necessarily to a closure) exists on the ledger at the end of the full execution (whether consumed or not).

It is not difficult to see that the guarantee can be ensured as follows (note there is a recursive relation between (Guarantee) and (Contract), as the latter assumes the former locally):

1) Keeping track of `DynamicRecord` and `ExternalRecord` registers input to the root call across call boundaries and cast instructions, and ensuring that, for each family of connected record registers, one member register is input to a function receiving it as a static `Record` (thus consuming it). We refer to this as the _global check_.

2) Ensuring that, in each function `f` in the execution, if a static `Record` `R_s` is minted locally and cast to a `DynamicRecord` `R_d` such that `R_d` is then output or passed to a callee _function_ of `f`,  then `R_s` is also output by `f`. We refer to this as the _local check_.

`ensure_records_exist` performs both checks. This is only done at runtime (on both the prover and verifier sides). Even though the local check might seem like it could be done at deployment time, this is not quite the case due to external closure calls. More specifically: a deployment-time local check would analyse the flow of record-like registers in external closure calls as they exist on the ledger at the time of deployment. However, those external closures could be updated later on and alter the flow of record-like registers (for instance, by casting them or swapping the order of input or output registers). This means the local check still has to be done at runtime (at least for programs involving external closure calls which receive or produce record-like values). If we are interested in _also_ including it at deployment time, this would not be too difficult due to some design choices in the algorithm.

In terms of the algorithm itself, `ensure_records_exist` explores the execution tree recursively starting at the root `Transition`. `Transition`s are explored with the private function `process_transition` in the same file. Closure calls are different enough from function ones in terms of the above checks that it was decided to process them with a separate function, `process_closure`. For instance, a closure can affect the local check of its caller, whereas a function cannot (for an interesting edge case, consider the case where a function mints a static record, passes it to an external closure which casts external-to-dynamic and outputs as dynamic; and the caller receives the latter and attempts to output it itself). An alternative design was discussed with @d0cd where both cases could be covered by the same function (introducing scopes to handle their differences appropriately), which we can circle back on if the bespoke closure treatment in the current implementation is not clear enough.

A fair number of test cases has added to the new file `synthesizer/src/vm/tests/test_v14/record_existence.rs`. I still intend to ensure (and document) that they cover every path of the ensure_records_exist algorithm.

I recently merged `staging` in order to include some changes (https://github.com/ProvableHQ/snarkVM/pull/2936) needed for one of the test cases. If this has polluted the diff too much, we may want to merge `staging` into `feat/dynamic_dispatch` as well.


rot256:
Still reading it, but an immediate first thought: why not use the standard "set union" data structure (https://en.wikipedia.org/wiki/Disjoint-set_data_structure) which would make this a O(n) pass? It looks to be quadratic in the worst case atm? Because we are iterating over all the families to find the one to which a register belongs.

.claude/CLAUDE.md
d0cd, line 1: Does this need to be delted?

.circleci/config.yml
```diff
@@ -89,6 +89,7 @@ commands:
       cache_key:
         type: string
         default: v4.5.0-rust-1.88.0-snarkvm-cache
+        default: v4.5.0-rust-1.88.0-snarkvm-cache
```
d0cd: These look like duplicates?

synthesizer/process/src/stack/call/dynamic.rs
```diff
@@ -138,6 +138,9 @@ impl<N: Network> CallTrait<N> for CallDynamic<N> {
         }
         // Else, throw an error.
         else {
+            // TODO (Antonio) remove
```
d0cd: Flagging TODOs.

synthesizer/process/src/verify_execution.rs
```diff
@@ -778,4 +786,532 @@ impl<N: Network> Process<N> {
...
+                        let destination_register = match cast.destinations().first() {
+                            Some(destination) => destination.locator(),
+                            _ => bail!(
+                                "Failed to retrieve destination register for cast to DynamicRecord instruction in closure {closure_name}"
+                            ),
+                        };
+
+                        if callee_locally_minted_static.contains(&operand_register) {
+                            // Case 2: Casting a locally minted static Record to a DynamicRecord. We ensure the latter is not output.
+                            if output_map.contains_key(&destination_register) {
```
rot256: A bit confusing that the validation logic for casting a locally minted static record to a dynamic record is different here than in the case of a function:

Here an `output_map` is initialized up-front and casts are checked eagerly (which is cleaner), where as in a function, this is deferred to the end and we check that `must_be_output` is empty after removing all the outputs (less elegant).

.claude/CLAUDE.md
Antonio95: I am unsure, I had to resolve a conflict on that file twice, both when I merged `staging` and today when I merged `feat/dynamic_dispatch`. Let me know if the file should be there (if so, I will pull it from `feat/dynamic_dispatch`)

.circleci/config.yml
Antonio95: Thank you, fixed.

synthesizer/process/src/verify_execution.rs
Antonio95: Good observation, but the situation is different:

- A closure cannot output a static `Record`. Therefore, the moment we cast (locally minted) static to dynamic, we make sure the latter is not output.

- A function can output a static `Record`. So, at the point a cast-static-to-dynamic instruction is encountered, we don't know yet whether the corresponding static Record must be output. This is the case if and only if the dynamic record is passed to a callee or output. Checking that eagerly here, aside from being less efficient (as the call instructions would be explored many times) is essentially impossible, since things like closure calls can cause the dynamic registers to effectively remap to other registers. As an example: cast `R_static` -> `R_dynamic`; pass `R_dynamic` to a closure, which simply outputs it; receive it at a different register `R_dynamic'` and try to output `R_dynamic'`. This last bit means `R_static` must be output. So it's not even enough to, at the time the cast-static-to-dynamic is encountered, check whether the dynamic register `R_dynamic` is passed to a callee function or output, since the relation  between `R_dynamic'` and `R_static` is only learnt while exploring the closure.

Antonio95:
> Still reading it, but an immediate first thought: why not use the standard "set union" data structure (https://en.wikipedia.org/wiki/Disjoint-set_data_structure) which would make this a O(n) pass? It looks to be quadratic in the worst case atm? Because we are iterating over all the families to find the one to which a register belongs.

Are you referring to the object `record_families`? If so, which part(s) have the O(n^2) asymptotic? Once I understand your point, I'm happy to switch to a disjoint-union structure. I still want to review my own argument (I already checked at some point) that the record families are indeed disjoint.

rot256:
> Are you referring to the object `record_families`? If so, which part(s) have the O(n^2) asymptotic? Once I understand your point, I'm happy to switch to a disjoint-union structure. I still want to review my own argument (I already checked at some point) that the record families are indeed disjoint.

Ah, I missed that there is only at most one family for every input in the root transition.

.claude/CLAUDE.md
d0cd: Lets keep it as it is on `feat/dynamic-dispatch`

synthesizer/process/src/verify_execution.rs
```diff
@@ -778,4 +786,531 @@ impl<N: Network> Process<N> {
...
+    /// The root transition must be last.
+    ///
+    /// Input `call_graph`: A copy of the call graph (which will be modified in
+    /// place). It is assumed to contain all transitions in `transitions`. All
+    /// children of a given Transition ID must appear in the same order as the
+    /// corresponding calls happen in the function.
+    pub fn ensure_records_exist<'a>(
+        &self,
+        transitions: impl ExactSizeIterator<Item = &'a Transition<N>> + DoubleEndedIterator + Clone,
+        mut call_graph: HashMap<N::TransitionID, Vec<N::TransitionID>>,
```
d0cd: Is there a way we can avoid needing a mutable call graph in this algorithm?
This way you can save on the clones on invocation.

synthesizer/process/src/verify_execution/mod.rs
```diff
@@ -69,6 +69,14 @@ impl<N: Network> Process<N> {
         // Construct the call graph of the execution.
         let call_graph = self.construct_call_graph(execution.transitions())?;
 
+        // From ConsensusVersion::V14 onwards, ensure that, for each non-closure
+        // function in the execution, all DynamicRecords and ExternalRecords
+        // received as inputs or from callees exist on the ledger at the end of
+        // the execution (whether spent or not).
+        if consensus_version >= ConsensusVersion::V14 {
```
d0cd: Nit. The comment on this style of check should be consistent throughout the codebase.

synthesizer/process/src/verify_execution.rs
```diff
@@ -778,4 +786,531 @@ impl<N: Network> Process<N> {
...
+        // Keeps track of DynamicRecords and ExternalRecords received as inputs
+        // to the root transition. Whenever a dynamic- or external-record
+        // register is connected to one of them (for instance, because of
+        // caller-callee input/output relations in a call or because of a cast
+        // instruction), the latter is added to the family of the former.
+        // Whenever any record in a family is found out to correspond to a
+        // record on the ledger, the entire family is removed from the vector.
+        let mut register_families: Vec<IndexSet<(N::TransitionID, u64)>> = Vec::new();
+
+        let root_transition = transitions.clone().last().ok_or_else(|| anyhow!("Empty transition list"))?;
```
d0cd: Do we need the clones on L820 and L822?

synthesizer/process/src/verify_execution.rs
```diff
@@ -778,4 +786,531 @@ impl<N: Network> Process<N> {
...
+        &self,
+        // Vector of families of registers connected to `DynamicRecord`s or `ExternalRecord`s input to the root transition.
+        register_families: &mut Vec<IndexSet<(N::TransitionID, u64)>>,
+        // TransitionID of the transition being processed
+        transition_id: &N::TransitionID,
+        // `None` for the root transition. For non-root transitions, `Some` containing:
+        //  - TransitionID of the caller
+        //  - indices of the caller registers where it passes inputs to this transition's call (with `None` for inputs that are not registers)
+        //  - indices of the caller registers where it receives outputs from this transition's call
+        caller_info: Option<(N::TransitionID, Vec<Option<u64>>, Vec<u64>)>,
```
d0cd: Are the indices better organized as a vec of tuples or a map, to ensure type-level length consistency?

synthesizer/process/src/verify_execution.rs
```diff
@@ -778,4 +786,531 @@ impl<N: Network> Process<N> {
...
+        // Map from TransitionID to the corresponding Transition
+        tid_to_transition: &HashMap<N::TransitionID, &Transition<N>>,
+        // Call graph of the execution
+        call_graph: &mut HashMap<N::TransitionID, Vec<N::TransitionID>>,
+    ) -> Result<()> {
+        let transition = tid_to_transition
+            .get(transition_id)
+            .ok_or_else(|| anyhow!("Missing transition with ID {transition_id}"))?;
+        let stack = self.get_stack(transition.program_id())?;
+        let function = stack.get_function(transition.function_name())?;
```
d0cd: Try to use `get_function_ref` when you can.
`get_function` clones the `Function`.
It's worth seeing if this optimization can be applied throughout the codebase.

synthesizer/process/src/verify_execution.rs
```diff
@@ -778,4 +786,531 @@ impl<N: Network> Process<N> {
...
+                    if let Instruction::Call(call) = instruction
+                        && !call.is_function_call(stack.as_ref())?
+                    {
+                        // Closure case
+                        let closure = {
+                            let operator = call.operator();
+                            match operator {
+                                CallOperator::Resource(closure_identifier) => {
+                                    // Local closure call
+                                    stack.as_ref().program().get_closure(closure_identifier)?
```
rot256: Just a heads up `get_closure` deep clones the closure, which takes `O(NUMBER_INSTRUCTIONS_IN_THE_CLOSURE)` time for every call to the closure. So in total: `O(NUMBER_INSTRUCTIONS_IN_THE_CLOSURE * NUMBER_OF_CLOSURE_CALLS)` time.


However: because the compiler sees that you are not mutating the result, it actually does not do the clone (in release mode).

I will let you decide on whether to rely on the compiler being smart enough to avoid this quadratic blowup.

rot256: The alternative is returning a reference to the (immutable) closure.

rot256:
@Antonio95 

It seems there is a missing check when passing a static record minted in a function to a closure (as an external record): we are only checking the `locally_minted_dynamic` from the caller, map to see if the static record must be output. There are two sets track "minting" in `process_transition`:

- `locally_minted_static`: registers holding Records created by `cast ... as X.record`.
- `locally_minted_dynamic`: map from DynamicRecord registers back to the static Record they were cast from.

When `process_closure` encounters a closure call, there are two code paths:

- the *cast* path (closure casts an ExternalRecord to DynamicRecord, producing a new register) and
- the *passthrough* path (same callee register appears as both input and output): "// Detecting connections of caller registers resulting from closure input-output relations not involving casts"

The cast path (Case 3, line 1235) checks both maps:

```rust
if caller_locally_minted_static.contains(caller_operand_register) {
   // Case 3: Effectively casting performing a static-to-dynamic cast on the caller side. We update the caller's local-check tracking
   // accordingly. Note the input operand in the closure could still be an ExternalRecord if the call to the closure is external
   caller_locally_minted_dynamic
        .insert(*caller_destination_register, *caller_operand_register);
}
```

But the passthrough path (Case 6, line 1263) only checks `caller_locally_minted_dynamic`:

```rust
// Case 6. Caller local-check update
if let Some(original_static) = caller_locally_minted_dynamic.get(&caller_input_register) {
    caller_locally_minted_dynamic.insert(*caller_output_register, *original_static);
}
```

In particular, I think analysis of this code would behave wrong?:

```
closure relay.aleo/passthrough:
    input r0 as example.aleo/coin.record;
    output r0 as example.aleo/coin.record;

function example.aleo/exploit:
    // State:
    // locally_minted_static: {}
    // locally_minted_dynamic: {}
    // must_be_output: {}

    cast self.signer 100u64 into r0 as coin.record

    // State:
    // locally_minted_static: {r0}
    // locally_minted_dynamic: {}
    // must_be_output: {}

    call relay.aleo/passthrough r0 into r1

    // process_closure Case 6:
    // locally_minted_dynamic.get(r0) => None, so r1 is not tracked
    //
    // State:
    // locally_minted_static: {r0}
    // locally_minted_dynamic: {}
    // must_be_output: {}

    call consumer.aleo/consume r1

    // State:
    // locally_minted_static: {r0}
    // locally_minted_dynamic: {}
    // must_be_output: {}
```

After `process_closure`, `locally_minted_dynamic` is still empty.

The function call to `consume` with `r1` finds it in neither map?

Antonio95:
> @Antonio95
> 
> It seems there is a missing check when passing a static record minted in a function to a closure (as an external record): we are only checking the `locally_minted_dynamic` from the caller, map to see if the static record must be output. There are two sets track "minting" in `process_transition`:
> 
> * `locally_minted_static`: registers holding Records created by `cast ... as X.record`.
> * `locally_minted_dynamic`: map from DynamicRecord registers back to the static Record they were cast from.
> 
> When `process_closure` encounters a closure call, there are two code paths:
> 
> * the _cast_ path (closure casts an ExternalRecord to DynamicRecord, producing a new register) and
> * the _passthrough_ path (same callee register appears as both input and output): "// Detecting connections of caller registers resulting from closure input-output relations not involving casts"
> 
> The cast path (Case 3, line 1235) checks both maps:
> 
> ```rust
> if caller_locally_minted_static.contains(caller_operand_register) {
>    // Case 3: Effectively casting performing a static-to-dynamic cast on the caller side. We update the caller's local-check tracking
>    // accordingly. Note the input operand in the closure could still be an ExternalRecord if the call to the closure is external
>    caller_locally_minted_dynamic
>         .insert(*caller_destination_register, *caller_operand_register);
> }
> ```
> 
> But the passthrough path (Case 6, line 1263) only checks `caller_locally_minted_dynamic`:
> 
> ```rust
> // Case 6. Caller local-check update
> if let Some(original_static) = caller_locally_minted_dynamic.get(&caller_input_register) {
>     caller_locally_minted_dynamic.insert(*caller_output_register, *original_static);
> }
> ```
> 
> In particular, I think analysis of this code would behave wrong?:
> 
> ```
> closure relay.aleo/passthrough:
>     input r0 as example.aleo/coin.record;
>     output r0 as example.aleo/coin.record;
> 
> function example.aleo/exploit:
>     // State:
>     // locally_minted_static: {}
>     // locally_minted_dynamic: {}
>     // must_be_output: {}
> 
>     cast self.signer 100u64 into r0 as coin.record
> 
>     // State:
>     // locally_minted_static: {r0}
>     // locally_minted_dynamic: {}
>     // must_be_output: {}
> 
>     call relay.aleo/passthrough r0 into r1
> 
>     // process_closure Case 6:
>     // locally_minted_dynamic.get(r0) => None, so r1 is not tracked
>     //
>     // State:
>     // locally_minted_static: {r0}
>     // locally_minted_dynamic: {}
>     // must_be_output: {}
> 
>     call consumer.aleo/consume r1
> 
>     // State:
>     // locally_minted_static: {r0}
>     // locally_minted_dynamic: {}
>     // must_be_output: {}
> ```
> 
> After `process_closure`, `locally_minted_dynamic` is still empty.
> 
> The function call to `consume` with `r1` finds it in neither map?

@rot256 Well spotted, that's a situation I already considered. I even coded a version of the algorithm that tracked that: revision [14302ef1bcedf0d552b9664299cd1a06cc878efc](https://github.com/ProvableHQ/snarkVM/commit/14302ef1bcedf0d552b9664299cd1a06cc878efc), "added treatment for Record -> Record remapping in external-closure calls". The algorithm became a bit cumbersome because it could then be the case that a `DynamicRecord` "came from" several static `Record`s, if the original static `Record` in the cast instruction were remapped to another register via an external-closure call.

Fortunately, it turns out one cannot pass a locally minted static `Record` to an external-closure call, as that creates a dependency cycle involving records (I tested this).

.claude/CLAUDE.md
Antonio95: fixed 👌

synthesizer/process/src/verify_execution.rs
Antonio95: Good point, fixed in 756127de0b4142d550ee7fb0363f8e0ec77a480e: https://github.com/ProvableHQ/snarkVM/compare/756127de0b4142d550ee7fb0363f8e0ec77a480e~1..756127de0b4142d550ee7fb0363f8e0ec77a480e

synthesizer/process/src/verify_execution.rs
Antonio95: The second one wasn't needed indeed, I've removed it - thanks. The first one seems more necessary as all we get (from  `execution.transitions()`) is an iterator and calling `last()` consumes it. However, we're not cloning the transitions themselves, only the iterator over them - which I understand is a very lean object.

If you think it's worth getting rid of the first clone too, we can e. g. construct `tid_to_transition` as an `IndexMap` instead of a `HashMap` and get the last key; or pass the root tid as a separate argument to the function.

synthesizer/process/src/verify_execution.rs
Antonio95: Ah, I wasn't aware of that. Fixed it 👌 (there was only one call): I can't find an analogue for `get_closure` (which is called twice), is that right?

synthesizer/process/src/verify_execution.rs
Antonio95: @d0cd this is related to your other comment on `get_function_ref` vs. `get_function`. If there is no analogue for closure, we can either leave things as they are (from what @rot256 says, the compiler optimises this in release mode) or construct a map `(program_name, closure_name) -> closure` and replace the call to `get_closure` to an attempt to read from the map if available, fetching the closure and populating the map otherwise. If we choose the latter, we may want to consider sharing that map across the entire `ensure_records_exist` stack.

synthesizer/process/src/verify_execution.rs
Antonio95: When you say length consistency, which objects are you referring to? The `Vec<Option<u64>>` and  `Vec<u64>` in `caller_info` refer to call inputs and outputs and are therefore unrelated length-wise.

synthesizer/process/src/verify_execution/mod.rs
Antonio95: Thank you, fixed. I chose this version to appear in the other places.

synthesizer/process/src/verify_execution.rs
d0cd: It might not exist, feel free to add it!

synthesizer/process/src/verify_execution.rs
d0cd: Ah I see. It's only the caller registers, not a map between the caller and callee.

Antonio95:
Heads up: @rot256 wrote to me commenting on some aspects of the design which were obscured by the implementation. After a short discussion with him, I've refactored things slightly according to our ideas. This is done in commit 051fbe6b4f02929495e1a2a7099a16156515ee1b. The changes are as follows:
- The method `ensure_records_exist` and its auxiliary functions and structures have been moved to a separate file inside the `verify_execution` module folder.
- Three simple structures have been created: `GlobalCheck`, `FunctionLocalCheck` and `ClosureLocalCheck`. These simply group together some maps and sets from the original implementation and make the code a bit more semantic by giving (method) names to the operations on those objects (for instance, `process_function` does not receive `FunctionLocalCheck` and `ClosureLocalCheck` arguments; and `process_closure` receives `caller_local_check: &mut FunctionLocalCheck`). However, no functional changes have been made to the algorithm*.

*There is one functional change independent from the above redesign: the addition of the `Input::Record` case when processing inputs, which detects `ExternalRecord (at caller) -> Record (at callee)` family materialisation. This got lost when I updated the code to account for the new input variants `RecordWithDynamicID` and `ExternalRecordWithDynamicID`. I've also added test case 7 in `record_existence.rs` with three subcases that check this functionality under various remapping/translations.

synthesizer/process/src/verify_execution.rs
Antonio95: Addressed in d41637278b81e7ded288957009e422e1a3d8bcce, which adds a `get_closure_ref` analogue of `get_function_ref`

synthesizer/process/src/verify_execution.rs
Antonio95: Addressed in d41637278b81e7ded288957009e422e1a3d8bcce, which adds a `get_closure_ref` analogue of `get_function_ref`. Due to the specifics of the borrow-checker, one now needs to call `process_closure(...)` with the same arguments once inside each of the two arms of the `match call.operator()` arm: if one simply tries to `let closure = match call.operator()` and then call `process_closure(...)` outside of it, the fact that now the closure is only a reference means the `stack.get_external_stack(...)` in the external-closure case would be dropped before the `closure` which points to it is used.

synthesizer/process/src/verify_execution/ensure_records_exist.rs
```diff
@@ -0,0 +1,647 @@
...
+// the check that, for each DynamicRecord cast from a locally minted static
+// Record and passed to a non-closure call or output, the static Record is
+// output.
+struct FunctionLocalCheck {
+    // Registers of static Records minted in this transition.
+    locally_minted_static: HashSet<u64>,
+    // For each DynamicRecord at r_j cast from a locally minted static Record at
+    // r_i, this map contains the entry r_j -> r_i. This also takes into account
+    // remaps of DynamicRecord caused by external-to-dynamic casts and
+    // dynamic-to-static casts occurring in closures.
```
d0cd: Do you mean static-to-dynamic?

synthesizer/process/src/verify_execution/ensure_records_exist.rs
```diff
@@ -0,0 +1,647 @@
...
+        for register in self.must_be_output.iter() {
+            if !actually_output.contains(register) {
+                return Some(*register);
+            }
+        }
+        None
+    }
+}
+
+// Auxiliary structure for the local record-existence check of a closure, i. e.
```
d0cd:
```suggestion
// Auxiliary structure for the local record-existence check of a closure, i.e.
```

synthesizer/process/src/verify_execution/ensure_records_exist.rs
```diff
@@ -0,0 +1,647 @@
...
+            izip!(caller_input_registers, input_registers, inputs)
+        {
+            if let Some(caller_input_register) = caller_input_register_opt {
+                match callee_input {
+                    Input::RecordWithDynamicID(..) | Input::Record(..) => {
+                        // Case 1: consuming a Record translating from a DynamicRecord or passed as an ExternalRecord
+                        // by the caller: the family is now known to exist on the ledger.
+                        global_check.mark_existing((*caller_tid, *caller_input_register));
+                    }
+                    // TODO (Antonio) add External -> Static case
```
d0cd: Flagging TODO.

synthesizer/process/src/verify_execution/ensure_records_exist.rs
```diff
@@ -0,0 +1,647 @@
...
+                        global_check.add_to_family(old_register, new_register);
+                    }
+                    _ => {}
+                }
+            }
+        }
+    } else {
+        // Root transition case
+
+        ensure!(
```
d0cd: This check could probably be factored out of both branches.

synthesizer/process/src/verify_execution/ensure_records_exist.rs
```diff
@@ -0,0 +1,647 @@
...
+            input_registers.len() == transition.inputs().len(),
+            "Mismatch in the number of inputs and registers in the root call"
+        );
+
+        // Initialise record_families with one (singleton) family per DynamicRecord and ExternalRecord input
```
d0cd:
```suggestion
        // Initialise the record families with one (singleton) family per DynamicRecord and ExternalRecord input
```

synthesizer/process/src/verify_execution/ensure_records_exist.rs
```diff
@@ -0,0 +1,647 @@
...
+        for (input, register) in transition.inputs().iter().zip(input_registers.iter()) {
+            if matches!(input, Input::DynamicRecord(..) | Input::ExternalRecord(..)) {
+                global_check.add_family((*transition_id, *register));
+            }
+        }
+
+        // Note: even if record_families is empty, we need to process the transitions to enforce the local check
```
d0cd: `record_families` doesn't exist as a variable anymore.

synthesizer/process/src/verify_execution/ensure_records_exist.rs
```diff
@@ -0,0 +1,647 @@
...
+                        local_check.cast_to_dynamic(operand_register, destination_register);
+                    }
+                    _ => {}
+                }
+            }
+            Instruction::Call(..) | Instruction::CallDynamic(..) => {
+                let caller_input_operands = if matches!(instruction, Instruction::Call(..)) {
+                    instruction.operands()
+                } else {
+                    &instruction.operands()[3..]
```
d0cd: Match more explicitly and explain where the 3 comes from and why it is safe.

synthesizer/process/src/verify_execution/ensure_records_exist.rs
```diff
@@ -0,0 +1,647 @@
...
+                                &caller_input_registers,
+                                &caller_output_registers,
+                            )?;
+                        }
+                        CallOperator::Locator(external_locator) => {
+                            // External closure call
+                            let closure_stack = stack.get_external_stack(external_locator.program_id())?;
+                            let closure = closure_stack.program().get_closure_ref(external_locator.resource())?;
+
+                            process_closure(
```
d0cd: You could probably consolidate the two branches.

synthesizer/process/src/verify_execution/ensure_records_exist.rs
```diff
@@ -0,0 +1,647 @@
...
+                                &caller_input_registers,
+                                &caller_output_registers,
+                            )?;
+                        }
+                    }
+                } else {
+                    // Function case
+
+                    let tid_callee = if let Some(tid) =
+                        call_graph.get(transition_id).unwrap().get(processed_function_calls)
```
d0cd: Make note on unwrap safety. This should be done for all unwraps.

synthesizer/process/src/verify_execution/ensure_records_exist.rs
```diff
@@ -0,0 +1,647 @@
...
+                    )?;
+                }
+            }
+            _ => {}
+        }
+    }
+
+    // Output processing
+
+    // In a first pass over the outputs, track which DynamicRecords coming from locally minted static records are directly output.
```
d0cd: Can the two passes be consolidated?

synthesizer/process/src/verify_execution/ensure_records_exist.rs
```diff
@@ -0,0 +1,647 @@
...
+    function.outputs().iter().for_each(|output| {
+        if let Operand::Register(output_register) = output.operand() {
+            if let Some(static_record) = local_check.get_static_source_of_dynamic(output_register.locator()) {
+                // TODO (Antonio) add test case
```
d0cd: Flagging TODO.

synthesizer/process/src/verify_execution/ensure_records_exist.rs
```diff
@@ -0,0 +1,647 @@
...
+        "Mismatch between the number of input registers in the call instruction and the callee's input types in call to closure {closure_name}"
+    );
+    ensure!(
+        caller_output_registers.len() == closure.outputs().len()
+            && caller_output_registers.len() == closure.output_types().len(),
+        "Mismatch between the number of output registers in the call instruction and the callee's output types in call to closure {closure_name}"
+    );
+
+    // Construct a map { callee register -> caller register } for the closure's inputs of type Record, DynamicRecord or ExternalRecord
+    let input_map = izip!(caller_input_registers, closure.inputs())
```
d0cd: Use zip_eq where possible.

synthesizer/process/src/verify_execution/ensure_records_exist.rs
```diff
@@ -0,0 +1,647 @@
...
+            }
+        })
+        .collect::<Result<HashMap<u64, u64>>>()?;
+
+    for instruction in closure.instructions() {
+        // Only cast instructions need to be processed at this stage.
+        if let Instruction::Cast(cast) = instruction {
+            match cast.cast_type() {
+                CastType::Record(..) => {
+                    // Case 1: minting a static Record locally. We keep track to ensure DynamicRecords cast from it are not output.
```
d0cd: In general, if we are number cases, they should be consolidated in documentation somewhere.

synthesizer/src/vm/tests/test_v14/record_existence.rs
```diff
@@ -0,0 +1,1041 @@
...
+fn test_existence_check() {
+    let rng = &mut TestRng::default();
+
+    let caller_private_key = sample_genesis_private_key(rng);
+    let caller_address = Address::try_from(&caller_private_key).unwrap();
+    let caller_view_key = ViewKey::<CurrentNetwork>::try_from(&caller_private_key).unwrap();
+
+    let network_field = Identifier::<CurrentNetwork>::from_str("aleo").unwrap().to_field().unwrap();
+    let program_base_field = Identifier::<CurrentNetwork>::from_str("base").unwrap().to_field().unwrap();
+    let decomission_function_field = Identifier::<CurrentNetwork>::from_str("decomission").unwrap().to_field().unwrap();
```
d0cd:
```suggestion
    let decommission_function_field = Identifier::<CurrentNetwork>::from_str("decomission").unwrap().to_field().unwrap();
```

synthesizer/src/vm/tests/test_v14/record_existence.rs
```diff
@@ -0,0 +1,1041 @@
...
+            input r0 as boolean.private;
+            
+            cast self.signer 1u16 2u16 r0 into r1 as map.record;
+
+            // Here we are effectively performing a static-to-dynamic cast, but through a closure (where it is in fact external-to-dynamic)
+            call remapper.aleo/remap r1 into r2;
+
+            call extension.aleo/dynamic_pass_through_function 3u64 r2 into r3 r4;
+
+            // This saves the call above from breaking the locak check
```
d0cd:
```suggestion
            // This saves the call above from breaking the local check
```

synthesizer/src/vm/tests/test_v14/record_existence.rs
```diff
@@ -0,0 +1,1041 @@
...
+
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+use super::*;
+
+// Test the record-existance check passes or fails as expected in a number of
```
d0cd:
```suggestion
// Test the record-existence check passes or fails as expected in a number of
```

synthesizer/process/src/verify_execution/ensure_records_exist.rs
```diff
@@ -0,0 +1,647 @@
...
+            self.families.iter().filter(|family| family.contains(&record)).count() <= 1,
+            "Multiple families contain register {} for transition ID {}",
+            record.1,
+            record.0
+        );
+
+        let family = self.families.iter().position(|family| family.contains(&record));
+
+        if let Some(family_index) = family {
+            self.families.remove(family_index);
```
d0cd: Can we safely swap remove?

synthesizer/process/src/verify_execution/ensure_records_exist.rs
```diff
@@ -0,0 +1,647 @@
...
+    caller_tid: &N::TransitionID,
+    // Closure being processed
+    closure: &ClosureCore<N>,
+    // Global check keeping track of families of registers connected to
+    // `DynamicRecord`s or `ExternalRecord`s input to the root transition.
+    global_check: &mut GlobalCheck<N>,
+    // Caller function's local check.
+    caller_local_check: &mut FunctionLocalCheck,
+    // Caller registers of inputs to the closure call (`None` for inputs that are not registers)
+    caller_input_registers: &Vec<Option<u64>>,
```
d0cd: You can probably turn this and the lower member into a slice.

synthesizer/process/src/verify_execution/ensure_records_exist.rs
```diff
@@ -0,0 +1,647 @@
...
+
+    // Returns whether a given register corresponds to a locally minted static Record.
+    fn is_locally_minted_static(&self, register: u64) -> bool {
+        self.locally_minted_static.contains(&register)
+    }
+
+    // Returns the register of the static Record from which a given
+    // DynamicRecord is cast, if any.
+    fn get_static_source_of_dynamic(&self, register: u64) -> Option<u64> {
+        self.locally_minted_dynamic.get(&register).cloned()
```
d0cd: You should be able to copy.

synthesizer/src/vm/tests/test_v14/record_existence.rs
```diff
@@ -0,0 +1,1041 @@
...
+        closure dynamic_mint_closure:
+            input r0 as u8;
+            input r1 as boolean;
+
+            cast {caller_address} r0 r1 into r2 as rover.record;
+            cast r2 into r3 as dynamic.record;
+
+            output r3 as dynamic.record;
+
+        function call_closure_mint:
```
d0cd: Do we need to call this in a test?

synthesizer/src/vm/tests/test_v14/record_existence.rs
```diff
@@ -0,0 +1,1041 @@
...
+            call dummy 10u16 false into r3 r4 r5;
+            call dynamic_pass_through_closure 10u64 r2 into r6 r7;
+
+            call.dynamic {program_base_field} {network_field} {decomission_function_field}
+                with r1 r6 (as dynamic.record dynamic.record)
+                into r8 (as boolean.public);
+
+            output r8 as boolean.public;
+
+        function mint_and_decomission_same:
```
d0cd: Do we need to call this in a test?

d0cd (changes requested):
Left some feedback. 
At a high-level the algorithm is solid. 
That said, I've found it a bit challenging to follow the code. 
It would be helpful to have a **concise** spec at the top of the file to guide readers.

We should ask ourselves how easy will this be for others to maintain.

synthesizer/process/src/verify_execution/ensure_records_exist.rs
Antonio95: Indeed, thank you. Fixed.

synthesizer/process/src/verify_execution/ensure_records_exist.rs
Antonio95: (Learnt and) fixed

synthesizer/process/src/verify_execution/ensure_records_exist.rs
Antonio95: Thank you, removed. The TODO itself had already been addressed and tested (test 7)

synthesizer/process/src/verify_execution/ensure_records_exist.rs
Antonio95: I only put it in the root-transition case because only there am I zipping the two iterators, but I guess such a cheap check can be done in all cases as a sanity check. Factored it out and replaced the "in the root call" by "
"in the call to {locator}"

synthesizer/process/src/verify_execution/ensure_records_exist.rs
Antonio95: Fixed 👌

synthesizer/process/src/verify_execution/ensure_records_exist.rs
Antonio95: Fixed 👌

synthesizer/process/src/verify_execution/ensure_records_exist.rs
Antonio95: Addressed

synthesizer/process/src/verify_execution/ensure_records_exist.rs
Antonio95: Initially there were branches only to get the closure description, but when I introduced `get_closure_ref` things broke because of the borrow checker (in the external-closure case, I was getting a pointer to a closure from a stack which itself was a reference; so I got the "creates a temporary value which is dropped..." error on `get_external_stack`; I even tried to hold the stack alongside the closure, but this didn't fix it because it is dropped anyway).

Unless you see some easy fix, what do you prefer? The current version which duplicates the call to `process_closure`, or the original one which called `get_closure` and owned the `ClosureCore`?

synthesizer/process/src/verify_execution/ensure_records_exist.rs
Antonio95: Done for all three `unwrap`s. In this one and the other instance of the same, I commented "The unwrap is safe (assuming a correct call graph) as we are processing a Transition which is part of the execution.". Let me know if you think the wording needs changes.

synthesizer/process/src/verify_execution/ensure_records_exist.rs
Antonio95: Yes, done 👌 I also slightly improved performance by only keeping track of locally minted, static-`Record` registers in `actually_output` (renamed to `locally_minted_output`).

synthesizer/process/src/verify_execution/ensure_records_exist.rs
Antonio95: Thank you, this was a leftover TODO. It is tested in many of the current test cases, e. g. test case 4.

synthesizer/process/src/verify_execution/ensure_records_exist.rs
Antonio95: Replaced all `zip`s and `izip`s by `zip_eq`s.

synthesizer/src/vm/tests/test_v14/record_existence.rs
Antonio95: Fixed

synthesizer/src/vm/tests/test_v14/record_existence.rs
Antonio95: Fixed all 20 misspellings of the word

synthesizer/src/vm/tests/test_v14/record_existence.rs
Antonio95: Fixed

synthesizer/process/src/verify_execution/ensure_records_exist.rs
Antonio95: Indeed, well spotted. I wasn't aware of that method, than you 👌 Done

synthesizer/process/src/verify_execution/ensure_records_exist.rs
Antonio95: Fixed, thank you 👌

synthesizer/process/src/verify_execution/ensure_records_exist.rs
Antonio95: The issue is that `self.locally_minted_dynamic.get` returns `Option<&u64>`. If `cloned` is problematic, I can return the `Option<&u64>` directly and dereference the `&u64` at the call site.

synthesizer/src/vm/tests/test_v14/record_existence.rs
Antonio95: No, removed.

synthesizer/src/vm/tests/test_v14/record_existence.rs
Antonio95: No, removed

synthesizer/process/src/verify_execution/ensure_records_exist.rs
d0cd: Can you use `.copied()`?

synthesizer/process/src/verify_execution/ensure_records_exist.rs
d0cd: Could something like this work? Might need some tweaks.
```
let (program_stack, identifier) = match call.operator() {
    CallOperator::Resource(identifier) => (stack, identifier),
    CallOperator::Locator(external_locator) => (
        stack.get_external_stack(external_locator.program_id())?,
        external_locator.resource(),
    ),
};
let closure = program_stack.program().get_closure_ref(identifier)?;
process_closure(
    transition_id,
    closure,
    global_check,
    &mut local_check,
    &caller_input_registers,
    &caller_output_registers,
)?;
```

synthesizer/process/src/verify_execution/ensure_records_exist.rs
```diff
@@ -0,0 +1,655 @@
...
+
+        let tid_to_transition: HashMap<N::TransitionID, &Transition<N>> =
+            transitions.map(|transition| (*transition.id(), transition)).collect();
+
+        let mut global_check = GlobalCheck::new();
+
+        // Recursively explore the execution, keeping track of record connections across the relevant casts and calls.
+        process_transition(self, &mut global_check, root_transition.id(), None, &tid_to_transition, call_graph)?;
+
+        // Ensure the local check passes.
```
d0cd:
```suggestion
        // Ensure the global check passes.
```

synthesizer/process/src/verify_execution/ensure_records_exist.rs
```diff
@@ -0,0 +1,655 @@
...
+}
+
+// Auxiliary function for `ensure_records_exist` that connects registers in the transition to the relevant record
+// families, tracking connections and marking families as existing if they are found to correspond to a record on the
+// ledger (global check). Furthermore, it also checks that, if a DynamicRecord cast from a locally minted static Record
+// is output or passed to a callee, the static one is output (local check).
+fn process_transition<N: Network>(
+    // Process being checked, from which function and closure definitions are retrieved
+    process: &Process<N>,
+    // Global check keeping track of families of registers connected to DynamicRecords or ExternalRecorda input to the
```
d0cd:
```suggestion
    // Global check keeping track of families of registers connected to DynamicRecords or ExternalRecords input to the
```

synthesizer/process/src/verify_execution/ensure_records_exist.rs
Antonio95: Ah, that almost works. One needs to clone `stack` (because `stack.get_external_stack()` returns something that's fully owned), but the former is `Arc`, so cloning is essentially free. Thank you 👌

synthesizer/process/src/verify_execution/ensure_records_exist.rs
Antonio95: Addressed in the latest commit

synthesizer/process/src/verify_execution/ensure_records_exist.rs
Antonio95: Done 👌 Is there any reason to prefer it to `cloned` here?

synthesizer/process/src/verify_execution/ensure_records_exist.rs
Antonio95: Fixed

synthesizer/process/src/verify_execution/ensure_records_exist.rs
Antonio95: Fixed, thanks

d0cd:
> Fortunately, it turns out one cannot pass a locally minted static `Record` to an external-closure call, as that creates a dependency cycle involving records (I tested this).

@Antonio95 I'm not sure this is correct. It should be possible to:
  1. Deploy `alpha.aleo` (`Token` record + mint function)
  2. Deploy `beta.aleo` (import `alpha.aleo`, closure `inspect_token` taking `alpha.aleo/Token.record`)
  3. Upgrade `alpha.aleo` to import `beta.aleo`, adding `mint_and_inspect` that mints a `Token` locally and calls `beta.aleo/inspect_token`

synthesizer/process/src/verify_execution/ensure_records_exist.rs
d0cd: They do the same thing, but `copied` is clearer for the reader.

Antonio95:
@rot256 (CC @d0cd) It is possible the `Record (at caller) <-> ExternalRecord (at callee)` boundary case is meant to be allowed (both at input and output boundaries) and the tests I ran which seemed to indicate otherwise actually had that behaviour because of an unrelated bug.

While we confirm, I have preemptively updated the record-existence check to handle that case: https://github.com/ProvableHQ/snarkVM/pull/3173/commits/51e70e13a081a9ab569eb34569f4c942e15092a8. The way I see it, this new information affects the previous design in the following ways (note: as far as I can tell, only the local check is affected)
- A static `Record` can now be passed to a function and not yet be known to materialise - because the function can receive it as an `ExternalRecord`. Therefore, passing a locally-minted static `Record` to a function now means we also add it to the list of static `Record`s that must be output.
- A static `Record` can now be effectively remapped to a new register within a single function's scope (including closures) via external-closure calls (this was never the case before because static `Records` cannot be output by closures). When this happens, any requirement that the original register were output must be relaxed to the requirement that any static-`Record` register connected to the original one is output. This relaxation must already be applied to the first point above. Note also that a single static-`Record` register could go through many such remappings in the scope of a single function and even "multiply" (if it is passed to several external-closure calls).
- Because of the above relaxation, a `DynamicRecord` can now effectively come from several static `Record`s.

I think the new design achieves that with relatively minimal changes. What do you think?

Antonio95:
@rot256 @d0cd The following changes to the specification have been made since the last version:

1. It has been confirmed the following patterns should be allowed:  call inputs regarded as a static `Record`s by the caller and `ExternalRecord`s by the callee; and call outputs regarded as static `Record`s by the caller and `ExternalRecord`s by the callee. This has the following implications:
    - When a locally minted static `Record` is input to a function call, it must be output by the caller as well. This is to guarantee the `ExternalRecord` received by the callee is known to materialise (note it can never be received as a static `Record`: this would constitute either a local static call or a dynamic call with a static-Record input).
2. It has been decided closures cannot output `DynamicRecord`s or `ExternaRecord`s (in addition to `Record`s, which they were already prevented from outputting). This means
    - The entire handling of closures (`ClosureLocalCheck` and `process_closure`) is gone, as closures can no longer affect the global check or the caller’s local check.
    - Static `Record` registers cannot be effectively remapped in the scope of a single function (as was the case before due to external-closure calls).
    - `DynamicRecord` registers cannot be effectively remapped either, and cast-to-dynamic instructions in closures are irrelevant (as the destinations cannot be output). This means those two situations no longer need to keep track of in a function’s local check.
3. The (function) local-check requirement has been made it actually *local*. More specifically, if a locally minted static `Record` `r0` is passed as an input to a function, then `r0` itself must be output. If, say, an external-function call effectively remaps `r0` to a static-`Record` register `r1` (which is now possible due to point 1 above, say, with an external function acting as an `ExternalRecord` pass-through), we do not consider outputting `r1` to be sufficient to materialise r0. In particular, we do not keep track of the relation between `r0` and `r1` (which would involve exploring all callees and their register flow as part of a function’s “local” check).
    
    This, together with 1, means that, upon encountering a record-like input to a function, it can be decided on the spot whether that breaks the local check (this was the case in process_closure before, but not in process_function, due to register remappings; and was the design Mathias favoured).
    
The description in the `updated ensure_records_exist.rs` reflects the most up-to-date state of assumptions. Tests have been reworked to fit the new specification. The new `Record` (at caller) → `ExternalRecord` (at callee) case has been tested (@d0cd  cf. test 2; In this example, I could get it to work. So I'm not sure what is different in the one I sent you the other day - although that should be more or less clear now you've tracked down the issue).

synthesizer/process/src/verify_execution/expl.txt
d0cd, line 1: Is this needed?

synthesizer/process/src/verify_execution/ensure_records_exist.rs
```diff
@@ -0,0 +1,515 @@
...
+//         - Case 6b: DynamicRecords cast from locally minted static Records and passed as inputs, in which case the
+//           static ones must be output by the caller (local check)
+//     - Case 7: DynamicRecords cast from locally minted static Records and output (local check)
+//     - Case 8: Connections between the caller registers where call outputs are received and call and the callee's
+//       output registers (global check).
+
+// Structure that keeps track of the global record-existence check of an execution, i.e. that each DynamicRecord and
+// ExternalRecord input to the root transition corresponds to a static Record that exists on the ledger at the end of
+// the execution.
+struct GlobalCheck<N: Network> {
```
raychu86, line 113: While it's nice that the documentation is extensive, IMO it is also hard to follow. 

This could be more concise and/or expressed as a list of rules to improve developer understanding. Program developers care about what they can/can't do rather than the justification

synthesizer/process/src/verify_execution/ensure_records_exist.rs
```diff
@@ -0,0 +1,515 @@
...
+//
+//   It can be shown that materialization of a non-static record input to the root transition cannot occur due to a
+//   connected static Record being output (by a function in the program containing the record definition) - at least,
+//   not without first materializing it by consuming it as a static-Record input.
+//
+//   2. Ensuring that, in each transition t in the execution, if a static Record R_s is
+//   minted locally and either
+//    - passed to a callee.
+//    - cast to a DynamicRecord R_d which is then passed to a callee or output.
+//   then R_d is itself output. We refer to this as the *local check*. Note that, in case b), if the callee
```
raychu86: Why does the `R_d` need to be output?

Previously in our `ExternalRecord` model, we don't require outputting the external record because it doesn't necessarily serve the developer. It also takes up one of the 16 outputs in the transition, without necessarily being useful.

raychu86:
> When a locally minted static Record is input to a function call, it must be output by the caller as well. This is to guarantee the ExternalRecord received by the callee is known to materialise (note it can never be received as a static Record: this would constitute either a local static call or a dynamic call with a static-Record input).


This seems like it could break a lot of the existing programs on-chain that do not output the external record. For instance a `multi_transfer` function that calls `transfer_private` many times internally.

synthesizer/process/src/verify_execution/ensure_records_exist.rs
Antonio95: Ah, that was a typo, thank you: it should be the static record `R_s` it was minted from that needs to be output. I have fixed it

In case your second part of the comment still applies after that fix/redefinition: no, the `ExternalRecord` received by the callee doesn't need to be output (by the callee). It's only the static `Record` minted by the caller and passed to the callee that then needs to be output (by the caller). Otherwise the guarantee (G) this PR is predicated on wouldn't be satisfied (with `t` being the callee):

> (Guarantee) Every non-static record (DynamicRecord or ExternalRecord) input to t or received by t from a function callee corresponds to a static Record which exists on the ledger at the end of the full execution (whether consumed or not).

Does that make sense?

synthesizer/process/src/verify_execution/expl.txt
Antonio95: No, fixed (must have ctrl-S'd by mistake)

synthesizer/process/src/verify_execution/ensure_records_exist.rs
Antonio95: I appreciat the feedback. There is a lot more documentation than what is included in the file and choosing how much/what to include has been a tricky balancing act.

The situation is as follows: the local check is a very concrete rule that developers should have no problem understanding and abiding by:

> If a function `f` mints a static Record `R_s` and either { passes it to a function call } or { casts it to a `DynamicRecord` `R_d` which it then outputs or passes to a function call  } (or both {}s) , then it must also output `R_d`.

A nice property of this restriction is that it's purely local and in particular cannot be broken by updates to other programs (e. g. the callees of `f`). The developer of `f` has full control over it. This is part of the latest changes we introduced (which, incidentally, now means we can introduce this checks at deployment time in the future if we think it'd be helpful; this wasn't the case with some of the our previous versions).

The global check is different. It refers to what happens to non-static record (`DynamicRecord`/`ExternalRecord`) inputs to the root transition of an execution. In particular, developers have little control over it. Instead, the burden is placed on the party triggering the execution. They should simply ensure that, IF they are supplying `DynamicRecord`s/`ExternalRecord`s to the root call, the path their execution is going to follow ensures the underlying static `Record`s are consumed at some point.

For instance, a developer is allowed to write functions such as:
```
function read_atmosphere_dynamic:
    input r0 as dynamic.record;

    get.dynamic.record r0.atmosphere into r1 as u16.private;

    output r1 as u16.private;
```
(or analogously for `ExternalRecord` inputs). This doesn't break the record-existence per se. However, if a party tries to execute that as the root call (and therefore only call of the transaction), the check will disallow it.

I think things can be summarised as follows: the record-existence check will pass if and only if the following two conditions hold:
 - The developers who wrote the functions appearing in the execution followed the local-check constraint.
 - The party authorising the root call (who can see the entire call graph which will result from it) makes sure the  any `DynamicRecord`s and `ExternalRecord`s they are passing to that root call are actually consumed (as static `Record`s) at some point in the execution. The worst that can happen (in terms of the global check) is that the party who tried to authorise the transaction receives an error of the form `Non-static record input at r1 of the root function milky_way/rover.aleo is not known to correspond to a record on the ledger`.

I agree this should be communicated clearly to the relevant groups.

CC @d0cd @vicsn (I think the above way to look at the situation, especially in terms of the programmer/authoriser split, is useful)

Antonio95:
> > When a locally minted static Record is input to a function call, it must be output by the caller as well. This is to guarantee the ExternalRecord received by the callee is known to materialise (note it can never be received as a static Record: this would constitute either a local static call or a dynamic call with a static-Record input).
> 
> This seems like it could break a lot of the existing programs on-chain that do not output the external record. For instance a `multi_transfer` function that calls `transfer_private` many times internally.

@raychu86 I think there might be a misunderstanding (in which case it's important we figure out what part of the documentation makes is sound like a different restriction than what is actually the case). Let's keep the definition of the local check in mind:

> If a function `f` mints a static Record `R_s` and either { passes it to a function call } or { casts it to a `DynamicRecord` `R_d` which it then outputs or passes to a function call  } (or both {}s) , then it must also output `R_d`.

In particular, a function `g` that receives an `ExternalRecord` at `rX` has no responsibilities towards it (because it is not a locally (meaning, in `g`) minted static `Record`), and is free to output it or pass it to its internal function calls. In a way (and this can be made precise with the formal analysis of the algorithm), the fact that all ancestors of `g` satisfy the contract `C` (that is, the local check) "protects" `g` and lets it rest assured that it is actually operating on an material `ExternalRecord`, i. e. one corresponding to a static `Record` on the ledger. For instance, if the `rX` received by `g` was minted as a static `Record`  inside `g`'s caller `f`, the local check on `f` ensures that `f` is actually outputting that `Record`, therefore materialising it.

As far as breaking current programs, I think Pranav ran a check and confirmed this wasn't the case.

In terms of the example you mentioned, I'm not sure what you specifically have in mind. Supposing it is something like (thanks to Victor for the hint)
```
function dex:
    input r0 as ext_program.aleo/some.record
    call transfer_private r0 into r1 r2
    call transfer_private r1 into r3 r4
    call transfer_private r2 into r5 r6
```
that sort of pattern is completely fine. This function could even be called as the root call of an execution, because the global check would also be satisfied (the `ExternalRecord` received by the root call at `r0` is indeed consumed in the course of the execution - specifically, in the first instruction).

synthesizer/process/src/verify_execution/ensure_records_exist.rs
raychu86: Yes that makes sense. 

You could in theory craft a record and perform a call with that crafted record without explicitly outputting it.

raychu86:
> that sort of pattern is completely fine. This function could even be called as the root call of an execution, because the global check would also be satisfied (the ExternalRecord received by the root call at r0 is indeed consumed in the course of the execution - specifically, in the first instruction).

@Antonio95 I misinterpreted the rule earlier and my above statement was incorrect. I thought it was to require outputting the external records like you defined in:
```
function dex:
    input r0 as ext_program.aleo/some.record
    call transfer_private r0 into r1 r2
    call transfer_private r1 into r3 r4
    call transfer_private r2 into r5 r6
 ```
 
But the actual rule is that if an ExternalRecord or DynamicRecord is used, then a corresponding static record must either be an input or output. This is fine and I'm glad you guys did a check of this for existing programs.

**Comment by rot256** [comment]
@Antonio95 

I don't understand:

> Because of the above relaxation,
> a DynamicRecord can now effectively come from several static Records.

You mean that multiple DynamicRecords can share a single underlying static Record?
Was that not already the case?

> This, together with 1, means that, upon encountering a record-like input to a function,
> it can be decided on the spot whether that breaks the local check.

Nice, that's much simpler.

File: synthesizer/process/src/verify_execution/ensure_records_exist.rs [hunk 35]
```diff
@@ -0,0 +1,548 @@
...
+        if let Some(family_index) = family {
+            self.families.swap_remove(family_index);
+        }
+    }
+
+    // Returns the families of registers that are not known to correspond to static Records on the ledger. This is only
+    // called once in the entire existence check, at the very end.
+    fn non_existent_families(&self) -> &[IndexSet<(N::TransitionID, u64)>] {
+        &self.families
+    }
```
rot256: Nit: maybe just a `.validate() -> Result<()>`

synthesizer/process/src/verify_execution/ensure_records_exist.rs
```diff
@@ -0,0 +1,548 @@
...
+            && !self.output_registers.contains(static_record)
+        {
+            // Case 6b)
+            bail!(
+                "In {}, DynamicRecord at r{register} passed to a function call is cast from a locally minted Record at r{static_record} which is not output",
+                self.locator
+            );
+        } else {
+            Ok(())
+        }
```
rot256, line 231: Nit: no need for `else` and `else if`:

```rust
if self.locally_minted_static.contains(&register) && !self.output_registers.contains(&register) {
    // Case 6a)
    bail!(
        "In {}, locally minted Record at r{register} is passed to a function call but not output",
        self.locator
    );
}
if let Some(static_record) = self.locally_minted_dynamic.get(&register)
    && !self.output_registers.contains(static_record)
{
    // Case 6b)
    bail!(
        "In {}, DynamicRecord at r{register} passed to a function call is cast from a locally minted Record at r{static_record} which is not output",
        self.locator
    );
}
Ok(())
```

synthesizer/process/src/verify_execution/ensure_records_exist.rs
```diff
@@ -0,0 +1,548 @@
...
+            && !self.output_registers.contains(static_record)
+        {
+            // Case 7)
+            bail!(
+                "In {}, output DynamicRecord at r{register} is cast from a locally minted Record at r{static_record} which is not output",
+                self.locator
+            );
+        } else {
+            Ok(())
+        }
```
rot256: Nit: don't need the else

synthesizer/process/src/verify_execution/ensure_records_exist.rs
```diff
@@ -0,0 +1,548 @@
...
+    } else {
+        // Root transition case.
+
+        // Initialise the record families with one (singleton) family per DynamicRecord and ExternalRecord input.
+        // Length check performed above.
+        for (input, register) in transition.inputs().iter().zip_eq(input_registers.iter()) {
+            if matches!(input, Input::DynamicRecord(..) | Input::ExternalRecord(..)) {
+                global_check.add_family((*transition_id, *register));
+            }
```
rot256: Nit: I think it is more readable if you explicitly populate the families upon creation of the `global_check` in `ensure_records_exist`. Instead of having one case in the recursion which is special and which we must understand to occur first for the rest of the analysis to work out.

synthesizer/process/src/verify_execution/ensure_records_exist.rs
```diff
@@ -0,0 +1,548 @@
...
+                        // static Record.
+                        local_check.cast_to_dynamic(operand_register, destination_register);
+                    }
+                    _ => {}
+                }
+            }
+            Instruction::Call(..) | Instruction::CallDynamic(..) => {
+                if let Instruction::Call(call) = instruction
+                    && !call.is_function_call(stack.as_ref())?
+                {
```
rot256: Tiny nit: I would prefer combining it with the outer match here and avoid the nested condition:

```rust
match instruction {
    ...
    Instruction::Call(call) if !call.is_function_call(stack.as_ref())? => {
       // this case
    }
    Instruction::Call(..) | Instruction::CallDynamic(..) => { 
       // the else case
    }
}
```

Antonio95:
> @Antonio95
> 
> I don't understand:
> 
> > Because of the above relaxation,
> > a DynamicRecord can now effectively come from several static Records.
> 
> You mean that multiple DynamicRecords can share a single underlying static Record? Was that not already the case?
> 
> > This, together with 1, means that, upon encountering a record-like input to a function,
> > it can be decided on the spot whether that breaks the local check.
> 
> Nice, that's much simpler.

In case this wasn't clear: the first quote in your message is outdated (it belongs to an older message). The message https://github.com/ProvableHQ/snarkVM/pull/3173#issuecomment-4040660759 superseded that. In particular, now that external closures cannot output records, these "effective remappings of record registers" (due to closure calls) are no longer possible. The comment "a DynamicRecord can now effectively come from several static Records" referred to a situation where e. g. `r0` (static) was cast into `r1` (dynamic); and then, within the same function, passed to an external closure and received into a different record `r2`. Hence `r1` would be virtually cast from `r0` and `r2`, which introduced some complications (the full algorithm which covered that case is in the history of this PR). However, now that such a static record can no longer be remapped to other static-record registers, these complications are gone.

synthesizer/process/src/verify_execution/ensure_records_exist.rs
```diff
@@ -0,0 +1,548 @@
...
+        for record in [old_register, new_register] {
+            debug_assert!(
+                self.families.iter().filter(|family| family.contains(&record)).count() <= 1,
+                "Multiple families contain register {} for transition ID {}",
+                record.1,
+                record.0
+            );
+        }
+
+        let family = self.families.iter_mut().find(|family| family.contains(&old_register));
```
rot256, line 139: Just an FYI: as I can see this only works because the analysis is run in the same order as the execution (e.g. depth first), meaning the invariant is that any register that the next instruction can depend on, has already been processed. Otherwise you would be silently dropping "connections" which should be tracked.

rot256 (approved):
I wonder if we even need to check that a locally minted record is passed to a function before enforcing that it is output: in those cases the creation of the record would have no possible side-effects. I think we considered this at one point, but can't remember why we decided against it?


The new version is much, much more readable.

synthesizer/process/src/verify_execution/ensure_records_exist.rs
Antonio95: You're completely right about this and it's an important principle of the algorithm we've had from the first designs 👌

synthesizer/process/src/verify_execution/ensure_records_exist.rs
Antonio95: Addressed 👌  I changed it to `.validate() -> Result<(), u64>` so that (one of) the incorrect record(s) can be returned for better error messaging.

synthesizer/process/src/verify_execution/ensure_records_exist.rs
Antonio95: Changed 👍

synthesizer/process/src/verify_execution/ensure_records_exist.rs
Antonio95: Changed 👌

synthesizer/process/src/verify_execution/ensure_records_exist.rs
Antonio95: Done! That required duplicating a bit of code in the recursion wrapper (i. e. in `ensure_records_exist` itself), but I agree its a better design. In particular it hardcodes the promise I made in one of the comments that "`global_check` is only populated once:...".

Thank you for the suggestion 👌

synthesizer/process/src/verify_execution/ensure_records_exist.rs
Antonio95: Ah, I forgot about the match + `if` syntax. Thanks for the call 👌

Antonio95:
@rot256 . Thanks a lot for the detailed suggestions 🙌  I've implemented it all in the latest commit, please let me know if you are satisfied with the changes.

As for the following: 
> I wonder if we even need to check that a locally minted record is passed to a function before enforcing that it is output: in those cases the creation of the record would have no possible side-effects. I think we considered this at one point, but can't remember why we decided against it?

There are deployed functions which mint a few instances of records, do some ternary operations over the fields and then mint the final record which they output (thanks goes to @d0cd, which checked this and informed me about it). That's an (admittedly surprising) flow we don't want to disallow, and it would be broken if we required those locally minted records to be output. That's why we added the extra condition that the record must be passed to a function call before we require it to be output.

synthesizer/src/vm/tests/test_v14/translation.rs
```diff
@@ -15,6 +15,8 @@
 
 use super::*;
 
+// TODO fix
```
vicsn: Should this stay?

synthesizer/process/src/stack/register_types/initialize.rs
```diff
@@ -45,10 +45,10 @@ impl<N: Network> RegisterTypes<N> {
         // Step 3. Check the outputs are well-formed.
         for output in closure.outputs() {
             // Ensure the closure output register is not a static record.
-            // Dynamic records are allowed as closure outputs.
+            // ExternalRecord and DynamicRecord are checked at V14+ deployment time (see `VM::check_transaction`).
```
vicsn: Update to V15

synthesizer/program/src/closure/mod.rs
```diff
@@ -179,8 +179,9 @@ impl<N: Network> ClosureCore<N> {
         // Ensure the maximum number of outputs has not been exceeded.
         ensure!(self.outputs.len() < N::MAX_OUTPUTS, "Cannot add more than {} outputs", N::MAX_OUTPUTS);
 
-        // Ensure the closure output register is not a record.
-        ensure!(!matches!(output.register_type(), RegisterType::Record(..)), "Output register cannot be a record");
+        // Ensure the closure output register is not a static record.
+        // ExternalRecord and DynamicRecord are checked at V14+ deployment time (see `VM::check_transaction`).
```
vicsn: Update to V15

synthesizer/src/vm/verify.rs
```diff
@@ -367,6 +367,24 @@ impl<N: Network, C: ConsensusStorage<N>> VM<N, C> {
                         "Invalid deployment transaction '{id}' - expected {num_functions} function and {num_records} record verifying keys after `ConsensusVersion::V14`"
                     );
                 }
+                if consensus_version >= ConsensusVersion::V14 {
```
vicsn: Should this be V15?

synthesizer/src/vm/verify.rs
```diff
@@ -367,6 +367,24 @@ impl<N: Network, C: ConsensusStorage<N>> VM<N, C> {
...
+                    // At V14+, closures must not output records.
```
vicsn: Should this be V15?

synthesizer/src/vm/verify.rs
```diff
@@ -367,6 +367,24 @@ impl<N: Network, C: ConsensusStorage<N>> VM<N, C> {
...
+                    for closure in deployment.program().closures().values() {
+                        for output in closure.outputs() {
+                            ensure!(
+                                !matches!(
+                                    output.register_type(),
+                                    RegisterType::Record(..)
+                                        | RegisterType::ExternalRecord(..)
+                                        | RegisterType::DynamicRecord
+                                ),
+                                "Invalid deployment transaction '{id}' - closure '{}' outputs a record type, which is not allowed at `ConsensusVersion::V14` or later",
```
vicsn: Should this be V15?

vicsn (changes requested):

vicsn (approved):

synthesizer/process/src/verify_execution/ensure_records_exist.rs
```diff
@@ -0,0 +1,566 @@
...
+                        .any(|output| {
+                            matches!(
+                                output.register_type(),
+                                RegisterType::ExternalRecord(..) | RegisterType::DynamicRecord
+                            )
+                        }),
+                };
+                ensure!(
+                    !has_forbidden_output,
+                    "Closure '{}' outputs ExternalRecord or DynamicRecord, which is disallowed at V14+",
```
d0cd:
```suggestion
                    "Closure '{}' outputs ExternalRecord or DynamicRecord, which is disallowed at V15",
```

synthesizer/src/vm/verify.rs
```diff
@@ -222,6 +222,8 @@ impl<N: Network, C: ConsensusStorage<N>> VM<N, C> {
                 //   - the argument bit size of futures does not exceed the maximum allowed size of u16::MAX.
                 //   - V3 deployments (amendments) are not allowed.
                 //   - the deployment has exactly one verifying key per function (no record verifying keys)
+                // If the `CONSENSUS_VERSION` is greater than or equal to `V15`, ensure that
```
d0cd: This comment should be moved to the lower block of consensus version checks.

d0cd (approved):
LGTM! Left a few nits and there still seem to be some bare `// TODO fix` comments in the repo.

PR #3177, opened 2026-03-09
Disallow constant outputs in dynamic calls

Antonio95:
The other day @vicsn spotted the following issue:

Declaring a function output as `output rX as TYPE.constant;` is allowed. In fact, one can deploy and successfully execute function with such questionable shapes as
```
    function foo:
        input r0 as u8.public;
        add r0 3u8 into r1;
        output r1 as u8.constant;
```
obtaining different outputs in different executions. This by itself is odd and we might want to disallow it altogether. However, if that function is called by another one (whether with `call` or `call.dynamic`; this can always be detected at compile time anyway because dynamic call still declare output types), there is an actual error: validation of the deployment certificate fails, likely because different constants are being sampled and baked into the circuit by the prover generating the certificate and the validator. This PR fixes that for the `call.dynamic` case, which can be done without version guards or breaking any previous changes.


synthesizer/program/src/logic/instruction/operation/call/dynamic.rs
```diff
@@ -353,6 +353,14 @@ impl<N: Network> Parser for CallDynamic<N> {
                 let (string, destinations) = many_m_n(1, N::MAX_OPERANDS, complete(parse_destination))(string)?;
                 // Parse the destination types from the string.
                 let (string, destination_types) = parse_value_types(string)?;
+
+                if destination_types.iter().any(|destination_type| matches!(destination_type, ValueType::Constant(..)))
```
d0cd: This check should be moved to `Self::new`.

d0cd: Might also be worth adding a redundant check over the operand types as well.

d0cd (approved):
The spirit of the PR sounds good. 
Left some feedback, and the tests in `synthesizer-program` need to be updated.

d0cd (changes requested):
Meant to request changes.

synthesizer/program/src/logic/instruction/operation/call/dynamic.rs
Antonio95: About the first comment: done here: https://github.com/ProvableHQ/snarkVM/commit/06c33a90e2f6d2ed5999dd52615f360ddeb7c0d1 Let me know if this is what you had in mind.

Antonio95: About the second comment: did you mean performing the same check for inputs to the dynamic call? (but why "redundant"?). If so, I've done it and expanded the test here:
https://github.com/ProvableHQ/snarkVM/pull/3177/commits/d3ee19e687f8d84a8e5973f38fd9c604bc8daf02

Antonio95: I'm bailing with "A constant cannot be passed in as input to a dynamic call.", which could be a bit confusing since I think one can hardcode a constant value and then pass it as public or private (sth like `with 10field (as field.public)` or `with 10field (as field.private)`). In any case, these messages don't seem to sever surface, the parser just fails giving the remaining string.

We can also defer the discussion on messaging and semantics when we address the root of the issue, i. e. function input/output declarations with type constant.

Antonio95:
Fixed the two failing tests in `synthesizer-program`. Switching from draft to ready for review (contingent on the couple small things I asked above).

synthesizer/program/src/logic/instruction/operation/call/dynamic.rs
```diff
@@ -69,10 +69,12 @@ impl<N: Network> CallDynamic<N> {
             operands.len().checked_sub(3).expect("operands.len() >= 3 is checked above") == operand_types.len(),
             "The number of operands and operand types must match"
         );
-        // Ensure that the operand types do not contain a future, dynamic future, record, or external record type.
+        // Ensure that the operand types do not contain a future, dynamic future, record, or external record type;
+        // or a constant type.
```
d0cd:
```suggestion
        // Ensure that the operand types do not contain a future, dynamic future, record, external record type, or a constant type.
```

synthesizer/program/src/logic/instruction/operation/call/dynamic.rs
```diff
@@ -93,9 +95,10 @@ impl<N: Network> CallDynamic<N> {
             destinations.len() == destination_types.len(),
             "The number of destination registers and destination types must match"
         );
-        // Ensure that the destination types do not contain a future, record, or external record type.
+        // Ensure that the destination types do not contain a future, record, or external record type; or a constant type.
```
d0cd:
```suggestion
        // Ensure that the destination types do not contain a future, record, external record type, or a constant type.
```

synthesizer/src/vm/tests/test_v14/call_dynamic.rs
```diff
@@ -3005,3 +3005,98 @@ fn test_batch_proof_instance_limit_exceeded() -> Result<()> {
 
     Ok(())
 }
+
+// Tests that programs containing dynamic calls with declared output types of the form "as <TYPE>.constant"
+// are disallowed.
```
d0cd:
```suggestion
// Tests that programs containing dynamic calls with declared output types of the form "as <TYPE>.constant" are disallowed.
```

d0cd (approved):
LGTM!

PR #3179, opened 2026-03-10
Postrelease v4.5.0 merge mainnet

vicsn:
## Motivation

Regular merge back into staging. Please review the files changed. They contain some hotfixes which went into canary/testnet. We should make sure that the correct logic ends up in staging.


Cargo.toml
```diff
@@ -182,255 +182,255 @@ test_consensus_heights = [ "snarkvm-console/test_consensus_heights", "snarkvm-sy
...
+version = "=4.5.0"
 
 [workspace.dependencies.snarkvm-synthesizer]
 path = "synthesizer"
-version = "=4.4.0"
+version = "=4.5.0"
 
 [workspace.dependencies.snarkvm-synthesizer-error]
 path = "synthesizer/error"
 version = "=4.4.0"
```
ljedrz, line 409: shouldn't this one be bumped as well?

ljedrz (commented):
LGTM with one caveat.

kaimast (approved):
I don't have an in-depth understanding of the external structs implementation, but everything looks good to me.

A simple `cargo update` should fix the security warning that makes CI fail.

mohammadfawaz (approved):
All the external struct related stuff look good to me!

Cargo.toml
vicsn: Interesting case, this was added to staging before the mainnet release. @raychu86 do your scripts take into account to grep for older crate versions to bump as well?

vicsn:
> A simple cargo update should fix the security warning that makes CI fail.

Can you make a separate PR? Thx!

PR #3191, opened 2026-03-16
Deps/update rand

ljedrz:
Update the `rand` dependency from `0.8` to `0.10`. It's a pretty heavy update, but the basic changes are simple:
- `RngCore` > `Rng`
- `Rng` > `RngExt`
- `Rng::gen` > `Rng::random`
- `Rng::gen_bool` > `Rng::random_bool`
- `Rng::gen_range` > `Rng::random_range`
- `Standard` > `StandardUniform`
- `Uniform::from` > `Uniform::new(_inclusive)`
- `rand::distributions` > `rand::distr`

There are 3 open points which I'll mark directly in the diffs.


Cargo.toml
```diff
@@ -477,6 +477,11 @@ version = "2.0"
 [workspace.dependencies.itertools]
 version = "0.14"
 
+[workspace.dependencies.k256]
+version = "0.14.0-rc.8"
```
ljedrz, line 481: note: this is a release candidate, but seems very close to a "full" release

ledger/committee/Cargo.toml
```diff
@@ -54,9 +54,9 @@ features = [ "preserve_order" ]
 workspace = true
 optional = true
 
-[dependencies.proptest]
-version = "1.0.0"
-optional = true
+# [dependencies.proptest]
```
ljedrz, line 57: this dependency still uses `0.9`

algorithms/src/polycommit/optional_rng.rs
```diff
@@ -24,32 +23,32 @@ use rand::RngCore;
...
     #[inline]
-    fn fill_bytes(&mut self, dest: &mut [u8]) {
-        self.0.as_mut().map(|r| r.fill_bytes(dest)).expect("Rng was invoked in a non-hiding context")
+    fn try_next_u64(&mut self) -> Result<u64, Self::Error> {
+        Ok(self.0.as_mut().map(|r| r.next_u64()).expect("Rng was invoked in a non-hiding context"))
     }
 
     #[inline]
-    fn try_fill_bytes(&mut self, dest: &mut [u8]) -> Result<(), rand::Error> {
+    fn try_fill_bytes(&mut self, dest: &mut [u8]) -> Result<(), Self::Error> {
```
ljedrz: this implementation is different than the original one, and I'm not sure if that logic can be preserved, for 2 reasons:
- the `TryRng` trait (the new supertrait of `Rng`) now requires `type Error` to be specified (which means we'd need to have the same error for `try_fill_bytes` and the `try_next_*` methods)
- `rand::Error` no longer exists

ljedrz:
The `wasm` issue should now be fixed.

ljedrz:
Not sure what the issue with `snarkvm-curves` is now, as the latest change was restricted to `snarkvm-wasm`.

ljedrz:
Apparently `target-cpu=native` is poison for CI, especially with mixed machines (which we are using). We might want to override that setting from `config.toml` in the CI config script by applying the `RUSTFLAGS` other than `target-cpu=native` over there directly (or just removing them).

vicsn:
@ljedrz whatever you prefer as the way forward regarding disabling target cpu in CI

ljedrz:
@vicsn done.

algorithms/src/polycommit/optional_rng.rs
```diff
@@ -24,32 +23,32 @@ use rand::RngCore;
...
         match &mut self.0 {
-            Some(r) => r.try_fill_bytes(dest),
-            None => Err(NonZeroU32::new(rand::Error::CUSTOM_START).unwrap().into()),
+            Some(r) => {
+                r.fill_bytes(dest);
+                Ok(())
+            }
+            None => todo!(),
```
vicsn: Is `try_fill_bytes` called from anywhere other than tests? If not can it be compile-guarded? If yes can we make a fool-proof interface?

vicsn (commented):
@kaimast I reviewed most of the PR, can you just double check utilities/src/rand.rs ?

algorithms/src/polycommit/optional_rng.rs
ljedrz: `OptionalRng` is currently used in `SonicKZG10::commit`, but it seems to be a workaround. It appears that these days it can just be removed.

vicsn (approved):
@kaimast no need to review I think it's good

PR #3194, opened 2026-03-18
Decrease sensitivity of retargeting algorithm

raychu86:
<!-- Thank you for submitting the PR! We appreciate you spending the time to work on these changes! -->

## Motivation

The retargeting algorithm for `coinbase_target` is sensitive to block time changes. A 30% increase in block time caused the coinbase target to drop abruptly and cascade downward. This PR decreases the sensitivity of the algorithm by increasing the anchor time from 25 to 35 seconds at `ConsensusVersion::V15`, which will:

  - Reduce the per-block target adjustment magnitude, making the coinbase target less volatile in response to block time fluctuations                                                                                            
  - Increase the half-life, meaning the target takes proportionally longer to halve or double (a sudden crash in prover participation will relieve difficulty pressure more gradually rather than causing a sharp drop)
  - Cause the `coinbase_target` to trend upward over time when block times are consistently below the anchor, as the larger anchor widens the gap between actual and expected block time
  
  
The coinbase reward algorithm does not change and still uses the original anchor time, which is now defined as `REWARD_ANCHOR_TIME`.

## Implementation

This PR replaces `N::ANCHOR_TIME` with `N::ANCHOR_TIMES`, a versioned array following the same pattern as `MAX_PROGRAM_SIZE` and `MAX_WRITES`. The V15 entry sets the anchor time to 35 seconds (from 25 seconds), activating on `ConsensusVersion::V15`.

`to_next_targets` now accepts a `ConsensusVersion` as its first parameter and looks up the active anchor time to use for retargeting. `ANCHOR_HEIGHT` is preserved as a hardcoded V1 literal since it is used exclusively by `coinbase_reward_v1`, which occurs well before `ConsensusVersion::V15`.
                                                                                                                                                                                                                                                        
## Test Plan

Tests have been added to verify that the coinbase target retargeting direction and magnitude differ between V1 (25s) and V14 (35s) anchor times, that the halving point shifts proportionally, and that doubling takes more blocks under the larger anchor time.

## Backwards compatibility

Existing behavior is unchanged before V15. The new anchor time is gated behind ConsensusVersion::V15.


ledger/block/src/verify.rs
```diff
@@ -402,7 +403,7 @@ impl<N: Network> Block<N> {
             timestamp,
             N::GENESIS_TIMESTAMP,
             N::STARTING_SUPPLY,
-            N::ANCHOR_TIME,
+            N::ANCHOR_TIMES[0].1,
             N::ANCHOR_HEIGHT,
```
ljedrz, line 406: shouldn't the `N::ANCHOR_HEIGHT` be hardcoded instead? the `N::ANCHOR_TIME` is used in `coinbase_reward_v2`

ledger/src/advance.rs
```diff
@@ -400,7 +401,7 @@ impl<N: Network, C: ConsensusStorage<N>> Ledger<N, C> {
             next_timestamp,
             N::GENESIS_TIMESTAMP,
             N::STARTING_SUPPLY,
-            N::ANCHOR_TIME,
```
ljedrz, line 403: ditto

ljedrz (changes requested):
I think the other anchor parameter should be hardcoded in the calls to `coinbase_reward`, as it's the 6th argument that is used in `coinbase_reward_v1`, and the 5th one (currently hardcoded to the original value) in `coinbase_reward_v2`.

ledger/block/src/verify.rs
raychu86: The intention was to keep the rewards/emissions the same and only update the retargeting algorithm.

I agree this is confusing.


`N::ANCHOR_HEIGHT` shouldn't need changing because it's only relevant in `coinbase_reward_v1` which is after the new `anchor_time` kicks in.

raychu86: Changes made in [717ad49](https://github.com/ProvableHQ/snarkVM/pull/3194/commits/717ad49a4015dc52a46959a89833b43ee0f8c2e8)

ljedrz (approved):

ledger/block/src/verify.rs
ljedrz: Simple and elegant :+1:.

ledger/block/src/helpers/target.rs
```diff
@@ -1903,4 +1909,222 @@ mod tests {
...
+            )
+            .unwrap();
+            previous_timestamp = next_timestamp;
+            num_blocks_v15 += 1;
+        }
+
+        // A larger anchor time shrinks the drift-to-anchor ratio (15/35 < 15/25),
+        // so more blocks are needed to double the target.
+        assert!(num_blocks_v15 > num_blocks_v1);
+    }
```
vicsn, line 2129: Can/should we add tests modelling the change in coinbase reward, if any?

vicsn (changes requested):

ledger/block/src/helpers/target.rs
ljedrz: IIUC, the reward remains unchanged

raychu86: Yes the coinbase reward will remain the same. Only the targeting will be different

vicsn (approved):

Issue #3193, opened 2026-03-17
[Bug] Potential deadlock in dry-run speculation while holding VM.process write lock through post-ratify


damons:
## 🐛 Bug Report

We observed a validator stall on the `v4.5.0` / `mainnet` release line that does not match the two previously diagnosed `current_block` deadlocks.

Live `strace` and `gdb` debugging on a hung validator on March 17, 2026 points instead at a deadlock centered on the VM `process` lock during dry-run speculation.

The current working theory is:

- `atomic_speculate_inner()` acquires `self.process.write()`
- it keeps that write guard for the full dry-run atomic batch
- it later enters `atomic_post_ratify::<false>()`
- post-ratify reads committee / delegated / bonded mappings via `FinalizeStore::get_mapping_speculative()`
- the RocksDB nested-map path parallelizes deserialization with Rayon
- meanwhile many worker threads are already blocked on `self.process.read()` via transaction verification and transaction spend estimation
- the speculate thread waits on Rayon-backed mapping work while still holding `process.write()`

That produces a plausible cycle where:

- the speculate thread holds `process.write()`
- reader-side workers cannot acquire `process.read()`
- post-ratify mapping work does not complete
- the speculate thread never drops `process.write()`

This appears to be a separate deadlock class from the earlier `current_block` re-entry issues.

## Steps to Reproduce

This is not yet reduced to a deterministic unit-test repro. The live repro we have so far is:

1. Run a `snarkOS` validator using the `snarkVM 4.5.0` release line on mainnet.
2. Wait for the validator to stall locally while peers continue advancing.
3. Attach `strace` and `gdb` to the hung process.
4. Observe:
   - many threads blocked in `lock_shared_slow`
   - readers blocked in `snarkvm_synthesizer::vm::verify::<impl VM>::check_transaction()`
   - readers blocked in `snarkos_node_bft_ledger_service::ledger::CoreLedgerService::transaction_spend_in_microcredits()`
   - one sequential-operation thread blocked in `atomic_speculate_inner()` while walking finalize-store mappings / waiting on a Rayon latch

From the live debug session, the blocked reader sites resolved to:

- `self.process.read().get_stack(transition.program_id())` in `synthesizer/src/vm/verify.rs`
- `execution_cost(&self.ledger.vm().process().read(), execution, consensus_version)` in `snarkOS/node/bft/ledger-service/src/ledger.rs`

And the blocked speculate path resolved to:

- `atomic_speculate_inner()`
[...]

PR #3196, opened 2026-03-23
[Fix] Reduce locking around the Process

ljedrz:
This PR is in response to https://github.com/ProvableHQ/snarkVM/issues/3193.

The current setup around the `Process` is unfortunate; its contents are either immutable or they are locked, but the `Process` itself is in an `RwLock`, and has some `&mut` methods. This is confusing, and might lead to complicated issues like the one in the link above.

While I initially considered just relaxing 2 instances of write guards to read guards, in the end I elected to remove the `RwLock` from the `Process`, and introduce an internal lock that can be used to maintain the current desired behavior of the write guards (blocking concurrent reads and writes from elsewhere), but with reduced (future) risk of deadlocks.

Note: since finalization and block insertion happen completely sequentially in a dedicated thread, the newly introduced lock is not a safeguard against concurrent use between them, but against any future use which may happen outside of those operations, but rely on the `Process` not being partially modified etc.

### DIFF
--- a/synthesizer/process/src/lib.rs
+++ b/synthesizer/process/src/lib.rs
@@ -112,6 +112,133 @@ pub struct Process<N: Network> {
     lock: Mutex<()>,
 }
 
+/// A wrapper ensuring exclusive access to the Process for the purposes of add_stacks.
+pub struct ProcessExclusiveGuard<'a, N: Network> {
+    process: &'a Process<N>,
+    #[cfg(not(feature = "locktick"))]
+    _guard: MutexGuard<'a, ()>,
+    #[cfg(feature = "locktick")]
+    _guard: LockGuard<MutexGuard<'a, ()>>,
+}
+
+impl<'a, N: Network> ProcessExclusiveGuard<'a, N> {
+    /// Adds a new stack to the process.
+    /// If the program already exists, then the existing stack is replaced and the original stack is returned.
+    /// Note. This method assumes that the provided stack is valid.
+    #[inline]
+    pub fn add_stack(&self, stack: Stack<N>) -> Option<Arc<Stack<N>>> {
+        // Get the program ID.
+        let program_id = *stack.program_id();
+        // Arc the stack first to limit the scope of the write lock.
+        let stack = Arc::new(stack);
+        // Insert the stack into the process, replacing the existing stack if it exists.
+        self.process.stacks.write().insert(program_id, stack)
+    }
+
+    /// Stages a stack to be added to the process.
+    /// The new stack is active, while the old stack is retained in `old_stacks`.
+    /// The `commit_stacks` method must be called to finalize the addition of the new stack.
+    /// The `revert_stacks` method can be called to revert the staged stacks.
+    #[inline]
+    pub fn stage_stack(&self, stack: Stack<N>) {
+        // Get the program ID.
+        let program_id = *stack.program_id();
+        // Arc the stack first to limit the scope of the write lock.
+        let stack = Arc::new(stack);
+        // If no entry in `old_stacks` exists for `program_id`, store the old stack.
+        // Note: If `old_stack` is `None`, it means that we are adding a new program to the process.
+        let old_stack = self.process.stacks.write().insert(program_id, stack);
+        let mut old_stacks = self.process.old_stacks.write();
+        if !old_stacks.contains_key(&program_id) {
+            old_stacks.insert(program_id, old_stack);
+        }
+    }
+
+    /// Commits the staged stacks to the process.
+    /// This finalizes the addition of the new stacks and clears the old stacks.
+    #[inline]
+    pub fn commit_stacks(&self) {
+        // Clear the old stacks.
+        self.process.old_stacks.write().clear();
+    }
+
+    /// Reverts the staged stacks, restoring the previous state of the process.
+    /// This will remove the new stacks and restore the old stacks.
+    #[inline]
+    pub fn revert_stacks(&self) {
+        // Restore the old stacks.
+        for (program_id, stack) in self.process.old_stacks.write().drain(..) {
+            // If the stack is `None`, remove the program from the process.
+            // Otherwise, insert the old stack back into the process.
+            if let Some(stack) = stack {
+                self.process.stacks.write().insert(program_id, stack);
+            } else {
+                self.process.stacks.write().shift_remove(&program_id);
+            }
+        }
+    }
+
+    /// Adds a new program to the process, verifying that it is a valid addition.
+    /// If the program exists, then the existing stack is replaced and discarded.
+    /// Note. This method should **NOT** be used by the on-chain VM to add new program, use `finalize_deployment` or `load_deployment` instead instead.
+    #[inline]
+    pub fn add_program(&self, program: &Program<N>) -> Result<()> {
+        // Initialize the 'credits.aleo' program ID.
+        let credits_program_id = ProgramID::<N>::from_str("credits.aleo")?;
+        // If the program is not 'credits.aleo', compute the program stack, and add it to the process.
+        if program.id() != &credits_program_id {
+            self.add_stack(Stack::new(self.process, program)?);
+        }
+        Ok(())
+    }
+
+    /// Adds a new program with the given edition to the process, verifying that it is a valid addition.
+    /// If the program exists, then the existing stack is replaced and discarded.
+    /// Note. This method should **NOT** be used by the on-chain VM to add new program, use `finalize_deployment` or `load_deployment` instead instead.
+    #[inline]
+    pub fn add_program_with_edition(&self, program: &Program<N>, edition: u16) -> Result<()> {
+        // Initialize the 'credits.aleo' program ID.
+        let credits_program_id = ProgramID::<N>::from_str("credits.aleo")?;
+        // If the program is not 'credits.aleo', compute the program stack, and add it to the process.
+        if program.id() != &credits_program_id {
+            let stack = Stack::new_raw(self.process, program, edition)?;
+            stack.initialize_and_check(self.process)?;
+            self.add_stack(stack);
+        }
+        Ok(())
+    }
+
+    /// Adds a set of programs and editions, in topological order, to the process, deferring validation of the programs until all programs are added.
+    /// If a program exists, then the existing stack is replaced and discarded.
+    /// Either all programs are added or none are.
+    /// Note. This method should **NOT** be used by the on-chain VM to add new program, use `finalize_deployment` or `load_deployment` instead instead.
+    #[inline]
+    pub fn add_programs_with_editions(&self, programs: &[(Program<N>, u16)]) -> Result<()> {
+        // Initialize the 'credits.aleo' program ID.
+        let credits_program_id = ProgramID::<N>::from_str("credits.aleo")?;
+        // Defer cleanup of the uncommitted stacks.
+        defer! {
+            self.revert_stacks()
+        }
+        // Initialize raw stacks for each of the programs, skipping `credits.aleo`.
+        for (program, edition) in programs {
+            if program.id() != &credits_program_id {
+                self.stage_stack(Stack::new_raw(self.process, program, *edition)?)
+            }
+        }
+        // For each stack, check and initialize it before adding it to the process.
+        for (program, _) in programs {
+            // Retrieve the stack.
+            let stack = self.process.get_stack(program.id())?;
+            // Initialize and check the stack for well-formedness.
+            stack.initialize_and_check(self.process)?;
+        }
+        // Commit the staged stacks.
+        self.commit_stacks();
+        Ok(())
+    }
+}
+
 impl<N: Network> Process<N> {
     /// Initializes a new process.
     #[inline]
@@ -146,7 +273,7 @@ impl<N: Network> Process<N> {
         lap!(timer, "Synthesize credits program keys");
 
         // Add the 'credits.aleo' stack to the process.
-        process.add_stack(stack);
+        process.lock().add_stack(stack);
 
         finish!(timer);
         // Return the process.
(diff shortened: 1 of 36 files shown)


vicsn (approved):
LGTM, let's see what Ray says. We'll of course run prerelease tests on the node with this new VM

raychu86 (commented):
Looks mostly fine to me, but does this new lock catch all the cases of when process is properly updated?

For instance, in `add_next_block_inner` we call `update_credits_verifying_keys` that updates the keys stored in the process.

synthesizer/process/src/lib.rs
```diff
@@ -83,21 +83,27 @@ use snarkvm_utilities::{defer, dev_println};
...
     /// The universal SRS.
     universal_srs: UniversalSRS<N>,
     /// The mapping of program IDs to stacks.
     stacks: Arc<RwLock<IndexMap<ProgramID<N>, Arc<Stack<N>>>>>,
     /// The mapping of program IDs to old stacks.
-    old_stacks: Arc<RwLock<IndexMap<ProgramID<N>, Option<Arc<Stack<N>>>>>>,
+    old_stacks: RwLock<IndexMap<ProgramID<N>, Option<Arc<Stack<N>>>>>,
+    /// A lock guarding the entire Process in case no concurrent reads or writes
+    /// on the object are to be permitted.
+    lock: Mutex<()>,
```
mitchmindtree, line 112: If the original lock was required to ensure atomic consistency of `Process` internals, it seems like this change just makes it easier for devs to forget to lock the process correctly when needed (looks like Ray might have highlighted an example already). Just want to highlight that under this new approach, nothing in the type system prevents e.g. accessing `stacks` or `old_stacks` without holding `lock`.

> and introduce an internal lock that can be used to maintain the current desired behavior of the write guards (blocking concurrent reads and writes from elsewhere), but with reduced (future) risk of deadlocks

It's not super clear to me that this will reduce deadlocking in the future (its still possible for 2 different threads/tasks to hold `Process` and end up waiting on eachother for this new lock guard). That said, maybe it does resolve #3193 which seems to be the priority here.

synthesizer/process/src/lib.rs
```diff
@@ -137,11 +147,23 @@ impl<N: Network> Process<N> {
...
+    #[cfg(feature = "locktick")]
+    pub fn lock(&self) -> LockGuard<MutexGuard<()>> {
+        self.lock.lock()
+    }
+
+    /// Guard the Process against any concurrent reads or writes.
+    #[cfg(not(feature = "locktick"))]
+    pub fn lock(&self) -> MutexGuard<()> {
+        self.lock.lock()
+    }
```
mitchmindtree: We might want to at least mention in doc comments that this is purely co-operative, i.e. nothing in the type-system is stopping another holder of a `Process` from writing just because you hold this lock.

mitchmindtree (commented):
I've left some thoughts inline as I was pinged for review - forgive me if I'm missing some context 🙏

---

Sidenote: In general, the amount of `Arc`s and locks throughout snarkVM feel like a code-smell and are difficult to reason about. It gives me the impression that a lot of state is unnecessarily lumped together, when it should be kept separate and composed/passed-by-value via functions as needed. But this probably warrants a deeper-dive into exactly how snarkOS uses `Process` and all of its parts across different threads.

ljedrz:
> Looks mostly fine to me, but does this new lock catch all the cases of when process is properly updated?

@raychu86 The intent is for that lock to be used in the future if such cases are to be introduced.

> It gives me the impression that a lot of state is unnecessarily lumped together, when it should be kept separate and composed/passed-by-value via functions as needed.

@mitchmindtree I would prefer for the Process to be elsewhere too, but indeed - to refactor it could be a heavy lift.

synthesizer/process/src/lib.rs
ljedrz: > If the original lock was required to ensure atomic consistency of Process internals, it seems like this change just makes it easier for devs to forget to lock the process correctly when needed

That was the _intent_, but the way it was done ended up being unidiomatic/misleading (possibly after later changes). Mixing exterior and interior mutability is typically an antipattern and very likely to introduce deadlocks.

> Just want to highlight that under this new approach, nothing in the type system prevents e.g. accessing stacks or old_stacks without holding lock.

Indeed; however, it is documented that its purpose is to lock out the internals of the process. I would have preferred to use an outer `RwLock` had it not been for the fact that the internals are either immutable or under locks as well.

> That said, maybe it does resolve https://github.com/ProvableHQ/snarkVM/issues/3193 which seems to be the priority here.

Precisely - it's not an ideal solution (that would require the `Process` to be decoupled from the `VM`, or the `VM` from the `Ledger`), but that would require a much larger refactoring.

ljedrz:
I [rebased the original](https://github.com/ProvableHQ/snarkVM/pull/3196/commits/97e0eedc933d27a3a5db1c5008c65938f0ded56e), and proposed enhanced futureproofing in the [new commit](https://github.com/ProvableHQ/snarkVM/pull/3196/commits/10ff1ec87c991a7ed1d8496727d68fd4d9e2812d).

The enhanced approach is as follows:
1. introduce a new `ProcessExclusiveGuard` wrapper object; only one such wrapper may exist at any given moment
2. move all the sensitive `Process` methods to `ProcessExclusiveGuard`, which enforces strong type safety for all the desired functionalities; the only changes inside them are caused by extra redirection - no logic is affected
3. adjust the existing calls to the aforementioned methods to include the wrapper transition; most of them are in tests

Please double-check that I've moved all the `Process` methods that are considered sensitive to `ProcessExclusiveGuard`.

vicsn (changes requested):
Please correct me if I'm wrong, but I think all of the iterations of this PR are susceptible to the following attack:
- We start `atomic_speculate` or `atomic_finalize` and halfway through we add or upgrade a `Stack` because of a finalized `Deployment`.
- Validators concurrently call `check_transaction` on a fresh incoming transaction. They read the pending `Stack`. Proof verification succeeds and we cache this in `partially_verified_transactions`. The `Proof` never gets checked again.
- For whatever reason, `atomic_speculate` or `atomic_finalize` reverts. Now `partially_verified_transactions` is corrupted and can lead to invalid state or a network fork. I'm not aware of how a a malicious validator could make this happen in practice, but IMO it's a reasonable assumption to make.

If the issue is limited to our `Stacks` management, then perhaps we can refactor `finalize.rs` to no longer tentatively add stacks, but instead to track them as `pending_stacks` which are only added to `process.stacks` at the end of `atomic_finalize`. @d0cd do you recall why we went the`revert_stacks` route? IMO there is no need to use the new `Stacks` until `atomic_finalize` finished.

This goes to show how brittle our VM logic is, and that for such a change we need a bit more detailed design document or PR README describing the guarantees of `Process` and it's contents.

The above particular issue would not have been possible with a more robust `TransactionCacheKey`, so making that change now: https://github.com/ProvableHQ/snarkVM/issues/3219

ljedrz:
> I think all of the iterations of this PR are susceptible to the following attack

The angle you're describing is basically status quo - this PR just removes the added deadlock potential.

> perhaps we can refactor finalize.rs to no longer tentatively add stacks, but instead to track them as `pending_stacks` which are only added to process.stacks at the end of `atomic_finalize`

That would work just fine as long as we don't need to depend on incremental status of the stacks during finalization.

d0cd:
@vicsn doesn't the `TransactionCacheKey` capture editions, which are incremented in the upgrade case that you described above?

We went with the `revert_stacks` approach to preserve the linear semantics of transactions. Using the new stacks after `atomic_finalize` creates implicit "delay" of upgrades until the end of the block.

ljedrz:
I proposed a delayed inclusion of the new stacks in https://github.com/ProvableHQ/snarkVM/pull/3196/commits/a3798b6a655fd5e53f183acdf51e9e7d353e2cf1, which solves potential stack incoherence issues and is simpler than the current approach.

ljedrz:
The CI errors are unrelated as far as I can tell:
```
Err:19 http://archive.ubuntu.com/ubuntu jammy-updates/universe amd64 Packages
  Hash Sum mismatch
  Hashes of expected file:
   - Filesize:1622817 [weak]
   - SHA256:e3d87b76559fb14dcbf9330344c31b2c6e5476488498bb06bf01b4b13092c2c2
   - SHA1:6df1852fa201fa855518d02c613eecde1292c360 [weak]
   - MD5Sum:2f39150c5bfb13ddf01eaa71672a75cb [weak]
  Hashes of received file:
   - SHA256:029dc96211412fd2fd1a0010fb35eb1ba39342eb039fb39d924135288c2ccf7d
   - SHA1:260eab80c959e97973ec41bbe4988e76697637e6 [weak]
   - MD5Sum:685b11abaf44916b422a3be301f5667e [weak]
   - Filesize:1622817 [weak]
  Last modification reported: Thu, 16 Apr 2026 03:07:21 +0000
  Release file created at: Thu, 16 Apr 2026 07:20:03 +0000
Get:20 http://archive.ubuntu.com/ubuntu jammy-updates/main amd64 Packages [4,262 kB]
Fetched 46.1 MB in 2min 10s (355 kB/s)                                         
Reading package lists... Done
W: https://download.docker.com/linux/ubuntu/dists/jammy/InRelease: Key is stored in legacy trusted.gpg keyring (/etc/apt/trusted.gpg), see the DEPRECATION section in apt-key(8) for details.
E: Failed to fetch http://archive.ubuntu.com/ubuntu/dists/jammy-updates/universe/binary-amd64/Packages.gz  Hash Sum mismatch
   Hashes of expected file:
    - Filesize:1622817 [weak]
    - SHA256:e3d87b76559fb14dcbf9330344c31b2c6e5476488498bb06bf01b4b13092c2c2
    - SHA1:6df1852fa201fa855518d02c613eecde1292c360 [weak]
    - MD5Sum:2f39150c5bfb13ddf01eaa71672a75cb [weak]
   Hashes of received file:
    - SHA256:029dc96211412fd2fd1a0010fb35eb1ba39342eb039fb39d924135288c2ccf7d
    - SHA1:260eab80c959e97973ec41bbe4988e76697637e6 [weak]
    - MD5Sum:685b11abaf44916b422a3be301f5667e [weak]
    - Filesize:1622817 [weak]
   Last modification reported: Thu, 16 Apr 2026 03:07:21 +0000
   Release file created at: Thu, 16 Apr 2026 07:20:03 +0000
E: Some index files failed to download. They have been ignored, or old ones used instead.
```

synthesizer/src/vm/finalize.rs
```diff
@@ -684,16 +684,16 @@ impl<N: Network, C: ConsensusStorage<N>> VM<N, C> {
...
-            let process = self.process.write();
+            // we choose to acquire it for the entire duration of this atomic batch.
+            let process = self.process.lock();
 
             // Revert any unstaged stacks, when the function returns.
             // Note. `commit_stacks` is called at the bottom of this function after successful finalization.
             //  The staged stacks are only reverted if the function returns an error.
             defer! {
-                process.revert_stacks();
+                process.cancel_staged_stacks();
```
vicsn, line 696: Would it be safer if `commit_stacks` was called after `atomic_finalize!` finishes?

vicsn (commented):
In line with Ray's [earlier comment](https://github.com/ProvableHQ/snarkVM/pull/3196#pullrequestreview-4010668740), would it be possible to make `update_credits_verifying_keys` a member of `ProcessExclusiveGuard`?

synthesizer/process/src/lib.rs
```diff
@@ -86,24 +86,145 @@ use aleo_std::prelude::{finish, lap, timer};
...
-#[derive(Clone)]
 pub struct Process<N: Network> {
     /// The universal SRS.
     universal_srs: UniversalSRS<N>,
     /// The mapping of program IDs to stacks.
     stacks: Arc<RwLock<IndexMap<ProgramID<N>, Arc<Stack<N>>>>>,
-    /// The mapping of program IDs to old stacks.
-    old_stacks: Arc<RwLock<IndexMap<ProgramID<N>, Option<Arc<Stack<N>>>>>>,
+    /// The mapping of program IDs to new, pending stacks.
+    new_stacks: RwLock<IndexMap<ProgramID<N>, Arc<Stack<N>>>>,
```
d0cd, line 109: I believe this change breaks the semantics of existing programs and could result in a failure to sync from scratch.
While it is reasonable to stage new stacks and commit at the end of block, you need to route `get_stack` to use the latest, possibly new stack.

synthesizer/src/vm/finalize.rs
ljedrz: I kept the preexisting structure to reduce the diff; it would be possible, though it would likely make the API "clunkier", as some of the methods on `Process` called by `atomic_finalize` may affect the stacks, meaning the pending stacks would have to be manually "carried around".

As for safety, the (unchanged) approach would only be troublesome if we panicked during finalization, but at that point the sequential processing thread would crash, so - unlike with a panicking task - it would not be recoverable (though a silently panicked task would be a big problem as well), meaning the `Process` wouldn't get "polluted".

ljedrz:
> would it be possible to make `update_credits_verifying_keys` a member of `ProcessExclusiveGuard`?

@vicsn Done in https://github.com/ProvableHQ/snarkVM/pull/3196/commits/276462a983152f13d111dd294f9158900bbf1c4e.

synthesizer/process/src/lib.rs
ljedrz: Also, it seems to break some of the `package` tests; perhaps they rely on this behavior.

vicsn: I had assumed this PR would not break existing finalization because of the following rule about **executions** in `should_abort_transaction`: `The transaction contains a transition that has been deployed or upgraded in this block.`.

However, as far as I can tell, consecutive **deployments** on staged stacks are possible... Maybe that scenario never happened yet in production, but now things are indeed getting very messy.

Let's revert a3798b6a6 -  relying on https://github.com/ProvableHQ/snarkVM/pull/3220 instead

In terms of ordering, let's merge this PR first, and only afterwards https://github.com/ProvableHQ/snarkVM/pull/3220

ljedrz: https://github.com/ProvableHQ/snarkVM/commit/a3798b6a655fd5e53f183acdf51e9e7d353e2cf1 was reverted.

vicsn (approved):

raychu86 (approved):

synthesizer/process/src/lib.rs
```diff
@@ -180,23 +159,156 @@ impl<N: Network> Process<N> {
...
+            // Initialize and check the stack for well-formedness.
+            stack.initialize_and_check(self.process)?;
+        }
+        // Commit the staged stacks.
+        self.commit_stacks();
+        Ok(())
+    }
+
+    /// Update the `credits.aleo` program in the VM with the latest verifying keys.
+    pub fn update_credits_verifying_keys(&self) -> Result<()> {
```
d0cd, line 244: Is there an edge case where a verifier can see inconsistent state between a `credits.aleo` keys update? Making an update here doesn't change the checksums or editions.

synthesizer/process/src/lib.rs
```diff
@@ -180,23 +159,156 @@ impl<N: Network> Process<N> {
...
+        self.process.old_stacks.write().clear();
     }
 
     /// Reverts the staged stacks, restoring the previous state of the process.
     /// This will remove the new stacks and restore the old stacks.
     #[inline]
     pub fn revert_stacks(&self) {
         // Restore the old stacks.
-        for (program_id, stack) in self.old_stacks.write().drain(..) {
+        for (program_id, stack) in self.process.old_stacks.write().drain(..) {
```
d0cd, line 172: Note that a verifier might read an inconsistent snapshot in between a reversion, not a correctness issue but wasted work.

synthesizer/process/src/lib.rs
```diff
@@ -86,74 +86,53 @@ use aleo_std::prelude::{finish, lap, timer};
...
-#[derive(Clone)]
 pub struct Process<N: Network> {
     /// The universal SRS.
     universal_srs: UniversalSRS<N>,
     /// The mapping of program IDs to stacks.
     stacks: Arc<RwLock<IndexMap<ProgramID<N>, Arc<Stack<N>>>>>,
     /// The mapping of program IDs to old stacks.
-    old_stacks: Arc<RwLock<IndexMap<ProgramID<N>, Option<Arc<Stack<N>>>>>>,
+    old_stacks: RwLock<IndexMap<ProgramID<N>, Option<Arc<Stack<N>>>>>,
+    /// A lock guarding the entire Process in case no concurrent reads or writes
```
d0cd: Is this comment accurate, do readers take the lock?

d0cd (approved):
LGTM, left some minor comments. It could be helpful to explicitly document the concurrency model/invariants of `Process` for future maintainers.

synthesizer/process/src/lib.rs
ljedrz: I can see that the only place we ever use that method is post-finalize when we're advancing to `ConsensusVersion::V8`; that path is perfectly serial, so if the only spot where we are concerned with the state of the proving or verifying keys of `credits.aleo` is finalization, an inconsistent read is impossible

synthesizer/process/src/lib.rs
ljedrz: it's a bit outdated, I'll adjust it

synthesizer/process/src/lib.rs
ljedrz: actually, we should be able to lock the stacks for the duration of the reverts - I haven't noticed it before, as I only moved this method under `ProcessExclusiveGuard`

synthesizer/process/src/lib.rs
ljedrz: https://github.com/ProvableHQ/snarkVM/pull/3196/commits/0a2bd2857f946db7ee4a1ca8d7ebe22aa01fd866

synthesizer/process/src/lib.rs
ljedrz: https://github.com/ProvableHQ/snarkVM/pull/3196/commits/1486f30994bb8ee27077bdca2cfdb2baeb866e7f

synthesizer/process/src/lib.rs
vicsn: Great catch, this function is not safe. Added some documentation here: 11caa356c and here: https://github.com/ProvableHQ/snarkVM/issues/3236#issuecomment-4370502616

PR #3201, opened 2026-03-26
[Feat] Introduce `commit.*.raw` instructions

raychu86:
<!-- Thank you for submitting the PR! We appreciate you spending the time to work on these changes! -->

## Motivation

This PR adds `.raw` variants of all commit instructions, activated at `ConsensusVersion::V15`. Raw variants commit over the untagged bit representation of the input, skipping the type-tagged serialization used by the standard variants.                                                                        
                                                                                                                                                                                                                                 
  New opcodes:                                                                                                                                                                                                                   
  - `commit.bhp256.raw`                                                                                                                                                                                                        
  - `commit.bhp512.raw`
  - `commit.bhp768.raw`                                                                                                                                                                                                          
  - `commit.bhp1024.raw`
  - `commit.ped64.raw`                                                                                                                                                                                                           
  - `commit.ped128.raw`                                                                                                                                                                                                        

## Test Plan

Unit tests for each raw variant covering parse, evaluate, and execute, and a divergence test confirming that raw and standard variants produce different outputs for the same input.

Integration tests verifying that programs using `commit.*.raw` are rejected before V15 and accepted at V15, and that `commit.ped128.raw` satisfies the Pedersen additive homomorphism property in both the function body and finalize.                                                                                                                                                       
                                                                                                                                                                                                                                 
## Backwards compatibility                                                                                                                                                                                                     

All new opcodes are gated behind `ConsensusVersion::V15`. Deployment of programs containing `commit.*.raw` is rejected before V15 via `Program::contains_v15_syntax()` enforced in `vm/verify.rs`. No existing behavior is affected.

### DIFF
--- a/synthesizer/program/src/lib.rs
+++ b/synthesizer/program/src/lib.rs
@@ -1323,6 +1323,36 @@ impl<N: Network> ProgramCore<N> {
         Ok(false)
     }
 
+    /// Returns `true` if a program contains any V15 syntax.
+    /// This includes:
+    /// 1. `commit.*.raw` opcodes (raw commit variants).
+    ///
+    /// This is enforced to be `false` for programs before `ConsensusVersion::V15`.
+    #[inline]
+    pub fn contains_v15_syntax(&self) -> bool {
+        // Helper to check if an opcode is a raw commit variant.
+        let has_op = |opcode: &str| opcode.starts_with("commit.") && opcode.ends_with(".raw");
+
+        // Determine if any function instructions contain the new syntax.
+        let function_contains = cfg_iter!(self.functions())
+            .flat_map(|(_, function)| function.instructions())
+            .any(|instruction| has_op(*instruction.opcode()));
+
+        // Determine if any closure instructions contain the new syntax.
+        let closure_contains = cfg_iter!(self.closures())
+            .flat_map(|(_, closure)| closure.instructions())
+            .any(|instruction| has_op(*instruction.opcode()));
+
+        // Determine if any finalize commands or constructor commands contain the new syntax.
+        let command_contains = cfg_iter!(self.functions())
+            .flat_map(|(_, function)| function.finalize_logic().map(|finalize| finalize.commands()))
+            .flatten()
+            .chain(cfg_iter!(self.constructor).flat_map(|constructor| constructor.commands()))
+            .any(|command| matches!(command, Command::Instruction(instruction) if has_op(*instruction.opcode())));
+
+        function_contains || closure_contains || command_contains
+    }
+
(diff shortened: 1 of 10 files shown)


synthesizer/program/src/logic/instruction/operation/commit.rs
```diff
@@ -397,4 +434,46 @@ mod tests {
...
+
+        assert_ne!(
+            standard, raw,
+            "commit.bhp256 and commit.bhp256.raw must produce different outputs for the same input"
+        );
+
+        // Check that committing on the field element is equivalent to the raw commit via the instruction.
+        let expected_commitment = N::commit_bhp256(&input_field.to_bits_le(), &randomizer).unwrap();
+        assert_eq!(Literal::Field(expected_commitment), raw);
+    }
```
Antonio95, line 478: It would be nice to have a test that checks that the raw commit methods return the same on structures with the same underlying field-element contents but which are different in terms of `Value` semantics (for instance, I think `struct {u8, u8}` and `[u8; 2]` should work?). Sth very minimal just as a sanity check

synthesizer/program/src/lib.rs
```diff
@@ -1323,6 +1323,36 @@ impl<N: Network> ProgramCore<N> {
...
+            .any(|instruction| has_op(*instruction.opcode()));
+
+        // Determine if any finalize commands or constructor commands contain the new syntax.
+        let command_contains = cfg_iter!(self.functions())
+            .flat_map(|(_, function)| function.finalize_logic().map(|finalize| finalize.commands()))
+            .flatten()
+            .chain(cfg_iter!(self.constructor).flat_map(|constructor| constructor.commands()))
+            .any(|command| matches!(command, Command::Instruction(instruction) if has_op(*instruction.opcode())));
+
+        function_contains || closure_contains || command_contains
```
Antonio95, line 1353: We could early return `true` before this if any of the three conditions are found to hold and spare the computation of the remaining one(s).

synthesizer/src/vm/tests/test_v15/commit_raw.rs
```diff
@@ -0,0 +1,182 @@
...
+    input r0 as field.public;
+    input r1 as scalar.public;
+    commit.bhp256.raw r0 r1 into r2 as field;
+    output r2 as field.public;
+
+constructor:
+    assert.eq true true;
+";
+
+// A program for testing the additive homomorphism property of `commit.ped128.raw`:
```
Antonio95, line 39: Cool test 👌

Antonio95 (approved):
LGTM, just two very minor comments.

PR #3205, opened 2026-03-27
Extend dev_skip_checks so vk public inputs don't have to match

vicsn:
## Motivation

We have a `dev_skip_checks` feature so someone testing deployments won't have to synthesize function and record circuits. In this workflow, the tests uses `leo deploy --skip-deploy-certificate` to use a fixed placeholder verifying key.

We recently added an optimistic check requiring that the public inputs specified by the verifying key and execution match. But this can only be valid if the deployer goes through the effort of synthesizing the circuit.

## Test Plan

Tested manually `leo deploy --skip-deploy-certificate` on an updated https://github.com/ProvableHQ/ARCs/tree/arc20_approvals


d0cd (approved):
LGTM!

PR #3207, opened 2026-03-27
Compute the cost of an execution based on the root function, inputs and signer address alone 

Antonio95:
This PR adds the function
```
pub fn execution_cost_for_call<A: Aleo, R: Rng + CryptoRng>(
    process: &Process<A::Network>,
    address: Address<A::Network>,
    program_id: ProgramID<A::Network>,
    function_name: Identifier<A::Network>,
    inputs: impl ExactSizeIterator<Item = impl TryInto<Value<A::Network>>>,
    rng: &mut R,
) -> Result<(MinimumCost, ExecuteCostDetails)>
```
and the necessary associated machinery. The function computes the (exact) cost of the `Execution` resulting from the given `program/address` function call by the given address on the given inputs. It does so without requiring the private key associated to the address and is therefore usable in more general situations than `execution_cost_for_authorization`.

# Design

 - `execution_cost_for_call` mocks an `Authorization` and subsequently calls on `execution_cost_for_authorization`.
 - Since the mocked `Authorization` must have the correct size (otherwise the resulting cost would be different), the exact same execution flow has to occur as in the real function call by the given (signer) address, as elements such as the execution graph and concrete input values.
 - The strawman approach of sampling a random private key and computing the resulting `Authorisation` fails for several reasons, the fundamental one being that the (signer) address can substantially affect the flow of an execution. The test case `test_cost_for_call_depending_on_signer` illustrates this.
 - Computing the correct the correct `Authorization` is impossible without the private key, which is unavailable. The proposed design follows the actual execution path without having the private key can be summarised as follows: A new variant `CallStack::AuthorizeMocked` has been added to `CallStack`. Most of the time, it behaves as `CallStack::Authorize`. However, (aside from some other minor differences) its flow samples a `Request` using the new  `Request::sample` where the `CallStack::Authorize` would call `Request::sign`. Some elements of the sampled request are correct (most crucially, the signer and input values; this is needed for the evaluation flow to succeed) whereas some are not (e.g. input IDs and the signature itself)

# Open questions for reviewers

- Would you prefer a design with only one `CallStack::Authorize` variant that includes a new flag `is_mocked: boolean` instead of keeping a separate `CallStack::AuthorizeMocked`? The code duplication with this latter approach is minimal and I think it's worth keeping the two flows well apart, but switching to the first approach would be easy and I'm happy to if there's interest.

- The `CallStack::AuthorizeMocked` flow constructs the transition using a new `Transition::from_unchecked` which does not check correctness of input IDs (which can't be done). I tried to also remove most of the output-ID computations in that function, but at least in some of the output variants this fails (for instance, when encrypting private records, the function checks the encryption key is as expected). If performance is critical, I can remove ID-computation work in some of the Output variants therein and simply mock the values (the result would be semi-mocked OutputIDs).

# Tests

- A new bespoke test in `synthesizer/src/vm/tests/test_v15/cost_for_call.rs` which tests cost-estimation correctness in a flow that depends on the signer (specifically: in the modulo-4 residue of the x coordinate of the address) in a non-trivial way. The details are explained therein.
- A check of cost-estimation correctness has been added to several of the tests in `test_v14` (including cases with input and output record-like values).
- A check of cost-estimation correctness has been added to all tests ran in `test_vm_execute_and_finalize`.


console/account/src/compute_key/mod.rs
```diff
@@ -43,6 +43,11 @@ pub struct ComputeKey<N: Network> {
 }
 
 impl<N: Network> ComputeKey<N> {
+    /// Creates a new `ComputeKey` from the given inputs without checking its correctness.
+    pub fn new_unchecked(pk_sig: Group<N>, pr_sig: Group<N>, sk_prf: Scalar<N>) -> Self {
```
vicsn: I can only imagine this is a performance optimization. At this low level IMO we should aim for safe by default APIs. Can we remove this function and you sample valid keys?

ledger/block/src/transition/mod.rs
```diff
@@ -425,6 +425,317 @@ impl<N: Network> Transition<N> {
         // Return the transition.
         Self::new(program_id, function_name, inputs, outputs, tpk, tcm, scm)
     }
+
+    /// Initializes a new transition from a request, response, and optional
+    /// dynamic outputs. It does not check correctness or consistency of the
+    /// provided values and should only be used for cost-estimation or testing
+    /// purposes.
+    pub fn from_unchecked(
```
vicsn, line 288: Can we reduce the # of lines of code of this function by either sampling more randomly, or abstracting some closures to shared functions?

synthesizer/process/src/stack/call/standard.rs
```diff
@@ -83,36 +83,58 @@ impl<N: Network> CallTrait<N> for Call<N> {
...
-                    program_checksum,
-                    false,
-                    rng,
-                )?;
-                // Add the request to the requests.
-                requests.push(request.clone());
-                // Add the request to the authorization.
-                authorization.push(request.clone())?;
+            // In Authorize (resp. AuthorizeMocked) mode, we need to compute the new request (resp. mocked request) and push it onto the call stack.
+            match &mut call_stack {
```
vicsn: Nittiest of personal nits: I think sticking to the original if let syntax may be better as we're really not trying to be exhaustive and its more similar to what we did before

synthesizer/process/src/stack/mod.rs
```diff
@@ -122,6 +122,8 @@ pub enum CallStack<N: Network> {
     Execute(Authorization<N>, Arc<RwLock<Trace<N>>>, Translations<N>),
     /// Execute a function and create the circuit assignment.
     PackageRun(Vec<Request<N>>, PrivateKey<N>, Assignments<N>),
+    /// Mock an evaluation for cost estimation.
+    AuthorizeMocked(Vec<Request<N>>, Address<N>, Authorization<N>),
```
vicsn: Nit: can this be moved after Authorize? If we keep this approach

utilities/src/rand.rs
```diff
@@ -52,6 +52,7 @@ pub trait UniformExt: Uniform {
 impl<T: Uniform> UniformExt for T {}
 
 /// A fast RNG used **solely** for testing and benchmarking, **not** for any real world purposes.
+#[derive(Clone)]
```
vicsn: @ljedrz is this a codesmell?

vicsn (changes requested):
Looks great! Few minor asks, keen to hear the other reviewer's takes

console/account/src/compute_key/mod.rs
Antonio95: It was indeed, since one needs to put a `ComputeKey` into the (dummy) `Signature` that goes into the `Request` object. I've now reverted this new method and sampled a private key and constructed the `ComputeKey` from it instead (at a small cost of two scalar-by-point multiplications). I agree introducing unchecked objects as this low level is at least questionable, thanks for the callout 👌

synthesizer/process/src/stack/call/standard.rs
Antonio95: Are you referring to replacing 
```
match &mut call_stack {
    CallStack::Authorize(requests, private_key, authorization) => { ... }
    CallStack::AuthorizeMocked(requests, address, authorization) => { ... }
    _ => {}
```
by
```
if let CallStack::Authorize(requests, private_key, authorization) = &mut call_stack {
    ...
}
if let CallStack::AuthorizeMocked(requests, address, authorization) = &mut call_stack {
    ...
}
```
?

I just made the change 👌  (lmk if you meant sth else). It's true these are "two special cases" rather than "one situation that needs to be handled differently for each of the enum variants".

Antonio95: NB: Just made an analogous change in `pub fn evaluate_function`, which had the same pattern.

synthesizer/process/src/stack/mod.rs
Antonio95: Just made the change in the def and in all the `match`es

utilities/src/rand.rs
Antonio95: I've preemptedly removed it. The only reason I introduced it is because I added the cost-estimation check, which needs an RNG, to `test_vm_execute_and_finalize`, and I was passing the same as the one used in the test cases themselves - which broke expectations. I've now sampled a separate RNG in `test_vm_execute_and_finalize` only for cost estimation and everything works without `TestRng: Clone` and without regenerating expectations.

vicsn (approved):
LGTM, looking forward to see what other reviewers think

synthesizer/process/src/cost.rs
```diff
@@ -141,6 +143,26 @@ pub fn execution_cost_for_authorization<N: Network>(
...
+pub fn execution_cost_for_call<A: Aleo, R: Rng + CryptoRng>(
+    process: &Process<A::Network>,
+    address: Address<A::Network>,
+    program_id: ProgramID<A::Network>,
+    function_name: Identifier<A::Network>,
+    inputs: impl ExactSizeIterator<Item = impl TryInto<Value<A::Network>>>,
+    consensus_version: ConsensusVersion,
+    rng: &mut R,
+) -> Result<(MinimumCost, ExecuteCostDetails)> {
+    let stack = process.get_stack(program_id).unwrap();
```
d0cd:
```suggestion
    let stack = process.get_stack(program_id)?;
```

console/program/src/request/sign.rs
```diff
@@ -191,4 +191,102 @@ impl<N: Network> Request<N> {
             is_dynamic,
         })
     }
+
+    /// Samples a `Request` with the given `signer`, `program_id`,
+    /// `function_name` and `inputs`. The fields `sk_tag`, `tvk`, `tcm`, `scm`,
+    /// `signature` and `input_ids` are random, but the size of the sampled
+    /// `Request` is the same as if it were correctly produced from the given
+    /// inputs and signed.
+    pub fn sample<R: Rng + CryptoRng>(
```
d0cd, line 200: Just a thought, `sample` makes it sound like a legitimate `Request` was created. Consider `mock`

ledger/block/src/transition/mod.rs
d0cd, line 108: An alternate design is to consider a `const` parameter that indicates whether checks are skipped or not. In this way, it is clear in all calls but allows you to dedup code.

synthesizer/process/src/stack/authorize.rs
```diff
@@ -130,6 +130,56 @@ impl<N: Network> Stack<N> {
...
+    /// Produces a mocked `Authorization` for a call to the given function on
+    /// the supplied inputs using the provided caller address. The resulting
+    /// `Authorization` has the same size as the one which would be produced
+    /// (and signed) using the private key corresponding to that address and can
+    /// therefore be used to compute the cost of the associated `Execution`, but
+    /// many of its values (such as the input IDs in the `Request`s) may not be
+    /// correct. This method does not check circuit satisfiability or `Request`
+    /// validity.
+    #[inline]
+    pub fn sample_authorization<A: circuit::Aleo<Network = N>, R: Rng + CryptoRng>(
```
d0cd, line 142: Similar comment on the connotation of "sample"

synthesizer/src/vm/tests/test_v14/mod.rs
```diff
@@ -176,3 +181,67 @@ fn add_and_test(
     // Add the next block to the VM.
     vm.add_next_block(&block).unwrap();
 }
+
+// Adds the given transactions to a new block and asserts all of them were
+// accepted, additionally checking that the cost estimations based on the
+// Authorization and the call target an inputs are correct.
+pub(crate) fn add_and_test_with_costs(
```
d0cd, line 153: If it's not too much overhead, you can just unify this with `add_and_test` to expand the tested cases.

d0cd (approved):
LGTM! Left some minor suggests for refactors and testing.
None are critical, so I'll leave it up to you

synthesizer/process/src/cost.rs
Antonio95: Thank you 👌

console/program/src/request/sign.rs
Antonio95: I actually initially named it `mock_sign`: https://github.com/ProvableHQ/snarkVM/pull/3207/changes/db32ae5f3dd6a1f3ff1c20f7c61c952af9c09a8a#diff-a270230d903d79f7b74a9c75841e5732a25d5fea1b082851ee6aa3da0dac7918L196

However, Victor suggested the `sample` naming in order to align it with other similar functionality in our blockchain (I initially pushed back a bit saying that `sample` sounded as if it would give you an "internally consistent" object, be he quickly found examples where that is not the case, e.g. https://github.com/ProvableHQ/snarkVM/blob/staging/console/program/src/data/literal/sample.rs#L37)

If you strongly prefer `mock`, just let me know and I'll change it back :)

ledger/block/src/transition/mod.rs
Antonio95: The (small) issue with that (or with passing a new `check_input_ids: bool` argument) is that it requires modifying preexisting calls. I also considered creating an internal function that receives that argument and making `Transition::from` a wrapper around that which simply passes `true`. If you still think either of those approaches is better than the current one, just let me know 👍 

(In terms of code deduplication, I did take the small step of pulling out the old `construct_output` closure common to both `from` and `from_unchecked` into a separate function, which is actually almost as long as the rest of `from_unchecked`: https://github.com/ProvableHQ/snarkVM/blob/feat/cost_without_pk/ledger/block/src/transition/mod.rs#L700)

synthesizer/process/src/stack/authorize.rs
Antonio95: As mentioned in the reply to the other comment, this was chosen to align with the terminology `sample` used throughout our codebase. Again, happy to change if you prefer `mock_authorization`.

synthesizer/src/vm/tests/test_v14/mod.rs
Antonio95: Done! in https://github.com/ProvableHQ/snarkVM/pull/3207/commits/dd3664b91ef36fab19c011c00ed0fba1c4f9ea2c. Changed `add_and_test` in all test_v14 and test_v15 tests to `add_and_test_with costs`.

PR #3211, opened 2026-04-07
Add estimate_execution_cost for fee estimation without private key

### LINKED ISSUE #3180: [Feature] Create a utility function to estimate fee without access to private key
## 🚀 Feature

We already have an `execution_cost_for_authorization`. However, to compute transitions in the Authorization, one needs access to a private key, which is not always possible (e.g. when the private key is in a hardware or MPC wallet).

To help figure out if we can compute fees regardless, some guiding questions:
- Can we create an Authorization with an ephemeral private key to run `vm.evaluate`? Even when the transitions create/consume records? This should be possible conceptually because we do something similar for `CheckDeployment`

Side notes:
- Transitions determine the size of the execution, which is needed for the cost
- We need the whole call graph to determine finalize costs

Nuel-osas:
## Summary

Adds `estimate_execution_cost` — a utility function that estimates execution cost in microcredits without requiring the caller's private key.

This enables fee estimation for:
- Hardware wallets (Ledger, Trezor) where the private key cannot be extracted
- MPC wallets where key shares are distributed across parties
- Any context where fee preview is needed before committing to a signing operation

## Approach

Generates an ephemeral `PrivateKey` and calls `authorize_unchecked` to produce a valid `Authorization` with transitions. Passes the authorization to the existing `execution_cost_for_authorization` for cost computation.

This follows the same ephemeral-key pattern used by `CheckDeployment` in `stack/deploy.rs` and `synthesize_key` in `stack/helpers/synthesize.rs`.

## Changes

- `synthesizer/process/src/cost.rs` — added `estimate_execution_cost` function
- Updated imports to include `PrivateKey`, `ProgramID`, `Value`, `circuit`

## Validation

```bash
cargo check -p snarkvm-synthesizer-process  # passes
```

Closes #3180

### DIFF
--- a/synthesizer/process/src/cost.rs
+++ b/synthesizer/process/src/cost.rs
@@ -15,11 +15,12 @@
 
 use std::collections::HashMap;
 
-use crate::{Authorization, FinalizeTypes, Process, Stack, StackRef, StackTrait};
+use crate::{circuit, Authorization, FinalizeTypes, Process, Stack, StackRef, StackTrait};
 
 use console::{
+    account::PrivateKey,
     prelude::*,
-    program::{FinalizeType, Identifier, LiteralType, PlaintextType},
+    program::{FinalizeType, Identifier, LiteralType, PlaintextType, ProgramID, Value},
 };
 use snarkvm_algorithms::snark::varuna::VarunaVersion;
 use snarkvm_ledger_block::{Deployment, Execution, Transaction};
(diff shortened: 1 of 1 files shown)


vicsn:
Conflicts with: https://github.com/ProvableHQ/snarkVM/pull/3207

Nuel-osas:
Thanks for the heads up @vicsn. Just looked at #3207 — that's a much more thorough approach (modifying `Request::sign` and the evaluation pipeline to support address-only authorization, plus proper record handling). Our PR was a lighter wrapper using an ephemeral key, which wouldn't handle record inputs.

Happy to close this in favor of #3207. Is there anything on that PR where an extra pair of eyes or testing help would be useful?

PR #3213, opened 2026-04-10
Fix missing request features for standard builds

vicsn:
## Motivation

CC @fedpre @iamalwaysuncomfortable 

## Test Plan

Tested locally.


ljedrz (approved):
Sneaky 👍.

PR #3214, opened 2026-04-10
[Feat] Tx rejection reasons (take 3)

ljedrz:
A successor to https://github.com/ProvableHQ/snarkVM/pull/3101 and https://github.com/ProvableHQ/snarkVM/pull/3159, this PR applies the architecture from the latter, and the storage from the former.

The goal is to have the transaction rejection reasons persisted in the storage, and not propagated inside blocks. This is somewhat complicated (at least with the current APIs), as transactions are rejected during speculation, at which point the rejection reasons go into the block template, and are available upon finalization (where actual persistent storage takes place).

The question is where to strip the rejection reasons from the block, or if it would be acceptable to adjust the APIs/objects just to avoid the reasons from _ever_ going into blocks.

Also @raychu86 am I correct in assuming that consensus versions are not a concern if the blocks don't carry transaction rejection reasons (meaning their serialization is unaffected)?


ljedrz:
The failing CI jobs would be fixed with https://github.com/ProvableHQ/snarkVM/pull/3196.

synthesizer/src/vm/finalize.rs
```diff
@@ -285,6 +285,9 @@ impl<N: Network, C: ConsensusStorage<N>> VM<N, C> {
         // Determine the maximum number of aborted transactions allowed in a block.
         let max_aborted_transactions = Transactions::<N>::max_aborted_transactions();
 
+        // Clear out any pending rejection reasons in case of errors in the previous iteration.
+        self.pending_rejected_reasons.write().clear();
```
vicsn, line 289: Can we log an ERROR if this happens?

synthesizer/src/vm/mod.rs
```diff
@@ -134,6 +134,8 @@ pub struct VM<N: Network, C: ConsensusStorage<N>> {
     partially_verified_transactions: Arc<RwLock<LruCache<TransactionCacheKey<N>, N::TransmissionChecksum>>>,
     /// The restrictions list.
     restrictions: Restrictions<N>,
+    /// The list of rejection reasons for pending confirmed transactions.
```
vicsn, line 137: /// The list of rejection reasons for pending confirmed transactions.
    /// TODO: it would be cleaner if these are passed along as an argument to `add_next_block`, but this requires a bigger refactor.

vicsn (changes requested):
After a lot of discussions we had, LGTM architecturally

ljedrz:
@vicsn I'll address these and any subsequent review comments once the aforementioned https://github.com/ProvableHQ/snarkVM/pull/3196 is merged.

vicsn:
> @vicsn I'll address these and any subsequent review comments once the aforementioned https://github.com/ProvableHQ/snarkVM/pull/3196 is merged.

The linked PR is now merged

ledger/block/src/transactions/rejected_reason/mod.rs
```diff
@@ -0,0 +1,83 @@
...
+    DuplicateProgramID(ProgramID<N>),
+
+    /// The transaction was rejected due to a failed finalize command. (program ID, edition, resource, index, command).
+    /// Note: We do not log the actual error message from the finalize command, as it may contain
+    /// sensitive information or lead to DOS vectors by storing string representations of large structs.
+    Finalize(ProgramID<N>, u16, Identifier<N>, usize, Box<Command<N>>),
+
+    /// The transaction was rejected due to a VM error not captured by a finalize command.
+    /// The programID and resource are logged if they are available.
+    NonFinalize(Option<(ProgramID<N>, u16)>, Option<Identifier<N>>),
```
vicsn: `VM` is probably a better name

vicsn (commented):
Generally LGTM.

@dgrkotsonis I wonder whether we should guard this behind the `history` feature so it's very explicit that this data is only available once a node operator indexes historic data, just like the other history data. What do you think? The downside is that now most local setups which want to use this feature for testing also need to build their nodes with history enabled.

ledger/block/src/lib.rs
```diff
@@ -145,7 +145,7 @@ impl<N: Network> Block<N> {
 
     /// Initializes a new block from the given previous block hash, block header, authority,
     /// ratifications, solutions, aborted solution IDs, transactions, and aborted transaction IDs.
-    pub fn from(
+    fn from(
```
mohammadfawaz, line 148: Is narrowing the visibility here intentional? This is considered a breaking change.

synthesizer/process/src/finalize.rs
```diff
@@ -170,34 +195,51 @@ impl<'a, N: Network> ProcessExclusiveGuard<'a, N> {
...
+            None::<(usize, Command<N>)>,
+        )?;
+        // if minimum_number_of_calls != execution.len() {
+        //     indexed_finalize_bail!(
+        //         Some((transition_program_id, *stack.program_edition())),
+        //         Some(transition_function_name),
+        //         "The number of transitions in the execution is incorrect. Expected {number_of_calls}, but found {}",
+        //         execution.len()
+        //     );
+        // }
```
mohammadfawaz: should this be deleted?

synthesizer/src/vm/finalize.rs
```diff
@@ -843,6 +890,12 @@ impl<N: Network, C: ConsensusStorage<N>> VM<N, C> {
                                         "Mismatch in finalize operations for a rejected execute - (found: {finalize_operations:?}, expected: {finalize:?})"
                                     ));
                                 }
+
+                                if let Some(rejected_reason) = self.pending_rejected_reasons.write().remove(fee_tx_id) {
+                                    store.insert_rejected_reason(**fee_tx_id, rejected_reason.clone()).map_err(
+                                        |_| "Couldn't store the reason behind a rejected deployment".to_string(),
```
mohammadfawaz: should this say "...behind a rejected execution" instead?

ledger/block/src/transactions/rejected/serialize.rs
```diff
@@ -21,13 +21,15 @@ impl<N: Network> Serialize for Rejected<N> {
...
                 Self::Deployment(program_owner, deployment) => {
+                    // Determine the number of fields based on whether the rejected reason is present.
                     let mut object = serializer.serialize_struct("Rejected", 3)?;
                     object.serialize_field("type", "deployment")?;
                     object.serialize_field("program_owner", program_owner)?;
                     object.serialize_field("deployment", deployment)?;
                     object.end()
                 }
                 Self::Execution(execution) => {
+                    // Determine the number of fields based on whether the rejected reason is present.
```
mohammadfawaz: Are the two comments here still relevant?

synthesizer/src/vm/finalize.rs
mohammadfawaz: `bond_validator` instead of `bond_public`?

ledger/block/src/transactions/rejected_reason/mod.rs
```diff
@@ -0,0 +1,83 @@
...
+/// The reason a transaction was rejected.
+#[derive(Clone, PartialEq, Eq)]
+pub enum RejectedReason<N: Network> {
+    /// The transaction was rejected due to a duplicate program ID deployment in the same block.
+    DuplicateProgramID(ProgramID<N>),
+
+    /// The transaction was rejected due to a failed finalize command. (program ID, edition, resource, index, command).
+    /// Note: We do not log the actual error message from the finalize command, as it may contain
+    /// sensitive information or lead to DOS vectors by storing string representations of large structs.
+    Finalize(ProgramID<N>, u16, Identifier<N>, usize, Box<Command<N>>),
```
mohammadfawaz: a 5-element tuple might be hard to parse. Is it maybe cleaner to promote to a struct instead?

synthesizer/src/vm/finalize.rs
```diff
@@ -3661,4 +3741,91 @@ finalize compute:
...
+        let rejected_transaction = next_block.transactions().iter().next().unwrap();
+        assert!(rejected_transaction.is_rejected());
+
+        // Add the next block to the VM.
+        vm.add_next_block(&next_block).unwrap();
+
+        // Check that the rejection reason was stored.
+        let tx_id = *rejected_transaction.id();
+        let rejection_reason = vm.finalize_store().get_rejected_reason(&tx_id).unwrap();
+        assert!(rejection_reason.is_some(), "Rejection reason should be stored");
```
mohammadfawaz, line 3827: Should we assert more stuff here? (assert the variant is `Finalize`, etc.)

synthesizer/src/vm/finalize.rs
```diff
@@ -403,34 +413,49 @@ impl<N: Network, C: ConsensusStorage<N>> VM<N, C> {
...
                                     .map(|(fee_tx, finalize)| {
                                         let rejected = Rejected::new_deployment(*program_owner, deployment);
                                         ConfirmedTransaction::rejected_deploy(counter, fee_tx, rejected, finalize)
+                                            .and_then(|confirmed_tx| {
+                                                // Store the rejection reason.
+                                                self.pending_rejected_reasons
+                                                    .write()
+                                                    .insert(confirmed_tx.id(), rejected_reason.clone());
+                                                store
+                                                    .insert_rejected_reason(*confirmed_tx.id(), rejected_reason)
```
mohammadfawaz, line 434: Since we're inside `atomic_finalize!(self.finalize_store(), FinalizeMode::DryRun, { .. })`, does the `DryRun` imply that this is a no-op?

synthesizer/src/vm/finalize.rs
```diff
@@ -475,19 +502,33 @@ impl<N: Network, C: ConsensusStorage<N>> VM<N, C> {
...
                                                 ConfirmedTransaction::rejected_execute(
                                                     counter, fee_tx, rejected, finalize,
                                                 )
+                                                .and_then(|confirmed_tx| {
+                                                    // Store the rejection reason.
+                                                    self.pending_rejected_reasons
+                                                        .write()
+                                                        .insert(confirmed_tx.id(), rejected_reason.clone());
+                                                    store
+                                                        .insert_rejected_reason(*confirmed_tx.id(), rejected_reason)
```
mohammadfawaz, line 528: Similarly, since we're inside `atomic_finalize!(self.finalize_store(), FinalizeMode::DryRun, { .. })`, does the `DryRun` imply that this is a no-op?

mohammadfawaz (commented):
Overall looks good! Just a few questions/nits.

ledger/block/src/lib.rs
ljedrz: It was just not used elsewhere (and I'd argue that it shouldn't be, otherwise one could construct a block with an invalid hash), so I considered it an API tightening. You're right about the breaking change, but we're not strictly adhering to semver, and it's a very niche function.

vicsn: Yeah it's probably fine, but to be sure @ljedrz can you draft a companion snarkOS PR?

ljedrz: There's nothing to do on snarkOS side other than update the SHA, I just checked locally.

synthesizer/process/src/finalize.rs
ljedrz: Ah, I remember now - it's something that has changed since Ray's implementation - there is no longer a `number_of_calls` variable, there's a `minimum_number_of_calls` instead, and I don't know what this means for the initial implementation. @raychu86 can you please advise?

synthesizer/src/vm/finalize.rs
ljedrz: Out of scope for this PR, but I can include it in my update.

synthesizer/src/vm/finalize.rs
ljedrz: Database-wise it is technically a noop; not sure about the other implications.

ledger/block/src/transactions/rejected_reason/mod.rs
ljedrz: It doesn't seem like it currently provides a lot more clarity in the code (I just tried it out), but it's a good idea on principle.

synthesizer/process/src/finalize.rs
vicsn: Because of dynamic dispatch, we cannot always statically determine the number of calls anymore.

The commented out code here is deprecated in favour of the two errors right below here, which you can try changing to indexed_finalize_bail

ljedrz: I only now saw the edit; will try it.

ljedrz: Yeah, worked perfectly fine: https://github.com/ProvableHQ/snarkVM/pull/3214/commits/b38316bb12aeef12861b9abd6d466c898aa65279.

vicsn:
One more request: can we ensure that we write rejection reasons only 1. If the `history` feature is enabled or 2. if `history` is not enabled then only write them from the latest `ConsensusVersion::V15` onwards? This may be paranoid, but this ensures that validator databases/environments are more consistent, making debugging easier.

ledger/block/src/transactions/rejected_reason/mod.rs
ljedrz: https://github.com/ProvableHQ/snarkVM/pull/3214/commits/f8f096f1b8959f99f84cbee68c7e5aaf3eaf6aa5

ljedrz:
@vicsn Done in https://github.com/ProvableHQ/snarkVM/pull/3214/commits/267c7470cd51f6c76d35f820b2128f70e585a290.

vicsn (approved):
LGTM again

mohammadfawaz (approved):
LGTM!

PR #3215, opened 2026-04-10
Log solution ID when post-ratify fails

kaimast:


dgrkotsonis (approved):

ledger/puzzle/src/lib.rs
```diff
@@ -128,10 +130,18 @@ impl<N: Network> Puzzle<N> {
 
     /// Returns the proof target given the solution.
     pub fn get_proof_target(&self, solution: &Solution<N>) -> Result<u64> {
+        let address = solution.address();
+
         // Calculate the proof target.
-        let proof_target = self.get_proof_target_unchecked(solution)?;
+        let proof_target = self
+            .get_proof_target_unchecked(solution)
+            .with_context(|| format!("Failed to get proof target for solution {address}"))?;
```
vicsn: Below we log id, elsewhere address, is that intended?

ledger/puzzle/src/lib.rs
```diff
@@ -128,10 +130,18 @@ impl<N: Network> Puzzle<N> {
...
         // Ensure the proof target matches the expected proof target.
-        ensure!(solution.target() == proof_target, "The proof target does not match the expected proof target");
+        ensure_equals!(
+            solution.target(),
+            proof_target,
+            "The proof target for solution {address} match the expected proof target"
```
raychu86: This is incorrect, should be solution ID instead of `address` here

ledger/puzzle/src/lib.rs
kaimast: It was not intentional but is fixed now.

ledger/puzzle/src/lib.rs
kaimast: Fixed

raychu86 (approved):

vicsn (approved):

PR #3217, opened 2026-04-13
Make parameter downloads more robust

vicsn:
## Motivation

A post-merge CI jobs recently failed to download parameters, which I could reproduce it locally. Seems that simple retries were not enabled. I first implemented a simple retry loop, but then I read this painful comment: ["It’s also a very common pattern to implement manually, and far too easy to do incorrectly."](https://seanmonstar.com/blog/reqwest-retries/). So going for the "proper" reqwest approach.

## Test Plan

Tried locally and enabling post-merge CI.

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -1279,8 +1279,7 @@ workflows:
       or pipeline.git.branch == "canary"
       or pipeline.git.branch == "testnet"
       or pipeline.git.branch == "mainnet"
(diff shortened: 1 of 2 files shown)


ljedrz (approved):
LGTM, the CI errors seem unrelated.

Note: it would be even more robust if we could create the `Client` just once and reuse it across the fetches, but that's outside of this scope.

PR #3218, opened 2026-04-14
feat: varuna performance optimisations

davencyw:
## Motivation

The following performance improvements were applied.

#### 1. Cross-product precomputation in the fourth round (`fourth.rs`)

The expressions `(α − row[i]) · (β − col[i])` were computed independently for the `b_poly` evaluations and the `f` inverses. This hoists the product into a single pass over `row_on_K`/`col_on_K` and reuses the resulting vector for both consumers, eliminating K redundant field multiplications per matrix.

#### 2. Closed-form domain summation (`prepare_third.rs`, `third.rs`)

Computing `∑_{h∈H} p(h)` via `evaluate_over_domain_by_ref` required a full O(n log n) FFT followed by a linear sum. For a polynomial of degree < 2n evaluated over a domain of size n, the identity `∑_{h∈H} h^k = n` iff `n | k` (and 0 otherwise) means the sum collapses to `n · (c₀ + cₙ)`, a constant-time read of two coefficients. Applied in both the third-round preparation and the third-round lineval computation.

#### 3. Equal-domain fast path in `apply_randomized_selector` (`selectors.rs`)

When source and target domains have the same size (true for every single-circuit proof), the selector polynomial is the identity. The existing code would multiply by the vanishing polynomial of the target domain and then divide by the vanishing polynomial of the source domain, an expensive round-trip that cancels exactly. The new fast path detects `src.size == target.size`, scales the polynomial by `combiner` directly, and proceeds to the division, avoiding the unnecessary multiply step entirely.

### Correctness

```
cargo test -p snarkvm-algorithms -- varuna
test result: ok. 22 passed; 0 failed; 0 ignored; 0 measured; 62 filtered out; finished in 204.07s
```

### Benchmark Results (`credits.aleo`)

| Function | Baseline | After | Δ (median) |
|---|---|---|---|
| `transfer_public` | 347.58 ms | 329.90 ms | **−4.9 %** |
| `transfer_private` | 2.9346 s | 2.8005 s | **−4.6 %** |
| `transfer_public_to_private` | 569.13 ms | 533.04 ms | **−6.3 %** |
| `transfer_private_to_public` | 2.9016 s | 2.7622 s | **−4.8 %** |
| `join` | 3.3261 s | 3.1319 s | **−5.8 %** |
| `split` | 2.8807 s | 2.7583 s | **−4.2 %** |

## Documentation
No documentation changes needed.

## Backwards compatibility
These changes are fully backwards compatible.

### DIFF
--- a/algorithms/src/snark/varuna/ahp/prover/round_functions/fourth.rs
+++ b/algorithms/src/snark/varuna/ahp/prover/round_functions/fourth.rs
@@ -168,6 +168,15 @@ impl<F: PrimeField, SM: SNARKMode> AHPForR1CS<F, SM> {
         let R_size = constraint_domain.size_as_field_element;
         let C_size = variable_domain.size_as_field_element;
 
+        // Precompute (alpha - row[i]) * (beta - col[i]) once and reuse for both b_poly evals
+        // and f inverses, saving K redundant field multiplications per matrix.
+        let cross_products: Vec<F> = row_on_K
+            .evaluations
+            .iter()
+            .zip_eq(&col_on_K.evaluations)
+            .map(|(&r, &c)| (alpha - r) * (beta - c))
+            .collect();
+
         let mut job_pool = snarkvm_utilities::ExecutionPool::with_capacity(2);
         job_pool.add_job(|| {
             let a_poly_time = start_timer!(|| format!("Computing a poly for {label}"));
@@ -182,14 +191,8 @@ impl<F: PrimeField, SM: SNARKMode> AHPForR1CS<F, SM> {
 
         job_pool.add_job(|| {
             let b_poly_time = start_timer!(|| format!("Computing b poly for {label}"));
-            let alpha_beta = alpha * beta;
             let b_poly = {
-                let evals: Vec<F> = row_on_K
-                    .evaluations
-                    .iter()
-                    .zip_eq(&col_on_K.evaluations)
-                    .map(|(&r, &c)| R_size * C_size * (alpha_beta - beta * r - alpha * c + r * c))
-                    .collect();
+                let evals: Vec<F> = cross_products.iter().map(|&cp| R_size * C_size * cp).collect();
                 EvaluationsOnDomain::from_vec_and_domain(evals, non_zero_domain)
                     .interpolate_with_pc(ifft_precomputation)
             };
@@ -199,8 +202,8 @@ impl<F: PrimeField, SM: SNARKMode> AHPForR1CS<F, SM> {
         let [a_poly, b_poly]: [_; 2] = job_pool.execute_all().try_into().unwrap();
 
         let f_evals_time = start_timer!(|| format!("Computing f evals on K for {label}"));
-        let mut inverses: Vec<_> =
-            row_on_K.evaluations.iter().zip_eq(&col_on_K.evaluations).map(|(r, c)| (alpha - r) * (beta - c)).collect();
+        // Move cross_products directly into inverses — no second pass over row/col needed.
+        let mut inverses = cross_products;
 
         let matrix_sumcheck_constants = v_R_i_alpha_v_C_i_beta * constraint_domain.size_inv * variable_domain.size_inv;
         batch_inversion_and_mul(&mut inverses, &matrix_sumcheck_constants);
--- a/algorithms/src/snark/varuna/ahp/prover/round_functions/fourth.rs
+++ b/algorithms/src/snark/varuna/ahp/prover/round_functions/fourth.rs
@@ -167,8 +167,9 @@ impl<F: PrimeField, SM: SNARKMode> AHPForR1CS<F, SM> {
             (&arithmetization.row, &arithmetization.col, &arithmetization.row_col_val);
         let R_size_C_size = constraint_domain.size_as_field_element * variable_domain.size_as_field_element;
 
-        // Precompute (alpha - row[i]) * (beta - col[i]) once and reuse for both b_poly evals
-        // and f inverses, saving K redundant field multiplications per matrix.
+        // Precompute (alpha - row[i]) * (beta - col[i]) once and reuse for both b_poly
+        // evals and f inverses, saving K redundant field multiplications per
+        // matrix.
         let cross_products: Vec<F> = row_on_K
             .evaluations
             .iter()
@@ -201,7 +202,8 @@ impl<F: PrimeField, SM: SNARKMode> AHPForR1CS<F, SM> {
         let [a_poly, b_poly]: [_; 2] = job_pool.execute_all().try_into().unwrap();
 
         let f_evals_time = start_timer!(|| format!("Computing f evals on K for {label}"));
-        // Move cross_products directly into inverses - no second pass over row/col needed.
+        // Move cross_products directly into inverses - no second pass over row/col
+        // needed.
         let mut inverses = cross_products;
 
         let matrix_sumcheck_constants = v_R_i_alpha_v_C_i_beta * constraint_domain.size_inv * variable_domain.size_inv;
--- a/algorithms/src/snark/varuna/ahp/prover/round_functions/fourth.rs
+++ b/algorithms/src/snark/varuna/ahp/prover/round_functions/fourth.rs
@@ -165,8 +165,7 @@ impl<F: PrimeField, SM: SNARKMode> AHPForR1CS<F, SM> {
     ) -> Result<(Sum<F>, Lhs<F>, Gpoly<F>, Apoly<F>, Bpoly<F>)> {
         let (row_on_K, col_on_K, row_col_val) =
             (&arithmetization.row, &arithmetization.col, &arithmetization.row_col_val);
-        let R_size = constraint_domain.size_as_field_element;
-        let C_size = variable_domain.size_as_field_element;
+        let R_size_C_size = constraint_domain.size_as_field_element * variable_domain.size_as_field_element;
 
         // Precompute (alpha - row[i]) * (beta - col[i]) once and reuse for both b_poly evals
         // and f inverses, saving K redundant field multiplications per matrix.
@@ -192,7 +191,7 @@ impl<F: PrimeField, SM: SNARKMode> AHPForR1CS<F, SM> {
         job_pool.add_job(|| {
             let b_poly_time = start_timer!(|| format!("Computing b poly for {label}"));
             let b_poly = {
-                let evals: Vec<F> = cross_products.iter().map(|&cp| R_size * C_size * cp).collect();
+                let evals: Vec<F> = cross_products.iter().map(|&cp| R_size_C_size * cp).collect();
                 EvaluationsOnDomain::from_vec_and_domain(evals, non_zero_domain)
                     .interpolate_with_pc(ifft_precomputation)
             };
@@ -202,7 +201,7 @@ impl<F: PrimeField, SM: SNARKMode> AHPForR1CS<F, SM> {
         let [a_poly, b_poly]: [_; 2] = job_pool.execute_all().try_into().unwrap();
 
         let f_evals_time = start_timer!(|| format!("Computing f evals on K for {label}"));
-        // Move cross_products directly into inverses — no second pass over row/col needed.
+        // Move cross_products directly into inverses - no second pass over row/col needed.
         let mut inverses = cross_products;
 
         let matrix_sumcheck_constants = v_R_i_alpha_v_C_i_beta * constraint_domain.size_inv * variable_domain.size_inv;
--- a/algorithms/src/snark/varuna/ahp/prover/round_functions/prepare_third.rs
+++ b/algorithms/src/snark/varuna/ahp/prover/round_functions/prepare_third.rs
@@ -132,11 +132,12 @@ impl<F: PrimeField, SM: SNARKMode> AHPForR1CS<F, SM> {
                             matrix_transpose,
                             *alpha,
                         )?;
-                        let sum = z_m_at_alpha
-                            .evaluate_over_domain_by_ref(circuit_specific_state.variable_domain)
-                            .evaluations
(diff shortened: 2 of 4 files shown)


algorithms/src/snark/varuna/ahp/prover/round_functions/fourth.rs
```diff
@@ -168,6 +168,15 @@ impl<F: PrimeField, SM: SNARKMode> AHPForR1CS<F, SM> {
         let R_size = constraint_domain.size_as_field_element;
         let C_size = variable_domain.size_as_field_element;
 
+        // Precompute (alpha - row[i]) * (beta - col[i]) once and reuse for both b_poly evals
+        // and f inverses, saving K redundant field multiplications per matrix.
+        let cross_products: Vec<F> = row_on_K
```
Antonio95, line 173: Correct (this was computed twice before, and one of them in an even less efficient form:` alpha * beta - r * beta - c * alpha + r * c`)

algorithms/src/snark/varuna/ahp/prover/round_functions/fourth.rs
```diff
@@ -182,14 +191,8 @@ impl<F: PrimeField, SM: SNARKMode> AHPForR1CS<F, SM> {
...
             let b_poly_time = start_timer!(|| format!("Computing b poly for {label}"));
-            let alpha_beta = alpha * beta;
             let b_poly = {
-                let evals: Vec<F> = row_on_K
-                    .evaluations
-                    .iter()
-                    .zip_eq(&col_on_K.evaluations)
-                    .map(|(&r, &c)| R_size * C_size * (alpha_beta - beta * r - alpha * c + r * c))
-                    .collect();
+                let evals: Vec<F> = cross_products.iter().map(|&cp| R_size * C_size * cp).collect();
```
Antonio95: Probably irrelevant as the compiler very likely optimises this away, but could you move the  `R_size * C_size` computation to outside the loop? I would even suggest simply defining `let R_size_C_size = constraint_domain.size_as_field_element * variable_domain.size_as_field_element;` at the start, since the individual `R_size` and `C_size`  aren't used anywhere else.

algorithms/src/snark/varuna/ahp/prover/round_functions/fourth.rs
```diff
@@ -199,8 +202,8 @@ impl<F: PrimeField, SM: SNARKMode> AHPForR1CS<F, SM> {
         let [a_poly, b_poly]: [_; 2] = job_pool.execute_all().try_into().unwrap();
 
         let f_evals_time = start_timer!(|| format!("Computing f evals on K for {label}"));
-        let mut inverses: Vec<_> =
-            row_on_K.evaluations.iter().zip_eq(&col_on_K.evaluations).map(|(r, c)| (alpha - r) * (beta - c)).collect();
+        // Move cross_products directly into inverses — no second pass over row/col needed.
```
Antonio95:
```suggestion
        // Move cross_products directly into inverses - no second pass over row/col needed.
```
changed the non-ascii m-dash

algorithms/src/snark/varuna/ahp/prover/round_functions/prepare_third.rs
```diff
@@ -132,11 +132,12 @@ impl<F: PrimeField, SM: SNARKMode> AHPForR1CS<F, SM> {
                             matrix_transpose,
                             *alpha,
                         )?;
-                        let sum = z_m_at_alpha
-                            .evaluate_over_domain_by_ref(circuit_specific_state.variable_domain)
-                            .evaluations
-                            .into_iter()
-                            .sum::<F>();
+                        // sum_{h∈H} p(h) = n*(c_0 + c_n) for deg(p) < 2n: sum_{h∈H} h^k = n iff n|k, else 0,
```
Antonio95:
```suggestion
                        // sum_{h in H} f(h) = n*(c_0 + c_n) for deg(p) < 2n, where c_0 and c_n are the coeffs of f at degree 0 and n, resp.
```

algorithms/src/snark/varuna/ahp/prover/round_functions/prepare_third.rs
```diff
@@ -132,11 +132,12 @@ impl<F: PrimeField, SM: SNARKMode> AHPForR1CS<F, SM> {
...
+                        // and in [0, 2n-2] only k=0 and k=n are multiples of n. Avoids an O(n log n) FFT.
```
Antonio95:
```suggestion
```

algorithms/src/snark/varuna/ahp/prover/round_functions/prepare_third.rs
```diff
@@ -132,11 +132,12 @@ impl<F: PrimeField, SM: SNARKMode> AHPForR1CS<F, SM> {
...
+                        let n = circuit_specific_state.variable_domain.size_as_field_element;
+                        let c_0 = z_m_at_alpha.coeffs.first().copied().unwrap_or_default();
+                        let c_n = z_m_at_alpha.coeffs.get(circuit_specific_state.variable_domain.size()).copied().unwrap_or_default();
+                        let sum = n * (c_0 + c_n);
```
Antonio95, line 154: This change is also correct. In general, given a finite subgroup `H` of order `n = |H|` of the group of units of a field and a polynomial `f` of degree `< k * n`  over that field (for some `k`), the sum `sum_{h in H} f(h)` equals `|H| * (c_0 + ... + c_(n * (k - 1))`, where `c_i` is the coeff of `x^i` in `f`. In this case, `z_m_at_alpha` is the product of `M(alpha, X)` and `z_{i, j}(X)`, each of which has `deg <= |variable_domain|`; therefore only `c_0` and `c_|variable_domain|` need to be added.

Pretty cool find and optimization :)

algorithms/src/snark/varuna/ahp/prover/round_functions/third.rs
```diff
@@ -389,7 +389,12 @@ impl<F: PrimeField, SM: SNARKMode> AHPForR1CS<F, SM> {
         z_m_at_alpha: Option<DensePolynomial<F>>,
     ) -> Result<LinevalInstance<F>> {
         let mut z_m_at_alpha = z_m_at_alpha.ok_or(anyhow::anyhow!(format!("Expected z_{_matrix_label}_at_alpha")))?;
-        let sum = z_m_at_alpha.evaluate_over_domain_by_ref(*variable_domain).evaluations.into_iter().sum::<F>();
+        // sum_{h∈H} p(h) = n*(c_0 + c_n) for deg(p) < 2n: sum_{h∈H} h^k = n iff n|k, else 0,
```
Antonio95:
```suggestion
        // sum_{h in H} f(h) = n*(c_0 + c_n) for deg(p) < 2n, where c_0 and c_n are the coeffs of f at degree 0 and n, resp.
```

algorithms/src/snark/varuna/ahp/prover/round_functions/third.rs
```diff
@@ -389,7 +389,12 @@ impl<F: PrimeField, SM: SNARKMode> AHPForR1CS<F, SM> {
...
+        // and in [0, 2n-2] only k=0 and k=n are multiples of n. Avoids an O(n log n) FFT.
```
Antonio95:
```suggestion
```

algorithms/src/snark/varuna/ahp/selectors.rs
```diff
@@ -116,6 +116,17 @@ pub(crate) fn apply_randomized_selector<F: PrimeField>(
...
+        // The multiplier simplifies to `combiner` (size/size = 1), and the subsequent
+        // mul_by_vanishing_poly(target) + divide_by_vanishing_poly(src) round-trip
+        // cancels exactly since v_target == v_src. This fires on every single-circuit proof.
+        if src_domain.size == target_domain.size {
+            poly.coeffs.iter_mut().for_each(|c| *c *= combiner);
+            let (h_i, xg_i) = poly.divide_by_vanishing_poly(*src_domain)?;
+            end_timer!(selector_time);
+            return Ok((h_i, Some(xg_i)));
+        }
+
```
Antonio95, line 124: Correct. I don't love the early-return structure, I think the following is a bit clearer:

```
        // Substituting in s_i, we get that:
        // \sum_i{poly_i}/v_H = \sum{h_i*v_H + x_g_i}
        // \sum_i{c_i*s_i*(poly_i/v_H - x_g_i)} = \sum{h_i*v_H}
        // \sum_i{c_i*(H_i.size()/H.size())*(poly_i/v_H_i - x_g_i*v_H/v_H_i)} =
        // \sum{h_i*v_H} \sum_i{c_i*(H_i.size()/H.size())*(poly_i/v_H_i} =
        // \sum{h_i*v_H} + \sum{c_i*x_g_i*(v_H/v_H_i)*(H_i.size()/H.size())}
        // (\sum_i{c_i*s_i*poly_i})/v_H = \sum{h_i*v_H} + \sum{c_i*s_i*x_g_i}
        // (\sum_i{c_i*s_i*poly_i})/v_H = h_1*v_H + x_g_1
        // That's what we're computing here.
        let selector_time = start_timer!(|| "Compute selector with remainder witness");

        let multiplier = if src_domain.size == target_domain.size { 
            combiner
        } else {
            combiner * src_domain.size_as_field_element * target_domain.size_inv
        };

        // Computing xg_i * s_i with s_i = (v_H / v_H_i) * (H_i.size() /
        // H.size()) without the last constant, which was already incorporated
        // into the multiplier. If the two domains are equal, s_i = 1 and we
        // skip this step.
        let updated_xg_i = if src_domain.size == target_domain.size {
            xg_i
        } else {
            xg_i = xg_i.mul_by_vanishing_poly(*target_domain);

            let (xg_i, remainder) = xg_i.divide_by_vanishing_poly(*src_domain)?;
            ensure!(
                remainder.is_zero(),
                "[Returning remainder witness] Failed to divide by vanishing polynomial - non-zero remainder ({remainder:?})"
            );

            xg_i
        };

        end_timer!(selector_time);
        Ok((h_i, Some(updated_xg_i)))
```

However, the functionality is exactly the same and I'm more than okay if you leave the original version (if you do change it, plz re-run the tests, I have not).

Antonio95 (approved):
Left very minor comments

algorithms/src/snark/varuna/ahp/prover/round_functions/fourth.rs
davencyw: Oh yes, good point. Fixed.

algorithms/src/snark/varuna/ahp/prover/round_functions/prepare_third.rs
davencyw: Thanks, fixed!

algorithms/src/snark/varuna/ahp/prover/round_functions/prepare_third.rs
davencyw: Thanks, fixed!

algorithms/src/snark/varuna/ahp/selectors.rs
davencyw: I like your implementation! Adjusted to yours now.

Antonio95 (approved):
Good to go! Thanks for implementing the suggestions 🙌

algorithms/src/snark/varuna/ahp/prover/round_functions/prepare_third.rs
```diff
@@ -132,11 +132,17 @@ impl<F: PrimeField, SM: SNARKMode> AHPForR1CS<F, SM> {
...
-                        let sum = z_m_at_alpha
-                            .evaluate_over_domain_by_ref(circuit_specific_state.variable_domain)
-                            .evaluations
-                            .into_iter()
-                            .sum::<F>();
+                        // sum_{h in H} f(h) = n*(c_0 + c_n) for deg(p) < 2n, where c_0 and c_n are the
+                        // coeffs of f at degree 0 and n, resp. and in [0, 2n-2]
+                        // only k=0 and k=n are multiples of n. Avoids an O(n log n) FFT.
+                        let n = circuit_specific_state.variable_domain.size_as_field_element;
+                        let c_0 = z_m_at_alpha.coeffs.first().copied().unwrap_or_default();
```
vicsn: Can we fail loudly and early? No defaults, just returning errors. We don't do this perrectly everywhere but better late than never :)

vicsn (changes requested):

algorithms/src/snark/varuna/ahp/prover/round_functions/prepare_third.rs
Antonio95: Hmm this is not a fail case, but the case where `z_m_at_alpha` is the zero polynomial. The unwrap or default makes sure we read its contstant term as zero when `z_m_at_alpha.coeffs` is empty (this is not an instance of: "things have failed so give me a default value and I'll sort things out").

davencyw: Good call, fixed now!

davencyw:
Thanks @vicsn for the pointer, let me know if there's more or if it looks good to you!

algorithms/src/snark/varuna/ahp/prover/round_functions/prepare_third.rs
vicsn: @Antonio95 makes sense, thx for clarifying

> Good call, fixed now!

@davencyw do you disagree with @Antonio95 ?

Antonio95:
The two modifications of `unwrap_or_default` should be undone: the first one covered the zero-polynomial case and the second one covers the "polynomial of degree < |variable domain|" case (which are valid)

davencyw:
Undone!

vicsn (approved):

PR #3221, opened 2026-04-15
Fix target adjustment tests

vicsn:
## Motivation

One test failed in the `merge-workflow`. The fix is consistent with my understanding of the impacts of the new anchor time: halvings will be reached slower but doublings will be reached faster.

And if I recall correctly, @ljedrz ' simulations indicated that with fast blocktimes, the target will not keep increasing but will settle on some higher steady state. @raychu86 is this in line with your expectations?

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -1273,13 +1273,18 @@ workflows:
       - utilities-derives
       - wasm
 
+  # Jobs specific to this branch to run
+  custom-workflow:
+    when: pipeline.git.branch == "fix_target_adjustment_tests"
+    jobs:
+      - ledger-block
+
   # Run additional but unlikely to fail checks on merges only
   merge-workflow:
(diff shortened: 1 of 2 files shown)


raychu86 (approved):
> halvings will be reached slower but doublings will be reached faster.

This is not exactly true. The half-life it's the number of seconds it takes for the mining difficulty target to either double or halve, depending on whether blocks are coming too slow or too fast. But the assuming consistent block times, the `drift` is also affected by the anchor_time. Which means it will be more drastic. 

The test fix is correct because technically the drift is larger (and it assumes that the target requirement is met every block).


| Drift | Half-life | Adjustment |
|---|---|---|
| Large | Large | Moderate |
| Large | Small | Aggressive |
| Small | Large | Tiny |
| Small | Small | Moderate |

PR #3222, opened 2026-04-17
Support arrays, structs, and identifiers in ternary operators

mohammadfawaz:
## Summary

Extends the `ternary` instruction to accept **array**, **struct**, and **identifier** operands (in addition to the pre-existing literal set), gated behind `ConsensusVersion::V15`.

- `console` and `circuit`: new `Ternary` impls for `Literal` and `Plaintext` (array/struct aware, recursive on fields/elements); new `Ternary` impl for `IdentifierLiteral`.
- `synthesizer-program`: `ternary` operation now handles non-literal operands; type-checking ensures both branches have matching array shape / struct layout.
- `synthesizer`: cost model updated; deployment validation via `check_no_non_literal_ternary` rejects non-literal / non-pre-V15-supported `ternary` before V15.

## Consensus gating

Before V15, deployments containing `ternary` on arrays, structs, or `identifier` operands are **aborted** — in `function`, `finalize`, `closure`, and `constructor` bodies. At V15 and later, they are accepted and executable.

## String is intentionally not supported

The `string` literal type is **not** a permitted `ternary` operand at any version:

- **Pre-V15**: `check_no_non_literal_ternary`'s pre-V15 allow-list excludes `string` (it was never in the original macro-based list).
- **V15+**: the new `ensure_no_string_leaves` check in `Ternary::output_types` recursively walks the branch plaintext type and rejects any `string` leaf (including inside arrays or structs, local or external).

Rationale: Aleo strings are variable-length byte arrays, but the byte-wise MUX that underlies circuit ternary requires both branches to have matching byte-lengths. During deployment, circuit synthesis samples strings with `Literal::sample(String, rng)`, which yields random lengths; supporting strings would require either a padded sampling strategy or a graceful length-check path, neither of which is a small change. If we ever want strings, it would be a follow-up with its own consensus gate.

## Tests

`synthesizer/src/vm/tests/test_v15/ternary_plaintext.rs` covers:
- V14 rejection / V15 acceptance for array, struct, identifier, finalize, closure, and constructor cases.
- Execution: arrays of various sizes, nested structs, identifiers, structs containing external records.
- Type-mismatch rejection (array size mismatch, struct field mismatch).
- `ternary` inside `finalize`.
- String rejection at V15 (both plain `string` and structs containing a `string` field).

Unit tests in `console/program/src/data/literal/ternary.rs` and `circuit/program/src/data/literal/ternary.rs` exercise `Literal::ternary` dispatch across all supported variants, including `Identifier`.

### DIFF
--- a/circuit/program/src/data/literal/mod.rs
+++ b/circuit/program/src/data/literal/mod.rs
@@ -21,6 +21,7 @@ mod cast_lossy;
 mod equal;
 mod from_bits;
 mod size_in_bits;
+mod ternary;
 mod to_bits;
 mod to_fields;
 mod to_type;
--- a/circuit/program/src/data/literal/ternary.rs
+++ b/circuit/program/src/data/literal/ternary.rs
@@ -1,0 +1,148 @@
+// Copyright (c) 2019-2026 Provable Inc.
+// This file is part of the snarkVM library.
+
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at:
+
+// http://www.apache.org/licenses/LICENSE-2.0
+
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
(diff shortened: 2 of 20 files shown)


mohammadfawaz:
Support for this in Leo is here: https://github.com/ProvableHQ/leo/pull/29349

Shows on average **7.8% drop in Aleo program sizes** across the Leo test suite.

PR #3224, opened 2026-04-17
[Fix] Restore past deterministic Rng behavior

### LINKED ISSUE #4209: [Bug] Calculated Proof Target Mismatch while syncing
## 🐛 Bug Report

While syncing a node from genesis with the CDN, I see a number of messages like:
`snarkos_node_cdn::blocks: Received blocks 260950 to 261000 (in 12.07s)
Apr 09 16:05:57 snarkos-slipstream-mainnet-client-0 snarkos[2011953]: 2026-04-09T16:05:57.627149Z DEBUG snarkos_node_cdn::blocks: Received blocks 261050 to 261100 (in 12.20s)
`
 before encountering the error `WARN snarkos_node_cdn::blocks: Failed to sync block(s) from the CDN — Failed to advance to block ab1gs9j3msvhl0j3g6h2qkyxqg4w0h4dr98mjlegztgla75hcv7wqgqps2dvj at height 259573 — Failed to add block to VM — Failed to finalize transactions - Failed to post-ratify - The proof target does not match the expected proof target` at which point the node never recovers and gets caught

## Steps to Reproduce

I am running on the branch associated with [this](https://github.com/ProvableHQ/snarkOS/pull/4175) draft PR, which was last rebased on staging at [this](https://github.com/ProvableHQ/snarkOS/commit/6aea8d98613932d58b68c235daa333a66be4c76f) commit (6aea8d98613932d58b68c235daa333a66be4c76f). 

The snarkVM version used on my feature branch points to the commit [(b3f661d}](https://github.com/ProvableHQ/snarkVM/commit/b3f661d1b4e266c3395feb68d948905760cb4de9) on my [slipstream](https://github.com/ProvableHQ/snarkVM/pull/3190) branch/PR

ljedrz:
The [rand update PR](https://github.com/ProvableHQ/snarkVM/pull/3191) resulted in some (at least) past proof targets to not match the expected values, leading to sync issues. This was tracked down to 2 functions whose behavior has slightly changed due to reproducibility-breaking optimizations (which were documented in the `0.9` update, but didn't list all the affected method names).

The solution was to vendor the past implementations for the affected methods.

With these changes, I've been able to sync with the mainnet past the height of 300k (which is beyond the first problematic block height 259573).

Fixes https://github.com/ProvableHQ/snarkOS/issues/4209.

### DIFF
--- a/utilities/src/rand.rs
+++ b/utilities/src/rand.rs
@@ -176,7 +176,10 @@ impl Drop for TestRng {
 }
 
 /// This impl is Lemire's method with approximate zone. It reproduces rand 0.8's
-/// `rng.gen_range(low..=high)` for `usize` on 64-bit. It mustn't be modified.
+/// `rng.gen_range(low..=high)` for `usize` on 64-bit. It mustn't be modified
+/// to maintain backwards compatibility. It is a direct reimplementation of
+/// `sample_single_inclusive` for a concrete type (`usize`) an inlined widening multiply
+/// from https://github.com/rust-random/rand/blob/937320c/src/distributions/uniform.rs.
 pub fn gen_range_inclusive_legacy(low: usize, high: usize, rng: &mut impl Rng) -> usize {
     debug_assert!(low <= high);
 
@@ -198,7 +201,9 @@ pub fn gen_range_inclusive_legacy(low: usize, high: usize, rng: &mut impl Rng) -
 }
 
 /// This impl reproduces rand 0.8's `slice.choose_weighted(rng, weight_fn)` for u16 weights.
-/// It mustn't be modified.
+/// It mustn't be modified to maintain backwards compatibility. It is a direct, "collapsed"
+/// reimplementation of `WeightedIndex::new(weights).sample(rng)` specifically for `u16` weights
+/// from https://github.com/rust-random/rand/blob/937320c/src/distributions/weighted_index.rs.
 pub fn choose_weighted_legacy<'a, T, R: Rng>(slice: &'a [T], weight_fn: impl Fn(&T) -> u16, rng: &mut R) -> &'a T {
     // WeightedIndex::new.
     let mut cumulative: Vec<u16> = Vec::with_capacity(slice.len());
--- a/utilities/src/rand.rs
+++ b/utilities/src/rand.rs
@@ -174,3 +174,25 @@ impl Drop for TestRng {
         println!("Called TestRng with seed {} {} times", self.seed, self.calls);
     }
 }
+
+/// This impl is Lemire's method with approximate zone. It reproduces rand 0.8's
+/// `rng.gen_range(low..=high)` for `usize` on 64-bit. It mustn't be modified.
+pub fn gen_range_inclusive_legacy(low: usize, high: usize, rng: &mut impl Rng) -> usize {
+    debug_assert!(low <= high);
+
+    let range = high.wrapping_sub(low).wrapping_add(1);
+
+    // Approximate zone: conservative but avoids division.
+    let zone = (range << range.leading_zeros()).wrapping_sub(1);
+
+    loop {
+        let v = rng.next_u64() as usize;
+        // Widening multiply: v * range as u128, split into (hi, lo).
+        let wide = (v as u128) * (range as u128);
+        let hi = (wide >> 64) as usize;
+        let lo = wide as usize;
+        if lo <= zone {
+            return low.wrapping_add(hi);
+        }
+    }
+}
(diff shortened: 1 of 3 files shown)


utilities/src/rand.rs
```diff
@@ -174,3 +174,59 @@ impl Drop for TestRng {
         println!("Called TestRng with seed {} {} times", self.seed, self.calls);
     }
 }
+
+/// This impl is Lemire's method with approximate zone. It reproduces rand 0.8's
+/// `rng.gen_range(low..=high)` for `usize` on 64-bit. It mustn't be modified.
```
vicsn:
```suggestion
/// `rng.gen_range(low..=high)` for `usize` on 64-bit. It mustn't be modified
/// to maintain backwards compatibility.
```

utilities/src/rand.rs
```diff
@@ -174,3 +174,59 @@ impl Drop for TestRng {
...
+        let hi = (wide >> 64) as usize;
+        let lo = wide as usize;
+        if lo <= zone {
+            return low.wrapping_add(hi);
+        }
+    }
+}
+
+/// This impl reproduces rand 0.8's `slice.choose_weighted(rng, weight_fn)` for u16 weights.
+/// It mustn't be modified.
```
vicsn:
```suggestion
/// It mustn't be modified to maintain backwards compatibility.
```

utilities/src/rand.rs
```diff
@@ -174,3 +174,59 @@ impl Drop for TestRng {
         println!("Called TestRng with seed {} {} times", self.seed, self.calls);
     }
 }
+
+/// This impl is Lemire's method with approximate zone. It reproduces rand 0.8's
+/// `rng.gen_range(low..=high)` for `usize` on 64-bit. It mustn't be modified.
+pub fn gen_range_inclusive_legacy(low: usize, high: usize, rng: &mut impl Rng) -> usize {
```
vicsn, line 183: Can we link to the original implementation to showcase that it's the same? Same for the function below.

Should we use the rand crate's proper updated method from the latest ConsensusVersion::V15 onwards so that we don't have to audit this code too hard?

vicsn (changes requested):

ljedrz:
> Should we use the rand crate's proper updated method from the latest `ConsensusVersion::V15` onwards?

IMO that depends purely on whether the difference in performance is significant enough in those callsites - with a `V15` change, we would still need to keep the legacy implementation at least for syncing purposes, so there isn't much benefit maintenance-wise.

utilities/src/rand.rs
ljedrz: links added, moving the rest of the conversation to the main thread

vicsn:
> with a V15 change, we would still need to keep the legacy implementation at least for syncing purposes, so there isn't much benefit maintenance-wise.

It think there is a benefit, because for any *new* state we'll be sure to use the proper externally maintained rng package

ljedrz:
I'd agree if it was crypto-sensitive, but this is just picking a random number from a range, so it suffices that it returns values deterministically. In addition, the current implementation might _also_ change in the future for performance reasons, which would require us to potentially track even more versions.

utilities/src/rand.rs
```diff
@@ -174,3 +174,70 @@ impl Drop for TestRng {
...
+    // The range is 0..=usize::MAX.
+    if range == 0 {
+        return rng.random::<u64>() as usize;
+    }
+
+    // Approximate zone: conservative but avoids division.
+    let zone = (range << range.leading_zeros()).wrapping_sub(1);
+
+    loop {
+        let v = rng.next_u64() as usize;
```
vicsn, line 196: Is the reference implementation directly sampling into a u128?

utilities/src/rand.rs
```diff
@@ -174,3 +174,70 @@ impl Drop for TestRng {
...
+    // WeightedIndex::new.
+    let mut iter = slice.iter();
+    let first = iter.next().unwrap();
+    let mut total: u16 = weight_fn(first);
+    let mut cumulative: Vec<u16> = Vec::with_capacity(slice.len() - 1);
+    for item in iter {
+        cumulative.push(total);
+        total += weight_fn(item);
+    }
+    assert!(total > 0);
```
vicsn, line 221: The validator should never panic. How likely is this to trigger? Can we return a Result?

utilities/src/rand.rs
ljedrz: no, it also sampled a `u64`; if there was a difference in sampling here, it would break compatibility (since the number of consumed bytes would differ), and the syncing would keep breaking

utilities/src/rand.rs
ljedrz: the original would return an error here, which we then unwrapped, so this is effectively the same; the set of instruction weights is hardcoded and thus never empty

vicsn (approved):

vicsn (approved):

PR #3225, opened 2026-04-17
[Fix] Check subDAG atomicity correctly for a chain of pending blocks

kaimast:
This PR fixes a bug during sync, when there is a block that does not immediately reach the availability threshold. In that case, the subDAG check was comparing against the wrong block, causing errors like the following:

```
ERROR snarkos_node_bft::sync: Block synchronization failed — The previous certificate should not be linked to the current certificate in block 17800788
```

### DIFF
--- a/ledger/tests/helpers/mod.rs
+++ b/ledger/tests/helpers/mod.rs
@@ -76,9 +76,13 @@ impl TestChainBuilder {
         let genesis_rng = &mut TestRng::from_seed(seed);
         let genesis_block = VM::from(store).unwrap().genesis_beacon(&private_key, genesis_rng).unwrap();
 
-        // Extract the private keys from the genesis committee by using the same RNG to sample private keys.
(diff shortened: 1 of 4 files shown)


vicsn (approved):

ledger/tests/helpers/mod.rs
```diff
@@ -76,9 +76,13 @@ impl TestChainBuilder {
         let genesis_rng = &mut TestRng::from_seed(seed);
         let genesis_block = VM::from(store).unwrap().genesis_beacon(&private_key, genesis_rng).unwrap();
 
-        // Extract the private keys from the genesis committee by using the same RNG to sample private keys.
+        // Reconstruct the private keys of the genesis committee.  genesis_beacon uses `private_key`
+        // as the first member, then samples (committee_size - 1) more from the seeded RNG.
         let genesis_rng = &mut TestRng::from_seed(seed);
```
kaimast, line 81: Interestingly, there was a bug in `TestChainBuilder` that has not caused issues before.

PR #3226, opened 2026-04-17
[Fix] Backport correct subDAG check to mainnet

kaimast:
#3225 backported to mainnet.


raychu86 (approved):
Looks fine,  @vicsn how critical is this? Does this require a new release?

PR #3238, opened 2026-04-30
feat: read-only `query` functions

mohammadfawaz:
## Summary

Read-only `query` functions: top-level program components that read finalize-store state and return typed plaintext to external callers. No transactions, no proofs, no state writes.

```aleo
query total_balance:
    input r0 as address.public;
    get.or_use balances[r0] 0u64 into r1;
    output r1 as u64.public;
```

Entry point (gated on `--features history`):

```rust
VM::evaluate_query_at_height(program_id, name, inputs, height) -> Result<Vec<Value>>
```

snarkOS calls this with `current_block_height()` for "latest" semantics, or any earlier height for historic queries.

## What's enforced

- **Read-only.** `add_command` rejects `set` / `remove` / `async` / `await` / `call` / `rand.chacha` / record-touching ops. The runtime adapter (`HistoricFinalizeStore`) additionally bails on every write entry point.
- **Plaintext-only.** Inputs and outputs reject records, futures, dynamic records, and dynamic futures.
- **Externally-callable only.** Programs cannot `call` queries; the only entry is `VM::evaluate_query_at_height` (or `Process::evaluate_query_at_height`).
- **Off-consensus.** Each call runs on the receiving node in response to an API request. Not part of block production / transaction execution; no fee paid.
- **Deploy-time bound.** Per-query worst-case compute is checked against `TRANSACTION_SPEND_LIMIT` so a deployer can't register a pathological query.

## Snapshot semantics, no contention

Mapping reads route through `FinalizeStore::get_historical_mapping_value`, which reads the per-key historical update map (entries at any given height are immutable once written). All reads in a query body are therefore pinned to the requested `height` — block production advancing past `height` mid-evaluation cannot disturb the result.

The `FinalizeGlobalState` is reconstructed from the actual block at `height` (round, timestamp, cumulative weights, previous hash), so `block.height`, `block.timestamp`, `random_seed`, etc. all see real values.

The query path does **not** take `Process::lock()` — it would serialize queries against block production. We rely on the immutability of historic-table entries instead. (Note: `Stack` itself has interior mutability around register-type caches, so a concurrent redeploy/upgrade of the same program could in principle perturb the structural side of a running query; this is documented as a known gap and a candidate for follow-up via a `StackSnapshot`-style fix.)

## Consensus gating (V15)

- New on-disk program component variant `6` (queries).
- `Program::contains_v15_syntax` flags programs with query blocks → pre-V15 deployments are rejected with a clear error (closes the variant-6 fork hazard).
- `query` is a V15 `RESTRICTED_KEYWORDS` entry (legacy programs that used `query` as an identifier still parse / deploy).
- New `Network::MAX_QUERIES` constant (numerically equal to `MAX_CLOSURES`).
- Symmetry walks (`contains_string_type`, `contains_external_struct`, `exceeds_max_array_size`) updated to iterate queries.
- `FinalizeRegistersState`'s `transition_id` and `nonce` are `Option`-typed; the query path passes `None`. Existing readers (`rand.chacha`, finalize await) bail explicitly when absent.

## `--features history` is required at runtime

The query module (`synthesizer/process/src/query.rs`) is itself `#[cfg(feature = "history")]` — without the feature, programs with query blocks can be parsed and deployed (so deployment validation still works), but queries cannot be evaluated. snarkOS production builds enable `--features history`. CI runs the relevant test partitions with `--features test,history` so the evaluation path is exercised.

## Known limitations

- Queries cannot call other queries, closures, or functions (queries are leaves; `is_call` is rejected at construction).
- No per-call runtime cost cap below the deploy bound — rate-limiting is expected at the snarkOS RPC layer.
- **Stack interior-mutability gap** (noted above): a concurrent redeploy of the same program could perturb structural state of a running query. Follow-up via a `StackSnapshot`-style fix.

## Tests

- End-to-end (`synthesizer/src/vm/tests/test_v15/queries.rs`): lifecycle, multi-input/output, branching, arity/unknown-program/unknown-query/type-mismatch, V14-rejected vs V15-accepted, string-typed input rejected, parse-time forbidden-command rejection. Evaluation tests are gated `#[cfg(feature = "history")]`.
- Restricted keyword (`synthesizer/src/vm/tests/test_v15/restricted_keywords.rs`): `query` rejected as program/function/struct/mapping/record name at V15.
- Process-level (`synthesizer/process/src/query.rs::tests`, gated): simple eval, default-fallback, uninitialized mapping, `block.timestamp` reads work, non-plaintext input rejected, pending-write isolation, historic-at-height returns the value applicable at that height.
- Program-level (`synthesizer/program/src/query/{mod,parse,input/parse,output/parse}.rs::tests`): construction bans, parse positive/negative, duplicate-name + name-collision rejections.

### DIFF
--- a/synthesizer/process/src/query.rs
+++ b/synthesizer/process/src/query.rs
@@ -19,7 +19,7 @@ use console::{
     program::{Identifier, ProgramID, Value},
 };
 use snarkvm_ledger_store::{FinalizeMode, FinalizeStorage, FinalizeStore, atomic_finalize};
-use snarkvm_synthesizer_program::{Command, FinalizeGlobalState, FinalizeStoreTrait, RegistersTrait, StackTrait};
+use snarkvm_synthesizer_program::{FinalizeGlobalState, FinalizeStoreTrait, RegistersTrait, StackTrait};
 
 impl<N: Network> Process<N> {
     /// Evaluates a query function in the program identified by `program_id` and returns its outputs.
@@ -253,14 +253,16 @@ fn evaluate_query_inner<N: Network>(
         registers.store(stack, input_stmt.register(), value)?;
     }
 
-    // Evaluate the commands. Queries cannot await; branches reuse the finalize-side
-    // `branch_to` helper so the two execution paths cannot drift.
+    // Evaluate the commands. Queries reject `await` at construction (`add_command`), so the
+    // dispatch is identical to `Finalize` / `Constructor` — we share `finalize_command_except_await`
+    // directly to avoid drift. `try_vm_runtime!` inside that helper also gives queries panic-catch
+    // protection, which is desirable on the off-consensus / RPC-exposed path.
     //
     // Termination & cost bounds (prototype):
     //   - The loop is bounded by `query.commands().len()`, which is itself bounded by
     //     `N::MAX_COMMANDS` (= `u16::MAX`).
-    //   - `branch_to` permits forward jumps only, so the counter strictly advances and
-    //     no command can re-execute. Termination is guaranteed.
+    //   - `branch_to` (used by the helper) permits forward jumps only, so the counter
+    //     strictly advances and no command can re-execute. Termination is guaranteed.
     //   - Deploy-time, `query_cost_for_single_query` enforces that the worst-case body
     //     cost is `<= TRANSACTION_SPEND_LIMIT`, so a deployed query cannot register an
     //     unboundedly expensive body.
@@ -269,22 +271,24 @@ fn evaluate_query_inner<N: Network>(
     //     to the deploy bound per call. Rate-limiting and indexing are expected to be
     //     handled at the snarkOS RPC layer, not here.
     let mut counter = 0;
+    let mut finalize_operations: Vec<snarkvm_synthesizer_program::FinalizeOperation<N>> = Vec::new();
     while counter < query.commands().len() {
         let command = &query.commands()[counter];
-        match command {
-            Command::Await(_) => bail!("'await' is forbidden in a query function"),
-            Command::BranchEq(branch) => {
-                counter = crate::finalize::branch_to(counter, branch, query.positions(), stack, &registers)?;
-            }
-            Command::BranchNeq(branch) => {
-                counter = crate::finalize::branch_to(counter, branch, query.positions(), stack, &registers)?;
-            }
-            other => {
-                other.finalize(stack, store, &mut registers)?;
-                counter += 1;
-            }
-        }
+        crate::finalize::finalize_command_except_await(
+            store,
+            stack,
+            &mut registers,
+            query.positions(),
+            command,
+            &mut counter,
+            &mut finalize_operations,
+            query.name(),
+        )?;
     }
+    // Defensive: queries reject all write-producing commands at construction, so no finalize
+    // operations should ever be emitted. Catches any future regression that allows a write
+    // through the type-check path.
+    debug_assert!(finalize_operations.is_empty(), "query produced finalize operations: {finalize_operations:?}");
 
     // Load the outputs.
     let mut outputs = Vec::with_capacity(query.outputs().len());
--- a/synthesizer/process/src/query.rs
+++ b/synthesizer/process/src/query.rs
@@ -79,17 +79,18 @@ pub fn evaluate_query<N: Network, P: FinalizeStorage<N>>(
         registers.store(stack, input_stmt.register(), value)?;
     }
 
-    // Evaluate the commands. Queries cannot await; branches are inlined.
+    // Evaluate the commands. Queries cannot await; branches reuse the finalize-side
+    // `branch_to` helper so the two execution paths cannot drift.
     let mut counter = 0;
     while counter < query.commands().len() {
         let command = &query.commands()[counter];
         match command {
             Command::Await(_) => bail!("'await' is forbidden in a query function"),
             Command::BranchEq(branch) => {
-                counter = evaluate_branch_eq(counter, branch, query.positions(), stack, &registers)?;
+                counter = crate::finalize::branch_to(counter, branch, query.positions(), stack, &registers)?;
             }
             Command::BranchNeq(branch) => {
-                counter = evaluate_branch_neq(counter, branch, query.positions(), stack, &registers)?;
+                counter = crate::finalize::branch_to(counter, branch, query.positions(), stack, &registers)?;
             }
             other => {
                 other.finalize(stack, store, &mut registers)?;
@@ -104,48 +105,6 @@ pub fn evaluate_query<N: Network, P: FinalizeStorage<N>>(
         outputs.push(registers.load(stack, output.operand())?);
     }
     Ok(outputs)
-}
-
-fn evaluate_branch_eq<N: Network>(
-    counter: usize,
-    branch: &snarkvm_synthesizer_program::BranchEq<N>,
-    positions: &std::collections::HashMap<Identifier<N>, usize>,
-    stack: &impl StackTrait<N>,
-    registers: &impl RegistersTrait<N>,
-) -> Result<usize> {
-    let first = registers.load(stack, branch.first())?;
-    let second = registers.load(stack, branch.second())?;
-    if first == second {
-        let position = branch.position();
-        match positions.get(position) {
-            Some(index) if *index > counter => Ok(*index),
-            Some(_) => bail!("Cannot branch to an earlier position '{position}'"),
-            None => bail!("The position '{position}' does not exist."),
-        }
-    } else {
-        Ok(counter + 1)
-    }
-}
-
-fn evaluate_branch_neq<N: Network>(
-    counter: usize,
-    branch: &snarkvm_synthesizer_program::BranchNeq<N>,
-    positions: &std::collections::HashMap<Identifier<N>, usize>,
-    stack: &impl StackTrait<N>,
-    registers: &impl RegistersTrait<N>,
-) -> Result<usize> {
-    let first = registers.load(stack, branch.first())?;
-    let second = registers.load(stack, branch.second())?;
-    if first != second {
-        let position = branch.position();
-        match positions.get(position) {
-            Some(index) if *index > counter => Ok(*index),
-            Some(_) => bail!("Cannot branch to an earlier position '{position}'"),
-            None => bail!("The position '{position}' does not exist."),
-        }
-    } else {
-        Ok(counter + 1)
-    }
 }
 
 #[cfg(test)]
--- a/synthesizer/process/src/query.rs
+++ b/synthesizer/process/src/query.rs
@@ -46,11 +46,15 @@ impl<N: Network> Process<N> {
 
     /// Evaluates a query function against historic finalize-store state at the given block
     /// height. Mirrors [`evaluate_query`], but routes mapping reads through the finalize store's
-    /// historical update map. Available only when snarkVM is built with `--features history`.
+    /// historical update map. The caller (typically `VM::evaluate_query_at_height`) is responsible
+    /// for constructing `state` from the historic block at `height` so query operands reading
+    /// block metadata see the historic block's values. Available only when snarkVM is built
+    /// with `--features history`.
     #[cfg(feature = "history")]
     #[inline]
     pub fn evaluate_query_at_height<P: FinalizeStorage<N>>(
         &self,
+        state: FinalizeGlobalState,
         store: &FinalizeStore<N, P>,
         program_id: impl TryInto<ProgramID<N>>,
         query_name: impl TryInto<Identifier<N>>,
@@ -60,7 +64,7 @@ impl<N: Network> Process<N> {
         let program_id = program_id.try_into().map_err(|_| anyhow!("Invalid program ID"))?;
         let query_name = query_name.try_into().map_err(|_| anyhow!("Invalid query function name"))?;
         let stack = self.get_stack(program_id)?;
-        evaluate_query_at_height(store, &stack, &query_name, inputs, height)
+        evaluate_query_at_height(state, store, &stack, &query_name, inputs, height)
     }
 }
 
@@ -97,20 +101,22 @@ pub fn evaluate_query<N: Network, P: FinalizeStorage<N>>(
 
 /// Evaluates a query function against historic finalize-store state at the given block
 /// `height`. Mapping reads route through `FinalizeStore::get_historical_mapping_value`, which
-/// reconstructs the value applicable at `height` from the per-key update log. The returned
-/// `FinalizeGlobalState` has its `block_height` set to `height` so any consensus-version
-/// behavior the query depends on resolves consistently with the historic state.
+/// reconstructs the value applicable at `height` from the per-key update log.
+///
+/// `state` is supplied by the caller — typically `VM::evaluate_query_at_height` constructs it
+/// from the historic block at `height` so query operands reading block metadata
+/// (`block.height`, `block.timestamp`, the random seed) reflect that block.
 ///
 /// Available only when snarkVM is built with `--features history`.
 #[cfg(feature = "history")]
 pub fn evaluate_query_at_height<N: Network, P: FinalizeStorage<N>>(
+    state: FinalizeGlobalState,
     store: &FinalizeStore<N, P>,
     stack: &Stack<N>,
     query_name: &Identifier<N>,
     inputs: Vec<Value<N>>,
     height: u32,
 ) -> Result<Vec<Value<N>>> {
-    let state = FinalizeGlobalState::for_query(height);
     let historic = HistoricFinalizeStore { store, height };
     atomic_finalize!(store, FinalizeMode::DryRun, {
         evaluate_query_inner(state, &historic, stack, query_name, inputs).map_err(|e| e.to_string())
@@ -314,8 +320,12 @@ mod tests {
 
     type CurrentNetwork = MainnetV0;
 
+    /// Builds a synthetic `FinalizeGlobalState` for tests. Production callers go through
+    /// `VM::evaluate_query`, which constructs the state from a real block at the call site.
     fn sample_finalize_state(block_height: u32) -> FinalizeGlobalState {
-        FinalizeGlobalState::for_query(block_height)
+        // Use `from` to avoid the BHP hash done by `new`. The seed is irrelevant for these
+        // tests (no rand.chacha) and the round/timestamp aren't read either.
+        FinalizeGlobalState::from(block_height as u64, block_height, None, [0u8; 32])
     }
 
     #[test]
@@ -577,10 +587,10 @@ query fixed_value:
     }
 
     #[test]
-    fn test_query_rejects_block_timestamp_operand() -> Result<()> {
-        // A query that reads `block.timestamp`. Deployment must fail at type-check time —
-        // queries have no meaningful timestamp, so the runtime would otherwise bail with a
-        // confusing "not available until V12" error on every call.
+    fn test_query_can_read_block_timestamp() -> Result<()> {
+        // Queries get a real `FinalizeGlobalState` from the calling VM (built from the
+        // current/historic block), so `block.timestamp` is a valid operand inside a query body.
+        // Drive it directly here at the process layer with a synthetic state.
         let program = Program::<CurrentNetwork>::from_str(
             r"
 program qy_block_ts.aleo;
@@ -595,13 +605,18 @@ query reads_ts:
         )?;
 
         let process = Process::<CurrentNetwork>::load()?;
-        // `Stack::new` runs `FinalizeTypes::from_query` over every query at construction time,
-        // so the rejection surfaces here.
-        let err = match Stack::new(&process, &program) {
-            Ok(_) => panic!("expected stack construction to reject block.timestamp in a query"),
-            Err(err) => err.to_string(),
-        };
-        assert!(err.contains("block.timestamp"), "unexpected error message: {err}");
+        let stack = Stack::new(&process, &program)?;
+        let finalize_store = FinalizeStore::<_, FinalizeMemory<_>>::open(aleo_std::StorageMode::new_test(None))?;
+
+        // Build a state with a non-trivial timestamp, mimicking what a real VM would supply.
+        let state = FinalizeGlobalState::from(1, 1, Some(1234567890), [0u8; 32]);
+        let outputs = evaluate_query(state, &finalize_store, &stack, &Identifier::from_str("reads_ts")?, vec![])?;
+
+        assert_eq!(outputs.len(), 1);
+        match &outputs[0] {
+            Value::Plaintext(Plaintext::Literal(Literal::I64(v), _)) => assert_eq!(**v, 1234567890),
+            other => panic!("expected i64 plaintext, got: {other}"),
+        }
         Ok(())
     }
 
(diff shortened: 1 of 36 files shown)


synthesizer/process/src/stack/mod.rs
```diff
@@ -377,6 +377,13 @@ impl<N: Network> Stack<N> {
             }
         }
 
+        // Type-check every query function. The result is not cached on the stack here;
+        // it is recomputed by the query evaluator. This is acceptable for the prototype
+        // and ensures that ill-typed queries are rejected at deploy time.
+        for query in self.program.queries().values() {
+            let _ = FinalizeTypes::from_query(self, query)?;
```
vicsn, line 384: Can this handle if a respective mapping used by the query is not yet initialized?

synthesizer/program/src/query/input/parse.rs
```diff
@@ -0,0 +1,76 @@
...
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+use super::*;
+
+impl<N: Network> Parser for Input<N> {
+    /// Parses a string into an input statement.
+    /// The input statement is of the form `input {register} as {plaintext_type}.public;`.
+    #[inline]
+    fn parse(string: &str) -> ParserResult<Self> {
```
vicsn, line 22: Some parse.rs files have good positive and negative unit tests, seems like low hanging fruit to add

synthesizer/program/src/query/mod.rs
```diff
@@ -0,0 +1,242 @@
...
+
+        // Reject any state-mutating or non-deterministic command.
+        ensure!(!command.is_write(), "Forbidden operation: query functions cannot use 'set' or 'remove'");
+        ensure!(!command.is_async(), "Forbidden operation: query functions cannot invoke an 'async' instruction");
+        ensure!(!command.is_await(), "Forbidden operation: query functions cannot 'await' a future");
+        ensure!(!command.is_call(), "Forbidden operation: query functions cannot 'call' another function");
+        ensure!(!command.is_cast_to_record(), "Forbidden operation: query functions cannot cast to a record");
+
+        // Reject `rand.chacha`. It is only meaningful with finalize global state.
+        if matches!(&command, Command::RandChaCha(_)) {
```
vicsn: Perhaps we can make an appropriate `fn is_*` abstraction for this as well

synthesizer/program/src/query/mod.rs
```diff
@@ -0,0 +1,242 @@
...
+            bail!("Forbidden operation: query functions cannot use 'rand.chacha'");
+        }
+
+        // Reject `async` instructions explicitly (already covered by is_async, but be paranoid).
```
vicsn: I prefer DRY and maintainable over paranoid

synthesizer/program/src/query/mod.rs
```diff
@@ -0,0 +1,242 @@
...
+
+        // Ensure the maximum number of commands has not been exceeded.
+        ensure!(self.commands.len() < N::MAX_COMMANDS, "Cannot add more than {} commands", N::MAX_COMMANDS);
+
+        // Reject any state-mutating or non-deterministic command.
+        ensure!(!command.is_write(), "Forbidden operation: query functions cannot use 'set' or 'remove'");
+        ensure!(!command.is_async(), "Forbidden operation: query functions cannot invoke an 'async' instruction");
+        ensure!(!command.is_await(), "Forbidden operation: query functions cannot 'await' a future");
+        ensure!(!command.is_call(), "Forbidden operation: query functions cannot 'call' another function");
+        ensure!(!command.is_cast_to_record(), "Forbidden operation: query functions cannot cast to a record");
```
vicsn: If we're disabling this, maybe we should also disable `GetRecordDynamic`. Maybe there is a simpler generalization like `is_instruction_for_record`.

We could/should impose the same limitation for `FinalizeCore` commands, not sure if that requires consensus gating. Maybe we're lucky that there is another rule preventing that from being deployed, or maybe nobody did yet and we can try and roll this out quickly.

synthesizer/program/src/lib.rs
```diff
@@ -251,6 +261,7 @@ impl<N: Network> ProgramCore<N> {
         "value",
         "async",
         "finalize",
+        "query",
```
vicsn: We have a `restricted_keywords_for_consensus_version` which may be more appropriate

synthesizer/program/src/lib.rs
```diff
@@ -797,6 +834,37 @@ impl<N: Network> ProgramCore<N> {
...
+    ///
+    /// # Errors
+    /// This method will halt if the query function name is already in use in the program.
+    /// This method will halt if the query function name is a reserved opcode or keyword.
+    #[inline]
+    fn add_query(&mut self, query: QueryCore<N>) -> Result<()> {
+        let query_name = *query.name();
+
+        // Reuse the closure cap as the prototype budget for query functions.
+        ensure!(self.queries.len() < N::MAX_CLOSURES, "Program exceeds the maximum number of query functions.");
```
vicsn: Seems worth it to me to make a new constant `MAX_QUERIES`

synthesizer/src/vm/mod.rs
```diff
@@ -302,6 +302,31 @@ impl<N: Network, C: ConsensusStorage<N>> VM<N, C> {
...
+    /// This is the most ergonomic call site for query functions: the caller supplies only the
+    /// program ID, query name, and inputs. The block height, finalize store, and stack are all
+    /// resolved from the VM. Queries are a node-local lookup; no transaction is produced.
+    #[inline]
+    pub fn evaluate_query(
+        &self,
+        program_id: impl TryInto<ProgramID<N>>,
+        query_name: impl TryInto<Identifier<N>>,
+        inputs: Vec<Value<N>>,
+    ) -> Result<Vec<Value<N>>> {
```
vicsn: Should we return the block height at which the query took place? What is common in Ethereum?

synthesizer/src/vm/mod.rs
```diff
@@ -302,6 +302,31 @@ impl<N: Network, C: ConsensusStorage<N>> VM<N, C> {
...
+        // Bind the query to the current finalize state. Prototype queries do not depend on the
+        // seed material, so a zero previous-block-hash is acceptable here; matches the pattern
+        // used by `sample_finalize_state` in the test helpers.
+        let block_height = self.block_store().current_block_height();
+        let block_round = block_height as u64;
+        let state = FinalizeGlobalState::from(block_round, block_height, None, [0u8; 32]);
+
+        // Delegate to `Process::evaluate_query`.
+        let process = self.process.read();
```
vicsn: Once you rebased on the latest staging we could opt to use `let process = self.process.lock()` to give a consistent state - but I wonder whether under continuous API requests a node would have a hard time to advance... I guess naïvely that issue can be dealt with at the snarkOS level through rate-limiting, and more sophisticated API providers can index or cache outside of the node.

synthesizer/process/src/query.rs
```diff
@@ -0,0 +1,293 @@
...
+
+    // Load the outputs.
+    let mut outputs = Vec::with_capacity(query.outputs().len());
+    for output in query.outputs() {
+        outputs.push(registers.load(stack, output.operand())?);
+    }
+    Ok(outputs)
+}
+
+fn evaluate_branch_eq<N: Network>(
```
vicsn: Is there any way to re-use logic of the existing finalize code, or at least to make sure these two different approaches stay in sync?

synthesizer/process/src/query.rs
```diff
@@ -0,0 +1,293 @@
...
+    // Resolve the query function in the stack's program.
+    let query = stack.program().get_query_ref(query_name)?;
+
+    // Compute the register types for the query body.
+    let types = FinalizeTypes::from_query(stack, query)?;
+
+    // Queries are read-only and externally-callable: no transition is associated. Use a default
+    // transition ID and a zero nonce. The block height in `state` is the only state-binding.
+    let transition_id = N::TransitionID::default();
+    let mut registers = FinalizeRegisters::new(state, transition_id, *query.name(), types, 0);
```
vicsn: Is it possible and/or worth it to adjust the abstraction so we don't have to fill in dummy data? Maybe as a cop-out we make some members optional

synthesizer/src/vm/mod.rs
```diff
@@ -302,6 +302,31 @@ impl<N: Network, C: ConsensusStorage<N>> VM<N, C> {
...
+        program_id: impl TryInto<ProgramID<N>>,
+        query_name: impl TryInto<Identifier<N>>,
+        inputs: Vec<Value<N>>,
+    ) -> Result<Vec<Value<N>>> {
+        // Bind the query to the current finalize state. Prototype queries do not depend on the
+        // seed material, so a zero previous-block-hash is acceptable here; matches the pattern
+        // used by `sample_finalize_state` in the test helpers.
+        let block_height = self.block_store().current_block_height();
+        let block_round = block_height as u64;
+        let state = FinalizeGlobalState::from(block_round, block_height, None, [0u8; 32]);
```
vicsn: Is it possible and/or worth it to adjust the abstraction so we don't have to fill in dummy data? Maybe as a cop-out we make more members optional

vicsn (changes requested):
Should we add queries to `deployment_cost` or do we deem the computation trivial?

synthesizer/process/src/stack/mod.rs
mohammadfawaz: Both paths already handle this — type-check is store-independent, and runtime bails via `store.contains_mapping_speculative`. Added a regression test in https://github.com/ProvableHQ/snarkVM/pull/3238/commits/c00e40d4e.

synthesizer/program/src/query/input/parse.rs
mohammadfawaz: Added positive + negative tests for the three new query parse files in https://github.com/ProvableHQ/snarkVM/pull/3238/commits/a1a58eb4f.

synthesizer/program/src/query/mod.rs
mohammadfawaz: Done — added `Command::is_rand_chacha` and switched to it in https://github.com/ProvableHQ/snarkVM/pull/3238/commits/684413e8b.

synthesizer/program/src/query/mod.rs
mohammadfawaz: Removed the redundant block in https://github.com/ProvableHQ/snarkVM/pull/3238/commits/b135e030d. The earlier `is_async()` check already covers it; coverage is preserved by `test_reject_async_instruction`.

synthesizer/program/src/query/mod.rs
mohammadfawaz: Added `is_instruction_for_record` (covers cast-to-record incl. dynamic + `get.record.dynamic`) and applied it to query/finalize/constructor in https://github.com/ProvableHQ/snarkVM/pull/3238/commits/fa80d73a2. No consensus gate needed: every newly-rejected pattern was already rejected at the type-check layer (`check_instruction_opcode`), so the set of valid programs is unchanged — the change just shifts the failure earlier.

synthesizer/program/src/lib.rs
mohammadfawaz: Moved `query` from `KEYWORDS` to `RESTRICTED_KEYWORDS` at V15 in https://github.com/ProvableHQ/snarkVM/pull/3238/commits/81b10cbac, with rejection tests mirroring the V14 pattern (`test_v15/restricted_keywords.rs`).

synthesizer/program/src/lib.rs
mohammadfawaz: Added `Network::MAX_QUERIES` in https://github.com/ProvableHQ/snarkVM/pull/3238/commits/df815fd91 (default-valued, same numeric value as `MAX_CLOSURES` to preserve current behavior).

synthesizer/src/vm/mod.rs
mohammadfawaz: `evaluate_query` now returns `(u32, Vec<Value<N>>)` in https://github.com/ProvableHQ/snarkVM/pull/3238/commits/ffb13bdb6. Eth `eth_call` itself does not return the block (it goes in the request), but here we always run against `current_block_height()`, so returning it lets callers attribute the answer without a follow-up call that could race chain progression.

synthesizer/src/vm/mod.rs
mohammadfawaz: Decided to NOT take the lock and documented the reasoning in https://github.com/ProvableHQ/snarkVM/pull/3238/commits/e88816f77: `process.lock()` returns the same exclusive `Mutex` held during `finalize_atomic_batch`, so taking it for a query would serialize concurrent queries and block chain advancement. The query relies on `Arc<Stack<N>>` immutability + finalize-store internal locking instead.

synthesizer/process/src/query.rs
mohammadfawaz: Done — query now calls `crate::finalize::branch_to` directly (made `pub(crate)`, was `fn`) in https://github.com/ProvableHQ/snarkVM/pull/3238/commits/bb277f832. Single source of truth, no possibility of drift.

synthesizer/process/src/query.rs
mohammadfawaz: Added `FinalizeRegisters::new_for_query` to encapsulate the dummies in https://github.com/ProvableHQ/snarkVM/pull/3238/commits/983fed97c. Both fields (`transition_id`, `nonce`) are only read by `rand.chacha`, which queries forbid — the dedicated constructor documents this at the call site.

synthesizer/src/vm/mod.rs
mohammadfawaz: Same encapsulation pattern — added `FinalizeGlobalState::for_query` in https://github.com/ProvableHQ/snarkVM/pull/3238/commits/d4e357d98. Only `block_height` is meaningful for queries; the others (`block_round`, `block_timestamp`, `random_seed`) are filled with safe defaults (only read by `rand.chacha`, which queries forbid).

mohammadfawaz:
Re: *"Should we add queries to `deployment_cost` or do we deem the computation trivial?"* — went with the latter for the fee component: queries are off-consensus (each call runs on a single node in response to an API request, not as part of block production / transaction execution), and their bytes are already counted via storage cost. To stay defensive though, https://github.com/ProvableHQ/snarkVM/pull/3238/commits/ed00cecf3 adds a deploy-time bound on each query's worst-case compute (mirrors the per-function `TRANSACTION_SPEND_LIMIT` check) so a deployer can't register a pathological query.

mohammadfawaz:
Addressing more issues in https://github.com/ProvableHQ/snarkVM/pull/3238/commits/3443b3e82:

- **BLOCKER (fork hazard):** `contains_v15_syntax` now flags programs containing a `query` block, gating the new on-disk component variant.
- **Symmetry walks (BUG x3):** `Program::contains_string_type` / `contains_external_struct` / `exceeds_max_array_size` now iterate `self.queries` (with matching helpers on `QueryCore` walking inputs/commands/outputs).
- **`for_query` cleanup:** `block_round` defaults to `0` instead of mirroring the height — unambiguous "no round" sentinel for any future opcode.
- **`block.timestamp` UX:** rejected at deploy-time inside the query type-check (clear error) instead of failing at every call with the misleading "not available until V12" runtime bail.
- **`evaluate_query` UX:** pre-validates each input is `Value::Plaintext`, naming the offending kind in the error.
- **Per-call cost cap:** documented existing termination + cost bounds + the known gap (rate-limiting deferred to snarkOS layer).
- **Doc nits:** `QueryCore::new` and `Process::evaluate_query` doc strings updated.
- **Tests:** V14-rejected / V15-accepted query deployment, string-typed query input rejection, `block.timestamp`-in-query rejection, non-plaintext input rejection.

synthesizer/program/src/lib.rs
vicsn: I scanned deployed canary/testnet/mainnet programs for query to see if anyone would run into trouble upgrading, but none use it in a disallowed place, so I think we're good.

synthesizer/src/vm/mod.rs
vicsn: If in Ethereum the `height` goes in the request, does that mean people can query historic state?

We do have a --history feature. Maybe we should add make this possible and in snarkOS add a feature-gated endpoint for this?

synthesizer/src/vm/mod.rs
vicsn: I think with regards to the `Stack` we're fine because a `query` only reads data from a single `Stack`.

The `finalize_store` needs more analysis:
- Ideally the `query` command operates on confirmed state (at a single block height...), not speculative state.
- The `query`should not write to the canonical `finalize_store` at any point.

@ljedrz can you have a look at the call to `self.process.evaluate_query` in this PR to see if it satisfies those two conditions? Should we be using `atomic_finalize!(self.finalize_store(), FinalizeMode::DryRun`, and would any safe solution inevitably run into lock contention when running queries on a live node?

synthesizer/process/src/query.rs
vicsn: Can we go even further and directly re-use `finalize_command_except_await`?

synthesizer/process/src/query.rs
vicsn: I think it would be better to make the members optional, so the compiler can guide us to all the places where `transition_id` and `nonce` are being used. I hope the diff won't be too big.

synthesizer/src/vm/mod.rs
vicsn: Actually can't we get all of the required correct info from the latest block? Even if we're not using it now, we might as well and set us up for easy adapdability in the future.

vicsn: This also allows re-enabling `Operand::BlockTimestamp`

synthesizer/src/vm/mod.rs
ljedrz: - at least several of the `finalize` methods (e.g. `Contains`, `Get`, `Set`, likely more) on `Command`s use speculative queries; this could be worked around by introducing methods that instead use confirmed db queries only
- these calls don't seem to perform writes to the store, but there's technically nothing stopping them from doing so unless there's some other ongoing atomic batch; this could be overcome if the aforementioned new methods used a different kind of `FinalizeStoreTrait`, one that doesn't expose any record updates

> Should we be using `atomic_finalize!(self.finalize_store(), FinalizeMode::DryRun)`

that would ensure that ultimately no writes are performed, and that there are no other ongoing writes

> would any safe solution inevitably run into lock contention when running queries on a live node?

indeed, that would halt attempts at block finalization

mohammadfawaz: Went with the `atomic_finalize!(DryRun)` solution for now in https://github.com/ProvableHQ/snarkVM/pull/3238/commits/fca420983 — gives confirmed-state reads + no-writes guarantee at the cost of contention with block finalization. Happy to swap to the more invasive confirmed-only-reader-trait approach if preferred.

synthesizer/src/vm/mod.rs
mohammadfawaz: Done in https://github.com/ProvableHQ/snarkVM/pull/3238/commits/57e800219 — added `evaluate_query_at_height(_, height)` on `Process` and `VM`, gated `#[cfg(feature = "history")]`. Reads route through a `HistoricFinalizeStore` adapter that delegates to `FinalizeStore::get_historical_mapping_value`; writes bail. snarkOS can expose this as a feature-gated endpoint.

synthesizer/process/src/query.rs
mohammadfawaz: Done in https://github.com/ProvableHQ/snarkVM/pull/3238/commits/295a13b5d — `finalize_command_except_await` is now `pub(crate)` + generic over `FinalizeStoreTrait`, and the query loop calls it directly. `branch_to` reverted to private since it's no longer needed externally.

synthesizer/process/src/query.rs
mohammadfawaz: Done in https://github.com/ProvableHQ/snarkVM/pull/3238/commits/14686c404 — `transition_id` and `nonce` are now `Option`-typed on the `FinalizeRegistersState` trait and the `FinalizeRegisters` struct. Readers (`rand.chacha`, finalize await) bail explicitly on `None`. Removed `new_for_query` since the `Option` types make the call site self-documenting.

synthesizer/src/vm/mod.rs
vicsn: Sorry I'm late to the game but I would suggest:
- we avoid the contention issue because we're seeing a lot of nodes with performance issues this week and its a big burden
- we dont need to worry about writes because it is forbidden at compile time anyway
- but it would be great if we can be sure to only read from atomically confirmed state :)

synthesizer/src/vm/mod.rs
mohammadfawaz: Done in https://github.com/ProvableHQ/snarkVM/pull/3238/commits/f3311a767 — `VM::evaluate_query` and `evaluate_query_at_height` now build `FinalizeGlobalState` from the real block (round / timestamp / cumulative weights / previous hash), matching the consensus finalize path. Removed `for_query` and the deploy-time `block.timestamp` rejection; replaced the negative test with a positive one confirming `block.timestamp` reads work.

synthesizer/src/vm/mod.rs
mohammadfawaz: If we use `ConfirmedFinalizeStore`, wouldn't it be possible to for two reads in the same `query` to see two different points in chain time?

mohammadfawaz: Unless we get the current block height then use `evaluate_query_at_height` but that would require `--feature history`.

mohammadfawaz: In other words, this could be possible?
```
  Query thread                              Finalize thread
  ============                              ===============
  T0: capture h = current_block_height()
  T1: Read 1 → get_value_confirmed(...)
      sees confirmed state at h
                                            T2: start_atomic()
                                            T3: stage writes (pending)
                                            T4: finish_atomic()  ← commit lands
  T5: Read 2 → get_value_confirmed(...)
      sees confirmed state at h+1
  T6: return (h, [val_from_h, val_from_h+1])
```

vicsn: You're right.

How would you feel about only enabling the history feature path? Our main goal is the ARC spec. Offering this for power user explorers/indexers is just a bonus, and they run with --history often anyway.

mohammadfawaz: Yeah we can do that. Would we gate only the "evaluate_query" function with the history feature? Or the whole PR? Maybe we could offer both options, one with history and one without that uses `atomic_finalize`?

vicsn: I'm ok at minimum only gating the required and the public history query interface which will be used by snarkOS.

Your call if you have a usecase in mind to keep the latest height version.

mohammadfawaz: Done in https://github.com/ProvableHQ/snarkVM/pull/3238/commits/cb89bb6d8 — dropped the non-history `evaluate_query` entirely; only `evaluate_query_at_height` remains, gated on `--features history`. snarkOS calls it with `current_block_height()` for "latest". Also updated CI to run the affected jobs with `--features history` so the new tests are exercised.

vicsn (approved):
LGTM!

PR #3241, opened 2026-05-04
[Fix] Avoid `reqwest::blocking` when fetching parameters

### LINKED ISSUE #2931: [Build] Warnings generated by `impl_remote!`
I see a lot of these warnings related to `impl_remote!` when building certain crates in snarkVM. The following is from a call to `cargo test --package=snarkvm-synthesizer-program sha`

```
warning: associated functions `store_bytes` and `remote_fetch` are never used
   --> parameters/src/macros.rs:48:12
    |
48  |         fn store_bytes(buffer: &[u8], file_path: &std::path::Path) -> Result<(), $crate::errors::ParameterError> {
    |            ^^^^^^^^^^^
...
72  |         fn remote_fetch(buffer: &mut Vec<u8>, url: &str) -> Result<(), $crate::errors::ParameterError> {
    |            ^^^^^^^^^^^^
...
355 |         impl $name {
    |         ---------- associated functions in this implementation
    |
   ::: parameters/src/testnet/mod.rs:179:1
    |
179 | impl_remote!(InclusionV0Prover, REMOTE_URL, "resources/", "inclusion", "prover", "credits_v0");
    | ---------------------------------------------------------------------------------------------- in this macro invocation
    |
    = note: this warning originates in the macro `impl_store_and_remote_fetch` which comes from the expansion of the macro `impl_remote` (in Nightly builds, run with -Z macro-backtrace for more info)
```

kaimast:
This change is needed for https://github.com/ProvableHQ/snarkOS/pull/4220.

#3169 changed `snarkvm-parameters` to use `reqwest::blocking` instead of `curl`. The former causes issues in some snarkOS test because it spawns a new `tokio` runtime, which does not work when called from an active `tokio` runtime.

[f584af0](https://github.com/ProvableHQ/snarkVM/pull/3241/commits/f584af09c996d7d6081e4591c5f0beab711c2922) changes the parameter fetching logic to use `ureq` instead which does not spawn a `tokio` runtime under the hood. 

Additionally, [47a827a](https://github.com/ProvableHQ/snarkVM/pull/3241/commits/47a827a3261669b4c97f64330bd1a9e1d7b294e4) cleans up the parameters macros and adds documentation. It also fixes #2931. The macros are still hard to read because of the many feature gates, but the change is an improvement.

### Testing
* All native parameter tests and ledger tests passed on my setup
* I verified that wasm tests pass ( `cd wasm && wasm-pack test --node`)
* For SGX, I verified that `snarkvm-console` builds. I could not build tests for parameters, but that seemed to be a pre-existing issue. I did not run any other tests, because I do not have SGX hardware,and did not have time to set up the simulator.

### Notes
* We should add documentation for `wasm` and `sgx` builds. I could not find any and used previous pull requests to figure out how to test them.
* It is unclear to me why there is a `wasm` feature, and we do not solely rely on `target_arch='wasm32'.

### DIFF
--- a/parameters/src/canary/mod.rs
+++ b/parameters/src/canary/mod.rs
@@ -19,6 +19,8 @@ pub use genesis::*;
 /// The restrictions list as a JSON-compatible string.
 pub const RESTRICTIONS_LIST: &str = include_str!("./resources/restrictions.json");
 
+/// The remote URLs to fetch parameters from.
+#[cfg_attr(not(any(feature = "filesystem", feature = "wasm")), allow(dead_code))]
 const REMOTE_URLS: [&str; 2] = ["https://parameters.provable.com/canary", "https://s3.us-west-1.amazonaws.com/canary.parameters"];
 
 // BondPublic
--- a/parameters/src/macros.rs
+++ b/parameters/src/macros.rs
@@ -13,45 +13,86 @@
 // See the License for the specific language governing permissions and
 // limitations under the License.
 
-#[macro_export]
-macro_rules! checksum {
-    ($bytes: expr) => {{
-        use sha2::Digest;
-        hex::encode(&sha2::Sha256::digest($bytes))
-    }};
+/// Computes a SHA-256 digest of the given byte slice and returns it as a lowercase hex string.
+/// Used to verify the integrity of downloaded or loaded parameter files.
+#[inline(always)]
+pub(crate) fn checksum(bytes: impl AsRef<[u8]>) -> String {
+    use sha2::Digest;
+    hex::encode(sha2::Sha256::digest(bytes))
 }
 
-#[macro_export]
-macro_rules! checksum_error {
-    ($expected: expr, $candidate: expr) => {
-        Err($crate::errors::ParameterError::ChecksumMismatch($expected, $candidate))
-    };
+/// Constructs a `ParameterError::ChecksumMismatch` error from an expected and a computed checksum.
+/// Used to report integrity failures when a loaded or downloaded file does not match its metadata.
+/// Generic over `T` so it can be returned from any `Result<T, ParameterError>`-returning function.
+#[cfg(any(feature = "filesystem", feature = "wasm"))]
+#[inline(always)]
+pub(crate) fn checksum_error<T>(expected: String, candidate: String) -> Result<T, crate::errors::ParameterError> {
+    Err(crate::errors::ParameterError::ChecksumMismatch(expected, candidate))
 }
 
-#[macro_export]
-macro_rules! remove_file {
-    ($filepath:expr) => {
-        // Safely remove the corrupt file, if it exists.
-        #[cfg(not(feature = "wasm"))]
-        if std::path::PathBuf::from(&$filepath).exists() {
-            match std::fs::remove_file(&$filepath) {
-                Ok(()) => println!("Removed {:?}. Please retry the command.", $filepath),
-                Err(err) => eprintln!("Failed to remove {:?}: {err}", $filepath),
+/// Removes a parameter file from disk (on targets where the filesystem is available)
+/// when it is found to be corrupt or mismatched.
+/// Prints a message on success so the user knows to retry; logs a warning on failure.
+#[inline(always)]
+pub(crate) fn remove_file(filepath: impl AsRef<std::path::Path>) {
+    cfg_if::cfg_if! {
+        if #[cfg(feature="wasm")] {
+            // No-op on wasm targets where filesystem access is unavailable.
+            let _ = filepath;
+        } else {
+            let filepath = filepath.as_ref();
+            if filepath.exists() {
+                match std::fs::remove_file(filepath) {
+                    Ok(()) => println!("Removed {:?}. Please retry the command.", filepath),
+                    Err(err) => eprintln!("Failed to remove {:?}: {err}", filepath),
+                }
             }
         }
-    };
+    }
 }
 
+/// Validates a locally loaded byte buffer against an expected size and checksum, then returns it.
+/// On a size mismatch the cached file is removed. On a checksum mismatch a `ChecksumMismatch`
+/// error is returned without removing the file.
+/// Used inside `impl_local!` after reading a compile-time embedded parameter file.
+#[inline(always)]
+pub(crate) fn load_bytes_local(
+    filepath: &str,
+    buffer: &[u8],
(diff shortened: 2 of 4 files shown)


parameters/Cargo.toml
```diff
@@ -93,15 +93,9 @@ version = "0.3.64"
...
-default-features = false
-features = ["blocking", "native-tls"]
-
-[target.'cfg(all(not(any(target_os = "android", target_os = "ios")), not(target_family = "wasm"), not(target_env = "sgx")))'.dependencies.reqwest]
-version = "0.13"
-default-features = false
-features = ["blocking", "rustls"]
+[target.'cfg(all(not(target_family = "wasm"), not(target_env = "sgx")))'.dependencies.ureq]
+workspace = true
+features = ["rustls"]
```
iamalwaysuncomfortable: This change will break mobile. We need to use `nativetls` for mobile targets, we'll test out the SDK changes based on this change, but the mobile changes need to be tested out as well if we're changing to ureq.

`wasm` should be safe since it uses a funky XMLHttpREquest object, but we'll test it explicitly.

parameters/src/testnet/mod.rs
```diff
@@ -19,6 +19,9 @@ pub use genesis::*;
 /// The restrictions list as a JSON-compatible string.
 pub const RESTRICTIONS_LIST: &str = include_str!("./resources/restrictions.json");
 
+/// The remote URLs to fetch parameters from.
+/// (only used when the `filesystem` feature is active)
+#[cfg_attr(not(feature = "filesystem"), allow(dead_code))]
```
iamalwaysuncomfortable: Why only gated for filesystem? Non-filesystems also download these.

parameters/src/macros.rs
```diff
@@ -68,41 +109,40 @@ macro_rules! impl_store_and_remote_fetch {
             Ok(())
         }
 
-        #[cfg(all(not(feature = "wasm"), not(target_env = "sgx")))]
+        #[cfg(all(feature = "filesystem", not(feature = "wasm"), not(target_env = "sgx")))]
```
iamalwaysuncomfortable, line 112: Why the sudden requirement of filesystem? Non-filesystem systems still perform downloads into buffers. I don't see any specific filesystem code in here, so why gate it behind this feature?

```suggestion
        #[cfg(all(not(feature = "wasm"), not(target_env = "sgx")))]
```

parameters/src/canary/mod.rs
```diff
@@ -19,6 +19,9 @@ pub use genesis::*;
 /// The restrictions list as a JSON-compatible string.
 pub const RESTRICTIONS_LIST: &str = include_str!("./resources/restrictions.json");
 
+/// The remote URLs to fetch parameters from.
+/// (only used when the `filesystem` feature is active)
+#[cfg_attr(not(feature = "filesystem"), allow(dead_code))]
```
iamalwaysuncomfortable: Why only when filesystem is active?

iamalwaysuncomfortable (changes requested):
As is this will break mobile without MobileTLS and needs to be explicitly tested.

parameters/src/macros.rs
kaimast: Unless I misunderstood the `impl_load_bytes_logic_remote` macro, `remote_fetch` is only invoked when `filesystem` or `wasm` is active.

parameters/src/testnet/mod.rs
kaimast: This is just to hide a warning during test builds about the constant not being used. 

The change does not remove the constant, but just suppresses the warning. I will update the comment.

parameters/Cargo.toml
kaimast: It now uses native-tls as before for those targets

kaimast:
We cannot test on mobile right now, but this PR still uses the same version of `native-tls` (0.2.18) under the hood. So the likelihood of mobile breaking is very small.

parameters/src/macros.rs
```diff
@@ -190,144 +289,109 @@ macro_rules! impl_load_bytes_logic_remote {
...
-                        if $expected_size != buffer.len() {
-                            return Err($crate::errors::ParameterError::SizeMismatch($expected_size, buffer.len()));
-                        }
-
-                        return Ok(buffer)
-                    } else {
-                        return Err($crate::errors::ParameterError::FilesystemDisabled);
-                    }
-                }
+                // We need either the `filesystem` or `wasm` feature to load parameters.
```
vicsn, line 378: The error says `FilesystemDisabled`, the comment says either `filesystem` or `wasm` are needed. Should we re-use `RemoteFetchDisabled` or make a new error type?

parameters/src/macros.rs
```diff
@@ -155,28 +195,87 @@ macro_rules! impl_store_and_remote_fetch {
...
-        return Ok($buffer.to_vec());
-    };
-}
-
+/// Implements the full remote-load flow for a parameter file.
+///
+/// On native `filesystem` targets: serves from the local cache directory if present; otherwise
+/// iterates through `$remote_urls` in order, retrying on transient errors, and caches the first
+/// verified download to disk.
+/// On wasm targets: iterates through `$remote_urls` using `XmlHttpRequest`, verifying the
```
vicsn: Should we mention sgx as well?

parameters/src/macros.rs
```diff
@@ -208,7 +208,74 @@ macro_rules! impl_store_and_remote_fetch {
 macro_rules! impl_load_bytes_logic_remote {
     ($remote_urls: expr, $local_dir: expr, $filename: expr, $metadata: expr, $expected_checksum: expr, $expected_size: expr) => {
         cfg_if::cfg_if! {
-            if #[cfg(all(feature = "filesystem", not(feature="wasm")))] {
+            if #[cfg(feature = "wasm")] {
```
vicsn, line 216: The refactor https://github.com/ProvableHQ/snarkVM/pull/3241/changes/8684be5c63164ac2f18610341443aecd1b1e1f25 is very hard to audit due to the inversion of multiple related feature flags and the number of lines of code. Can we leave a more comprehensive refactor for a separate PR so we don't block snarkOS?

vicsn (changes requested):

raychu86:
How relevant is this PR now that https://github.com/ProvableHQ/snarkVM/pull/3248 is merged?

PR #3242, opened 2026-05-05
Move spend limit check to finalize phase

vicsn:
## Motivation

This is a lenient version of the block spend limit check, allowing for `num_certificates * proposal_spend_limit` worth of compute (same as snarkOS allows at the moment).

## Test Plan

Besides CI, we should ensure:
- [x] sync test passing

## Backwards compatibility

The new check is guarded by a consensus version.


synthesizer/src/vm/finalize.rs
```diff
@@ -1008,15 +1073,34 @@ impl<N: Network, C: ConsensusStorage<N>> VM<N, C> {
...
-                aborted_transactions.push((*transaction, "Fee transactions are not allowed in speculate".to_string()));
-                continue;
-            }
+            let compute_spend = match transaction {
+                Transaction::Deploy(_, _, _, deployment, _) => {
+                    let (_, cost_details) = deployment_cost(self.process(), deployment, consensus_version)?;
+                    deploy_compute_cost_in_microcredits(cost_details, consensus_version)?
+                }
+                Transaction::Execute(_, _, execution, _) => {
+                    let (_, cost_details) = execution_cost(self.process(), execution, consensus_version)?;
```
mohammadfawaz: Is there potentially a change in behavior here if `deployment_cost` or `execution_cost` fail?  I assume we previously we have caught such issues in `should_abort_transaction` which updates `aborted_transactions`.

synthesizer/src/vm/mod.rs
```diff
@@ -492,6 +494,15 @@ impl<N: Network, C: ConsensusStorage<N>> VM<N, C> {
         // Determine if the block timestamp should be included.
         let block_timestamp = (block.height() >= N::CONSENSUS_HEIGHT(ConsensusVersion::V12).unwrap_or_default())
             .then_some(block.timestamp());
+        // Determine the block spend limit.
+        let block_spend_limit = if let Authority::Quorum(subdag) = block.authority() {
```
mohammadfawaz: Nit: should we have a utility to help compute `block_spend_limit`? The computation is repeated 3 times in this PR.

synthesizer/src/vm/finalize.rs
```diff
@@ -365,6 +365,14 @@ impl<N: Network, C: ConsensusStorage<N>> VM<N, C> {
             let mut deployment_payers: IndexSet<Address<N>> = IndexSet::new();
             // Initialize a list of the successful deployments.
             let mut deployments = IndexSet::new();
+            // Initialize the previous transactions spend.
+            let mut previous_transactions_spend = 0;
+            // Determine the transaction spend limit.
+            let consensus_version = N::CONSENSUS_VERSION(state.block_height()).unwrap();
+            let transaction_spend_limit =
+                consensus_config_value_by_version!(N, TRANSACTION_SPEND_LIMIT, consensus_version).unwrap();
```
mohammadfawaz, line 420: Nit: unwraps should be justified in a comment (looks like this is also an AGENTS.md rule)

synthesizer/src/vm/finalize.rs
```diff
@@ -976,6 +1023,23 @@ impl<N: Network, C: ConsensusStorage<N>> VM<N, C> {
...
+            let id = transaction.id();
+            if compute_spend > transaction_spend_limit {
+                return Some(format!(
+                    "Transaction '{id}' exceeds the transaction spend limit with compute_spend: '{compute_spend}'"
+                ));
+            }
+            if let Some(block_spend_limit) = block_spend_limit {
+                if previous_transactions_spend.saturating_add(compute_spend) > block_spend_limit {
+                    return Some(format!(
+                        "Transaction '{id}' exceeds the block spend limit with compute_spend: '{compute_spend}'"
```
mohammadfawaz: Maybe this message should be more elaborate to also emit print `previous_transactions_spend` and `block_spend_limit`

synthesizer/src/vm/finalize.rs
```diff
@@ -377,6 +385,32 @@ impl<N: Network, C: ConsensusStorage<N>> VM<N, C> {
                     continue 'outer;
                 }
 
+                let compute_spend = match transaction {
```
mohammadfawaz: Should this be gated with a consensus version check like we're doing below?

ledger/src/check_next_block.rs
```diff
@@ -231,6 +231,16 @@ impl<N: Network, C: ConsensusStorage<N>> Ledger<N, C> {
         // Determine if the block timestamp should be included.
         let block_timestamp = (block.height() >= N::CONSENSUS_HEIGHT(ConsensusVersion::V12).unwrap_or_default())
             .then_some(block.timestamp());
+        // Determine the block's spend limit.
+        let block_spend_limit = if let Authority::Quorum(subdag) = block.authority() {
```
raychu86: This should be gated behind a `ConsensusVersion` like `block_timestamp`, where `block_spend_limit` is None before `ConsensusVersion::V15`

synthesizer/src/vm/finalize.rs
```diff
@@ -1008,15 +1073,34 @@ impl<N: Network, C: ConsensusStorage<N>> VM<N, C> {
...
 
         // Abort the transactions that are have duplicates or are invalid. This will prevent the VM from performing
         // verification on transactions that would have been aborted in `VM::atomic_speculate`.
         for transaction in transactions.iter() {
-            // Abort the transaction early if it is a fee transaction.
-            if transaction.is_fee() {
-                aborted_transactions.push((*transaction, "Fee transactions are not allowed in speculate".to_string()));
-                continue;
-            }
+            let compute_spend = match transaction {
```
raychu86: Why not calculate the `compute_spend` directly in `should_abort_transaction`?

This allows errors in `prepare_for_speculate` to fail speculation on the whole block instead of aborting the individual transactions

synthesizer/src/vm/mod.rs
```diff
@@ -492,6 +494,15 @@ impl<N: Network, C: ConsensusStorage<N>> VM<N, C> {
...
+            Some(
+                subdag.values().map(|certificates| certificates.len() as u64).sum::<u64>()
```
raychu86: This doesn't exactly translate to `batch_spend_limit` for each batch. 

Now it's that the entire block can't exceed `num_certificates * batch_spend_limit`, so some batches can have more as long as the avg is <= `batch_spend_limit`

vicsn: Indeed, it is a bit more lenient. I don't see issues with it though. An alternative would be to compute and pass around a mapping of `{proposal => {limit, transactions}}` which seems like a lot of work.

synthesizer/src/vm/finalize.rs
vicsn: Indeed, this will also interplay with https://github.com/ProvableHQ/snarkVM/pull/3214

ledger/src/check_next_block.rs
vicsn: 5d7fb010e

synthesizer/src/vm/finalize.rs
vicsn: 2ae975169

synthesizer/src/vm/mod.rs
vicsn: fff0ef146

synthesizer/src/vm/finalize.rs
vicsn: 2ae975169

synthesizer/src/vm/finalize.rs
vicsn: more duplication and lines of code

raychu86: It should be the same lines of code, it just needs to live in the individual call to `should_abort_transaction`, rather than as a loop in `prepare_for_speculate`. 

It will also better scope what happens to the individual transaction, rather than causing the entire block to fail serialization (and thus halt block production). Though this is not a likely case, if don't expect the cost calculations to fail.

vicsn: Sorry for the small previous answer, I was rushing to push my last changed before logging off.

I agree that moving the compute computation only inside of `should_abort_transaction` would be cleaner. I took the liberty to do some refactoring to make the code more digestable: 1b57bd7f5

synthesizer/src/vm/mod.rs
vicsn: Also note that since my last commits, I'm now also tracking `compute_spend` even for transactinos which may not be ultimately accepted. Seems like a very clean setup to me.

```
PrepareSpeculateResult::Finalize(compute_spend) => {
    if consensus_version >= ConsensusVersion::V15 {
        // Track the compute_spend used so far.
        block_spend = block_spend.saturating_add(compute_spend);
    }
}
```

synthesizer/src/vm/finalize.rs
vicsn: The [final diff](https://github.com/ProvableHQ/snarkVM/pull/3242/changes) is a lot more digestable at this point than reviewing commit per commit

mohammadfawaz:
LGTM! Deferring final approval to @raychu86.

synthesizer/src/vm/finalize.rs
```diff
@@ -996,60 +1090,42 @@ impl<N: Network, C: ConsensusStorage<N>> VM<N, C> {
...
-                        fee.payer().map(|payer| deployment_payers.insert(payer));
-                        deployments.insert(*deployment.program_id());
+                PrepareSpeculateResult::Abort(abort_reason) => {
+                    // Store the aborted transaction.
+                    aborted_transactions.push((*transaction, abort_reason));
+                }
+                PrepareSpeculateResult::Finalize(compute_spend) => {
+                    if consensus_version >= ConsensusVersion::V15 {
+                        // Track the compute_spend used so far.
+                        block_spend = block_spend.saturating_add(compute_spend);
```
raychu86: Transactions that fail subsequent verification via `check_transaction` still consumes block spend.

`should_abort_transaction` is not more stateful than before, so

synthesizer/src/vm/tests/test_v15/mod.rs
```diff
@@ -15,14 +15,12 @@
...
 // Tests for the record-existence check.
 mod record_existence;
-
 // Tests on the input/output behaviour of closures and related functionality.
 mod closure_records;
 // Tests on the use of `commit_*_raw` instruction variants.
 mod commit_raw;
-
-// Additional test for cost estimation without a private key.
-mod cost_for_call;
```
raychu86, line 25: The actual `cost_for_call.rs` file was not removed

synthesizer/program/src/logic/finalize_global_state/mod.rs
```diff
@@ -25,6 +25,8 @@ pub struct FinalizeGlobalState {
     block_timestamp: Option<i64>, // TODO (raychu86): Consider adding the entire Metadata here instead.
     /// The block-specific random seed.
     random_seed: [u8; 32],
+    /// The block spend limit.
+    block_spend_limit: Option<u64>,
```
raychu86, line 29: Is there a reason why this needs to live in `FinalizeGlobalState` instead of just being declared during `VM::finalize`?

The objects in `FinalizeGlobalState` were originally intended for objects that the actual finalize operations could pull from or execute on.

raychu86: The `block_spend_limit` also is not included in the pre-image of the `seed`. And it should not need to be.

vicsn: TL;DR: to reduce refactoring.

I'm computing the spend limit based on the number of certificates in the block in order to achieve more fairness. Our consensus algorithm moves in mysterious ways - under heavy load sometimes anchors may be skipped, a block contains 3x the amount of certificates, and I think we'd rather want to optimize for guaranteed transaction inclusion than stable blocktimes at this point in time. We don't have access to the number of certificates in `atomic_speculate`.

With that background in mind, the other IMO inferior options are:
1. pass a new block_spend_limit variable along to `atomic_speculate`, `speculate`, and likely more call sites, which touches 10+ files and 30+ functions. 
3. similar to the `pending_rejection_reasons` (which carries over data from `atomic_speculate` to `add_next_block`, introduce a `pending_block_spend`, which carries over data into `atomic_speculate`.

synthesizer/src/vm/tests/test_v15/mod.rs
vicsn: My bad a73d81e8b

synthesizer/src/vm/finalize.rs
vicsn: Great catch, this would have allowed a malicious validator injecting aborted txs to easily DoS. We only have to check this during finalization: 71b4ec5ef

vicsn:
@raychu86 I further improved the spend_limit calculation and ran a regression test to see it would have been sufficient for all historic blocks: 27e1842

synthesizer/src/vm/mod.rs
raychu86: Can this be attacked by a malicious validator? 

A validator could just fill their batch with to exceed the `batch_spend_limit`, and then everyone else's transactions will be aborted. Granted, this would require them to spend the fee.

synthesizer/src/vm/finalize.rs
raychu86: Nice this looks good now I think.

raychu86:
@vicsn Fixes look good. My only remaining concern is - https://github.com/ProvableHQ/snarkVM/pull/3242#discussion_r3268387482

synthesizer/src/vm/mod.rs
vicsn: Gave it some more thought and tried some prototyping. Sticking to a `batch_spend_limit` would really be better, but also a lot more refactoring, so I'd suggest we wrap this up and leave it as future work. This PR + the snarkOS PR is already an improvement over the status quo in that it prevents a halt under honest conditions.

Added one more commit e026b3c41 which clarifies the TODOs, tightly integrates `fn batch_spend_limit` again, and to reduce the need for further regression analysis it keeps the batch spend limit equal to the transaction spend limit.

ledger/narwhal/batch-header/src/lib.rs
```diff
@@ -69,12 +69,17 @@ impl<N: Network> BatchHeader<N> {
...
-    /// This implies the block spend limit is bounded at batch_spend_limit * N::NUM_MAX_CERTIFICATES` * MAX_GC_ROUNDS.
-    // TODO: div by 20 is temporary until we can dial in what the limit should be.
-    #[deprecated(note = "Use `Subdag::spend_limit` instead")]
     pub fn batch_spend_limit(height: u32) -> u64 {
-        consensus_config_value!(N, TRANSACTION_SPEND_LIMIT, height).unwrap() * Self::MAX_TRANSMISSIONS_PER_BATCH as u64
-            / 20
+        // TODO(vicsn) a more robust setup would bound the batch spend limit further.
+        // For example to: 5_f64 * credits_per_second_of_runtime / max_certificates.
+        if height >= N::CONSENSUS_HEIGHT(ConsensusVersion::V15).unwrap() {
+            consensus_config_value!(N, TRANSACTION_SPEND_LIMIT, height).unwrap()
```
raychu86: I don't understand this. This is now saying that after ConsensusVersion::V15, a batch can't spend more than a singular transaction's `TRANSACTION_SPEND_LIMIT`

synthesizer/src/vm/mod.rs
raychu86: Can you make an issue for this?

ledger/narwhal/batch-header/src/lib.rs
vicsn: In the previous versions of this PR it was much lower, but in order to reduce the scope of this PR and to avoid further regression analysis, I think a minimum of `TRANSACTION_SPEND_LIMIT` suffices. Documented this in more detail in the follow-up issue: https://github.com/ProvableHQ/snarkVM/issues/3264

synthesizer/src/vm/mod.rs
vicsn: https://github.com/ProvableHQ/snarkVM/issues/3264

ledger/narwhal/batch-header/src/lib.rs
raychu86: > In the previous versions of this PR it was much lower

I don't see this as an argument for why we should be reducing this. This makes it easier for malicious validators to fill blocks with their own transactions. 

In practice, we also aren't seeing the number of transactions being the bounds for stable block times.

raychu86:
Note: There are some failing CI and merge conflicts

ledger/narwhal/batch-header/src/lib.rs
raychu86: I'm fine with reducing it for now, as long as we have a plan to scale this back up s.t. our TPS is not hindered

vicsn: > I don't see this as an argument for why we should be reducing this.

Even in the current state of this PR, **honest** users can spam us to a blocktime of `batch_spend_limit_as_seconds` x `max_certificates` x `min_rounds_per_block` = 1 x 40 x 2 = 80 seconds, which is undesirably big.

> This makes it easier for malicious validators to fill blocks with their own transactions.

But your point is also valid. Updated this issue noting that we may want to make the `transaction_spend_limit` a fraction of `batch_spend_limit` again: https://github.com/ProvableHQ/snarkVM/issues/3264

raychu86 (approved):
I have some reservations about the new limitations as it enables some possible malicious behavior, but this should resolve more serious issues.

console/network/src/consensus_heights.rs
```diff
@@ -56,7 +56,8 @@ pub enum ConsensusVersion {
     V14 = 14,
     /// V15: Introduces the record-existence check and `commit.*.raw` instruction variants.
     ///      Increase the anchor time to 35.
-    ///      Unconditionally stores transaction rejection reasons.
+    ///      Moves the block's spend limit check to the finalize phase.
+    ///      Supports storing of transaction rejection reasons.
```
vicsn: TODO: need to bump to V16

Antonio95: @vicsn Done in https://github.com/ProvableHQ/snarkVM/pull/3242/commits/2fd0aeb7b781389a6512c77900c7efb40de0896a

ledger/narwhal/subdag/src/lib.rs
```diff
@@ -397,4 +429,71 @@ mod tests {
...
+        let subdag_n = subdag_with_cert_count(n, &mut rng);
+        let subdag_2n = subdag_with_cert_count(2 * n, &mut rng);
+
+        let limit_n = subdag_n.spend_limit(v16_height).unwrap();
+        let limit_2n = subdag_2n.spend_limit(v16_height).unwrap();
+
+        // 2 * floor(x) ≤ floor(2x) ≤ 2 * floor(x) + 1
+        assert!(limit_2n >= 2 * limit_n, "limit_2n={limit_2n} limit_n={limit_n}");
+        assert!(limit_2n <= 2 * limit_n + 1, "limit_2n={limit_2n} limit_n={limit_n}");
+    }
```
Antonio95: All operands involved in `spend_limit` are integers (at >= V16), so no decimal/floor machinery is needed. Here's the same test with a simpler assert and a corrected description, which also passes:

```
    /// Doubling the number of certificates must double the spend limit.
    #[test]
    fn test_spend_limit_proportional_to_cert_count() {
        let v16_height = CurrentNetwork::CONSENSUS_HEIGHT(ConsensusVersion::V16).unwrap();
        let mut rng = TestRng::default();

        let n: usize = 10;
        let subdag_n = subdag_with_cert_count(n, &mut rng);
        let subdag_2n = subdag_with_cert_count(2 * n, &mut rng);

        let limit_n = subdag_n.spend_limit(v16_height).unwrap();
        let limit_2n = subdag_2n.spend_limit(v16_height).unwrap();

        assert!(limit_2n == 2 * limit_n, "limit_2n={limit_2n} limit_n={limit_n}");
    }
```

Antonio95 (approved):
Only a nit in one of the tests. The rest of the points have been clarified privately.

PR #3243, opened 2026-05-05
Cleanup verification for Testnet

raychu86:
## Motivation

  This PR introduces `is_pre_accepted_testnet_transaction`, a Testnet-only exemption that marks those transactions as pre-verified, bypassing the corrected proof check for historical chain continuity. 

**The exemption is a no-op on all other networks.**


kaimast (approved):

vicsn (approved):
LGTM. Not packing this into a release until there's a high need

PR #3249, opened 2026-05-07
Backwards-compatible verifier profiling

Antonio95:
This PR removes the non-backwards compatible optimisation from PR #3237 (Varuna sponge initialisation with SHA) and leaves the other three in (preprocessed BHP windows, more efficient Poseidon S-boxes and parallel processing of transition outputs). All the details can be found in the description of the cited PR.

As part of the undone changes, the ConsensusVersion::V16 and VarunaVersion::V3 introduced in said PR have been removed.

Furthermore, the number of BHP chunks preprocessed together has been switched from 5 to 4 as the latter made some CLI flows fail (it has been confirmed that the size of the updated BHP hashers are as described in the cited PR's description, i.e. very small; but because the CLI runs many tests in parallel, the switch from 0.13 GB (for N = 4) to 0.8 GB (for N = 5) was too much).

The full PR #3237 can be merged at a later stage, once a new Varuna version is more convenient to introduce.

The updated benchmarks are as follows (cf. the other PR for an explanation):

Screenshot 2026-05-07 at 22 30 51


vicsn:
@ljedrz this PR touches a few places you're familiar with

.circleci/config.yml
```diff
@@ -1285,7 +1285,7 @@ workflows:
       or pipeline.git.branch == "canary"
       or pipeline.git.branch == "testnet"
       or pipeline.git.branch == "mainnet"
-      or pipeline.git.branch == "copilot/adjust-transaction-cache-key"
+      or pipeline.git.branch == "profile_verifier_multirecord"
```
ljedrz:
```suggestion
      or pipeline.git.branch == "profile_verifier_multirecord_compatible"
```

algorithms/src/snark/varuna/varuna.rs
```diff
@@ -137,19 +138,52 @@ impl<E: PairingEngine, FS: AlgebraicSponge<E::Fq, 2>, SM: SNARKMode> VarunaSNARK
...
+            VarunaVersion::V3 => {
+                let mut sponge = FS::new_with_parameters(fs_parameters);
+
+                let mut digest = Sha256::new();
+
+                digest.update(Self::PROTOCOL_NAME);
+                for (batch_size, inputs) in inputs_and_batch_sizes.values() {
+                    digest.update((*batch_size as u64).to_le_bytes());
+                    for input in inputs.iter() {
+                        for val in input.to_bytes_le().unwrap() {
```
ljedrz: `to_bytes_le` allocates; if possible, it would be best to perform this conversion without allocating, especially if there can be many inputs

algorithms/src/snark/varuna/varuna.rs
```diff
@@ -137,19 +138,52 @@ impl<E: PairingEngine, FS: AlgebraicSponge<E::Fq, 2>, SM: SNARKMode> VarunaSNARK
...
+                            digest.update([val]);
+                        }
+                    }
+                }
+                for circuit_specific_commitments in circuit_commitments {
+                    digest.update(circuit_specific_commitments.to_bytes_le().unwrap());
```
ljedrz: ditto (iterate over bytes instead of allocating if possible)

console/algorithms/src/bhp/hasher/hash_uncompressed.rs
```diff
@@ -29,7 +29,12 @@ impl<E: Environment, const NUM_WINDOWS: u8, const WINDOW_SIZE: u8> HashUncompres
     /// the BHP commitment scheme, as it is typically not used by applications.
     fn hash_uncompressed(&self, input: &[Self::Input]) -> Result<Self::Output> {
         // Ensure the input size is at least the window size.
-        ensure!(input.len() > Self::MIN_BITS, "Inputs to this BHP must be greater than {} bits", Self::MIN_BITS);
+        ensure!(
+            input.len() > Self::MIN_BITS,
```
ljedrz, line 33: shouldn't this be `>=`? `MIN_BITS` sounds like a valid value; alternatively, `MIN_BITS` could be incremented by `1`

console/algorithms/src/bhp/hasher/mod.rs
```diff
@@ -35,6 +43,11 @@ pub struct BHPHasher<E: Environment, const NUM_WINDOWS: u8, const WINDOW_SIZE: u
     bases: Arc<Vec<Vec<Group<E>>>>,
     /// The bases lookup table for the BHP hash.
     bases_lookup: Arc<Vec<Vec<[Group<E>; BHP_LOOKUP_SIZE]>>>,
+    /// The preprocessed combinations of elements in the bases lookup table.
+    // For each group of BHP_NUM_COMBINED_CHUNKS contiguous
+    // BHP_LOOKUP_SIZE-element tuples in `bases_lookup`, this contains
+    // all possible BHP_LOOKUP_SIZE^BHP_NUM_COMBINED_CHUNKS cross-tuple sums.
+    combined_bases_lookup: Arc<Vec<Vec<Vec<Group<E>>>>>,
```
ljedrz:
```suggestion
    combined_bases_lookup: Arc<[Vec<Vec<Group<E>>>]>,
```
it's almost always preferable to collapse `Arc<Vec<T>>` into `Arc<[T]>`

console/algorithms/src/bhp/hasher/mod.rs
ljedrz, line 43: is this object cloned often? if that happens in any hot spot, it might be best to introduce an intermediate `BHPHasherInner` object in order to reduce the number of `Arc`s involved in cloning (since with these changes, it involves 4 atomic operations) to 1

ljedrz (commented):
Left some comments.

algorithms/src/snark/varuna/varuna.rs
Antonio95: This belonged to the previous, backwards-breaking PR. The current PR reverts sponge initialisation to how it happens in staging, and in particular does not have the highlighted code. I will link your comment in the other PR in case we get back to it, though (which I hope we will!).

algorithms/src/snark/varuna/varuna.rs
Antonio95: ditto^2 :)

console/algorithms/src/bhp/hasher/hash_uncompressed.rs
Antonio95: (Keen eye!) I agree this is a bad name, but it's unrelated to this PR: it's already the case in staging. https://github.com/ProvableHQ/snarkVM/blob/staging/circuit/algorithms/src/bhp/hasher/hash_uncompressed.rs#L31-L34. I would leave it as is for now in order to avoid breaking changes.

console/algorithms/src/bhp/hasher/mod.rs
Antonio95: Thanks, I didn't know 👌  Done in 7f9433cc3c816b49932ed5dd7eb67865e95f611d.

console/algorithms/src/bhp/hasher/mod.rs
Antonio95: From what I can tell, it is cloned:
 - Once [here](https://github.com/ProvableHQ/snarkVM/blob/profile_verifier_multirecord_compatible/circuit/network/src/v0.rs#L68), when initialising the Network's local-thread hashers (separate for Mainnet, Testnet and Canary)
 - In `MerkleTree::clone()` [here](https://github.com/ProvableHQ/snarkVM/blob/profile_verifier_multirecord_compatible/console/collections/src/merkle_tree/mod.rs#L76). This is in turn called a few MerkleTree methods, most notably in `prepare_update` inside [`update`](https://github.com/ProvableHQ/snarkVM/blob/profile_verifier_multirecord_compatible/console/collections/src/merkle_tree/mod.rs#L287).
 - In a few tests, which is probably not critical
 
 I've seen you've some work on these so you probably have a decent feeling whether this is too costly. My intuition says that the gain from updating 4 ref counts to 1 is neligible in the `update` and similar paths, which already do a lot of work (e.g. hashing). In exchange, the price of switching to sth like this
 ```
 #[derive(Debug, PartialEq, Deserialize, Serialize)]
#[serde(bound = "E: Serialize + DeserializeOwned")]
struct BHPHasherInner<E: Environment, const NUM_WINDOWS: u8, const WINDOW_SIZE: u8> {
    bases: Vec<Vec<Group<E>>>,
    bases_lookup: Vec<Vec<[Group<E>; BHP_LOOKUP_SIZE]>>,
    combined_bases_lookup: Box<[Vec<Vec<Group<E>>>]>,
    random_base: Vec<Group<E>>,
}
#[derive(Clone, Debug, PartialEq, Deserialize, Serialize)]
#[serde(bound = "E: Serialize + DeserializeOwned")]
pub struct BHPHasher<E: Environment, const NUM_WINDOWS: u8, const WINDOW_SIZE: u8> {
    inner: Arc<BHPHasherInner<E, NUM_WINDOWS, WINDOW_SIZE>>,
}
```
would be
 1. A change in serialization, I'm not 100% sure if that could cause any unanticipated issues;
 2. [These accessors](https://github.com/ProvableHQ/snarkVM/blob/profile_verifier_multirecord_compatible/console/algorithms/src/bhp/hasher/mod.rs#L216-L224) would no longer work, the signature would have to change - and they are public-facing.
 
 What do you think?

Comment (ljedrz) [reply console/algorithms/src/bhp/hasher/mod.rs:43]: These sites don't seem too hot (and if that ever changes, it shouldn't be too hard to find it in profiling), so it should be ok.

Comment (Antonio95) [reply console/algorithms/src/bhp/hasher/mod.rs:43]: Okay, thanks for the reassurance. Let's keep it as it is for now then 👌

**Comment by Antonio95** [comment]
Update:

With the V17 changes merged in, the last benchmark `prepare_advance_multirecord` failed: the two programs were too large and failed to deploy.
- In order to enable larger programs, `genesis_with_fixed_committee_seed` (which was created with 4 validators) has been changed to use `CurrentNetwork::LATEST_MAX_CERTIFICATES()`, which is 104 at the time of writing. This means consensus rounds are fuller and the maximum 10-second synthesis limit is achieved. This sufficed to make the child program deploy.
- The caller program had a total combined density of ~24_000_000, higher than the maximum achievable of ~15_000_00. It has been shrunk from 30 calls to the child program to 18, which makes the flow work.

Benchmarks now execute successfully. `check_transaction_multirecord` shows essentially the same times as originally reported in parallel, and ~25% worse (than reported) in `--serial` due to unrelated changes. `prepare_advance_multirecord` cannot be compared with the original reported times due to the changes in the program.

File: console/algorithms/src/bhp/hasher/mod.rs [hunk 6]
```diff
@@ -15,17 +15,25 @@
...
+use rayon::prelude::*;
 use serde::{Deserialize, Serialize};
 use std::sync::Arc;
 
 /// The BHP chunk size (this implementation is for a 3-bit BHP).
 pub(super) const BHP_CHUNK_SIZE: usize = 3;
 pub(super) const BHP_LOOKUP_SIZE: usize = 1 << BHP_CHUNK_SIZE;
 
+// The amount of chunks (i.e. bit triplets) to preprocess together in the lookup table.
+pub(super) const BHP_NUM_COMBINED_CHUNKS: usize = 4;
```
vicsn, line 37: Should we document here why we chose this number

console/algorithms/src/bhp/hasher/hash_uncompressed.rs
```diff
@@ -49,19 +54,49 @@ impl<E: Environment, const NUM_WINDOWS: u8, const WINDOW_SIZE: u8> HashUncompres
             Cow::Borrowed(input)
         };
 
-        // Compute sum of h_i^{sum of (1-2*c_{i,j,2})*(1+c_{i,j,0}+2*c_{i,j,1})*2^{4*(j-1)} for all j in segment}
-        // for all i. Described in section 5.4.1.7 in the Zcash protocol specification.
```
vicsn, line 53: Should we keep the reference to the zcash doc? So a reader knows there's some prior art and security analysis

vicsn (approved):
Mostly LGTM, two minor comments

vicsn:
Also some new merge conflicts remain

console/algorithms/src/bhp/hasher/hash_uncompressed.rs
Antonio95: Done in 398709a3d1fb129a14350ee96b698b48c6930abe

console/algorithms/src/bhp/hasher/mod.rs
Antonio95: Done in 398709a3d1fb129a14350ee96b698b48c6930abe

PR #3250, opened 2026-05-09
docs/test(query): qualify Stack snapshot claim + add name-collision tests

mohammadfawaz:
Follow-up to #3238.

## Changes

- **Doc fix:** `VM::evaluate_query_at_height` and `evaluate_query_at_height` (process layer) previously claimed the result was fully snapshot-consistent at the requested height and that we relied on \`Arc<Stack<N>>\` immutability. ljedrz pointed out the latter is misleading — \`Stack\` has interior mutability for its register-type / finalize-type / old-stacks caches, so a concurrent redeploy of the same program could perturb the structural side mid-query. The mapping-value side IS snapshot-consistent (via the per-key historical update map); only the structural side is not. Doc updated to be honest about the gap; a future \`StackSnapshot\`-style fix is the path to closing it.
- **Tests:** add three name-rejection tests in \`synthesizer/program/src/parse.rs\` — duplicate query names, query/function name collision, query/closure name collision. The behavior was already in place via \`add_query\`'s \`is_unique_name\` check; these lock it in.

### DIFF
--- a/synthesizer/process/src/query.rs
+++ b/synthesizer/process/src/query.rs
@@ -55,19 +55,17 @@ impl<N: Network> Process<N> {
 }
 
 /// Evaluates a query function against historic finalize-store state at the given block
-/// `height`. Mapping reads route through `FinalizeStore::get_historical_mapping_value`, which
-/// reconstructs the value applicable at `height` from the per-key update log (a confirmed-only
-/// table — entries are only written during finalize, never modified). Pinning every read to
-/// `height` therefore gives true snapshot semantics: block production advancing past `height`
-/// during evaluation cannot disturb the result, and we don't need to take an atomic batch on
-/// the store (so this path doesn't contend with block finalization either).
+/// `height`. Mapping reads route through `FinalizeStore::get_historical_mapping_value` and
+/// are snapshot-consistent at `height` without contending with block finalization.
 ///
-/// `state` is supplied by the caller — typically `VM::evaluate_query_at_height` constructs it
-/// from the historic block at `height` so query operands reading block metadata
-/// (`block.height`, `block.timestamp`, the random seed) reflect that block.
+/// `state` is supplied by the caller — typically built from the historic block at `height`
+/// so block-metadata operands reflect that block.
 ///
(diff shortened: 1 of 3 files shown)


synthesizer/process/src/query.rs
```diff
@@ -55,19 +55,17 @@ impl<N: Network> Process<N> {
...
 ///
-/// `state` is supplied by the caller — typically `VM::evaluate_query_at_height` constructs it
-/// from the historic block at `height` so query operands reading block metadata
-/// (`block.height`, `block.timestamp`, the random seed) reflect that block.
+/// `state` is supplied by the caller — typically built from the historic block at `height`
+/// so block-metadata operands reflect that block.
 ///
-/// Available only when snarkVM is built with `--features history`. snarkOS calls this with
-/// `current_block_height()` for "latest", or any earlier height for historic queries.
+/// Caveat: the live `Stack` has interior mutability, so a concurrent redeploy of the same
```
vicsn, line 64: Note that if we allow query calls from within finalize blocks, this would not be an issue because those would be run serially, one transaction at a time.

@copilot can you make an issue and draft PR for allowing query calls from finalize blocks?

vicsn (approved):

vicsn:
Actually, the point about Stack mutability reveals a bigger issue.

The API-exposed query will be evaluating _historic_ state on the _latest_ known `Stack` due to:
```
        let stack = self.get_stack(program_id)?;
        evaluate_query_at_height(state, store, &stack, &query_name, inputs, height)
```

It would be actually correct if we fetch the historic deployment and recreate its potentially historic `Stack`. That may require fetching from or iterating over `ProgramMap`, then getting the transaction ID from `ReverseIDMap` and then checking which block that transaction ID was included to see which edition is appropriate.

If that program imports other programs, instantiation of the respective `Stack` may require fetching imported programs of the appropriate editions... and in the worst case this requires recreating huge nested Stacks, which is not desirable. However, at the moment queries cannot call other queries, so perhaps we can find a way to instantiate a query's Stack with existing or new abstractions so it doesn't import anything...

PR #3253, opened 2026-05-11
feat(view): allow finalize bodies to call `view` functions

### LINKED ISSUE #3252: [Feature] Allow query to be called from function or finalize
## 🚀 Feature

Note that in the function scope, we currently only allow a single finalization. So adding a query becomes awkward to say the least, and requires appropriate in-circuit checks.

mohammadfawaz:
Closes #3252.

A function's `finalize` body can now `call` a view function — same-program or imported. Views remain leaves (cannot call other views).

```aleo
finalize compute_total:
    input r0 as address.public;
    call my_token.aleo/get_balance r0 into r1;  // call a view
    add r1 100u64 into r2;
    set r2 into total[r0];
```

Views may also declare zero outputs and serve as cross-program preconditions (Aleo analogue of Solidity's `function require_member(address) external view { require(...); }`):

```aleo
view require_member:
    input r0 as address.public;
    get.or_use members[r0] false into r1;
    assert.eq r1 true;

finalize transfer:
    input r0 as address.public;
    call vw_acl.aleo/require_member r0;   // no `into`; aborts the tx if the guard fails
    ...
```

## What's enforced

- **Construction**: `Finalize::add_command` permits `Call`; `call.dynamic` rejected via `Command::is_dynamic_call`. Constructors still forbid `call` entirely. View inputs/commands/outputs are all `many0` (matches `function`); body constraints (no record-touching ops, no state writes, no `async`/`await`/`call`/`rand.chacha`) are enforced by `ViewCore::add_command`.
- **Type-check**: `Opcode::Call(_)` in a finalize body is allowed only when the target resolves to a view; each operand's type is checked against the view's declared input type, and the destination count must match the view's output count. `call.dynamic` is bailed in the finalize type-checker.
- **Runtime**: `Call::finalize` loads inputs from the caller's `FinalizeRegisters`, dispatches the view body via a new `StackTrait::evaluate_view` hook against the live finalize store with the caller's `FinalizeGlobalState`, and writes outputs back. A view-body failure surfaces as a finalize rejection.
- **Cost**: `cost_per_command` folds the callee's `view_cost_for_single_view` into the caller's finalize cost, keeping `TRANSACTION_SPEND_LIMIT` as the combined bound.
- **V15 gate**: `Program::contains_v15_syntax` flags any `call` in a finalize body; arity relaxations ride the same gate.


synthesizer/process/src/lib.rs
```diff
@@ -36,7 +36,10 @@ mod deploy;
 mod evaluate;
 mod execute;
 mod finalize;
-#[cfg(feature = "history")]
+// The view module is unconditionally compiled because finalize bodies can call views
```
vicsn: Superfluous docs

synthesizer/program/src/finalize/mod.rs
```diff
@@ -167,8 +167,10 @@ impl<N: Network> FinalizeCore<N> {
 
         // Ensure the command is not an async instruction.
         ensure!(!command.is_async(), "Forbidden operation: Finalize cannot invoke an 'async' instruction");
-        // Ensure the command is not a call instruction.
-        ensure!(!command.is_call(), "Forbidden operation: Finalize cannot invoke a 'call'");
+        // Allow `call` only when the target resolves to a `view` function (enforced later by the
+        // type-check at `Stack::new`). `call.dynamic` remains forbidden because its target type
+        // is determined at runtime and would let arbitrary functions be invoked from finalize.
```
vicsn: No it's because we'd first need to design a way to track dynamic spend / gas usage

synthesizer/program/src/logic/instruction/operation/call/standard.rs
```diff
@@ -211,8 +211,39 @@ impl<N: Network> Call<N> {
             }
         };
 
+        // If the operator is a view function, retrieve the view and compute the output types.
+        // Views are externally-callable and declared as `FinalizeType::Plaintext(_)` for both
+        // inputs and outputs (enforced by `ViewCore::add_input`/`add_output`); we therefore
+        // pattern-match `FinalizeType::Plaintext` directly when constructing the `RegisterType`.
+        if let Ok(view) = program.get_view_ref(name) {
```
vicsn, line 273: Could it be worth creating a new `FinalizeCall` operation, giving us extra clarity and assurance that old edge case `Call` behaviour doesn't leak into the finalize-version of doing calls? Or would it just be a bunch of code duplication for little benefit?

vicsn (changes requested):

synthesizer/process/src/lib.rs
mohammadfawaz: Fixed in b557ddad9c063493d9df943601987fba2ea41476

synthesizer/program/src/finalize/mod.rs
mohammadfawaz: Fixed in b557ddad9c063493d9df943601987fba2ea41476

synthesizer/program/src/logic/instruction/operation/call/standard.rs
mohammadfawaz: Addressed in 05934278aa9a6282e05c9a841b3570150b3e6559 — moved the view-target handling into a dedicated `Call::output_types_for_view` so the transition-side `Call::output_types` never sees views. Creating a separate `FinalizeCall` doesn't feel worth it 🤔

vicsn: Summary of our discussion: I'd be ok if we revert 0593427, and maybe we can move the finalize logic into `Call::finalize`

mohammadfawaz: Reverted in https://github.com/ProvableHQ/snarkVM/pull/3253/changes/b74ddbdf1f0a76113ce4e705f6ee3bc243818de5
and the second part is done in https://github.com/ProvableHQ/snarkVM/pull/3253/changes/8f0ac1070f4142a94d5f9a10bc4f5600dde8e08e

The small downside is that `FinalizeStoreTrait` now has to be passed to every `fn finalize` even though it's only used by the one corresponding to `Call`

vicsn: Great, it's a bit awkward, but I think now acceptable.

One nit: Could we opt for a simpler approach and pass in None instead of sampling the dummy finalize store?

mohammadfawaz: https://github.com/ProvableHQ/snarkVM/pull/3253/changes/935efcff4afd633e51ece2ea98a5ee441238afb2

vicsn (changes requested):
We need to update `check_upgrade_is_valid`. It's a pity that logic is so hidden and not tightly integrated.

We currently prevent a function from defining more than MAX_TRANSITIONS static calls in `get_minimum_number_of_calls`. Will we run into DoS trouble if someone defines a finalize scope calling into MAX_COMMANDS views? Defining a `MAX_CALLS = MAX_TRANSITIONS` is probably acceptable UX.

We could also consider changing `check_program_is_well_formed` to allow a deployment of programs with only views, not sure if this will break existing assumptions in the code. For example, deployments could then have 0 verifying keys. I'm leaning towards leaving this as future work.

mohammadfawaz:
Re: [vicsn's review](https://github.com/ProvableHQ/snarkVM/pull/3253#pullrequestreview-4343710022)

> We need to update `check_upgrade_is_valid`. It's a pity that logic is so hidden and not tightly integrated.
>
> We currently prevent a function from defining more than MAX_TRANSITIONS static calls in `get_minimum_number_of_calls`. Will we run into DoS trouble if someone defines a finalize scope calling into MAX_COMMANDS views? Defining a `MAX_CALLS = MAX_TRANSITIONS` is probably acceptable UX.
>
> We could also consider changing `check_program_is_well_formed` to allow a deployment of programs with only views, not sure if this will break existing assumptions in the code. For example, deployments could then have 0 verifying keys. I'm leaning towards leaving this as future work.

Addressed in fae8a28d6: `MAX_CALLS = 32` enforced in `Finalize::add_command`, and `check_upgrade_is_valid` now requires each old view to remain in the new program with the same input/output types (body mutable, same policy as functions/finalize).

Filed #3271 to track the view-only deployment relaxation as future work.

vicsn (approved):

PR #3258, opened 2026-05-13
Fix synthesis of circuits with `Scalar` outputs

### LINKED ISSUE #2801: [Bug] Deployment can fail depending on sampled values

Antonio95:
Closes https://github.com/provableHQ/snarkVM/issues/2801

**Note:** Before merging, confirm the `merge` CLI flow triggers `synthesizer/tests/test_vm_execute_and_finalize.rs` (this PR deletes the `#[test]` annotation preceding the `#[test_log::test]`, which was causing duplicate execution of that long test).

### The issue

Aleo programs which `cast` a value of type one of {`Field`, `Group`, `Address`} to `Scalar` and output the latter often cause errors during deployment verification.

The reason is that (unlike a `console::Scalar`) the circuit representation of a `Scalar` actually contains a `Field` element, which has ~2 more bits (i.e. takes values up to  ~4 times larger) than a `Scalar`. When a function’s circuit is constructed (e.g. during key synthesis and verifying-key verification), the resulting `Response` is ejected, which in particular ejects the outputs therein. Since inputs to the function are sampled randomly during synthesis, an input of one of the three types mentioned at the start and cast to a `Scalar` which is subsequently output often results in an underlying `Field` value which panics at `eject` time.

Only this case of `cast` is problematic. For instance, when `cast`ing `U16` to `U8`, constraints are added to the circuit to ensure the most significant 8 bits are 0, and only the 8 least significant bits form the new `U8` value. Constraints are not enforced during key synthesis, and ejecting the 8 bits of the `U8` always results in valid `console::U8`.

There are two subtler places where problematic `Scalar`s were being ejected other than when ejecting the response: in the `load_circuit` and `store_circuit` calls in the the circuit-construction part of the `cast` instruction. These two functions perform a type check of the circuit value received against the destination register’s type. This was done by first ejecting the circuit value into a console one and then calling `StackTrait`’s `matches_value_type` (which only accepts console types). This ejection also resulted in an error once the aforementioned one was fixed.

### The fix

Part of the design philosophy has been not to add any special-case treatment to `Scalar` type for improved future maintainability.

- In `CheckDeployment` and `Synthesize` mode, we no longer eject the `Response` at the point where we were. Instead of returning that, we now return an `Option` which is only populated in modes other than the two above, returning `None` in those two.
- At subsequent points down the execution flow, in  `CheckDeployment` and `Synthesize` mode, we sample random outputs. These are the only part of the (no longer present) `Response` needed at that stage, and only their type is relevant.
- The `eject` + `matches_value_type` calls in `load_circuit` and `store_circuit` have been replaced by a new `circuit_matches_value_type`. This function performs the same type exploration as `matches_value_type` but acts on circuit values and does not eject. Although this results in some code duplication, it does not result in a performance loss but the opposite: the `eject` + `matches_value_type` flow effectively resulted in two separate explorations of the same type tree (plus any relevant value conversions in `eject`), as opposed to the single exploration in the new `circuit_matches_value_type` (and no value conversions).

The test file `synthesizer/src/vm/tests/test_v15/scalar_outputs.rs` has been added, which tests key synthesis and verification in functions with previously failing casts. This has been checked on the three problematic source types, both for private and public inputs and outputs, as well as in closures.

Note this PR does not need to be `ConsensusVersion`-gated, since the fix will not cause deployments which would be rejected to be accepted (validators with pre-fix code simply panic, which does not lead to an inconsistent blockchain state with those who accept). Expectations of `test_vm_execute_and_finalize` had to be regenerated since the sampling of random request outputs (where before a `Request` was being ejected instead) desynchronised the state of the `rng` with respect to pre-change executions and resulted in outputs different than expected. The expectation regeneration was inspected to be correct (and not, e.g., overwrite previously-expected-to-fail cases with passing ones or vice-versa).


synthesizer/process/src/stack/registers/registers_circuit.rs
```diff
@@ -237,4 +237,384 @@ impl<N: Network, A: circuit::Aleo<Network = N>> RegistersCircuit<N, A> for Regis
             Register::Access(..) => bail!("Cannot store to a register access: '{register}'"),
         }
     }
+
+    /// Checks that the given circuit value matches the layout of the register type. This is a circuit analogue of [`StackTrait::matches_register_type`].
+    fn circuit_matches_register_type(
```
vicsn, line 250: This PR is perfect except for the massive amount of duplication in this file. Could we perhaps use a macro to deduplicate? See for example the magic we did in `macro_rules! do_hash`, `macro_rules! do_commit`, `macro_rules! do_ecdsa_verification`

vicsn (changes requested):

synthesizer/src/vm/tests/test_v15/scalar_outputs.rs
```diff
@@ -0,0 +1,93 @@
...
+    // Since |scalar field| ~ |base field|/4, we run each test enough times N =
+    // 5 that a value which does not fit inside a Scalar will be sampled with a
+    // high enough probability 1 - 1/4^N.
+    for _ in 0..5 {
+        // The three types which can be cast to a Scalar but do not always fit
+        // inside one are Field, Group and Address.
+        test_program_with_cast_source_type("field", &vm, &caller_private_key, rng);
+        test_program_with_cast_source_type("group", &vm, &caller_private_key, rng);
+        test_program_with_cast_source_type("address", &vm, &caller_private_key, rng);
+    }
```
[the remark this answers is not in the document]
Antonio95: I already considered this and checked the test's runtime, which was 57 seconds. I've now made some changes to bring it down to 33s.

synthesizer/process/src/stack/registers/mod.rs
```diff
@@ -19,7 +19,7 @@ mod registers_trait;
 use crate::{CallStack, RegisterTypes};
 use console::{
     network::prelude::*,
-    program::{Entry, Literal, Plaintext, Register, Request, Value},
+    program::{Entry, Literal, Plaintext, PlaintextType, Register, RegisterType, Request, Value},
```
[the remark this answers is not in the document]
Antonio95: They are used in the child module `registers_circuit.rs`, which imports `super::*`, which is why clippy was happy. I've now moved the two imports to the child module anyway.

synthesizer/process/src/stack/registers/registers_circuit.rs
Antonio95: I agree with your call. It required hours of macro surgery work, but I'm happy to report I've got it working in the latest commit 🙌  
https://github.com/ProvableHQ/snarkVM/pull/3258/changes/030e388f003d06ae852d3b44a06e29355b7af15f

synthesizer/process/src/stack/call/standard.rs
```diff
@@ -328,11 +328,20 @@ impl<N: Network> CallTrait<N> for Call<N> {
                         // Push the request onto the call stack.
                         call_stack.push(request.clone())?;
 
-                        // Execute the request.
+                        // Synthesize the circuit. Note the response will be empty.
                         let response = substack.execute_function::<A, R>(call_stack, console_caller, root_tvk, rng)?;
 
+                        // TODO (Antonio) only place affected, sort out
```
vicsn: Todo left

synthesizer/src/vm/mod.rs
```diff
@@ -1280,7 +1280,7 @@ function a:
             // Note: `deployment_transaction_ids` is sorted lexicographically by transaction ID, so the order may change if we update internal methods.
             assert_eq!(
                 deployment_transaction_ids,
-                vec![deployment_1.id(), deployment_2.id(), deployment_4.id(), deployment_3.id()],
+                vec![deployment_4.id(), deployment_1.id(), deployment_3.id(), deployment_2.id()],
```
vicsn: Can you remove this assert? I've thought for years its useless and creates busywork to just update the expectation

synthesizer/process/src/stack/registers/registers_circuit.rs
vicsn: Very nice

vicsn (changes requested):
You may have forgotten to commit the new macro

synthesizer/src/vm/mod.rs
Antonio95: @vicsn I think part of what that test wants to check is that `VM::from` imports programs (i.p. depending on one another) correctly regardless of the order `vm.transaction_store()` returns them in. So, in order for the check `assert!(VM::from(vm.store.clone()).is_ok());` in line 1289 to be meaningful, one has to check `vm.transaction_store()` is not already returning them in the correct import order (1, 2, 3, 4) by chance.

Which of the two following options do you favour?
 - Replacing the concrete hardcoded order in the check to the check "make sure `vm.transaction_store()` returns them in an order different from 1, 2, 3, 4". This would maintain the meaningfulness of the test and make it so that, on average, if any PR changes things that affect that order (i.p. the random transaction IDs), this test will need to be updated with probability only 1 / 4! = 1/24 (= the chance that the random changes result in the order 1, 2, 3, 4).
 - Removing this check and living with the risk of the test not being meaningful during some periods in the future. This doesn't seem too bad in the sense that this has been tested over and over, but who knows if future changes could potentially break it (in which case it's likely other tests would break anyway).

synthesizer/process/src/stack/call/standard.rs
Antonio95: On it. I've been doing my own review of the PR for soundness (which is why it's still in draft mode).

For context, that TODO is related to the fact that the check in https://github.com/ProvableHQ/snarkVM/pull/3258/changes#diff-ba4fccc0a22e2dd004e835aef17e46aab9e4ad41797cb811e164fea71be0072dL551-L554
now only happens for modes other than `CallStack` and `Synthesize` and I want to make sure this is okay.

Antonio95: Confirmed this is okay now. All places where `execute_function` is called (where the linked check no longer occurs) in `CheckDeployment` or `Synthesize` mode either
 - Are running in a different mode than those, so the check still occurs
 - or don't use the returned `Response` object.

The only exception is the call in this block: https://github.com/ProvableHQ/snarkVM/blob/fix/scalar-outputs/synthesizer/process/src/stack/call/standard.rs#L310-L344. But in that case, the outputs are now constructed manually later in that function, specifically using the `output.value_type()` we were checking against before the changes: https://github.com/ProvableHQ/snarkVM/blob/fix/scalar-outputs/synthesizer/process/src/stack/call/standard.rs#L508-L522.

Antonio95 (commented):
Double checked everything.

Antonio95:
(All flows pass with the exception of the exact same test cases which are already failing in `staging`)

synthesizer/src/vm/mod.rs
vicsn: Removing is fine

vicsn (approved):

synthesizer/src/vm/mod.rs
Antonio95: Done 👌

mohammadfawaz:
If I understand correctly, this fixes a class of programs that didn't use to deploy and now does. Can you explain further why this doesn't need a consensus version check? Are we worried about a window where some nodes have upgraded and others haven't?

synthesizer/src/vm/tests/test_v15/scalar_outputs.rs
```diff
@@ -0,0 +1,95 @@
...
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+use super::*;
+
+// Tests that a function which receives a Field/Group/Address input, casts it to
+// a Scalar and outputs the latter, deploys correctly.
+#[test]
```
mohammadfawaz: A few extra test suggestions: 
- Cross-function `call` returning a scalar: function A calls function B where B outputs a scalar (`call.dynamic` too).
- Nested scalar (struct/array/record entry).
- Negative test that the new `circuit_matches_value_type` still rejects a type-mismatched value.
- Test execution here, not just deployment.

synthesizer/process/src/stack/helpers/type_matching.rs
```diff
@@ -0,0 +1,391 @@
...
+        $matches_future_internal:ident,
+    ) => {
+        // Checks that the given stack value matches the layout of the register type.
+        pub(super) fn $matches_register_type $($declaration_generics)*(
+            stack: &impl StackTrait<N>,
+            value: &$($type_path)::+::Value<$type_generic>,
+            register_type: &RegisterType<N>,
+        ) -> Result<()> {
+
+            let value_kind = match value {
```
mohammadfawaz: `value_kind` is only used in the final `bail!` arm but it's built on every call. Maybe move its definition into the `_ =>` arm?

synthesizer/process/src/stack/execute.rs
```diff
@@ -591,7 +601,12 @@ impl<N: Network> Stack<N> {
         // If the circuit is in `Authorize` mode, then save the transition.
         if let CallStack::Authorize(_, _, authorization) = registers.call_stack_ref() {
             // Construct the transition.
-            let transition = Transition::from(&console_request, &response, &output_types, &output_registers)?;
+            let transition = Transition::from(
+                &console_request,
+                console_response.as_ref().unwrap(),
```
mohammadfawaz: Maybe justify why this unwrap is safe (an expect maybe)

synthesizer/process/src/stack/execute.rs
```diff
@@ -617,7 +632,12 @@ impl<N: Network> Stack<N> {
             registers.ensure_console_and_circuit_registers_match()?;
 
             // Construct the transition.
-            let transition = Transition::from(&console_request, &response, &output_types, &output_registers)?;
+            let transition = Transition::from(
+                &console_request,
+                console_response.as_ref().unwrap(),
```
mohammadfawaz: Same here. `expect` instead of `unwrap`?

Antonio95:
> If I understand correctly, this fixes a class of programs that didn't use to deploy and now does. Can you explain further why this doesn't need a consensus version check? Are we worried about a window where some nodes have upgraded and others haven't?

Added a comment to that effect to the PR description, as discussed.

Antonio95:
@mohammadfawaz I've made the following changes:
- Small suggestion implementations: changed the two `unwrap`s you mentioned to `expect`s and moved the `value_type` computation to the `_ =>` arm.
- Run `cargo update`
- Moved the test file `scalar_outputs.rs` from `test_v15/` to `test_v18/` (for timing-tracking purposes, even if the changes are not consensus-gated) and adding some of the test cases you requested:
  - Scalar read from an array
  - Scalar read from a record
  - Scalar read from a dynamic record (with `get.record.dynamic`)
  - Scalar read from a struct
  - Executions of all deployed functions. Even though in principle an execution should never be problematic (because a scalar there must really have a correct underlying base-field element in order for the proof to verify, and the problem this PR fixes is ejection of wrongly sampled Scalars), I agree testing it is a good sanity check.
   
  I haven't added any "Cross-function call returning a scalar" tests because the issue this PR addresses relates to verification of individual-function circuits and therefore call flows in actual executions are, in principle, unrelated.
 - Separately, relatively thorough testing of the new `console_matches_register_type` (in `synthesizer/process/src/stack/helpers/stack_trait.rs::test_console_matches_register_type`) and `circuit_matches_register_type` (in `synthesizer/process/src/stack/registers/registers_circuit.rs::test_circuit_matches_register_type`).

Let me know if you are okay with the introduced tests, I think we're close to done here!

mohammadfawaz (approved):
Great! Thank you. Before merging, do you mind checking if the `check-cargo-semver-checks` failure is something we should worry about?

Antonio95:
I have checked (the flow takes a while to run) and the semver checks are
1) unrelated to this PR 
2) already present in recent `staging` commits
3) not immediate to fix

So I think we can leave them as far as this PR is concerned

PR #3259, opened 2026-05-14
[Feature] Add support for hotswapping committe in devnets

kaimast:
We recently introduced hotswapped committees for devnets in snarkOS. However, that implementation was only added to the BFT logic, not BlockSync. 

On devnets, I observed that, if a node fell behind or started slightly later than others, it would not be able to sync properly. The cleanest way to fix this is to make snarkVM itself aware of the hot swap and at which height it happened.
Being able to explicitly set the hotswap height also enables restarting nodes in devnet tests without problems.

For this behavior to be enabled, snarkvm needs to be compiled with the `devnet` feature and the ledger must be initialized using `Ledger::load_with_dev_committee`.

### DIFF
--- a/ledger/Cargo.toml
+++ b/ledger/Cargo.toml
@@ -109,9 +109,9 @@ test-helpers = [
 test_targets = [
   "snarkvm-console/test_targets",
 ]
-timer = [ "aleo-std/timer" ]
+dev-committee = []
 dev-print = [ "snarkvm-utilities/dev-print" ]
-
+timer = [ "aleo-std/timer" ]
 
 [dependencies.snarkvm-console]
 workspace = true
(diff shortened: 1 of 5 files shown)


ledger/Cargo.toml
```diff
@@ -109,6 +109,7 @@ test-helpers = [
 test_targets = [
   "snarkvm-console/test_targets",
 ]
+devnet = []
```
vicsn: Can we follow existing naming convention and call this e.g. `test_committee`

vicsn (changes requested):

ledger/Cargo.toml
ljedrz: or `dev_committee`, but yeah, it needs to be more precise

ljedrz (commented):
So far so good; do we call `load_with_dev_committee` with `Some` anywhere?

vicsn (approved):

PR #3265, opened 2026-05-21
Ensure awaits.len() == futures.len() - 1

vicsn:
## Motivation

As a result of an earlier bug report, thought of adding this defense in depth check.

While adding this, also took a full day to research soundness of Dynamic Future handling, and could not find any issues.

## Test Plan /  Backwards compatibility

Besides CI, a sync test could reveal whether this invariant was hypothetically broken in the wild. Because it operates on stateful finalize scope, we can't easily test this in parallel, so I'll just batch it up into our regular E2E sync testing before the next release.

### DIFF
--- a/synthesizer/process/src/finalize.rs
+++ b/synthesizer/process/src/finalize.rs
@@ -299,8 +299,10 @@ impl<'a, N: Network> ProcessExclusiveGuard<'a, N> {
             // Check that the total number of `Await` commands evaluated during
             // finalization matches the number of `Future`s defined in the
             // execution's transitions' outputs.
-            let total_futures =
-                execution.transitions().filter(|t| t.outputs().iter().any(|o| o.future().is_some())).count();
+            let total_futures = execution
+                .transitions()
+                .filter(|t| t.outputs().last().and_then(|output| output.future()).is_some())
+                .count();
             let expected_total_awaits = total_futures.saturating_sub(1);
             if total_awaits != expected_total_awaits {
                 indexed_finalize_bail!(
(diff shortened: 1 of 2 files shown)


synthesizer/process/src/finalize.rs
```diff
@@ -360,8 +377,10 @@ fn finalize_fee_transition<N: Network, P: FinalizeStorage<N>>(
         false => HashMap::new(),
     };
 
-    // Finalize the transition.
-    finalize_transition(state, store, stack, fee, call_graph, Default::default())
+    // Finalize the transition. The fee path has no awaits, so the total_awaits count is unused here.
+    let (finalize_operations, _total_awaits) =
+        finalize_transition(state, store, stack, fee, call_graph, Default::default())?;
```
mohammadfawaz, line 384: Maybe worth asserting that `_total_awaits` is 0 here.

synthesizer/process/src/finalize.rs
```diff
@@ -291,9 +293,24 @@ impl<'a, N: Network> ProcessExclusiveGuard<'a, N> {
...
             // of calls matches the number of transitions.
-            let mut finalize_operations =
+            let (mut finalize_operations, total_awaits) =
                 finalize_transition(state, store, &stack, transition, call_graph, dynamic_future_to_future)?;
 
+            // Check that the total number of `Await` commands evaluated during
+            // finalization matches the number of `Future`s defined in the
+            // execution's transitions' outputs.
+            let total_futures =
+                execution.transitions().filter(|t| t.outputs().iter().any(|o| o.future().is_some())).count();
```
mohammadfawaz: I guess technically it doesn't matter since we're expecting at most a single `Future` output anyways, but in other places we do `transition.outputs().last().and_then(|output| output.future())` instead of using `any(..)`.

mohammadfawaz (commented):
Does this pose a forking risk given that it's not gated by a consensus version? I realize gating this check in `Process` is not possible since `Process` is consensus version agnostic, which leaves us with changing this to a static check in `Program` but I'm not sure that's easy to do given that some awaits could be conditional (behind branch instructions). What are your thoughts on this?

vicsn:
> Does this pose a forking risk given that it's not gated by a consensus version?

Well yes, but also if anyone manages to breach this particular defense in depth check then they can potentially mint money out of thin air. So if our regression syncing tests pass, we'll be sure no one managed to craft such a pathway, and I don't think we have to consensus gate.

synthesizer/process/src/finalize.rs
vicsn: 901be22a0

synthesizer/process/src/finalize.rs
vicsn: 901be22a0

mohammadfawaz (approved):
LGTM! Thanks.

PR #3266, opened 2026-05-21
history feature optimization

vicsn:
## Motivation

I understood you're already using this

### DIFF
--- a/ledger/store/src/program/finalize.rs
+++ b/ledger/store/src/program/finalize.rs
@@ -311,15 +311,8 @@ pub trait FinalizeStorage<N: Network>: 'static + Clone + Send + Sync {
                 self.mapping_update_map()
                     .insert((program_id, mapping_name, key.clone(), current_height), value.clone())?;
 
-                // Ensure the update height history does not already exist.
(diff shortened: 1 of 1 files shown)


ledger/store/src/program/finalize.rs
```diff
@@ -311,15 +311,8 @@ pub trait FinalizeStorage<N: Network>: 'static + Clone + Send + Sync {
...
                     .insert((program_id, mapping_name, key.clone(), current_height), value.clone())?;
 
-                // Ensure the update height history does not already exist.
-                let height_update_key = (program_id, mapping_name, key.clone());
-                if self.mapping_update_heights_map().contains_key_speculative(&height_update_key)? {
-                    bail!(
-                        "Illegal operation: the history of height updates for '{program_id}/{mapping_name}/{key}' already exists in storage"
-                    );
-                }
-
```
mohammadfawaz, line 321: What was this protecting against exactly? Say we have this for a given mapping. `k` is the key:
```
insert(k) @ height 10 -> remove(k) @ height 20 -> insert(k) @ height 30
```
would this be rejected? And that's no longer the case?

ljedrz (approved):

PR #3270, opened 2026-05-22
Fix timestamp sampling

vicsn:
## Motivation

Fix a regression breaking snarkOS CI introduced by https://github.com/ProvableHQ/snarkVM/pull/3255

AI opted for a new library to prevent adding a dependency.


ljedrz (approved):
LGTM; one note: if we're looking for maximum reproducibility, perhaps the impact of the timestamp could be removed via feature-gating; that being said, the effect might be miniscule

PR #3276, opened 2026-05-24
Await testing backwards compatibility

vicsn:
## Motivation

You were right @mohammadfawaz - ran an E2E sync test and ran into a halt because the total_awaits invariant was violated on testnet for testing/exposition reasons. Now only enabling it from `ConsensusVersion::V15` onwards and will try running it against mainnet without the consensus guard.


mohammadfawaz (commented):
Do we also need to gate
```
    if total_awaits != 0 {
        indexed_finalize_bail!(
            Some((*fee.program_id(), *stack.program_edition())),
            Some(*fee.function_name()),
            "Fees must not have any awaits"
        );
    }
```
on line 388?

vicsn:
> Do we also need to gate on line 388?

Syncing succeeded fine without it, but cleaned it up anyway. Failing CI is unrelated.

mohammadfawaz (approved):
LGTM!

PR #3284, opened 2026-06-02
Sign V2

Antonio95:
This PR
 - Introduces a `Signature::sign_v2` function which adds a prefix to the message being signed.
 - Introduces `Signature::sign_bits_v2` and `Signature::sign_bytes_v2`, which operate similarly to their original (non-v2) counterparts but 1) add the length of the message to the message itself and 2) sign using `Signature::sign_v2` (i.e. with a prefix) under the hood.
 - Introduces `Signature::sign_bits_raw_v2` and `Signature::sign_bytes_raw_v2`, analogous to the above but not including message length (i.e. incorporating 2) but not 1)).
 - Introduces a small restriction in the inputs allowed to `Signature::sign`.
 - Adds relevant tests that the above work as expected and some rejection tests.
 
Note that all `Signature::sign*/verify*` functions have also `PrivateKey::sign*/verify*` wrappers. `Request::sign/verify` are unaffected.

The last clean build is the next-to-last commit at the time of writing: c534e88ca.

This PR is in draft mode until we decide:
 - Whether to deprecate the original `Signature::sign`.
 - If so, what to do about the many uses of it present in snarkVM, snarkOS, the SDK and any other relevant repos.

Some options to handle the resulting complications include:
 - Not marking `sign` as deprecated, and simply documenting/making developers aware `sign_v2` is the preferred option from now on.
 - Marking it as deprecated and doing one of the following:
   - Changing all uses across the relevant repos to `sign_v2`. Note signatures are used in some delicate (e.g. consensus-related) paths.
   - Annotating every current use in the relevant repos with `#[allow(deprecated)]` (deciding on a case-by-case basis whether to switch to `sign_v2` instead)
   - Hiding the deprecation annotation behind the non-presence of a new feature (e.g. `sign_v1`) and updating our `Cargo.toml`s to import the crate with that flag. This makes it less likely external developers would use the deprecated function by mistake (they would have to activate the feature flag in the dependency). As a slightly negative aspect of this approach, this means future developers of our own repos would have to be aware that `sign_v2` is the preferred option from now onwards (since the presence of the feature flag in the import means the deprecation notice would not be displayed).

The last commit at the time of writing, 96e926924, adds the deprecation notice to `Signature::sign` and, for most of its current uses, it switches to sign_v2 or annotates them with `#[allow(deprecated)]`. Note this is only for illustrative purposes.


circuit/account/src/lib.rs
```diff
@@ -72,8 +72,8 @@ pub(crate) mod helpers {
 
         // Generate a signature.
         let message: Vec<_> = (0..num_fields).map(|_| Uniform::rand(rng)).collect();
-        let signature = console::Signature::sign(&private_key, &message, rng).unwrap();
-        assert!(signature.verify(&address, &message));
+        let signature = console::Signature::sign_v2(&private_key, &message, rng).unwrap();
```
vicsn, line 75: Is this arithmetization equivalent / backwards compatible?

console/account/src/signature/verify.rs
```diff
@@ -15,10 +15,110 @@
...
+        // Pack the bits into field elements.
+        match message.chunks(Field::<N>::size_in_data_bits()).map(Field::from_bits_le).collect::<Result<Vec<_>>>() {
+            // TODO (Antonio) re-introduce
+            // Ok(fields) => self.verify_v2(address, &fields),
+            Ok(fields) => {
+                for f in fields.iter() {
+                    println!("  f: {f}");
+                }
+                self.verify_v2(address, &fields)
+            }
```
mohammadfawaz, line 109: Leftover debug code?

console/account/src/signature/verify.rs
```diff
@@ -142,16 +253,88 @@ mod tests {
...
+            assert!(!signature_v1.verify_bits_v2(&address, &message));
+            assert!(!signature_v2.verify_bits(&address, &message));
+            assert!(!signature_v2.verify_bits_raw_v2(&address, &message));
+            assert!(!signature_raw_v2.verify_bits_v2(&address, &message));
+        }
+        Ok(())
+    }
+
+    #[test]
+    fn test_sign_and_verify_bits_v2_padding() -> Result<()> {
```
mohammadfawaz, line 287: No test would catch a change to `SIGNATURE_V2_PREFIX` or the length encoding so maybe worth checking against hardcoded values?

console/account/src/lib.rs
```diff
@@ -215,10 +232,23 @@ mod tests {
...
+            let expected_signature_v2 = private_key.sign_bits_v2(&message, &mut rng).unwrap();
+
+            // Serialize
+            let expected_string = &expected_signature_v2.to_string();
+            let candidate_string = serde_json::to_string(&expected_signature).unwrap();
+            assert_eq!(expected_string, serde_json::Value::from_str(&candidate_string).unwrap().as_str().unwrap());
+
+            // Deserialize
+            assert_eq!(expected_signature_v2, serde_json::from_str(&candidate_string).unwrap());
+            assert_eq!(expected_signature, Signature::<CurrentNetwork>::from_str(expected_string).unwrap());
```
mohammadfawaz, line 246:
```suggestion
            let candidate_string = serde_json::to_string(&expected_signature_v2).unwrap();
            assert_eq!(expected_string, serde_json::Value::from_str(&candidate_string).unwrap().as_str().unwrap());

            // Deserialize
            assert_eq!(expected_signature_v2, serde_json::from_str(&candidate_string).unwrap());
            assert_eq!(expected_signature_v2, Signature::<CurrentNetwork>::from_str(expected_string).unwrap());
```

console/account/src/lib.rs
```diff
@@ -157,10 +158,15 @@ mod tests {
             let signature = private_key.sign_bits(&message, &mut rng).unwrap();
             let verification = signature.verify_bits(&address, &message);
             assert!(verification);
+            let signature_v2 = private_key.sign_bits_v2(&message, &mut rng).unwrap();
+            let verification_v2 = signature_v2.verify_bits_v2(&address, &message);
+            assert!(verification_v2);
         }
+
```
mohammadfawaz, line 165:
```suggestion
        }
```

PR #3288, opened 2026-06-04
[Feat] Add a compile-time flag for credits-only history

ljedrz:
This PR adds a new compile-time flag - `HISTORY_CREDITS_ONLY` - which, if present, causes the `history` feature to become restricted to `credits.aleo`.

### DIFF
--- a/ledger/store/src/program/finalize.rs
+++ b/ledger/store/src/program/finalize.rs
@@ -44,6 +44,9 @@ use std::sync::{Arc, atomic::AtomicU32};
 #[cfg(any(feature = "history", feature = "history-staking-rewards", feature = "slipstream-plugins"))]
 use std::{borrow::Cow, sync::atomic::Ordering};
 
+#[cfg(feature = "history")]
+const CREDITS_HISTORY_ONLY: bool = option_env!("HISTORY_CREDITS_ONLY").is_some();
+
 /// Serialized form of a mapping replacement, captured before storage consumes the entries.
 #[cfg(feature = "slipstream-plugins")]
 struct SerializedMappingEntries {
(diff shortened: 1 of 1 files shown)


ledger/store/src/program/finalize.rs
```diff
@@ -305,22 +305,26 @@ pub trait FinalizeStorage<N: Network>: 'static + Clone + Send + Sync {
...
-
-                // Ensure the update height history does not already exist.
-                let height_update_key = (program_id, mapping_name, key.clone());
-                if self.mapping_update_heights_map().contains_key_speculative(&height_update_key)? {
-                    bail!(
-                        "Illegal operation: the history of height updates for '{program_id}/{mapping_name}/{key}' already exists in storage"
-                    );
-                }
+                if !(option_env!("HISTORY_CREDITS_ONLY").is_some()
+                    && program_id != ProgramID::from_str("credits.aleo")?)
```
vicsn: This seems insufficient to me, we want two criteria to be true:
1. Store all history if HISTORY_CREDITS_ONLY is not set
2. Store only credits.aleo history if HISTORY_CREDITS_ONLY is set

The way I'm reading it, it doesn't satisfy either:
```
HISTORY_CREDITS_ONLY: Some && credits.aleo: true => false && false ==> no update
HISTORY_CREDITS_ONLY: None && credits.aleo: true => true && false ==> no update
HISTORY_CREDITS_ONLY: Some && credits.aleo: false => false && true ==> no update
HISTORY_CREDITS_ONLY: None && credits.aleo: false => true && true ==> update all but credits.aleo
```

Might the following be better (perhaps with negligible suboptimal performance):
```
const credits_only = option_env!("HISTORY_CREDITS_ONLY").is_some();
const is_credits = program_id == ProgramID::from_str("credits.aleo")?;
if !credits_only || (credits_only && is_credits) {
  // store shit
}
```

vicsn (changes requested):

ledger/store/src/program/finalize.rs
ljedrz: changed `!(a && !b)` to `!a || b` in https://github.com/ProvableHQ/snarkVM/pull/3288/commits/f4b69d5d6d83c8d605efa744cf4393df14110055

PR #3293, opened 2026-06-05
Future-proof `Scalar` conversion to bits

Antonio95:
(Edit: Converted to draft pending investigation on the possibility of bad-circuit construction)

A `Scalar` circuit value is stored as an underlying `Field` (i.e. base-field) element. No range checks (i.e. checks that the bit-decomposition of the `Field` corresponds to the canonical representation of a `Scalar` and not, say, to a larger value) occur upon injection. These checks do happen whenever `Scalar::to_bits` is called in-circuit (which is also triggered by many other circuit operations)

This is not an issue today because the Aleo-instructions language (used to construct transition circuits), record translation and record inclusion all result in circuits satisfying the following: every `Scalar` injected into the circuit is converted `to_bits` (and therefore range-checked) at some later point during synthesis. As an example, private and public transition inputs are wrapped in a `Plaintext` that gets converted `to_bits` when the circuit processes the input, i.e. hashes it (public) or encrypts it (private).

This PR future-proofs synthesis against failure to convert to bits (i.e. to range-check) all `Scalar`s. See also the note at the end of the next section for potential independent uses this PR could have in the future.

### Solution

- Each time a `Scalar` is injected into a circuit, it is added to a tracking set.
- Each time an in-circuit `Scalar` is converted to bits (which might in turn be caused by many higher-level operations), it is removed from the tracking set.
- During synthesis of transition, translation and inclusion (but not: inclusion V0) circuits, right before ejection of the R1CS or assignment (`eject_r1cs_and_reset` / `eject_assignment_and_reset`), all `Scalar`s not yet converted to bits are so (and i.p. are range-checked).

The reason for this flow as opposed to simply range-checking upon injection is that the latter would not be backwards-compatible (right now, many deployed circuits perform some operations between `Scalar` injection and whatever operation it is which triggers `to_bits`), but the implemented flow is (cf. next section).

Going into some more detail, it was not possible to have the tracking set simply contain the variable identifier of still-unconverted `Scalar`s: because tracking occurs at a lower level than most of our Aleo-type-specific circuit machinery, it is not aware of the existence of a `to_bits` function it can call directly. For that reason, when tracking a new `Scalar`, an abstract function is stores alongside it, which is always `to_bits` in the case of interest. This abstract function which the lower-level knows nothing about is what is called in `convert_unconverted_values` before ejection.

**Note:** The `*_unconverted_values` machinery implemented can also be used in the future whenever we need to be able to a) track circuit values, b) untrack circuit values, and c) guarantee that a certain closure is called before the end of synthesis for all still-untracked values.

### Backwards-compatibility

It has been checked that all `mainnet`, test and `canary` circuits in the regressions repository already convert all of their `Scalar`s to bits. In particular, their synthesis remains unaffected after the implemented change, the original certificates also validate and no redeployments are needed.

This means the change does not need to be version guarded.

### Tests

The main checks were carried out in a separate repo and consisted of the aforementioned regression checks that the updated synthesis doesn’t affect existing transition/translation circuits (or the inclusion one).

A new small test has been added: `test_unconverted_scalar_rejection`, which checks `eject_r1cs_and_reset` and `eject_assignment_and_reset` indeed reject circuits with still-unconverted `Scalar`s. This fact was also temporarily tested while running the regression tests.

### DIFF
--- a/synthesizer/process/src/stack/execute.rs
+++ b/synthesizer/process/src/stack/execute.rs
@@ -567,6 +567,9 @@ impl<N: Network> Stack<N> {
             }
         }
 
+        // Convert all tracked, still-unconverted values to bits.
+        A::convert_unconverted_values();
+
         // Eject the circuit assignment and reset the circuit.
         let assignment = A::eject_assignment_and_reset();
 
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -1290,6 +1290,7 @@ workflows:
       or pipeline.git.branch == "testnet"
       or pipeline.git.branch == "mainnet"
       or pipeline.git.branch == "ensure_finalize_scopes_match"
+      or pipeline.git.branch == "scalar_conversion"
     jobs:
       - check-unused-dependencies # This can be cleaned up before releases
       - check-cargo-semver-checks # This can be cleaned up before releases
--- a/circuit/environment/src/canary_circuit.rs
+++ b/circuit/environment/src/canary_circuit.rs
@@ -286,8 +286,9 @@ impl Environment for CanaryCircuit {
 
     /// Returns the R1CS circuit, resetting the circuit.
     fn eject_r1cs_and_reset() -> R1CS<Self::BaseField> {
-        // Clear the unconverted-scalars diagnostic.
-        Self::clear_unconverted_scalars();
+        // Ensure all Scalars have been converted to bits.
+        Self::check_unconverted_values().unwrap();
+
         CANARY_CIRCUIT.with(|circuit| {
             // Reset the witness mode.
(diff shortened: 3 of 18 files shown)


synthesizer/process/src/stack/execute.rs
```diff
@@ -567,6 +567,9 @@ impl<N: Network> Stack<N> {
             }
         }
 
+        // Convert all tracked, still-unconverted values to bits.
+        A::convert_unconverted_values();
```
mohammadfawaz, line 571: This runs after the is_satisfied() check above, so any constraint added by the conversion here would not be covered by that check and would only fail at prove time. Should the conversion happen before is_satisfied() so a newly added range check is still gated by it?

PR #3294, opened 2026-06-05
feat(synthesizer): evaluate views against the program edition live at the queried height

### LINKED ISSUE #3292: [Feature] Support offchain API calls to view functions on historic program deployments
## 🐛 Bug Report

Currently we only use the most recent deployment / Stack.

mohammadfawaz:
Historic view queries (`--features history`) evaluated against the *latest* program edition, so after a program upgrade a view at an earlier height ran against the wrong body.

`VM::evaluate_view_at_height` now resolves the program edition that was live at the queried `height` and evaluates the view against that body. Mapping reads were already pinned to `height`; this closes the remaining gap where the body always came from the latest edition. All view-query entry points now behave consistently: they query against the historic program version.

**Breaking change:** removes the lower-level `Process::evaluate_view_at_height`. Resolving the edition live at a height requires the deployment and block stores, which only the `VM` has — so a `Process`-level entry point can't produce a correct historic result. Callers that already hold a resolved `Stack` can use the free `evaluate_view_at_height` function, which is unchanged.

Closes #3292

### DIFF
--- a/synthesizer/src/vm/mod.rs
+++ b/synthesizer/src/vm/mod.rs
@@ -345,13 +345,55 @@ impl<N: Network, C: ConsensusStorage<N>> VM<N, C> {
     /// snarkOS calls this with `current_block_height()` for "latest", or any earlier height
     /// for historic views. `height` must satisfy `height <= current_block_height()`.
     ///
-    /// Caveat: the `Stack` itself uses interior mutability, so a concurrent redeploy of the
-    /// same program could perturb its structural caches mid-view. Mapping values are
-    /// snapshot-consistent at `height`; program structure is not. Known gap; a future
-    /// `StackSnapshot`-style fix would close it.
+    /// The view body is taken from the program edition live at `height`. For the latest edition regardless of
+    /// `height`, use [`VM::evaluate_view_at_height_using_latest_edition`].
     #[cfg(feature = "history")]
     #[inline]
-    pub fn evaluate_view_at_height(
+    pub fn evaluate_view_at_height_using_historic_edition(
+        &self,
+        program_id: impl TryInto<ProgramID<N>>,
+        view_name: impl TryInto<Identifier<N>>,
+        inputs: Vec<Value<N>>,
+        height: u32,
+    ) -> Result<Vec<Value<N>>> {
+        let program_id = program_id.try_into().map_err(|_| anyhow!("Invalid program ID"))?;
+        let view_name = view_name.try_into().map_err(|_| anyhow!("Invalid view function name"))?;
+        let state = self.finalize_state_for_block(height)?;
+        let edition = self.resolve_program_edition_at_height(&program_id, height)?;
+        let latest_stack = self.process.get_stack(program_id)?;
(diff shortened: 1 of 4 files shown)


synthesizer/src/vm/mod.rs
```diff
@@ -345,13 +345,55 @@ impl<N: Network, C: ConsensusStorage<N>> VM<N, C> {
...
     ///
-    /// Caveat: the `Stack` itself uses interior mutability, so a concurrent redeploy of the
-    /// same program could perturb its structural caches mid-view. Mapping values are
-    /// snapshot-consistent at `height`; program structure is not. Known gap; a future
-    /// `StackSnapshot`-style fix would close it.
+    /// The view body is taken from the program edition live at `height`. For the latest edition regardless of
+    /// `height`, use [`VM::evaluate_view_at_height_using_latest_edition`].
     #[cfg(feature = "history")]
     #[inline]
-    pub fn evaluate_view_at_height(
```
mohammadfawaz, line 354: I realize this is a breaking change and I'm not thrilled about that, but I thought it's worth it since the old function name is now misleading or at least unclear.

synthesizer/src/vm/mod.rs
```diff
@@ -345,13 +345,55 @@ impl<N: Network, C: ConsensusStorage<N>> VM<N, C> {
...
+        )
+    }
+
+    /// As [`VM::evaluate_view_at_height_using_historic_edition`], but takes the view body from the latest
+    /// loaded program edition instead of resolving the edition live at `height`. Mapping reads are still
+    /// pinned to `height`. Cheaper (no edition resolution or stack rebuild); after an upgrade the two entry
+    /// points can differ for the same `height`. Available only with `--features history`.
+    #[cfg(feature = "history")]
+    #[inline]
+    pub fn evaluate_view_at_height_using_latest_edition(
```
vicsn: I think we can get rid of this function, we have the functionality now to accurately evaluate at a historic height so let's just use it.

However, do we also have a non-history gated function, e.g. something like `evaluate_view_at_tip` or `evaluate_current_view`?

vicsn (changes requested):

synthesizer/src/vm/mod.rs
vicsn: I think evaluate_view_at_height was fine. Especially if we get rid of the other historic function

mohammadfawaz: https://github.com/ProvableHQ/snarkVM/pull/3294/commits/b18815a441f0c5300b07b0637f5d607550c1ed7a

synthesizer/src/vm/mod.rs
mohammadfawaz: > However, do we also have a non-history gated function, e.g. something like evaluate_view_at_tip or evaluate_current_view?

The challenge with this was the possibility of the block heigh incrementing in the middle of evaluating a view, which is why we only enabled historic queries.

vicsn (approved):

PR #3295, opened 2026-06-08
Authorize multiple requests coming from a root call

Antonio95:
This PR introduces the function ~`authorize_multiple_requests`~ `authorize_requests`, which takes in a vector of `Request`s and produces the corresponding `Authorization` by exploring the call graph and computing the resulting `Transition`s without needing a private key.  Throughout this graph exploration, the stack checks the `Request`s it encounters match the current state of the registers (most notably, the target of and inputs to each call).

`authorize_requests ` assumes that

- the `Request`s all come from a single root call and appear in DFS pre-order (i.e. as present in e.g. an `Authorization` produced by the usual evaluation/execution methods).
- the `Request`s are fully populated and correct. In particular, the program and function, `tvk`, `tcm`, input values, input IDs and signature are all consistent.

The tests introduced (which required much more extensive design work than the functionality itself) also show how to use this function in conjunction with other parts of the VM, most notably `sample_authorization`.

### Design

The design is relatively simple. A new

```rust
CallStack::AuthorizeRequests(
  requests: Vec<Request<N>>,
  current_index: Arc<RwLock<usize>>,
  authorization: Authorization<N>,
);
```

variant has been introduced. `requests` is fully populated from the start. `current_index` starts at 0 and tracks which request in requests is being explored at the moment (note it must be shared across the entire exploration). `authorization` is handled as in the `CallStack::Authorize` case: it is populated with the `Request`s in the order they are explored (and thus ends up with a copy of `requests`) and the `Transition`s (in post-order) alongside the transition call graph.

When evaluating a `Call` or `CallDynamic` instruction, `current_index` is increased by one and, instead of computing and signing a new `Request` using the input values in the registers, `requests[current_index]` is retrieved and checked to be consistent with the register values (inputs, call targets, etc.; note conversion to the callee view is needed in the case of dynamic calls before comparing) and to have a correctly computed signature and input IDs. Most other parts of `CallStack::AuthorizeRequests` mirror `CallStack::Authorize`.

### Tests

Since verifying the new functionality requires complex graphs and `Request`s (especially with different types of inputs), instead of adding new tests for it, it has been incorporated into many `test_v14` and `test_v15` test cases. Specifically, the check detailed below is now performed in every `add_and_test_with_costs` call where the `inputs` argument is `Some`, which occurs 193 times at the time of writing.

The expanded test machinery does the following:

1. Start with the full `Execution` computed from the usual `vm.execute` machinery which is verified by the VM and added to a block as before.
2. Call `sample_authorization` using the information about the root call and its inputs, obtaining an `Authorization` which includes a list of mocked `Request`s.
3. Retrieve from the execution transaction the `tvk`s of the original `Request`s using the public `tpk` and the caller’s view key.
4. Use the (correct) `tvk`s to populate each of the mocked `Requests` with correct values, including computing input IDs and a `Request` signature. Note the latter required adding a method mirroring `Request::sign` which receives the `tvk` instead of deriving it from a randomly sampled `r`. That new method, alongside most of the machinery needed for this new test flow, is confined to the file `synthesizer/src/vm/tests/test_v14/authorize_requests_helpers.rs`.
    
    Note that all input/output values of the mocked `Request` can be retained (they are already correct in the mocked object) except for those that depend on the `tvk`: output `Record`s of any transition in the execution (which contain a `tvk`-dependent `nonce`; the payload itself is correct), and input `Record`s, `RecordsExternal` and `ExternalRecord`s corresponding to output `Record`s of previous transactions in the same execution (for the same reason).
    
5. With the correct, fully populated requests, call `authorize_requests ` and check the resulting `Authorization` matches the original one. All `Transition` values except for the `tpk` can be compared directly.

Note point 3 is optional: one could instead sample fresh `tvk`s and would still obtain an `Authorization` which would be consistent with the original one in most regards (call graph, input and output values (except for those detailed in the second part of point 4 above), etc.). However, other values such as input and output IDs, could not be compared with the fresh-`tvk` approach, so step 3 was implemented in this test flow for extra thoroughness.

The 193 calls where `authorize_requests` is checked involve complex call graphs (with up to 11 transitions), many different types of inputs, dynamic calls involving record translation, closures vs. function calls, etc.

Edit: in 35489e09d, I added to the `authorize_requests` test flow the sanity check that the resulting `Authorization` can be proved and verified correctly, which was the case in all instances. It was reverted in the commit immediately after.


synthesizer/process/src/stack/evaluate.rs
```diff
@@ -324,8 +328,10 @@ impl<N: Network> Stack<N> {
         )?;
         finish!(timer);
 
-        // If the circuit is in `Authorize` or `AuthorizeMocked` mode, then save the transition.
-        if let CallStack::Authorize(_, _, authorization) = registers.call_stack_ref() {
+        // If the circuit is in `Authorize`, `AuthorizeMocked` or `AuthorizeRequests` mode, then save the transition.
```
vicsn, line 331: AuthorizeMocked is mentioned in the comment but not the let statement

synthesizer/process/src/stack/authorize.rs
```diff
@@ -205,4 +205,45 @@ impl<N: Network> Stack<N> {
         // Return the authorization.
         Ok(authorization)
     }
+
+    /// Authorizes a number of `Request`s resulting from a root call and populated with correct data
+    /// (`tvk`, input IDs, signature, etc.), checking they are correctly related. The `Request`s
+    /// must be in DFS pre-order (as present, for instance, in `Authorization`s).
+    #[inline]
+    pub fn authorize_multiple_requests<A: circuit::Aleo<Network = N>, R: Rng + CryptoRng>(
```
vicsn:
```suggestion
    pub fn authorize_requests<A: circuit::Aleo<Network = N>, R: Rng + CryptoRng>(
```

synthesizer/process/src/verify_execution/mod.rs
```diff
@@ -620,119 +620,165 @@ impl<N: Network> Process<N> {
...
+            |transition| transition.function_name(),
+        )
+    }
+
+    /// Constructs a call graph keyed by the transition view key (`tvk`) of each request.
+    ///
+    /// This mirrors [`Self::construct_call_graph`], but operates on `Request`s instead of
+    /// `Transition`s. The requests must be provided in the same DFS reverse-post-order as the
+    /// transitions of the corresponding `Execution`.
+    pub fn construct_request_call_graph<'a>(
```
vicsn: Is this used anywhere? It would be great if we can avoid the abstraction/refactor if it's not needed.

vicsn (changes requested):

synthesizer/process/src/stack/evaluate.rs
Antonio95: This comment refers to both that if-left block and one immediately after (which is for `AuthorizeMocked`), both of which are about saving the `Transition` and refer to similar flows (`::Authorize*`). Let me know if you'd like me to split it into two comments, each referring to one if-let block.

synthesizer/process/src/verify_execution/mod.rs
```diff
@@ -620,119 +620,165 @@ impl<N: Network> Process<N> {
...
             // At this point, the algorithm must maintain the following invariant:
             // - The stack is either empty, or the top entry is incomplete.
             match traversal_stack.last_mut() {
-                // If the stack is empty, then push the `transition` to the top of the stack.
+                // If the stack is empty, then push the `object` to the top of the stack.
                 None => {
                     ensure!(
                         counter == 0,
                         "Invalid traversal - execution contains multiple disconnected transition trees"
                     );
```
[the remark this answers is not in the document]
Antonio95: Irrelevant now that this file has been reverted to its `staging` state in b77580761054e180bf132c42062da805cb3680e8

synthesizer/process/src/stack/authorize.rs
```diff
@@ -205,4 +205,45 @@ impl<N: Network> Stack<N> {
...
+        &self,
+        requests: Vec<Request<N>>,
+        rng: &mut R,
+    ) -> Result<Authorization<N>, StackAuthError> {
+        let timer = timer!("Stack::authorize_multiple_requests");
+
+        assert!(!requests.is_empty(), "No requests provided");
```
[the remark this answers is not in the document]
Antonio95: Fixed in b77580761054e180bf132c42062da805cb3680e8 (`ensure!` doesn't work directly because the function returns a `StackAuthError`, used `if` instead)

synthesizer/process/src/verify_execution/mod.rs
Antonio95: That was a stale change from the earlier version where I assumed the `requests` would be in post-order. I've now removed it, and more generally reverted the file to its `staging` state in b77580761. If we ever need this functionality and we happen to remember this commit ever existed, we can pull it from here :)

synthesizer/process/src/stack/authorize.rs
Antonio95: I started on `authorize_requests` and then changed to `authorize_multiple_requests` to make the distinction from the preexisting `authorize_request` more clear. I've now changed it to the `authorize_requests` you propose in 737b8a196, but lmk if I've convinced you into the `authorize_multiple_requests` camp and I will revert.

vicsn (approved):
LGTM! Let's wait for Mo's review

vicsn:
Going ahead with merging already because we need it for the testnet release

PR #3298, opened 2026-06-11
Add `Operand::ComponentChecksum`

mohammadfawaz:
## Summary

Adds a new `Operand::ComponentChecksum(Option<ProgramID>, Identifier)` that resolves to the SHA3-256 checksum of a single named program component's source — a **function, closure, or view** — mirroring the existing whole-program `Operand::Checksum`. This enables pinning the behavior of an individual component (e.g. in a constructor) within an otherwise upgradable program.

The keyword is just `checksum`; the kind is resolved by name, since component names are unique within a program:

- `foo/checksum` — the component `foo` (function/closure/view) in the current program
- `token.aleo/transfer/checksum` — the component `transfer` in an external program

Parsing is unambiguous against the whole-program checksum because a `ProgramID` always carries a network suffix (`.aleo`) and an `Identifier` never does: `foo/checksum` is a component, `token.aleo/checksum` is a program checksum, `token.aleo/foo/checksum` is an external component.

Usable in finalize / constructor / view scope only; rejected in all non-finalize scopes and as struct/array/record members, matching `Checksum`. Gated at `ConsensusVersion::V16`.

## Changes

- `Operand::ComponentChecksum` variant: parse, display, and byte (de)serialization (discriminant 13).
- `to_checksum()` on `FunctionCore`, `ClosureCore`, and `ViewCore`, all delegating to a shared `source_code_checksum` helper alongside `ProgramCore::to_checksum`.
- Resolution in the finalize register loader via a per-`Stack` `component_checksums` cache (computed once at initialization from all functions, closures, and views — no per-load re-hash).
- Deploy-time validation: a `<name>/checksum` whose name is not a function, closure, or view in the target program (or whose external program is not imported) is rejected at deployment instead of failing on every execution.
- `contains_v16_syntax` scans finalize, constructor, **and view** command + output operands, so the operand cannot bypass the pre-V16 deploy gate via a view.

## Tests

`test_v16/component_checksum.rs`:
- Resolves function, closure, and view: constructor records all three `<name>/checksum` values, each verified against the component's `to_checksum()`.
- Checksum is deterministic, ignores unrelated functions, and changes when the component body changes.
- Pin across upgrade: an upgrade that leaves `foo` unchanged is accepted; one that changes `foo` is rejected by the constructor's pinned-checksum assertion.
- Rejected pre-V16 (V15), including when hidden inside a view; accepted at V16.
- Dangling reference (name that is not a component) is rejected at deployment.

### DIFF
--- a/synthesizer/process/src/stack/finalize_types/initialize.rs
+++ b/synthesizer/process/src/stack/finalize_types/initialize.rs
@@ -67,6 +67,8 @@ impl<N: Network> FinalizeTypes<N> {
         }
 
         // Type-check the outputs: each output operand must resolve and match the declared type.
+        // Output operands do not flow through `check_command`, so the component checksum check is repeated here.
+        // Only views have outputs; finalize and constructor scopes have none.
         for output in view.outputs() {
             // If the operand is `Operand::ComponentChecksum`, ensure the referenced component exists.
             if let Operand::ComponentChecksum(program_id, name) = output.operand() {
--- a/synthesizer/process/src/stack/finalize_types/initialize.rs
+++ b/synthesizer/process/src/stack/finalize_types/initialize.rs
@@ -68,6 +68,10 @@ impl<N: Network> FinalizeTypes<N> {
 
         // Type-check the outputs: each output operand must resolve and match the declared type.
         for output in view.outputs() {
+            // If the operand is `Operand::ComponentChecksum`, ensure the referenced component exists.
+            if let Operand::ComponentChecksum(program_id, name) = output.operand() {
+                Self::check_component_checksum(stack, program_id, name)?;
+            }
             let actual = finalize_types.get_type_from_operand(stack, output.operand())?;
             if &actual != output.finalize_type() {
                 bail!(
@@ -228,8 +232,8 @@ impl<N: Network> FinalizeTypes<N> {
     ) -> Result<()> {
         // Check the operands.
         for operand in command.operands() {
-            // If the operand is `Operand::Checksum`, `Operand::Edition`, or `Operand::ProgramOwner` and it contains a program ID,
-            // ensure that the program ID is imported by the current program.
+            // If the operand is `Operand::Checksum`, `Operand::Edition`, or `Operand::ProgramOwner` and it contains a
+            // program ID, ensure that the program ID is imported by the current program.
             match operand {
                 Operand::Checksum(program_id) | Operand::Edition(program_id) | Operand::ProgramOwner(program_id) => {
                     if let Some(program_id) = program_id {
@@ -237,6 +241,10 @@ impl<N: Network> FinalizeTypes<N> {
                             bail!("External program '{program_id}' is not imported by '{}'.", stack.program_id());
                         }
                     }
+                }
+                // If the operand is `Operand::ComponentChecksum`, ensure the referenced component exists.
+                Operand::ComponentChecksum(program_id, name) => {
+                    Self::check_component_checksum(stack, program_id, name)?
                 }
                 _ => {}
             }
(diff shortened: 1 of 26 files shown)


synthesizer/process/src/stack/finalize_types/initialize.rs
```diff
@@ -68,6 +68,10 @@ impl<N: Network> FinalizeTypes<N> {
 
         // Type-check the outputs: each output operand must resolve and match the declared type.
         for output in view.outputs() {
+            // If the operand is `Operand::ComponentChecksum`, ensure the referenced component exists.
+            if let Operand::ComponentChecksum(program_id, name) = output.operand() {
```
vicsn, line 74: Why only check for view outputs, not for view/finalize operations?

synthesizer/program/src/logic/instruction/operand/parse.rs
```diff
@@ -34,6 +34,17 @@ impl<N: Network> Parser for Operand<N> {
                 |(_, index)| Self::AleoGeneratorPowers(index),
             ),
             map(tag("aleo::GENERATOR"), |_| Self::AleoGenerator),
+            // Note that `Operand::ComponentChecksum` (`<name>/checksum`) must be parsed before `Operand::Checksum`,
+            // `Operand::Edition`, `Operand::ProgramOwner`, and `Operand::ProgramID`s, since a component name may
+            // collide with those keywords and may be prefixed with a program ID. This is unambiguous since a program
+            // ID always carries a network suffix (e.g. `.aleo`) while an identifier never does.
```
vicsn: I'm not following, do you have an example of a collision which this particular ordering protects against?

synthesizer/program/src/lib.rs
```diff
@@ -1437,6 +1437,32 @@ impl<N: Network> ProgramCore<N> {
...
+        let command_contains = cfg_iter!(self.functions())
+            .flat_map(|(_, function)| function.finalize_logic().map(|finalize| finalize.commands()))
+            .flatten()
+            .chain(cfg_iter!(self.constructor).flat_map(|constructor| constructor.commands()))
+            .flat_map(|command| command.operands())
+            .any(|operand| matches!(operand, Operand::ComponentChecksum(..)));
+
+        // Check each view command operand and view output operand, since views (allowed since V15) also
+        // accept finalize-context operands.
+        let view_contains = self.views.values().any(|view| {
```
vicsn: Why not chain this into the above section?

synthesizer/src/vm/tests/test_v16/component_checksum.rs
```diff
@@ -0,0 +1,425 @@
...
+    };
+    assert_eq!(expected, stored);
+
+    // Upgrade with `foo` unchanged (adding `bar`). The pinned checksum still matches, so it is accepted.
+    let v1 = program(FOO_UNCHANGED, true);
+    let deployment = vm.deploy(&caller_private_key, &v1, None, 0, None, rng).unwrap();
+    let block = sample_next_block(&vm, &caller_private_key, &[deployment], rng).unwrap();
+    assert_eq!(block.transactions().num_accepted(), 1);
+    assert_eq!(block.transactions().num_rejected(), 0);
+    vm.add_next_block(&block).unwrap();
```
vicsn, line 224: Can we check that bar's checksum is now present in the program's Stack? Can we add a test which updates bar and that bar's checksum is updated as expected in the program's Stack?

vicsn (changes requested):

synthesizer/program/src/logic/instruction/operand/parse.rs
mohammadfawaz: I'll simplify the comments but basically if you have a function called `checksum` and you write `checksum/checksum`. The ordering doesn't matter in most cases though.

synthesizer/process/src/stack/finalize_types/initialize.rs
mohammadfawaz: Right so everything else gets checked in `check_command` earlier. `check_command` already contains a call to `check_component_checksum`. 

Views need special treatment here because they can return but of course `finalize` and `constructor` can't.

mohammadfawaz: https://github.com/ProvableHQ/snarkVM/pull/3298/commits/21f1765521459af6429346bb8c69e9cd8fb61550

synthesizer/program/src/logic/instruction/operand/parse.rs
mohammadfawaz: https://github.com/ProvableHQ/snarkVM/pull/3298/commits/21f1765521459af6429346bb8c69e9cd8fb61550

synthesizer/program/src/lib.rs
mohammadfawaz: https://github.com/ProvableHQ/snarkVM/pull/3298/commits/21f1765521459af6429346bb8c69e9cd8fb61550

synthesizer/src/vm/tests/test_v16/component_checksum.rs
mohammadfawaz: https://github.com/ProvableHQ/snarkVM/pull/3298/commits/21f1765521459af6429346bb8c69e9cd8fb61550

vicsn (approved):

Antonio95 (approved):

PR #3301, opened 2026-06-12
feat(synthesizer): add VM::evaluate_view for latest-height view queries

mohammadfawaz:
Adds `VM::evaluate_view`, a latest-height view evaluator that does not require `--features history` (unlike `evaluate_view_at_height`).

Reads route through `LatestFinalizeStore`, which reads confirmed values, so a concurrent speculation's uncommitted batch is never observed. A view that reads mappings holds a new `finalize_lock` (read) for its evaluation while block commits hold it (write), so it cannot straddle a commit and observe a torn snapshot across reads. A view that reads no mappings skips the lock.

Performance: a minor slow down only happens on if a stateful view (a view that reads mappings) is mid-evaluation at the exact moment a block commits. Otherwise, there should be no impact.

### DIFF
--- a/synthesizer/process/src/lib.rs
+++ b/synthesizer/process/src/lib.rs
@@ -37,6 +37,7 @@ mod evaluate;
 mod execute;
 mod finalize;
 mod view;
+pub use view::evaluate_view_with_stack;
 #[cfg(feature = "history")]
 pub use view::evaluate_view_with_stack_at_height;
 mod verify_deployment;
--- a/synthesizer/process/src/view.rs
+++ b/synthesizer/process/src/view.rs
@@ -18,7 +18,6 @@ use console::{
     network::prelude::*,
     program::{Identifier, Value},
 };
-#[cfg(feature = "history")]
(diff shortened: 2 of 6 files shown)


vicsn:
> Performance: a minor slow down only happens on if a stateful view (a view that reads mappings) is mid-evaluation at the exact moment a block commits.

Really appreciate the prototype but I'm afraid this contention will come to bite us and may cause a lot of debugging work.

If only because a rayon bug caught us twice already which deadlocks when spawning many reads while another thread tries to write.

We also recently spent a decent amount of time reducing nested locking in snarkVM/OS to make things simpler to reason about and this PR would be a step backwards.

The other solution which Ray mentioned (looping until we're sure the view was consistent) has a lot of overhead.

I'd vote for making an issue in snarkOS to deprecate the non-history endpoint which we hastily added once we're sure the history node is available to take the load.

mohammadfawaz:
@vicsn alright sounds good! I'll close for now then and revisit in the future if the need arises.

PR #3309, opened 2026-06-18
Fix history indexing performance

vicsn:
## Motivation

My apologies, I did not manually review any of this. Various node operators reported that the history feature was very slow: 
```
Over hours:we see exactly therocksdb compaction backlogEdgardo reported (0.5 → 0.35, L0 piling up).
Not I/O-bound at steady state:disk ~12% util, sub-ms latency; the finalize thread is inR(notD) — CPU-bound.
Tried, no steady-state gain:RocksDB parallelism patch (increase_parallelism→cores,max_subcompactions,max_open_files=-1),performance governor + EPP, exclusive best-core pinning, CCD cache locality.
```

AI analysis, this PR implements option A:
```
Root cause: O(n) read-modify-write of a growing per-key heights vector
The history feature keeps two maps. The expensive one is MappingUpdateHeightsMap, which stores a Vec<u32> of every height at which a key was ever updated. On every mapping write during finalize, the code does a full read-modify-write of that whole vector.
For any key updated frequently, that vector grows by one u32 per update. Each update then costs, where n = number of prior updates to that key:

get_confirmed → bincode-deserialize the entire vector (O(n))
.into_owned() → clone the whole vector (O(n))
insert → on finish_atomic, bincode-serialize the entire vector again into the shared WriteBatch (O(n) CPU + O(n) bytes written)
So each write is O(n), and cumulatively O(n²) over the life of the chain. n grows with every block, which is exactly why it degrades over hours rather than being bad from cold start.
The smoking gun is that the MappingUpdateMap already stores (program, mapping, key, height) -> value for each update. The heights vector is redundant — it exists only because bincode-encoded keys aren't byte-order-preserving by height (integers are little-endian, confirmed in map.rs create_prefixed_key + bincode::serialize), so you can't cheaply range-scan to find "latest height ≤ target." So the fix is to make appends O(1) and recover the lookup a different way.

Option A — backward compatible, no migration (recommended): Stop writing/reading MappingUpdateHeightsMap entirely. On read, derive the update heights by prefix-scanning the existing MappingUpdateMap ((program, mapping, key, *)) and sorting in memory, then do the same binary-search. Writes drop to O(1) (one insert, no RMW, no growing blob) — kills the CPU blowup and the write amplification immediately. Existing data already has all the per-height entries, so old history DBs keep working with no migration. Cost: view queries get an O(n) prefix scan (rare path, acceptable). The stale heights map is just ignored/lazily dropped.

Option B — optimal reads, needs migration: Drop the heights map and re-encode the height suffix of MappingUpdateMap keys as big-endian so RocksDB key order is numeric, enabling seek_for_prev for O(log n) reads. Best of both (O(1) write, O(log n) read, no write-amp growth), but requires a custom key-encoding helper (new abstraction → approval) and a one-time migration of existing history DBs.


```

## Test Plan

- [ ] We will ask a party to test in production

## Backwards compatibility

This is supposed to be backwards compatible.

### DIFF
--- a/ledger/store/src/helpers/memory/internal/map.rs
+++ b/ledger/store/src/helpers/memory/internal/map.rs
@@ -380,6 +380,29 @@ impl<
     fn values_confirmed(&'a self) -> Self::Values {
         self.map.read().clone().into_values().map(Cow::Owned)
     }
+
+    ///
+    /// Returns the confirmed keys whose serialized form begins with the serialized `prefix`.
+    ///
+    fn get_keys_confirmed_with_prefix<Q>(&'a self, prefix: &Q) -> Result<Vec<K>>
+    where
+        Q: Serialize,
+    {
+        // Serialize the prefix; the BTreeMap keys are sorted by their serialized bytes.
+        let prefix = bincode::serialize(prefix)?;
+
+        // Seek to the prefix and collect every key sharing it.
+        // Note: The 'unwrap' is safe here, because the keys are defined by us.
+        let map = self.map.read();
+        let mut keys = Vec::new();
+        for (key, _) in map.range(prefix.clone()..) {
+            if !key.starts_with(&prefix) {
+                break;
+            }
+            keys.push(unchecked_deserialize(key)?);
+        }
+        Ok(keys)
+    }
 }
 
 impl<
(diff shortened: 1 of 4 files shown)


ljedrz (approved):
LGTM 👍.

vicsn:
Already merged into mainnet

PR #3310, opened 2026-06-20
Extend `sample_authorization` to facilitate multi-request flows

Antonio95:
This PR introduces `sample_authorization_extended`, which returns the same mocked authorization as `sample_authorization` alongside some extra information, most notably record-tracking information which allows, for instance, an authorizer-signer pair to interactively populate the mocked requests in a multi-request flow.

As an important example of how to use this, a showcase function `construct_authorization` is provided in the tests which mimics such an authoriser-signer flow. This is tested in `test_transfer_private_3_flow` with a specific execution (`ldgbatcher_p28.aleo/transfer_private_3`, which calls three children which pass records among themselves) - but cf. “Tests” below.

The original `sample_authorization` maintains the prototype and behaviour for backwards compatibility (it is still used e.g. when estimating execution costs), but it is now a wrapper around `sample_authorization_extended` which simply discards the rest of the data (note: the overhead to compute that extra piece of unnecessary data is negligible).

All changes in this PR are backwards-compatible and don't need to be ConsensusVersion-guarded.

### Design

The main changes needed for `sample_authorization_extended` are in `CallStack::AuthorizeMocked`, which now also tracks two mappings: one with information about static records minted  during the execution, and one with information about static, external or dynamic records input to any call in the execution. At the end of `sample_authorization_extended`, the two are collated together in order to obtain the desired tracking mapping: from (index of minting request, output register of the record) to a list of (index of receiving request, input index of the (static/external/dynamic) record). The mechanism used to collate the two original mappings is the record nonce, which is sampled at random when records are minted in `AuthorizeMocked` mode but is then handled consistently (with the made-up value) throughout the entire flow. This solves some technical challenges, most notably the fact that a record can be passed to a callee before being minted (for instance: mint static + cast to dynamic + pass to dynamic call, then output). This nonce-based tracking turns out to be very robust, as indicated in the “Tests” section below.

In the example flow from `construct_authorization`, after each round of authoriser-signer communication, i.e. once the authoriser has received the signature and tvk from the signer and populates the request with correct values, it also uses the record-tracking information (from `sample_authorization_extended`) to patch the nonces of record-like inputs to subsequent requests with new nonces computed from the actual tvk.

### Tests

The test  `test_transfer_private_3_flow` checks that the `construct_authorization` flow results in an authorisation with the expected shape and which, more importantly, verifies (in particular, the signature, input IDs and input values are all correct and consistent, including intra-transaction records).

However, `construct_authorization` has been tested on almost all of `test_v14` and `test_v15`. This important because those tests stretch the boundaries of our dynamic dispatch, translation, call graph and record manipulation machinery with quite complex executions. In all cases (where `add_and_test_with_costs` is called, which is 276 times), construct `construct_authorization` produced an authorization which proved and verified correctly. This extensive testing was done in commmit `1662737f8` and has since been removed to avoid the significant overhead they entail in the already long `test_v1*` work.


synthesizer/process/src/stack/mod.rs
```diff
@@ -114,8 +115,14 @@ pub type Translations<N> = Arc<RwLock<Vec<Vec<(TranslationAssignment<N>, Proving
...
     /// Authorize an `Execute` transaction.
     Authorize(Vec<Request<N>>, Option<PrivateKey<N>>, Authorization<N>),
-    /// Mock an evaluation for cost estimation.
-    AuthorizeMocked(Vec<Request<N>>, Address<N>, Authorization<N>),
+    /// Mock an evaluation.
+    AuthorizeMocked(
+        Vec<Request<N>>,
+        Address<N>,
+        Authorization<N>,
+        Arc<RwLock<HashMap<N::Field, (usize, u64)>>>,
```
vicsn: Use named types

vicsn (changes requested):

synthesizer/process/src/stack/authorize.rs
```diff
@@ -147,6 +150,50 @@ impl<N: Network> Stack<N> {
...
+        &self,
+        address: Address<A::Network>,
+        program_id: ProgramID<A::Network>,
+        function_name: Identifier<A::Network>,
+        inputs: impl ExactSizeIterator<Item = impl TryInto<Value<A::Network>>>,
+        rng: &mut R,
+    ) -> Result<
+        (
+            Authorization<N>,
+            HashMap<(usize, u64), Vec<(usize, usize)>>,
```
vicsn: Can we use named types

synthesizer/process/src/stack/authorize.rs
```diff
@@ -170,14 +217,66 @@ impl<N: Network> Stack<N> {
...
+            if let Some(consumer_requests_and_indices) = input_records.get(nonce_x) {
+                record_tracking.insert(*minter_request_and_register, consumer_requests_and_indices.clone());
+            }
+        }
+
+        // Collect the names of (all) static Record inputs and the program checksums
+        let mut record_names = HashMap::new();
+        let mut program_checksums = HashMap::new();
+
+        for (request_index, request) in authorization.to_vec_deque().iter().enumerate() {
```
vicsn: Can all this logic become a member of authorization?

vicsn (changes requested):

synthesizer/process/src/stack/authorize.rs
```diff
@@ -147,6 +150,50 @@ impl<N: Network> Stack<N> {
...
+    ///    after conversion to an external or dynamic record) passed as the
+    ///    `l_1`-th input to the `k_1`-th request, the `l_2`-th input to the
+    ///    `k_2`-th request, etc.
+    ///  - HashMap<(usize, usize), Identifier<CurrentNetwork>>: record-name
+    ///    information for *all* input static records to any of the resulting
+    ///    requests. Entry `(n, m) -> r_name` in the returned map indicates that
+    ///    the `m`-th input of the `n`-th request in the transaction is a static
+    ///    `Record` with name `r_name`.
+    ///  - HashMap<usize, Field<CurrentNetwork>>: Program-checksum information:
+    ///    entry `n -> c` in the returned map indicates that the `n`-th request
```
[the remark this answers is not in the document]
Antonio95: Obsolete as per af16498e9cd3ff5af06793536043ded7294ec621

synthesizer/process/src/stack/authorize.rs
```diff
@@ -147,6 +150,50 @@ impl<N: Network> Stack<N> {
...
+    /// `sample_authorization` alongside some extra information necessary to
+    /// populate the mocked `Request`s. These additional outputs are as follows:
+    ///
+    ///  - HashMap<(usize, u64), Vec<(usize, usize)>>: Record-tracking
+    ///    information on records which are both minted by a request in the
+    ///    transaction and received (possibly as external or dynamic) by other
+    ///    requests in the transaction. Entry `(n, m) -> [(k_1, l_1), (k_2,
+    ///        l_2), ...]` in the returned map indicates that the `n`-th
+    ///    transaction output a static record at register `m` which was (possiby
+    ///    after conversion to an external or dynamic record) passed as the
```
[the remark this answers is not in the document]
Antonio95: Fixed in af16498e9cd3ff5af06793536043ded7294ec621

synthesizer/src/vm/tests/test_v16/construct_authorization/construct_authorization_flow.rs
```diff
@@ -0,0 +1,324 @@
...
+                    _ => unreachable!("Tracked record inputs must be of type Record or DynamicRecord"),
+                };
+                corrected_inputs[*consumer_request_index][*consumer_input_index] = corrected;
+            }
+        }
+    }
+
+    // ********************************************************************************************
+    // Step 3: Call authorize_requests to obten correct authorizations
+
```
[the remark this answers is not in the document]
Antonio95: Fixed in af16498e9cd3ff5af06793536043ded7294ec621

synthesizer/process/src/stack/mod.rs
Antonio95: Done in af16498e9cd3ff5af06793536043ded7294ec621, I've called them `RecordFlowTracker` `RecordNameTracker` and `ProgramChecksumTracker` and moved the explanation of what they mean from the function to the new type aliases.

synthesizer/process/src/stack/authorize.rs
Antonio95: Done in af16498e9cd3ff5af06793536043ded7294ec621

synthesizer/process/src/stack/authorize.rs
Antonio95: Replaced the loop by

```
        let record_names = authorization.collect_record_names(self)?;
        let program_checksums = authorization.collect_program_checksums(self)?;
```
in 33e58f0ded343982a6907df6f8bd8751b0ab109e, let me know if that's what you had in mind. I thought it  cleaner to split it into two methods, especially taking into account that the resulting extra work of calling `get_stack_global` twice per request is negligible.

synthesizer/process/src/stack/authorize.rs
```diff
@@ -147,6 +157,23 @@ impl<N: Network> Stack<N> {
         inputs: impl ExactSizeIterator<Item = impl TryInto<Value<A::Network>>>,
         rng: &mut R,
     ) -> Result<Authorization<N>, StackAuthError> {
+        self.sample_authorization_extended::<A, R>(address, program_id, function_name, inputs, rng)
+            .map(|(authorization, _, _, _)| authorization)
+    }
+
+    /// Produces a mocked `Authorization` with the same properties as
+    /// `sample_authorization` alongside some extra information necessary to
+    /// populate the mocked `Request`s.
```
Antonio95: TODO: explain why the extended information may be useful.

synthesizer/process/src/stack/authorize.rs
```diff
@@ -147,6 +157,23 @@ impl<N: Network> Stack<N> {
...
+    #[inline]
+    pub fn sample_authorization_extended<A: circuit::Aleo<Network = N>, R: Rng + CryptoRng>(
```
Antonio95: sample_authorization_with_record_tracking

vicsn (approved):

PR #3313, opened 2026-06-24
Revert anchor times

vicsn:
## Motivation

From activation of ConsensusVersion::V15 June 15th, the cumulative daily total_reward from Solutions dropped by 10x. The PR which influenced the puzzle logic was was https://github.com/ProvableHQ/snarkVM/pull/3194 

Without understanding any of the underlying logic, the only thing we should worry about in this PR is that it won't have unintended side effects, that it "simply" reverts reward logic to what it was before. Sometimes, **lowering** a value in our software is very dangerous and is harder than **increasing** a value.

## Test Plan

I hope it's trivial enough that we don't need extra tests

## Backwards compatibility

Guarded the revert by a consensus version

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -1324,8 +1324,7 @@ workflows:
       or pipeline.git.branch == "canary"
       or pipeline.git.branch == "testnet"
       or pipeline.git.branch == "mainnet"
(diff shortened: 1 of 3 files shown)


mohammadfawaz (approved):
LGTM!

PR #3317, opened 2026-06-30
Fix stacks supplied to the record-existence check

Antonio95:
This PR fixes an issue where the program stacks available to `ensure_records_exist` did not contain all programs necessary for the check. Specifically, that function needs two types of stacks:

- the stacks of the programs in the execution
- the stacks of the programs containing any external closures called in any of the transitions

Only the first kind was being supplied. So, in executions where a stack of the second kind corresponds to a program which is not itself the program of a transition (i.e. of the first kind), execution failed.

**Note:** This PR is in draft mode until a decision is made on ConsensusVersion gating.

### Fix

All direct (i.e. not recursive) imports of the transitions’ programs are also added to the stacks. This always covers all stacks of the second kind above (and possibly more; but fetching the stack corresponds to cloning an `Arc` and is therefore negligible at the numbers involved in proving and verification).

### Testing

Since the source of the issue is evident from inspection of `stack` reads in `ensure_records_exist`, only one test has been added: `test_external_closure_stack_fetching`. This test fails in the commit it was added (`48633cf98`), showing there is an issue; and passes after the implemented fixes.


Antonio95:
Update: now gated behind ConsensusVersion::V18, which is not present in the branch (hence the CI failures). It will be automatically fixed once https://github.com/ProvableHQ/snarkVM/pull/3316 containing V18 is merged. It has been checked that gating behind V17 works and behaves as expected: the test fails at V15 and passes at V17.

synthesizer/process/src/lib.rs
```diff
@@ -517,6 +517,40 @@ impl<N: Network> Process<N> {
...
+            // Fetch those direct imports of the program which have not been fetched before. Note
+            // this cannot be done inside the conditional block above since then direct imports of
+            // direct imports which are themselves called later on would be missed (consider the
+            // transition list [A, B] where A imports B and B imports C).
+            for import_program_id in stack.program().imports().keys() {
+                if !execution_stacks.contains_key(import_program_id) {
+                    execution_stacks.insert(*import_program_id, self.get_stack(import_program_id)?);
+                }
+            }
+        }
```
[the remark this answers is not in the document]
Antonio95: Done (in a better way) in 4c318ee2c

synthesizer/src/vm/tests/test_v15/record_existence.rs
```diff
@@ -903,3 +903,85 @@ fn test_existence_check() {
...
+            call ext_closure_lib.aleo/add_values r0 r1 into r2;
+            output r2 as u64.public;
+
+        constructor:
+            assert.eq true true;
+        ",
+    )?;
+
+    // Initialize the VM at V15.
+    let vm = sample_vm_at_height(CurrentNetwork::CONSENSUS_HEIGHT(ConsensusVersion::V15)?, rng);
```
vicsn: Should this be V18?

vicsn (approved):

synthesizer/src/vm/tests/test_v15/record_existence.rs
Antonio95: Indeed, moved to a new `test_v18` and bumped in the test in f51b22b. Contingent on the CI flow, the PR is now ready.

PR #3318, opened 2026-07-01
Speed up history reads

### LINKED ISSUE #3315: [Bug]   Mapping history lookups are extremely slow for frequently-updated keys
## History Feature Performance

The `history` feature's historical-mapping read path is **O(total number of updates ever made to a key)** and **independent of the requested block height**, making it unusable for frequently-updated mappings. A single `credits.aleo/committee/{validator}/history/{height}` lookup takes **~79 seconds** (returning a 44-byte value), while the same call for a rarely-updated key (e.g. `unbonding`) returns in milliseconds.

Root cause is in `FinalizeStore::get_historical_mapping_value` → `get_mapping_update_heights` (`ledger/store/src/program/finalize.rs`, L808–L869, introduced by commit `09f729f7` / PR #3148 "Revamp historical mapping storage"):

```rust
let mut heights = self.storage.mapping_update_map()
    .get_keys_confirmed_with_prefix(&(program_id, mapping_name, mapping_key))?   // scans EVERY recorded update for this key
    .into_iter().map(|(_, _, _, height)| height).collect::<Vec<_>>();
...
heights.sort_unstable();   // sorts the entire per-key history on every request
```

The requested `height` is only applied *afterward*, in a `binary_search` over the fully-materialized vector — so requesting an early height does the same full scan. Two compounding issues:

1. **No bound on the scan** — it loads every historical update of the key, not just those `<= height`.
2. **Height is encoded little-endian in the map key**, so the prefix scan can't return rows in numeric order and can't early-terminate; it must read all rows and sort in memory every call (acknowledged in the code's own comments).

For `credits.aleo/committee` (rewritten essentially every round), this is on the order of millions of entries → ~79s per request, and it grows unbounded as the chain advances. This is the read-path counterpart to the write-side concerns already raised in #3311 about the same `09f729f7` revamp.

## Steps to Reproduce

1. Build snarkVM (via snarkOS v4.7.5) with `--features "history"` and sync an Aleo mainnet archive node.
2. Query a frequently-updated key, e.g. the `committee` mapping for an active validator, against the local REST endpoint:

   ```bash
   curl -s -o /dev/null -w '%{http_code} %{time_total}s\n' --max-time 120 \
     'http://127.0.0.1:3030/mainnet/program/credits.aleo/mapping/committee/aleo1anfvarnm27e2s5j6mzx3kzakx5eryc69re96x6grzkm9nkapkgpq4vyy5t/history/19576759'
   # -> 200 79.049254s   (44-byte response)
   ```

3. Confirm it's height-independent — all three return ~79s:

   ```bash
   .../committee/aleo1anf...pq4vyy5t/history/1000000     # 200 78.979746s
   .../committee/aleo1anf...pq4vyy5t/history/19576759    # 200 78.968959s
   .../committee/aleo1anf...pq4vyy5t/history/19661400    # 200 79.043978s  (~ chain tip)
   ```

4. Confirm it's specific to high-update-count keys (same node, same endpoint):

   | mapping | current value | `/history/{height}` |
   |---|---|---|
   | `unbonding/aleo16dl5...tvpsz59q4s` | `200 0.000603s` | `200 0.004442s` |
   | `committee/aleo1anf...pq4vyy5t` | `200 0.000492s` | `200 79.049254s` |

## Expected Behavior

A historical-mapping lookup should resolve in roughly constant time regardless of how many times the key has been updated — e.g. an indexed/range seek to the most recent update `<= height` (one read), not a full prefix scan + in-memory sort of the key's entire update history. Expected: low-millisecond response for any `(program, mapping, key, height)`, matching current-value reads and the `unbonding/history` case above. Instead, `committee/history` takes ~79s and degrades further as the chain grows.

## Your Environment

- snarkVM Version: 4.7.3 (git rev `a108e8159c8b26a19fbe25bd51a5de9a645796a3`, as pinned by snarkOS v4.7.5; features `history,history-staking-rewards`)
- Rust Version: 1.88.0
- Computer OS: Ubuntu 24.04

vicsn:
## Motivation

Closes https://github.com/ProvableHQ/snarkVM/issues/3315

Notes:
- The type change (u32 → [u8; 4]) is serialization-neutral in bincode.
- Reads using the old binary-search path look up with .to_le_bytes() and find the correct data.

## Test plan

TODO: write a script which starts a old history node network, upgrades to new history nodes, and queries some data

### DIFF
--- a/ledger/store/src/helpers/memory/internal/map.rs
+++ b/ledger/store/src/helpers/memory/internal/map.rs
@@ -347,6 +347,27 @@ impl<
     }
 
     ///
+    /// Returns the (key, value) pair for the greatest confirmed key ≤ the given key,
+    /// or `None` if no such entry exists in this map.
+    ///
+    /// Uses `BTreeMap::range` for O(log n) performance.
+    ///
+    fn get_floor_confirmed<Q>(&'a self, key: &Q) -> Result<Option<(Cow<'a, K>, Cow<'a, V>)>>
+    where
+        K: Borrow<Q>,
+        Q: PartialEq + Eq + Hash + Serialize + ?Sized,
+    {
+        let key_bytes = bincode::serialize(key)?;
+        // Note: The 'unwrap' is safe here, because the keys are defined by us.
+        Ok(self
+            .map
+            .read()
+            .range(..=key_bytes)
+            .next_back()
+            .map(|(k, v)| (Cow::Owned(unchecked_deserialize(k).unwrap()), Cow::Owned(v.clone()))))
+    }
+
+    ///
     /// Returns an iterator visiting each key-value pair in the atomic batch.
     ///
     fn iter_pending(&'a self) -> Self::PendingIterator {
--- a/ledger/store/src/helpers/memory/program.rs
+++ b/ledger/store/src/helpers/memory/program.rs
@@ -47,12 +47,9 @@ pub struct FinalizeMemory<N: Network> {
     key_value_map: NestedMemoryMap<(ProgramID<N>, Identifier<N>), Plaintext<N>, Value<N>>,
     /// The rejection reason map.
     rejected_reason_map: MemoryMap<Field<N>, RejectedReason<N>>,
-    /// The historical mapping value update map.
+    /// The historical mapping value map (keyed by big-endian block height).
     #[cfg(feature = "history")]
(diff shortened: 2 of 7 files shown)


vicsn:
Superseded by: https://github.com/ProvableHQ/snarkVM/pull/3323

PR #3319, opened 2026-07-01
Avoid translation in the root call

Antonio95:
Note: this PR is in draft more until a few more test cases are added.

In the design of dynamic dispatch, no translation should occur for inputs/outputs of the root call. The introduction of the `Input::RecordwithDynamicID` and `Input::ExternalRecordwithDynamicID` variants means it is the submitter that selects which inputs they tell the verifier correspond to translated `Record`s and `ExternalRecord`s by setting those input variants. Correctness is checked in `to_transition_verifier_inputs`, but only for call *instructions* and, in particular, not for the root call. This PR adds that check. More details are discussed in an internal document.

# Changes

Replaced the first of the two checks (the one in `verify_execution`), which is the only one that applies to the root call, by a more strict type-matching function: the new `matches_type`. For non-root calls it behaves the same as the original check (i.e. it performs the non-strict variant of the check) as we do not know at that point whether each non-root call is dynamic or not. For the root call, it performs the more strict check which, among other things, disallows `Input::RecordwithDynamicID` and `Input::ExternalRecordwithDynamicID`. Note this is consensus-gated to V18, i.e. up until V17 inclusive, the check has has the same behaviour as before.

The check in `to_transition_verifier_inputs`, i.e. at the call site in the caller, hasn’t been modified. This was already robust as implemented. It uses is_type and cannot be modified to `matches_type` since the latter refers to callee view. Modifying `matches_type` so it encompasses both cases would be unnecessarily complicated.

# Tests

At the moment one test has been added: `test_external_record_with_dynamic_id_input_to_root`, which checks that a root call with an `ExternalRecordwithDynamicID` input is rejected. More test cases are coming.


synthesizer/process/src/verify_execution/mod.rs
```diff
@@ -178,25 +178,38 @@ impl<N: Network> Process<N> {
             ensure!(function.inputs().len() == num_inputs, "The number of transition inputs is incorrect");
             ensure!(function.outputs().len() == num_outputs, "The number of transition outputs is incorrect");
 
+            // Obtain the is_dynamic flag used by Input::matches_type and Output::matches_type. At
+            // this point, we only want to enforce strict matching for the root call: we do not know
+            // which of the non-root ones are dynamic or not yet. Strict matching is not applied
+            // before ConsensusVersion::V18.
+            let is_dynamic_for_types =
+                if !reverse_call_graph.contains_key(transition.id()) && consensus_version >= ConsensusVersion::V18 {
```
vicsn: In other places in the code we determine the root_transition as `execution.transitions().last().unwrap();` It will be easier for long-term maintenance to re-use that. Though some tests also use `transaction.transitions().nth(num_transitions - 2)`, not sure why those are valid...

synthesizer/process/src/verify_execution/mod.rs
```diff
@@ -178,25 +178,38 @@ impl<N: Network> Process<N> {
...
+                    Some(false)
```
vicsn: Do I see correctly that we never actually exercise the path where this is `true`? Should we have a test for that code path?

ledger/block/src/transition/output/mod.rs
```diff
@@ -417,6 +417,68 @@ impl<N: Network> Output<N> {
...
+    /// `Record`, `ExternalRecord`, `DynamicRecord`, `RecordWithDynamicID` and
+    /// `ExternalRecordWithDynamicID` outputs are handled according to the value of `is_dynamic`:
+    ///  - `Some(true)` enforces type correspondence assuming the outputs refer to a dynamic call.
+    ///  - `Some(false)` enforces it assuming a static call.
+    ///  - `None` is flexible and makes no assumption about the type of call: both dynamic-call and
+    ///    static-call matching patterns return `true`.
+    ///
+    /// For outputs of an invalid type according to `is_dynamic`, this function returns false
+    /// regardless of the value type.
+    pub fn matches_type(&self, expected_value_type: &ValueType<N>, is_dynamic: Option<bool>) -> bool {
```
vicsn: If this is always about a caller/callee comparison, might be worth renaming as follows. 

And I'm not sure if `matches_type` is appropriate compared to other places where `matches` implies more exact matches. What do you think of `valid_as_argument`?

```suggestion
    pub fn matches_type(&self, callee_value_type: &ValueType<N>, is_dynamic: Option<bool>) -> bool {
```

ledger/block/src/transition/input/mod.rs
```diff
@@ -246,6 +246,67 @@ impl<N: Network> Input<N> {
...
+    /// `ExternalRecord`, `DynamicRecord`, `RecordWithDynamicID` and `ExternalRecordWithDynamicID`
+    /// inputs are handled according to the value of `is_dynamic`:
+    ///  - `Some(false)` enforces type correspondence assuming the inputs refer to a dynamic call.
+    ///  - `Some(true)` enforces it assuming a static call.
+    ///  - `None` is flexible and makes no assumption about the type of call: both dynamic-call and
+    ///    static-call matching patterns return `true`.
+    ///
+    /// For inputs of an invalid type according to `is_dynamic`, this function returns false
+    /// regardless of the value type.
+    pub fn matches_type(&self, expected_value_type: &ValueType<N>, is_dynamic: Option<bool>) -> bool {
```
vicsn: If this is always about a caller/callee comparison, might be worth renaming as follows.

And I'm not sure if `matches_type` is appropriate compared to other places where `matches` implies more exact matches. What do you think of `valid_as_argument`?

```suggestion
    pub fn matches_type(&self, callee_value_type: &ValueType<N>, is_dynamic: Option<bool>) -> bool {
```

vicsn (changes requested):

ledger/block/src/transition/input/mod.rs
```diff
@@ -246,6 +246,67 @@ impl<N: Network> Input<N> {
...
         )
     }
+
+    /// Returns `true` if the input matches the given value type as seen by the callee. `Record`,
+    /// `ExternalRecord`, `DynamicRecord`, `RecordWithDynamicID` and `ExternalRecordWithDynamicID`
+    /// inputs are handled according to the value of `is_dynamic`:
+    ///  - `Some(false)` enforces type correspondence assuming the inputs refer to a dynamic call.
+    ///  - `Some(true)` enforces it assuming a static call.
+    ///  - `None` is flexible and makes no assumption about the type of call: both dynamic-call and
+    ///    static-call matching patterns return `true`.
```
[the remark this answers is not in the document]
Antonio95: Correct, fixed

synthesizer/process/src/verify_execution/ensure_records_exist.rs
```diff
@@ -296,6 +296,8 @@ impl<N: Network> Process<N> {
             .iter()
             .zip_eq(input_registers.iter())
             .filter_map(|(input, register)| {
+                // We do not need to match Input::ExternalRecordWithDynamicID here as that variant
+                // can never occur at the root call.
                 if matches!(input, Input::DynamicRecord(..) | Input::ExternalRecord(..)) {
                     Some((*root_transition_id, *register))
                 } else {
```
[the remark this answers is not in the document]
Antonio95: I'd argue this is actually desired behaviour: if that situation happens, the record-existence check executed early will pass (because it is running on the incorrect assumption that `Input::ExternalRecordWithDynamicID` doesn't occur as the root call) and the subsequent, dynamic/static-strict type checks will detect the incorrect `Input::ExternalRecordWithDynamicID` and reject with the latter reason, which is the meaningful one.

ledger/block/src/transition/input/mod.rs
```diff
@@ -246,6 +246,67 @@ impl<N: Network> Input<N> {
...
+            (Self::Constant(..), ValueType::Constant(..))
+                | (Self::Public(..), ValueType::Public(..))
+                | (Self::Private(..), ValueType::Private(..))
+        ) {
+            return true;
+        }
+
+        if is_dynamic == Some(true) {
+            // Unmatched variants:
+            //  - Self::Record: cannot be the calle's view of a dynamic-call input, so we return
```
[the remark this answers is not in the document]
Antonio95: Fixed

synthesizer/process/src/verify_execution/mod.rs
Antonio95: Okay, I've changed it to what you requested to be consistent and make it easier to identify (although the index is maybe a bit more fragile long-term). And I assume the `num_transitions - 2` case is when you also have the fee transition at the end :)

Antonio95:
Update: when running the tests, I realised we were never strictly testing callee types for static calls. So I have now added the check in the input-to-child-transitions part of `verify_execution` so that it is run both for dynamic and static calls. Note that, for non-root transitions, this is a more strict version of the `matches_type` with `is_dynamic = None` done at the start of all transitions anyway (not at the call site). Still, this is such a cheap check that I think it is worth rejecting early if possible and repeating it with stricter type-matching at the call site.

The tests now check incorrect `[External]RecordWithDynamicID` inputs to non-root transitions coming from static calls are also rejected as they should.

ledger/block/src/transition/input/mod.rs
Antonio95: I've changed it to `valid_as_argument` for `Input` and `valid_as_output` for `Output`, I agree `matches` can be a bit misleading when `is_dynamic = None` (so non-strict enforcement is done)

Antonio95:
Update: I've also run backwards-compatible tests by (locally/temporarily) removing the consensus-version gating and checking all test_v14 and test_v15 tests pass with this stricter type checking, as they should.

ledger/block/src/transition/output/mod.rs
Antonio95: Done 👌

synthesizer/process/src/verify_execution/mod.rs
Antonio95: This is done with the latest changes, cf. https://github.com/ProvableHQ/snarkVM/pull/3319#issuecomment-4917853661. Sometimes testing helps you find more bugs! :)

synthesizer/process/src/verify_execution/mod.rs
```diff
@@ -178,25 +178,39 @@ impl<N: Network> Process<N> {
             ensure!(function.inputs().len() == num_inputs, "The number of transition inputs is incorrect");
             ensure!(function.outputs().len() == num_outputs, "The number of transition outputs is incorrect");
 
+            // Obtain the is_dynamic flag used by Input::valid_as_argument and Output::valid_as_output. At
+            // this point, we only want to enforce strict matching for the root call: we do not know
+            // which of the non-root ones are dynamic or not yet. Strict matching is not applied
+            // before ConsensusVersion::V18.
+            let is_dynamic_for_types = if transition.id() == execution.transitions().last().unwrap().id()
```
mohammadfawaz: Could be worth using `expect` instead of `unwrap` here.

synthesizer/process/src/verify_execution/mod.rs
```diff
@@ -178,25 +178,39 @@ impl<N: Network> Process<N> {
...
+                && consensus_version >= ConsensusVersion::V18
+            {
+                Some(false)
+            } else {
+                None
```
mohammadfawaz: At V18 non-root transitions are checked here with `None` and again strictly in `to_transition_verifier_inputs`. Can we drop the redundant check here for non-root?

ledger/block/src/transition/input/mod.rs
```diff
@@ -246,6 +246,67 @@ impl<N: Network> Input<N> {
...
+        } else {
+            // is_dynamic = None: flexible matching
+            matches!(
+                (self, callee_value_type),
+                (Self::Record(..), ValueType::Record(..))
+                    | (Self::RecordWithDynamicID(..), ValueType::Record(..))
+                    | (Self::ExternalRecord(..), ValueType::ExternalRecord(..))
+                    | (Self::ExternalRecordWithDynamicID(..), ValueType::ExternalRecord(..))
+                    | (Self::DynamicRecord(..), ValueType::DynamicRecord)
+            )
```
mohammadfawaz: Can we do this instead?

```suggestion
            self.is_type(callee_value_type)
```

ledger/block/src/transition/output/mod.rs
```diff
@@ -417,6 +417,68 @@ impl<N: Network> Output<N> {
...
+        } else {
+            // is_dynamic = None: flexible matching
+            matches!(
+                (self, callee_value_type),
+                (Self::Record(..), ValueType::Record(..))
+                    | (Self::RecordWithDynamicID(..), ValueType::Record(..))
+                    | (Self::ExternalRecord(..), ValueType::ExternalRecord(..))
+                    | (Self::ExternalRecordWithDynamicID(..), ValueType::ExternalRecord(..))
+                    | (Self::DynamicRecord(..), ValueType::DynamicRecord)
+            )
```
mohammadfawaz: Can we do this instead?

```suggestion
            self.is_type(callee_value_type)
```

synthesizer/process/src/verify_execution/mod.rs
Antonio95: This line happens inside a `for transition in execution.transitions()` loop, so the `unwrap` is safe. Would you like me to document it is (or do you mean that precisely because it is safe, it should be done with an `expect` and the safety documented should be documented in the error string)?

ledger/block/src/transition/output/mod.rs
Antonio95: Idem

ledger/block/src/transition/input/mod.rs
Antonio95: Good call, done in 9dc471f4ad73459f0d9c973f99855593099f1b4c

synthesizer/process/src/verify_execution/mod.rs
Antonio95: Done in 9dc471f4ad73459f0d9c973f99855593099f1b4c, specifically here: https://github.com/ProvableHQ/snarkVM/blob/fix/translated_type_tracking/synthesizer/process/src/verify_execution/mod.rs#L183-L195.

I think there could be a case for keeping the weak check here to exit early, but it's a pretty weak argument. I like your approach better 👍

synthesizer/process/src/verify_execution/mod.rs
mohammadfawaz: I personally thing `expect` is almost always better than `unwrap` :) It at least makes it clear why the unwrapping is safe.

mohammadfawaz:
I would imagine the `cargo-audit` failures would go away if you merge with `staging`.

synthesizer/process/src/verify_execution/mod.rs
Antonio95: This part is now gone, but I changed the other `unwrap` to an `expect` in ee3080a61ab09876e8acc83d458269e776b48a4e

Antonio95:
All working incl. `cargo-audit` with the exception of the unrelated `semver-check` failure.

vicsn (approved):
LGTM!

synthesizer/process/src/verify_execution/mod.rs
```diff
@@ -183,26 +183,45 @@ impl<N: Network> Process<N> {
...
+            // Before ConsensusVersion::V18, we check the input and output types match the ones
+            // defined in the function under the weaker type equivalence. At ConsensusVersion::V18
+            // and higher, we perform strict type checking for the root transition only: for
+            // non-root transitions, strict type checking happens at the call site in
+            // `to_transition_verifier_inputs` instead.
+            // Note: We only need to check that the variant matches because we already check the
+            // hashes in the `Input::verify` and `Output::verify` functions. Note: The length checks
+            // above already verify that the counts match, so `zip_eq` here is safe and acts as a
+            // defense-in-depth assertion.
+            if consensus_version < ConsensusVersion::V18 || is_root_transition {
```
vicsn: We need to introduce a new `ConsensusVersion::V19`

vicsn (changes requested):

vicsn (approved):

Issue #3230, opened 2026-04-23
[Feature] Enable pruning of historic state


vicsn:
## 🚀 Feature

Add a new public `prune_rocksdb_state<S, T>` method, which takes in a rocksdb store and key type, as well as a range.

When called, it will queue a new type of SequentialOperation, which prunes the respective data.

The main benefits of the smaller state, is that initial participation and recovery will be much faster, making the network more robust.

vicsn:
@ljedrz what do you think of the proposed architecture? The only issue I see is that from a performance/blocking perspective we'd have to be mindful that we don't prune too much at once.

ljedrz:
Ok, so this would primarily apply to the `AuthorityMap` under `BlockStorage`, right? Too bad it's keyed by the block hash, making it unordered, but if this is used consistently (i.e. only the last N entries are ever stored), it would still be quite fast. That being said, if it's used consistently, then only `Map::remove` would be needed to achieve this, perhaps with the addition of a single removal pass of past entries in case some old entry is found, for which we could just introduce a single atomic block that removes past ranges in a loop (which could still just use `Map::remove` underneath), until only the desired number is left.

vicsn:
One more idea: we can mark any relevant getters to said data behind a compile-time `history` compilation flag, and make sure we only allow deleting certain data types when `not(history)`.

ljedrz:
Also, I just realized that I don't see why this would have to be a sequential operation; if we're only removing old data (that's presumably unused), I don't think it could negatively impact anything else.

eranrund:
I've spent some time trying to wrap my head around this, and I'm wondering what your thoughts are regarding how `get_block` is currently implemented and used.

Right now, [`get_block`](https://github.com/ProvableHQ/snarkVM/blob/a56e12e0ee73df7e6405ea8f5fcf1ffdb4cbe13d/ledger/store/src/block/mod.rs#L990-L1043) reconstructs a complete `Block`, including its [authority data](https://github.com/ProvableHQ/snarkVM/blob/a56e12e0ee73df7e6405ea8f5fcf1ffdb4cbe13d/ledger/store/src/block/mod.rs#L1008-L1010). I don't think we would want it to construct a partial block.


Consequently, if we prune old entries from `authority_map`, `get_block` will fail for blocks outside the retention window. This appears to have several implications:

[...]

PR #3323, opened 2026-07-07
Speed up history reads

vicsn:
## Motivation

Closes https://github.com/ProvableHQ/snarkVM/issues/3315

- The type change (u32 → [u8; 4]) is serialization-neutral in bincode.
- Reads using the old binary-search path look up with .to_le_bytes() and find the correct data.

## Backwards compatibility

Wrote a script which builds snarkOS without and with this fix, calls `http://localhost:3033/testnet/program/credits.aleo/mapping/bonded/aleo12ux3gdauck0v60westgcpqj7v8rrcr3v346e4jtq04q7kkt22czsh808v2/history/1` on both versions on the same node to see if the approach is backwards compatible.


ljedrz (commented):
The overall plan is sound, but shouldn't we prefer a simple migration instead of handling the legacy format? For anyone running `--features=history` and having any values in the legacy map, it would be a simple one-time pass after startup, and we'd be able to avoid the extra legacy-handling logic. Such a mini-migration plumbing could also prove to be useful in the future.

In addition, I'd recommend introducing at least a type alias for the BE height, just so its purpose (or at least content) is immediately clear.

vicsn:
> The overall plan is sound, but shouldn't we prefer a simple migration instead of handling the legacy format? For anyone running --features=history and having any values in the legacy map, it would be a simple one-time pass after startup, and we'd be able to avoid the extra legacy-handling logic. Such a mini-migration plumbing could also prove to be useful in the future.

For now I'll defer this because I don't have time. https://github.com/provableHQ/snarkVM/issues/3230 may also touch this conceptually, so we can build on top of that in the future.

> In addition, I'd recommend introducing at least a type alias for the BE height, just so its purpose (or at least content) is immediately clear.

063966b9b

ljedrz (approved):

PR #3325, opened 2026-07-08
Be more lenient about ConsensusVerison usage

vicsn:
## Motivation

`test_quorum_block_spend_limit_aborts_excess_transactions` failed because it broke the assumption that test consensus versions are enabled one after another. Since this test in question already passed once, and in general these `test_v*.rs` tests can be pruned one day, I think it's fine to make this one more lenient.


ljedrz (approved):
LGTM, let's see the CI

PR #3334, opened 2026-07-23
Load the credits record translation verifying key

howardwu:
Dynamic calls can translate the native `credits.aleo/credits` record, but fresh verifiers previously lacked its translation key. This loads the existing network key and enables verification at Consensus V18.

## Changes

- Load the native credits translation key during `Process::load` and credits-key refreshes.
- Return the same key from the deployment-store lookup paths.
- Reject native credits record translation before V18 and allow it from V18 onward.
- Cover fresh-process verification and key load, lookup, and refresh behavior.

The translation circuit, key bytes, and record encoding are unchanged. Ordinary credits calls and other record translations are unaffected.

## Verification

- Focused process, ledger-store, and dynamic credits regressions pass.
- Workspace Clippy and nightly formatting pass.

This is the node-side dependency for ProvableHQ/amm-v3#58.

### DIFF
--- a/synthesizer/process/src/verify_execution/mod.rs
+++ b/synthesizer/process/src/verify_execution/mod.rs
@@ -252,11 +252,21 @@ impl<N: Network> Process<N> {
         // Construct the list of verifier inputs.
         let mut verifier_inputs: Vec<_> = verifier_inputs.values().cloned().collect();
 
+        // Prepare the native credits record identifiers for the V18 activation check.
+        let credits_program_id = ProgramID::from_str("credits.aleo")?;
+        let credits_record_name = Identifier::from_str("credits")?;
+
         // Construct the batch of translation verifier inputs.
         let batch_translation_inputs = Translation::prepare_verifier_inputs(
             execution.transitions(),
             &transition_map,
             &|(program_id, record_name)| {
+                ensure!(
+                    consensus_version >= ConsensusVersion::V18
+                        || program_id != &credits_program_id
+                        || record_name != &credits_record_name,
+                    "The credits record translation is not enabled before ConsensusVersion::V18"
+                );
                 execution_stacks
                     .get(program_id)
                     .ok_or_else(|| anyhow!("Missing stack for program '{program_id}'"))
--- a/synthesizer/src/vm/tests/test_v14/call_dynamic.rs
+++ b/synthesizer/src/vm/tests/test_v14/call_dynamic.rs
@@ -92,8 +92,8 @@ fn test_dynamic_calls_to_credits_aleo() -> Result<()> {
                 ",
     )?;
 
-    // Initialize the VM.
-    let vm = sample_vm_at_height(CurrentNetwork::CONSENSUS_HEIGHT(ConsensusVersion::V14)?, rng);
+    // Initialize the VM at V18, when native credits record translation is enabled.
+    let vm = sample_vm_at_height(CurrentNetwork::CONSENSUS_HEIGHT(ConsensusVersion::V18)?, rng);
 
     // Deploy the program.
     println!("Deploying program: {}", program.id());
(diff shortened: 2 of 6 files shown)


howardwu:
Commit `9be737398` makes fresh nodes able to verify dynamic calls that translate the native `credits.aleo/credits` record. It loads the network-provided translation key through both process and deployment-store paths, and the regression proves the execution in one process before verifying it in another. The full all-feature pre-commit hook passed.

howardwu:
The current cargo-workflow failure is unrelated to this change. Its audit job rejects the base branch’s existing crossbeam-epoch 0.9.18 lockfile under RUSTSEC-2026-0204; the PR and base have the same Cargo.lock blob, and the base failed the same audit. Formatting, clippy, all-target builds, and the other component workflows passed. The dependency update belongs in a separate staging PR, after which this branch can be updated.

raychu86:
This will require ConsensusVersion::V18 gating to prevent validator forking.

raychu86:
A random thought: We could in theory use an amendment transaction to force the stack to hold the translation keys for `credits.aleo/credits`. This wouldn't be a very clean solution, as `credits.aleo` has been historically deployment-less

synthesizer/src/vm/tests/test_v14/call_dynamic.rs
```diff
@@ -90,12 +92,13 @@ fn test_dynamic_calls_to_credits_aleo() -> Result<()> {
                 ",
     )?;
 
-    // Initialize the VM.
-    let vm = sample_vm_at_height(CurrentNetwork::CONSENSUS_HEIGHT(ConsensusVersion::V14)?, rng);
+    // Initialize the VM at V18, when native credits record translation is enabled.
```
Antonio95, line 95: Maybe this should be split into a test which stays at V14 here, and we modify the expectation to catch the error (it's possible we have to increase the block spacing between the V14 and V18 *test* consensus heights); and a new separate test in test_v18 which operates as the updated version here. I can add an issue, this does not seem urgent at all.

synthesizer/process/src/verify_execution/mod.rs
```diff
@@ -252,11 +252,21 @@ impl<N: Network> Process<N> {
...
+        // Prepare the native credits record identifiers for the V18 activation check.
+        let credits_program_id = ProgramID::from_str("credits.aleo")?;
+        let credits_record_name = Identifier::from_str("credits")?;
+
         // Construct the batch of translation verifier inputs.
         let batch_translation_inputs = Translation::prepare_verifier_inputs(
             execution.transitions(),
             &transition_map,
             &|(program_id, record_name)| {
+                ensure!(
```
Antonio95, line 264: I see this is the only consensus guard. 
- I think (at least) this guard is correct if we assume there are no paths which could cause a validator to synthesise the credits translation key before this is called (such as in many of the tests we had, where the same vm was used to produce proofs and to check them - which is why the vm had the verifying keys at transaction-checking time). Is this a safe assumption?
- On the flip side, is this the only place that needs guarding? More concretely, are there other places where a pre-fix verifier could attempt to fetch this key and fail? (which could result in a fork).

I'm going to temptatively say yes to both. Specifically about the second one, I also asked an agent which, after a few mins thinking, came saying "The credits translation verifying key is only ever used for verification through the closure at lines 263‑274 (...)" and "The other get_verifying_key-by-name sites don't touch credits.aleo/credits". Let's keep an eye in the coming days for possible errors coming from this consensus guard, though.

Antonio95 (approved):
Left two minor comments, no changes needed for now. Green light from me.

raychu86 (approved):
Logic changes and consensus gating LGTM, can merge once there is confirmation on final test checks

niklaslong (approved):
LGTM, one nit though: the verifying key is loaded into a fresh VM but the proving key isn't, this means a prover would re-synthesise both as the guard in [synthesize.rs](https://github.com/ProvableHQ/snarkVM/blob/ee6c11509fc51097f0da30b6eb281ea543fb8d41/synthesizer/process/src/stack/helpers/synthesize.rs#L125) gates on both?

PR #3335, opened 2026-07-23
Enforce canonical subDAG certificate ordering

howardwu:
`to_subdag_root` hashes certificate IDs in each round in `IndexSet` insertion order.

However, `IndexSet` equality ignores order, so the existing DFS sanity check could accept two structurally equal subDAGs that produce different roots.

## What changed

- Reconstruct the canonical left-to-right DFS order using the same reverse-push/LIFO traversal as snarkOS `order_dag_with_dfs`.
- Compare certificate iterators explicitly instead of relying on order-insensitive `IndexSet` equality.
- Enforce the check during checked block construction and block verification starting in consensus V18.
- Preserve pre-V18 acceptance behavior and leave the Merkle-root algorithm unchanged.

## Expected result

At V18 and later, every accepted quorum block must use the single certificate order derived from the leader certificate and its ordered predecessor IDs. Validators therefore hash the same ordered leaves for the same certificate graph.

## Limits

This does not make `IndexSet` sort certificates, nor does it sort certificate IDs. The fix validates the producer's canonical DFS order before accepting a block.

V18 is currently unactivated in the public network height tables, so activation still depends on scheduling that consensus version.

### DIFF
--- a/ledger/narwhal/subdag/src/lib.rs
+++ b/ledger/narwhal/subdag/src/lib.rs
@@ -325,6 +325,13 @@ pub mod test_helpers {
 
     type CurrentNetwork = MainnetV0;
 
+    /// Orders the given subDAG using the same left-to-right DFS traversal as the Narwhal block producer.
+    pub fn order_subdag_with_dfs<N: Network>(
+        subdag: &BTreeMap<u64, IndexSet<BatchCertificate<N>>>,
+    ) -> BTreeMap<u64, IndexSet<BatchCertificate<N>>> {
+        super::order_subdag_with_dfs(subdag)
+    }
+
     /// Returns a sample subdag, sampled at random.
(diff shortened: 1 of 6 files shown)


ljedrz (approved):

ledger/narwhal/subdag/src/lib.rs
```diff
@@ -452,6 +489,50 @@ mod tests {
         }
     }
 
+    #[test]
+    fn test_certificate_order_consensus_gate() {
```
Antonio95, line 500: This could optionally moved to `test_v18`

Antonio95 (approved):

Antonio95:
Some CI failures seem like they could be related to these changes, e.g. `tests::test_subdag_with_long_branch`: `Error: Subdag certificates are not canonically ordered in block 26`. It would be good to double check.

raychu86 (approved):

PR #3336, opened 2026-07-24
Fix stack handling in casts involving external structs

### LINKED ISSUE #3330: FinalizeTypes::matches_struct rejects valid cast to external struct with member-access operands
## Summary

`FinalizeTypes::matches_struct` validates a `cast` to an **external struct** using a single `Stack`, and reuses that stack both to look up the struct definition **and** to resolve the operand register types. When it is called with the *external* program's stack (which is correct for the struct definition), it then resolves the cast's operand member accesses against that same external stack — so a member access on an external-struct register calls `get_external_stack(<external program>)` while already on that program's stack, and fails:

```
Attempted to get the main program '<external program>' as an external program.
```

The analogous `RegisterTypes::matches_struct` (functions/transitions) takes **two** stacks and does not have this problem — which is why the bug is finalize-only.

## Version

`v4.8.1` (commit `b7f0859`)

## Reproducer (pure Aleo)

Add both programs to a `Process` (dependency order: `lib.aleo` first).

`lib.aleo`
```aleo
program lib.aleo;

struct S:
    hi as u128;
    lo as u128;

function noop:

constructor:
    assert.eq edition 0u16;
```

`consumer.aleo`
```aleo
import lib.aleo;
program consumer.aleo;

function run:
    input r0 as lib.aleo/S.private;
    async run r0 into r1;
    output r1 as consumer.aleo/run.future;

finalize run:
    input r0 as lib.aleo/S.public;
    cast r0.hi r0.lo into r1 as lib.aleo/S;   // <-- member-access operands into an external-struct cast
    assert.eq r1.lo r0.lo;

constructor:
    assert.eq edition 0u16;
```

`process.add_program(&consumer)` fails during finalize type-checking with the error above. The same construction in a **transition** body (`function`, not `finalize`) validates fine, e.g. `cast r0.hi r1.lo into r2 as lib.aleo/S`.

## Root cause

The two struct-cast type-checkers diverge (paths in `synthesizer/process/src/stack`):

- **Transition** — `register_types/matches.rs::matches_struct` takes two stacks:
  ```rust
  pub fn matches_struct(&self, operands_stack: &Stack<N>, stack: &Stack<N>, operands, struct_) -> Result<()>
  ```
  Call site `register_types/initialize.rs:550`:
  ```rust
  self.matches_struct(stack, &*external_stack, instruction.operands(), struct_)?;
  ```
  Operand types are resolved against `operands_stack` (the caller), the struct definition against `stack` (external). Correct.

- **Finalize** — `finalize_types/matches.rs::matches_struct` takes only one stack:
  ```rust
  pub fn matches_struct(&self, stack: &Stack<N>, operands, struct_) -> Result<()>
  ```
  Call site `finalize_types/initialize.rs:999`:
  ```rust
  self.matches_struct(&*external_stack, instruction.operands(), struct_)?;
  ```
  Inside, operand register types are resolved with `self.get_type(stack, register)` where `stack` is the **external** stack. For a member-access operand, `get_type` reaches the `ExternalStruct(..)` member branch (`finalize_types/mod.rs:186`):
  ```rust
  let external_stack = stack.get_external_stack(locator.program_id())?;
  ```
  Here `stack` already *is* that external program's stack, so `program_id == self.program.id()` and `get_external_stack` bails (`helpers/stack_trait.rs:265`).

So this is an inconsistency between the two parallel type-checkers: `RegisterTypes::matches_struct` was given a separate stack for operand-type resolution vs. struct-definition lookup, but `FinalizeTypes::matches_struct` was not.

## Suggested fix

Give `FinalizeTypes::matches_struct` the same two-stack signature as `RegisterTypes::matches_struct` (an `operands_stack` for `get_type` on the operands, and the external `stack` for the struct definition), and update the `finalize_types/initialize.rs:999` call site to pass both. Cast semantics for external structs in finalize should match transitions.

## Context

Found via a Leo miscompile ([ProvableHQ/leo#29586](https://github.com/ProvableHQ/leo/issues/29586)). Leo currently works around it in code generation by spilling member-access operands into plain registers before an external-struct cast in a finalize, but the underlying defect is here in snarkVM.

Antonio95:
Closes #3330, see details therein.

This PR is in draft mode until a decision is made on consensus gating.

The issue boiled down to the fact that `FinalizeTypes::matches_struct` (here: https://github.com/ProvableHQ/snarkVM/blob/staging/synthesizer/process/src/stack/finalize_types/matches.rs#L20) only received one stack when actually two separate ones could be involved (one for the target struct and one for the operand(s') struct(s)). The `Stack(id)` type needs information about the program it refers to in order to be uniquely determined.

This problem had already been detected and fixed for the function (i.e. private) scope, corresponding to `RegisterTypes::matches_struct` (here: https://github.com/ProvableHQ/snarkVM/blob/staging/synthesizer/process/src/stack/register_types/matches.rs#L20).

This PR fixes the issue by porting those simple changes to `FinalizeTypes::matches_struct`. It also adds many intentionally designed test cases involving relatively complex casts (in `synthesizer/src/vm/tests/test_v18/external_struct_stacks.rs`). Aside from the pure struct case, that file includes test functions focused on arrays and records containing external structs. No issues have been found than in `FinalizeTypes::matches_struct`, but this part of the language is difficult to test exhaustively so we should keep an eye out in the future.

Aside from the fix, the PR includes several non-functional changes to a couple of functions involved in the check to improve code clarity (e.g. renaming the stacks received by `matches_struct`).


synthesizer/src/vm/tests/test_v19/external_struct_stacks.rs
```diff
@@ -0,0 +1,840 @@
...
+            a as program_a.aleo/struct_a;
+
+        function noop:
+
+        constructor:
+            assert.eq edition 0u16;
+        ",
+    )?;
+
+    // This is a distinct program from program_b.aleo, but both declare a struct with the same name
```
mohammadfawaz, line 61: Maybe add a rejection case where same-named structs have different member layouts; the current definitions are equivalent, so a stack mix-up could still pass. A constructor or view case would also cover the other scopes sharing `FinalizeTypes`.

synthesizer/src/vm/tests/test_v18/external_struct_stacks.rs
```diff
@@ -0,0 +1,840 @@
...
+            cast r2.b.a into r3 as program_b2.aleo/struct_b;
+
+        constructor:
+            assert.eq edition 0u16;
+        ",
+    )?;
+
+    let vm = sample_vm_at_height(CurrentNetwork::CONSENSUS_HEIGHT(ConsensusVersion::V18)?, rng);
+
+    let transaction = vm.deploy(&deployer_private_key, &program_a, None, 0, None, rng)?;
```
mohammadfawaz: Could most cases use `Process::add_program` instead of full deployment/proof/execution flows? The bug occurs during stack type checking; keeping one end-to-end case and making the rest focused tests would reduce CI time and duplication.

synthesizer/src/vm/tests/test_v19/external_struct_stacks.rs
```diff
@@ -0,0 +1,840 @@
...
+            input r0 as program_a.aleo/struct_a.public;
+            async run_11 r0 into r1;
+            output r1 as program_c.aleo/run_11.future;
+
+        finalize run_11:
+            input r0 as program_a.aleo/struct_a.public;
+
+            cast r0 into r1 as program_b.aleo/struct_b;
+            cast r0 r1 into r2 as struct_c;
+
```
mohammadfawaz: Trailing spaces on this blank line can be removed.

Antonio95:
After the internal discussions on consensus gating, I think I have a solution which is strictly consensus-preserving (i.e. the new code at < V19 runs *exactly* like the old code) backwards compatibility while not threading the consensus version to any of the delicate internal places we wanted to avoid.

The idea is:

1. The internal call to `matches_struct` for `FinaliseTypes::check_instruction_opcode` is left in its new, corrected form (as corrected in the first version of the PR: it receives two stacks and passes the correct values; mirroring its `RegisterTypes` counterpart).
2. Inside `VM::check_transaction`, we add one more case to the legacy checks for `consensus_version < ConsensusVersion::V19`: https://github.com/ProvableHQ/snarkVM/blob/fix/external_struct_stacks/synthesizer/src/vm/verify.rs#L368.
    1. Note this check cannot happen inside `synthesizer/program/src/lib.rs`, since achieving *strict* consensus preservation means one truly needs the `stack` type (which is not available to `synthesizer_program`: that would create a circular dependency).
    2. For efficiency, this < V19 legacy check is guarded by an instruction-level check on whether any finalize-like scope (`finalize`, `view`, `constructor`) contains a cast-to-external-struct instruction: https://github.com/ProvableHQ/snarkVM/blob/fix/external_struct_stacks/synthesizer/program/src/lib.rs#L1066.
    3. The legacy check performs the minimal amount of work needed to mimic the pre-V19 behaviour: instead of creating a full stack for the program being checked, it creates a lightweight stack and only populates it with the relevant data for the legacy check (mostly, types of output registers in finalise-like scopes). It then goes through every cast-to-external-struct instruction and makes the call `finalize_types.matches_struct(&*external_stack,&*external_stack,cast.operands(),struct_,)?;`, which exactly mimics the old code (which made a call to the original `matches_struct`, which received only one `Stack`: `external_stack`). Note that other stacks relevant to this check (the external stacks which may have to be explored during type resolution) must exist beforehand and therefore aren’t affected by the fact that, for the *current* program, we are only building a lightweight stack.
    
    I’ve implemented this check in as much of a self contained way as possible within `VM::check_transaction`. This has only required `pub`-exposing a couple of new things (e.g. methods of `FinaliseTypes`) in exchange for not having to add the legacy-logic-specific into e.g. `FinaliseTypes` methods.
    

**Small caveat:** The new legacy check has only been added to the verifier side, to `VM::check_transaction`. Nothing analogous has been added to `VM::deploy`, which in particular uses the new, correct `matches_struct` when judging the validity of the deployment it is building (even at `ConsensusVersion` < V19). This means that a program with a problematic cast-to-external instruction will fail to build the deployment with the old code, but succeed to build it (even at `ConsensusVersion` < V19) with the new code. From what I have seen, this should not be an issue as we have many other legacy (and the opposite kind) checks we perform in `VM::check_transaction` and `VM::deploy`. What matters is that `VM::check_transaction` has a consistent behaviour.

As part of these latest changes, I have done the following:

1. Checked that all `test_v*` pass with the new version of the code but always being executed **at consensus version < V19** (I simply temporarily added enough space between the test consensus heights for V18 and V19 to ensure the latter was never reached). This is a sanity check that the changes don’t break anything.
2. Expanded `test_v19::external_struct_stacks.rs` with test cases for the view and constructor scopes.
3. Crucially: checked that, for all test cases in `test_v19::external_struct_stacks.rs` (new and old), the finalize/view/constructor scopes that are rejected with the new code at consensus version < V19 are exactly the same as those that failed in the old code. This was indeed the case: exactly four out of 33 cases were rejected in the new and old code (`views/view_2`, `constructor/constructor_case_2`, `only/run_10`, `only/run_16`).
    
    NB: This can’t be tested with the clean version of the code (since one needs to develop each test finalize/view/constructor scopes separately to see which ones fail), but I left in the commit which enables this: 141bbd4d6 “temporary changes to print exact rejections”.
    
    Also, because of the “Small caveat” mentioned earlier, the new-code-vs.-old-code check could not be completely symmetric: in the new code, acceptance rejection was judged based on `vm.check_transaction` (`vm.deploy` always succeeds); in the old code, it was based on `vm.deploy`.
    
4. Ran timing tests to ensure that pre-V19, the new legacy check doesn’t significantly impact verifying times. In other words, I’ve run `check_transaction` on the old and new code at pre-V19. I’ve only done so for the `external_struct_stacks.rs`, (note that, for programs whose finalize scopes don’t cast an external structs, the only new pre-V19 computation is precisely the check of whether they contain any such instructions, which guards the full legacy check). The results are in the attached file (the serial numbers are much more meaningful than the parallel ones, since tests are themselves run in parallel and therefore affect each other’s measurements). The change in verification time (for the accepting cases) between the old and the new code ranges between -8.6% (yes, decrease) and +12.6 (the average is ~0.8%), which means that, random noise aside, the effect of the new check should be negligible (as expected by pure inspection of the code). Let me know if you want more reliable numbers (e.g. benching each case several times to get a less noisy average).

I have not addressed the original review comments yet, will do that shortly.

[times.txt](https://github.com/user-attachments/files/30538079/times.txt)

ljedrz:
@Antonio95 [related](https://github.com/ProvableHQ/snarkVM/pull/3355), especially the Veria AI review comment.

synthesizer/src/vm/tests/test_v19/external_struct_stacks.rs
```diff
@@ -0,0 +1,840 @@
...
+        function run_1:
+            input r0 as u8.private;
+
+            cast r0 into r1 as program_a.aleo/struct_a;
+            cast r1 into r2 as program_b.aleo/struct_b;
+            cast r1 r2 into r3 as struct_c;
+
+            output r3 as struct_c.private;
+
+        // Case 2 / function: Similar to case 1 but with a cast involving a external-struct member read (r1.val) into an external-struct.
```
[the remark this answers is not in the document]
Antonio95: Done in cdc93576e8c8f91e54f3aa2f515d260dfe9a01a3

synthesizer/src/vm/tests/test_v19/external_struct_stacks.rs
```diff
@@ -0,0 +1,840 @@
...
+        // Case 7 / function: Tests a three-level struct access involving external structs.
+        function run_7:
+            cast 24u8 into r0 as program_a.aleo/struct_a;
+            cast r0 into r1 as program_b.aleo/struct_b;
+            cast r0 r1 into r2 as struct_c;
+            cast r2.b.a.val into r3 as u8;
+
+            output r3 as u8.private;
+
+        // Case 8 / function: Tests a casts to external struct one of whose operands involves a two-level struct access.
```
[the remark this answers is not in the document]
Antonio95: Done in cdc93576e8c8f91e54f3aa2f515d260dfe9a01a3

synthesizer/src/vm/tests/test_v19/external_struct_stacks.rs
Antonio95: Done 👌

synthesizer/src/vm/tests/test_v18/external_struct_stacks.rs
Antonio95: Good idea 👌  Done, this is the resulting state of the file: https://github.com/ProvableHQ/snarkVM/blob/a66fb3bd97bf249c0470829ddcdd33bf920326eb/synthesizer/src/vm/tests/test_v19/external_struct_stacks.rs

synthesizer/src/vm/tests/test_v19/external_struct_stacks.rs
Antonio95: I've written a test with one possible instance of that situation, cf. https://github.com/ProvableHQ/snarkVM/blob/a66fb3bd97bf249c0470829ddcdd33bf920326eb/synthesizer/src/vm/tests/test_v19/external_struct_stacks.rs#L702-L782 (and in particular the instruction on line 776 therein). Let me know if that's what you had in mind.

synthesizer/src/vm/tests/test_v19/external_struct_stacks.rs
```diff
@@ -0,0 +1,840 @@
...
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+use super::*;
+
+// Checks that various instances of casts to external structs or from using external-structs members
+// function as expected. This is tested in both the function (i.e. private, RegisterType) setting as
+// well as the public (i.e. finalise, FinaliseType) one.
```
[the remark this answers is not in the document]
Antonio95: Done

synthesizer/src/vm/tests/test_v19/external_struct_stacks.rs
```diff
@@ -0,0 +1,1033 @@
...
+    let err_new = {
+        let vm_new = sample_vm_at_height(CurrentNetwork::CONSENSUS_HEIGHT(ConsensusVersion::V19).unwrap(), rng);
+
+        vm_new.process().lock().add_program(&program_a).unwrap();
+        vm_new.process().lock().add_program(&program_b).unwrap();
+        vm_new.process().lock().add_program(&program_b2).unwrap();
+        vm_new.process().lock().add_program(&program_c).unwrap_err().to_string()
+    };
+
+    assert_eq!(err_old, err_new);
```
mohammadfawaz, line 828: These branches do not exercise different consensus versions. Can we test the pre-V19 path through `VM::check_transaction` with a deployment produced pre-V19, then assert the intended V17/V19 outcomes.

Antonio95: Done here, is this what you had in mind? 9801bd47a2ab9799ad7880b7a7fdd1dacce44f96

PR #3342, opened 2026-07-29
fix(tests): omit concrete outputs from rewritten expectations

### LINKED ISSUE #3341: Rewrite-expectations output handling in updated test_vm_execute_and_finalize
We recently modified those tests so they don't check for specific output values (which were very fragile due to randomness affecting e.g. record nonces). They still check whether the transaction succeeded or failed as expected.

We should update the operation of the `REWRITE_EXPECTATIONS` flag accordingly (e.g. not to write concrete output values).

sethryanrollins:
This updates `REWRITE_EXPECTATIONS` to leave out the concrete output values that the tests no longer compare. Those values can change because of randomness, causing noisy expectation diffs. The existing expectation files were regenerated without the unused additional output data, and tests were added to cover the rewrite behavior. Closes #3341.


vicsn (commented):
Mostly LGTM! Thank you! Can you resolve the merge conflict with staging?

sethryanrollins:
@vicsn can do! Should be able to when I get into work in a few

vicsn (approved):

PR #3349, opened 2026-07-31
[Perf] Faster block tree (de)serialization

ljedrz:
Found while working on https://github.com/ProvableHQ/snarkOS/pull/4361: restarting the nodes took way longer than I anticipated, and it boils down to a flat ~48s (on my machine, ~125s in the CI) cost of loading the block tree cache, which - while still a big improvement for very large ledgers - is a slowdown for small ones.

This PR removes this flat (de)serialization "tax", by introducing a `MerkleTreeState` object, which is a subset of the `MerkleTree`, sans its hashers, which can be recreated without having to deserialize them from the cache file.

Below is a comparison of the block tree load times between the current state and this PR:

| leaves | from scratch | legacy file | legacy load | new file | new load |
|-|-|-|-|-|-|
| 1 | 0.35 ms | 21,712,808 B | 48.07 s | 112 B | 0.34 ms |
| 100 | 2.85 ms | 21,720,040 B | ~48 s | 7,344 B | 0.28 ms |
| 10,000 | 19.4 ms | 22,557,032 B | ~48 s | 844,336 B | 1.34 ms |
| 1,000,000 | 1.295 s | 87,267,176 B | ~48 s | 65,554,480 B | 97.7 ms |

Why haven't we noticed this sooner? Variable window size was added to BHP on 30.04.2026 (dd09be7e3, efa368be4), along with setup and size metrics, which increased the size of the serialized block tree from ~210KiB to 20.7MiB, so what originally cost ~0.5s to deserialize now costs ~48s. Note: those changes were most likely expected, and have their own benefits, with this fallout just being an expected side effect.

### DIFF
--- a/console/collections/src/merkle_tree/mod.rs
+++ b/console/collections/src/merkle_tree/mod.rs
@@ -31,7 +31,7 @@ use locktick::parking_lot::Mutex;
 #[cfg(not(feature = "locktick"))]
 use parking_lot::Mutex;
 use serde::{Deserialize, Serialize};
-use std::{collections::BTreeMap, mem};
+use std::{borrow::Cow, collections::BTreeMap, mem};
 
 #[cfg(not(feature = "serial"))]
 use rayon::prelude::*;
@@ -50,8 +50,6 @@ use rayon::prelude::*;
 /// Padding levels are then added as needed to reach the full `DEPTH`, each of
 /// which is constructed by hashing the root of the previous level together with
 /// `e`.
-#[derive(Deserialize, Serialize)]
-#[serde(bound = "E: Serialize + DeserializeOwned, LH: Serialize + DeserializeOwned, PH: Serialize + DeserializeOwned")]
 pub struct MerkleTree<E: Environment, LH: LeafHash<Hash = PH::Hash>, PH: PathHash<Hash = Field<E>>, const DEPTH: u8> {
     /// The leaf hasher for the Merkle tree.
     leaf_hasher: LH,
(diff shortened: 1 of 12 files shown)


vicsn (approved):
Excellent!

console/collections/src/merkle_tree/mod.rs
```diff
@@ -178,6 +196,80 @@ impl<E: Environment, LH: LeafHash<Hash = PH::Hash>, PH: PathHash<Hash = Field<E>
...
+
+        // Recompute the root hash, by iterating from the root level up to `DEPTH`.
+        let mut root_hash = tree[0];
+        for _ in 0..padding_depth {
+            // Update the root hash, by hashing the current root hash with the empty hash.
+            root_hash = path_hasher.hash_children(&root_hash, &empty_hash)?;
+        }
+        ensure!(root_hash == root, "The Merkle tree state has an invalid root");
+
+        Ok(Self {
```
Antonio95, line 262: From what I gather, this initialisation function checks consistency between the "internal root" `tree[0]` (the root of the largest subtree including all non-empty leaves) and the full tree's root (obtained by hashing the previous value with the empty hash a number of times). But it does not check consistency  between `tree` (the digests of the implicit leaves) and the internal root.

Is my understanding correct? And if so, is there any particular reason for that half-check design? It seems it would make more sense to either check consistency in both cases, or in neither - but I might be missing something.

Antonio95 (commented):
Great stuff 👌 I think the reason the issue hadn't been caught earlier is that the changes which caused it were merged into `staging` only recently (https://github.com/ProvableHQ/snarkVM/pull/3249).

I have left only minor comment regarding the consistency check in `from_state`. Happy to approve once we discuss that one.

console/collections/src/merkle_tree/mod.rs
ljedrz: this is the same logic as in the existing `prepare_append` (L#343), just double-checked against the state's root for internal consistency, since we only ever use it against our own cached tree (i.e. we trust it); strictly speaking, we could skip it, but it's relatively cheap

Antonio95: I see, so it's a cheap sanity check 👌 And doing the whole digests-to-tree[0] check is not justified in terms of time vs. necessity. Approving now.

PR #3355, opened 2026-08-05
[Feat] Align declared plaintext type sizes during deployment verification

ljedrz:
Deployment verification did not bound the size of the `PlaintextType`s a program declares, while `Plaintext::to_fields` caps encoded data at `MAX_DATA_SIZE_IN_FIELDS` fields at runtime. This PR aligns them.

Gated at `V19`.

### DIFF
--- a/console/network/src/consensus_heights.rs
+++ b/console/network/src/consensus_heights.rs
@@ -69,7 +69,8 @@ pub enum ConsensusVersion {
     /// V18: Enables native credits record translation, introduces block-wide deployment limits,
     ///      and enforces canonical subDAG certificate ordering.
     V18 = 18,
-    /// V19: Adds more accurate type checking for the root call.
+    /// V19: Adds more accurate type checking for the root call, and bounds the size of every
+    /// `PlaintextType` declared in a deployed program.
     V19 = 19,
     /// V20: TBD
     V20 = 20,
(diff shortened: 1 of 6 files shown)


vicsn (approved):
LGTM, the Veria suggestion can be tackled in a subsequent PR

vicsn:
@cbeck88 can you review, and potentially rebase https://github.com/ProvableHQ/snarkVM/pull/3357 . Do you think both PRs are useful or should we pick one?

cbeck88:
This looks good to me, and I agree with you @vicsn that the Veria suggestion can be a follow on.

I think 3355 and 3357 are independent though. This involves imposing limits on plaintext type size where none existed before, so its a behavior change, and is consensus gated.

3357 is fixing an efficiency problem in type checking, where a small program can take exponential time to type check because the routine is naive and doesn't remember work that it already did before.

There are several follow ups for 3357 -- size_in_bits_raw and types_equivalent have the same recursive structure without memoization, so the same pathological types can cause those functions to take a huge amount of time.

I think we should merge this, and merge 3357, and I don't think they will conflict. And afterwards go on to try to fix size_in_bits and types_equivalent, each in their own PR.

cbeck88 (approved):

PR #3356, opened 2026-08-05
`rand.chacha` cost

Antonio95:
This PR implements a more meaningful cost model for the finalize-only command `rand.chacha`.

The operation of `rand.chacha` comprises the following steps:

1. Compute a “pre-seed” including a few always-present values (but execution-dependent, e.g. block height, …) and up to two additional operands passed to the command.
2. Call `hash_bhp1024` on the pre-seed and output 32 bytes, which constitutes the seed.
3. Instantiate a `ChaCha20` RNG with the seed
4. Derive from the RNG as many bytes as necessary to compute a value of the command’s destination type.
5. Actually compute that value.

Before this PR, the cost of this command was a fixed 25_000 microcredits.

### Changes

- Cost model: dirty test code was added to evaluate what the correct pricing should be, cf. 98032ef43 and e61b36494 (the changes have since been cleaned up). As it turns out,
    - The always-present pre-seed is about 750 bits, which is roughly equivalent to 3 field elements.
    - Steps 3-5 above are a) essentially free for the destination types which do not boil down to a group element (negligible cost wrt. 1-2) and b) not cheap in the opposite case (destination in {Group, Address}). In the latter case, steps 3-5 take about twice as long as 1-2 (when there are no additional operand seeds).
    
    In light of that, the cost function computes:
    
    - A “seed component” reflecting the cost of the `hash_bhp1024` call. This is computed the same way the cost for `hash.bhp1024` is, with the difference that we append three “field” operands to capture the ever-present part of the input (the aforementioned ~750 bits).
    - An “output component” capturing steps 3-5 above, which is 0 for non-Group/Adress destinations and 115_000 for those two types. This number comes from the aforementioned tests and turns out to yield quite accurate values (judging based on execution time).  Cf. 98032ef43 and e61b36494, and in particular the table https://github.com/ProvableHQ/snarkVM/blob/e61b36494d4805280e61052fa32a24093d9a8889/synthesizer/program/src/logic/command/rand_chacha.rs#L429-L432. I’m happy to elaborate on the logic.
- Consensus gating:
    
    This change needs to be consensus-gated as the change in cost may result in different validator behaviour in the presence of the same transaction. `cost_per_command` now also receives the consensus version as an argument (https://github.com/ProvableHQ/snarkVM/blob/fix/rand_chacha_cost/synthesizer/process/src/cost.rs#L721). The way I have coded it, each `match` arm is free to use it as it sees fit, and in particular right now only one does (`Command::RandChaCha`). *If you think we could do something more robust, e.g. have each arm necessarily encode a vector of consensus versions (including only the `ConsensusVersion`s where the cost function jumps), let me know.*
    
    Unfortunately, because `cost_per_command` is reached through many intermediate cost functions, basically now all functions in that file need to receive it as well. Note the file’s publicly exposed functions already received it anyway, so this is only a change in internal wiring. Tests which called the internal cost functions directly have been updated to also pass a consensus version (the choice of which version to pass has been relatively meaningful, e.g. the earliest version for which that intermediate cost function is selected by a higher-level one).
    
- The PR includes no dedicated tests, since the actual change is quite minimal (as mentioned above, testing was done to determine the cost model). *Let me know if you want me to add any concrete tests or bring back any of those cost-determination tests.*


synthesizer/process/src/cost.rs
```diff
@@ -45,13 +45,13 @@ pub fn deployment_cost<N: Network>(
...
+        deployment_cost_v4(process, deployment, consensus_version)
     } else if consensus_version >= ConsensusVersion::V16 {
-        deployment_cost_v3(process, deployment)
+        deployment_cost_v3(process, deployment, consensus_version)
     } else if consensus_version >= ConsensusVersion::V10 {
-        deployment_cost_v2(process, deployment)
+        deployment_cost_v2(process, deployment, consensus_version)
     } else {
-        deployment_cost_v1(process, deployment)
+        deployment_cost_v1(process, deployment, consensus_version)
```
vicsn, line 54: Suggested refactor: since we only need the new changes from the new consensus_version, maybe it'll be simpler if we introduce a new `deployment_cost_v5` and `execution_cost_v4`, and from within those functions we route to new appropriate finalize_cost wrappers.

synthesizer/process/src/cost.rs
```diff
@@ -1122,7 +1137,31 @@ pub fn cost_per_command<N: Network>(
...
             cost_in_size(stack, finalize_types, [command.key_operand()], MAPPING_PER_BYTE_COST, mapping_base_cost)
         }
-        Command::RandChaCha(_) => Ok(25_000),
+        Command::RandChaCha(command) => {
+            if consensus_version >= ConsensusVersion::V19 {
+                let seed_component = {
+                    let mut bhp_operands = command.operands().to_vec();
+                    for _ in 0..3 {
+                        bhp_operands.push(Operand::Literal(Literal::Group(Group::generator())));
+                    }
```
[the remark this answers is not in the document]
Antonio95: Good call, fixed

synthesizer/process/src/cost.rs
```diff
@@ -1122,7 +1137,31 @@ pub fn cost_per_command<N: Network>(
...
+                    cost_in_size(stack, finalize_types, &bhp_operands, HASH_BHP_PER_BYTE_COST, HASH_BHP_BASE_COST)
```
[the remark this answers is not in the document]
Antonio95: Since `rand.chacha` calls `bhp1024`, we are using the cost we'd assign to the latter call, which is computed using `cost_in_size` as above. In any case, I've checked `hash.bhp1028 [u8; 1024]` and `hash.bhp1024 [[[[[u8; 4]; 4]; 4]; 4]; 4]` run in roughly the same time (i.e. the nesting and size/type element prefixation doesn't seem to have that much of an impact), so it is okay they are assigned the same cost.

synthesizer/process/src/cost.rs
```diff
@@ -1122,7 +1133,31 @@ pub fn cost_per_command<N: Network>(
         Command::GetOrUseDynamic(command) => {
             cost_in_size(stack, finalize_types, [command.key_operand()], MAPPING_PER_BYTE_COST, mapping_base_cost)
         }
-        Command::RandChaCha(_) => Ok(25_000),
+        Command::RandChaCha(command) => {
+            if consensus_version.is_some_and(|version| version >= ConsensusVersion::V21) {
+                let seed_component = {
+                    let mut bhp_operands = command.operands().to_vec();
+                    for _ in 0..3 {
```
vicsn, line 1140: Can you add a comment "The always-present pre-seed is about 750 bits, which is roughly equivalent to 3 field elements."

vicsn (approved):
LGTM, one nit

synthesizer/process/src/cost.rs
```diff
@@ -45,7 +45,7 @@ pub fn deployment_cost<N: Network>(
     consensus_version: ConsensusVersion,
 ) -> Result<(MinimumCost, DeployCostDetails)> {
     if consensus_version >= ConsensusVersion::V18 {
-        deployment_cost_v4(process, deployment)
+        deployment_cost_v4(process, deployment, consensus_version)
```
raychu86, line 48: Safety here seems fine, but why not just make this a `deployment_cost_v5` and `execution_cost_v4`?

PR #3357, opened 2026-08-05
fix exponential-time recursive type checking issue

cbeck88:
## Motivation

it was pointed out that, check_plaintext_type can be forced to do an enormous amount of work, if e.g. struct T1 contains 32 fields of type T2, which contains 32 fields of type T3, etc. up to a depth of 12. Instead of checking each type once, the routine rechecks the same types again and again, and checking a small leo program can take hundreds of trillions of operations, leading to a persistent DOS risk.

this is trivially fixed by memoizing when a type has been checked.

## Test Plan

we added a unit test that checks that the bad test case type checks in a small amount of time.

## Backwards compatibility

I think there is no backwards compatibility risk, since there's no behavior change, only efficiency improvement


iamalwaysuncomfortable:
Can I just say I've missed reading these PR descriptions?

vicsn (approved):

PR #3360, opened 2026-08-06
Remove redundant assert

eranrund:
We are trying to reduce panics in favor of standard `Err` results. The function modified here already has an `ensure!` call that checks the exact condition the `assert_eq!` used to check, however the `assert_eq!` happens before `ensure!` gets to run. I don't think we want that.


ljedrz (approved):

vicsn (approved):
I don't think this is worth the writer+reviewer's time. If we'd want to, AI can rampage through the code and remove panic sites, guided by a Rust no_panic attribute. But it's not a concern because of our try_vm_runtime hack

PR #3365, opened 2026-08-10
Harden deployment verification: seed remaining ambient RNG sources (follow-on to #3346)

cbeck88:
## Motivation

Follow-on hardening for #3346:

PR #3346 makes the `CheckDeployment` burner — and thus `self.signer` — deterministic, closing the fork in which a `call.dynamic` target chosen from `self.signer` makes the same deployment verify on some validators and abort on others. Since that change already alters `CheckDeployment` synthesis for every deployment, this PR closes additional ambient-RNG sources reachable during deployment verification, so that verification is a pure function of the deployment. (This fixes all such similar issues that I could find at this time.)

- **`get.record.dynamic` — an independent second instance of the same fork.** `compute_or_sample_path` samples its dummy entry value from `rand::rng()` on the not-present branch, which is reached during `CheckDeployment` synthesis (a sampled dynamic record carries `data = None`). That value is stored to a register a program can feed into a `call.dynamic` target (closure vs. non-closure), so the same deployment can verify on some validators and abort on others. Because the exploiting program never reads `self.signer`, **#3346's burner fix does not close it.** It is now drawn from a `ChaChaRng` seeded by the record root and entry identifier, both deterministic across validators verifying the same deployment.

- **`Request::sign` nonce and record-translation sampling** in `verify_deployment` are now drawn from the deployment-ID-seeded RNG rather than the ambient `rng`. These are defense-in-depth: once the burner key is seeded neither currently changes a verification outcome, but this removes any remaining dependence of verification on the ambient RNG. The now-unused `rng` argument is renamed `_rng` (signature retained for API compatibility, but we could also remove it in this PR if reviewers want).

Deliberately left out of scope: the Synthesize-mode (key-generation) burner in `synthesize.rs` also uses the ambient RNG, but `synthesize_key` runs during `deploy()`/proving — never during validator block verification — so it is not a consensus-fork source.

## Test Plan

Adds `synthesizer/src/vm/tests/test_v19/get_record_dynamic_targets.rs`, mirroring #3346's `closure_dynamic_targets.rs` but driving the `call.dynamic` target from a `get.record.dynamic` read instead of from `self.signer`. It deploys such programs and asserts repeated runs of `verify_deployment` agree on each fixed deployment. It fails on #3346 as-is (the `get.record.dynamic` value is still ambient) and passes with this fix.

```
cargo test -p snarkvm-synthesizer --features test --lib \
  test_conditional_dynamic_call_target_deployment_via_get_record_dynamic -- --nocapture
```

## Backwards compatibility

For the same reason as #3346 doesn't gate this behind a `ConsensusVersion` change, this one doesn't need that either. The change only alters the verification outcome of future deployments, and only for programs that would have caused the consensus network to fork.

### DIFF
--- a/synthesizer/process/src/stack/deploy.rs
+++ b/synthesizer/process/src/stack/deploy.rs
@@ -89,7 +89,10 @@ impl<N: Network> Stack<N> {
         &self,
         consensus_version: ConsensusVersion,
         deployment: &Deployment<N>,
-        rng: &mut R,
+        // Retained for API compatibility. Deployment verification is intentionally deterministic:
+        // all randomness is drawn from `seeded_rng` (seeded by the deployment ID) so that every
+        // validator computes the same result.
+        _rng: &mut R,
     ) -> Result<()> {
         let timer = timer!("Stack::verify_deployment");
 
@@ -202,7 +205,9 @@ impl<N: Network> Stack<N> {
                 is_root,
                 program_checksum,
                 false,
-                rng,
+                // Draw the request nonce from the seeded RNG as well, so no part of the
+                // `CheckDeployment` synthesis depends on the ambient RNG. See the burner note above.
+                &mut seeded_rng,
             )?;
             lap!(timer, "Compute the request for {}", function.name());
             // Initialize the assignments.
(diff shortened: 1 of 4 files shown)


Antonio95 (commented):
Great 👌

Antonio95 (approved):

PR #3366, opened 2026-08-11
Lower testnet MAX_CERTIFICATES to 40 at consensus V19

vicsn:
## Summary

- Set testnet `MAX_CERTIFICATES` to 40 when consensus V19 activates.
- Before V19, testnet keeps the current limit of 100 certificates per batch.
- Mainnet and canary keep their current limits at V19 (40 and 100).
- Update consensus constant tests so testnet can lower this limit at a new version.

## Why

Testnet currently allows up to 100 certificates per batch. Mainnet caps this at 40. Lowering testnet to 40 at V19 aligns testnet with mainnet and allows us to use a higher synthesis and spend limit.

## Backwards compatibility

The code contains a comment:
```
    //  Note: This value must **not** decrease without considering the impact on serialization.
    //  Decreasing this value will break backwards compatibility of serialization without explicit
    //  declaration of migration based on round number rather than block height.
```
Based on manual and AI review I don't see any issues because we only have 26 validators active on testnet. @ljedrz can you double check problematic usage of `LATEST_MAX_CERTIFICATES` in snarkOS? 

## Test plan

- [x] Pre-commit hook passes (`cargo clippy` and `cargo fmt --check`)
- [x] `cargo test -p snarkvm-console-network --lib consensus_constants --features test`

### DIFF
--- a/console/network/src/consensus_heights.rs
+++ b/console/network/src/consensus_heights.rs
@@ -69,8 +69,8 @@ pub enum ConsensusVersion {
     /// V18: Enables native credits record translation, introduces block-wide deployment limits,
     ///      and enforces canonical subDAG certificate ordering.
     V18 = 18,
-    /// V19: Adds more accurate type checking for the root call, and bounds the size of every
-    /// `PlaintextType` declared in a deployed program.
+    /// V19: Adds more accurate type checking for the root call, bounds the size of every
+    /// `PlaintextType` declared in a deployed program, and updates the number of validators.
     V19 = 19,
     /// V20: TBD
     V20 = 20,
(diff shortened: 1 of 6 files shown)


console/network/src/consensus_heights.rs
```diff
@@ -610,7 +610,6 @@ mod tests {
         consensus_config_returns_some::<CanaryV0>();
 
         max_certificates_increasing::<MainnetV0>();
-        max_certificates_increasing::<TestnetV0>();
```
ljedrz, line 613: perhaps we could special-case `V19`? if might be useful in the future

vicsn: [5808aa6](https://github.com/ProvableHQ/snarkVM/pull/3366/commits/5808aa650d605504cee29a84e7a1ebcaa9b7c195)

Antonio95 (approved):

PR #3367, opened 2026-08-11
[Refactor] Split the construction of the instruction set

### LINKED ISSUE #4386: [Bug] snarkvm-ledger-puzzle-epoch CI failures
I'm suspecting an OOM.

ljedrz:
`instruction_set<N>()` is currently a 2,700-line body with 7,771 calls to `vec![]`, which has a high MIR, borrow-checking, and metadata encoding cost, leading to frequent compile-time OOMs on snarkOS side ([related issue](https://github.com/ProvableHQ/snarkOS/issues/4386)).

The list was hashed (best-effort: `Debug` impls, as not everything impls `Display`) before and after chunking in order to ensure parity. I left that test in, so that we don't accidentally reorder entries in the future and have a convenient baseline to compare against.

Fixes https://github.com/ProvableHQ/snarkOS/issues/4386 (once snarkOS points to it).


vicsn (approved):
Epic

PR #3368, opened 2026-08-11
Fix ci integration

vicsn:
## Motivation

The synthesizer-integration job failed with:
```
    2026-08-11T10:12:30.758196Z  WARN snarkvm_synthesizer::vm::finalize: Speculation safely aborted a transaction - Struct 'Woo' is not defined. (at1042f3w77kuc7kucdtyg9zawdyq4kmxr6c3f6vvwa9h49ld69jsgsdyukg4)
    Speculation safely aborted a transaction - Struct 'Woo' is not defined. (at1042f3w77kuc7kucdtyg9zawdyq4kmxr6c3f6vvwa9h49ld69jsgsdyukg4)
    thread 'test_vm_execute_and_finalize' panicked at synthesizer/tests/test_vm_execute_and_finalize.rs:164:13:
    assertion failed: aborted_transaction_ids.is_empty()
```

The V19 check check_program_plaintext_sizes aborted deployments that used nested external structs.

Root cause: When sizing a type like child.aleo/BooHoo, the helper always looked up member types (Woo) on the current program. Woo lives only in child.aleo, so lookup failed with Struct 'Woo' is not defined, and speculation aborted the deploy.

That showed up in test_vm_execute_and_finalize for cases like future_with_external_struct / nested external structs.

Fix: Resolve each struct against the program that declares it (switch to the external stack for ExternalStruct, same pattern as plaintext_size_in_bytes / check_plaintext_type_inner), instead of always using the importer’s program.

Not sure if this conflicts with https://github.com/ProvableHQ/snarkVM/pull/3336

## Test Plan

Running extended CI on this PR.


Antonio95 (approved):
The logic seems correct, and in particular so does the handling of the `PlaintextType::ExternalStruct` case.

PR #3371, opened 2026-08-11
Bump the Rust toolchain to 1.96 and clear the resulting lints

cbeck88:
## Motivation

Updates the MSRV in rust-toolchain.toml, Cargo.toml, and .circleci/config.yml.

The immediate motivation is that some AVX-512 intrinsics are only stable from Rust 1.89.

But more generally, Rust 1.88 is from 1 year ago, and we may be missing compiler bug fixes or performance improvements.

The newer compiler reports 287 clippy findings in existing code.

Most are cleared mechanically by `cargo clippy --fix`; the bulk is `mismatched_lifetime_syntaxes` (149), plus collapsible `if`, useless `vec!`, and manual impls of derivable traits. The rest needed a human:

  - `cloned_ref_to_slice_refs` (28 sites): `&[x.clone()]` becomes `slice::from_ref(&x)`. Three sites in the instruction-operation macros are *not* convertible and the lint had to be suppressed instead.
  - `bigint_256.rs` / `bigint_384.rs`: `_addcarry_u64` and `_subborrow_u64` are safe functions on this toolchain, so the `unsafe` blocks wrapping them are now redundant.
  - `finalize.rs`: `notify_staking_reward` collapses into a let-chain.
  - `non_canonical_partial_ord_impl` now fires on the impl block rather than the method, so the existing allow moves with it.

## Test Plan

clippy and fmt fix and the tests are passing locally

## Backwards compatibility

No behavioural change is intended; every edit is a lint fix.

Note: we likely shouldn't merge this until the *next* release cycle

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -24,7 +24,7 @@ executors:
   # The default docker image used in the main-workflow
   rust-docker:
     docker:
-      - image: cimg/rust:1.88.0 # Attention - Change the MSRV in Cargo.toml and rust-toolchain as well
+      - image: cimg/rust:1.96.0 # Attention - Change the MSRV in Cargo.toml and rust-toolchain as well
     environment:
       RUSTFLAGS: "" # override the config.toml flags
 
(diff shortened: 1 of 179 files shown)


.circleci/config.yml
```diff
@@ -1324,7 +1324,7 @@ workflows:
       or pipeline.git.branch == "canary"
       or pipeline.git.branch == "testnet"
       or pipeline.git.branch == "mainnet"
-      or pipeline.git.branch == "fix_ci_integration"
+      or pipeline.git.branch == "chore/rust-1.96-toolchain"
```
cbeck88, line 1327: ^

cbeck88: i'm not sure if it will trigger because it's a PR from my fork. if it's better to open the PR directly from a branch in ProvableHQ/snarkVM i can reopen it i guess

PR #3372, opened 2026-08-11
Fix/memoize size in bits and types equivalent

cbeck88:
## Motivation

Prevent exponential recursion in `size_in_bits` and `types_equivalent`.

This is essentially the same issue https://github.com/ProvableHQ/snarkVM/pull/3357 but appearing in different functions.

This was also found by an AI scanner.

The fix is very similar, to use memoization to remember subcalls that we already evaluated so that we don't redo them.

However, the data that is memoized is a little more complex because we need to remember the total "height" we have seen so far; if we don't do that, then it is a subtle change in behavior from what the code does currently.

## Test Plan

We add regression tests that test that pathological cases evaluate correctly in a short amount of time.

## Backwards compatibility

This should be backwards compatible for the same reason that 3357 was -- we aren't changing behavior, we are just more efficiently determining the same answer.

### DIFF
--- a/console/program/src/data_types/finalize_type/size_in_bits.rs
+++ b/console/program/src/data_types/finalize_type/size_in_bits.rs
@@ -13,7 +13,7 @@
 // See the License for the specific language governing permissions and
 // limitations under the License.
 
-use crate::{DynamicFuture, Identifier, StructType};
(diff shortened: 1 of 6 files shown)


vicsn (approved):
LGTM, change is conceptually straightforward, existing test suite is quite extensive and new tests look to correctly target the DoS.

PR #3375, opened 2026-08-13
Revert to constraint-based deployment cap

vicsn:
## Summary
- Rollback to the old deployment cap model
- Bump the max cap from 2^21 to 2^22.
- Add consensus version V19, whose only job is to introduce this new deployment cap

## Test plan

- [x] CI mostly passed, only red jobs include:
  - a minor semver incompatibility exists
  - spurious failures not present in other runs or fixed on staging
- [x] Regression test passed, all existing program deployments fit within the new limit
- [x] Local snarkOS network deployment test passed, targeted program deployments fit within the new limit


synthesizer/process/src/stack/deploy.rs
```diff
@@ -120,11 +120,26 @@ impl<N: Network> Stack<N> {
...
-        if consensus_version < ConsensusVersion::V18 {
-            // Check that the number of combined variables does not exceed the deployment limit.
-            ensure!(deployment.num_combined_variables()? <= N::MAX_DEPLOYMENT_VARIABLES);
-            // Check that the number of combined constraints does not exceed the deployment limit.
-            ensure!(deployment.num_combined_constraints()? <= N::MAX_DEPLOYMENT_CONSTRAINTS);
+        // Enforce per-transaction variable and constraint limits, except at V18 where a block-wide
+        // synthesis limit is used instead.
+        let (variable_limit, constraint_limit) = if consensus_version < ConsensusVersion::V18 {
+            (Some(N::MAX_DEPLOYMENT_VARIABLES), Some(N::MAX_DEPLOYMENT_CONSTRAINTS))
+        } else if consensus_version >= ConsensusVersion::V19 {
```
Antonio95: Can this case be placed first to break early in the future?

Antonio95 (approved):

Antonio95 (approved):

Antonio95 (approved):

console/network/src/lib.rs
```diff
@@ -150,16 +150,24 @@ pub trait Network:
...
     /// The maximum number of constraints in a deployment. This limit was enforced at the transaction level up to
-    /// consensus version V16 (inclusive). This corresponds to ~0.5 second single-threaded runtime at mainnet
-    /// launch reference validator hardware.
+    /// consensus version V16 (inclusive), skipped at V18 in favor of a block-wide synthesis limit, and
+    /// replaced by `MAX_DEPLOYMENT_CONSTRAINTS_V2` from V19. This corresponds to ~0.5 second single-threaded
+    /// runtime at mainnet launch reference validator hardware.
     const MAX_DEPLOYMENT_CONSTRAINTS: u64 = 1 << 21; // 2,097,152 constraints
+    /// The maximum number of variables in a deployment. Enforced at the transaction level from consensus
+    /// version V19.
+    const MAX_DEPLOYMENT_VARIABLES_V2: u64 = 1 << 22; // 4,194,304 variables
```
raychu86, line 164: nit: We usually use arrays of tuples where the first element is the consensus version and the second is the value


In this case, it's fine as is for now, but should be refactored later.

raychu86 (approved):

console/network/src/lib.rs
raychu86: Created an [issue](https://github.com/ProvableHQ/snarkVM/issues/3378).

snarkworld (approved):

PR #3385, opened 2026-08-18
Stabilize Puzzle::prove benchmark CI reporting

vicsn:
## Summary
- Run `Puzzle::prove` in CI with 100 samples, 3s warmup, and 10s measurement so the bencher median is not dominated by the old 10-sample default.
- Keep `Puzzle::check_solutions` on the cheap path; document the local command in `ledger/puzzle/README.md`.

A 940% `Puzzle::prove` alert between `c04138c` and `f1f867b` did not reproduce locally (later SHA was faster). The 200% github-action-benchmark gate was firing on n=10 noise.

## Test plan
- [x] Confirm the `Benchmark ledger/puzzle` step still emits one `Puzzle::prove` bencher line and the `Puzzle::check_solutions` line(s).
- [x] After merge, treat the first post-change comparison as a new baseline for `Puzzle::prove` (sample size changed).

### DIFF
--- a/.github/workflows/benchmarks.yml
+++ b/.github/workflows/benchmarks.yml
@@ -115,7 +115,11 @@ jobs:
       - name: Benchmark ledger/puzzle
         run: |
           set -o pipefail
-          cargo bench --package=snarkvm-ledger-puzzle --bench puzzle --features=setup -- --output-format bencher | tee -a output.txt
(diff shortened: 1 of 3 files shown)


cbeck88 (approved):
looks fine to me, thank you

PR #3386, opened 2026-08-18
Apply a conservative spend_limit to beacon blocks similar to production environments

vicsn:
## Summary

1. Apply `spend_limit` to beacon blocks. Use a conservative `min_certificates` = `2 * ((MAX_CERTIFICATES + 2) / 3)`.
2. Set testnet `MAX_CERTIFICATES` to 40 from consensus V20. This will make the `spend_limit` in testnet beacon blocks reflect reality more closely.

Quorum `spend_limit` does not change.

## Test plan

- [x] `cargo test -p snarkvm-ledger-narwhal-subdag --lib --features test-helpers -- test_spend_limit test_min_limits`
- [x] `cargo test -p snarkvm-console-network --lib consensus_constants --features test`

## Backwards compatibility

The code contains a comment:
```
    //  Note: This value must **not** decrease without considering the impact on serialization.
    //  Decreasing this value will break backwards compatibility of serialization without explicit
    //  declaration of migration based on round number rather than block height.
```
Based on manual and AI review we don't see any issues because we only have 26 validators active on testnet. https://github.com/ProvableHQ/snarkVM/pull/3366#issuecomment-5250733492

### DIFF
--- a/ledger/narwhal/subdag/src/lib.rs
+++ b/ledger/narwhal/subdag/src/lib.rs
@@ -258,15 +258,14 @@ impl<N: Network> Subdag<N> {
         }
     }
 
-    /// Returns a lower-bound certificate count for a subdag at `block_height`.
+    /// Returns a lower-bound certificate count for a subdag with `max_certificates` per round.
     ///
-    /// For `N = MAX_CERTIFICATES(block_height)` written as `N = 3f + 1`, this is the integer
-    /// `2 * (f + 1)`. In general (including `N = 3f + 1 + k` with `0 <= k < 3`), this is two
-    /// rounds of the availability threshold: `2 * ((N + 2) / 3)`.
+    /// For `N = max_certificates` written as `N = 3f + 1`, this is the integer `2 * (f + 1)`.
+    /// In general (including `N = 3f + 1 + k` with `0 <= k < 3`), this is two rounds of the
+    /// availability threshold: `2 * ((N + 2) / 3)`.
     #[inline]
-    pub fn min_certificates(block_height: u32) -> u64 {
-        // unwrap: `MAX_CERTIFICATES` is defined for every consensus height.
-        let n = consensus_config_value!(N, MAX_CERTIFICATES, block_height).unwrap() as u64;
+    pub fn min_certificates(max_certificates: u16) -> u64 {
+        let n = max_certificates as u64;
         // `(N + 2) / 3 = f + 1` when `N = 3f + 1 + k` with `0 <= k < 3`.
         n.saturating_add(2).saturating_div(3).saturating_mul(2)
     }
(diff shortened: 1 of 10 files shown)


ledger/narwhal/subdag/src/lib.rs
```diff
@@ -258,6 +258,52 @@ impl<N: Network> Subdag<N> {
...
+        n.saturating_add(2).saturating_div(3).saturating_mul(2)
+    }
+
+    /// Returns the block spend limit for a subdag with `min_certificates` at `block_height`.
+    ///
+    /// Used for beacon blocks, which have no subdag but must still enforce block-wide limits.
+    #[inline]
+    pub fn min_spend_limit(block_height: u32) -> Option<u64> {
+        // unwrap: `CONSENSUS_HEIGHT` is defined for every `ConsensusVersion`.
+        if block_height >= N::CONSENSUS_HEIGHT(ConsensusVersion::V16).unwrap() {
```
eranrund, line 279: A bit of a nit, but we might want to test this in https://github.com/ProvableHQ/snarkVM/blob/fix/committee-and-beacon-spend-limit/synthesizer/src/vm/tests/test_v16/block_spend_limit.rs

That file currently contains a comment saying `Block-level compute spend limits for quorum blocks`, however there will now be a spend limit for beacon blocks as well, starting at V16.

ledger/narwhal/subdag/src/lib.rs
```diff
@@ -657,4 +703,28 @@ mod tests {
...
+    #[test]
+    fn test_min_limits_use_min_certificates() {
+        let v16_height = CurrentNetwork::CONSENSUS_HEIGHT(ConsensusVersion::V16).unwrap();
+        let v18_height = CurrentNetwork::CONSENSUS_HEIGHT(ConsensusVersion::V18).unwrap();
+
+        for height in [v16_height, v18_height, u32::MAX] {
+            let n = consensus_config_value!(CurrentNetwork, MAX_CERTIFICATES, height).unwrap() as u64;
+            let min_certs = n.saturating_add(2).saturating_div(3).saturating_mul(2);
+            assert_eq!(Subdag::<CurrentNetwork>::min_certificates(height), min_certs);
+            if n % 3 == 1 {
```
eranrund: Heads up that the crate does not enable the `snarkvm-console/test` feature, so we get the following:
```
height=19860000 n=40 min_certs=28
height=20794000 n=40 min_certs=28
height=4294967295 n=40 min_certs=28
```
I am not sure whether that is on purpose.

If we do enable the feature via dev-dependencies we will get:
```
height=19 n=104 min_certs=70
height=21 n=104 min_certs=70
height=4294967295 n=105 min_certs=70
```
but in that case `n % 3` will never be `1`.

eranrund (approved):
LGTM, some minor comments on test-related stuff.

ledger/narwhal/subdag/src/lib.rs
vicsn: We recently started getting issues with the duration of our tests because we don't cache block and tx creation. So I personally prefer to leave this type of test be. :)

ledger/narwhal/subdag/src/lib.rs
vicsn: Yeah that seems silly. I think the main value is unit testing behaviour of `min_certificates`, so sharpened this in: 852091a19

eranrund (approved):

PR #3387, opened 2026-08-19
Reject Signature::sign for request-like messages

vicsn:
Reject `Signature::sign` when `message[1] == hash_psd2([message[0]])`, so it cannot collide with the `Request::sign` preimage.

This implements the urgent part of https://github.com/ProvableHQ/snarkVM/pull/3284.


cbeck88 (approved):
looks good to me

PR #3388, opened 2026-08-24
[CI] Testnet v4.10.0

snarkworld:
## Motivation

This PR updates the testnet version to `v4.10.0` and sets the testnet `ConsensusVersion::V20` activation height to `19_374_000`.


vicsn (approved):
Pushed a commit to fix cargo audit

PR #3390, opened 2026-08-25
ci: pin the nightly that check-fmt formats with

cbeck88:
`install_rust_nightly` runs `rustup toolchain install nightly-...`, which resolves to whatever nightly exists on the day the job runs. `check-fmt` then formats with it. So the definition of "correctly formatted" is a moving target, and a pull request that changes nothing about formatting can go red because rustfmt's output moved that morning. It also means a contributor whose local nightly differs from the one CI happened to resolve gets a green check locally and a red one on the branch, with no way to tell why.

The formatter has to be a separate toolchain: `.rustfmt.toml` uses imports_layout, imports_granularity, overflow_delimited_expr, reorder_impl_items and style_edition, none of which exist on stable, while rust-toolchain.toml pins 1.88. Since it is a second toolchain, it wants pinning for the same reasons the first one does.

`fmt_nightly` sits with the other constants this file already keeps as pipeline parameters, so the version lives in one place and bumping it is one edit. `--profile minimal --component rustfmt` is added because the job needs the formatter and nothing else.

This is insurance rather than a repair. The tree reports zero diffs under all three formatters it was checked against --

    rustfmt 1.8.0-nightly  (74509131e8 2025-04-29)
    rustfmt 1.9.0-nightly  (7e46c5f6fb 2026-04-01)   <- the pin
    rustfmt 1.10.0-nightly (e7769602ac 2026-08-24)

so rustfmt's output for this codebase has not moved in sixteen months. The point is that nothing currently stops it from moving, and when it does the failure lands on whichever pull request happens to be open.

AGENTS.md names the pinned toolchain everywhere it previously said `cargo +nightly`, so that the documented commands and CI agree. CONTRIBUTING.md still says plain `cargo fmt`, which was already inaccurate for the reason above; it is left for a separate change rather than edited in passing here.

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -16,6 +16,22 @@ parameters:
   twoxlarge:
     type: string
     default: 2xlarge
+  # The toolchain `check-fmt` formats with.
+  #
+  # `.rustfmt.toml` uses options that only exist on nightly -- imports_layout,
+  # imports_granularity, overflow_delimited_expr, reorder_impl_items and
+  # style_edition -- so the formatter cannot be the 1.88.0 this project builds
+  # with, and `cargo fmt` without a toolchain silently drops all five rather
+  # than refusing. The formatter is therefore a second toolchain, and pinning
+  # it is the same act as pinning the first.
+  #
+  # Resolves to rustfmt 1.9.0-nightly (7e46c5f6fb 2026-04-01); rustup's dated
+  # nightlies are built from the previous day's commit. Bumping it may reformat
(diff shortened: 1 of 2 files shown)


vicsn (approved):
LGTM, can you make the same change for snarkOS? 

I dont think the veria report is an urgent issue but please independently think about it. I think the main attack vector is exfiltration of CI secrets, though currently CI is not triggerable by external contributors. I would opt to document it as a defense in depth issue.

PR #3391, opened 2026-08-26
fix(serialize): cap eager Vec<T> deserialization reserve

cbeck88:
## Motivation

Vec<T>'s CanonicalDeserialize reads a u64 length prefix off the wire and eagerly reserves that many elements' worth of capacity before reading any actual elements. Since the length is attacker-controlled on any P2P-deserialized payload, a small message can force a large upfront allocation. Cap the eager reserve to a small hint; the vector still reaches its true size via push, bounded by how much data the reader actually has.

## Test Plan

N/A

## Backwards compatibility

N/A

### DIFF
--- a/utilities/src/serialize/impls.rs
+++ b/utilities/src/serialize/impls.rs
@@ -421,7 +421,10 @@ impl<T: CanonicalDeserialize> CanonicalDeserialize for Vec<T> {
     ) -> Result<Self, SerializationError> {
         let len = u64::deserialize_with_mode(&mut reader, compress, validate)?;
         let mut values = Vec::new();
-        let _ = values.try_reserve(len as usize);
+        // `len` comes straight off the wire, so only pre-reserve a small hint from it; a huge
+        // value must not force a large upfront allocation. The vector still reaches its real
+        // size via `push`, which only grows as far as the reader actually has data for.
+        let _ = values.try_reserve(len.min(1024) as usize);
         for _ in 0..len {
             values.push(T::deserialize_with_mode(&mut reader, compress, Validate::No)?);
         }


utilities/src/serialize/impls.rs
```diff
@@ -421,7 +421,10 @@ impl<T: CanonicalDeserialize> CanonicalDeserialize for Vec<T> {
     ) -> Result<Self, SerializationError> {
         let len = u64::deserialize_with_mode(&mut reader, compress, validate)?;
         let mut values = Vec::new();
-        let _ = values.try_reserve(len as usize);
+        // `len` comes straight off the wire, so only pre-reserve a small hint from it; a huge
```
ljedrz, line 424: Which network objects involve this code path? snarkOS deserialization is almost universally based on `FromBytes`/`ToBytes`, I wouldn't expect to see this coming via the wire at all, rather from the hardcoded parameters.

cbeck88: here is one path (snarkOS staging 77308d1):

Varuna `Proof` is the one place in the transaction encoding where `FromBytes` hands off to `CanonicalDeserialize`,
so the `Vec<T>` impl does end up on a wire-decoded path.

`Proof<N>`'s `FromBytes` defers to the inner varuna proof: synthesizer/snark/src/proof/bytes.rs#L18-L31:

```rust
impl<N: Network> FromBytes for Proof<N> {
        let version = u8::read_le(&mut reader)?;
        if version != 1 { return Err(error("Invalid proof version")); }
        let proof = FromBytes::read_le(&mut reader)?;   // <-- varuna Proof<E>
        Ok(Self { proof })
    }
```

…whose `FromBytes` is just `CanonicalDeserialize`: algorithms/src/snark/varuna/data_structures/proof.rs#L377-L381:

```rust
impl<E: PairingEngine> FromBytes for Proof<E> {
    fn read_le<R: Read>(mut r: R) -> io::Result<Self> {
        Self::deserialize_compressed(&mut r)
            .map_err(|err| into_io_error(anyhow::Error::from(err).context("could not deserialize Proof")))
    }
```

…and the first field it reads is a `Vec<u64>` - proof.rs#L344-L351:

```rust
impl<E: PairingEngine> CanonicalDeserialize for Proof<E> {
    fn deserialize_with_mode<R: Read>(mut reader: R, compress: Compress, validate: Validate)
        -> Result<Self, SerializationError>
    {
        let batch_sizes: Vec<u64> = CanonicalDeserialize::deserialize_with_mode(&mut reader, compress, validate)?;
```

So the first 8 bytes after the proof version byte are the length that reaches the `try_reserve` this PR touches.
This can lead to `try_reserve(800 MB)` , but not with the change in this PR.

cbeck88: Here's another:

```rust
  // proof.rs:364
  pc_proof: CanonicalDeserialize::deserialize_with_mode(&mut reader, compress, validate)?,
```

`pc_proof: sonic_pc::BatchLCProof<E>` → `BatchProof<E>(Vec<kzg10::KZGProof<E>>)` (data_structures.rs:401), derived `CanonicalDeserialize` → the generic `Vec<T>` impl this PR patches.

cbeck88: There's also a path via `Deployment` -> `Certificate`:

`Deployment::read_le` reads up to `MAX_FUNCTIONS + MAX_RECORDS = 341` entries of `(Identifier, (VerifyingKey, Certificate))`. `Certificate::read_le` →
  `deserialize_compressed` → `derived` → `BatchLCProof` → the same uncapped `Vec<KZGProof>`.

ljedrz (approved):
LGTM as a hardening.

The reason why I didn't expect this to be present in network messages is that they had all been fuzz-tested, and there were neither OOMs, nor did the memory "jump" during those runs; the latter is most likely due to the allocator not actually dedicating that many bytes if the payload doesn't actually support it, and those are restricted by the network message framing.

PR #3393, opened 2026-08-26
Fix/bound varuna proof lengths

cbeck88:
## Motivation

We read serialized `varuna::Proof` and `sonic_pc::BatchProof` from the wire, and so bounding the maximum sizes of these that will be accommodated by the deserializer helps to harden the node against abuse.

`varuna::Proof` and `sonic_pc::BatchProof` both begin with a length taken straight from the reader, and neither bounds it.

`Proof::deserialize_with_mode` reads `batch_sizes` as a length-prefixed `Vec<u64>`. Those values are not inert — they become loop counts and expected lengths for everything that follows (`Commitments`, `Evaluations`, the third and fourth prover messages), and they are carried on the deserialized `Proof` for the verifier. `BatchProof` reads its evaluation proofs the same way, and is reached twice over: once as `Proof.pc_proof`, and once as `Certificate.pc_proof` via `BatchLCProof`.

Both are reachable from ordinary deserialized objects — a proof through any execution or fee transaction, a certificate through any deployment — so the counts driving that work are whatever the sender wrote.

These are protocol constants, and other code already depends on their values: `SonicKZG10::batch_check` requires exactly one evaluation proof per distinct query point name, and the verifier computes `batch_size - 1`. This PR checks them at the point where the length is read, so that a malformed proof is rejected on its first few bytes rather than surviving into code that assumes the check already happened. `CircuitVerifyingKey` already does exactly this for its commitments (`MAX_CIRCUIT_COMMITMENTS = 12`); these are the two remaining length-prefixed reads on the same path.

To be precise: This is **not** a fix for a live memory-exhaustion bug. The generic `Vec<T>` deserializer now caps its `try_reserve` at 1024 elements (#3391, since merged to staging), and `deserialize_vec_without_len` collects into a `Result<Vec<_>, _>`, whose size hint reserves nothing — so an implausible length dies on the first short read. The value here is validation, failing fast, and not leaving the property resting on those two incidental details, and consistency with the circuit verifying key.

Complements #3391, which approached the same class from the other end by bounding the generic `Vec<T>` reserve. The two are independent: this PR bounds the specific reachable sites with their semantic limits, #3391 is a backstop for sites nobody has audited.

## Test Plan

New tests in `algorithms/src/snark/varuna/data_structures/proof.rs`:

- `test_proof_batch_sizes_are_bounded` — covers all six rejection paths (zero circuits, a zero-instance circuit, too many circuits, one circuit holding more than the whole batch allows, an instance count that would wrap the running total past zero, and too many instances in total). It also asserts that an in-bounds blob fails with `IoError` rather than `InvalidData`, so the test distinguishes "the counts were rejected" from "the buffer was short."
- `test_batch_proof_length_is_bounded` — round-trips every length the protocol can produce (0–3), rejects 4 on read, and rejects 4 on write.
    -  `test_serializing_proof` previously built a 10-element `BatchProof`; it now builds at most `MAX_BATCH_PROOF_LEN`. That was the only place in the tree constructing an out-of-spec one.

Verified on an idle machine, built from the pushed commit:

- `cargo test -p snarkvm-algorithms -p snarkvm-console-network --lib` — 87/87
  and 7/7, including every varuna end-to-end `prove_and_verify_*` test and the
  hiding variants, so real proofs still round-trip through the new bounds.
- `cargo +nightly-2026-04-02 fmt --all -- --check` clean.
- The compile-time assertions were checked to actually fire, not merely
  compile: lowering either `MAX_CIRCUITS_PER_BATCH_PROOF` or
  `MAX_INSTANCES_PER_BATCH_PROOF` below the consensus limit it pins produces
  `error[E0080]: evaluation of constant value failed`.

## Documentation

N/A

## Backwards compatibility

No consensus version guard needed.

The wire format is unchanged — this only changes which byte strings are accepted, and every bound sits above anything either consensus era can produce:

- Circuits ≤ 128. From V14 this follows from `MAX_BATCH_PROOF_INSTANCES = 128`, since every circuit holds at least one instance. Before V14 that limit did not apply, so the bound comes from the transitions instead: an execution carries fewer than `Transaction::MAX_TRANSITIONS` (32) transitions, one circuit each, plus one for inclusion — at most 32.
- Instances across the whole batch ≤ 1024, by the same two-era split as the circuit bound:
  - From V14, `Network::MAX_BATCH_PROOF_INSTANCES` (128) caps the total directly. It counts every instance an execution proves — transition, record-translation and inclusion alike (`verify_execution/mod.rs:321-331`).
  - Before V14 that limit did not apply, but neither did record translation: it requires `dynamic` syntax, which `contains_v14_syntax` refuses to deploy before V14 (`vm/verify.rs:341-344`), and pre-V14 deployments may not carry record verifying keys (`vm/verify.rs:347-352`). So a batch holds only function and inclusion instances: at most 31 transitions × `MAX_INPUTS` (16) = 496 inclusion instances, plus one per transition for the function circuits — fewer than 528.

  Deliberately not tightened to 128 — pre-V14 proofs can legitimately exceed that, and tightening would reject historical blocks. There is no separate per-circuit cap: a single oversized circuit is caught by this same total.
- Evaluation proofs ≤ 3, one per distinct query point name. The Varuna AHP queries at three ("alpha", "beta", "gamma"), the certificate at one ("challenge"). Unchanged since before mainnet launch, and independent of circuit count, hiding mode and `VarunaVersion`.

BatchProof now also refuses to serialize more than three, so the round trip stays total. Nothing in the repository constructs one that large; test_templates.rs holds a BatchProof in TestComponents but never serializes it.

### DIFF
--- a/algorithms/src/polycommit/sonic_pc/data_structures.rs
+++ b/algorithms/src/polycommit/sonic_pc/data_structures.rs
@@ -396,9 +396,58 @@ impl<'a, E: PairingEngine> CommitterUnionKey<'a, E> {
     }
 }
 
+/// The maximum number of evaluation proofs in a `BatchProof`.
+///
+/// `SonicKZG10::batch_open` emits exactly one proof per distinct query point
+/// name. The Varuna AHP queries at three names ("alpha", "beta", "gamma") and
+/// the verifying key certificate at one ("challenge"), so no `BatchProof` this
+/// crate can produce holds more than three. `batch_check` already rejects any
+/// other count; bounding it here refuses the length before it is used.
+pub const MAX_BATCH_PROOF_LEN: u64 = 3;
+
 /// Evaluation proof at a query set.
-#[derive(Clone, Debug, Default, PartialEq, Eq, Hash, CanonicalSerialize, CanonicalDeserialize)]
+///
+/// `CanonicalDeserialize` and `Valid` are written out rather than derived, so
+/// that the length prefix can be bounded by `MAX_BATCH_PROOF_LEN` before any
+/// element is read. Both are needed because deriving `CanonicalDeserialize`
+/// is also what derives `Valid`.
+#[derive(Clone, Debug, Default, PartialEq, Eq, Hash, CanonicalSerialize)]
 pub struct BatchProof<E: PairingEngine>(pub(crate) Vec<kzg10::KZGProof<E>>);
+
+impl<E: PairingEngine> Valid for BatchProof<E> {
+    fn check(&self) -> Result<(), SerializationError> {
+        Valid::check(&self.0)
+    }
+
+    fn batch_check<'a>(batch: impl Iterator<Item = &'a Self> + Send) -> Result<(), SerializationError>
+    where
+        Self: 'a,
+    {
+        let batch: Vec<_> = batch.collect();
+        Valid::batch_check(batch.iter().map(|v| &v.0))
+    }
+}
+
+impl<E: PairingEngine> CanonicalDeserialize for BatchProof<E> {
+    fn deserialize_with_mode<R: Read>(
+        mut reader: R,
+        compress: Compress,
+        validate: Validate,
+    ) -> Result<Self, SerializationError> {
+        let len = u64::deserialize_with_mode(&mut reader, compress, validate)?;
+        if len > MAX_BATCH_PROOF_LEN {
+            return Err(SerializationError::InvalidData);
+        }
+        let mut proofs = Vec::with_capacity(len as usize);
+        for _ in 0..len {
+            proofs.push(kzg10::KZGProof::deserialize_with_mode(&mut reader, compress, Validate::No)?);
+        }
+        if let Validate::Yes = validate {
+            kzg10::KZGProof::<E>::batch_check(proofs.iter())?;
+        }
+        Ok(BatchProof(proofs))
+    }
+}
 
 impl<E: PairingEngine> BatchProof<E> {
     pub fn is_hiding(&self) -> bool {
--- a/algorithms/src/polycommit/sonic_pc/data_structures.rs
+++ b/algorithms/src/polycommit/sonic_pc/data_structures.rs
@@ -440,8 +440,7 @@ impl<E: PairingEngine> Valid for BatchProof<E> {
     where
         Self: 'a,
     {
-        let batch: Vec<_> = batch.collect();
-        Valid::batch_check(batch.iter().map(|v| &v.0))
+        Valid::batch_check(batch.map(|v| &v.0))
     }
 }
 
--- a/algorithms/src/polycommit/sonic_pc/data_structures.rs
+++ b/algorithms/src/polycommit/sonic_pc/data_structures.rs
@@ -396,23 +396,40 @@ impl<'a, E: PairingEngine> CommitterUnionKey<'a, E> {
     }
 }
 
-/// The maximum number of evaluation proofs in a `BatchProof`.
+/// The maximum number of evaluation proofs a `BatchProof` will serialize.
 ///
-/// `SonicKZG10::batch_open` emits exactly one proof per distinct query point
-/// name. The Varuna AHP queries at three names ("alpha", "beta", "gamma") and
-/// the verifying key certificate at one ("challenge"), so no `BatchProof` this
-/// crate can produce holds more than three. `batch_check` already rejects any
-/// other count; bounding it here refuses the length before it is used.
+/// `SonicKZG10::batch_open` emits one proof per distinct query point name, so
+/// the count follows from the caller's query set rather than from the scheme.
+/// Both callers in this crate sit at or below three: the Varuna AHP queries at
+/// "alpha", "beta" and "gamma", and the verifying key certificate at
+/// "challenge". `batch_check` already rejects any other count.
+///
+/// This is enforced on serialization as well as deserialization, so that the
+/// round trip stays total -- whatever writes will read back. A caller with a
+/// larger query set would have to raise it.
 pub const MAX_BATCH_PROOF_LEN: u64 = 3;
 
 /// Evaluation proof at a query set.
 ///
-/// `CanonicalDeserialize` and `Valid` are written out rather than derived, so
-/// that the length prefix can be bounded by `MAX_BATCH_PROOF_LEN` before any
-/// element is read. Both are needed because deriving `CanonicalDeserialize`
-/// is also what derives `Valid`.
-#[derive(Clone, Debug, Default, PartialEq, Eq, Hash, CanonicalSerialize)]
+/// `CanonicalSerialize`, `CanonicalDeserialize` and `Valid` are written out
+/// rather than derived, so that the length can be bounded by
+/// `MAX_BATCH_PROOF_LEN` on both sides. All three are needed because deriving
+/// `CanonicalDeserialize` is also what derives `Valid`.
+#[derive(Clone, Debug, Default, PartialEq, Eq, Hash)]
 pub struct BatchProof<E: PairingEngine>(pub(crate) Vec<kzg10::KZGProof<E>>);
+
+impl<E: PairingEngine> CanonicalSerialize for BatchProof<E> {
+    fn serialize_with_mode<W: Write>(&self, mut writer: W, compress: Compress) -> Result<(), SerializationError> {
+        if self.0.len() as u64 > MAX_BATCH_PROOF_LEN {
+            return Err(SerializationError::InvalidData);
+        }
+        CanonicalSerialize::serialize_with_mode(&self.0, &mut writer, compress)
+    }
+
+    fn serialized_size(&self, compress: Compress) -> usize {
+        CanonicalSerialize::serialized_size(&self.0, compress)
+    }
+}
 
 impl<E: PairingEngine> Valid for BatchProof<E> {
     fn check(&self) -> Result<(), SerializationError> {
--- a/algorithms/src/snark/varuna/data_structures/proof.rs
+++ b/algorithms/src/snark/varuna/data_structures/proof.rs
@@ -34,21 +34,31 @@ use std::mem::size_of;
 
 /// The maximum number of circuits in a batch proof.
 ///
-/// Every circuit in a batch contributes at least one instance, so the number of
-/// circuits is bounded by the consensus limit on the total number of instances
-/// in a batch proof, `Network::MAX_BATCH_PROOF_INSTANCES`. That limit lives in
-/// `snarkvm-console-network`, which this crate cannot depend on, so the value
-/// is repeated here: if it is ever raised, this constant has to be raised
-/// first, or newly produced proofs stop being deserializable.
+/// Two different arguments bound this, because the consensus limit on the total
+/// number of instances in a batch proof, `Network::MAX_BATCH_PROOF_INSTANCES`
+/// (128), is only applied from `ConsensusVersion::V14` and so says nothing
+/// about older proofs:
+///
+/// - From V14, every circuit holds at least one instance, so the number of
+///   circuits cannot exceed that limit of 128.
+/// - Before V14, a batch held one circuit per distinct function in the
+///   execution plus one for inclusion, and an execution carries fewer than
+///   `Transaction::MAX_TRANSITIONS` (32) transitions, giving at most 32.
+///
+/// 128 therefore covers both eras. It repeats a value from
+/// `snarkvm-console-network`, which this crate cannot depend on; a compile-time
+/// assertion beside that limit keeps the two from drifting apart.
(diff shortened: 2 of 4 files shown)


algorithms/src/polycommit/sonic_pc/data_structures.rs
```diff
@@ -396,10 +396,76 @@ impl<'a, E: PairingEngine> CommitterUnionKey<'a, E> {
...
+    fn check(&self) -> Result<(), SerializationError> {
+        Valid::check(&self.0)
+    }
+
+    fn batch_check<'a>(batch: impl Iterator<Item = &'a Self> + Send) -> Result<(), SerializationError>
+    where
+        Self: 'a,
+    {
+        let batch: Vec<_> = batch.collect();
+        Valid::batch_check(batch.iter().map(|v| &v.0))
```
ljedrz:
```suggestion
        Valid::batch_check(batch.map(|v| &v.0))
```
wouldn't this work, since `batch` is already an iterator?

algorithms/src/polycommit/sonic_pc/data_structures.rs
```diff
@@ -396,10 +396,76 @@ impl<'a, E: PairingEngine> CommitterUnionKey<'a, E> {
...
+/// `SonicKZG10::batch_open` emits one proof per distinct query point name, so
+/// the count follows from the caller's query set rather than from the scheme.
+/// Both callers in this crate sit at or below three: the Varuna AHP queries at
+/// "alpha", "beta" and "gamma", and the verifying key certificate at
+/// "challenge". `batch_check` already rejects any other count.
+///
+/// This is enforced on serialization as well as deserialization, so that the
+/// round trip stays total -- whatever writes will read back. A caller with a
+/// larger query set would have to raise it.
+pub const MAX_BATCH_PROOF_LEN: u64 = 3;
```
ljedrz, line 410: a general note that'll apply to the other `const`s as well: I'd feel more confident if instead of a concrete value, it was derived from one that's rooted in the network/proving setup; in this case we might be able to introduce some `enum` that would hold "alpha", "beta", and "gamma" (and use that one in the original location), and derive the `3` from the number of variants in that `enum`; that way, it would automatically be adjusted if that number ever changes; for the other `const`s it's simpler - I'd derive the values from the reference `const`s

algorithms/src/snark/varuna/data_structures/proof.rs
```diff
@@ -347,8 +379,29 @@ impl<E: PairingEngine> CanonicalDeserialize for Proof<E> {
...
+            let batch_size = u64::deserialize_with_mode(&mut reader, compress, validate)?;
+            if batch_size == 0 || batch_size > MAX_INSTANCES_PER_CIRCUIT {
+                return Err(SerializationError::InvalidData);
+            }
+            // Cannot overflow, being at most 128 * 512.
+            total_instances += batch_size;
+            if total_instances > MAX_INSTANCES_PER_BATCH_PROOF {
+                return Err(SerializationError::InvalidData);
+            }
+            batch_sizes.push(usize::try_from(batch_size)?);
```
ljedrz, line 413: this seems like a change in behavior, no? the former code would truncate, while this one would break for sizes > `u32` on 32-bit architectures; not saying the former behavior was correct, just pointing out a difference

ljedrz (commented):
LGTM in general, left some suggestions; I'm not too knowledgeable about this code, so I'd recommend for someone else to greenlight it.

algorithms/src/snark/varuna/data_structures/proof.rs
cbeck88: yes, i didn't think i was changing behavior, let me review

algorithms/src/polycommit/sonic_pc/data_structures.rs
cbeck88: yup thank you

algorithms/src/polycommit/sonic_pc/data_structures.rs
cbeck88: one problem for `MAX_BATCH_PROOF_LEN` is that there's actually a 4th point "challenge" but a single proof only actually holds 3. so the enum approach would give a value of 4.

another issue about the enum is it mixes Varuna concepts into sonic_pc which breaks separation of concerns

--

we do currently test that the constant is correct, because `prove_and_verify_*` serializes real proofs, so if this constant were too tight, those tests would fail.

i think we could do some more specific test where we test this constant against `QuerySet::to_set()` or something to ensure it doesn't drift

cbeck88: i'll check if it's possible for the others

algorithms/src/snark/varuna/data_structures/proof.rs
cbeck88: i think its not a change in behavior because we test `batch_size > MAX_INSTANCES_PER_CIRCUIT` above, and return an error in that case. so at this line we already know `batch_size < 512`.

it might make the code a little cleaner if we change it to `batch_size as usize`, or change the type of `batch_sizes` so that conversions are not necessary. i'll iterate on it

algorithms/src/polycommit/sonic_pc/data_structures.rs
cbeck88: for `const MAX_BATCH_PROOF_INSTANCES`, it seems that the crate dependencies don't let us do that:

`snarkvm-console-network` depends on `snarkvm-algorithms`, not the reverse, so algorithms cannot see
  MAX_BATCH_PROOF_INSTANCES.

The `const _: () = { assert!(...) }` block at console/network/src/lib.rs:644 is the cleanest way I can see right now to enforce that the constant matches the code

algorithms/src/snark/varuna/data_structures/proof.rs
cbeck88: it turns out we can't really change the type because varuna uses usize for batch_size everywhere, so that breaks the api. i think the best thing is a comment explaining that this conversion can't fail.

algorithms/src/snark/varuna/data_structures/proof.rs
```diff
@@ -32,6 +32,44 @@ use std::{
...
+/// compile-time assertion in `snarkvm-console-network` pins it above the 496.
+pub const MAX_INSTANCES_PER_CIRCUIT: u64 = 512;
+
+/// The maximum number of instances across all circuits in a batch proof.
+///
+/// The inclusion circuit contributes the 496 above, and each function circuit
+/// one instance per transition, so a batch holds fewer than 528. As with the
+/// per-circuit bound this keeps room to spare, and a compile-time assertion in
+/// `snarkvm-console-network` pins it above that sum.
+pub const MAX_INSTANCES_PER_BATCH_PROOF: u64 = 1024;
```
cbeck88, line 72: this and the two other constants above are tested in the const assertions block in console/network...

algorithms/src/snark/varuna/data_structures/proof.rs
```diff
@@ -32,6 +32,44 @@ use std::{
...
+///
+/// The largest legitimate batch is the inclusion circuit, which holds one
+/// instance per input record. An execution carries at most
+/// `Transaction::MAX_TRANSITIONS - 1` transitions, one slot being held back
+/// for the fee (see `Transaction::check_execution_size`), so at most
+/// 31 * `Network::MAX_INPUTS` (16) = 496 records. Proofs predating the
+/// consensus limit on total instances can reach that, so this is deliberately
+/// looser than the 128 above; tightening it would reject historical blocks. A
+/// compile-time assertion in `snarkvm-console-network` pins it above the 496.
+pub const MAX_INSTANCES_PER_CIRCUIT: u64 = 512;
```
vicsn: Thank you for making this!

Shall we get rid of `const MAX_INSTANCES_PER_CIRCUIT` as it adds only a marginal amount of security and has different naming?

vicsn (changes requested):

console/network/src/lib.rs
```diff
@@ -651,10 +651,16 @@ const _: () = {
...
         assert!(N::MAX_BATCH_PROOF_INSTANCES as u64 <= snarkvm_algorithms::snark::varuna::MAX_CIRCUITS_PER_BATCH_PROOF);
-        // Across the whole batch: the inclusion circuit holds one instance per input record over
-        // every transition but the fee -- `Transaction::MAX_TRANSITIONS` is `MAX_FUNCTIONS + 1`
-        // (see `test_transaction_depth_is_correct`), so that is `MAX_FUNCTIONS` transitions --
-        // plus one instance per transition for the function circuits.
+        // From `ConsensusVersion::V14` this limit caps an execution's batch directly, counting
+        // transition, record-translation and inclusion instances alike.
+        assert!(
+            N::MAX_BATCH_PROOF_INSTANCES as u64 <= snarkvm_algorithms::snark::varuna::MAX_INSTANCES_PER_BATCH_PROOF
+        );
```
cbeck88, line 658: @Antonio95 this is the new compile time assertion relevant to your comment. it connects the consensus network maximum on instances (declared on the network type N) to the constant we added here for the deserializer to use

algorithms/src/polycommit/sonic_pc/data_structures.rs
```diff
@@ -396,10 +396,75 @@ impl<'a, E: PairingEngine> CommitterUnionKey<'a, E> {
...
```
Antonio95, line 410: I agree with @ljedrz that having the constant 3 here feels somewhat arbitrary. I'm not advocating to make this module super general (e.g. to prepare it for use by external developers), but nothing about a sonic PCS proof intrinsically says it should be a batched proof for at most three points.

@cbeck88 @ljedrz one approach I think is more clean (but might be too annoying to implement) is to make `BatchProof` generic on `N: usize`, which would refer to the maximum number of points that are opened. We could also use the chance to convert the `Vec<kzg10::KZGProof<E>>` vector it wraps around into an array (but we likely don't want to change how it is serialised for backwards compatibility, i.e. we may still want to preface the serialisation of the array with the length; also, since I just claimed `N` would be the *maximum* of points it can refer to, it might be good to leave it as a vector anyway). If we change the module constant to a generic  constant `BatchProof<N>`, then executions would come with a `BatchProof<3>` and certificates would come with a `BatchProof<2>`.

What do you (both) think? In any case, I'm not strongly opposed to the much simpler, hardcoded-3 approach from the PR if we want to merge quickly.

console/network/src/lib.rs
Antonio95: Do these two constants represent the same thing? (two separate variants of this question: before V14 and at V14+). The fact that `N::MAX_BATCH_PROOF_INSTANCES` was originally not documented is somewhat annoying.

Antonio95 (commented):
Seems largely okay to me, but let's discuss the comments I left before going ahead.

For reference, when I was trying to convince myself this doesn't need to be consensus-gated, I came to the question of whether, before this change, a malicious prover could add spurious points to the batched Sonic PCS proof (which the verifier would simply ignore) - this would be an issue since such a proof would be accepted before this change but rejected after it. Fortunately, that could not be done before either due to this assertion: https://github.com/ProvableHQ/snarkVM/blob/staging/algorithms/src/polycommit/sonic_pc/mod.rs#L503.

@cbeck88 I'm wondering about consensus gating on the other front this PR adds, i.e. the total number of changes. Let's discuss this in this PR's thread on `N::MAX_BATCH_PROOF_INSTANCES`.

console/network/src/lib.rs
Antonio95: In particular, are we now bounding the total number of instances of a batched proof by 128? (which is certainly much less than the total number of function, inclusion and translation proofs an Execution could theoretically contain).

If so, why make the new bound 1028, i.e. why allow the deserialiser to read a lot more instances than we will allow later down the pipeline? Is it to encompass both the pre- and post-V14 situations (because the deserialiser doesn't have access to the ConsensusVersion)

algorithms/src/polycommit/sonic_pc/data_structures.rs
cbeck88: > but nothing about a sonic PCS proof intrinsically says it should be a batched proof for at most three points.

thank you, this helps to clarify what is wrong with the current code

> one approach I think is more clean (but might be too annoying to implement) is to make BatchProof generic on `N: usize`

my gut is that this is too disruptive of a change for the benefit that it is worth

maybe it's acceptable if there is a generic `BatchProof<N: usize>` or something, but we re-export it like

```
using BatchProof = sonic_pcs::BatchProof<3>;
```

from the consumer of `sonic_pcs` ?

But it's still kind of ugly, and will lead to uglier error messages.

I want to take a step back and think some more about it, maybe there is another way that the decoder can work that doesn't require us to impose explicit bounds like this.

PR #3394, opened 2026-08-26
Perf/cpu proving algos

cbeck88:
## Motivation

An agent noticed that we don't use Ruffini's rule for division when doing CPU proving and use a naive algorithm instead.

This kicked off a swarm of agents looking for similar algorithmic improvements to improve the speed of CPU proving, and they found 12 improvements. I reviewed each of these changes and they seem simple and correct to me, and the measurement process we converged on seems very sound to me.

### Speed up CPU proving: −8% on a private transfer

Twelve changes to the prover, each of which deletes arithmetic or allocation
without changing a single field element that comes out. Measured end to end on
`credits.aleo/transfer_private`:

| threads | before | after | change |
|---:|---:|---:|---:|
| **16** | 5.863 s | 5.390 s | **−8.06%** (−474 ms) |
| **4** | 12.545 s | 12.055 s | **−3.91%** (−490 ms) |

The first commit adds the benchmark these numbers come from, so the claim is
reproducible from this branch:

```
cargo bench -p snarkvm-ledger --bench transaction --features bench-prove \
    -- 'Execute\(transfer_private\)'
```

### Why

Proving is dominated by multi-scalar multiplication, which is already well
optimized, so there is no single large win available here. The work *around* the
MSM is a different matter: several routines on the proving path do more
arithmetic, or more allocation, than the operation they implement requires.
Individually most are small. Together they are the numbers above.

Two things this helps:

1. **Anyone proving locally.** Proof latency is the whole experience for a user
   generating their own transactions.
2. **This repository's own tests and CI.** Proving dominates the synthesizer and
   ledger suites, so anything that shortens a proof shortens all of them.

The saving is *larger* on more cores — −8.1% at 16 threads against −3.9% at 4 —
and the absolute saving is nearly identical at both (474 ms and 490 ms). That is
the signature of deleted work rather than improved parallelism: most of these
changes remove **serial** work, so by Amdahl's law they matter more as core count
rises, and the benefit grows on larger machines instead of shrinking.

### The changes

Each is a single commit, and each stands on an argument you can check by reading
it, independent of the measurement.

| # | commit | what it deletes |
|---|---|---|
| 1 | add the proving benchmark | *(no behaviour change)* |
| 2 | divide by the vanishing polynomial in closed form | `p = q(Xⁿ−1)+r` matched coefficient-wise is `n` independent carry chains; for `deg p < 2n` — every call the prover makes — the quotient is literally `p[n..]`. Generic long division rediscovers this one coefficient at a time |
| 3 | Ruffini's rule for `X − z` | one multiply-add per coefficient instead of two multiplications, a clone of the dividend, and a degree re-check every iteration. The remainder falls out as `self(z)` |
| 4 | Horner's rule for `evaluate` | `deg` multiplications instead of `2·deg`, and no vector of powers the size of the polynomial |
| 5 | scale the rowcheck in place | `lhs += &(&rowcheck * c)` onto a zero polynomial is three passes and two allocations; the accumulation it is named for only ever has one term |
| 6 | fuse `add_assign`'s zero branch | copy-then-scale walks the coefficients twice to produce what one walk produces |
| 7 | `has_lc_eval` | the "is this polynomial present?" probe was answered by evaluating it and discarding the value, twice per matrix |
| 8 | hoist the Lagrange coefficients | they depend only on the constraint domain and `alpha`, both fixed across the `a`/`b`/`c` jobs, and were built three times per circuit |
| 9 | borrow the FFT precomputations | root-of-unity tables are circuit data, fixed for the life of the process, and were deep-copied per circuit per proof |
| 10 | selector `xg · (v_H / v_H_i)` as a repeat | the quotient of the two vanishing polynomials is `1 + Xᵐ + X²ᵐ + …`, so the product is `xg` repeated. It was a multiply and a divide over the larger domain to compute a memcpy |
| 11 | block the Lagrange setup loop | one dependency chain the length of the domain, where each block can seed itself with `gˢᵗᵃʳᵗ` |
| 12 | collapse the opening's two accumulation levels | both levels are linear, so `Σⱼ dⱼ(Σᵢ cⱼᵢPᵢ) = Σᵢ(Σⱼ dⱼcⱼᵢ)Pᵢ` — one pass per term instead of two |

Two of these change behaviour in ways worth calling out rather than burying:

- **Horner (4) widens `evaluate`.** It never calls `degree()`, so it evaluates a
  polynomial carrying trailing zero coefficients where the powers form hit
  `degree()`'s `assert!` and panicked. `coeffs` is a public field and callers do
  construct such polynomials, so the tolerance is wanted — but it is a widening,
  and a test pins it.
- **The opening collapse (12) is only sound because the sponge is squeezed and
  never absorbed** between `open_combinations` and the KZG open, so every
  challenge is fixed before any polynomial arithmetic. Squeeze order is preserved
  exactly: groups in query-set order, combinations in label order within a group,
  and the trailing squeeze per group.

### Correctness

Every change computes the same field elements by a cheaper route, so no proof
changes and there is no consensus impact.

Each commit brings tests that compare the new path against the one it replaces —
over random inputs and over the boundary cases: empty, constant, degree zero,
exact multiples, coefficient vectors carrying trailing zeros, and domains from 2
to 256. The Varuna suite (25 prove-and-verify tests), the polycommit suite (19)
and the FFT suite all pass at every commit, as does
`cargo clippy --workspace --all-targets --all-features -- -D warnings`.

For the opening change in particular, prove-and-verify is the load-bearing test:
a reordered Fiat–Shamir squeeze does not produce a slower proof, it produces one
that fails verification.

### How it was measured

**The benchmark proves at the consensus version mainnet actually runs, and
asserts that it does.** This turned out to matter more than expected.
`execute_authorization` derives the Varuna version from the current block height,
and a benchmark on a freshly created chain sits at height 1 — `ConsensusVersion::V1`,
so **Varuna V1**, a prover the network stopped using at height 6,135,000, and one
that never calls `prover_prepare_third_round` at all. The benchmark therefore
advances to `ConsensusVersion::V19` and asserts it got there.

Reaching that version also means the genesis records cannot be spent: they are
Version 0, and from `ConsensusVersion::V8` onward the credits circuit expects
Version 1. Spending one anyway produces an assignment that does not satisfy the
constraints, which surfaces a long way from its cause — as a non-zero remainder
when Varuna divides by the vanishing polynomial. The benchmark mints a spendable
record with `transfer_public_to_private` instead.

**Hardware.** Intel Xeon w9-3475X (Sapphire Rapids-WS), 36 cores / 72 threads,
82.5 MiB L3, 125 GiB RAM, Ubuntu 24.04.4 (kernel 6.17), rustc 1.88.0, governor
`powersave` with turbo enabled. Built with the repository's own
`.cargo/config.toml`, which sets `-C target-cpu=native`.

**Protocol.** Criterion's `--save-baseline`/`--baseline` runs one arm to
completion and then the other, which on an interactive machine is exposed to
drift — run medians moved by up to 4% across a single session here. Instead both
arms' benchmark binaries were built once, checksummed to confirm they differ, and
then alternated block by block with the arm order flipping each block. Each block
is one `sample_size(10)` run per arm; the table reports ten blocks.

**A null was run at each thread count** — the same binary against itself — so the
noise floor is measured under the same conditions as the treatment rather than
assumed:

| run | mean | median | SD | *t* | blocks favouring |
|---|---:|---:|---:|---:|---:|
| null, 16 threads | −0.022% | +0.178% | 1.04% | −0.07 | 5/10 |
| **change, 16 threads** | **−8.064%** | **−8.042%** | 1.01% | **−25.3** | **10/10** |
| null, 4 threads | −0.171% | −0.745% | 3.10% | −0.17 | 6/10 |
| **change, 4 threads** | **−3.910%** | **−3.994%** | 0.90% | **−13.8** | **10/10** |

Both nulls land on zero, with blocks splitting 5/10 and 6/10 — what no effect
looks like. Both treatments are unanimous across ten blocks.

The 4-thread null is noticeably noisier than the treatment it calibrates
(SD 3.10% against 0.90%), which suggests it caught a disturbed period on the
machine. It is reported as measured rather than re-run, because the conclusion
does not depend on it: −3.91% at *t* = −13.8 sits more than a standard deviation
of the null's own spread away from zero, with every block agreeing.

**The benchmark is off by default.** It sits behind a `bench-prove` feature
because proving takes seconds: enabled, it would add tens of minutes to the
`benchmarks.yml` job, needs the universal SRS, and is variable enough to risk the
benchmark action's alert threshold — the same three reasons that workflow already
gives for skipping the `varuna` and `store`/`advance` benchmarks.

Two small instrumentation fixes ride along in the first commit, because without
them the profile cannot be read: `prover_prepare_third_round` was emitting
`AHP::Prover::ThirdRound`, the same label `prover_third_round` uses, so two
distinct phases were indistinguishable in the tree; and `open_combinations` had
no timer, so its cost landed in its caller's self time. The `profiler` feature is
also now forwarded from `snarkvm-algorithms` through `snarkvm-synthesizer` and
`snarkvm-ledger`, so the instrumentation that already exists can be reached from
a benchmark that drives the whole stack.

### On splitting this up

These are presented together because the headline number is a property of the set
and each commit on its own moves less than a single benchmark run can resolve.
Every commit is self-contained and independently revertible, so if you would
rather review them as separate pull requests — or want per-change measurements
for any that look suspect — say which and I will split it accordingly.

## Test Plan

Extensive unit test coverage with each individual change, in the commit that introduced the change.
Benchmarks show that we get a measurable win overall.

## Documentation

N/A

## Backwards compatibility

Backwards compatible, no consensus guard needed.

### DIFF
--- a/algorithms/src/fft/domain.rs
+++ b/algorithms/src/fft/domain.rs
@@ -272,16 +272,33 @@ impl<F: FftField> EvaluationDomain<F> {
             }
             u
         } else {
-            let mut l = (t_size - one) * self.size_inv;
-            let mut r = one;
+            let l0 = (t_size - one) * self.size_inv;
             let mut u = vec![F::zero(); size];
             let mut ls = vec![F::zero(); size];
-            for i in 0..size {
-                u[i] = tau - r;
-                ls[i] = l;
-                l *= &self.group_gen;
-                r *= &self.group_gen;
-            }
+
+            // Both sequences are powers of `group_gen`, so a block can seed itself with
+            // `group_gen^start` rather than walking there from zero. Written serially this
+            // is one dependency chain the length of the domain, which neither
+            // more cores nor wider registers can shorten.
+            //
+            // Same shape as `distribute_powers_and_mul_by_const` above.
+            #[cfg(not(feature = "serial"))]
+            let chunk = core::cmp::max(size / max_available_threads(), 1024);
+            #[cfg(feature = "serial")]
+            let chunk = core::cmp::max(size, 1);
+
+            cfg_chunks_mut!(u, chunk).zip(cfg_chunks_mut!(ls, chunk)).enumerate().for_each(
+                |(i, (u_chunk, ls_chunk))| {
+                    let mut r = self.group_gen.pow([(i * chunk) as u64]);
+                    let mut l = l0 * r;
+                    u_chunk.iter_mut().zip(ls_chunk.iter_mut()).for_each(|(tau_minus_r, l_i)| {
+                        *tau_minus_r = tau - r;
+                        *l_i = l;
+                        r *= &self.group_gen;
+                        l *= &self.group_gen;
+                    });
+                },
+            );
 
             batch_inversion(u.as_mut_slice());
             cfg_iter_mut!(u, 1_000).zip_eq(ls).for_each(|(tau_minus_r, l)| {
--- a/algorithms/src/fft/polynomial/dense.rs
+++ b/algorithms/src/fft/polynomial/dense.rs
@@ -102,6 +102,37 @@ impl<F: Field> DensePolynomial<F> {
         } else if point.is_zero() {
             return self.coeffs[0];
         }
+
+        // Horner's rule: `degree` multiplications and no allocation, where
+        // materialising the powers of the point needs two multiplications per
+        // coefficient -- once to build the power, once to weight the
+        // coefficient -- and a vector the size of the polynomial to hold them.
+        //
+        // It is fully serial, where the powers form parallelised the weighting.
+        // That is the trade, and it is worth taking: the deleted half of the
+        // arithmetic outweighs the parallelism given up, and the allocation it
+        // also deletes is proportional to the polynomial.
+        //
+        // Unlike the powers form this tolerates trailing zero coefficients,
+        // which reach it because `coeffs` is public and callers do build such
+        // polynomials. The powers form called `degree()` and panicked on that
+        // input; `evaluate_tolerates_trailing_zeros` pins the new behaviour.
+        let mut acc = F::zero();
+        for coeff in self.coeffs.iter().rev() {
+            acc = acc * point + *coeff;
+        }
+        acc
+    }
+
+    /// The powers form that `evaluate` replaced, kept so the equivalence test
+    /// has something to compare against.
+    #[cfg(test)]
+    fn evaluate_by_powers(&self, point: F) -> F {
+        if self.is_zero() {
+            return F::zero();
+        } else if point.is_zero() {
+            return self.coeffs[0];
+        }
         let mut powers_of_point = Vec::with_capacity(1 + self.degree());
         powers_of_point.push(F::one());
         let mut cur = point;
@@ -983,5 +1014,47 @@ mod monic_linear_divide_tests {
         let (q, r) = DensePolynomial::from_coefficients_vec(vec![c]).divide_by_monic_linear(z);
         assert!(q.is_zero());
         assert_eq!(r, c);
+    }
+}
+
+#[cfg(test)]
+mod evaluate_tests {
+    use super::vanishing_divide_tests::rand_poly;
+    use snarkvm_curves::bls12_377::Fr;
+    use snarkvm_fields::{One, Zero};
+    use snarkvm_utilities::rand::{TestRng, Uniform};
+
+    /// Horner agrees with the powers form it replaced, on well-formed input.
+    ///
+    /// Field addition is associative, so the tree reduction the powers form
+    /// used and Horner's serial accumulation agree exactly rather than
+    /// approximately -- `assert_eq!` is the right strength here, not an
+    /// epsilon.
+    #[test]
+    fn evaluate_agrees_with_the_powers_form() {
+        let rng = &mut TestRng::default();
+        for len in [0usize, 1, 2, 5, 16, 17, 33, 70] {
+            let p = rand_poly(len, rng);
+            for point in [Fr::zero(), Fr::one(), Fr::rand(rng), Fr::rand(rng)] {
+                assert_eq!(p.evaluate(point), p.evaluate_by_powers(point), "len={len}");
+            }
+        }
+    }
+
+    /// `evaluate` tolerates trailing zero coefficients. The powers form it
+    /// replaced panicked in `degree()` on exactly this input, so the tolerance
+    /// is a deliberate widening rather than an accident.
+    #[test]
+    fn evaluate_tolerates_trailing_zeros() {
+        let rng = &mut TestRng::default();
+        for len in [1usize, 5, 17] {
+            for pad in [1usize, 4, 20] {
+                let mut p = rand_poly(len, rng);
+                let point = Fr::rand(rng);
+                let expected = p.evaluate(point);
+                p.coeffs.resize(p.coeffs.len() + pad, Fr::zero());
+                assert_eq!(p.evaluate(point), expected, "len={len} pad={pad}");
+            }
+        }
     }
 }
--- a/algorithms/src/fft/polynomial/dense.rs
+++ b/algorithms/src/fft/polynomial/dense.rs
@@ -367,9 +367,15 @@ impl<'a, F: Field> AddAssign<(F, &'a DensePolynomial<F>)> for DensePolynomial<F>
     #[allow(clippy::suspicious_op_assign_impl)]
     fn add_assign(&mut self, (f, other): (F, &'a DensePolynomial<F>)) {
         if self.is_zero() {
+            // One pass, not two. The scale has to happen either way; the copy
+            // does not have to be a separate walk over the coefficients.
+            //
+            // `clear()` rather than assigning a fresh vector, because `self`
+            // being zero does not mean it is empty -- a polynomial whose
+            // coefficients are all zero reaches here too, and its allocation is
+            // worth keeping.
             self.coeffs.clear();
-            self.coeffs.extend_from_slice(&other.coeffs);
-            self.coeffs.iter_mut().for_each(|c| *c *= &f);
+            self.coeffs.extend(other.coeffs.iter().map(|c| *c * f));
         } else if other.is_zero() {
             // return
         } else if self.degree() >= other.degree() {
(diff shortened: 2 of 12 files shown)


algorithms/src/fft/polynomial/dense.rs
```diff
@@ -102,6 +118,37 @@ impl<F: Field> DensePolynomial<F> {
         } else if point.is_zero() {
             return self.coeffs[0];
         }
+
+        // Horner's rule: `degree` multiplications and no allocation, where
```
Antonio95, line 122: I would personally remove most, if not all, of this comment.

algorithms/src/fft/polynomial/dense.rs
```diff
@@ -158,15 +205,77 @@ impl<F: PrimeField> DensePolynomial<F> {
         DensePolynomial::from_coefficients_vec(shifted)
     }
 
-    /// Divide `self` by the vanishing polynomial for the domain `domain`.
-    /// Returns the quotient and remainder of the division.
+    /// Divides by the monic linear polynomial `X - z`, returning the quotient
```
Antonio95, line 208: I would replace the entire comment by:

/// Divides by the monic linear polynomial `X - z` using Ruffini's rule, returning the quotient and the remainder `self(z)`.

algorithms/src/fft/domain.rs
```diff
@@ -272,16 +272,33 @@ impl<F: FftField> EvaluationDomain<F> {
...
             let mut u = vec![F::zero(); size];
             let mut ls = vec![F::zero(); size];
-            for i in 0..size {
-                u[i] = tau - r;
-                ls[i] = l;
-                l *= &self.group_gen;
-                r *= &self.group_gen;
-            }
+
+            // Both sequences are powers of `group_gen`, so a block can seed itself with
```
Antonio95, line 279: I feel like this comment is not very useful, especially the second sentence. Instead, it would be illustrative to say what this now-less-obvious code is doing, which is a parallel version of the very simple computation `l_0, l_0 * gen...` and `tau - 1, tau - r, tau - r^2....` (right?).

algorithms/src/fft/domain.rs
```diff
@@ -272,16 +272,33 @@ impl<F: FftField> EvaluationDomain<F> {
...
+            //
+            // Same shape as `distribute_powers_and_mul_by_const` above.
+            #[cfg(not(feature = "serial"))]
+            let chunk = core::cmp::max(size / max_available_threads(), 1024);
+            #[cfg(feature = "serial")]
+            let chunk = core::cmp::max(size, 1);
+
+            cfg_chunks_mut!(u, chunk).zip(cfg_chunks_mut!(ls, chunk)).enumerate().for_each(
+                |(i, (u_chunk, ls_chunk))| {
+                    let mut r = self.group_gen.pow([(i * chunk) as u64]);
```
Antonio95, line 292: If we're being sharp, I think these values can be computed outside the loop for added efficiency. Suppose there are 10 chunks. I think this line is computing
 - `r = gen^0` in the first iteration
 - `r = gen^10` in the second iteration
 - `r = gen^20` in the third iteration (this is the square of the r from the second iteration)
 - `r = gen^30` in the fourth iteration (this is the square of the r from the second iteration)

So what about computing `gen^10` outside the loop and seeding `r` with `(gen^10)^i` inside it?

algorithms/src/snark/varuna/ahp/prover/round_functions/second.rs
```diff
@@ -111,7 +110,21 @@ impl<F: PrimeField, SM: SNARKMode> AHPForR1CS<F, SM> {
                     let mut rowcheck = multiplier_2.multiply().unwrap();
                     rowcheck.coeffs.iter_mut().zip(&z_c.coeffs).for_each(|(ab, c)| *ab -= c);
 
-                    instance_lhs += &(&rowcheck * instance_combiner);
+                    // `instance_lhs` is this and nothing else -- one job, one instance combiner,
```
Antonio95, line 113: This comment is unnecessarily verbose except for the last clarification on leading zeros. Perhaps simply:

Initialise `instance_lhs` as `instance_combiner * rowcheck`. The trim is necessary The trim is necessary as `apply_randomized_selector` below is degree-sensitive.

algorithms/src/snark/varuna/ahp/selectors.rs
```diff
@@ -133,15 +133,34 @@ pub(crate) fn apply_randomized_selector<F: PrimeField>(
...
+            // `xg * (v_H / v_H_i)` is a concatenation, not a multiplication.
+            //
+            // With `m = |H_i|` and `n = |H|`, both powers of two and `m` dividing `n`, the
+            // quotient `v_H / v_H_i` is `(X^n - 1) / (X^m - 1) = 1 + X^m + X^2m
+            // + ... + X^(n-m)`. Multiplying `xg` by that places a copy of `xg`
+            // at every multiple of `m`, and since `deg(xg) < m` the copies do
+            // not overlap and nothing has to be added. So the result is `xg`'s
+            // coefficients repeated `n/m` times.
+            //
+            // Written as a multiply followed by a divide it was two passes over an
```
Antonio95, line 145: This last paragraph describing how the code used to work should likely go.

algorithms/src/snark/varuna/ahp/selectors.rs
```diff
@@ -133,15 +133,34 @@ pub(crate) fn apply_randomized_selector<F: PrimeField>(
         let updated_xg_i = if src_domain.size == target_domain.size {
             xg_i
         } else {
-            let xg_i = xg_i.mul_by_vanishing_poly(*target_domain);
-
-            let (xg_i, remainder) = xg_i.divide_by_vanishing_poly(*src_domain)?;
+            // `xg * (v_H / v_H_i)` is a concatenation, not a multiplication.
```
Antonio95, line 136: The `xg` in this comment (incl. below) should be `xg_i` (this is consistent with the surrounding code but also clearer, since we often label individual-proof/batch objects with `_i` and batched ones without subindices).

I would remove this first line - we've already said what we're computing in the previous comment, and the next paragraph already explains why it can be computed as we now do.

algorithms/src/snark/varuna/ahp/selectors.rs
Antonio95, line 129: The old code and the new code seem to be computing `xg_i * (v_H / v_H_i)`, but I can't see where the scaling by `H_i.size() / H.size()` happens. Is the old comment wrong, or am I missing sth?

Antonio95 (changes requested):
I have reviewed all changes except for the new `LazyCombination`, which adds a new structs and so I want to spend a few more mins on it (at least it is prover-side only).

I have left a few comments, many of which relate to the new code comments and only a couple of which are functional. Overall, I'm happy with the changes, but let's discuss the things I brought up before moving forward.

algorithms/src/fft/polynomial/dense.rs
cbeck88: fixed: https://github.com/ProvableHQ/snarkVM/pull/3399/changes/dc3a78a9cc4d77e31a344f605e1348631217ccbe

algorithms/src/fft/domain.rs
cbeck88: fixed: https://github.com/ProvableHQ/snarkVM/pull/3399/changes/dc3a78a9cc4d77e31a344f605e1348631217ccbe

algorithms/src/snark/varuna/ahp/prover/round_functions/second.rs
cbeck88: fixed: https://github.com/ProvableHQ/snarkVM/pull/3399/changes/dc3a78a9cc4d77e31a344f605e1348631217ccbe

algorithms/src/snark/varuna/ahp/selectors.rs
cbeck88: fixed: https://github.com/ProvableHQ/snarkVM/pull/3399/changes/dc3a78a9cc4d77e31a344f605e1348631217ccbe

algorithms/src/fft/domain.rs
cbeck88: fix for the hoist is: https://github.com/ProvableHQ/snarkVM/pull/3399/changes/747eaba3678cacbf0ce9ed21a0f2c3a710818962

algorithms/src/snark/varuna/ahp/selectors.rs
cbeck88: `xg -> xg_i` is: https://github.com/ProvableHQ/snarkVM/pull/3399/changes/1f13369f9816865e5e8c57594e11450b5c06bc46

other comment fixes: https://github.com/ProvableHQ/snarkVM/pull/3399/changes/dc3a78a9cc4d77e31a344f605e1348631217ccbe

algorithms/src/snark/varuna/ahp/selectors.rs
cbeck88: i think the comment was a little stale

The scaling is folded into multiplier before the division:

```rust
let multiplier = combiner * src_domain.size_as_field_element * target_domain.size_inv;
```

`src_domain.size_as_field_element * target_domain.size_inv` is exactly
`|H_i| / |H|`, and because it is applied to poly before the division, both
`h_i` and `xg_i` come out already scaled.

That is pre-existing code and this branch does not change it.

Antonio95: Makes sense, thanks for finding the place where it was done 🙌  Resolving

PR #3395, opened 2026-08-27
Bump the Rust toolchain to 1.96 and clear the resulting lints

cbeck88:
## Motivation

Updates the MSRV in rust-toolchain.toml, Cargo.toml, and .circleci/config.yml.

The immediate motivation is that some AVX-512 intrinsics are only stable from Rust 1.89.

But more generally, Rust 1.88 is from 1 year ago, and we may be missing compiler bug fixes or performance improvements.

The newer compiler reports 287 clippy findings in existing code.

Most are cleared mechanically by `cargo clippy --fix`; the bulk is `mismatched_lifetime_syntaxes` (149), plus collapsible `if`, useless `vec!`, and manual impls of derivable traits. The rest needed a human:

  - `cloned_ref_to_slice_refs` (28 sites): `&[x.clone()]` becomes `slice::from_ref(&x)`. Three sites in the instruction-operation macros are *not* convertible and the lint had to be suppressed instead.
  - `bigint_256.rs` / `bigint_384.rs`: `_addcarry_u64` and `_subborrow_u64` are safe functions on this toolchain, so the `unsafe` blocks wrapping them are now redundant.
  - `finalize.rs`: `notify_staking_reward` collapses into a let-chain.
  - `non_canonical_partial_ord_impl` now fires on the impl block rather than the method, so the existing allow moves with it.

## Test Plan

clippy and fmt fix and the tests are passing locally

## Backwards compatibility

No behavioural change is intended; every edit is a lint fix.

---

Supersedes #3371, which was opened from a fork branch where CI did not trigger correctly. Same commits, rebased onto current `staging`.

### DIFF
--- a/synthesizer/src/vm/mod.rs
+++ b/synthesizer/src/vm/mod.rs
@@ -203,7 +203,7 @@ impl<N: Network, C: ConsensusStorage<N>> VM<N, C> {
             })
             .collect::<Result<Vec<_>>>()?;
         // Sort the deployment transaction IDs by their block heights.
-        deployment_ids.sort_unstable_by(|(_, a), (_, b)| a.cmp(b));
+        deployment_ids.sort_unstable_by_key(|(_, a)| *a);
 
         // Load the deployments in order of their block heights.
         const PARALLELIZATION_FACTOR: usize = 256;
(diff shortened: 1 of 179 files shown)


synthesizer/src/vm/mod.rs
```diff
@@ -203,7 +203,7 @@ impl<N: Network, C: ConsensusStorage<N>> VM<N, C> {
             })
             .collect::<Result<Vec<_>>>()?;
         // Sort the deployment transaction IDs by their block heights.
-        deployment_ids.sort_unstable_by(|(_, a), (_, b)| a.cmp(b));
+        deployment_ids.sort_unstable_by_key(|(_, a)| *a);
```
ljedrz:
```suggestion
        deployment_ids.sort_unstable_by_key(|(_, h)| *h);
```
nit

ljedrz (approved):
LGTM

Antonio95 (approved):
LGTM, the version bump will make rust analyzer's life a lot easier. The two CI errors are harmless.

vicsn (approved):
LGTM!

PR #3397, opened 2026-08-28
fix(curve): reject non canonical infinity encoding

cbeck88:
## Motivation

A compressed Short Weierstrass point sets a flag bit to mark the point at
infinity. When that flag is set the x-coordinate carries no information and must
be zero — but `deserialize_with_mode` ignores it entirely:

https://github.com/ProvableHQ/snarkVM/blob/55c15e2d35ea5d6fbcb8c79d26bb3ed111172a87/curves/src/templates/macros.rs#L127-L133

This means that there are many non-canonical representations of the identity point, which are accepted by the deserializer. The BLS12-381serialization spec requires of decoders, for exactly this reason: an infinity
encoding whose remaining bits are not zero must be rejected.

The same predicate is already applied correctly thirty lines away, in
`Affine::from_random_bytes`:

https://github.com/ProvableHQ/snarkVM/blob/55c15e2d35ea5d6fbcb8c79d26bb3ed111172a87/curves/src/templates/short_weierstrass_jacobian/affine.rs#L117-L120

```rust
// If x is valid and is zero and only the infinity flag is set, then parse this
// point as infinity. For all other choices, get the original point.
if x.is_zero() && flags.is_infinity() {
```

This change brings the deserializer into line with it.

## What this is not

It is **not** a soundness break. The mapping is many-to-one *onto the identity*:
every non-canonical encoding decodes to the same point, so a verifier sees
identical values whichever encoding arrived. It does not let anyone change what
a proof asserts, make an invalid proof verify, or produce a point of their
choosing. It is byte-level malleability and nothing more.

## Test Plan

In the generic `sw_tests<P>` harness, so it covers every Short Weierstrass curve
rather than BLS12-377 G1 alone. It asserts the desired behaviour, so it fails
without the fix:

```
test bls12_377::tests::test_g1_projective_curve ... FAILED
a non-canonical infinity encoding was accepted (Validate::Yes).
[1, 0, 0, 0]... and [0, 0, 0, 0]... are two encodings of one value.
```

- It checks the **canonical** infinity encoding still deserializes *first*. A fix
  that rejected both would break every proof in existence, and that should fail
  loudly rather than pass quietly.
- It runs under `Validate::No` as well as `Validate::Yes`, because the rejection
  must come from the structure of the encoding rather than the subgroup check.
- It mutates byte 0, not the last byte — the flag lives in the high bits of the
  final byte, so touching byte 0 changes x and leaves the flag set.

## Backwards compatibility

This is backwards compatible in that the serializer only emits canonical points, so the
blockchain itself should only consist of canonical encodings. The transactions that
we have consensed on would contain only canonical encodings, since they go through
at least one serialization deserialization round trip when they are gossipped to other nodes.
So this will not break our ability to parse historical blocks and transactions.

### DIFF
--- a/curves/src/templates/macros.rs
+++ b/curves/src/templates/macros.rs
@@ -126,6 +126,11 @@ macro_rules! impl_sw_curve_serializer {
                 let point = if let Compress::Yes = compress {
                     let (x, flags) = P::BaseField::deserialize_with_flags::<_, SWFlags>(&mut reader)?;
                     if flags.is_infinity() {
+                        // The x-coordinate must be zero when the infinity flag is
+                        // set, or the encoding is non-canonical.
+                        if !x.is_zero() {
+                            return Err(snarkvm_utilities::serialize::SerializationError::InvalidData);
+                        }
                         Self::zero()
                     } else {
                         Affine::<P>::from_x_coordinate(x, flags.is_positive().unwrap())
--- a/curves/src/templates/short_weierstrass_jacobian/tests.rs
+++ b/curves/src/templates/short_weierstrass_jacobian/tests.rs
@@ -17,7 +17,7 @@ use std::io::Cursor;
 
 use super::{Affine, Projective};
 use crate::{AffineCurve, ProjectiveCurve, ShortWeierstrassParameters};
-use snarkvm_fields::Zero;
+use snarkvm_fields::{One, Zero};
 use snarkvm_utilities::{
     Compress,
     TestRng,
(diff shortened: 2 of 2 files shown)


vicsn (approved):
LGTM, pending regression testing to get extra assurance this doesn't break syncing.

IMO we don't have to sync the entire chain E2E, just the last 100.000 blocks would already be good evidence - and we don't expect new entrants to sync from genesis anyway but to use snapshots.

PR #3398, opened 2026-09-02
perf: avoid duplicate deserialization allocation

### LINKED ISSUE #3389: [Perf] Deserialization incurs duplicate allocation
## 🚀 Feature

https://github.com/ProvableHQ/snarkVM/blob/eff84712bee8f958d357a4316b98a61981513e13/utilities/src/bytes.rs#L128-L129

Perhaps there is a way to avoid the duplication, see:
```
/// Deserializes a dynamically-sized byte array.
    pub fn deserialize_with_size_encoding<D: Deserializer<'de>>(deserializer: D, name: &str) -> Result<T, D::Error> {
        let mut buffer = Vec::with_capacity(32);
        deserializer.deserialize_bytes(FromBytesVisitor::new(&mut buffer, name))?;
        FromBytes::read_le(&*buffer).map_err(de::Error::custom)
    }
```

zakazaka95:
## Summary
- deserialize borrowed and temporary byte slices directly into `FromBytes`
- consume an owned byte buffer without copying it into a second allocation
- retain the preallocated sequence fallback and checked/unchecked behavior

Closes #3389

## Performance
A release harness compiling the production `bytes.rs` deserialized a 4 KiB value 100,000 times:

- before: 200,000 allocations, 40.02 ms
- after: 0 allocations, 11.56 ms

## Testing
- production-file round trips through bincode bytes and JSON sequences
- checked and `read_le_unchecked` paths
- standalone harness Clippy with `-D warnings`
- `cargo +nightly-2026-04-10 fmt --all -- --check`
- `git diff --check`

Full workspace checks could not resolve the locked dependency graph on this Windows host because Cargo's Schannel backend could not fetch the uncached `aleo-std`; upstream CI remains the full-crate gate.

### DIFF
--- a/utilities/src/bytes.rs
+++ b/utilities/src/bytes.rs
@@ -171,9 +171,7 @@ impl<'de, T: FromBytes> FromBytesDeserializer<T> {
 
     /// Deserializes a dynamically-sized byte array.
     pub fn deserialize_with_size_encoding<D: Deserializer<'de>>(deserializer: D, name: &str) -> Result<T, D::Error> {
-        let mut buffer = Vec::with_capacity(32);
-        deserializer.deserialize_bytes(FromBytesVisitor::new(&mut buffer, name))?;
-        FromBytes::read_le(&*buffer).map_err(de::Error::custom)
+        deserializer.deserialize_bytes(FromBytesValueVisitor::new(name, false))
     }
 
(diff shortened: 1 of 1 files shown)


cbeck88 (approved):
looks good to me

vicsn (approved):

PR #3399, opened 2026-09-02
Speed up CPU proving: -8% on a private transfer

cbeck88:
## Motivation

An agent noticed that we don't use Ruffini's rule for division when doing CPU proving and use a naive algorithm instead.

This kicked off a swarm of agents looking for similar algorithmic improvements to improve the speed of CPU proving, and they found 12 improvements. I reviewed each of these changes and they seem simple and correct to me, and the measurement process we converged on seems very sound to me.

### Speed up CPU proving: −8% on a private transfer

Twelve changes to the prover, each of which deletes arithmetic or allocation
without changing a single field element that comes out. Measured end to end on
`credits.aleo/transfer_private`:

| threads | before | after | change |
|---:|---:|---:|---:|
| **16** | 5.863 s | 5.390 s | **−8.06%** (−474 ms) |
| **4** | 12.545 s | 12.055 s | **−3.91%** (−490 ms) |

The first commit adds the benchmark these numbers come from, so the claim is
reproducible from this branch:

```
cargo bench -p snarkvm-ledger --bench transaction --features bench-prove \
    -- 'Execute\(transfer_private\)'
```

### Why

Proving is dominated by multi-scalar multiplication, which is already well
optimized, so there is no single large win available here. The work *around* the
MSM is a different matter: several routines on the proving path do more
arithmetic, or more allocation, than the operation they implement requires.
Individually most are small. Together they are the numbers above.

Two things this helps:

1. **Anyone proving locally.** Proof latency is the whole experience for a user
   generating their own transactions.
2. **This repository's own tests and CI.** Proving dominates the synthesizer and
   ledger suites, so anything that shortens a proof shortens all of them.

The saving is *larger* on more cores — −8.1% at 16 threads against −3.9% at 4 —
and the absolute saving is nearly identical at both (474 ms and 490 ms). That is
the signature of deleted work rather than improved parallelism: most of these
changes remove **serial** work, so by Amdahl's law they matter more as core count
rises, and the benefit grows on larger machines instead of shrinking.

### The changes

Each is a single commit, and each stands on an argument you can check by reading
it, independent of the measurement.

| # | commit | what it deletes |
|---|---|---|
| 1 | add the proving benchmark | *(no behaviour change)* |
| 2 | divide by the vanishing polynomial in closed form | `p = q(Xⁿ−1)+r` matched coefficient-wise is `n` independent carry chains; for `deg p < 2n` — every call the prover makes — the quotient is literally `p[n..]`. Generic long division rediscovers this one coefficient at a time |
| 3 | Ruffini's rule for `X − z` | one multiply-add per coefficient instead of two multiplications, a clone of the dividend, and a degree re-check every iteration. The remainder falls out as `self(z)` |
| 4 | Horner's rule for `evaluate` | `deg` multiplications instead of `2·deg`, and no vector of powers the size of the polynomial |
| 5 | scale the rowcheck in place | `lhs += &(&rowcheck * c)` onto a zero polynomial is three passes and two allocations; the accumulation it is named for only ever has one term |
| 6 | fuse `add_assign`'s zero branch | copy-then-scale walks the coefficients twice to produce what one walk produces |
| 7 | `has_lc_eval` | the "is this polynomial present?" probe was answered by evaluating it and discarding the value, twice per matrix |
| 8 | hoist the Lagrange coefficients | they depend only on the constraint domain and `alpha`, both fixed across the `a`/`b`/`c` jobs, and were built three times per circuit |
| 9 | borrow the FFT precomputations | root-of-unity tables are circuit data, fixed for the life of the process, and were deep-copied per circuit per proof |
| 10 | selector `xg · (v_H / v_H_i)` as a repeat | the quotient of the two vanishing polynomials is `1 + Xᵐ + X²ᵐ + …`, so the product is `xg` repeated. It was a multiply and a divide over the larger domain to compute a memcpy |
| 11 | block the Lagrange setup loop | one dependency chain the length of the domain, where each block can seed itself with `gˢᵗᵃʳᵗ` |
| 12 | collapse the opening's two accumulation levels | both levels are linear, so `Σⱼ dⱼ(Σᵢ cⱼᵢPᵢ) = Σᵢ(Σⱼ dⱼcⱼᵢ)Pᵢ` — one pass per term instead of two |

Two of these change behaviour in ways worth calling out rather than burying:

- **Horner (4) widens `evaluate`.** It never calls `degree()`, so it evaluates a
  polynomial carrying trailing zero coefficients where the powers form hit
  `degree()`'s `assert!` and panicked. `coeffs` is a public field and callers do
  construct such polynomials, so the tolerance is wanted — but it is a widening,
  and a test pins it.
- **The opening collapse (12) is only sound because the sponge is squeezed and
  never absorbed** between `open_combinations` and the KZG open, so every
  challenge is fixed before any polynomial arithmetic. Squeeze order is preserved
  exactly: groups in query-set order, combinations in label order within a group,
  and the trailing squeeze per group.

### Correctness

Every change computes the same field elements by a cheaper route, so no proof
changes and there is no consensus impact.

Each commit brings tests that compare the new path against the one it replaces —
over random inputs and over the boundary cases: empty, constant, degree zero,
exact multiples, coefficient vectors carrying trailing zeros, and domains from 2
to 256. The Varuna suite (25 prove-and-verify tests), the polycommit suite (19)
and the FFT suite all pass at every commit, as does
`cargo clippy --workspace --all-targets --all-features -- -D warnings`.

For the opening change in particular, prove-and-verify is the load-bearing test:
a reordered Fiat–Shamir squeeze does not produce a slower proof, it produces one
that fails verification.

### How it was measured

**The benchmark proves at the consensus version mainnet actually runs, and
asserts that it does.** This turned out to matter more than expected.
`execute_authorization` derives the Varuna version from the current block height,
and a benchmark on a freshly created chain sits at height 1 — `ConsensusVersion::V1`,
so **Varuna V1**, a prover the network stopped using at height 6,135,000, and one
that never calls `prover_prepare_third_round` at all. The benchmark therefore
advances to `ConsensusVersion::V19` and asserts it got there.

Reaching that version also means the genesis records cannot be spent: they are
Version 0, and from `ConsensusVersion::V8` onward the credits circuit expects
Version 1. Spending one anyway produces an assignment that does not satisfy the
constraints, which surfaces a long way from its cause — as a non-zero remainder
when Varuna divides by the vanishing polynomial. The benchmark mints a spendable
record with `transfer_public_to_private` instead.

**Hardware.** Intel Xeon w9-3475X (Sapphire Rapids-WS), 36 cores / 72 threads,
82.5 MiB L3, 125 GiB RAM, Ubuntu 24.04.4 (kernel 6.17), rustc 1.88.0, governor
`powersave` with turbo enabled. Built with the repository's own
`.cargo/config.toml`, which sets `-C target-cpu=native`.

**Protocol.** Criterion's `--save-baseline`/`--baseline` runs one arm to
completion and then the other, which on an interactive machine is exposed to
drift — run medians moved by up to 4% across a single session here. Instead both
arms' benchmark binaries were built once, checksummed to confirm they differ, and
then alternated block by block with the arm order flipping each block. Each block
is one `sample_size(10)` run per arm; the table reports ten blocks.

**A null was run at each thread count** — the same binary against itself — so the
noise floor is measured under the same conditions as the treatment rather than
assumed:

| run | mean | median | SD | *t* | blocks favouring |
|---|---:|---:|---:|---:|---:|
| null, 16 threads | −0.022% | +0.178% | 1.04% | −0.07 | 5/10 |
| **change, 16 threads** | **−8.064%** | **−8.042%** | 1.01% | **−25.3** | **10/10** |
| null, 4 threads | −0.171% | −0.745% | 3.10% | −0.17 | 6/10 |
| **change, 4 threads** | **−3.910%** | **−3.994%** | 0.90% | **−13.8** | **10/10** |

Both nulls land on zero, with blocks splitting 5/10 and 6/10 — what no effect
looks like. Both treatments are unanimous across ten blocks.

The 4-thread null is noticeably noisier than the treatment it calibrates
(SD 3.10% against 0.90%), which suggests it caught a disturbed period on the
machine. It is reported as measured rather than re-run, because the conclusion
does not depend on it: −3.91% at *t* = −13.8 sits more than a standard deviation
of the null's own spread away from zero, with every block agreeing.

**The benchmark is off by default.** It sits behind a `bench-prove` feature
because proving takes seconds: enabled, it would add tens of minutes to the
`benchmarks.yml` job, needs the universal SRS, and is variable enough to risk the
benchmark action's alert threshold — the same three reasons that workflow already
gives for skipping the `varuna` and `store`/`advance` benchmarks.

Two small instrumentation fixes ride along in the first commit, because without
them the profile cannot be read: `prover_prepare_third_round` was emitting
`AHP::Prover::ThirdRound`, the same label `prover_third_round` uses, so two
distinct phases were indistinguishable in the tree; and `open_combinations` had
no timer, so its cost landed in its caller's self time. The `profiler` feature is
also now forwarded from `snarkvm-algorithms` through `snarkvm-synthesizer` and
`snarkvm-ledger`, so the instrumentation that already exists can be reached from
a benchmark that drives the whole stack.

### On splitting this up

These are presented together because the headline number is a property of the set
and each commit on its own moves less than a single benchmark run can resolve.
Every commit is self-contained and independently revertible, so if you would
rather review them as separate pull requests — or want per-change measurements
for any that look suspect — say which and I will split it accordingly.

## Test Plan

Extensive unit test coverage with each individual change, in the commit that introduced the change.
Benchmarks show that we get a measurable win overall.

## Documentation

N/A

## Backwards compatibility

Backwards compatible, no consensus guard needed.

---

## This supersedes #3394

Same branch, opened from this repository rather than a fork so the merge flow
can run, as @Antonio95 asked. The review on #3394 is unchanged and still worth
reading; three commits on top of what was reviewed there address it:

| commit | review thread |
|---|---|
| `review: trim the new code comments` | the comments on `evaluate`, `divide_by_monic_linear`, the Lagrange setup, the rowcheck, and the selector |
| `review: name the selector's remainder witness xg_i in its comment` | `xg` -> `xg_i`, matching the surrounding code |
| `perf: hoist the per-chunk seed exponentiation out of the Lagrange loop` | seeding a chunk with `(g^chunk)^i` instead of `g^(i*chunk)` |

Only the third touches code, and only to reach the same field element by a
shorter ladder. Everything else is comment text.

One thread has no commit against it, because the answer is already in the file.
@Antonio95 asked where the `H_i.size() / H.size()` scaling happens in
`apply_randomized_selector`, since neither the old code nor the new one appears
to do it. It is folded into `multiplier` before the division:

```rust
let multiplier = combiner * src_domain.size_as_field_element * target_domain.size_inv;
```

`src_domain.size_as_field_element * target_domain.size_inv` is exactly
`|H_i| / |H|`, and because it is applied to `poly` before the division, both
`h_i` and `xg_i` come out already scaled. The comment above `updated_xg_i` says
so -- "without the last constant, which was already incorporated into the
multiplier". That is pre-existing code and this branch does not change it.

Also worth flagging on the hoist: `distribute_powers_and_mul_by_const` directly
above it still uses the wide `g.pow([(i * chunk) as u64])` form. The two loops
are deliberately the same shape, so they now differ in this one detail. Happy
to give it the same treatment, or to drop the hoist and keep them matching --
the saving either way is tens of multiplications against a chunk of at least a
thousand.

https://claude.ai/code/session_01L9mqNeg6kEF6qE5rAmz487

### DIFF
--- a/algorithms/src/fft/domain.rs
+++ b/algorithms/src/fft/domain.rs
@@ -276,10 +276,10 @@ impl<F: FftField> EvaluationDomain<F> {
             let mut u = vec![F::zero(); size];
             let mut ls = vec![F::zero(); size];
 
-            // Both sequences are powers of `group_gen`, so a block can seed itself with
-            // `group_gen^start` rather than walking there from zero. Written serially this
-            // is one dependency chain the length of the domain, which neither
-            // more cores nor wider registers can shorten.
+            // Fill `u[i] = tau - g^i` and `ls[i] = l0 * g^i`, where `g = group_gen`. Both
+            // sequences are powers of `g`, so a block can seed itself with `g^start`
+            // instead of walking there from zero, which is what lets the fill
+            // run in parallel.
             //
             // Same shape as `distribute_powers_and_mul_by_const` above.
             #[cfg(not(feature = "serial"))]
--- a/algorithms/src/fft/domain.rs
+++ b/algorithms/src/fft/domain.rs
@@ -272,16 +272,33 @@ impl<F: FftField> EvaluationDomain<F> {
             }
             u
         } else {
-            let mut l = (t_size - one) * self.size_inv;
-            let mut r = one;
+            let l0 = (t_size - one) * self.size_inv;
             let mut u = vec![F::zero(); size];
             let mut ls = vec![F::zero(); size];
-            for i in 0..size {
-                u[i] = tau - r;
-                ls[i] = l;
-                l *= &self.group_gen;
-                r *= &self.group_gen;
-            }
+
+            // Both sequences are powers of `group_gen`, so a block can seed itself with
+            // `group_gen^start` rather than walking there from zero. Written serially this
+            // is one dependency chain the length of the domain, which neither
+            // more cores nor wider registers can shorten.
+            //
+            // Same shape as `distribute_powers_and_mul_by_const` above.
+            #[cfg(not(feature = "serial"))]
+            let chunk = core::cmp::max(size / max_available_threads(), 1024);
+            #[cfg(feature = "serial")]
+            let chunk = core::cmp::max(size, 1);
+
+            cfg_chunks_mut!(u, chunk).zip(cfg_chunks_mut!(ls, chunk)).enumerate().for_each(
+                |(i, (u_chunk, ls_chunk))| {
+                    let mut r = self.group_gen.pow([(i * chunk) as u64]);
+                    let mut l = l0 * r;
+                    u_chunk.iter_mut().zip(ls_chunk.iter_mut()).for_each(|(tau_minus_r, l_i)| {
+                        *tau_minus_r = tau - r;
+                        *l_i = l;
+                        r *= &self.group_gen;
+                        l *= &self.group_gen;
+                    });
+                },
+            );
 
             batch_inversion(u.as_mut_slice());
             cfg_iter_mut!(u, 1_000).zip_eq(ls).for_each(|(tau_minus_r, l)| {
--- a/algorithms/src/fft/domain.rs
+++ b/algorithms/src/fft/domain.rs
@@ -287,9 +287,13 @@ impl<F: FftField> EvaluationDomain<F> {
             #[cfg(feature = "serial")]
             let chunk = core::cmp::max(size, 1);
 
+            // The chunks run unordered, so each seeds itself rather than carrying `r` from
+            // its predecessor. `chunk_gen^i` is the same seed as `g^(i*chunk)` from a
+            // ladder over `log2(chunks)` bits instead of 64.
+            let chunk_gen = self.group_gen.pow([chunk as u64]);
             cfg_chunks_mut!(u, chunk).zip(cfg_chunks_mut!(ls, chunk)).enumerate().for_each(
                 |(i, (u_chunk, ls_chunk))| {
-                    let mut r = self.group_gen.pow([(i * chunk) as u64]);
+                    let mut r = chunk_gen.pow([i as u64]);
                     let mut l = l0 * r;
                     u_chunk.iter_mut().zip(ls_chunk.iter_mut()).for_each(|(tau_minus_r, l_i)| {
                         *tau_minus_r = tau - r;
--- a/algorithms/src/fft/polynomial/dense.rs
+++ b/algorithms/src/fft/polynomial/dense.rs
@@ -102,6 +102,37 @@ impl<F: Field> DensePolynomial<F> {
         } else if point.is_zero() {
             return self.coeffs[0];
         }
+
+        // Horner's rule: `degree` multiplications and no allocation, where
+        // materialising the powers of the point needs two multiplications per
+        // coefficient -- once to build the power, once to weight the
+        // coefficient -- and a vector the size of the polynomial to hold them.
+        //
+        // It is fully serial, where the powers form parallelised the weighting.
+        // That is the trade, and it is worth taking: the deleted half of the
+        // arithmetic outweighs the parallelism given up, and the allocation it
+        // also deletes is proportional to the polynomial.
+        //
+        // Unlike the powers form this tolerates trailing zero coefficients,
+        // which reach it because `coeffs` is public and callers do build such
+        // polynomials. The powers form called `degree()` and panicked on that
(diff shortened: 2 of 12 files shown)


Antonio95 (approved):
Approved as all the comments from my review in the original PR https://github.com/ProvableHQ/snarkVM/pull/3394 have been addressed. Thank you!

Issue #3229, opened 2026-04-23
[Bug] LLVM miscompilation on aarch64-linux-gnu breaks Varuna proof generation and verification


davencyw:
## 🐛 Bug Report

LLVM miscompiles `EvaluationDomain::evaluate_all_lagrange_coefficients` on `aarch64-unknown-linux-gnu` when building with `opt-level=3 + lto="thin" + incremental=true`. The Lagrange coefficients returned by the function violate the partition-of-unity property (they no longer sum to 1), which silently corrupts Varuna proof generation and verification — proofs generated in release mode fail to verify.

The miscompilation is a **thin LTO cross-crate optimization bug**. Step-by-step isolation shows that every individual operation inside `evaluate_all_lagrange_coefficients` produces correct results when called separately. The loop that builds the input vectors, `batch_inversion` , and the final element-wise multiply all work fine in isolation, even in release mode. The bug only manifests when LLVM optimizes the *composed* function within the `snarkvm-algorithms` crate during the thin LTO link step. The same logic compiled inline in a separate crate produces correct results.

**All four conditions are required to trigger the bug:**

| Condition | Value |
|-----------|-------|
| Target | `aarch64-unknown-linux-gnu` |
| `opt-level` | `3` |
| `lto` | `"thin"` |
| `incremental` | `true` |

It does **not** reproduce on `aarch64-apple-darwin` (macOS Apple Silicon).
Switching `incremental` to `false` resolves this bug.

**Affected code:**
- `algorithms/src/fft/domain.rs`: `EvaluationDomain::evaluate_all_lagrange_coefficients`

**Affected profiles:** The workspace `Cargo.toml` sets `opt-level = 3`, `lto = "thin"`, `incremental = true` for `[profile.release]`, `[profile.bench]`, and `[profile.test]` all three are affected on aarch64-linux-gnu.

## Steps to Reproduce

1. Check out the [`fix/llvm-miscompilation-repro`](https://github.com/ProvableHQ/snarkVM/tree/fix/llvm-miscompilation-repro) branch on an `aarch64-unknown-linux-gnu` machine (e.g. GCP c4a).
2. Run the minimal reproducer in release mode:
   ```bash
   cd repro
   cargo run --release
   ```
   Output: `Lagrange coefficients (domain 2^8)... FAIL (sum != 1)` (fails for all domain sizes from 2^8 to 2^17)
3. Confirm it passes in debug mode:
   ```bash
   cargo run
   ```
[...]

PR #3402, opened 2026-09-03
build: disable incremental compilation for the release and bench profiles

cbeck88:
Sets `incremental = false` for `[profile.release]` and `[profile.bench]`. No source changes.

Refs #3229.

## Motivation

Cargo's default for the release profile is `incremental = false`, and the rustc book states that incremental compilation "inhibits certain optimizations (for example by increasing the amount of codegen units) and is therefore not recommended for release builds". (https://doc.rust-lang.org/rustc/codegen-options/index.html#incremental)

 `-C incremental` raises the default codegen-units from 16 to 256 per crate, so each crate is split into many more, much smaller LLVM modules, and `lto = "thin"` is then left to recover the lost inlining across them at link time. That is both a weaker optimization pipeline and a far less exercised one: it is the configuration implicated in #3229, where thin LTO miscompiles `evaluate_all_lagrange_coefficients` on aarch64-unknown-linux-gnu. (https://doc.rust-lang.org/rustc/codegen-options/index.html#codegen-units)

The rebuild-time cost is small, since incremental caching only covers the pre-LTO codegen phase while the thin LTO link is redone in full on every build, and is the majority of the build time in such a build.

## History

This combination (lto=thin, incremantal=true) first appeared in 2020 very early without explanation: 79ff75d67
It was later mirrored onto the dev profile. 37625d014

This all sat untouched for 6 years until we fixed the test profile to not use lto: 9a8301341

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -661,14 +661,14 @@ version = "2"
 [profile.release]
 opt-level = 3
 lto = "thin"
-incremental = true
+incremental = false
 
 [profile.bench]
 opt-level = 3
 debug = false
 rpath = false
 lto = "thin"
-incremental = true
+incremental = false
 debug-assertions = false
 
 [profile.dev]


ljedrz (approved):
Yeah I think that's fair; it might make compilation for dev purposes (where we mostly use `--release` anyway) slower, but we should prioritize the performance of built binaries.

PR #3405, opened 2026-09-04
test(curves): assert every serialized curve point deserializes back

cbeck88:
## Motivation

#3397 tightened the compressed Short Weierstrass deserializer so that a set
infinity flag with a non-zero x-coordinate is rejected. While sweeping for the
same problem elsewhere, we came close to merging the analogous tightening of the
*uncompressed* deserializer — and that one would have broken the library's
ability to read back what it had just written.

The uncompressed serializer emits the identity as `x = 0`, `y = 1` with the
infinity flag set:

https://github.com/ProvableHQ/snarkVM/blob/70ac1ff8b/curves/src/templates/macros.rs#L86-L95

So a check requiring both coordinates to be zero when the flag is set rejects
our own output, for the single most common point there is.

Nothing in the test suite would have caught that. There was no assertion
anywhere that the identity survives a serialization round trip, and no
serialization coverage at all for the projective types, which carry their own
`CanonicalSerialize` and `CanonicalDeserialize` impls.

## What this adds

`serialization_round_trip_test` in `curves/src/traits/tests_group.rs`, generic
over `AffineCurve` so one harness covers BLS12-377 G1, BLS12-377 G2 and
Edwards BLS12 rather than a single curve. For each point it asserts:

- serialize followed by deserialize returns an equal value, across all four
  `Compress` × `Validate` combinations;
- the property holds in both affine and projective form;
- `serialized_size` agrees with the number of bytes actually written.

The random half is a `proptest` over a 32-byte scalar seed, run through a
`TestRunner` so it stays generic over the curve.

**The half that does the real work is the explicit degenerate table.** Every bug
of this shape lives on a point that a random scalar reaches with probability
around 2^-250 — the identity above is the clearest case. Random sampling alone
would not have caught the patch we nearly merged, and would not catch the next
one either. The table covers the identity, the small multiples, `r - 1`, and a
projective point that has not been normalized.

## Test Plan

Reapplying the strict-uncompressed patch described above turns G1 and G2 red,
naming the point and the mode:

```
test bls12_377::tests::test_g1_serialization_round_trip ... FAILED
test bls12_377::tests::test_g2_serialization_round_trip ... FAILED

identity (affine): the deserializer rejected this library's own output
  (Compress::No, Validate::Yes): InvalidData
```

Removing the mutation returns the suite to green: 65 passed in
`snarkvm-curves`, with `cargo clippy -p snarkvm-curves --all-targets --
-D warnings` and `cargo +nightly fmt --check` clean.

## Scope

The harness is deliberately restricted to prime-order-subgroup points, which is
what `Validate::Yes` admits. On-curve points *outside* the subgroup — the
order-2 and order-4 points, whose coordinates include a zero — do not all round
trip today, so widening the harness to cover them fails on the current tree. A
doc comment on the entry point records this so the next person does not walk
into it. Those cases, and the canonicality gaps they expose, are a separate
follow-up.

## Backwards compatibility

Test-only. No runtime code changes, so no consensus impact. The single
non-test change is a `proptest` dev-dependency on `snarkvm-curves`, taken from
the existing workspace pin (`=1.6`, already in `Cargo.lock` via
`ledger/committee`); `Cargo.lock` gains one line and no new transitive crates.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -3596,6 +3596,7 @@ version = "4.9.0"
 dependencies = [
  "bincode",
  "criterion",
+ "proptest",
  "rand 0.10.2",
  "rustc_version",
  "serde",
--- a/curves/Cargo.toml
+++ b/curves/Cargo.toml
@@ -57,6 +57,9 @@ workspace = true
 [dev-dependencies.criterion]
 workspace = true
 
+[dev-dependencies.proptest]
+workspace = true
+
 [dev-dependencies.rand]
 workspace = true
 
--- a/curves/src/bls12_377/tests.rs
+++ b/curves/src/bls12_377/tests.rs
@@ -647,6 +647,11 @@ fn test_g1_projective_group() {
 }
 
 #[test]
+fn test_g1_serialization_round_trip() {
+    serialization_round_trip_test::<G1Affine>();
+}
+
+#[test]
 fn test_g1_generator() {
     let generator = G1Affine::prime_subgroup_generator();
     assert!(generator.is_on_curve());
@@ -668,6 +673,11 @@ fn test_g2_projective_group() {
     let a: G2Projective = rng.random();
     let b: G2Projective = rng.random();
     projective_test(a, b, &mut rng);
+}
+
+#[test]
+fn test_g2_serialization_round_trip() {
+    serialization_round_trip_test::<G2Affine>();
 }
 
 #[test]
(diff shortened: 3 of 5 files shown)


ljedrz (approved):
LGTM structure-wise, it would be best for someone more familiar with curve points to weigh in on the scope of the test too.

Antonio95 (approved):

PR #3408, opened 2026-09-04
Record a storage schema version, and repair the history encoding

cbeck88:
Draft, for 4.10.1. **Supersedes #3404, and is independent of it** — this branch is cut from `testnet` and contains none of that one's commits, so it merges on its own. It shares ideas with #3404 (byte-reversal collision handling, whole-group chunking, read-all-before-write-any batches) but the driver and the resume mechanism differ, and #3404's measurements, adversarial review and defect repros are why this one is shaped the way it is.

## Why this exists

A partner running an archive node with `--features history` upgraded to v4.9.0 and hit a write-performance regression — roughly **500× write amplification** (3.6–3.9 GB/s against ~7 MB/s healthy) and a compaction sawtooth that never drained. They diagnosed it themselves, reverted #3323, and have been running that patched build since.

**We do not know how long they ran the unpatched build, nor what state their database is in now.** They reverted quickly — the report says the signature reproduced within ~60 seconds on a fully-synced node — but "quickly" is not a number, and every entry written during that window is in a different on-disk encoding from everything around it. Nor do we know whether they ran `--history` before v4.7.5, which determines whether their keys carry a heights row and are therefore immune to the mixing described below.

That uncertainty is the reason for this work. The point is not only to repair a known-bad database, but to be able to **tell** whether a given database is repairable at all, and to say so before touching it.

## The two changes that caused it

**[#3312](https://github.com/ProvableHQ/snarkVM/pull/3312) — "Fix history indexing performance"** ([`a108e815`](a108e8159c8b26a19fbe25bd51a5de9a645796a3)) removed all three sites that wrote `MappingUpdateHeightsMap`, replacing the index with a prefix scan. Self-consistent at the time: every entry was little-endian and the scan decoded little-endian. Writes stopped rewriting an ever-growing vector, so this was a genuine improvement. Shipped in snarkOS v4.7.5 and v4.8.0.

**[#3323](https://github.com/ProvableHQ/snarkVM/pull/3323) — "Speed up history reads"** ([`58a1359e`](58a1359ea0d0acfa177b5ac00cd9854ae393ab93)) introduced big-endian height keys, using **the presence of a heights row as the sentinel for "this key is legacy"**:

- write path, [`finalize.rs#L371`](https://github.com/ProvableHQ/snarkVM/blob/58a1359ea0d0acfa177b5ac00cd9854ae393ab93/ledger/store/src/program/finalize.rs#L371) and [`#L453`](https://github.com/ProvableHQ/snarkVM/blob/58a1359ea0d0acfa177b5ac00cd9854ae393ab93/ledger/store/src/program/finalize.rs#L453)
- read path, [`finalize.rs#L875`](https://github.com/ProvableHQ/snarkVM/blob/58a1359ea0d0acfa177b5ac00cd9854ae393ab93/ledger/store/src/program/finalize.rs#L875)

Shipped in snarkOS v4.8.1, v4.9.0 and v4.9.1.

That sentinel was **already false for three weeks of production data**. Keys written by v4.7.5/v4.8.0 have no heights row and are little-endian, so #3323's `else` branch treats them as big-endian. Two consequences:

1. **Reads return values from unrelated heights** — silently, with a 200.
2. **The write amplification returns.** For a key that still carries a row from ≤v4.7.4, `a108e815` had frozen that row and made writes O(1). #3323 sees the frozen row, concludes "legacy", and resumes the read-modify-write. That is what the partner measured.

## Seeing it fail

#3404 carries a committed test that demonstrates the corruption directly — a read that is **correct before a migration and wrong after it**, because an unlisted little-endian entry is byte-identical to a migrated big-endian one:

**[`test_migration_exposes_unlisted_entries_as_wrong_heights`](https://github.com/ProvableHQ/snarkVM/blob/006708fbf6b09b369b95f46a6c24188839f391f1/ledger/store/src/program/finalize.rs#L2413)**

```
seeded: heights 100 and 1000 listed; an unlisted LE entry at height 256
query at height 65,536 after migration -> Some(256u64)
correct answer                          -> Some(1000u64)
```

`LE(256)` and `BE(65_536)` are the same four bytes. That test is the shortest statement of why a migration driven by the heights row is not enough, and it is worth reading before the rest of this description.

## Why the heights map cannot drive this

Three incompatible layouts shipped in three weeks, and nothing on disk records which one wrote a given entry:

| snarkOS | height encoding | `MappingUpdateHeightsMap` |
|---|---|---|
| ≤ v4.7.4 | little-endian | written, lists every entry |
| **v4.7.5, v4.8.0** | little-endian | **not written** |
| v4.8.1+ | big-endian for keys with no row; little-endian for keys with one | frozen, or appended again |

The middle row is the problem. Those entries have no heights row, so a heights-driven migration cannot see them — and on a node that reverted the v4.8.1 change they are *most* of its history. This migration reads `MappingUpdateMap` itself; the heights map is only ever deleted, never trusted. `test_heights_map_is_ignored_and_dropped` proves that against a row which omits a real height, invents two that never existed, and points at a key with no entries.

## What is here

Errors are returned, never panicked: snarkOS wraps ledger loading in `spawn_blocking!`, which turns a panic into `anyhow!("[tokio::spawn_blocking] {error}")` — so a panicking migration reports "task N panicked" where the error chain should carry the remedy. We observed exactly that on a real ledger.

**A storage schema version.** A metadata map (last `DataID` variant, per that enum's append-only rule) records the version a database was last written under, plus migration resume state. `RocksDB::open` refuses a database from a newer build instead of misreading it, and otherwise applies migrations in order, advancing the version only once each completes. It runs whatever cargo features are set, since a migration works on raw keys under a prefix — a build that never opens the typed map can still repair it. It only protects forward from this release; older builds do not consult it.

**Migration v0 → v1.** A suffix byte-reversal on raw keys: `HeightBytes` is a `[u8; 4]` written bare by bincode as the final field, so the value bytes never move. The typed API would deserialize a `Plaintext` and a `Value` per entry only to re-serialize both identically, at ~4.1 KB peak per entry — unaffordable where a few hundred staking keys accrue one entry each per block forever.

**Both legacy paths removed.** The write path is the half that matters: it read the heights row, appended, and rewrote it on every update. No rows survive the migration, so removing it is what actually ends the amplification.

## Measurements — and a correction

An xhigh review found that the sweep below measures the wrong axis, and I have confirmed it. Peak memory is **O(entries in the largest single key)**, not flat: `scan_heights`'s vector, its bincode copy, the `lookup` HashSet and the colliding/isolated partition each hold one element per entry *of the key being migrated*. The sweep scaled key count while holding depth at <= 4,000, so it never varied the dimension that drives memory.

At a constant ~900k entries, varying only the shape:

| shape | entries | peak growth |
|---|---:|---:|
| 447 keys x 2,000 | 894,000 | 0.076 GiB |
| 20 keys x 45,000 | 900,000 | 0.110 GiB |
| 5 keys x 180,000 | 900,000 | 0.118 GiB |

So a real 6.5M-height staking key is on the order of low single-digit GB, **not the ~0.1 GiB originally claimed here**. Still far below the ~10.9 TiB the typed approach would need, and still workable, but the earlier figure was wrong and the benchmark needs a depth axis before any number here should be trusted.

Throughput, which the sweep does measure correctly: 300-550k entries/s.

## Two things a reviewer should attack

**Resume semantics.** An xhigh review found six defects here, all sharing one cause: progress was recorded in a write separate from the data it described. `write_chunk` now puts the cursor in the same `WriteBatch` as its deletes and inserts, which required the work order to be deterministic in the recorded heights alone so a resumed run rebuilds the identical batch list and skips committed batches whole. Three metadata orderings were inverted for the same reason. Every substantive defect found in this work so far has been in the interruption path, and most were found by someone other than me. Specifically here: a key's original heights are recorded to metadata *before* anything moves, because once partly migrated a key's entries no longer say what they were — a moved entry sits at `BE(h) == LE(byte_reverse(h))` and reads back as a plausible height of its own. And `migrate` writes a completion sentinel before returning, since the runner advances the version afterwards and a crash in between would otherwise re-run a finished migration over big-endian data.

**Classifying a mixed database.** An earlier revision refused any database containing big-endian entries. Running it against a real three-format ledger showed that was wrong: it panicked on an entry that byte-reversed to height 53,442, and the reason is structural. A key first written during the v4.7.5/v4.8.0 window has no heights row — that is what defines the window — so when a v4.8.1+ build took over it took the big-endian branch for exactly those keys. **Single keys hold both encodings, and that is the expected shape of any node that ran v4.7.5/v4.8.0 and then upgraded**, so refusing the mixture would refuse the databases this exists to fix.

Entries are classified per key against the **chain tip, read from the committee store** — `CurrentRound` then `RoundToHeight`, both in this same database and reachable by the raw prefix access the migration already uses. A reading above the tip cannot be a height, which settles almost every entry.

Where both readings are plausible:

- a **palindromic** suffix reads as the same height either way, so there is nothing to decide;
- a key with **no big-endian entry** was never touched by a v4.8.1+ build, so everything under it is little-endian;
- a key with **no little-endian entry** was only ever written big-endian;
- a key showing **neither** — a mapping written exactly once, at an ambiguous height — is judged against the ledger as a whole, which usually excludes one encoding;
- only a ledger genuinely holding both, on a key that shows neither, leaves the entry undecidable, and it is reported rather than guessed at.

That fourth case was found by review, and it mattered: without it such an entry was assumed little-endian, so a correct big-endian entry was deleted and rewritten at a height it never had — on ledgers that had nothing to migrate, while `--check` reported them as safe.

There is deliberately no boundary heuristic. An earlier revision placed such entries by which side of the encoding changeover they fell on; it produced two classification defects and earns very little once the bound is a real tip rather than a constant, so it was removed. (Thanks to @snarkworld for catching that this description still described it.)

## What the residual case costs

Measured on real ledgers built from stock binaries:

| ledger | tip | undecidable |
|---|---:|---:|
| genesis-synced, three formats | 76,517 | **6** |
| 20.8M rung, v4.8.0 → v4.9.1 | 20,847,646 | **2,887** |

The count is approximately `mixed keys × little-endian window ÷ 128`, so it grows linearly with how long an operator ran v4.8.0 before upgrading. The second figure came from a five-hour window; a month-long one would be far larger.

Deciding these needs evidence the storage layer does not have — whether a given block actually updated a given mapping key, which is recorded in that block's finalize operations. A phase to do that was written and removed as unjustified complexity; these measurements are the argument for building it, and the decision is open.

## Operator impact

Worth carrying into release notes; there is no CHANGELOG in this repo.

**A node will not start on a ledger whose history has not been migrated.** It reports the schema version it found, the one it needs, and the command to run. This is the intended behaviour: the alternative is the node reading its own history incorrectly, silently, which is the fault being fixed.

**This applies regardless of whether the build has the `history` feature.** The gate is on the data, not on the build. A build that never reads history could stamp the schema version without migrating and would never notice — but that stamp is what a later history-enabled build consults, and it would then skip the migration and misread every entry. So a node carrying unmigrated history must migrate it even if it has no interest in it.

The operator most likely to be caught out is one who ran `--history` at some point, stopped, and now runs a build without it. If that population turns out to be real rather than hypothetical, a `--drop-history` mode that discards the maps and stamps the version would be the humane answer; it is not built, because it destroys data.

**Ledgers with no history are unaffected** — one seek finds nothing, the version is stamped in passing, and the node starts normally. A node that never enabled the feature will not notice this change.

**Migration is not automatic.** It can rewrite billions of entries and take hours, so it is a deliberate operator action:

```
snarkvm-migrate-db --check <ledger-dir>   # read-only; can this be migrated, and how long?
snarkvm-migrate-db <ledger-dir>           # do it, with the node stopped
```

`--check` is safe against a copy of a production ledger. The migration is resumable: interrupt it and run it again.

## Verification

`cargo test -p snarkvm-ledger-store --features history,rocks` — **128 passed, 0 failed, 6 ignored.**

Not yet run against a database produced by real binaries. A three-format ledger built from stock v4.7.4 → v4.8.0 → v4.9.1 exists for exactly that and is the next step.

https://claude.ai/code/session_01GhrVwfiKhGgcRQXUTZfC5p


ledger/store/src/helpers/rocksdb/internal/mod.rs
```diff
@@ -48,6 +51,130 @@ use std::{
...
+        found <= STORAGE_VERSION,
+        "This ledger was written by a newer version of snarkVM (storage schema v{found}; this \
+         build understands v{STORAGE_VERSION}). Upgrade snarkVM, or resync from genesis."
+    );
+    if found == STORAGE_VERSION {
+        return Ok(());
+    }
+
+    // Nothing recorded, and nothing to record: stamp it and carry on.
+    if !history_migration::has_history(database, network_id)? {
```
ljedrz: won't this resolve to `false` (i.e. `has_history == true`) even after migration? I'm missing a call to
```
put_metadata(database, network_id, MetadataKey::StorageVersion, &STORAGE_VERSION.to_le_bytes())?;
```
after the migration logic

snarkworld (commented):
Reviewed at 2e7edf8. Two defects reproduced locally against this branch, plus clippy is failing; details inline.

**Blocking**
1. The storage version is never stamped after a migration (ljedrz's comment above). Confirmed with a test: seed one entry, `migrate()`, read `StorageVersion` back → 0, `has_history` → true. The node refuses forever and rerunning the tool prints success while doing nothing (it returns early on the `CURSOR_COMPLETE` sentinel).
2. `observed_tip` is a lower bound on the chain tip, not an upper bound on real heights, so the newest entries can be misclassified as already migrated. Confirmed with a test (inline on `observed_tip`).
3. `cargo clippy -p snarkvm-ledger-store --features rocks --all-targets -- -D warnings` fails: dead `MetadataKey::StorageMigrationHandoff`, dead `delete_metadata`, dead `Encoding`, and `enum_variant_names` on `MetadataKey`.

**Should be addressed or acknowledged**
4. The PR description says a mixed key is resolved "by the boundary between" its LE and BE entries. `classify_entries` has no boundary logic; every ambiguous entry on a mixed key is reported as undecidable (inline).
5. `check_storage_version` is not gated on the `history` feature, so a build that never reads the map is still forced to migrate before it can start. Fine if intentional, but it needs a release note.
6. The tool infers the network id from the directory suffix; dev ledgers are `.ledger-{network}-{id}` so it picks up the dev id (inline).
7. The map is scanned three times per run: `plan()` in the tool, `plan()` again inside `migrate()`, then `classify_entries` per key (inline).

**Process**
8. New binary plus a new `tracing-subscriber` dependency; per AGENTS.md both need explicit sign-off.

No consensus exposure found: storage only, and the appended `DataID` variant respects the append-only rule.

ledger/store/src/helpers/rocksdb/internal/history_migration.rs
```diff
@@ -0,0 +1,1129 @@
...
+
+    // The heights map is not consulted by anything after this point, and keeping it would leave the
+    // read-modify-write path that made it grow in the first place.
+    drop_map(database, &heights_context)?;
+    let mut batch = rocksdb::WriteBatch::default();
+    batch.put(
+        metadata_key(network_id, MetadataKey::StorageMigrationCursor),
+        bincode::serialize(&(Vec::<u8>::new(), CURSOR_COMPLETE, tip))?,
+    );
+    batch.delete(metadata_key(network_id, MetadataKey::StorageMigrationHeights));
```
snarkworld, line 587: Nothing here (or in `migrate_db.rs` after `migrate()` returns) writes `StorageVersion`. The only stamp on the branch is the fresh-database path in `check_storage_version`, and `has_history` is still true after migration because the entries stay under the `MappingUpdate` prefix. So the node refuses on the next start, and rerunning the tool returns early on `CURSOR_COMPLETE` and prints "Migration complete" without doing anything.

Repro I ran on this branch:
```rust
seed(&db, b"mu", 1, b"v");
migrate(&db, NETWORK).unwrap();
// get_metadata_u32(StorageVersion) == 0, has_history == true
```

Suggest putting `StorageVersion = STORAGE_VERSION` in this same `WriteBatch`, and deleting `StorageMigrationCursor` rather than leaving it at `CURSOR_COMPLETE`. The doc on the cursor says it is "meaningful only to the migration for the version currently being applied," but nothing resets it when the version advances, so a future v1 → v2 migration that reads it would return early and skip its own work. An end-to-end test that opens via `RocksDB::open` after migrating would have caught this; none of the current tests exercise that path.

ledger/store/src/helpers/rocksdb/internal/history_migration.rs
```diff
@@ -0,0 +1,1129 @@
...
+/// of an entry's two readings is its real height, so the true tip is at least the *smaller* of
+/// them. Taking the largest such value over every entry therefore yields a bound that no real
+/// height exceeds, which is the property that matters: judged against it, every entry keeps at
+/// least one plausible reading, so a correct ledger can never be condemned as corrupt.
+///
+/// Deriving it matters more than it looks. Judged against [`CEILING_HEIGHT`] alone, an entry is
+/// ambiguous whenever its height's low byte is small -- one entry in thirty-two. Judged against the
+/// real tip, only entries whose *both* readings are real heights are ambiguous, which on a
+/// 76,514-block ledger is a handful of byte patterns rather than 1,672.
+fn observed_tip(database: &rocksdb::DB, update_context: &[u8]) -> Result<u32> {
```
snarkworld: The doc says this "yields a bound that no real height exceeds." It does not: it is the max over entries of the *smaller* reading, which is a lower bound on the chain tip, and a real height can sit above it. When that happens the entry's true reading fails the `<= tip` test and its byte-reversed reading passes, so it is classified as already big-endian and left in place.

Repro I ran on this branch (all little-endian, a node that stopped at 131_328 = 0x00020100):
```rust
for h in [100u32, 131_327, 131_328] { seed(&db, b"lambda", h, ..); }
observed_tip(..)          // 131_327
classify_entries(..)      // little = [100, 131_327], big = 1
migrate(&db, NETWORK).unwrap();
read_migrated(&db, b"lambda", 131_328)  // None
read_migrated(&db, b"lambda", 66_048)   // Some("v131328")
```

The trigger is a top height whose byte-reversal is below it: for heights under 2^24 that is low byte zero and second byte less than third, roughly 1 in 512 stopping heights. Since the staking keys are updated every block, every one of them at the tip goes wrong together. If the key also holds ambiguous entries, `saw_big` flips and the symptom is instead a spurious "resync from genesis".

The storage layer does know the tip: the block store's height-keyed maps are in the same database, and the tool already works on raw prefixes. Reading the real tip from there removes the inference entirely.

ledger/store/src/helpers/rocksdb/internal/history_migration.rs
```diff
@@ -0,0 +1,1129 @@
...
+        // whichever build wrote it, the entry is already at the key a big-endian write would give
+        // it. Counted as migrated rather than moved, to avoid a write that changes nothing.
+        if as_little == as_big {
+            big += 1;
+            continue;
+        }
+        match (saw_little, saw_big) {
+            (_, false) => little.push(as_little),
+            (false, true) => big += 1,
+            (true, true) => undecidable.push(Undecidable { little: as_little, big: as_big }),
```
snarkworld: The PR description says a key holding both encodings is "placed by the boundary between them, since every little-endian entry predates every big-endian one" and only entries fitting both sides are reported. There is no boundary check here: any ambiguous entry on a key with `saw_little && saw_big` is undecidable. That is safer than described, but it refuses more ledgers than the description claims, and reviewers evaluating the residual-case argument are evaluating logic that is not in the code. Either implement the boundary test (LE reading valid iff `as_little < min(certain BE)`, BE reading valid iff `as_big > max(certain LE)`, decide when exactly one holds) or update the description.

ledger/store/src/helpers/rocksdb/internal/history_migration.rs
```diff
@@ -0,0 +1,1129 @@
...
+    if matches!(&resume, Some((_, index, _)) if *index == CURSOR_COMPLETE) {
+        return Ok(());
+    }
+
+    let tip = match &resume {
+        Some((_, _, tip)) => *tip,
+        None => {
+            // Nothing has been written yet, so this is the moment to refuse. Everything is
+            // classified up front; only once that succeeds does anything move.
+            let plan = plan(database, network_id)?;
```
snarkworld: On a fresh run every entry is read five times before the last write: `plan()` in the tool (`migrate_db.rs:121`) is two passes, one in `observed_tip` and one in the per-key classify loop; this `plan()` call is the same two passes again; and `classify_entries` at line 430 is a fifth read of each key. On the archive-node shape this exists for that is hours of extra wall time. Consider having the tool pass its `RepairPlan` in, or having `plan()` return the per-key `little` vectors so the loop does not re-derive them. The tool should also read `StorageVersion` and the cursor (two point reads) before scanning at all, so an already-migrated or resuming ledger is not walked end to end just to print a status line.

ledger/store/src/helpers/rocksdb/internal/history_migration.rs
```diff
@@ -0,0 +1,1129 @@
...
+//! # Resuming
+//!
+//! A key's entries are ambiguous once partially migrated: a moved entry sits at
+//! `BE(h) == LE(byte_reverse(h))` and reads back as a plausible height of its own. So the migration
+//! does not re-derive its work from the data. Before touching a key it records that key's original
+//! heights in storage metadata, and advances an index into that list as batches commit. An
+//! interrupted run reads the list back and continues from the index, never re-deriving what it has
+//! already moved.
+
+use super::{MapID, MetadataKey, PREFIX_LEN, ProgramMap, get_metadata, metadata_key};
```
snarkworld: Nit: crate convention is std first, external crates second, crate-local third. This file has `super` first, then `serde`, then `anyhow`, then `std`.

ledger/store/src/helpers/rocksdb/internal/mod.rs
snarkworld: This check is not gated on the `history` feature, so a build that never opens the typed map is still refused until the operator migrates data it does not read. The PR body presents that as a feature ("a build that never opens the typed map can still repair it"), which is fine, but it is a startup-blocking behavior change for non-history nodes with old history data on disk and should be called out in the release notes.

ledger/store/src/bin/migrate_db.rs
```diff
@@ -0,0 +1,163 @@
...
+            _ => positional.push(arg),
+        }
+    }
+
+    let Some(path) = positional.first().cloned() else { bail!("{USAGE}") };
+    let network_id = match positional.get(1) {
+        Some(explicit) => explicit.parse()?,
+        None => {
+            let name = std::path::Path::new(&path).file_name().and_then(|name| name.to_str());
+            match name.and_then(|name| name.rsplit_once('-')).and_then(|(_, id)| id.parse().ok()) {
```
snarkworld: Dev-mode ledgers are named `.ledger-{network}-{id}` (see `aleo-std-storage`), so `rsplit_once('-')` yields the dev id, not the network. For `.ledger-1-3` the tool infers network 3, finds zero keys under that prefix, prints an estimate of 0.0 minutes, runs a no-op migration that writes `CURSOR_COMPLETE` under network 3, and prints "Migration complete" while the network-1 history is untouched. Either parse `ledger-{n}` / `.ledger-{n}-{id}` explicitly, or range-check the inferred id and require the explicit argument otherwise.

ledger/store/Cargo.toml
```diff
@@ -124,6 +129,11 @@ optional = true
 [dependencies.tracing]
 workspace = true
 
+[dependencies.tracing-subscriber]
```
snarkworld, line 136: New dependency (and a new `[[bin]]` above). Per AGENTS.md both need explicit approval; flagging so it is a conscious decision rather than an implicit one.

ledger/store/src/helpers/rocksdb/internal/mod.rs
snarkworld: Confirmed on 2e7edf8. Seeded one entry, ran `migrate()`, read `StorageVersion` back: 0, and `has_history` is still true because migrated entries stay under the `MappingUpdate` prefix. So the node refuses on the next start, and rerunning the tool returns early on `CURSOR_COMPLETE` and prints success without doing anything.

The one-line `put_metadata` is necessary but I would also delete `StorageMigrationCursor` in the same `WriteBatch`: nothing resets it when the version advances, so a future v2 migration that reads the leftover sentinel would skip its own work. Details in my review on `history_migration.rs:452`.

ledger/store/Cargo.toml
cbeck88: true, and we should discuss whether we want `snarkvm-db-migrate` to be a standalone binary (so that the operator explicitly runs migration) or something that runs when the node starts up unconditionally.

usually in webapps you want to have a human watching when migration happens, because if migration goes wrong you can get corrupted data base or data loss. for a blockchain it might not be important to do that because all this data is highly replicated anyways by design, and backed up on CDNs etc.

ledger/store/src/helpers/rocksdb/internal/history_migration.rs
cbeck88: it used to do that but we lost the write when refactoring in one of the last two commits, fixing now

ledger/store/src/helpers/rocksdb/internal/mod.rs
cbeck88: Confirmed and fixed in 1055c3e6 — thank you, this was a complete functional failure.

Moving the migration out of the node removed the runner that advanced `StorageVersion`, and nothing replaced it. Reproduced exactly as you describe: seed one entry, `migrate()`, read it back → 0, `has_history` → true.

`StorageVersion` is now written in the same `WriteBatch` that disposes of the migration's working state, and I took snarkworld's further point as well — the `CURSOR_COMPLETE` sentinel is gone entirely, since the version is the record the node already checks, and a leftover sentinel could have made a future v1→v2 migration skip its own work.

ledger/store/src/helpers/rocksdb/internal/history_migration.rs
cbeck88: Fixed in 1055c3e6, and I took both halves of the suggestion.

`StorageVersion` and the deletion of `StorageMigrationCursor` now commit in the same batch, and the `CURSOR_COMPLETE` sentinel is removed rather than left behind — your point about a future v1→v2 migration reading the leftover and returning early is exactly right, and it is a better reason to drop the sentinel than the one I would have given.

You are also right about the missing test. The old one asserted only that a second run was harmless, which the sentinel satisfied while the version was never written. The replacement asserts the record itself: version stamped, cursor gone, working state gone. That is the assertion whose absence hid this.

ledger/store/src/helpers/rocksdb/internal/history_migration.rs
cbeck88: Confirmed and fixed in 1055c3e6. Your repro reproduces verbatim: tip 131,327, the entry at 131,328 classified as already-migrated, and its value readable at 66,048 afterwards.

You are right that the doc claimed an upper bound on real heights while the function computes a lower bound on the tip. What I had actually proved is the weaker statement that no entry is implausible under *both* readings — and that is what the test asserted, so it passed while the doc said something false. The two claims were about the same quantity and only one of them was true.

And you are right that the tip was never out of reach. The committee store shares the database, so `CurrentRound` → `RoundToHeight` gives it through the same raw prefix access the migration already uses. `observed_tip` and its ceiling constant are gone, replaced by reading it. The test now covers your case directly, including asserting that nothing appears at 66,048.

ledger/store/src/helpers/rocksdb/internal/history_migration.rs
cbeck88: You are right, and it is the description that is wrong rather than the code. The boundary heuristic existed in an earlier revision, produced two classification defects, and was removed once the plausibility bound became real — at which point it earned nothing. The description was not updated with it.

I am correcting the PR body rather than implementing the boundary test. Your reading of the tradeoff matches mine: refusing more ledgers is the safer error, and with the tip now read rather than inferred the ambiguous population is small enough that the heuristic would be recovering very little at the cost of the logic that went wrong twice.

Thank you for catching that reviewers were being asked to evaluate an argument that was not in the code.

ledger/store/src/helpers/rocksdb/internal/history_migration.rs
cbeck88: Fixed in 1055c3e6. `plan()` is now computed once by the tool and passed into `migrate()`, so the fresh-run path plans once rather than twice.

I also took the second half: the tool reads `StorageVersion` and the cursor as point lookups before scanning, so an already-migrated ledger prints its status without a walk, and a resuming one goes straight back to work. On the archive shape this exists for, that was hours spent to print a line.

The remaining per-key `classify_entries` in the migrate loop I have left. Having `plan()` return the per-key height vectors would remove it, but those are unbounded — a staking key on a real archive carries millions of heights, and holding them all is the memory profile this design exists to avoid.

ledger/store/src/bin/migrate_db.rs
cbeck88: Fixed in 1055c3e6. Parsing is now explicit: strip a leading `.`, require the `ledger-` prefix, take the first component after it, and range-check against the known networks. Anything outside that is refused with a request for the explicit argument rather than assumed.

The failure mode you describe is what made this worth fixing properly rather than patching the split — it succeeded. Wrong network inferred, no history found under that prefix, a cursor stamped under a network nobody owns, and "Migration complete" printed while the real history sat untouched. A tool that reports success having done nothing is worse than one that fails.

ledger/store/src/helpers/rocksdb/internal/history_migration.rs
cbeck88: Fixed in 1055c3e6 — imports reordered to std, external, crate-local.

ledger/store/Cargo.toml
cbeck88: Signed off on both the `[[bin]]` and the `tracing-subscriber` dependency — thank you for flagging them rather than letting them pass implicitly.

The larger question they point at is genuinely open, and worth settling before this merges: **should the migration be a standalone binary at all, or a `snarkos ledger migrate` subcommand?**

The case for standalone, which is what is here now:

- No node machinery — no config, keys, or network. It is a data tool operating on a directory.
- The migration logic lives in `snarkvm-ledger-store` and works on raw prefixes, so a binary beside it has no dependencies a node would drag in.
- Easy to run against a copy, which is how anyone sensible will first use it.

The case for a subcommand, which I now think is stronger than I judged when I built this:

- **Distribution.** Operators receive snarkOS releases. They do not build `snarkvm-ledger-store` with `--features rocks`, and a binary that lives in this repo has no route to them without one being invented.
- **Discovery.** `snarkos --help` is where someone looks after the node tells them to migrate. A tool named in an error message but absent from the binary they have is a poor handoff.
- I had thought a compliance constraint favoured standalone, on the grounds that a partner told us they cannot run non-stable builds. Re-reading it, that constraint is about *patched* builds, not about snarkOS as such — a stable release carrying a `ledger migrate` subcommand is precisely what they can run. So that argument does not hold, and I should not have leaned on it.

What does not change either way: the work belongs in `snarkvm-ledger-store` where the storage layout is, and the node refuses to start on a schema it does not understand rather than migrating implicitly. The question is only which binary the entry point hangs off.

I have no strong preference and would rather defer to whoever owns the release and packaging story. If the answer is a subcommand, the change is small — the two entry points are thin wrappers over `plan()` and `migrate()`.

ledger/store/src/helpers/rocksdb/internal/mod.rs
cbeck88: Intentional, and on reflection it is a correctness requirement rather than a convenience — so I have recorded the reasoning in the code (085847297) rather than leaving it to be inferred.

The gate is on the **data**, not on which features the build was compiled with. A build that never reads history could stamp the version without migrating and would never notice the difference. But that stamp is exactly what a later history-enabled build consults, and it would then skip the migration and read little-endian entries as big-endian — silent wrongness, which is the failure this whole change exists to end. Blocking a non-history node that carries unmigrated history is the price of the version meaning what it says.

You are right that it needs calling out. There is no CHANGELOG in this repo, so I have added an **Operator impact** section to the PR description; tell me if release notes are kept somewhere I have not found and I will move it.

Worth naming the sharp edge for whoever writes the release notes: the operator most likely to be surprised is one who ran `--history` at some point, stopped, and now runs a build without it. They have no interest in the history data, yet must migrate it before the node will start. If that turns out to be a real population rather than a hypothetical one, the humane answer is probably a `--drop-history` mode on the tool that deletes the maps and stamps the version — it is a few lines, but it discards data, so I would not add it speculatively.

ledger/store/Cargo.toml
vicsn: In general, I think it makes sense for migrations not to be run automatically/unconditionally:
- As Chris said, we may want to have a human in the loop in case of corruption
- We don't want a quorum of validators to ever directly activate this and to stall consensus for the duration of the migration

As an exception, the design in https://github.com/ProvableHQ/snarkVM/pull/3329 *is* run automatically/unconditionally but it's a special case: because it only prunes data there is no interference with consensus. And anyhow we can decide later on the final architecture for that PR.

That leaves the question of snarkVM binary v.s. snarkOS migrate. One more advantage of the `snarkOS migrate` approach is that this could potentially also cover migrations of `BFTPersistentStorage`. If `BFTPersistentStorage` would be moved into snarkVM this benefit would be void, but that would be yet another huge work package.

So I'm in favour of a `snarkOS migrate` command - for which most logic can live in snarkVM, "close" to the db logic.

ledger/store/src/helpers/rocksdb/internal/mod.rs
```diff
@@ -48,6 +51,147 @@ use std::{
...
+/// The storage schema version this build writes and understands.
+///
+/// Bump this whenever the on-disk layout changes in a way that a build expecting the previous
+/// version would read incorrectly. A database records the version it was last written under, so a
+/// build can refuse a database from the future instead of silently misreading it.
+///
+/// Note this only protects forward from the release that introduced it: builds older than that do
+/// not consult the record at all. Three incompatible historical-mapping layouts shipped within
+/// three weeks with nothing on disk to distinguish them, which is the situation this exists to
+/// prevent recurring.
```
vicsn, line 63: I feel like this is a bit too much, see: https://github.com/ProvableHQ/snarkOS/pull/4433/changes#diff-a54ff182c7e8acf56acfd6e4b9c3ff41e2c41a31c9b211b2deb9df75d9a478f9R24

ledger/store/src/helpers/rocksdb/internal/mod.rs
```diff
@@ -48,6 +51,147 @@ use std::{
...
+/// Returns whether the migration from `version` to `version + 1` has anything to do here.
+///
+/// Lets the version be stamped in passing when every outstanding migration is a no-op, without that
+/// shortcut being tied to what any one of them happens to be about.
+fn has_work(database: &rocksdb::DB, network_id: u16, version: u32) -> Result<bool> {
+    match version {
+        0 => history_migration::has_history(database, network_id),
+        other => bail!("No storage migration is defined for schema v{other}"),
+    }
+}
```
vicsn, line 138: Mild preference for the following setup:
- call it `fn requires_migration`
- currently there is no compile-time check whether we're including all possible versions and we have to manually remember to add entries here. We could:
  - replace `const STORAGE_VERSION: u32 = 1` by an `enum StorageVersions` similar to what we do for `enum ConsensusVersions`.
  - match here on the enum, for the latest one we should always return `Ok(false)`

ledger/store/src/helpers/rocksdb/internal/mod.rs
```diff
@@ -48,6 +51,147 @@ use std::{
...
+    ensure!(
+        found <= STORAGE_VERSION,
+        "This ledger was written by a newer version of snarkVM (storage schema v{found}; this \
+         build understands v{STORAGE_VERSION}). Upgrade snarkVM, or resync from genesis."
+    );
+    if found == STORAGE_VERSION {
+        return Ok(());
+    }
+
+    // Nothing recorded, and nothing to record: stamp it and carry on.
```
vicsn, line 164: What?

cbeck88: Yeah some of the comments are wierd and it's partly the code getting manipulated by tools that don't respect the comments

cbeck88: I'm mostly focused on correctness so far but let's get the comments right too. There was still a correctness bug yesterday I think

ledger/store/src/helpers/rocksdb/internal/mod.rs
cbeck88: yeah that seems reasonable

vicsn:

PR #3409, opened 2026-09-05
fix(program): reject trailing bytes inside length-delimited sub-objects

cbeck88:
## Motivation

`Future`, `Plaintext` and `Record` all embed a sub-object in a `u16`-delimited
region. Every one of them reads the length, takes that many bytes, and parses the
sub-object out of them — without checking that the sub-object consumed the whole
region:

```rust
let num_bytes = u16::read_le(&mut reader)?;
let mut bytes = Vec::new();
(&mut reader).take(num_bytes as u64).read_to_end(&mut bytes)?;
let entry = Argument::read_le_internal(&mut bytes.as_slice(), depth)?;
```

Anything left over in that region is silently dropped. The writer emits the true
length:

```rust
let bytes = argument.to_bytes_le()?;
u16::try_from(bytes.len())?.write_le(&mut writer)?;
```

So a region can be inflated by up to 65,535 bytes of arbitrary padding and the
value still parses — to **the same value**, which then re-serializes shorter.

**One transaction, many byte encodings, one transaction ID.** Since a
`TransmissionID` is a hash of the transmitted bytes, that is two TransmissionIDs
for one transaction. Struct members and array elements nest, so the number of
encodings of a single transaction is combinatorial rather than merely large.

snarkVM already guards this at the top level — `deserialize_transaction_strict`
rejects trailing bytes after a transaction. The same guard is simply missing on
these inner regions.

## The four sites

| File | Region |
|---|---|
| `console/program/src/data/future/bytes.rs:55` | future argument |
| `console/program/src/data/plaintext/bytes.rs:52` | struct member |
| `console/program/src/data/plaintext/bytes.rs:74` | array element |
| `console/program/src/data/record/bytes.rs:49` | record entry |

Each gains two checks: the declared region must be entirely present, and the
sub-object must consume all of it.

The truncation check matters for the same reason as the trailing check. Without
it, a region declared longer than the remaining stream yields a short read, and
the value re-serializes with the true smaller length — the same divergence from
the other direction.

## How this was found

A fuzzing campaign on the re-encoding oracle: parse, re-encode, require byte
identity. Five artifacts, all at the future-argument site:

| artifact | declared | actual | delta |
|---|---|---|---|
| `04194fa1a3` | 35 | 3 | −32 |
| `3daf94a166` | 36 | 5 | −31 |
| `69ebd5da88` | 36 | 6 | −30 |
| `81a2b3ed9f` | 36 | 5 | −31 |
| `ad256562ac` | 36 | 3 | −33 |

Every delta is exactly `declared - actual`. Each artifact reports
`the re-encoded bytes parse again; same transaction id: true`.

The other three sites were then confirmed by construction rather than assumed —
each has a regression test that fails on an unpatched tree.

## Test plan

Four tests, one per site, each asserting the unpadded encoding still round trips
*before* asserting padded ones are rejected — so a fix that rejected both would
fail loudly rather than pass quietly.

On an unpatched tree all four fail:

```
a future argument declaring 1 bytes more than it uses must be rejected
a struct member declaring 1 bytes more than it uses must be rejected
an array element declaring 1 bytes more than it uses must be rejected
a record entry declaring 1 bytes more than it uses must be rejected
```

With the patch: 4 pass, and `snarkvm-console-program`'s full suite is 303
passed / 0 failed. Applied to the vendored tree, all five artifacts are rejected.

## Impact: a reachable commit stall, not only malleability

The padding survives `deserialize_transaction_strict`, because that check catches
trailing bytes only at the outermost level and this padding sits *inside* the
`u16` regions. That makes a second valid `(transaction_id, checksum)` pair
constructible, and a `TransmissionID` is exactly that pair — the checksum is
SHA3 over the raw transmitted bytes
(`ledger/narwhal/data/src/lib.rs:53-65`).

The consequence is worse than duplicate IDs. A byzantine validator can author a
batch header committing to `(id, padded_checksum)`:

1. Honest validators fetch the padded bytes and validate them. The transaction ID
   matches, and the checksum is computed on the still-padded buffer, so it
   matches too. They sign and certify the batch.
2. The transmission is then canonicalized — the padding is dropped *after* the
   checksum check.
3. At commit, `decouple_transmissions` asserts the canonical checksum equals the
   committed one. It does not. `prepare_advance_to_next_quorum_block` bails, and
   the leader cannot build the block.

This patch turns it into a clean rejection at parse time.

*(Traced by reading the path, not by executing it. A test driving a padded batch
header through to commit would confirm it, and is worth writing.)*

## Backwards compatibility

Compatible. Both sides were audited rather than assumed.

**Nothing on disk carries padding.** In snarkVM every persistence path
re-serializes from the parsed object: block storage decomposes into typed maps
(`FutureMap = Option<Future<N>>` and friends,
`ledger/store/src/transition/output.rs:38-53`), and RocksDB writes bincode over
`to_bytes_le()` (`ledger/store/src/helpers/rocksdb/internal/map.rs:87`). The one
place raw bytes are retained, `Data::Buffer`, is confined to narwhal
transmissions and never persisted. On the snarkOS side every producer
canonicalizes before writing; even the proposal cache, which genuinely round
trips bytes through a file, can only ever receive canonical ones.

**A padded encoding cannot already be on chain.** Block verification recomputes
the transmission checksum from the canonical re-serialization
(`ledger/block/src/verify.rs:631-640`), so a block including a padded
transmission fails verification on every node today. A syncing mainnet is
evidence none has been committed. The genesis block parses unchanged with this
patch applied.

**What does change is what a node accepts from a peer.** An old node would sign a
batch carrying a padded transmission where a new node would not. That
disagreement is only reachable by an adversary — no honest client emits padding,
since `to_bytes_le` cannot produce it — and under that same attack today the
batch is certified and then stalls at commit. Refusing it earlier is the better
failure, and it degrades toward the current behaviour rather than away from it.

### DIFF
--- a/console/program/src/data/future/bytes.rs
+++ b/console/program/src/data/future/bytes.rs
@@ -53,8 +53,18 @@ impl<N: Network> Future<N> {
             // Read the argument bytes.
             let mut bytes = Vec::new();
             (&mut reader).take(num_bytes as u64).read_to_end(&mut bytes)?;
+            // Ensure the declared region is entirely present.
+            if bytes.len() != num_bytes as usize {
+                return Err(error("Failed to read future: argument is truncated"));
+            }
             // Recover the argument.
-            let entry = Argument::read_le_internal(&mut bytes.as_slice(), depth)?;
+            let mut slice = bytes.as_slice();
+            let entry = Argument::read_le_internal(&mut slice, depth)?;
+            // Ensure the argument fills its region. Bytes left unread would be dropped
+            // on re-serialization, giving one value a second encoding.
+            if !slice.is_empty() {
+                return Err(error("Failed to read future: argument has trailing bytes"));
+            }
             // Add the argument.
             arguments.push(entry);
         }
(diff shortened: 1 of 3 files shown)


ljedrz (approved):
LGTM, it's a sound hardening.

PR #3410, opened 2026-09-06
fix(narwhal): reject repeated entries in count-prefixed collections

cbeck88:
During deserialization, `BatchHeader`, `BatchCertificate` and `Subdag` each read a count and then insert
that many entries into an `IndexSet` or a `BTreeMap`:

```rust
let mut transmission_ids = IndexSet::new();
for _ in 0..num_transmission_ids {
    transmission_ids.insert(TransmissionID::read_le(&mut reader)?);   // return value discarded
}
```

The collection absorbs a repeat silently — `insert`'s return value is discarded,
and `collect` into a set has nowhere to report one. So an encoding carrying a
duplicate parses to the same object and re-serializes shorter: **one value, two
encodings.**

Measured on a sampled batch header:

```
962 bytes in  ->  ACCEPTED  ->  re-encodes to 937 bytes, same batch ID
```

## Why nothing downstream catches it

A batch header's ID is recomputed on read and compared against the wire value
(`bytes.rs`), and the signature is verified over that recomputation. **Both
pass.** The batch ID is computed over the *deduplicated* set, so the duplicate
never reaches the preimage. Every guard operates on the parsed object, and the
duplicate only ever existed in the bytes.

## The five sites

| File | Collection |
|---|---|
| `batch-header/src/bytes.rs` | `transmission_ids` |
| `batch-header/src/bytes.rs` | `previous_certificate_ids` |
| `batch-certificate/src/bytes.rs` | `signatures` |
| `subdag/src/bytes.rs` | `certificates` per round |
| `subdag/src/bytes.rs` | rounds (`BTreeMap`) |

Each now checks the collection's length against the declared count.

## Order is deliberately left alone

Sorting on write was considered and rejected: it would rewrite every historical
batch ID. `compute_batch_id` hashes the transmission IDs **in insertion order**
(`to_id.rs`), and the signature covers the result.

That order is arbitrary — a proposal drains its worker FIFO, so it reflects
arrival order at one validator — but it is *authenticated*, which is what
matters. Reordering is already rejected today; only duplication slipped through,
because dedup happens before the ID is computed.

**The exception is `Subdag`'s rounds.** The writer iterates a `BTreeMap`, so it
already emits ascending order; the reader now requires it. That rejects nothing
an honest node produces and closes reordering as well as repetition.

## Test plan

Five regression tests, one per site, each asserting the honest encoding still
round trips *before* asserting the tampered one is refused — so a fix that
rejected both would fail loudly rather than pass quietly.

All five fail on an unpatched tree:

```
a batch header repeating a transmission ID must be rejected
a batch header repeating a previous certificate ID must be rejected
a certificate repeating a signature must be rejected
a subdag repeating a certificate within a round must be rejected
a subdag whose rounds are not ascending must be rejected
```

With the patch: batch-header 5 passed, batch-certificate 8 passed, subdag 19
passed. Full pre-commit hook green (`clippy --workspace --all-targets
--all-features -D warnings` plus fmt).

## Backwards compatibility

No honest encoding is affected. `to_bytes_le` writes `collection.len()` and
iterates that same collection, so it can never emit a duplicate, and the subdag
writer already emits ascending rounds. Nothing already on chain can carry one.

Related to #3409, which fixes the same class — a non-canonical encoding accepted
— in `Future`/`Plaintext`/`Record`.

---

### DIFF
--- a/ledger/narwhal/batch-certificate/src/bytes.rs
+++ b/ledger/narwhal/batch-certificate/src/bytes.rs
@@ -31,9 +31,15 @@ impl<N: Network> BatchCertificate<N> {
         let mut signature_bytes = vec![0u8; num_signatures as usize * Signature::<N>::size_in_bytes()];
         reader.read_exact(&mut signature_bytes)?;
         // Read the signatures.
-        cfg_chunks!(signature_bytes, Signature::<N>::size_in_bytes())
+        let signatures = cfg_chunks!(signature_bytes, Signature::<N>::size_in_bytes())
             .map(|data| Signature::read_le_with_unchecked(data, unchecked))
-            .collect::<Result<IndexSet<_>, _>>()
+            .collect::<Result<IndexSet<_>, _>>()?;
+        // Ensure no signature was repeated. The set absorbs a repeat, so it would
+        // otherwise give one certificate a second encoding.
+        if signatures.len() != num_signatures as usize {
+            return Err(error("Duplicate signature in batch certificate"));
+        }
+        Ok(signatures)
     }
 }
 impl<N: Network> FromBytes for BatchCertificate<N> {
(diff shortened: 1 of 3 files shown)


ljedrz (approved):
LGTM, hardening.

edit: seems like there may be some test glitches to iron out first, though.

vicsn (approved):

Issue #3415, opened 2026-09-08
[Feature] Reduce transaction latency variance


vicsn:
## 🚀 Feature

When a validator receives a transaction, it may call `check_transaction` several times. In principle, multiple calls should not influence latency much due to `check_transactions_cache`. 

If you call the broadcast API endpoint with the additional synchronous `?check_transaction=true`, we observe a funky distribution of response times:
<img width="1018" height="382" alt="Image" src="https://github.com/user-attachments/assets/b240c52b-a72e-42fc-9ff0-755ea878731c" />

`snarkos_consensus_transmission_latency` shows a similar issue:

<img width="1051" height="320" alt="Image" src="https://github.com/user-attachments/assets/2fee8be0-059f-4bd4-af28-8dcfab6ca18b" />

Some possible contenders:

Site in check_transaction | Shared with prepare/speculate? | Multi-second stall?
-- | -- | --
self.process.lock() | No — never taken | No
process.get_stack(s) / contains_program | stacks RwLock; speculate only writes on deploy stage_stack / revert_stacks | Unlikely (brief writes)
Uniqueness via transition_store / block_store | Concurrent reads vs speculate | Unlikely
block_store().current_block_height() (tree.read()) | Tree is not written during prepare | No
current_block | Prepare holds a read for the whole call; check_transaction does not use it | No for verification
Public-fee finalize_store().get_value_speculative | Same key_value_map atomic_batch mutex as speculate dry-run | Possible jitter, not a clean multi-second exclusive wait
partially_verified_transactions LRU | Short peek/push | No
Process::verify_execution / verify_fee (Varuna) | Same rayon pool as prepare_for_speculate’s parallel check_transaction | Yes — CPU / pool saturation

Solution ideas:
- we should have better instrumentation to confirm whether latency spikes correlate to CPU saturation, and how much duplicate transaction verification is happening
  - if we have CPU saturation, we could introduce a separate rayon threadpool for hot path transaction verification (within block creation) v.s. slow path transaction verification (incoming broadcasts)
  - if we do duplicate transaction verfication, we can be consider introducing smarter verification flags/locks/guards

vicsn:
Relevant instrumentation PRs:
- rocksdb: https://github.com/ProvableHQ/snarkVM/pull/3296
- snarkVM: https://github.com/ProvableHQ/snarkVM/pull/3416
- snarkOS: https://github.com/ProvableHQ/snarkOS/pull/4451

[...]

PR #3416, opened 2026-09-08
feat(metrics): instrument check_transaction cache and speculate latency

vicsn:
## Summary
- Adds VM metrics for `check_transaction` cache hit/miss, duration (labeled by `cache`), and in-flight count, plus in-flight/duration gauges for `speculate`, `prepare_for_speculate`, and `atomic_speculate`.
- This is the instrumentation slice of #3415: it lives in snarkVM because the partial-verification cache and both verification paths are here. Host CPU remains a GCE/snarkOS overlay; thread-pool splits belong in a follow-up once these series confirm the bottleneck.
- Enable with the existing `metrics` feature (`snarkvm/metrics` already forwards into synthesizer). Overlay `snarkvm_vm_check_transaction_duration_seconds{cache="hit"}` with `snarkvm_vm_prepare_for_speculate_in_flight` and instance CPU: slow cache hits during speculate mean contention, while a high miss ratio means duplicate proof work.

## Test plan
- [ ] `cargo test -p snarkvm-metrics`
- [ ] `cargo test -p snarkvm-synthesizer --features metrics --lib -- test_verify`
- [ ] On a metrics-enabled node, confirm `snarkvm_vm_check_transaction_cache_{hit,miss}_total` increment under `?check_transaction=true` broadcasts
- [ ] Overlay cache-hit duration with `prepare_for_speculate_in_flight` and GCE CPU during a latency spike


synthesizer/src/vm/finalize.rs
```diff
@@ -1432,6 +1442,11 @@ impl<N: Network, C: ConsensusStorage<N>> VM<N, C> {
         state: FinalizeGlobalState,
         rng: &mut R,
     ) -> Result<(Vec<&'a Transaction<N>>, Vec<(&'a Transaction<N>, String)>)> {
+        #[cfg(feature = "metrics")]
+        let _prepare_metrics = snarkvm_metrics::vm::TimedInFlight::enter(
+            snarkvm_metrics::vm::PREPARE_FOR_SPECULATE_IN_FLIGHT,
+            snarkvm_metrics::vm::PREPARE_FOR_SPECULATE_DURATION_SECONDS,
+        );
```
cbeck88, line 1449: its a bit strange that this measures a variable number of transactions, but i can see why you might want this

suggestion, what if we also had a gauge / histogram for the number of transactions in the `transactions` function parameter here ?

cbeck88 (approved):

PR #3417, opened 2026-09-08
Rebuild the history by replaying blocks, instead of re-encoding it

cbeck88:
Draft, for 4.10.1. **An alternative to #3408, not a successor to it** — same base (`testnet`), same storage-schema-version machinery, but it repairs the history by discarding and replaying it rather than by re-encoding the entries in place. Both branches merge independently; the point is to measure this one on the cluster and pick.

## Why an alternative

#3408 classifies every historical mapping entry as little- or big-endian and byte-reverses the little-endian ones. Reviewing the direction turned up a property of the corruption that no amount of classification can get past:

**The two encodings collide, and the collision already destroyed data on the running node.**

`LE(256)` and `BE(65_536)` are the same four bytes, and `HeightBytes` is the final field of the raw key. A mapping key written at height 256 while the node ran a little-endian build, and again at height 65,536 after it upgraded, produced *one* raw key — the second write overwrote the first, and height 256's value is gone. That happened months ago, on the partner's node, before any repair ran.

So a repair that relocates surviving entries inherits those holes. It cannot be complete, however good its classifier is.

**How many.** The collision needs `byte_reverse(h)` to also be a real height, which for a ~20.8M tip forces the low byte of `h` to be ≤ 1 — roughly 0.6% of the little-endian window. `credits.aleo/{committee, delegated, bonded}` reach that bound exactly, because `replace_mapping` rewrites every key in the mapping on every reward block. On #3408's measured ledger (20.8M tip, five-hour window, ~100 mixed keys) that arithmetic lands at ~2,200 entries, against the 2,887 it reports as "undecidable" — close enough that those are plausibly the same entries, seen from the other side: not "we cannot decide which height this is", but "one of these two heights lost its value".

Filling them requires regenerating the values, and the values are recoverable: the blocks are still on disk.

## What this does

Discards the finalize and committee stores, then replays every block already in local storage. No network, no 7 TB resync.

It is **self-verifying**, which is the main reason to prefer it, and the verification covers both halves of a block:

- Per confirmed transaction, `VM::atomic_finalize` already compares every recomputed finalize operation against the one the block records (`synthesizer/src/vm/finalize.rs:1030`, `:1060`) and aborts the batch on a mismatch.
- Per block, the replay checks its finalize root against the header's. That half was missing at first: the operations returned by pre- and post-ratify were discarded, and on a live node their coverage comes from `check_next_block`, which a replay never runs — so `replace_mapping` of the `credits.aleo` staking mappings, the majority of an archive node's history, was being rebuilt against nothing.

A replay that completes is one that agreed with the chain on every transaction *and* every ratification of every block. There is no classifier to review, no heuristic, no undecidable residue, and no boundary inference.

**The mapping state goes with the history**, necessarily: reproducing the update at height `h` needs the state as it stood at `h - 1`, and the only state on disk is the one at the tip. Blocks, transactions, transitions and deployments are never touched — they are the source.

## Shape

- **A storage schema version**, unchanged in substance from #3408: a metadata map records the version a database was last written under, `RocksDB::open` refuses one from a newer build, and a ledger with no history is stamped in passing so a fresh node never sees a prompt for work that does not exist. An explicit `allow_downlevel_open` is the one bypass, for the rebuild itself.
- **Both legacy paths removed**, as in #3408. The write path is the half that matters: retiring the heights row is what ends the amplification.
- **`VM::replay_block`**, a new `SequentialOperation` variant that finalizes a block without inserting it — re-inserting would append to the block Merkle tree a second time. It shares `finalize_state` with `add_next_block_inner` so the two cannot drift.
- **`VM::rebuild_finalize_state`**, the driver. Bound to `ConsensusDB` rather than any `ConsensusStorage`, so it cannot be called on a memory-backed VM and go wipe a ledger on disk.

## Progress reporting

Requested, and easier to do honestly here than in #3408: the work is a known number of blocks, so percentage and ETA are direct rather than inferred from an entries-per-second constant (#3408 had to correct its estimate downward by 7×). Every 30 seconds:

```
Rebuilding: block 8214000/20847646 (39.4%), 214 blocks/s, 16h24m remaining, 10h39m elapsed
```

The rate comes from the most recent window, not from genesis. Early blocks are near-empty and late ones carry the chain's real load, so a running average is optimistic for hours — the opposite of what an operator planning a maintenance window needs.

## Safety properties

- **Block completeness is checked before anything is discarded.** A ledger missing a block cannot be rebuilt, and discovering that at height 12,000,000 with the finalize state already gone would turn a ledger that merely reads history wrongly into one that serves nothing.
- **The feature set is checked before anything is discarded.** The clear works on raw prefixes, so it removes history whether or not the build was compiled to write any. Without the guard the failure is silent and total: history cleared, replay writes none, schema version stamped over the result. A build without `history` (or `history-staking-rewards`) refuses a ledger that holds it — and because an emptied history map is indistinguishable from one that never existed, the clear *records* which optional maps it discarded, so a run begun by a `history` build cannot be finished by one without it.
- **Resume needs no cursor.** The resume point is the committee store's current height, which ratification writes inside the same atomic batch as the block's mapping updates — the state's own account of where it got to, rather than a record kept alongside the data it describes and able to disagree with it. `CommitteeStorage::insert` additionally requires each height to follow the last, so a replay resuming in the wrong place is refused by the store. Every substantive defect found in #3408's lineage was in the interruption path; this removes the mechanism those defects lived in.
- **A half-finished rebuild refuses to serve.** An in-progress flag is set before the clear and cleared after the version is stamped, because an empty finalize store cannot distinguish itself from a fresh ledger and only one of the two must refuse reads.

## Operator impact

Same as #3408 in shape: a node will not start on an unrebuilt ledger, it reports what to run, and the gate is on the data rather than on the build's features. The command is intended to be `snarkos developer rebuild-history <ledger-dir>` in a companion snarkOS PR; the driver stays here, where the finalize semantics and the test infrastructure are.

For testing before that lands, this carries a small binary: `cargo build --release --bin rebuild_db --features rebuild,history`. Deliberately a binary and not an example — an example is built with the crate's dev-dependencies unified into the graph, which would hand an operator a tool compiled with `snarkvm-synthesizer/test`: the wrong `Process` for the tip height, and confirmed transactions capped at eight. The `rebuild` feature turns on the synthesizer's rocks support from a real dependency, and pulls in `tracing-subscriber` optionally for the progress log. It drops out once snarkOS has the command.

## One thing that deliberately did not change

`snarkvm-ledger`'s `rocks` feature still does not enable `snarkvm-synthesizer/rocks`. Adding it there would flip `add_next_block_inner` onto its atomic-batch path for anyone building that crate on its own, which is unrelated to this work. The new opt-in `rebuild` feature pairs them instead. Real consumers are unaffected either way: the umbrella `snarkvm` crate's `rocks` already pairs the two.

## What a reviewer should attack

**The process is now built during the replay** (`VM::from_without_deployments`), and this commit is the one most in need of review — it landed after the review pass.

`VM::from` preloads every deployment at its latest edition and amendment count, so replaying an earlier deployment of a revised program checked it against the wrong program. The cluster confirmed this is not hypothetical: a real 20.8M-block mainnet archive carries `xqc_ccc.aleo` at edition 1, and the tool refused it. The process now starts with `credits.aleo` alone and the replay adds each program as it reaches that program's deployment, which is how a node builds it while syncing. After the change the same ledger reports `Nothing blocks a rebuild of this ledger`.

The guard changed shape with it. It no longer refuses any ledger holding a revised program — a blanket ban that would have excluded most of mainnet's future — but refuses a VM that holds programs loaded from storage, naming the constructor to use. Covered by `test_rebuild_replays_a_deployed_program`, which deploys a program, executes it twice and rebuilds through a non-preloading VM; the program's stack exists during that replay only because its own deployment was replayed first.

**Pre-V8 credits verifying keys.** `VM::from` selects `Process::load_v0()` or `load()` from the *tip* height, so replaying pre-V8 blocks runs with post-V8 keys. Finalize verifies no proofs, so this should not matter — but "should not" is why it is listed.

**Runtime.** Unmeasured; that is what the cluster test is for. The replay is inherently sequential — block `h` consumes the state left by `h - 1`, `atomic_finalize` is pinned to the sequential-ops thread, and `CommitteeStorage::insert` enforces `next_height == current_height + 1`. Block *loading* can be parallelised with a prefetch window, which is the obvious next step if I/O turns out to dominate; it is deliberately not in this PR, because correctness comes first and the measurement should tell us whether it is worth it.

**No `--check` mode yet.** The pre-flight completeness check covers the part that matters before anything is destroyed, but there is no read-only report.

## Verification

`cargo test -p snarkvm-ledger --features rocks,history,test --test history-rebuild` — 3 passed.

Each test winds the recorded schema version back to 0 first. Without that they are vacuous: a ledger built in a test is stamped at the current version as it is created, since it has no history and the startup gate records the version in passing, so the rebuild would take its "nothing to do" path and every assertion would hold over a chain nothing had touched. Each test also plants a mapping that no replay would recreate, so a rebuild that silently does not run fails the test instead of passing it.

- `test_rebuild_reproduces_history` — the rebuilt history matches the one the chain wrote, key for key and height for height, read back through the public path a node would serve from.
- `test_rebuild_resumes_after_interruption` — a rebuild stopped halfway resumes where it stopped rather than restarting or skipping.
- `test_rebuild_is_idempotent` — a finished rebuild is a no-op, not a second pass.

### Review round

A review pass found two defects worth naming here, both fixed:

- **The clear discarded rejection reasons and could never write them back.** They are keyed by transaction id rather than height, so the encoding fault never touched them, and `atomic_finalize` only writes one for a transaction id held in `pending_rejected_reasons` — which speculation populates and a replay never runs. They should not have been in the clear list at all.
- **The clear wiped the committee maps last, and the committee height is the resume point.** An interruption during the mapping-update clear — the longest phase, and the one followed by a synchronous compaction — left the mapping state gone while the committee still read as the tip. The next run resumed at `tip + 1`, replayed nothing, satisfied its own completeness check vacuously, and stamped the schema version over a destroyed ledger while reporting success. The committee maps now clear first, so any interruption resumes from genesis.

Also fixed: a second run on an already-rebuilt ledger discarded a healthy finalize state and replayed the whole chain; an upgraded program is now refused before anything is discarded rather than aborting the replay midway; the replay aborts the finalize batch when `finish_atomic` itself fails; and the schema-gate bypass no longer outlives the rebuild that needed it.

### On real ledgers

Run against the fixtures built for #3408, on the same m7i.4xlarge.

**The defect reproduces, and the rebuild repairs it.** On the genesis-synced three-format ledger (tip 77,672, complete from genesis), stock `snarkos-be` 4.9.1 reads `credits.aleo/bonded` for `aleo1vfukg8ky…` as frozen at 12,180,637,827 across heights 33,000–53,000. The same binary reads both sides of the experiment, since after a rebuild no heights rows remain and it takes the big-endian path.

After replaying all 77,672 blocks:

| height | before | after |
|---:|---|---|
| 30,000 | 11,959,892,352 | 11,959,892,352 |
| 40,000 | **12,180,637,827** | 15,260,647,937 |
| 50,000 | **12,180,637,827** | 18,573,197,004 |
| 55,000 | **12,180,637,827** | 20,238,778,565 |
| 60,000 | **12,180,637,827** | 21,892,412,609 |
| 70,000 | **12,180,637,827** | 25,199,566,409 |
| 77,000 | **12,180,637,827** | 27,520,361,383 |

The rebuilt series is monotone and lands against the live tip value of 27,732,555,245. The run took **2h17m for 77,672 blocks (9.45 blocks/s) at 875 MB peak RSS**, with no finalize-operation mismatch on any block.

(Raising `--rest-rps` matters when sampling many heights: the node's default of 10 requests per second otherwise throttles a tight query loop, and the rejections are indistinguishable from missing history in the v1 REST error format.)

**`--check` on the 20.8M-block archive**: `Blocks 20847647 (tip 20847646), history present, staking rewards absent` → `Nothing blocks a rebuild of this ledger`. It is complete from genesis, so it can be replayed.

**Opening a 1.7 TB ledger** costs 117 s and **10.3 GB peak RSS** before any work begins, against #3408's tool peaking at 1.12 GB for its entire migration. That tool worked on raw RocksDB with `max_open_files` bounded; this one goes through the typed store and block tree, and does not bound open files. Worth fixing before this is put in front of an operator.

**Throughput**: uncontended, the replay holds ~11 blocks/s on early-chain blocks, which would put 20.85M blocks near 24 days against the ~57-day genesis resync measured on the same hardware. That figure is a **ceiling, not a forecast** — early blocks are nearly empty and replay cost scales with transactions finalized. `perf` shows it hash-bound inside finalize (`keccakf` ~13%, `Fp256::mul_assign` ~11%), which is the same work a node does, so there is no pathology to remove and block prefetching would not help. A run at realistic load is in progress.

https://claude.ai/code/session_01E5mtJ1Qr73t8KRHZP8RBkA

### DIFF
--- a/ledger/Cargo.toml
+++ b/ledger/Cargo.toml
@@ -99,8 +99,7 @@ locktick = [
 metrics = [ "snarkvm-ledger-committee/metrics", "snarkvm-ledger-store/metrics" ]
 prop-tests = [ "snarkvm-ledger-committee/prop-tests" ]
 rocks = [ "snarkvm-ledger-store/rocks" ]
-# The offline finalize-state rebuild, and the tool that drives it. Opt-in and separate from `rocks`
-# so that enabling RocksDB storage does not also change `add_next_block`'s atomic-batch path for
+# Separate from `rocks`: pairing them there would change `add_next_block`'s atomic-batch path for
 # anyone building this crate on its own.
 rebuild = [ "rocks", "snarkvm-synthesizer/rocks", "dep:tracing-subscriber" ]
 serial = [
--- a/ledger/Cargo.toml
+++ b/ledger/Cargo.toml
@@ -16,10 +16,10 @@ categories = [ "cryptography", "web-programming" ]
 license = "Apache-2.0"
 edition = "2024"
 
-[[example]]
+[[bin]]
 name = "rebuild_db"
-path = "examples/rebuild_db.rs"
-required-features = [ "rocks" ]
+path = "src/bin/rebuild_db.rs"
+required-features = [ "rebuild" ]
 
 [[test]]
 name = "block-cache"
@@ -99,6 +99,10 @@ locktick = [
 metrics = [ "snarkvm-ledger-committee/metrics", "snarkvm-ledger-store/metrics" ]
 prop-tests = [ "snarkvm-ledger-committee/prop-tests" ]
 rocks = [ "snarkvm-ledger-store/rocks" ]
+# The offline finalize-state rebuild, and the tool that drives it. Opt-in and separate from `rocks`
+# so that enabling RocksDB storage does not also change `add_next_block`'s atomic-batch path for
+# anyone building this crate on its own.
+rebuild = [ "rocks", "snarkvm-synthesizer/rocks", "dep:tracing-subscriber" ]
 serial = [
   "snarkvm-console/serial",
   "snarkvm-ledger-authority/serial",
@@ -248,6 +252,11 @@ workspace = true
 [dev-dependencies.snarkvm-synthesizer]
 workspace = true
 features = [ "rocks", "test" ]
+
+[dependencies.tracing-subscriber]
+version = "0.3"
+features = [ "env-filter", "std" ]
+optional = true
 
 [dev-dependencies.tracing-subscriber]
 version = "0.3"
--- a/ledger/Cargo.toml
+++ b/ledger/Cargo.toml
@@ -29,7 +29,7 @@ required-features = [ "rocks" ]
 [[test]]
 name = "history-rebuild"
 path = "tests/history_rebuild.rs"
-required-features = [ "rocks", "history", "test" ]
+required-features = [ "rebuild", "history", "test" ]
 
 [[bench]]
 name = "block"
@@ -251,7 +251,7 @@ workspace = true
 
 [dev-dependencies.snarkvm-synthesizer]
 workspace = true
-features = [ "rocks", "test" ]
+features = [ "test" ]
 
 [dependencies.tracing-subscriber]
 version = "0.3"
--- a/ledger/Cargo.toml
+++ b/ledger/Cargo.toml
@@ -16,10 +16,20 @@ categories = [ "cryptography", "web-programming" ]
 license = "Apache-2.0"
 edition = "2024"
 
+[[example]]
+name = "rebuild_db"
+path = "examples/rebuild_db.rs"
+required-features = [ "rocks" ]
+
 [[test]]
 name = "block-cache"
 path = "tests/block_cache.rs"
 required-features = [ "rocks" ]
+
+[[test]]
+name = "history-rebuild"
+path = "tests/history_rebuild.rs"
+required-features = [ "rocks", "history", "test" ]
 
 [[bench]]
 name = "block"
@@ -237,7 +247,7 @@ workspace = true
 
 [dev-dependencies.snarkvm-synthesizer]
 workspace = true
-features = [ "test" ]
+features = [ "rocks", "test" ]
 
 [dev-dependencies.tracing-subscriber]
 version = "0.3"
--- a/ledger/examples/rebuild_db.rs
+++ b/ledger/examples/rebuild_db.rs
@@ -1,0 +1,99 @@
+// Copyright (c) 2019-2026 Provable Inc.
+// This file is part of the snarkVM library.
+
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at:
+
+// http://www.apache.org/licenses/LICENSE-2.0
+
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+//! Rebuilds a ledger's finalize state by replaying its blocks.
+//!
+//! ```text
+//! cargo build --release --example rebuild_db --features rocks,history
+//! ./target/release/examples/rebuild_db ~/.aleo/storage/ledger-0
+//! ```
+//!
+//! An example rather than a binary so that it can borrow the crate's existing
+//! `tracing-subscriber` dev-dependency instead of adding one, and because it is temporary: the
+//! operator-facing form of this is `snarkos developer rebuild-history`, which calls the same
+//! `VM::rebuild_finalize_state`. Nothing here is more than argument parsing and a log subscriber.
+//!
+//! The node must be stopped. RocksDB permits a single writer, so this refuses to open a ledger a
+//! node still holds, and the rebuild rewrites the state a running node would be reading.
+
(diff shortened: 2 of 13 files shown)


vicsn:

PR #3423, opened 2026-09-09
build: remove lld config

brunoffranca:
Removes the lld linker configuration from the cargo build settings. No longer necessary since lld became the default in Rust 1.90.0.

https://blog.rust-lang.org/2025/09/18/Rust-1.90.0/#lld-is-now-the-default-linker-on-x86-64-unknown-linux-gnu

Follow-up to: https://github.com/ProvableHQ/snarkOS/pull/4441


cbeck88 (approved):
thank you

PR #3425, opened 2026-09-09
hardening(curve): consult the sign flag when deserializing x = 0 on Edwards

cbeck88:
Split out of #3407, which bundled this with a Short Weierstrass fix that needs its own backwards-compatibility argument. #3407 is closed in favour of this pair; the Short Weierstrass half is #3422.

**This is the half that is not reachable from consensus.** The reachable one is #3422.

## The bug

A compressed twisted Edwards encoding is `x` plus a flag naming the sign of `y`. At `x = 0`, `y^2 = 1` has two solutions: the identity `(0, 1)` and the order-two point `(0, -1)`. The deserializer special-cased `x = 0` and returned the identity without ever reading the flag.

So the identity had two accepted encodings, and `(0, -1)` deserialized to the identity instead of to itself — a silently wrong value rather than an error. `Validate::Yes` catches neither: the identity is in the prime-order subgroup, and `(0, -1)` had already been replaced by the time the subgroup check ran.

## The fix

Delete the special case. `from_x_coordinate(0, greatest)` already returns `(0, 1)` for a clear flag and `(0, -1)` for a set one, so it was redundant as well as wrong. There is no `x` for which this newly returns `None`.

## Backwards compatibility

**No consensus object is encoded with this macro, so no historical bytes change meaning.** Aleo `group` values — every `Address`, record `nonce` and `group` literal — are encoded as the bare affine x-coordinate via `Group`'s `ToBytes`, with no flag byte, and decoded via `Group::from_x_coordinate`, which subgroup-checks the two candidates. `Group` has no `CanonicalSerialize` impl, no type deriving `CanonicalSerialize` holds a twisted Edwards point, and every `deserialize_compressed` call site in the repo is a Varuna or KZG structure over the pairing curve.

Within the library the encoder is unchanged and only ever emitted the clear flag at `x = 0`, so everything it has produced still decodes to the same point. Only the flag-set spelling changes meaning, and it was never emitted.

@Antonio95 asked on #3407 for that last claim to be checked rather than asserted:

```sh
git log --oneline -S 'mask |= 1 << 7' -- utilities/src/serialize/flags.rs
git log --oneline -S 'if x == P::BaseField::zero()' -- curves/src/templates/macros.rs
git log --oneline -S 'EdwardsFlags::from_y_sign' -- curves/src/templates/macros.rs
```

Each returns exactly one commit: `2e1312c12`, the initial commit. Bit layout, special case and writer are unchanged since the repository began.

The only commit that ever changed decoder behaviour is `e8ef1d2db` (May 2022), which narrowed `SWFlags::from_u8` from `(_, true) => Infinity` to `(true, true) => None`, commented *"we only want **one** way to serialize the point at infinity"*. Decoder narrowed, encoder untouched — the same move as this PR and as #3397.

The other half of @Antonio95's question — whether a point not produced by this library could be sitting in historical data — applies to the Short Weierstrass side, which is the side that reaches consensus. The backtest answering it is on #3422.

## Tests

Two tests in the generic `edwards_test<P>` harness, so they cover every twisted Edwards curve. Both written first, both fail on an unfixed tree.

- `edwards_non_canonical_identity_is_rejected` — asserts the canonical encoding still deserializes *first*, so a fix that rejected both would fail loudly rather than pass quietly; then that the flag-set encoding is not a second spelling of the identity. Runs under `Validate::No` as well, since the rejection must come from the encoding rather than the subgroup check.
- `edwards_order_two_point_round_trips` — the round-trip failure directly, both compression modes, under `Validate::No`.

Also corrects a doc comment from #3405 that this fix makes false.

`cargo test -p snarkvm-curves`: 65 passed, 0 failed.

## Left alone

The order-four points `(±i, 0)` accept either flag for the same reason. Not consensus-reachable, and `Validate::Yes` rejects them anyway for being outside the prime-order subgroup.

## Review history

The discussion on #3407 predates the split and quotes a body that described both halves at once. Nothing from it is lost: @Antonio95's encoding-history question is answered above, and the historical-data question is answered on #3422.

### DIFF
--- a/curves/src/templates/macros.rs
+++ b/curves/src/templates/macros.rs
@@ -248,11 +248,12 @@ macro_rules! impl_edwards_curve_serializer {
                 let point = if let Compress::Yes = compress {
                     let (x, flags): (P::BaseField, EdwardsFlags) = P::BaseField::deserialize_with_flags(&mut reader)?;
 
-                    if x == P::BaseField::zero() {
-                        Self::zero()
-                    } else {
-                        Affine::<P>::from_x_coordinate(x, flags.is_positive()).ok_or(SerializationError::InvalidData)?
-                    }
+                    // The flag names the sign of y, and it is consulted at every
+                    // x. At x = 0 it is the only thing separating the identity
+                    // (0, 1) from the order-two point (0, -1); reading the
+                    // identity out of x = 0 without looking at the flag would
+                    // give the identity a second encoding.
+                    Affine::<P>::from_x_coordinate(x, flags.is_positive()).ok_or(SerializationError::InvalidData)?
                 } else {
                     let x = P::BaseField::deserialize_uncompressed(&mut reader)?;
                     let y = P::BaseField::deserialize_uncompressed(&mut reader)?;
--- a/curves/src/templates/twisted_edwards_extended/tests.rs
+++ b/curves/src/templates/twisted_edwards_extended/tests.rs
@@ -53,6 +53,88 @@ where
     edwards_curve_serialization_test::<P>(rng);
     edwards_from_random_bytes::<P>(rng);
     edwards_from_x_and_y_coordinates::<P>(rng);
+    edwards_non_canonical_identity_is_rejected::<P>();
+    edwards_order_two_point_round_trips::<P>();
+}
+
+/// Returns `(0, -1)`, the order-two point.
+///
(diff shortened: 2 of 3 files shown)


Antonio95 (approved):
This seems correct. Thanks also for thoroughly checking my concerns (historical data, consensus reachability, etc.)

PR #3427, opened 2026-09-10
build: improve dev compile times

brunoffranca:
## Summary

Changes the dev profile configuration to improve compile times. Also removes profile configuration that is redundant.

Follow-up to https://github.com/ProvableHQ/snarkOS/pull/4445.

## Changes

- Changed the dev profile optimization level to 0 and the debug info to line tables only. Improved compile time for full builds by 39% (from 32.2s to 19.8s).
- Removed all configuration that was equal to the default profile values already. https://doc.rust-lang.org/cargo/reference/profiles.html#default-profiles

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -800,9 +800,9 @@ dependencies = [
 
 [[package]]
 name = "der"
-version = "0.8.0"
+version = "0.8.2"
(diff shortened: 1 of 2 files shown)


cbeck88 (approved):
thank you

PR #3428, opened 2026-09-11
[Feat] Increase maximum program mappings

raychu86:
## Motivation

This PR increases the maximum number of mappings a program can have.

- Convert Network::MAX_MAPPINGS into a versioned consensus configuration.
- Preserve maximum of 31 mappings through V20.
- Increase maximum to 128 mappings beginning with V21.
- Add LATEST_MAX_MAPPINGS() for parsing and program construction.
- Enforce active consensus limit during deployment verification.
- Extend consensus configuration and boundary test coverage.

## Test Plan

Tests have been added to enforce that the new maximum mappings value is properly enforced before and after V21.

## Backwards compatibility

New value will be enabled after `ConsensusVersion::V21`


vicsn (approved):
LGTM!

A quorum majority of honest validators will wait for the consensusversion activation before including this into their batch proposals.

PR #3431, opened 2026-09-13
chore: the licence and locktick checks are a test, not a build script

cbeck88:
## Motivation

`build.rs` walks the whole repository on every build of the root package, checking that each
`.rs` file carries the licence header and that lock imports have `locktick` counterparts. Both
are repository lints. Neither is a build step, and being one costs more than it looks.

**The script declares no `cargo:rerun-if-changed`.** With none, cargo fingerprints the package
by the newest mtime among its files, so `snarkvm` goes out of date whenever anything in the
tree is touched and the walk runs again — not once per push, but on every build of the root
package by anyone. That is the same shape as `curves/build.rs`, deleted in #3429.

It emits nothing besides: no `rustc-cfg`, no link directive, not one `cargo:` line. Nothing
reads it, so moving it changes no build.

## Why a test

A test is where this repository already keeps this kind of check:

- `synthesizer/tests/utilities/tests/mod.rs` already walks the tree with `WalkDir`, and
  `walkdir` is already a `dev-dependency` there.
- CI already runs the root package's tests through the existing `snarkvm` job, so **this adds
  no job**.
- `cargo test` still gives it to anyone working in the repository.

`walkdir` moves from `build-dependencies` to `dev-dependencies`. `Cargo.lock` is untouched.

## The diff is a rename

git renders it as `build.rs => tests/sources.rs`, and the only hunks are the `include_bytes!`
path and `main` becoming two `#[test]` entry points:

```
 Cargo.toml                   |  2 +-
 build.rs => tests/sources.rs | 16 +++++++++++-----
```

The walk, the constants, `DIRS_TO_SKIP`, the licence-year check, the lock-balance arithmetic
and both error messages are unchanged — so "is this the same check?" is answerable by
inspection, which seemed the thing worth optimising for in a lint move.

## Checked

```
cargo test -p snarkvm --test sources
  test licenses_are_present ... ok
  test locks_have_locktick_counterparts ... ok
  test result: ok. 2 passed; finished in 0.02s
```

`cargo clippy -p snarkvm --test sources -- -D warnings` clean, `cargo +nightly fmt --all --check`
clean.

I also ran the original `build.rs` as a standalone binary against the same tree to confirm the
behaviour did not drift while moving it — both pass a clean tree, both flag a mangled header,
both flag `use parking_lot::Mutex;` and `use tokio::sync::RwLock;`, and both stay silent on
`use parking_lot::MutexGuard;` and on a balanced `parking_lot` + `locktick` pair.

## The trade

A missing header now fails in a test run rather than at the desk of whoever next builds the
root package. That is where `check-fmt` and `check-clippy` already put this class of feedback.

---

An earlier revision of this PR did the same thing with a Python script and a new CircleCI job;
this is smaller, adds no job, and keeps the check in the language the repository actually uses
for it. Supersedes #3430.


vicsn (approved):
Just to confirm, primary motivation is to speed up local development? Makes sense!

PR #3432, opened 2026-09-14
WIP storage version, and dropping broken history feature flag.

cbeck88:
## Motivation

Following up on decision to drop the `history` feature, and make `history-staking-rewards` a runtime flag not a build time flag. (See release notes for 4.10.x, and issues #3404 , #3408 , #3417 )

(Note: I'm not yet convinced that history staking rewards is not broken, this is never tested in CI. If it's found to be badly broken then we should remove it also in this commit.)

This adds a storage version table to the database so that we can create a system of database migrations for future changes to the database schema and layout.

## Test Plan

This is mostly deletion, but the storage version stuff needs to have unit test coverage.

## Documentation

The snarkOS companion PR needs to explain how a snarkOS node operator performs DB migrations, but this PR itself doesn't need documentation update. We will also need to document how to use history-staking-rewards after it stops being a build flag.

## Backwards compatibility

We're explicitly dropping a broken feature. We're otherwise backwards compatible, modulo changes to the `history-staking-rewards` feature flag.


ledger/store/src/helpers/rocksdb/internal/schema.rs
```diff
@@ -0,0 +1,379 @@
...
+}
+
+/// Every migration, in order. The `from` versions must be `0, 1, ..., STORAGE_VERSION - 1`.
+const fn migrations() -> [Migration; STORAGE_VERSION as usize] {
+    [Migration {
+        from: 0,
+        description: "drop the tables of the removed `history` feature",
+        has_work: drop_history_tables::has_work,
+        apply: drop_history_tables::apply,
+    }]
```
ljedrz, line 89: if possible, this should be a one-liner sourced from `drop_history_tables`

ledger/store/src/helpers/rocksdb/internal/schema.rs
```diff
@@ -0,0 +1,379 @@
...
+    prefix[2..].copy_from_slice(&u16::from(map_id).to_le_bytes());
+    prefix
+}
+
+/// Returns the half-open raw key range `[from, to)` that contains exactly the keys starting with
+/// the given prefix.
+///
+/// `to` is the byte-wise successor of the prefix, which is the right bound regardless of how the
+/// prefix's fields are encoded (a `u16` map ID is little-endian, so `id + 1` would not do).
+fn prefix_range(prefix: &[u8; PREFIX_LEN]) -> (Vec<u8>, Vec<u8>) {
```
ljedrz, line 138: no need for vectors here

ledger/store/src/helpers/rocksdb/internal/schema.rs
```diff
@@ -0,0 +1,379 @@
...
+        } else {
+            *byte += 1;
+            break;
+        }
+    }
+    (prefix.to_vec(), to)
+}
+
+/// Returns whether no key starts with the given prefix.
+fn is_prefix_empty(database: &rocksdb::DB, prefix: &[u8; PREFIX_LEN]) -> Result<bool> {
```
ljedrz, line 154: nit: I'd name this `is_prefix_present`, as an existing prefix is never empty

ljedrz (commented):
Solid approach structure-wise, left some minor comments.

ledger/store/src/program/finalize.rs
```diff
@@ -802,6 +657,19 @@ impl<N: Network, P: FinalizeStorage<N>> FinalizeStore<N, P> {
         &self.block_height
     }
 
+    /// Sets whether each block's staking rewards are recorded in the staking rewards map.
+    ///
+    /// Shared by every clone of this store. Recording only covers blocks finalized while the
+    /// setting is on; earlier heights have no rows, and turning it off leaves a gap.
+    pub fn set_record_staking_rewards(&self, record: bool) {
```
vicsn, line 664: Would it make sense to enforce that this can only be set if staking rewards data is present for the current tip (with an exception for if we start from genesis)? That would ensure there are no holes in the staking rewards data.

synthesizer/process/src/view.rs
```diff
@@ -28,45 +27,36 @@ use snarkvm_synthesizer_program::{
...
 ///
-/// Evaluates whatever `stack` it is given; the caller must supply the stack for the edition live
-/// at `height`. Prefer `VM::evaluate_view_at_height`, which resolves the edition.
-///
-/// Available only with `--features history`.
-#[cfg(feature = "history")]
-pub fn evaluate_view_with_stack_at_height<N: Network, P: FinalizeStorage<N>>(
+/// Evaluates whatever `stack` it is given; the caller must supply the stack for the program's
+/// current edition. Prefer `VM::evaluate_view`, which resolves the stack.
+pub fn evaluate_view_with_stack<N: Network, P: FinalizeStorage<N>>(
```
vicsn, line 34: Just for your awareness, this can technically return non-atomic data, i.e. a function might be evaluated partially on state at height X and partially on state at height X+1. There were a few debates on this in the PRs introducing the view functions. snarkOS currently exposes `evaluate_view_latest` which is prone to this issue.

AFAIK the only way to avoid this in our current architecture would be to hold a `process.lock` during evaluation, but then a node which is hammed with a few `/view/` requests will incur performance hits during syncing.

I would personally be fine if we document this issue in the comments and in a github issue, but leave it be for now, because for most usecases in e.g. Ethereum people `view` static/const data. It is future work for our VM or Leo to support const view functions.

vicsn:

PR #3435, opened 2026-09-16
fix(ledger/query): bound `RestQuery`, reuse its connection, retry a lost request, skip the empty state-path query

cbeck88:
## What was broken

- `RestQuery` set no timeout, and ureq sets none by default (`Timeouts::default()` leaves `global`, `connect`, `recv_response` and `recv_body` all `None`). A node that accepts the connection and then goes silent blocks the calling thread indefinitely, and a thread blocked in a synchronous read cannot be interrupted from outside. The async path (`reqwest::get`) had the same gap.
- Every query built a fresh `ureq::Agent` / `reqwest::Client`, so each request paid for a new pool and, on HTTPS, a new TLS handshake.
- `get_state_paths_for_commitments` sent `statePaths?commitments=` for an empty slice. Every execution with no record inputs (every public transfer) reaches it that way, and at least one public endpoint answers the empty query with a 502, which turns every public execution into a failed `Trace::prepare`.

## What changes

1. **Skip the empty query.** Zero commitments returns an empty `Vec` without a request, sync and async. `Query::get_state_paths_for_commitments` already had this guard for the sync arm; the async wrapper now has it too.
2. **Three bounds, both paths.** Defaults: connect 10s, stall 30s, total 300s. `stall` is `timeout_recv_body` / reqwest `read_timeout`, which resets on every successful read, so it ends a response that has stopped moving without capping how large a response may be. `total` is the only bound that catches a peer which takes the request and never sends anything (there is no gap between reads to measure). 300s is sized for the largest response these routes return over a slow link. All three are settable through `RestQuery::with_timeouts(connect, stall, total)`.
   - The first bounds commit used `timeout_recv_response`; the second replaces it, because ureq keeps checking the `RecvResponse` deadline while the body is read, so on a GET it is an absolute cap on the whole response, not a first-byte bound. The commit message has the measurement.
3. **Hold the client.** `RestQuery` owns a `ureq::Agent` (cheap to clone, shares an `Arc<ConnectionPool>`) built in `From<http::Uri>`. `reqwest::Client` is fallible to build, so it lives in a `OnceLock` shared by clones and is built on first use. The bounds move onto the agent config, and the async client reads them back from there so they live in one place.
4. **Keep the wasm build working.** reqwest's wasm `ClientBuilder` has no `connect_timeout` / `read_timeout`, so those two sit under `cfg(not(target_arch = "wasm32"))`; without that, `cargo check -p snarkvm-wasm --target wasm32-unknown-unknown` fails with E0599 after the client change (verified both ways on this branch). The total bound goes on each request via `RequestBuilder::timeout`, which both backends have, so a wasm build is bounded too.
5. **Retry once a request lost on a pooled connection, on both paths.** A peer can close a pooled connection between two queries (ureq's `max_idle_age` is 15s) and neither client retries that: ureq has no retry at all, and hyper-util's `retry_canceled_requests` only covers a request it never began writing, so one written to a socket the peer had closed surfaces as `IncompleteMessage`. Sync retries on `ureq::Error::Io`; async on a request error that is neither a timeout nor a connect failure. Only the call is retried, never the body read, never on a timeout, and the second attempt gets only what is left of `total`, so the bound is paid once. Off wasm only.

## Tests

Six new tests bind a `127.0.0.1` listener each and drive a stub server: a node that never answers, a node that stops mid-answer, a timed-out query is not retried, a query lost on a pooled connection is retried, a query lost late is retried within the total bound, and two queries share one accepted connection. All use short bounds via `bounded_query` so a regression fails in seconds rather than hanging. `cargo test -p snarkvm-ledger-query --all-features`: 15/15. The async retry was verified the same way with a temporary tokio dev-dependency (not included): without it the second query on an abandoned pooled connection fails. `cargo clippy -p snarkvm-ledger-query --all-targets --all-features -D warnings` and `cargo +nightly fmt --check` clean.

## Behaviour change for callers

Any `RestQuery` consumer (snarkOS client paths) now gets finite waits by default. A LAN validator and a proving service reaching across the internet may want different numbers; `with_timeouts` is there for that. One asymmetry, documented on `with_timeouts`: reqwest's `read_timeout` also covers the wait for the first byte, ureq's `timeout_recv_body` does not, so a node that never answers fails after `stall` on the async path and after `total` on the sync path.

Also on ureq 3.3, `connect` covers the TCP connect only; the TLS handshake runs in the send phase, where `total` is the bound that applies.

Note on ureq versions: the stall bound relies on ureq 3.3.0's `timeout_recv_body` resetting per read. ureq 3.4.1 (#1194) changed it to a total budget for the body; the lock pins 3.3.0, so this is correct as merged, but a bump to 3.4 should revisit the stall semantics.

The parameter-download side of the same problem (`parameters/src/macros.rs`) is a separate PR.


ledger/query/src/query/rest.rs
```diff
@@ -27,18 +27,108 @@ use anyhow::{Context, Result, anyhow, bail, ensure};
...
+/// Resets on every successful read, so it bounds a peer that stops mid-answer
+/// without capping how large an answer may be.
+const DEFAULT_STALL_TIMEOUT: Duration = Duration::from_secs(30);
+
+/// How long a whole request may take.
+///
+/// The only bound that covers a peer which accepts the connection, takes the
+/// request and then sends nothing at all: the stall bound above measures gaps
+/// between reads, and a response that never starts has none.
+const DEFAULT_TOTAL_TIMEOUT: Duration = Duration::from_secs(300);
```
cbeck88, line 41: previously there were no timeouts on these at all, i looked at Reqwest defaults and made it a little more generous

vicsn (commented):
@iamalwaysuncomfortable let us know what yoh think, and can you trigger sdk CI on this snarkVM branch?

PR #3438, opened 2026-09-17
Prevent off-thread speculative reads from stalling atomic_speculate

vicsn:
## Motivation

Fleet metrics from #3415 / #3416 showed 21–23 of 30 validators stalling together for ~50s, with `atomic_speculate` tailing to 25s while the thread doing that work used ~0.0001 cores. That is a blocked thread, not rayon/CPU saturation.

`check_transaction` public-fee checks were calling `get_value_speculative` on `credits.aleo/account`. That takes the per-map `atomic_batch` mutex and scans the pending dry-run. `atomic_speculate` uses the same mutex to enqueue finalize writes. `parking_lot::Mutex` is unfair, so a broadcast burst of fee checks can starve the sequential speculate thread across most of the fleet at once.

This PR:
- Consults the pending atomic batch only on the thread that started it; other threads observe confirmed state.
- Checks public-fee balances against confirmed storage.
- Adds `ATOMIC_BATCH_LOCK_WAIT_SECONDS` and `ATOMIC_BATCH_OFF_THREAD_SPECULATIVE_READ_TOTAL`.

Related: #3415, #3416.

## Test Plan

- [x] `cargo test -p snarkvm-ledger-store --lib -- speculative_reads`
- [x] `test_speculative_reads_from_other_threads_ignore_atomic_batch` — off-thread readers do not observe pending inserts
- [x] `test_off_thread_speculative_reads_do_not_block_atomic_writer` — 8 readers scanning a 20k-entry batch while the owner inserts 5k more
- [x] `test_pregenerated_public_fee_checks_do_not_stall_atomic_speculate` — 64 TestnetV0 executions from the local zip, concurrent speculative fee-balance reads vs `atomic_speculate` (skips if the zip is absent)

Contention-test wall time on this machine (`optimized + debuginfo`):

| | writer elapsed |
|---|---|
| before (3 runs) | 71ms, **4.44s**, **4.72s** |
| after (3 runs) | **164µs, 192µs, 174µs** |

Typical speedup is about **25,000×** (4.5s → 170µs). The 71ms run is the unfair mutex getting lucky; the 4.5s runs match a blocked sequential thread.

## Documentation

No docs changes. The new metrics are for overlaying with `atomic_speculate` latency in snarkOS.

## Backwards compatibility

No consensus fork. Off-thread speculative reads now see committed state instead of an in-progress dry-run, which is the correct isolation for admission checks. Same-thread dry-run behavior is unchanged.


ljedrz (approved):
Not only is it a great perf win, the new logic is quite straightforward. LGTM!

ledger/store/src/helpers/mod.rs
```diff
@@ -23,6 +23,77 @@ pub(crate) mod test_helpers;
...
+    static NEXT_THREAD_KEY: AtomicU64 = AtomicU64::new(1);
+
+    /// Returns a stable nonzero identifier for the current thread.
+    pub(crate) fn current_thread_key() -> u64 {
+        THREAD_KEY.with(|cell| {
+            let mut key = cell.get();
+            if key == 0 {
+                key = NEXT_THREAD_KEY.fetch_add(1, Ordering::Relaxed);
+                // `0` is reserved for "no owner".
+                if key == 0 {
```
ljedrz, line 46: love this one 😁

synthesizer/src/vm/finalize.rs
```diff
@@ -4191,4 +4191,90 @@ finalize compute:
...
+        use std::{
+            process::Command,
+            sync::{
+                Arc,
+                atomic::{AtomicBool, Ordering},
+            },
+            time::{Duration, Instant},
+        };
+
+        let zip = std::path::PathBuf::from(env!("CARGO_MANIFEST_DIR")).join(
```
vicsn: Todo: remove this hardcoded file

PR #3439, opened 2026-09-18
Introduce dev_skip_state_root_check

vicsn:
## Motivation

It is useful for testing purposes to pregenerate a lot of transactions once, and to be able to re-use and submit them against a variety of test networks of different committee sizes. However, snarkOS built with `--features test_network` will create a unique genesis block with the given `--dev-num-validators` and a genesis state root which is unique to that committee size. Acceptance of pregenerated transactions is tied to a specific `global_state_root`, and therefore limited to the particular committee size. 

There are a couple of alternative approaches we could take in order to re-use pregenerated transactions:
1. Introduce a dev feature flag to skip state root checks (this PR + https://github.com/ProvableHQ/snarkOS/pull/4463) was the quickest and easiest approach.
2. Introduce a dev feature flag to set the `test_network` genesis state root to a hardcoded value in `BlockStore::insert`. This is somewhat cleaner than 1. but doesn't allow me to re-use transactions I already pregenerated in the past.
3. Adjusting `test_network` to always automatically generate a small 4-validator genesis block, and then bond-in to `--dev-num-validators`. This will require at least two blocks (one to send transactions to would-be-validators, one to bond them in) and new transaction pregeneration logic specifically for this purpose.
4. Adjusting our stress-tests always generate a small 4-validator genesis block, and then bond-in `--dev-num-validators`.

## Note on the naming

Currently our development feature flags use a combination of `test` v.s. `dev` and underscores v.s. dashes. We could clean this up separately one day, but happy to document a strong opinion now.

## Test Plan

- [x] `cargo test -p snarkvm-synthesizer --lib -- test_check_transaction_unknown_global_state_root` rejects an execution and a deployment with an unknown root (default)
- [x] same test with `--features dev_skip_state_root_check` accepts both

## Documentation

No docs changes. The feature is a compile-time opt-in for local/dev builds.

## Backwards compatibility

Default path is unchanged. The skip is gated behind `dev_skip_state_root_check` and is not a consensus change.


cbeck88:
I didn’t understand what’s wrong with 2, why can’t you use transactions you generated in the past? Or it’s just that, you want to use testnet or mainnet transactions for this in a dev network

cbeck88:
I see, because of the bonding?

cbeck88:
That seems like a good long term state — skipping the state root check like this seems mainly useful in performance tests. But then skipping required checks is undesirable.

Its also desirable to be able to submit lots of invalid transactions and test network performance under such pressure, so route 2->3->4 seems good to help that

vicsn:
I agree that 2. is the cleaner and more robust approach, and I'm glad you don't have the sunk cost I did to be blind to that. Closing this PR

PR #3440, opened 2026-09-18
Prevent off-thread speculative reads from stalling atomic_speculate

vicsn:
## Motivation

Testnet backport of the stall fix from #3438 (`b77715957f`).

Fleet metrics from #3415 / #3416 showed 21–23 of 30 validators stalling together for ~50s, with `atomic_speculate` tailing to 25s while the thread doing that work used ~0.0001 cores. That is a blocked thread, not rayon/CPU saturation.

`check_transaction` public-fee checks were calling `get_value_speculative` on `credits.aleo/account`. That takes the per-map `atomic_batch` mutex and scans the pending dry-run. `atomic_speculate` uses the same mutex to enqueue finalize writes. `parking_lot::Mutex` is unfair, so a broadcast burst of fee checks can starve the sequential speculate thread across most of the fleet at once.

This PR:
- Consults the pending atomic batch only on the thread that started it; other threads observe confirmed state.
- Checks public-fee balances against confirmed storage.
- Adds `ATOMIC_BATCH_LOCK_WAIT_SECONDS` and `ATOMIC_BATCH_OFF_THREAD_SPECULATIVE_READ_TOTAL`.

Also bumps `.cargo/release-version` to `v4.10.1`.

Related: #3415, #3416, #3438.

## Test Plan

- [x] `cargo test -p snarkvm-ledger-store --lib -- speculative_reads`
- [x] `test_speculative_reads_from_other_threads_ignore_atomic_batch` — off-thread readers do not observe pending inserts
- [x] `test_off_thread_speculative_reads_do_not_block_atomic_writer` — 8 readers scanning a 20k-entry batch while the owner inserts 5k more

## Documentation

No docs changes. The new metrics are for overlaying with `atomic_speculate` latency in snarkOS.

## Backwards compatibility

No consensus fork. Off-thread speculative reads now see committed state instead of an in-progress dry-run, which is the correct isolation for admission checks. Same-thread dry-run behavior is unchanged.


vicsn:
Companion snarkOS draft: https://github.com/ProvableHQ/snarkOS/pull/4465

PR #3441, opened 2026-09-18
Cut sequential-thread occupancy for self-constructed blocks

vicsn:
## Summary
- Self-constructed blocks occupied the sequential VM thread for two dry-runs plus a RealRun. Construct-path speculate now keeps the finalize batch, check skips a second dry-run when the block hash matches, and `add_next_block` finishes that batch instead of replaying RealRun.
- Speculative reads use an IndexMap overlay beside the ordered write log so pending lookups stay O(1) as the batch grows.
- Sequential queue-wait is labeled by op, and speculate wall time is split into prepare / queue / inner.

Untrusted and sync blocks still dry-run. Speculate of N+1 still cannot overlap apply of N.

## Test plan
- [x] `cargo test -p snarkvm-ledger --lib -- test_self_constructed_check_and_advance`
- [x] `cargo test -p snarkvm-ledger --features rocks --lib -- test_self_constructed_check_and_advance`
- [x] Overlay / speculative map tests (memory + rocks)
- [ ] Validator construct-then-advance: confirm sequential occupancy is one finalize pass, leftover keep is discarded if construct is abandoned


ledger/store/src/helpers/memory/internal/map.rs
```diff
@@ -50,6 +50,8 @@ pub struct MemoryMap<
     /// Owner of the in-progress atomic batch (`0` = none).
     atomic_owner: Arc<AtomicU64>,
     atomic_batch: Arc<Mutex<Vec<(K, Option<V>)>>>,
+    /// Latest pending value per key, rebuilt from `atomic_batch` on rewind.
+    pending: Arc<Mutex<IndexMap<K, Option<V>>>>,
```
ljedrz: couldn't this just be `IndexMap<K, usize>` to avoid the duplication of values?

ledger/store/src/helpers/rocksdb/internal/nested_map.rs
```diff
@@ -505,48 +514,27 @@ impl<
...
+        // If there is no atomic batch in progress on this thread, then return the confirmed key-value pairs.
         if !crate::helpers::atomic_owner::consults_atomic_batch(self.is_atomic_in_progress(), &self.atomic_owner) {
-            return Ok(key_values);
+            return self.get_map_confirmed(map);
         }
 
-        // Retrieve the atomic batch.
         let start = Instant::now();
-        let operations = self.atomic_batch.lock().clone();
+        let pending = self.pending.lock();
```
ljedrz: why do we want to prolong the guard here? this method (`get_map_speculative`) is used in read-only contexts too, and the contents can't change in the meantime

ledger/store/src/helpers/memory/internal/map.rs
ljedrz: also, since all atomic speculation happens sequentially in a dedicated thread (`ensure_sequential_processing`), it could be sharing the same mutex as `atomic_batch`

ledger/store/src/helpers/rocksdb/internal/map.rs
```diff
@@ -401,11 +412,10 @@ impl<
     {
         // Return early if there is no atomic batch in progress on this thread.
         if crate::helpers::atomic_owner::consults_atomic_batch(self.is_atomic_in_progress(), &self.atomic_owner) {
-            // We iterate from the back of the `atomic_batch` to find the latest value.
             let start = Instant::now();
-            let batch = self.atomic_batch.lock();
+            let pending = self.pending.lock();
```
ljedrz: same comment as https://github.com/ProvableHQ/snarkVM/pull/3441/changes#r4046961622

ledger/store/src/helpers/memory/internal/nested_map.rs
```diff
@@ -50,6 +50,8 @@ pub struct NestedMemoryMap<
     /// Owner of the in-progress atomic batch (`0` = none).
     atomic_owner: Arc<AtomicU64>,
     atomic_batch: Arc<Mutex<Vec<(M, Option<K>, Option<V>)>>>,
+    /// Latest pending values, rebuilt from `atomic_batch` on rewind.
+    pending: Arc<Mutex<crate::helpers::pending_overlay::NestedPending<M, K, V>>>,
```
ljedrz: same comment as https://github.com/ProvableHQ/snarkVM/pull/3441/changes#r4046999377, and it applies to the RocksDB side as well

ledger/store/src/helpers/memory/internal/nested_map.rs
```diff
@@ -377,41 +387,22 @@ impl<
             return self.get_map_confirmed(map);
         }
 
-        // Retrieve the confirmed key-value pairs for the given map.
-        let mut key_values = self.get_map_confirmed(map)?;
+        let start = Instant::now();
+        let pending = self.pending.lock();
```
ljedrz: same comment as https://github.com/ProvableHQ/snarkVM/pull/3441#discussion_r4046961622

ledger/store/src/helpers/rocksdb/internal/nested_map.rs
```diff
@@ -505,48 +514,27 @@ impl<
...
-                            None => key_values.push((k, v)),
-                        }
-                    }
-                    // Clear the key-value pairs for the map.
-                    (None, None) => key_values.clear(),
-                    // Remove the key-value pair for the key.
-                    (Some(k), None) => key_values.retain(|(key, _)| key != &k),
-                    (None, Some(_)) => unreachable!("Cannot remove a key-value pair from a map without a key."),
-                }
+        for (k, v) in pending.entries_for_map(map) {
```
ljedrz: if I'm not mistaken, this may change the order of entries, which may be a compatibility issue; `self.atomic_batch` contains the full ordered log of insertions and deletions, while `self.pending` may remove an earlier entry or replace it with a newer value in place; if this method is used for any hash computation etc, it would require a consensus version bump

synthesizer/src/vm/mod.rs
```diff
@@ -630,13 +634,21 @@ impl<N: Network, C: ConsensusStorage<N>> VM<N, C> {
...
+            self.discard_kept_speculation_inner();
+            #[cfg(feature = "rocks")]
+            self.block_store().pause_atomic_writes()?;
+        }
 
         // First, insert the block.
         if let Err(insert_error) = self.block_store().insert(&block) {
-            if cfg!(feature = "rocks") {
+            if using_kept_batch {
+                self.finalize_store().abort_atomic();
```
ljedrz: for future-proofing, I'd promote this to something like
```
let mut kept_guard = using_kept_batch.then(|| defer(|| self.finalize_store().abort_atomic()));
```
which would be "disarmed" once `finish_atomic` succeeds

synthesizer/src/vm/mod.rs
```diff
@@ -702,6 +745,8 @@ impl<N: Network, C: ConsensusStorage<N>> Drop for VM<N, C> {
     fn drop(&mut self) {
         // Check if this the final external reference to `VM`.
         if Arc::strong_count(&self.sequential_ops_tx) == 1 {
+            // Abort any kept finalize batch before shutting down the sequential thread.
+            self.discard_kept_speculation();
```
ljedrz: note: this will cause the `VM` to no longer cleanly drop in async contexts; this is probably fine in the snarkOS implementation, but might break some tests there

ljedrz (commented):
Left a few comments.

ljedrz (commented):
It looks like most of the concerns have been mitigated, but I'll want to do one more pass to be sure.

ledger/store/src/helpers/mod.rs
```diff
@@ -94,6 +94,240 @@ pub(crate) mod atomic_owner {
...
+            V: Clone,
+        {
+            get_flat(&self.log, &self.pending, key)
+        }
+    }
+
+    /// Latest nested pending values, plus maps that were fully removed in the log.
+    pub(crate) struct NestedPending<M, K, V> {
+        /// Latest `(map, serialized-key)` write. `None` is a key deletion.
+        entries: IndexMap<(M, Vec<u8>), (K, Option<V>)>,
```
ljedrz:
```suggestion
        entries: IndexMap<(M, Vec<u8>), usize>,
```
just as it was done with the non-nested maps

synthesizer/src/vm/helpers/sequential_op.rs
```diff
@@ -83,12 +151,100 @@ impl<N: Network, C: ConsensusStorage<N>> VM<N, C> {
...
     AddNextBlock(Block<N>),
-    AtomicSpeculate(FinalizeGlobalState, i64, Option<u64>, Vec<Ratify<N>>, Solutions<N>, Vec<Transaction<N>>),
+    AtomicSpeculate {
+        state: FinalizeGlobalState,
+        time_since_last_block: i64,
+        coinbase_reward: Option<u64>,
+        ratifications: Vec<Ratify<N>>,
+        solutions: Solutions<N>,
+        transactions: Vec<Transaction<N>>,
+        keep: bool,
```
ljedrz:
```suggestion
        keep: Option<SpeculationId>,
```
I think that assigning this value sooner would result in less plumbing, and would provide us with a better insight into the order of events

ljedrz:
I think we should also test it with the `slipstream-plugins` feature on, this could impact it.

ljedrz (approved):
LGTM, though I'd remove the plugin in a dedicated PR.

vicsn:
Pregenerated prepare/check/advance bench (6000 TestnetV0 `transfer_public` executions, 30 blocks) ran on GitHub Actions against staging and this branch.

| Step | [staging baseline](https://github.com/ProvableHQ/snarkVM/actions/runs/35630583902) | [this PR](https://github.com/ProvableHQ/snarkVM/actions/runs/35631490201) | Change |
| --- | ---: | ---: | ---: |
| `prepare_advance_to_next_quorum_block` | 8.72 s | 9.32 s | +6.9% |
| `check_next_block` | 5.66 s | 3.79 s | −33% |
| `advance_to_next_block` | 2.18 s | 0.62 s | −71% |
| **sum** | **16.56 s** | **13.73 s** | **−17%** |

Fixture: [`ci-pregenerated-txs-testnet-40val-6000`](https://github.com/ProvableHQ/snarkVM/releases/tag/ci-pregenerated-txs-testnet-40val-6000) (same zip as `gs://provable-pregenerated-transactions/pregenerated_transactions/transactions-testnet-40val-10-22-9349107e39bb93a84f9a539602ae3cca1372557e_test_network-6000-20.zip`). Dispatch `Run snarkVM Benchmarks` on a branch to rerun; snarkVM does not have `GCLOUD_SERVICE_KEY`, so CI currently downloads the GitHub release when that secret is empty.

ledger/store/src/helpers/mod.rs
```diff
@@ -94,6 +94,251 @@ pub(crate) mod atomic_owner {
...
+pub(crate) mod pending_overlay {
+    use indexmap::{IndexMap, IndexSet};
+    use serde::Serialize;
+    use std::hash::Hash;
+
+    /// Rebuilds the latest pending log index per key from an ordered write log.
+    pub(crate) fn rebuild_flat<K: Clone + Eq + Hash, V>(log: &[(K, Option<V>)]) -> IndexMap<K, usize> {
+        let mut pending = IndexMap::with_capacity(log.len());
+        for (i, (key, _)) in log.iter().enumerate() {
+            pending.insert(key.clone(), i);
```
vicsn: Are duplicate inserts a problem?

ledger/store/src/helpers/mod.rs
```diff
@@ -94,6 +94,251 @@ pub(crate) mod atomic_owner {
...
+            self.log.is_empty()
+        }
+
+        pub(crate) fn clear(&mut self) {
+            self.log.clear();
+            self.pending.clear();
+        }
+
+        pub(crate) fn push(&mut self, key: K, value: Option<V>) {
+            self.pending.insert(key.clone(), self.log.len());
```
vicsn: Are duplicate inserts a problem?

synthesizer/src/vm/helpers/sequential_op.rs
```diff
@@ -83,12 +151,104 @@ impl<N: Network, C: ConsensusStorage<N>> VM<N, C> {
...
+        *self.self_constructed.lock() =
+            Some(SelfConstructed { id, hash: None, finalize_operations, parked_stacks, batch_kept });
+    }
+
+    /// Associates a constructed block hash with the matching construct-path speculate.
+    pub fn bind_self_constructed_hash(&self, id: SpeculationId, hash: N::BlockHash) {
+        if let Some(constructed) = self.self_constructed.lock().as_mut()
+            && constructed.id == id
+        {
+            constructed.hash = Some(hash);
```
vicsn: Can this be overwritten? Should we panic if it already exists?

utilities/src/defer.rs
```diff
@@ -22,6 +22,11 @@ impl<F: FnOnce()> Defer<F> {
     pub fn new(f: F) -> Self {
         Defer { f: Some(f) }
     }
+
+    /// Prevents the deferred closure from running on drop.
+    pub fn disarm(mut self) {
```
vicsn: Is this extra complexity really needed?

synthesizer/process/src/lib.rs
```diff
@@ -184,15 +184,35 @@ impl<'a, N: Network> ProcessExclusiveGuard<'a, N> {
     /// This will remove the new stacks and restore the old stacks.
     #[inline]
     pub fn revert_stacks(&self) {
-        // Restore the old stacks.
+        let _ = self.park_staged_stacks();
+    }
+
+    /// Reverts staged stacks and returns the new stacks so they can be restored later.
+    #[inline]
+    pub fn park_staged_stacks(&self) -> IndexMap<ProgramID<N>, Arc<Stack<N>>> {
```
vicsn: Why the extra abstraction and complexity?

vicsn:
Let's merge https://github.com/ProvableHQ/snarkVM/pull/3439 first so we're sure the benchmark logic is acceptable

synthesizer/src/vm/mod.rs
```diff
@@ -641,6 +636,7 @@ impl<N: Network, C: ConsensusStorage<N>> VM<N, C> {
         // instead of pausing and replaying RealRun.
```
vicsn: So we still need/use DryRun and RealRun?

vicsn:
Re-benched `4e0a6dd` against a fresh baseline. **On the current one-block workload this branch is 21% slower overall**, which contradicts the −17% posted earlier.

| Step | Baseline | This PR | Change |
| --- | --- | --- | --- |
| `prepare_advance_to_next_quorum_block` | 174.00 ms | 285.64 ms | **+64.2%** |
| `check_next_block` | 122.49 ms | 114.21 ms | −6.8% |
| `advance_to_next_block` | 45.63 ms | 13.53 ms | **−70.3%** |
| **sum** | **342.13 ms** | **413.38 ms** | **+20.8%** |

[PR run](https://github.com/ProvableHQ/snarkVM/actions/runs/35742318253) · [baseline run](https://github.com/ProvableHQ/snarkVM/actions/runs/35742672049)

### Why the earlier table isn't comparable

The bench changed between the two measurements. `PREGENERATED_TX_LIMIT=0` used to mean "all 6000 executions" across 30 blocks; on `4e0a6dd` it means **one full block** (`MAX_TRANSMISSIONS_PER_BATCH * PREGENERATED_NUM_VALIDATORS`, 200 transmissions). `perf/pregenerated-bench-baseline` still carries the old harness, so it can no longer serve as a control.

I pushed `perf/pregenerated-bench-baseline-v2` = `staging` + a cherry-pick of `2a7e8ff` (the bench commit only, no optimization commits). That commit is self-contained — it brings `dev_skip_state_root_check` and the bench declaration with it — so the two runs differ only by the four optimization commits.

### Reading it

`advance_to_next_block` is down 70%, which is the change doing what it set out to do: the RealRun replay is gone. The problem is `prepare_advance_to_next_quorum_block` at +64%, and it more than eats the gain. Keeping the finalize batch open across construct is evidently not free, and on a single block there are no later blocks to amortise it against.

That also explains the divergence from the 30-block result: there, `check_next_block` saved 33% and the prepare overhead was 6.9%, so the arithmetic went the other way. Whether the branch is a win therefore depends on which workload is representative — the one-block figure is the pessimistic end.

### Caveats

Both numbers are **single iterations** (`+/- 0`, no repetition) on shared `ubuntu-24.04` runners, so treat ±10% as noise. A 21% gap is probably outside that, but a +64% regression on one step is worth reproducing before acting on it. The two runs did land on different runners minutes apart.

Worth deciding before merge: whether to re-point the bench at multiple blocks so prepare-side overhead is amortised the way it would be in production, since that is the regime the PR is aimed at.

PR #3444, opened 2026-09-22
fix(vm): release sequential worker after final external clone

paulbalaji:
The sequential worker retains a cloned VM containing the sender on which it waits. Dropping every external VM therefore leaves the worker and its Process alive. Checking `Arc::strong_count` in each VM destructor also cannot reliably identify the final owner when clones drop concurrently.

External VM clones now share an Arc-owned queue; the worker owns no queue sender. Final queue drop closes the sender and joins the worker once. Thread identity is stored independently so queued operations can still check their execution thread during shutdown. A worker panic is logged during cleanup rather than propagated from Drop, including when cleanup runs during another unwind.

Validation: **7 regression tests passed**—Mainnet/Testnet/Canary lifetimes, concurrent final-clone drops, successful queued-operation drain during concurrent final-clone drops, worker panic cleanup, and cleanup during an existing unwind. The drain test holds the Process lock while queuing eight operations and preparing four final clones for concurrent drop, then checks every successful response and Process release. The lifetime tests observe weak Process references without sleeps; all four failed on the original v4.8.1 dependency.

This patch is based on v4.8.1, which Hyperlane currently consumes. Public VM methods and execution/consensus code are unchanged. Final external drop waits for queued operations, as the original destructor intended. The exact patch is also mirrored under [hyperlane-xyz/snarkVM](https://github.com/hyperlane-xyz/snarkVM/tree/hyperlane-v4.8.1-vm-lifecycle.1).

Local commands: `cargo test -p snarkvm-synthesizer --test vm_lifecycle -- --test-threads=1` and `cargo test -p snarkvm-synthesizer --lib vm::helpers::sequential_op::tests` (test profile: opt-level 0, no LTO or debug symbols).


ljedrz (approved):
LGTM as a hardening 👍.

